PR #517, opened 2024-01-21
Enable cross_entropy loss to handle dense targets

AtomicVar:
## Proposed changes

Dense targets means probabilities or one-hot encodings.

Currently, the `cross_entropy` loss only supports targets to be class indices. However, in PyTorch, the [`CrossEntropyLoss`](https://pytorch.org/docs/stable/generated/torch.nn.CrossEntropyLoss.html) supports targets to be both class indices and probabilities (or one-hot encodings). In TensorFlow, two seperate cross-entropy loss layers are provided: [`CategoricalCrossentropy`](https://www.tensorflow.org/api_docs/python/tf/keras/losses/CategoricalCrossentropy) and [`SparseCategoricalCrossentropy`](https://www.tensorflow.org/api_docs/python/tf/keras/losses/SparseCategoricalCrossentropy), to support two different types of targets.

This PR enhances the `cross_entropy` loss to support both sparse targets (class indices) and dense targets (probabilities or one-hot encodings). It can infer what type of targets from the shape, like PyTorch does.

## Checklist

### DIFF
--- a/python/mlx/nn/losses.py
+++ b/python/mlx/nn/losses.py
@@ -31,12 +31,13 @@ def cross_entropy(
     Computes the cross entropy loss.
 
     Args:
-        logits (array): The unnormalized predicted logits.
-        targets (array): The ground truth values. These can be class indices or probabilities for
-            each class, depending on its shape. If ``targets`` are class indices, then the shape of
-            ``targets`` should match the shape of ``logits``, excluding the ``axis`` dimension.
-            If ``targets`` are probabilities (or one-hot encoded), then the shape of ``targets``
-            should be identical to the shape of ``logits``.
+        logits (array): The unnormalized logits.
+        targets (array): The ground truth values. These can be class indices or
+            probabilities for each class. If the ``targets`` are class indices,
+            then ``targets`` shape should match the ``logits`` shape with
+            the ``axis`` dimension removed. If the ``targets`` are probabilities
+            (or one-hot encoded), then the ``targets`` shape should be the same as
+            the ``logits`` shape.
         weights (array, optional): Optional weights for each target. Default: ``None``.
         axis (int, optional): The axis over which to compute softmax. Default: ``-1``.
         label_smoothing (float, optional): Label smoothing factor. Default: ``0``.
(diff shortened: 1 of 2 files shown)


python/mlx/nn/losses.py
```diff
@@ -32,20 +32,59 @@ def cross_entropy(
...
+        >>> logits = mx.array([[2.0, -1.0], [-1.0, 2.0]])
+        >>> targets = mx.array([0, 1])
+        >>> cross_entropy(logits, targets)
+        array([0.0485873, 0.0485873], dtype=float32)
+
+        >>> # Probabilities (or one-hot vectors) as targets
+        >>> logits = mx.array([[2.0, -1.0], [-1.0, 2.0]])
+        >>> targets = mx.array([[0.9, 0.1], [0.1, 0.9]])
+        >>> cross_entropy(logits, targets)
+        array([0.348587, 0.348587], dtype=float32)
```
awni, line 63: ❤️

python/mlx/nn/losses.py
```diff
@@ -62,7 +101,9 @@ def cross_entropy(
 
     # Apply weights if provided
     if weights is not None:
-        if weights.shape != targets.shape:
+        if (targets_as_probs and weights.shape != _drop_dim(targets.shape, axis)) or (
+            not targets_as_probs and weights.shape != targets.shape
+        ):
```
awni, line 106: I think you could simplify this and check `weights.shape == loss.shape` ?

python/mlx/nn/losses.py
```diff
@@ -62,7 +101,9 @@ def cross_entropy(
...
             raise ValueError(
                 f"Weights with shape {weights.shape} is not the same as "
                 f"targets with shape {targets.shape}."
```
awni, line 109: Please update this error message since now the weight should have the same shape as the loss but not necessarily the target.

python/mlx/nn/losses.py
```diff
@@ -32,20 +32,59 @@ def cross_entropy(
...
+    # Check shapes in two cases: targets as class indices and targets as probabilities
+    if (targets_as_probs and targets.shape != logits.shape) or (
+        not targets_as_probs and targets.shape != _drop_dim(logits.shape, axis)
+    ):
+        raise ValueError(
+            f"Targets shape {targets.shape} does not match logits shape {logits.shape}."
+        )
+
+    if targets_as_probs:
+        score = mx.sum(logits * targets, axis=axis)  # mx.inner doesn't support axis
```
awni, line 84: nit: `mx.inner` is unlikely to support axis since numpy is the same, so the comment is not necessary

awni: You could (and probably should) use `mx.tensordot` though.

awni (changes requested):
Thanks for adding this! I left just a few comments, please address then we can merge.

python/mlx/nn/losses.py
AtomicVar: Thanks for your advice, but I cannot figure out how to use `mx.tensordot` to do `mx.sum(a * b, axis=axis)`. For example, if both `a` and `b` has shape `[2, 3, 4]`, with `mx.sum(a * b, axis=1)` I can get the desired output with shape `[2, 4]` (`3` is reduced). But with `mx.tensordot(a, b, axes=[[1], [1]])`, it produces shape `[2, 4, 2, 4]`. I would like to use something like `np.einsum` to do `mx.einsum("ijk,ijk->ik", a, b)` but MLX doesn't have a Einsum.

AtomicVar:
@awni Thank you for taking the time to review the PR. I've pushed a new commit to address to problems, except the `mx.tensordot` thing. Let me know if this is acceptable.

python/mlx/nn/losses.py
awni: Indeed, sorry `tensordot` is not the right call, sorry for the bad suggestion. Let's keep the sum reduce as you have it actually, that is simple and likely fine perf wise.

awni (approved):
Thanks for the addition, looks great!!

PR #519, opened 2024-01-22
Buffer Donation

awni:
## Proposed changes

Donate input buffers to outputs when possible.


mlx/transforms.cpp
```diff
@@ -262,6 +262,7 @@ void eval(const std::vector<array>& outputs) {
     auto stream = arr.primitive().stream();
     std::vector<std::shared_future<void>> arr_deps;
     for (auto& in : arr.inputs()) {
+      // TODO that's a bug
```
awni, line 265: This is unrelated to this PR, but I believe the below is a race condition (input's primitive can be wiped before the output task is scheduled). It's only relevant for multi-stream stuff. Will have to test and fix but makes sense for a follow up.

python/tests/test_ops.py
```diff
@@ -1665,7 +1665,7 @@ def test_tensordot(self):
 
     def test_inner(self):
         self.assertCmpNumpy([(3,), (3,)], mx.inner, np.inner)
-        self.assertCmpNumpy([(1, 1, 2), (3, 2)], mx.inner, np.inner)
+        self.assertCmpNumpy([(1, 1, 2), (4, 2)], mx.inner, np.inner)
```
awni, line 1668: @jagrit06 something strange here. If I use `3` the test does not work and it gives bogus results. It feels like it's reading from memory that it should not be in the `gemv` specialization. The shapes are `[1, 2]` and `[4, 2].T`.

It's also possible I messed something up, but every other test clears, and benchmarks and examples still work. And this test fails even when I use zero buffers (which don't have any donation). It also passes if I route to our `gemm` instead of `gemv` and it passes if I use MPS `gemm` as well.

Would you mind taking a look at that case?

awni:
Will append benchmarks to this comment:

### In place reduces runaway memory allocations:

This benchmark now uses only 8GB instead of around 80-90GB before this change:

```python
    n = 8192
    x = mx.zeros((n, n, 16))

    def abs_inplace(x):
        for _ in range(20):
            x = mx.abs(x)
        return x
```

python/tests/test_ops.py
jagrit06: Alright, I'll hop on it

awni: Reverted this change since @jagrit06 fixed in #523

awni:
I did a benchmark with the Transformer model to see memory use changes.

To highlight the impact a bit I changed a coupe things:
1. Use context=512, batch size 1 so the model uses more memory
2. Use Adam which does a lot of binary / unary stuff
3. Use a step function so grads can also be donated since it doesn't hold a reference to them e.g.

```python
  def step(inputs, targets):
        loss, grads = loss_and_grad_fn(inputs, targets)
        model.update(optimizer.apply_gradients(grads, model))
        return loss
```

And do `mx.eval(model.parameters())` outside that function.

Dim | Main | Inplace |
---   | ---- | ----- |
1024 | 3.75 | 3.05 |
2048 | 11.13 | 9.32 |

Commands for reference 
```
python main.py --gpu --batch_size 1 --context_size 512
python main.py --gpu --batch_size 1 --context_size 1024
```

awni:
@angeloskath couple questions:

1. This PR is a draft, I didn't clean up / do all the CPU because I want to get your input on how it's done first. If you think it's roughly good as is I will clean it up and complete it.
2. I suspect the impact of in place could be greater if we used a better order in evaluations so that we let go of references in the tape earlier (related to the priority queue stuff). Presumably the whole backward pass consists of potentially donateable buffers for the activations / deltas and we don't get much of it.

angeloskath:
I think it is great as it is!

I also think that I am having similar issues both with the priority queue evaluation as well as the checkpointing so I will spend some time trying to figure this out. Meaning, how can we evaluate the graph making sure that we are using a good path (not optimal but at least good). One thing that comes to mind is maybe when doing the dfs we should first go down the path of a detachable array?

awni:
Ok, this is ready to be reviewed. I'm happy with the GPU memory saving results. We don't see a huge gain on the CPU since it's all sequential anyway allocations free up quickly.

The additional complexity is not too bad.. worth the benefit IMO. In certain cases this can really save a ton of memory (for example when you have a chain of unary / binary ops).

awni:
The CPU is slightly faster which is nice:

```
python main.py                                            
Training a transformer with 153.883 M parameters
Iter 20: Train loss 7.104, It/sec 0.284
```

```
python main.py 
Training a transformer with 153.883 M parameters
Iter 10: Train loss 7.829, It/sec 0.296
Iter 20: Train loss 7.422, It/sec 0.302
```

angeloskath (approved):
Everything looks really good! The results are obviously fantastic.

My only question is why do we donate with move sometimes? Is it to avoid the copy of the `shared_ptr`? It seems like a lot of code (like 1-2 if statements) could be removed if we just copied the buffer 🤷

awni:
Actually that is an excellent question and I will make a note about it because is critical and highly non obvious. I think it is worth documenting the source code so I will put it there.

angeloskath:
The other thing I thought was that it could be that the same array can be donated twice but then I am not sure that that's safe right?

angeloskath:
> The other thing I thought was that it could be that the same array can be donated twice but then I am not sure that that's safe right?

Actually that's stupid. It can never be donated twice... Anyway I 'll wait to read the comments :-)

awni:
(Sorry for being a bit pedantic, I'm writing this for future me and friends as well)

The basic idea is the following. This applies to the Metal back-end only. CPU buffer donation is simple and does not require moving the buffer out of the input.

An array's data is donate-able if it satisfies two criteria:

1. It has a descriptor `use_count` of `1`
2. It has a data `use_count` of `1`

To get the descriptor `use_count` to be `1` for donation we now detach the array as soon as the kernel has been put in the command buffer. That removes references to the input arrays as soon as we are "done" with them.

However, because of the above, in order to keep the array's input data buffers around until the GPU is done computing the array itself, we have to hold a strong references to them. We get these reference via `Array:data_shared_ptr` and put them into the completion handler.

However, if the output array received a donated buffer from it's input AND we hold the buffer in the  completion handler, then the data `use_count` will be `2`, which precludes the buffer from any future donation.  Concretely, something like:

```python
for _ in range(4):
    x = mx.abs(x)
```

would have a pattern of only every other array donating it's buffer instead of just using the same buffer throughout.

One way to fix this is to move the buffer out of the input as soon as it's donated. This has the effect of letting only the output array hold a strong reference to it, since a `nullptr` will now be added to the completion handler.

awni:
@jagrit06 @angeloskath one other thing to be cognizant of with this change. In Python land we should discourage patterns  that hold references to arrays when they aren't needed otherwise buffer donation won't happen. (It's kind of nice as this is somewhat intuitive IMO). But in the following the first pattern is better from a memory reuse standpoint than the second since the reference to `grads` is gone before the `eval` is called.:

```python
def step(input, target):
    loss, grads = loss_vallue_and_grad(model, input, target)

for input, target in dataset:
    step(input, target)
    mx.eval(model.parameters(), optimizer.state)
```

```python
for input, target in dataset:
    loss, grads = loss_vallue_and_grad(model, input, target)
    mx.eval(model.parameters(), optimizer.state)
```

angeloskath:
> However, if the output array received a donated buffer from its input AND we hold the buffer in the completion handler, then the data use_count will be 2, which precludes the buffer from any future donation.

Makes perfect sense! Thanks.

Regarding avoiding holding unnecessary references to things, even in PT-land it makes sense to delete or set things to `None` after one is done using them. So the wrapping function although elegant is not needed. The following would have the same effect.

```python
for input, target in dataset:
    loss, grads = loss_vallue_and_grad(model, input, target)
    del grads
    mx.eval(model.parameters(), optimizer.state)
```

awni:
Of course, you can use `del` as well, that's totally legit.

angeloskath:
Feel free to merge this whenever you want :-)

awni:
PS I apologize for any rebasing headaches this change causes. The surface area is pretty large 🥲

sck-at-ucy:
I am sorry if I am creating noise by jumping in the discussion, but reading through your comments was very educational for me. Is this a good practice to apply in general 

```
for input, target in dataset:
    loss, grads = loss_vallue_and_grad(model, input, target)
    del grads
    mx.eval(model.parameters(), optimizer.state)
```

I mean to `del grads`. I understand what it does, is there any reason not to always do this? 

Also a question about `mx.eval(model.parameters(), optimizer.state)`. A few times I forget to include it in my code and I did not see a difference. Is it only under certain scenarios that this matters? Naive questions... 😬🫣

angeloskath:
@sck-at-ucy for cases where there is memory pressure I would say it is a good practice. There is nothing to lose so in general it could be done.

Regarding omitting the eval, if you have a print or some other implicit evaluation then you may not see a difference. However, if you don't then the graph is gonna keep growing to a point where it doesn't make much sense. Even if you do have a print, it can be more efficient to evaluate everything explicitly rather than one by one especially for shared computation. For instance,

```python
x, y = my_complicated_multi_output_network(a, b, c)
# this will evaluate x and then y
print(x)
print(y)
# the following will evaluate them together. At the very least the
# overhead of python to C++ will be shared.
mx.eval(x, y)
```

sck-at-ucy:
Thanks this clarifies things in my mind

PR #520, opened 2024-01-22
Add the Mean Squared Logarithmic Error (MSLE) loss function

AtomicVar:
## Proposed changes

Mean Squared Logarithmic Error (MSLE) is a loss function used in regression problems. Some ML frameworks/libraries contain this loss function, including:

- [TensorFlow (tf.keras.losses.MeanSquaredLogarithmicError)](https://www.tensorflow.org/api_docs/python/tf/keras/losses/MeanSquaredLogarithmicError)
- [torchmetrics (MeanSquaredLogError)](https://lightning.ai/docs/torchmetrics/stable/regression/mean_squared_log_error.html)
- [scikit-learn (mean_squared_log_error)](https://scikit-learn.org/stable/modules/model_evaluation.html#mean-squared-log-error)

This PR adds a `msle_loss` function to `nn.losses` module.

## Checklist

### DIFF
--- a/docs/src/python/nn/losses.rst
+++ b/docs/src/python/nn/losses.rst
@@ -19,6 +19,7 @@ Loss Functions
    l1_loss
    log_cosh_loss
    mse_loss
+   msle_loss
    nll_loss
    smooth_l1_loss
    triplet_loss
--- a/python/mlx/nn/losses.py
+++ b/python/mlx/nn/losses.py
@@ -171,6 +171,49 @@ def mse_loss(
     return _reduce(loss, reduction)
 
 
+def msle_loss(
+    inputs: mx.array, targets: mx.array, reduction: Reduction = "mean"
+) -> float:
+    r"""
+    Calculate the mean squared logarithmic error (MSLE) between predictions and
+    targets.
+
+    The loss is given by:
+
+    .. math::
+        \text{MSLE} = \frac{1}{N}\sum_i^N (\log(1 + y_i) - \log(1 + \hat{y_i}))^2
+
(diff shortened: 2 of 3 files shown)


Stealeristaken (changes requested):
Maybe adding 
```python
 if (targets < 0).any() or (inputs < 0).any():
        raise ValueError(
            "Mean Squared Logarithmic Error cannot be used when "
            "targets contain negative values."
        )
```

 could be nice and 
```python
loss=mse_loss(mx.log1p(targets), mx.log1p(inputs))
```


can be more optimized.

awni:
If you do something like `if (targets < 0).any()` it will eval the graph, so that is not a good idea. It won't break gradient transforms, but it could be very slow.

I think the implementation is fine the way you have it, that said I don't think we should include this in `nn.losses` yet. It is pretty niche and also reasonably easy to implement so I don't see a strong need to include it here. We can revisit if that  changes in the future. In the meantime, I will close this.

Stealeristaken:
> If you do something like `if (targets < 0).any()` it will eval the graph, so that is not a good idea. It won't break gradient transforms, but it could be very slow.
> 
> I think the implementation is fine the way you have it, that said I don't think we should include this in `nn.losses` yet. It is pretty niche and also reasonably easy to implement so I don't see a strong need to include it here. We can revisit if that changes in the future. In the meantime, I will close this.

I know its a bit late response sorry but what do you mean by `it will eval the graph` and why is it a bad idea? Couldn't figure out

awni:
You can read about how lazy evaluation works in MLX in the [docs](https://ml-explore.github.io/mlx/build/html/usage/lazy_evaluation.html). I think it should answer your question.

PR #521, opened 2024-01-22
Fix package installation for CI

danilopeixoto:
## Proposed changes

`python setup.py install` (deprecated) command is attempting to install dependencies based on remote name after building and installing the local package. We can use `python setup.py develop` or `python -m pip install .` instead.

Using `python setup.py develop` for this request.

https://app.circleci.com/pipelines/github/ml-explore/mlx/691/workflows/c86339cd-d578-4df6-acd3-7ce7043e83fb/jobs/1832

## Checklist


awni:
🤔 I think I'm missing something: `python setup.py install` runs fine on my machine. Do you know which command is causign it to fail?

danilopeixoto:
I don’t know exactly… But I suspect the production release works because the `python setup.py install` command finds dependencies of `mlx==0.0.10` (version set in `setup.py` for `PYPI_RELEASE=1`) in PyPi from previous upload. This does not work when the version is not in PyPi (for undefined `PYPI_RELEASE`). However, the command should not search dependencies on remote such as `python setup.py develop` or `python -m pip install .`.

angeloskath:
The failure is the check for the installation version and maybe setup tools don't support versions like `major.minor.patch.whatever` so it fails. It succeeds when `PYPI_RELEASE=1` cause then the version string is normal.

My suggestion @danilopeixoto is to change the `install` to `develop` and we 'll fix it properly when we move to `pypa/build` or something like that. The reason for develop rather than `install` in order to reuse the mlx build rather than build it twice one for install and one for `bdist_wheel`.

angeloskath (changes requested):
Left it as a comment as well, let's change that `install` to `develop` for now and we should revisit the whole build shortly.

angeloskath:
Yeah sorry we can merge this. @danilopeixoto the comment regarding semver versioning is not exactly correct (our versioning is fine, for instance devXXX builds are even uploadable to PyPI) but I don't mind keeping it. @awni wdyt?

.circleci/config.yml
```diff
@@ -136,6 +136,10 @@ jobs:
             pip install pybind11-stubgen
             pip install numpy
             pip install twine
+      # TODO: Implement new build system and/or versioning scheme compatible with
+      # Python packages and Semantic Versioning (https://semver.org).
+      # Following this, modify the Python package installation command to transition
+      # from 'development' to 'production' mode."
```
awni, line 142:
```suggestion
      # TODO: Modify the Python package installation command to transition
      # from 'development' to 'production' mode."
```

awni, line 142:
```suggestion
      # TODO: Update build system to switch away from setup.py develop
```

awni:
Ah I didn't notice that. Yea let's remove the bit about semantic versioning (left a suggestion in line)

angeloskath (approved):

PR #523, opened 2024-01-22
Fix out of bounds reads in gemv kernel

jagrit06:
## Proposed changes

* Added a check against out of bounds matrix reads in the gemv kernel 
* No significant performance regressions observed (tested on M3 max)

## Checklist


awni (approved):
Thanks for the quick fix @jagrit06

PR #525, opened 2024-01-23
Use `axes` in Tensordot instead of `dims

awni:
## Proposed changes

This aint PyTorch after all 😄


angeloskath (approved):
Oops how did this slip through... 🤔

PR #526, opened 2024-01-23
Implement Diagonal operator

PRESIDENT810:
## Proposed changes

Implement Diagonal operator

This is still in development as I'm not sure what is the best way to implement Diagonal's vjp API.

## Checklist


awni:
Thanks for adding this!

I am not sure this is worth making a primitive for however. As you can see it ends up being a lot of code (and you have to write all the functions VJP, JVP, etc). One way to implement `diagonal` is to do:

1. A gather/take to extract the diagonal
2. A scatter on an array of zeros to construct a diagonal matrix

We already have those ops/primitives so that would reduce the code a bunch and automatically get you VJPs, JVPs, etc.

PR #533, opened 2024-01-23
mx.diag implementation

### LINKED ISSUE #503: [Feature] `mx.diag`
Follow the [NumPy API](https://numpy.org/doc/stable/reference/generated/numpy.diag.html)

- C++ op with bindings in python
- tests

ManishAradwad:
## Proposed changes

closes #503 

## Checklist

### DIFF
--- a/mlx/array.h
+++ b/mlx/array.h
@@ -365,7 +365,7 @@ class array {
   // The ArrayDesc contains the details of the materialized array including the
   // shape, strides, the data type. It also includes
   // the primitive which knows how to compute the array's data from its inputs
-  // and a the list of array's inputs for the primitive.
+  // and a list of array's inputs for the primitive.
   std::shared_ptr<ArrayDesc> array_desc_{nullptr};
 };
 
--- a/mlx/ops.cpp
+++ b/mlx/ops.cpp
@@ -980,15 +980,15 @@ array diagonal(
     int axis1 = 0,
     int axis2 = 1,
     StreamOrDevice s) {
-  auto ndim = a.ndim();
-  auto ax1 = (axis1 < 0) ? axis1 + ndim : axis1;
+  int ndim = a.ndim();
+  int ax1 = (axis1 < 0) ? axis1 + ndim : axis1;
   if (ax1 < 0 || ax1 >= ndim) {
     std::ostringstream msg;
     msg << "Invalid axis1 " << axis1 << " for array with " << ndim
         << " dimensions.";
     throw std::out_of_range(msg.str());
   }
-  auto ax2 = (axis2 < 0) ? axis2 + ndim : axis2;
+  int ax2 = (axis2 < 0) ? axis2 + ndim : axis2;
   if (ax2 < 0 || ax2 >= ndim) {
     std::ostringstream msg;
     msg << "Invalid axis2 " << axis2 << " for array with " << ndim
@@ -1031,11 +1031,18 @@ array diagonal(
   }
 
   std::vector<int> slice_sizes;
-  for (int i = 0; i < a_ndim; ++i) {
-    slice_sizes.push_back(1);
+  for (int i = 0; i < (a_ndim - 2); ++i) {
+    slice_sizes.push_back(a_moved.shape(i));
   }
+  slice_sizes.push_back(1);
+  slice_sizes.push_back(1);
 
   array result = gather(a_moved, indices, axes, slice_sizes, s);
+
+  // To handle the case where diagonal has only one element
+  if (result.shape(0) == 1) {
+    return squeeze(result, -1, s);
+  }
 
   return squeeze(result, s);
 }
--- a/mlx/ops.cpp
+++ b/mlx/ops.cpp
@@ -974,6 +974,66 @@ array transpose(const array& a, StreamOrDevice s /* = {} */) {
   return transpose(a, std::move(axes), to_stream(s));
 }
 
+array diagonal(
+    const array& a,
+    int offset = 0,
+    int axis1 = 0,
+    int axis2 = 1,
+    StreamOrDevice s) {
+  auto ndim = a.ndim();
+  if (a.shape(0) != a.shape(1)) {
+    throw std::invalid_argument("Input must be 2-d and square.");
+  }
+  if (axis1 < 0 || axis2 < 0) {
+    throw std::invalid_argument(
+        "Invalid axis argument provided. Axis value should be non-negative.");
+  }
+  auto ax1 = (axis1 < 0) ? axis1 + ndim : axis1;
+  if (ax1 < 0 || ax1 >= ndim) {
+    std::ostringstream msg;
+    msg << "Invalid axis1 " << axis1 << " for array with " << ndim
+        << " dimensions.";
+    throw std::out_of_range(msg.str());
+  }
+  auto ax2 = (axis2 < 0) ? axis2 + ndim : axis2;
+  if (ax2 < 0 || ax2 >= ndim) {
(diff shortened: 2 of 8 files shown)


ManishAradwad:
@awni i need your help for extracting the diagonal from 2d array. i wasn't sure about how we can extract diagonal without indexing the input 2d array and i couldn't find the syntax to index an mlx array. 

I'll also add the tests for `mx.diag` in the next few days.

awni:
The op you are looking for is [`gather`](https://github.com/ml-explore/mlx/blob/main/mlx/ops.h#L843).

Although I'm uncertain about this PR vs #526 ? It seems like this implementation is closer to how we should do it (via an op not a primitive). Should we close #526? @PRESIDENT810 could you comment?

One thing that would be nice with this one is to do `diagonal` (e.g. following NumPy) and then you can simply have `diag` call `diagonal`. Does that make sense?

PRESIDENT810:
@awni I think only using a op instead of a primitive is better so free to close my PR.

And @ManishAradwad feel free to reference anything that helps. I think apart from the primitive thing other parts should be working fine.

awni:
Thanks, sounds like a great plan!

PRESIDENT810:
Also @ManishAradwad if you are uncertain about how to use gather to extract the diagonal array, I took some time to read JAX's implementation which I think might be helpful

```
def diagonal(a: ArrayLike, offset: int = 0, axis1: int = 0,
             axis2: int = 1) -> Array:
  util.check_arraylike("diagonal", a)
  a_shape = shape(a)
  if ndim(a) < 2:
    raise ValueError("diagonal requires an array of at least two dimensions.")
  offset = core.concrete_or_error(operator.index, offset, "'offset' argument of jnp.diagonal()")

  a = moveaxis(a, (axis1, axis2), (-2, -1))

  diag_size = max(0, min(a_shape[axis1] + min(offset, 0),
                         a_shape[axis2] - max(offset, 0)))
  i = arange(diag_size)
  j = arange(abs(offset), abs(offset) + diag_size)
  return a[..., i, j] if offset >= 0 else a[..., j, i]
```
So I think one way to do this is:
1. Move axis1 and axis2 so they become the last two axes of the input array
2. Compute what should be the diagonal size of that two axes
3. Construct the indices of elements you want to extract for the two diagonal axes
4. Use `gather` is to extract all elements for the first n-2 dims, and extract elements using the indices you construct in step 3 to get the diagonal elements.

Feel free to ping me if you need any help.

ManishAradwad:
Thanks a lot for your instructions @PRESIDENT810. I'm still stuck at using `gather` to extract the diagonal elements(step 4 above). I tried implementing it but was not able to get the right elements.

I didn't understand why we need to extract all elements for the first n-2 dims. Also, I think I'm messing the `slice_sizes` argument of gather. Is it fixed as `diag_size` or there is more to it?

PRESIDENT810:
@ManishAradwad 
According to numpy's doc:
> If a has more than two dimensions, then the axes specified by axis1 and axis2 are used to determine the 2-D sub-array whose diagonal is returned. The shape of the resulting array can be determined by removing axis1 and axis2 and appending an index to the right equal to the size of the resulting diagonals.
So if a has n dims with shape (d1, d2 ... dn) and selected axis x, y, we need to keep the unselected n-2 dims and extract the diagonal matrix of the 2D subarray, whose two axes should be axis1 and axis2. The result should be of shape (d1, d2 ... dn, diag_size), with dx and dy taken out from d1, d2 ... dn. 

For example, if you run `np.arange(12).reshape(3, 2, 2).diagonal(0, 1, 2)`, the result should be
```
array([[ 0,  3],
       [ 4,  7],
       [ 8, 11]])
```
with shape (3, 2,). This is because 3 is the unselected dim and 2 is the diag_size. For each 2*2 subarray you need to take the element corresponding to the first dimension (which is 3).

As for gather, I'm not really sure how to use it either. Let me take some time to look at it.

ManishAradwad:
I got it working for 2-d arrays. But still, not getting the correct results for higher dimensional arrays. will keep working on it.

awni:
I'm sorry our C++ [gather](https://ml-explore.github.io/mlx/build/html/usage/function_transforms.html#automatic-differentiation) is undocumented.

The arguments to pay attention to are `indices`, `axes` and `slice_sizes`:

- Indices: the indices you want to extract. This would be e.g. `arange(shape(axis1))` and `arange(shape(axis2))`
- Axes: the axes corresponding to the indices `{axis1, axis2}`
- Slice sizes: the size of each slice. It should be a `1` in the dim corresponding to `axis1` and `axis2` and the full shape everywhere else.

ManishAradwad:
I have updated the code according to the above description. The dimensions returned for higher dimensional inputs are correct, but the diagonal values extracted are still incorrect. 

@awni @PRESIDENT810 Please check if I'm populating the input arrays to gather properly.  

Also, if the diagonal to be returned contains only one value like [1] then it returns a scalar value(just 1) since I'm squeezing the `results` array. I have added the code to handle that case, but now it returns a 2d array like [[1]]. How can i fix this?

PRESIDENT810:
@ManishAradwad 
> Also, if the diagonal to be returned contains only one value like [1] then it returns a scalar value(just 1) since I'm squeezing the results array. I have added the code to handle that case, but now it returns a 2d array like [[1]]. How can i fix this?
My guess is you can just use `reshape` to make it the right shape.

PRESIDENT810:
Hummm I don't quite understand how to use `gather` either.
To get `a[..., indices1, indices2]` I'm using
```
x = array({0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11}, {3, 2, 2});
auto indices0 = array({0});
auto indices1 = arange(2);
auto indices2 = arange(2);
auto out = gather(x, {indices0, indices1, indices2}, {0, 1, 2}, {3, 1, 1});
out = reshape(out, {3, 2});
```
But the outcome is 
```
array([[0, 4],
           [8, 3],
           [7, 11]], dtype=int32)
```
where as the correct result of `a[..., indices1, indices2]` should be
```
array([[0, 3],
           [4, 7],
           [8, 11]], dtype=int32)
```
I find the output's dimensions are confusing. @awni could you please throw some light on this?

Also I hope we can have a `gather_nd` C++ operator API just like `mlx_gather_nd` in Python binding, which would make things easier here.

PRESIDENT810:
Ah I've got it! Looks like `gather` behaves differently than I thought and I need to do a `moveaxis` to make it work. 

I've successfully implemented it in #562, which supports higher dimensional array and passes all tests. @ManishAradwad @awni Mind if take a look? I think that one should be working!

ManishAradwad:
Great @PRESIDENT810 ! I will modify this PR to use the `diagonal` implemented by you in the `diag` function. Or maybe you can include `diag` in your PR. Any way is fine by me. @awni , do we need a separate PR for `diag`?

PRESIDENT810:
@ManishAradwad You can push your `diag` implementation to my [feat_diagonal](https://github.com/PRESIDENT810/mlx/tree/feat_diagonal) branch so they will be included in the same PR. I just invited you to collaborate in my mlx fork repo.

ManishAradwad:
Sure, I'll push to your branch. 

> Also I think apart from diagonal tests we still need a few separate tests for diag. You can add that to ops_test.cpp or I can handle it after you pushed your implementation.

Yeah, I'll also add the tests for `diag` on the same branch

awni:
Nice to see the collaboration and progress on this! I assume we will close this PR in favor of #562?

ManishAradwad:
Yup, I'll close it.

PR #539, opened 2024-01-24
[Fix] mx.allclose bug with infinite values

### LINKED ISSUE #522: [BUG] mx.allclose fails for inf == inf.
**Describe the bug**

`mx.allclose` always fails when `inf` is in the inputs but it should pass if the `inf`s are the same sign.

**To Reproduce**

Include code snippet:

```shell
>>> import mlx.core as mx
>>> a = mx.array(float("inf"))
>>> mx.allclose(a, a)
array(False, dtype=bool)
>>> import numpy as np
>>> np.allclose(a, a)
True
```

This occasionally causes some of our float16 tests to fail when an inf pops up (even if it's expected).

Rifur13:
## Proposed changes

This fixes #522 . 

Handle infinite and NaN values gracefully, following the Numpy API. Added a parameter `equal_nan` that treats `NaN` as equal if compared.

Added `mx.isclose`, which returning the results of a element-wise comparisons in a boolean array.

## Checklist

### DIFF
--- a/mlx/ops.cpp
+++ b/mlx/ops.cpp
@@ -1156,7 +1156,7 @@ array allclose(
     double atol /* = 1e-8 */,
     bool equal_nan /* = false */,
     StreamOrDevice s /* = {}*/) {
-    return all(isclose(a, b, rtol, atol, equal_nan, s), s);
+  return all(isclose(a, b, rtol, atol, equal_nan, s), s);
 }
 
 array isclose(
@@ -1167,30 +1167,32 @@ array isclose(
     bool equal_nan /* = false */,
     StreamOrDevice s /* = {}*/) {
   // |a - b| <= atol + rtol * |b|
-    auto rhs = add(array(atol), multiply(array(rtol), abs(b, s), s), s);
-    auto lhs = abs(subtract(a, b, s), s);
-    auto out = less_equal(lhs, rhs, s);
+  auto rhs = add(array(atol), multiply(array(rtol), abs(b, s), s), s);
+  auto lhs = abs(subtract(a, b, s), s);
+  auto out = less_equal(lhs, rhs, s);
 
-    // Correct the result for infinite values.
-    auto any_inf = isinf(a, s) || isinf(b, s) || isneginf(a, s) || isneginf(b, s);
-    auto both_inf = (isinf(a, s) && isinf(b, s)) || (isneginf(a, s) && isneginf(b, s));
-    auto any_nan = isnan(a, s) || isnan(b, s);
+  // Correct the result for infinite values.
+  auto any_inf = isinf(a, s) || isinf(b, s) || isneginf(a, s) || isneginf(b, s);
+  auto both_inf =
+      (isinf(a, s) && isinf(b, s)) || (isneginf(a, s) && isneginf(b, s));
+  auto any_nan = isnan(a, s) || isnan(b, s);
 
-    // Convert all elements where either value is infinite to False.
-    out = out && logical_not(any_inf, s);
+  // Convert all elements where either value is infinite to False.
+  out = out && logical_not(any_inf, s);
 
-    // Convert all the elements where both values are infinite and of the same sign to True.
-    out = out || both_inf;
+  // Convert all the elements where both values are infinite and of the same
+  // sign to True.
+  out = out || both_inf;
 
-    // Convert all the elements where either value is NaN to False.
-    out = out && logical_not(any_nan, s);
+  // Convert all the elements where either value is NaN to False.
+  out = out && logical_not(any_nan, s);
 
-    if (equal_nan) {
-      auto both_nan = isnan(a, s) && isnan(b, s);
-      out = out || both_nan;
-    }
+  if (equal_nan) {
+    auto both_nan = isnan(a, s) && isnan(b, s);
+    out = out || both_nan;
+  }
 
-    return out;
+  return out;
 }
 
 array all(const array& a, bool keepdims, StreamOrDevice s /* = {}*/) {
--- a/mlx/ops.cpp
+++ b/mlx/ops.cpp
@@ -1119,10 +1119,7 @@ array isnan(const array& a, StreamOrDevice s /* = {} */) {
 }
 
 array isinf(const array& a, StreamOrDevice s /* = {} */) {
-  if (is_integral(a.dtype())) {
-    return full(a.shape(), false, bool_, s);
-  }
-  return equal(a, array(std::numeric_limits<float>::infinity(), a.dtype()), s);
+  return logical_or(isposinf(a, s), isneginf(a, s), s);
 }
 
 array isposinf(const array& a, StreamOrDevice s) {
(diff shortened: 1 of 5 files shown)


python/tests/test_ops.py
```diff
@@ -855,6 +855,21 @@ def test_allclose(self):
...
         self.assertTrue(mx.allclose(a, b, 0.01, 0.1).item())
 
+        c = mx.array(float("inf"))
+        self.assertTrue(mx.allclose(c, c).item())
+
+    def test_isclose(self):
+        a = mx.array([float("inf"), float("inf"), float("-inf")])
+        b = mx.array([float("inf"), float("-inf"), float("-inf")])
+
+        self.assertListEqual(mx.isclose(a, b).tolist(), [True, False, True])
```
awni, line 865: 👍

mlx/ops.cpp
```diff
@@ -1154,11 +1154,45 @@ array allclose(
...
+    bool equal_nan /* = false */,
     StreamOrDevice s /* = {}*/) {
   // |a - b| <= atol + rtol * |b|
   auto rhs = add(array(atol), multiply(array(rtol), abs(b, s), s), s);
   auto lhs = abs(subtract(a, b, s), s);
-  return all(less_equal(lhs, rhs, s), s);
+  auto out = less_equal(lhs, rhs, s);
+
+  // Correct the result for infinite values.
+  auto any_inf = isinf(a, s) || isinf(b, s) || isneginf(a, s) || isneginf(b, s);
```
awni, line 1175: `isinf` is true if `isposinf` is true or `isneginf` is true so you don't need to also check `isneginf` here

mlx/ops.cpp
```diff
@@ -1154,11 +1154,45 @@ array allclose(
...
+  auto both_inf =
+      (isinf(a, s) && isinf(b, s)) || (isneginf(a, s) && isneginf(b, s));
```
awni, line 1177: You should use `isposinf` here?

mlx/ops.cpp
```diff
@@ -1154,11 +1154,45 @@ array allclose(
...
+
+  // Convert all elements where either value is infinite to False.
+  out = out && logical_not(any_inf, s);
+
+  // Convert all the elements where both values are infinite and of the same
+  // sign to True.
+  out = out || both_inf;
+
+  // Convert all the elements where either value is NaN to False.
+  out = out && logical_not(any_nan, s);
```
awni, line 1188: I believe this is not needed. If any element is NaN than the `<` operator will be `false` always.

```
>>> 1 <= mx.array(float("nan"))
array(False, dtype=bool)
>>> mx.array(float("nan")) <= 1
array(False, dtype=bool)
```

mlx/ops.cpp
```diff
@@ -1154,11 +1154,45 @@ array allclose(
...
   auto rhs = add(array(atol), multiply(array(rtol), abs(b, s), s), s);
   auto lhs = abs(subtract(a, b, s), s);
-  return all(less_equal(lhs, rhs, s), s);
+  auto out = less_equal(lhs, rhs, s);
+
+  // Correct the result for infinite values.
+  auto any_inf = isinf(a, s) || isinf(b, s) || isneginf(a, s) || isneginf(b, s);
+  auto both_inf =
+      (isinf(a, s) && isinf(b, s)) || (isneginf(a, s) && isneginf(b, s));
+  auto any_nan = isnan(a, s) || isnan(b, s);
```
awni, line 1178: Remove this as not needed.

python/src/ops.cpp
```diff
@@ -1668,6 +1669,46 @@ void init_ops(py::module_& m) {
...
+
+        Returns a boolean array where two arrays are element-wise equal within a tolerance.
+
+        Infinite values are treated as equals if they're of the same sign. NaN are never treated as equals.
+
+        Two values are considered equal if:
+
+        .. code-block::
+
+         all(abs(a - b) <= (atol + rtol * abs(b)))
```
awni, line 1700:
```suggestion
         abs(a - b) <= (atol + rtol * abs(b))
```

python/src/ops.cpp
```diff
@@ -1668,6 +1669,46 @@ void init_ops(py::module_& m) {
...
+      "atol"_a = 1e-8,
+      py::kw_only(),
+      "equal_nan"_a = false,
+      "stream"_a = none,
+      R"pbdoc(
+        isclose(a: array, b: array, /, rtol: float = 1e-05, atol: float = 1e-08, *, equal_nan: bool = False, stream: Union[None, Stream, Device] = None) -> array
+
+        Returns a boolean array where two arrays are element-wise equal within a tolerance.
+
+        Infinite values are treated as equals if they're of the same sign. NaN are never treated as equals.
```
awni, line 1694:
```suggestion
        Infinite values are considered equal if they have the same sign, NaN values are
        not equal unless ``equal_nan`` is ``True``.
```

awni: Could you also update the `allclose` doc with the same comment?

python/src/ops.cpp
```diff
@@ -1668,6 +1669,46 @@ void init_ops(py::module_& m) {
...
+        Note unlike :func:`array_equal`, this function supports numpy-style
+        broadcasting.
+
+        Args:
+            a (array): Input array.
+            b (array): Input array.
+            rtol (float): Relative tolerance.
+            atol (float): Absolute tolerance.
+            equal_nan (bool): If ``True``, NaNs are treated as equal.
+              Defaults to ``False``.
```
awni, line 1711:
```suggestion
            equal_nan (bool): If ``True``, NaNs are considered equal.
              Defaults to ``False``.
```

awni (changes requested):
Nice thank you!!  Left a few comments, please address and then I will check again.

python/tests/test_ops.py
Rifur13: Thanks for the review!

mlx/ops.cpp
```diff
@@ -1154,11 +1154,45 @@ array allclose(
     const array& b,
     double rtol /* = 1e-5 */,
     double atol /* = 1e-8 */,
+    bool equal_nan /* = false */,
+    StreamOrDevice s /* = {}*/) {
+  return all(isclose(a, b, rtol, atol, equal_nan, s), s);
+}
+
+array isclose(
```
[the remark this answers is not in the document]
Rifur13: Done.

mlx/ops.cpp
Rifur13: Done.

mlx/ops.cpp
Rifur13: Done.

mlx/ops.cpp
Rifur13: Yes. `isinf` and `isposinf` actually have the same implementation. `isinf` is wrong so I fixed it and added a test.

mlx/ops.cpp
Rifur13: Yes. `isinf` has the wrong implementation so I fixed it too and added a test.

mlx/ops.cpp
```diff
@@ -1119,10 +1119,7 @@ array isnan(const array& a, StreamOrDevice s /* = {} */) {
 }
 
 array isinf(const array& a, StreamOrDevice s /* = {} */) {
-  if (is_integral(a.dtype())) {
-    return full(a.shape(), false, bool_, s);
-  }
-  return equal(a, array(std::numeric_limits<float>::infinity(), a.dtype()), s);
+  return logical_or(isposinf(a, s), isneginf(a, s), s);
```
awni, line 1122: 👍 thanks for fixing that + the tests.

awni (approved):
Thanks for fixing this, much appreciated!

PR #541, opened 2024-01-24
Custom VJP and checkpointing

angeloskath:
Introduces `custom_vjp`, `checkpoint` and `depends`.

Looking at the value and grad graph of some simple functions it seems to work however I don't see the correct memory usage pattern in real world training so I will be looking for the bug.

Aforementioned simple function and graphs

```python
import mlx.core as mx
def f(a, b):
    return (a.square() + b.square()).sqrt()

def g(a, b):
    c = mx.checkpoint(f, a, b)
    d = mx.checkpoint(f, b, c)
    return d;

a = mx.array(0.3)
b = mx.array(0.4)
mx.eval(a, b)
mx.export("g1.png", mx.value_and_grad(g)(a, b))
mx.export("g2.png", mx.value_and_grad(lambda a, b: f(b, f(a, b)))(a, b))
```

g
g2


angeloskath:
Ok, I believe this is ready for review.

The main thing to talk about is the change in `eval()`. The depth heuristic works quite well but it adds a bit of overhead so most things have had a small regression (~1%). The next steps would be to measure how much of this regression is coming from the extra DFS or simply due to the different graph processing order. The first checks are promising as the regression becomes smaller and smaller the larger the dimensions.

mlx/ops.h
```diff
@@ -1094,4 +1094,26 @@ array addmm(
...
     StreamOrDevice s = {});
 
+/** Implements the identity function but allows injecting dependencies to other
+ * arrays to ensure that they will be computed before anything that depends on
+ * the inputs. */
+std::vector<array> depends(
+    const std::vector<array>& inputs,
+    const std::vector<array>& dependencies);
+
+/** Allows to have call a function and define a custom VJP for it. */
```
jagrit06, line 1104: Typo

mlx/transforms.cpp
```diff
@@ -216,13 +218,43 @@ void eval(const std::vector<array>& outputs) {
...
+    }
+    depth++;
+    depth_map[id] = depth;
+    for (auto& s : a.siblings()) {
+      depth_map[s.id()] = depth;
+    }
+    return depth;
+  };
+
+  recurse = [&](const array& a, bool largest_branch_first) {
```
jagrit06, line 238: Most of our primitives don't have too many inputs as of yet, so the process to build the tape in eval should be fine for that - but it might be nice to profile the time spent there in some common use cases just to be sure

python/src/array.cpp
```diff
@@ -39,6 +39,10 @@ py::list to_list(array& a, size_t index, int dim) {
 }
 
 auto to_scalar(array& a) {
+  {
+    py::gil_scoped_release nogil;
```
awni, line 43: Does it make sense to add these to the `save*` functions as well? They all do evals.

mlx/ops.h
```diff
@@ -1094,4 +1094,26 @@ array addmm(
...
+        const std::vector<array>&,
+        const std::vector<array>&,
+        const std::vector<array>&)> fun_vjp,
+    const std::vector<array>& args);
+
+/** Checkpoint a function such that to compute its gradient we run the function
+ * again and we don't have to keep all the intermediate states in memory. */
+std::vector<array> checkpoint(
+    std::function<std::vector<array>(const std::vector<array>&)> fun,
+    const std::vector<array>& args);
```
awni, line 1117: This has the semantics of a transform as opposed to an op, I think it makes more sense in `transforms.h`?

python/src/ops.cpp
```diff
@@ -4,6 +4,7 @@
 #include <ostream>
 #include <variant>
 
+#include <pybind11/functional.h>
```
awni, line 7: Any reason you added that?

python/src/transforms.cpp
```diff
@@ -10,6 +10,7 @@
 
 #include "mlx/array.h"
 #include "mlx/graph_utils.h"
+#include "mlx/ops.h"
```
awni, line 13: nit: remove if you make `checkpoint` a transform

mlx/transforms.cpp
```diff
@@ -216,13 +218,43 @@ void eval(const std::vector<array>& outputs) {
...
+    std::vector<int> recursion_order(a.inputs().size());
+    std::iota(recursion_order.begin(), recursion_order.end(), 0);
+    std::sort(
+        recursion_order.begin(),
+        recursion_order.end(),
+        [&a, &depth_map, largest_branch_first](int left, int right) {
+          int depth_left = depth_map[a.inputs()[left].id()];
+          int depth_right = depth_map[a.inputs()[right].id()];
+          bool large_left = depth_left >= depth_right;
+          return largest_branch_first == large_left;
```
awni, line 252: Nit: the naming here confused me a little, makes it sound like a binary tree. Maybe change `left`, `right` to simply `i` and `j` (or something like that)

mlx/transforms.cpp
```diff
@@ -244,7 +276,8 @@ void eval(const std::vector<array>& outputs) {
     }
   };
 
-  recurse(synchronizer);
+  compute_depths(synchronizer);
+  recurse(synchronizer, false);
```
awni, line 280: Could you explain why `largest_branch_first` starts `false` and then switches to `true`?

mlx/transforms.cpp
```diff
@@ -216,13 +218,43 @@ void eval(const std::vector<array>& outputs) {
...
+    int depth = 0;
+    for (auto& in : a.inputs()) {
+      depth = std::max(depth, compute_depths(in));
+    }
+    depth++;
+    depth_map[id] = depth;
+    for (auto& s : a.siblings()) {
+      depth_map[s.id()] = depth;
+    }
+    return depth;
```
awni, line 235: Did you consider trying to count the aggregate array size of the branch rather than the depth?

awni (changes requested):
@angeloskath this is awesome! I looked at the updated eval but did not check the checkpointing yet. I left a few comments. Would be great to add some benchmarks as well so we understand and have a record of what we are gaining in terms of memory and losing in terms of graph time with this change. 

Another thought which you can feel free to disagree if it doesn't make sense, but it might be safer to split this into two PRs. One for the eval + gil and another for the checkpointing + custom VJP primitives. Maybe they really go together, but since there is a lot going on and the eval changes are quite critical, it will be easier to revert / pinpoint issues just in case.

awni:
Btw another option which I think you will hate but would be faster for eval (if it's an issue) is to have a depth counter on the array itself. From an implementation standpoint it is not that messy (array adds one to its inputs depth, set depth to 0 on detach).

awni:
Depth tracking looks nice actually! Simplifies the eval considerably :), hopefully it got rid of most of the eval overheads

mlx/ops.h
```diff
@@ -1105,4 +1105,26 @@ array addmm(
...
+    const std::vector<array>& dependencies);
+
+/** Allows to have call a function and define a custom VJP for it. */
+std::vector<array> custom_vjp(
+    std::function<std::vector<array>(const std::vector<array>&)> fun,
+    std::function<std::vector<array>(
+        const std::vector<array>&,
+        const std::vector<array>&,
+        const std::vector<array>&)> fun_vjp,
+    const std::vector<array>& args);
```
awni, line 1122: this is super cool! custom vjps ✅

awni: I would probably say this also is more like a transform than an op though. We may also just want a new file..

It's composable right? Like I could wrap a custom vjp of a function in another function and make a custom vjp (or transform it)?

angeloskath:
I am attaching some speed stats for the change. I don't think there is eval overhead anymore rather different access patterns due to the evaluation order being different. TL;DR small perf benefit for the GPU and small regression on the MNIST cpu benchmark with many layers.

```
Transformer LM GPU (it/s)

Batch | Before | After
------+--------+------
 2    | 5.258  | 5.279
 4    | 2.844  | 2.857
 8    | 1.481  | 1.486
16    | 0.736  | 0.746

Transformer LM CPU (it/s)

Batch | Before | After
------+--------+------
 1    | 0.671  | 0.673
 2    | 0.430  | 0.430
 4    | 0.236  | 0.235

MNIST CPU (s/epoch average of all 10)

Layers | Before | After
-------+--------+------
  2    | 0.074  | 0.073
  4    | 0.094  | 0.094
  8    | 0.126  | 0.132
 16    | 0.185  | 0.199

MNIST GPU (s/epoch average of all 10)

Layers | Before | After
-------+--------+------
  2    | 0.152  | 0.153
  4    | 0.190  | 0.195
  8    | 0.264  | 0.259
 16    | 0.442  | 0.434
```

mlx/ops.h
angeloskath: Transformations like getting the vjp of course works. Vmap should be easy but I didn't implement it. We just need to vmap the passed functions.

Yeah I think both this and checkpoint should go in transforms.

angeloskath:
One more thing that I think is pretty cool :-) LoRA finetuning on the full 32 layers.

The memory is almost the same as when doing inference like 10 more GB or sth. It increases after a while due to the caching allocator (it's hard to have memory pressure in the M2 Ultra :-) ). I disabled the cache and it is very consistent <50GB no change from inference.

```
(mlx) akatharopoulos@mlx-m2-macstudio-0:~/dev/mlx-examples/llms (main)
↳ python -m mlx_lm.lora --model /tmp/mistral-fp16/ --data ../lora/data/ --lora-layers 32 --batch-size 4 --train --checkpoint
Loading pretrained model
Total parameters 7245.140M
Trainable parameters 3.408M
Loading datasets
Training
Starting training..., iters: 1000
Iter 1: Val loss 2.331, Val took 15.609s
Iter 10: Train loss 2.122, It/sec 0.595, Tokens/sec 236.770
Iter 20: Train loss 1.501, It/sec 0.560, Tokens/sec 224.438
Iter 30: Train loss 1.360, It/sec 0.530, Tokens/sec 215.787
Iter 40: Train loss 1.225, It/sec 0.590, Tokens/sec 226.239
Iter 50: Train loss 1.132, It/sec 0.582, Tokens/sec 229.315
Iter 60: Train loss 1.053, It/sec 0.595, Tokens/sec 233.254
Iter 70: Train loss 1.090, It/sec 0.507, Tokens/sec 203.514
Iter 80: Train loss 1.091, It/sec 0.527, Tokens/sec 209.517
Iter 90: Train loss 1.079, It/sec 0.602, Tokens/sec 225.924
Iter 100: Train loss 1.122, It/sec 0.578, Tokens/sec 217.987
Iter 110: Train loss 0.997, It/sec 0.614, Tokens/sec 235.025
Iter 120: Train loss 0.956, It/sec 0.560, Tokens/sec 216.283
Iter 130: Train loss 0.992, It/sec 0.581, Tokens/sec 227.525
Iter 140: Train loss 0.995, It/sec 0.577, Tokens/sec 222.622
Iter 150: Train loss 0.970, It/sec 0.519, Tokens/sec 209.280
Iter 160: Train loss 0.908, It/sec 0.565, Tokens/sec 219.820
Iter 170: Train loss 0.946, It/sec 0.577, Tokens/sec 235.551
Iter 180: Train loss 0.906, It/sec 0.596, Tokens/sec 214.070
Iter 190: Train loss 0.952, It/sec 0.576, Tokens/sec 228.200
Iter 200: Train loss 0.998, It/sec 0.583, Tokens/sec 219.485
Iter 200: Val loss 1.037, Val took 15.655s
Iter 200: Saved adapter weights to adapters.npz.
Iter 210: Train loss 0.965, It/sec 0.612, Tokens/sec 230.304
Iter 220: Train loss 0.903, It/sec 0.600, Tokens/sec 236.875
Iter 230: Train loss 0.894, It/sec 0.574, Tokens/sec 221.001
Iter 240: Train loss 0.859, It/sec 0.574, Tokens/sec 217.876
Iter 250: Train loss 0.886, It/sec 0.526, Tokens/sec 215.773
Iter 260: Train loss 0.831, It/sec 0.572, Tokens/sec 215.109
Iter 270: Train loss 0.762, It/sec 0.565, Tokens/sec 226.375
```

mlx/transforms.cpp
angeloskath: Changed to `i`, `j`. Honestly, I think `left`, `right` was less confusing. Left as in smaller index in the array. Return true if it shouldn't be swapped.

The whole sort API is kinda bad tbh. For instance the above check can lead to crashes because it has larger or equal instead of larger. It is now fixed as well.

python/src/ops.cpp
angeloskath: It was suggested by Pybind to fix some casting error but it's not needed. It wasn't even then.

mlx/transforms.cpp
angeloskath: So the concept of going down the largest branch is that there are more things to detach and less likely that we 'll be keeping things around for a long time (the time it takes to process the big branch).

Obviously it is a heuristic and not a particularly good one at that. One can probably make up graphs where going down the smallest branch is the most efficient memory-wise.

Regarding the reverse direction for the synchronizer that is pretty straightforward. We know that we 'll keep all of these no matter what. As a result start from the smallest graph which will retain in memory the smallest amount of arrays to compute the second smallest graph etc. A good example is the parameter updates. If we were starting from the largest output graph, then we would keep all the gradients in memory. Otoh starting from the smallest we can throw the gradient away immediately cause we were done with the update.

So to sum up, the heuristic is as follows: Given N subgraphs where you keep only the root node of each subgraph. Start from the smallest one but process it largest branch first.

angeloskath:
I think I addressed everything, added some tests in CPP. We probably need some python ones as well but we can add them when we also expose custom vjp in python. @awni re-review whenever convenient.

mlx/transforms.cpp
awni: > Changed to i, j. Honestly, I think left, right was less confusing. Left as in smaller index in the array

I see.. maybe I misunderstand the invariants of `std::sort`, but I thought `i` and `j` were just arbitrary indices, not not necessarily true that `i  < j`.

mlx/transforms.cpp
awni: Your explanation makes sense, I'm still pondering a bit why the strategy is a generally good one. 

Sorting the outputs by reverse depth makes sense, largest last. This is basically reverse params but more principled.

I'm wondering if there is a big difference from doing just that vs also sorting on the children?

Thinking out loud a bit:

- Largest last makes sense when there is a path between outputs for the reasons you said
- When there is no path between the outputs, presumably the order shouldn't matter much (independent sub graphs)
- I'm struggling a bit with to come up with cases where shortest last is helpful. I understand your reasoning from above, so it would imply in the graphs you were looking at we have lot's of cases where the longest path is basically fully detachable but the shortest path branch is not?

mlx/transforms.cpp
angeloskath: Looking back into the docs, it actually is neither. The docs say `comp(*(it + n), *it)` should return `false` for any valid `it` and `n` after sort is done. So I guess if `comp(left, right)` should return `true` if we need to swap.

mlx/transforms.cpp
angeloskath: Precisely, it depends on the large subgraph. For instance, in the following

```
 F   G    H
  \   \    \
A - B - C - D - E
```

doing shortest first keeps `H`, `G` and `F` around before computing `A`.

mlx/transforms.cpp
angeloskath: This may indeed be a more robust heuristic. I think we should consider it as a TODO though cause the current one works and we can keep at it for a while trying new things without much benefit...

mlx/primitives.cpp
```diff
@@ -797,6 +797,43 @@ std::pair<std::vector<array>, std::vector<int>> Cosh::vmap(
   return {{cosh(inputs[0], stream())}, axes};
 }
 
+std::vector<array> CustomVJP::vjp(
+    const std::vector<array>& primals,
+    const std::vector<array>& cotangents,
+    const std::vector<int>& argnums,
+    const std::vector<array>& outputs) {
+  std::vector<array> inputs(primals.begin(), primals.end() - outputs.size());
```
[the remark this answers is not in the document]
angeloskath: Not a dumb question at all. It was the only way I could think of doing depends and `CustomVJP`. I don't think it's too inelegant but I am open to suggestions.

So the custom vjp in pseudocode is implemented as follows (assuming unary functions `f` and `df`):

```
y = f(x)
z = stop_gradient(y)
g = custom_vjp_primitive(df, inputs=[x, z])
assert(g == z)
```

Computing the vjp of `g` wrt `x` (assuming the above was wrapped in a function) would return the following (again in pseudocode):

```
dx, dz = custom_vjp_primitive_vjp(
        primals=[x, z],
        cotangents=[c],
        outputs=[g]
) = df(x), c  # <--- that's the slicing part. getting x from [x, z]
```

It needs `z` as input in order to return it as output and `x` as input in order to compute the `vjp`. It is similar to the `Depends` primitive in that it is an identity primitive that injects dependencies in the graph or affects the transformations.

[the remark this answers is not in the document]
angeloskath: Something to probably change is the possibility of donating `z` to `g` via move shared buffer. Same goes for depends.

mlx/ops.h
```diff
@@ -1105,4 +1105,13 @@ array addmm(
     const float& beta = 1.f,
     StreamOrDevice s = {});
 
+/**
+ * Implements the identity function but allows injecting dependencies to other
+ * arrays. This ensures that these other arrays will also have been computed
+ * when the outputs of this function are computed.
+ */
+std::vector<array> depends(
```
awni, line 1113: I might be wrong about this, but it looks like we could use `depends` as a replacement for `Synchronizer`?

mlx/transforms.cpp
```diff
@@ -713,4 +749,57 @@ std::function<array(const array&)> vmap(
...
+}
+
+std::vector<array> checkpoint(
+    std::function<std::vector<array>(const std::vector<array>&)> fun,
+    const std::vector<array>& args) {
+  auto vjp_fun = [fun](
+                     const std::vector<array>& primals,
+                     const std::vector<array>& cotangents,
+                     const std::vector<array>& outputs) -> std::vector<array> {
+    auto [__, vjps] = vjp(fun, depends(primals, outputs), cotangents);
```
awni, line 798: 🤯  I had no idea you could use underscore there.

python/mlx/nn/utils.py
```diff
@@ -31,3 +34,53 @@ def wrapped_value_grad_fn(*args, **kwargs):
...
+    """
+
+    @wraps(fn)
+    def checkpointable_fn(*args, **kwargs):
+        checkpoint = False
+        module = None
+        if isinstance(args[0], Module):
+            module = args[0]
+            if hasattr(module, "checkpoint"):
+                checkpoint = module.checkpoint
```
awni, line 65: This is intended to be a property you can add to any module right?

So usage:

```python
lin = nn.Linear(10, 10)
lin.checkpoint = True
```

python/mlx/nn/utils.py
```diff
@@ -31,3 +34,53 @@ def wrapped_value_grad_fn(*args, **kwargs):
         return value, grad
 
     return wrapped_value_grad_fn
+
+
+def maybe_checkpoint(fn: Callable):
```
awni, line 39: What's the usage with `maybe_checkpoint`? Why not always checkpoint the module when you use this wrapper?

And if you don't want a check pointed module, don't call it?

Then you don't really need to pass in kwargs or add an attribute to the module which I think adds a lot keep track of in the API.

awni: Concretely, usage like the following seems a bit more intuitive to me.

```
m = nn.chekcpoint(m) if args.checkpoint else m
outputs = m(inputs)
```

And I think we could make it so that it only works with modules to simplify a bit?

awni (commented):
This is really some magic here! I think the core implementation is LGTM.

I have some questions/suggestions about the C++/Python `checkpoint` API and Module level `maybe_checkpoint` API. I might be missing some aspects of the implementation that constrain the usage. But let me know what you think.

mlx/primitives.cpp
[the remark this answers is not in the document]
awni: Thanks for the explanation. Indeed, we should update our `Copy` and other identity primitives to do donation. I forgot about `Copy` when I was adding it. I can do it as a follow on after this lands.

mlx/transforms.cpp
```diff
@@ -713,4 +749,57 @@ std::function<array(const array&)> vmap(
...
+  }
+
+  return array::make_arrays(
+      shapes,
+      dtypes,
+      std::make_shared<CustomVJP>(to_stream(s), std::move(fun_vjp)),
+      inputs);
+}
+
+std::vector<array> checkpoint(
```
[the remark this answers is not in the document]
angeloskath: Yeah I think you are right. I was having a bunch of GIL related issues and thought that calling immediately would somehow help resolve them or sth. I will update it to a proper function transformation. It is how I am using it anyway.

mlx/ops.h
angeloskath: Yes we can. Since we are creating the array manually, we don't even need to have an output. Nobody ever calls eval on the Synchronizer anyway.

mlx/transforms.cpp
angeloskath: It isn't anything special, just a valid C++ identifier. That means next call you can't use it since it is already declared once in the local scope. I wish `std::ignore` would work in this case but since this kind of tuple destructuring is a declaration it can't work...

It reads nice though doesn't it?

python/mlx/nn/utils.py
angeloskath: Yeah. Basically, one can control checkpointing by either setting an instance property or passing the flag as a keyword argument.

python/mlx/nn/utils.py
angeloskath: So I agree that we should add a `nn.checkpoint` that does what `mx.checkpoint` does (what it will do) but for modules.

Regarding `maybe_checkpoint`, I think it is quite common to want to optionally enable checkpointing. For instance the same code may be running models from 100M to 10B parameters and for the 10B you almost certainly need checkpointing but for the 100M you almost certainly don't. It is like having a `Checkpointable` mixin module.

angeloskath: I would do it the way it is in the transformer now. The module is checkpointable and its container passes checkpoint to it or we can loop and activate checkpointing with something like

```python
def set_checkpointing(m):
    if hasattr(m, "checkpoint"):
        m.checkpoint = True

model.apply_to_modules(set_checkpointing)
```

For the LoRA example since we are looping to set LoRA layers I simply added a loop that did the above, something like

```python
for l in model.layers:
    l.checkpoint = True
```

awni: Why not just do something like:

```
if config.checkpoint:
  model.layers = [nn.checkpoint(l) for l in model.layers]
```

You could do the same with `model.apply_to_modules` right?

Is there something that you get with the attribute that is hard to do otherwise?

python/mlx/nn/layers/transformer.py
```diff
@@ -176,10 +179,11 @@ def __init__(
             for i in range(num_layers)
```
awni, line 179: You could do the checkpointing here based on the argument using `nn.checkpoint`?

python/mlx/nn/utils.py
awni: > I would do it the way it is in the transformer now.

It seems like it would work just as well in that example to checkpoint in the constructor based on the flag and wouldn't require the addition of `maybe_checkpoint` and the `checkpoint` attribute...but I might be missing something in terms of `nn.checkpoint` works.

angeloskath: Well ... removing the checkpointing. I know it is kind of unlikely. I am not married to the attribute... I think it might be worth doing something else or even just removing the attribute and relying on the kwarg.

angeloskath:
@awni The C++ interface returning a function makes perfect sense.

For the python one, honestly I think a `checkpointable` decorator basically `maybe_checkpoint` but renamed is a much better experience... I think we should have both the `nn.utils.checkpoint` function that I wrote now and a `nn.utils.checkpointable` decorator that can decorate `__call__` to be a checkpointed call if the keyword argument `checkpoint` is present.

awni:
> For the python one, honestly I think a checkpointable decorator basically maybe_checkpoint but renamed is a much better experience...

I'm not at all against a decorator at all. Another option is to use `mx.checkpoint` to decorate `__call__`. If I understand correctly you don't like that as much because you can't easily toggle it on and off?

python/mlx/nn/utils.py
```diff
@@ -26,8 +29,37 @@ def inner_fn(params, *args, **kwargs):
...
 
+    @wraps(fn)
     def wrapped_value_grad_fn(*args, **kwargs):
         value, grad = value_grad_fn(model.trainable_parameters(), *args, **kwargs)
         return value, grad
 
     return wrapped_value_grad_fn
+
+
+def checkpoint(module: Module, fn: Callable):
```
awni, line 40: Should we make `fn` `Optional` and use `__call__` if it is not set? It seems like that will be the most common case right?

awni: Or maybe better as question: why not hard code `fn` to the module itself and remove it as an argument? Is it a flexibility thing (like I can use `Module.forward`)?

angeloskath: Yeah this goes for using `mx.checkpoint` as a decorator above as well. Since the modules are not pure functions, the `nn.checkpoint` is basically a helper for "Checkpoint this function wrt this module's parameters".

Once could presumably want to checkpoint a function even completely outside of the module. Something like the following:

```python
def decode(model, x):
    ... call model(x) a bunch of times ...
```

Btw I agree that `fn` should default to `None` that gets set to `__call__`.

angeloskath:
@awni should be good to merge.

python/mlx/nn/utils.py
```diff
@@ -26,8 +29,42 @@ def inner_fn(params, *args, **kwargs):
...
+        fn (Callable, optional): The function to checkpoint. If not provided it
+            defaults to the provided module.
+
+    Returns:
+        A callable that saves the inputs and outputs during the forward pass
+        and recomputes all intermediate states during the backward pass.
+    """
+    if fn is None:
+        # Capturing module instead of module.__call__ allows someone to
+        # monkey-patch __call__ later on and the correct method will be used
```
awni, line 57: 👍

awni (approved):
🚀

PR #543, opened 2024-01-24
More helpful error message in vjp transform + concate bug

### LINKED ISSUE #487: ValueError invoked by mlx.core.grad
**Describe the bug**
When I tried to calculate the gradient of x * * 2 at x=3 using mlx.core.grad, I encountered this problem and wanted to know if it was my method of using the function that was problematic because I saw the document saying, 'Returns: A function which has the same input arguments as fun and returns the gradient (s).' According to the document, there seems to be no problem with my usage method.

**To Reproduce**

Include code snippet
```python
import mlx.core as mx

def square(x):
    return x ** 2
    
print(square((mx.array([3], dtype=mx.float32))))
grad_square = mx.grad(square)

gradient_at_3 = grad_square(mx.array([3], dtype=mx.float32))

print(gradient_at_3)
```
image

**Expected behavior**
The result is array ([6], dtype=float32)

 - OS Version: MacOS 14.2.1
 - Version 0.7.0

### LINKED ISSUE #540: There is no crash when two arrays of different dimensions are used as inputs to mlx.core.concatenate.
**Describe the bug**
There is no crash when two arrays of different dimensions are used as inputs to mlx.core.concatenate. According to the principle, two arrays with different dimensions cannot be concatenated. The implementation of mlx, numpy and tensorflow is as follows:

**To Reproduce**

Include code snippet
```python
import mlx.core as mx

array1 = mx.array([[1, 2, 3],[1, 2, 3],[1, 2, 3]])
array2 = mx.array([4, 5, 6])
axis = 0
concatenated = mx.concatenate([array1, array2], axis=axis)
print(concatenated)
```
image

```python
import numpy as np

np_array1 = np.array([[1, 2, 3],[1, 2, 3],[1, 2, 3]])
np_array2 = np.array([4, 5, 6])
axis = 0
np_concatenated = np.concatenate([np_array1, np_array2], axis=axis)
print(np_concatenated)
```
image

```python
import tensorflow as tf

tf_tensor1 = tf.constant([[1, 2, 3],[1, 2, 3],[1, 2, 3]])
tf_tensor2 = tf.constant([4, 5, 6])
tf_concatenated = tf.concat([tf_tensor1, tf_tensor2], axis=0)
print(tf_concatenated)

```
image

**Expected behavior**
A dimension check should be added here to report a crash.

 - OS Version: MacOS 14.2.1
 - Version 0.10.0
 - numpy 1.24.3
 - tensorflow 2.13.0

### LINKED ISSUE #542: error message is useless when using 'mlx.core.concatenate'
**Describe the bug**
After the test at https://github.com/ml-explore/mlx/issues/540, I tried to change axis to 1 to see if the code would work, but I found that the crash error message was useless. And I can make the code work properly with the reshape operation, where I think the error messages can be a little clearer.

**To Reproduce**

Include code snippet
```python
import mlx.core as mx

array1 = mx.array([[1, 2, 3],[1, 2, 3],[1, 2, 3]])
array2 = mx.array([4, 5, 6])
axis = 1
concatenated = mx.concatenate([array1, array2], axis=axis)
print(concatenated)
```
image

```python
import mlx.core as mx

array1 = mx.array([[1, 2, 3], [1, 2, 3], [1, 2, 3]])
array2 = mx.array([4, 5, 6])
array2_reshaped = mx.reshape(array2, (3, 1))
axis = 1
concatenated = mx.concatenate([array1, array2_reshaped], axis=axis)
print(concatenated)
```
image

**Expected behavior**
A clear error message

 - OS Version: MacOS 14.2.1
 - Version 0.10.0

awni:
## Proposed changes

Closes #487 

```
import mlx.core as mx

def f(x):
    return x.sum(keepdims=True)

mx.eval(mx.grad(f)(mx.array([0, 1])))
```

Will now print:

```
ValueError: [vjp] Output shape (1) does not cotangent shape (). If you are using grad your function must return a scalar.
```

Also closes #540 which was a bug that let you concatenate arrays with different number of dims. Added test for that as well.

Also closes #542 with a message like:

```
ValueError: [concatenate] All the input arrays must have the same number of dimensions. However, got arrays with dimensions 2 and 1.
```

### DIFF
--- a/mlx/transforms.cpp
+++ b/mlx/transforms.cpp
@@ -337,12 +337,20 @@ std::pair<std::vector<array>, std::vector<array>> vjp(
       }
     }
     if (cotan_index >= cotans.size()) {
-      throw std::invalid_argument(
-          "[vjp] Number of outputs with gradient does not match number of cotangents.");
+      std::ostringstream msg;
+      msg << "[vjp] Number of outputs to compute gradients for ("
+          << outputs.size() << ") does not match number of cotangents ("
+          << cotans.size() << ").";
+      throw std::invalid_argument(msg.str());
     }
     if (out.shape() != cotans[cotan_index].shape()) {
-      throw std::invalid_argument(
-          "[vjp] Output shape does not match shape of cotangent.");
+      std::ostringstream msg;
+      msg << "[vjp] Output shape " << out.shape()
+          << " does not cotangent shape " << cotans[cotan_index].shape() << ".";
+      if (outputs.size() == 1 && out.size() == 1) {
+        msg << " If you are using grad your function must return a scalar.";
+      }
+      throw std::invalid_argument(msg.str());
     }
     output_cotan_pairs.emplace_back(i, cotan_index++);
   }
--- a/mlx/transforms.cpp
+++ b/mlx/transforms.cpp
@@ -346,7 +346,8 @@ std::pair<std::vector<array>, std::vector<array>> vjp(
     if (out.shape() != cotans[cotan_index].shape()) {
       std::ostringstream msg;
       msg << "[vjp] Output shape " << out.shape()
-          << " does not cotangent shape " << cotans[cotan_index].shape() << ".";
+          << " does not match cotangent shape " << cotans[cotan_index].shape()
+          << ".";
       if (outputs.size() == 1 && out.size() == 1) {
         msg << " If you are using grad your function must return a scalar.";
       }
--- a/mlx/ops.cpp
+++ b/mlx/ops.cpp
@@ -669,26 +669,27 @@ array concatenate(
     int axis,
     StreamOrDevice s /* = {} */) {
   if (arrays.size() == 0) {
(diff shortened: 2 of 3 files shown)


mlx/transforms.cpp
```diff
@@ -337,12 +337,20 @@ std::pair<std::vector<array>, std::vector<array>> vjp(
...
+          << outputs.size() << ") does not match number of cotangents ("
+          << cotans.size() << ").";
+      throw std::invalid_argument(msg.str());
     }
     if (out.shape() != cotans[cotan_index].shape()) {
-      throw std::invalid_argument(
-          "[vjp] Output shape does not match shape of cotangent.");
+      std::ostringstream msg;
+      msg << "[vjp] Output shape " << out.shape()
+          << " does not cotangent shape " << cotans[cotan_index].shape() << ".";
```
angeloskath, line 349: typo: does not **match**

angeloskath (approved):

PR #558, opened 2024-01-26
Added deepcopy implementation.

hazemessamm:
## Proposed changes

Added `deepcopy` implementation to create more than one copy for the same model.

## Checklist

- [] I have updated the necessary documentation (if needed)

### DIFF
--- a/python/mlx/nn/layers/base.py
+++ b/python/mlx/nn/layers/base.py
@@ -1,12 +1,11 @@
 # Copyright © 2023 Apple Inc.
 
 import textwrap
+from copy import deepcopy
 from typing import Any, Callable, List, Optional, Tuple, Union
 
 import mlx.core as mx
 from mlx.utils import tree_flatten, tree_unflatten
-from copy import deepcopy
-
 
 
 class Module(dict):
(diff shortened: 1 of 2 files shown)


python/mlx/nn/layers/base.py
```diff
@@ -90,6 +91,34 @@ def __getattr__(self, key: str):
...
+        cls = self.__class__
+        instance_copy = cls.__new__(cls)
+        memo[id(self)] = instance_copy
+
+        for k, v in self.items():
+            if isinstance(v, mx.array):
+                obj = mx.array(v)
+            elif isinstance(v, list):
+                obj = [
+                    mx.array(elem) if isinstance(elem, mx.array) else deepcopy(elem)
```
hazemessamm, line 104: This solution might not be the best but I did not find a method that copies `mx.array` so I just copied it by calling `mx.array(old_mx_array)` to create a new array.

angeloskath:
@hazemessamm out of curiosity why would we want a deep copy?

In general weights are not updated in place so keeping a copy of the parameters dictionary means you have a copy of the weights. Even if someone changes them. Basically what does this PR make possible with the deep copy except for the ability to do something like

```
linear.weight *= 2
```

Keep in mind that doing

```
old_weights = linear.parameters()
linear.weight = linear.weight * 2
```

does not affect `old_weights` at all.

hazemessamm:
Thanks for the response and explanatikn I did not know that, I was working on an Exponential Moving Average module where I wanted to take a deep copy of the model so I can update it and avoid changing anything in the original model. Do you think is there a way to do that without deep copying the model?

angeloskath:
Oh yeah exponential moving average would be extremely easy to do in MLX. Literally the following code:

```python
def ema_update(ema_params, model, alpha=0.9):
    return tree_map(lambda a, b: a * alpha + (1-alpha) * b, ema_params, model.parameters())

ema_parameters = model.parameters()
for batch in iterate_batches(data):
    loss, grad = loss_and_grad(model, batch)
    optimizer.update(model, grad)
    del grad  # to save some memory cause we don't need it any more
    ema_parameters = ema_update(ema_parameters, model)
    mx.eval(model.parameters(), ema_parameters, optimizer.state)
```

To use the ema weights afterwards one needs simply to do `model.update(ema_parameters)` and then use the model normally.

I will close this PR since I think a deep copy is almost never needed and for sure not for EMA :-) . Feel free to reopen if you think there is something that needs a deep copy.

Issue #557, opened 2024-01-26
mlx.core.flatten lacks dimension check


Redmept1on:
**Describe the bug**
When I tried to use the mlx.core.flatten operation on two-dimensional arrays, there was no crash when I set end_axis to more than 2.

**To Reproduce**

Include code snippet
```python
import mlx.core as mx

# Create an array for flattening
array_a = mx.array([[1, 2], [3, 4]])

flattened = mx.flatten(array_a, start_axis=0, end_axis=6)
print(flattened)
```
![image](https://github.com/ml-explore/mlx/assets/84373897/06fc43fe-69d0-408f-8a6f-95b1d29428dc)

In addition, there is another problem, which I think is error message. The default start_axis is 0 and end_axis is -1. When I try to set start_axis to 2, the following error occurs, I think there is a problem with this error message. Because the default value start_axis is less than end_axis.

```python
import mlx.core as mx

# Create an array for flattening
array_a = mx.array([[1, 2], [3, 4]])

flattened = mx.flatten(array_a, start_axis=2, end_axis=-1)
print(flattened)
```
![image](https://github.com/ml-explore/mlx/assets/84373897/54f75223-f4f9-4d30-98cc-620fe77a23c7)

**Desktop (please complete the following information):**
 - OS Version: MacOS 14.2.1
 - Version 0.10.0

awni:
- We clamp the axis to the beginning and end so if `end_axis` is really large we will clip it. That's intentional, not a bug. I don't know if it's a desirable feature or not, but I will leave it that way for now unless we have a good reason to change it.
- In the second case a -1 axis is treated as `1` so that is why you see that message.

I will update the docs to be more clear to help with this.

Redmept1on:
Thanks for explaining！

PR #559, opened 2024-01-26
mlx.core.flatten lacks dimension check

saurabhaloneai:
issue #557

 previous code :
```
if (end_ax < start_ax) {
    throw std::invalid_argument(
        
        "start_axis must be less than or equal to end_axis");
  } ``
```

after adding the error msg : 

```
if (end_ax < start_ax) {
    throw std::invalid_argument(
        "[flatten] Invalid combination of start_axis and end_axis. "
        "start_axis must be less than or equal to end_axis");
  }
```

-----
ig it will solve the issue 
-----

## Checklist

### DIFF
--- a/mlx/ops.cpp
+++ b/mlx/ops.cpp
@@ -320,7 +320,8 @@ array flatten(
   }
   if (end_ax < start_ax) {
     throw std::invalid_argument(
-        "[flatten] start_axis must be less than or equal to end_axis");
+        "[flatten] Invalid combination of start_axis and end_axis. "
+        "start_axis must be less than or equal to end_axis");
   }
   if (start_ax >= ndim) {
     std::ostringstream msg;


angeloskath (changes requested):
If we are improving the error message we might as well make available the provided `end_ax` and `start_ax` as is done in other ops. Something like `start_axis must be less... to end_axis but start_axis=N and end_axis=M`. You can use the `std::ostringstream` to create the error message.

saurabhaloneai:
Something like this right ?

```
if (start_axis >= end_axis) {
        std::ostringstream errorMessage;
        errorMessage << "Error: start_axis must be less than end_axis, but start_axis=" << start_axis << " and end_axis=" << end_axis << ".";
        
        throw std::invalid_argument(errorMessage.str());
    }
```

angeloskath:
@saurabhaloneai let me know if you 'll work on it or I can close since the current message is not that bad.

saurabhaloneai:
Thank you for your feedback. If the current message meets your expectations, I'm happy to consider the task closed :)

PR #561, opened 2024-01-26
fix cache key in RoPE

davidkoski:
the values were being computed for D but the cache key was D // 2

## Proposed changes

the values were being computed for `D` but the cache key was `D // 2`.  This makes a new variable for `half_D` so that it can use the original value of `D` for the cache key.

This should be _slightly_ faster because we will get cache hits and more importantly we won't get false hits if you pass in a D (dimension) which happens to be half of the original D.

Tested with `mistral` from mlx-examples and I see the same results.

## Checklist

### DIFF
--- a/python/mlx/nn/layers/positional_encoding.py
+++ b/python/mlx/nn/layers/positional_encoding.py
@@ -104,9 +104,9 @@ class RoPE(Module):
         dtype=mx.float32,
     ):
         if (N, D, offset, base, scale, dtype) != cls._cos_sin_theta_key:
-            D = D // 2
+            half_D = D // 2
             positions = mx.arange(offset, N, dtype=dtype) * scale
-            freqs = mx.exp(-mx.arange(0.0, D, dtype=dtype) * (math.log(base) / D))
+            freqs = mx.exp(-mx.arange(0.0, half_D, dtype=dtype) * (math.log(base) / half_D))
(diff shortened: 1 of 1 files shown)


python/mlx/nn/layers/positional_encoding.py
```diff
@@ -104,9 +104,9 @@ def create_cos_sin_theta(
         dtype=mx.float32,
     ):
         if (N, D, offset, base, scale, dtype) != cls._cos_sin_theta_key:
-            D = D // 2
+            half_D = D // 2
             positions = mx.arange(offset, N, dtype=dtype) * scale
-            freqs = mx.exp(-mx.arange(0.0, D, dtype=dtype) * (math.log(base) / D))
+            freqs = mx.exp(-mx.arange(0.0, half_D, dtype=dtype) * (math.log(base) / half_D))
             theta = mx.reshape(positions, (-1, 1)) * mx.reshape(freqs, (1, -1))
             cls._cos_sin_theta_key = (N, D, offset, base, scale, dtype)
```
davidkoski, line 111: Note that originally it was using `D // 2` as the cache key here

angeloskath: Nice catch! Thanks.

angeloskath (approved):
This would have gone unnoticed for a while! Thanks!

PR #562, opened 2024-01-26
Implement diagonal operator

### LINKED ISSUE #503: [Feature] `mx.diag`

PRESIDENT810:
## Proposed changes

This implements mx.diagonal in operator level, inspired by @ManishAradwad.

The API strictly follows numpy's design:
```
Return specified diagonals
If `a` is 2-D, then a 1-D array containing the diagonal and of the
same type as `a` is returned unless `a` is a `matrix`, in which case
a 1-D array rather than a (2-D) `matrix` is returned in order to
maintain backward compatibility.
If ``a.ndim > 2``, then the dimensions specified by `axis1` and `axis2`
are removed, and a new axis inserted at the end corresponding to the
diagonal.
Args:
  a (array): Input array
  offset (int, optional): Offset of the diagonal from the main diagonal.
    Can be positive or negative. Default: ``0``.
  axis1 (int, optional): Axis to be used as the first axis of the 2-D sub-arrays from which
      the diagonals should be taken. Default: ``0``.
  axis2 (int, optional): Axis to be used as the second axis of the 2-D sub-arrays from which
      the diagonals should be taken. Default: ``1``.
Returns:
    array: The diagonal array.
```

Closes #503 

## Checklist


ManishAradwad:
I'm stuck at setting up the C++ API of MLX. I followed the steps given [here](https://ml-explore.github.io/mlx/build/html/install.html#c-api), but in the C++ mlx code I write, I get a "cannot include mlx" error at `#include <mlx>`. Do I have to do any other steps after the instructions at the above link?

PRESIDENT810:
@ManishAradwad Which file do you want to add `#include <mlx>`? You can't do that in a header file if `mlx.h` already have included that (which causes a cyclic inclusion). The way C++'s `include` works is similar to copy and paste that file's content into where you use `include`. So if you write `#include <mlx>` in for example, `mlx/ops.h`, because `mlx.h` includes `mlx/ops.h`, there will be problems.

You can just add your `array diag(const array& a, int k, StreamOrDevice s) ` below my `diagonal` function and I don't you need to include other things. And `diag`'s test should go into `ops_test.cpp` below my `diagonal` tests.

ManishAradwad:
No, that's fine. I'm talking about using the mlx library in a cpp code. For instance I want to run a test.cpp file in which I wrote some cpp code using mlx.

PRESIDENT810:
@ManishAradwad
I would suggest you just run the C++ tests within mlx repo. If you want to use mlx library in a separate project you will need to link to libmlx, which is a bit complex.

To run mlx's C++ tests you can just follow the instruction and do `make test`.
Or if you are using IDE with opening mlx as a CMake project, you can just choose to run the CTest target in your IDE.

ManishAradwad:
Added the `diag` implementation with tests

PRESIDENT810:
@ManishAradwad Awesome! Thanks for your great work!

There's only one last thing I'm not sure. Looks like numpy only has `numpy.diag` function but not `numpy.ndarray.diag`, in contrast `diagonal` has been both implemented as a numpy function and a ndarray method. @awni Could you please share some thought on this? Do we need to strictly align with numpy's API and keep only `mx.diag`, or make it a method on `mx.array` too?

ManishAradwad:
just tried the `diag` method on `mx.array` and it works just like `mx.diag(a)`

Screenshot 2024-01-28 at 1 05 06 PM

awni (approved):
Thanks for adding this!! Looks great!

I made a few changes to the diagonal implementation to simplify and make it more efficient. You can check it if you are interested. (It helps to know how our `gather` works 😄 , I will try and add some documentation to the C++ function).

awni (approved):

PR #565, opened 2024-01-27
No gil eval

angeloskath:
There are a couple of reasons for releasing the GIL every time we are in `eval`.

Besides being able to use the `Load` primitive with python streams, primitives that keep around a python object need to get the GIL in the destructor in order to properly release their resources without crashing.

Since we are releasing it in eval I went ahead and set the loading from `npz` to be lazy if we own the python stream. Initially I wanted to check reference counts to check if we own the stream passed as argument but that in Python seems pretty impossible, passing `mx.load(open(...))` the stream ends up with 3 references somehow (I guess 1 the call 1 pybind and the 1 that should have been).

### DIFF
--- a/python/src/array.cpp
+++ b/python/src/array.cpp
@@ -39,6 +39,10 @@ py::list to_list(array& a, size_t index, int dim) {
 }
 
 auto to_scalar(array& a) {
+  {
+    py::gil_scoped_release nogil;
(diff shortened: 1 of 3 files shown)


awni (approved):
Lazy loading, nice!! LGTM 🚀

Issue #293, opened 2023-12-26
Add int64 reductions


klokare:
Ran into an issue with the sum function when using int64.

the following runs ok

```python
import mlx.core as mx
y = mx.ones((70), dtype=mx.int32)
mx.sum(y)
```

the following crashes every time on my MacBook Pro M3 Pro / 18GB

```python
import mlx.core as mx
y = mx.ones((70), dtype=mx.int64)
mx.sum(y)
```

with error message "libc++abi: terminating due to uncaught exception of type std::runtime_error: [Reduce::eval_gpu] Does not support int64"

can you confirm this is intended behaviour? or if it something that is planned for a future version?

awni:
Yes, int64 is not yet supported in our reductions, that's expected. It's definitely possible to add, but not the highest priority at the moment. Happy to take a contribution for this though if someone is interested in adding it!

vj-krish:
@awni we can close this now.

awni:
Thanks @vj-krish !!

angeloskath:
Fixed in #569.

PR #569, opened 2024-01-28
Add GPU support for uint64/int64 reductions

vj-krish:
Decompose reductions into 2 compute dispatches
since Metal doesn't have full support for atomics
with int64, uint64.

Github Issue #293.

## Proposed changes

This change adds support for GPU reduction operators on arrays whose element type is int64/uint64. Please see GitHub issue #293 

## Checklist

### DIFF
--- a/mlx/backend/metal/kernels/reduce.metal
+++ b/mlx/backend/metal/kernels/reduce.metal
@@ -29,110 +29,12 @@ template <typename T, typename Op>
 ///////////////////////////////////////////////////////////////////////////////
 
 template <typename T, typename U, typename Op, int N_READS=REDUCE_N_READS>
-[[kernel]] void all_reduce(
-    const device T *in [[buffer(0)]],
-    device mlx_atomic<U> *out [[buffer(1)]],
-    const device size_t& in_size [[buffer(2)]],
-    uint gid [[thread_position_in_grid]],
-    uint lid [[thread_position_in_threadgroup]],
-    uint grid_size [[threads_per_grid]],
-    uint simd_per_group [[simdgroups_per_threadgroup]],
-    uint simd_lane_id [[thread_index_in_simdgroup]],
-    uint simd_group_id [[simdgroup_index_in_threadgroup]]) {
-  // NB: this kernel assumes threads_per_threadgroup is at most
-  // 1024. This way with a simd_size of 32, we are guaranteed to
-  // complete the reduction in two steps of simd-level reductions.
-
+inline U per_thread_all_reduce(
+    const device T *in,
+    const device size_t& in_size,
+    uint gid,
+    uint grid_size) {
   Op op;
-  threadgroup U local_vals[simd_size];
-      
-  U total_val = Op::init;
-
-  in += gid * N_READS;
-      
-  int r = 0;
-  for(; r < (int)ceildiv(in_size, grid_size * N_READS) - 1; r++) {
-    U vals[N_READS] = {op.init}; 
-
-    for(int i = 0; i < N_READS; i++) {
-      vals[i] = static_cast<U>(in[i]);
-    }
-    for(int i = 0; i < N_READS; i++) {
-      total_val = op(vals[i], total_val);
-    }
-
-    in += grid_size * N_READS;
-  }
-
-  // Separate case for the last set as we close the reduction size 
-  size_t curr_idx = (gid + r * (size_t)grid_size) * N_READS;
-  if (curr_idx < in_size) {
-    int max_reads = in_size - curr_idx;
-    T vals[N_READS];
-
-    for(int i = 0, idx = 0; i < N_READS; i++, idx++) {
-      idx = idx < max_reads ? idx : max_reads - 1;
-      vals[i] = in[idx];
-    }
-    for(int i = 0; i < N_READS; i++) {
-      U val = i < max_reads ? vals[i] : Op::init; 
-      total_val = op(static_cast<U>(val), total_val);
-    }
-  }
-
-  // Reduction within simd group
-  total_val = op.simd_reduce(total_val);
-  if (simd_lane_id == 0) {
-    local_vals[simd_group_id] = total_val;
-  }
-  
-  // Reduction within thread group
-  threadgroup_barrier(mem_flags::mem_threadgroup);
-  total_val = lid < simd_per_group ? local_vals[lid] : op.init;
-  total_val = op.simd_reduce(total_val);
-
-  // Reduction across threadgroups
-  if (lid == 0) {
-    op.atomic_update(out, total_val);
-  }
-}
-
-#define instantiate_all_reduce(name, itype, otype, op) \
-  template [[host_name("all_reduce_" #name)]] \
-  [[kernel]] void all_reduce<itype, otype, op>( \
-      const device itype *in [[buffer(0)]], \
-      device mlx_atomic<otype> *out [[buffer(1)]], \
-      const device size_t& in_size [[buffer(2)]], \
-      uint gid [[thread_position_in_grid]], \
-      uint lid [[thread_position_in_threadgroup]], \
-      uint grid_size [[threads_per_grid]], \
-      uint simd_per_group [[simdgroups_per_threadgroup]], \
-      uint simd_lane_id [[thread_index_in_simdgroup]], \
-      uint simd_group_id [[simdgroup_index_in_threadgroup]]);
-
-///////////////////////////////////////////////////////////////////////////////
-// All reduce without atomics
-///////////////////////////////////////////////////////////////////////////////
-
-template <typename T, typename U, typename Op, int N_READS=REDUCE_N_READS>
-[[kernel]] void all_reduce_no_atomics(
-    const device T *in [[buffer(0)]],
-    device U *out [[buffer(1)]],
-    const device size_t& in_size [[buffer(2)]],
-    uint gid [[thread_position_in_grid]],
-    uint lid [[thread_position_in_threadgroup]],
-    uint grid_size [[threads_per_grid]],
-    uint simd_per_group [[simdgroups_per_threadgroup]],
-    uint simd_lane_id [[thread_index_in_simdgroup]],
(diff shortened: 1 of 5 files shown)


angeloskath:
@vj-krish from a first skim that looks awesome!

As you mentioned they are quite separated although there is huge possibility of code reuse. From a quick look the kernels are the same except for using atomics for writing the output. I would add a template parameter and an if statement to route to each case inside the kernel.

The dispatch may be more involved. Feel free to add the results of a couple of runs similar to the one below

```
python benchmarks/python/comparative/bench_mlx.py sum_axis --size 1024x1024 --axis 1 --dtype int64
```

It should be approximately twice slower than int32 for most cases.

vj-krish:
> As you mentioned they are quite separated although there is huge possibility of code reuse. From a quick look the kernels are the same except for using atomics for writing the output. I would add a template parameter and an if statement to route to each case inside the kernel.

Yeah, I'll push a follow up commit that refactors for re-use. Pushed this as the first commit, so it's easier to review the actual change stand alone.

> python benchmarks/python/comparative/bench_mlx.py sum_axis --size 1024x1024 --axis 1 --dtype int64
Sure thing, I'll do that!

>

vj-krish:
**Row reduce**
```
python benchmarks/python/comparative/bench_mlx.py sum_axis --size 1024x1024x1024 --axis 2 --dtype int32
124.9713659286499

python benchmarks/python/comparative/bench_mlx.py sum_axis --size 1024x1024x1024 --axis 2 --dtype int64
209.6609399318695
```

**Col Reduce**

```
python benchmarks/python/comparative/bench_mlx.py sum_axis --size 1024x1024x1024 --axis 0,1 --dtype int32
98.03353309631348

python benchmarks/python/comparative/bench_mlx.py sum_axis --size 1024x1024x1024 --axis 0,1 --dtype int64
232.3654432296753
```

**All Reduce**

```
python benchmarks/python/comparative/bench_mlx.py sum_axis --size 1024x1024x1024 --axis 0,1,2 --dtype int32
115.88422560691833

python benchmarks/python/comparative/bench_mlx.py sum_axis --size 1024x1024x1024 --axis 0,1,2 --dtype int64
228.8528048992157
```

mlx/backend/metal/kernels/reduce.metal
```diff
@@ -212,6 +306,217 @@ template <typename T, typename U, typename Op, int N_READS=REDUCE_N_READS>
...
+
+  // Prepare next level
+  if (simd_lane_id == 0) {
+    local_vals[simd_group_id] = total_val;
+  }
+  threadgroup_barrier(mem_flags::mem_threadgroup);
+
+  // Reduction within thread group
+  // Only needed if thread group has multiple simd groups
+  if(ceildiv(reduction_size, N_READS) > simd_size) {
```
vj-krish, line 317: I've changed this from,

`if (reduction_size > sims_size)`

since each thread reduces N_READS element.

mlx/backend/metal/kernels/reduce.metal
```diff
@@ -212,6 +306,217 @@ template <typename T, typename U, typename Op, int N_READS=REDUCE_N_READS>
...
+    uint3 lid,
+    uint gid_x,
+    uint3 lsize,
+    uint3 gsize) {
+
+  Op op;
+  U total_val = Op::init;
+
+  uint base_offset = (tid.y * lsize.y + lid.y) * N_READS;
+  for(uint r = 0; r < N_READS && (base_offset + r) < reduction_size; r++) {
```
vj-krish, line 383: Previously, there were 2 loops here. I've fused them into 1 in this change.

mlx/backend/metal/kernels/reduce.metal
```diff
@@ -111,6 +110,101 @@ template <typename T, typename U, typename Op, int N_READS=REDUCE_N_READS>
...
+  // NB: this kernel assumes threads_per_threadgroup is at most
+  // 1024. This way with a simd_size of 32, we are guaranteed to
+  // complete the reduction in two steps of simd-level reductions.
+
+  Op op;
+  threadgroup U local_vals[simd_size];
+
+  U total_val = Op::init;
+
+  if (gid * N_READS < in_size) {
```
vj-krish: Added this check to ensure we don't go OOB, given we are rounding thread group size to simd width.

mlx/backend/metal/kernels/reduce.metal
```diff
@@ -24,11 +24,59 @@ template <typename T, typename Op>
...
+template <typename T, typename U, typename Op, int N_READS=REDUCE_N_READS>
+inline U per_thread_all_reduce(
+    const device T *in,
+    const device size_t& in_size,
+    uint gid,
+    uint grid_size) {
+  Op op;
+  U total_val = Op::init;
+
+  if (gid * N_READS < in_size) {
```
vj-krish, line 40: Added this check to ensure we don't go OOB, given we are rounding thread group size to simd width.

mlx/backend/metal/reduce.cpp
```diff
@@ -36,37 +41,88 @@ void all_reduce_dispatch(
...
+      ((thread_group_size + simd_size - 1) / simd_size) * simd_size;
 
   // If the number of thread groups needed exceeds 1024, we reuse threads groups
   uint n_thread_groups = safe_div(mod_in_size, thread_group_size);
   n_thread_groups = std::min(n_thread_groups, 1024u);
   uint nthreads = n_thread_groups * thread_group_size;
 
+  // Allocate intermediate array to store partial reduction results from all
+  // thread groups
+  array* output = &out;
```
angeloskath, line 77: No need for that or the pointer to the intermediate array. See https://github.com/ml-explore/mlx/blob/65d0b8df9fb02a30445fc5431aa2d8e73ca0af2f/mlx/backend/metal/matmul.cpp#L251 for how to allocate an intermediate array.

mlx/backend/metal/reduce.cpp
```diff
@@ -36,37 +41,88 @@ void all_reduce_dispatch(
...
+  set_array_buffer(compute_encoder, in, 0);
+  set_array_buffer(compute_encoder, *output, 1);
+  compute_encoder->setBytes(&in_size, sizeof(size_t), 2);
+
   MTL::Size group_dims = MTL::Size(thread_group_size, 1, 1);
   MTL::Size grid_dims = MTL::Size(nthreads, 1, 1);
 
   compute_encoder->dispatchThreads(grid_dims, group_dims);
+
+  if (is_out_64b_int) {
```
angeloskath, line 102: I think this would simplify quite a lot if it was clearly split up in two separate cases. Normal and atomic free. The normal one should be the same as before and for the atomic free you would allocate a new array and call the kernel twice one in with the intermediate and one with the out.

Importantly, the second call is always a single threadgroup! Otherwise there is a bug. So no need for the thread group math. You could check that it indeed fits in a single threadgroup if you want for extra safety. Also the 2nd call can be avoided if the input array is smaller than `1024 * N_READS` which is quite a few cases.

One more important thing is that the intermediate array is currently freed as soon as the function exits (which is before the kernel is run). The way we deal with intermediate arrays is we hold a copy in a completion handler and we clear it so that it gets freed after the kernel is run. See https://github.com/ml-explore/mlx/blob/65d0b8df9fb02a30445fc5431aa2d8e73ca0af2f/mlx/backend/metal/matmul.cpp#L251 or https://github.com/ml-explore/mlx/blob/65d0b8df9fb02a30445fc5431aa2d8e73ca0af2f/mlx/backend/metal/softmax.cpp#L80 for examples.

angeloskath (changes requested):
The kernels look great!

I went through the all reduce dispatch and left some comments. I will go through the rest probably tomorrow.

mlx/backend/metal/reduce.cpp
vj-krish: I did that so we conditionally create an intermediate array, since we don't need it for the path that uses atomics.

mlx/backend/metal/reduce.cpp
```diff
@@ -101,10 +164,24 @@ void row_reduce_general_dispatch(
   }
   int ndim = shape.size();
 
+  array* output = &out;
+
+  // Allocate intermediate array to store partial reduction results
+  std::unique_ptr<array> intermediate;
```
angeloskath, line 170: Same as above regarding freeing the intermediate.

mlx/backend/metal/reduce.cpp
```diff
@@ -2,6 +2,7 @@
 
 #include <algorithm>
 #include <cassert>
+#include <memory>
```
angeloskath, line 5: I assume this is for the `unique_ptr` which shouldn't be needed.

angeloskath (changes requested):
The row and col reduce look great as well. The same general comments. The intermediate array allocation and free needs to be changed and I would make the two branches of clearer in the dispatch code.

Thanks for adding these they look and perform great!

mlx/backend/metal/reduce.cpp
vj-krish: > I think this would simplify quite a lot if it was clearly split up in two separate cases. Normal and atomic free. The normal one should be the same as before and for the atomic free you would allocate a new array and call the kernel twice one in with the intermediate and one with the out.

Sure, I can do that. But I think there's going to some amount of duplication if we go down that path.

> Importantly, the second call is always a single threadgroup! Otherwise there is a bug. So no need for the thread group math. You could check that it indeed fits in a single threadgroup if you want for extra safety. Also the 2nd call can be avoided if the input array is smaller than 1024 * N_READS which is quite a few cases.

Ack, that's a good catch. I'll address that.

> One more important thing is that the intermediate array is currently freed as soon as the function exits (which is before the kernel is run).
Ack. I'm guessing this is for required so the reuse allocator in MLX doesn't recycle it until prior use is complete (since the underlying Metal resource should be retained by the driver once we bind it.) I'll add a completion handler block .

angeloskath (approved):
Looks great, tests pass! Thanks for the contribution!

We can simplify a bit more when we do an optimization pass for small reductions (that's not int64 specific all types).

PR #571, opened 2024-01-28
Compile primitive

awni:
## Proposed changes

Compiled primitive with primitive-level fusion

- [x] Handle streams and max depth
- [x] Implement compile  vmap
- [x] Implement Compile vjp and jvp
- [x] Do the right thing for CPU compile (either skip compile entirely or fix the simulated tape to detach arrays and do donation properly)


awni:
I think this is safe to review now. 
- Tests pass
- I'm happy with how CPU is handled (it still compiles but detaches etc)
- Transforms over Compiled work and for the sake of keeping the diff manageable, I would prefer to do that it in a follow on if everyone is 👌 .

mlx/compile.cpp
```diff
@@ -327,6 +467,216 @@ void compile_simplify(
...
+    if (global_cache.find(arr.id()) != global_cache.end()) {
+      continue;
+    }
+
+    // Two pass recursion:
+    // First pass:
+    //  - Collect all the primitives which we can fuse with
+    //  - Keeps a cache of fusable primitives which may be added out of
+    //    DAG order. We have to determine if all of a fused primitive's
+    //    outptus are also in the fused section, and this may not be the
```
angeloskath, line 505: nit: typo outptus -> outputs

mlx/compile.cpp
```diff
@@ -327,6 +467,216 @@ void compile_simplify(
   }
 }
 
+// Extract sub-graphs of the graph that can be compiled
+// and replace them with a Compiled Primitive.
+void compile_fuse(
```
angeloskath, line 472: Very very nice!

angeloskath (approved):
This is perfect! Also very well tested, I'm super hyped!

mlx/compile.h
```diff
@@ -0,0 +1,28 @@
+// Copyright © 2023-2024 Apple Inc.
+
+#pragma once
+
+#include "mlx/array.h"
+
+namespace mlx::core {
+
+enum class CompileMode { disabled, no_simplify, no_fuse, enabled };
```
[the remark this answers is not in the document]
awni: BTW I didn't expose this in Python yet.. but we can (and just not advertise it) if it's helpful for debugging. Mostly added it so I could test simplify and fusion separately.

PR #579, opened 2024-01-29
Propagate nans in binary ops

### LINKED ISSUE #570: negative dimensions should not be allowed in 'mlx.core.full'.
**Describe the bug**
When the mlx.core.full parameter shape contains negative values, the program should crash instead of returning an empty array.

**To Reproduce**

```python
import mlx.core as mx

shape = (-1, 2)
vals = mx.array([-2,2])

full_array = mx.full(shape, vals)
print(full_array)
```
image

```python
import numpy as np

shape = (-1, 2)
vals = np.array([2,2])

np_full_array = np.full(shape, vals)
print(np_full_array)
```
image

**Expected behavior**
The program should crash and return an error message.

 - OS Version: MacOS 14.2.1
 - Version 0.10.0

### LINKED ISSUE #572: abort occurs when performing an mlx.core.inner operation on an empty array using the cpu
**Describe the bug**
abort occurs when performing an mlx.core.inner operation on an empty array using the cpu

**To Reproduce**
```python
import mlx.core as mx

a_mx = mx.array([], dtype=mx.float32)
b_mx = mx.array([], dtype=mx.float32)

result_gpu = mx.inner(a_mx, b_mx)
print("Inner product on GPU:", result_gpu)

mx.set_default_device(mx.cpu)
result_cpu = mx.inner(a_mx, b_mx)
print("Inner product on CPU:", result_cpu)

```
image

 - OS Version: MacOS 14.2.1
 - Version 0.10.0

### LINKED ISSUE #575: mlx.core.logaddexp got an wrong output on the gpu when dealing with nan
**Describe the bug**
mlx.core.logaddexp got an wrong output on the gpu when dealing with nan, the output should be nan.

**To Reproduce**
```python
import mlx.core as mx

a_mx = mx.array([float('nan'), 0.2, 0.3], dtype=mx.float32)
b_mx = mx.array([1, 2, 3], dtype=mx.float32)

logaddexp_gpu = mx.logaddexp(a_mx, b_mx)
print("logaddexp on GPU:", logaddexp_gpu)

mx.set_default_device(mx.cpu)
logaddexp_cpu = mx.logaddexp(a_mx, b_mx)
print("logaddexp on CPU:", logaddexp_cpu)
```
image

**Expected behavior**
array([nan, 2.15298, 3.06504], dtype=float32)

 - OS Version: MacOS 14.2.1
 - Version 0.0.10

awni:
### Propagate NaN in `logaddexp`, `maximum`, `minimum`

Closes #575 

Notice for the CPU maximum and minimum I removed `vDSP`, just using the default op. I was expecting this to have very little impact on timing since it's such a memory bandwidth bound op and the reads should be pretty well vectorized. To my surprise it's actually quite a bit faster to NOT use `vDSP`. Any ideas why that might be? 🤔 

After change
```
Timing maximum ... 4.97326 msec
```

Before change:
```
Timing maximum ... 6.62306 msec
```

### Handle empty matmuls on CPU:

Closes #572 

### Throw on negative indices in full

Closes #570

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -42,7 +42,7 @@ jobs:
       - run:
           name: Run Python tests
           command: |
-            python3 -m unittest discover python/tests
+            python3 -m unittest discover python/tests -v
       # TODO: Reenable when extension api becomes stable
       # - run:
       #     name: Build example extension
--- a/benchmarks/python/single_ops.py
+++ b/benchmarks/python/single_ops.py
@@ -44,6 +44,13 @@ def time_matmul():
     time_fn(mx.matmul, a, b)
 
 
+def time_maximum():
+    a = mx.random.uniform(shape=(32, 1024, 1024))
+    b = mx.random.uniform(shape=(32, 1024, 1024))
+    mx.eval(a, b)
+    time_fn(mx.maximum, a, b)
+
+
 def time_negative():
     a = mx.random.uniform(shape=(10000, 1000))
     mx.eval(a)
@@ -101,6 +108,7 @@ if __name__ == "__main__":
 
     time_add()
     time_matmul()
+    time_maximum()
     time_exp()
     time_negative()
     time_logsumexp()
--- a/mlx/backend/accelerate/matmul.cpp
+++ b/mlx/backend/accelerate/matmul.cpp
@@ -46,6 +46,11 @@ inline void matmul_cblas_general(
   size_t N = b.shape(-1);
   size_t K = a.shape(-1);
 
+  if (K == 0) {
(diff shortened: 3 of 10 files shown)


angeloskath:
I have no idea. vDSP should be faster as we are not using `simd_float16` in the default op so only whatever the compiler auto vectorizes which won't be much judging from prior experience.

For simple binary ops we could do a default vectorized cpu op that we can use. It will at least make sure to do vectorized reads even if it resorts to an unrolled inner loop for performing the operation.

awni:
I'm going to merge this for now. Given the above results, it's probably worth revisiting our use of some of the VDSP binary and unary ops. Certainly simplifies things if we don't use them and let alone the speed improvement.

PR #581, opened 2024-01-29
Update Compute Pipeline Creation API

jagrit06:
## Proposed changes

Add options to make metal functions and compute pipeline specializations with function constants and function linking 

## Checklist


angeloskath (approved):
I am not sure I see where everything will be used but I see the pieces needed by v1 of the compile primitive so I am hyped :-)

PR #584, opened 2024-01-30
Fix binary op dispatch

### LINKED ISSUE #583: Different result from Numpy in simple add operator
**Describe the bug**
I'm trying to add two differently shaped arrays, but the results are unexpected. To confirm the discrepancy in the output, I compared the results with NumPy.

**To Reproduce**

Include code snippet
```
import mlx.core as mx
import numpy as np

left_shape = [1, 1, 2, 1, 1, 2]
right_shape = [2, 1]

mlx_left = mx.ones(left_shape)
mlx_right = mx.ones(right_shape)

np_left = np.ones(left_shape)
np_right = np.ones(right_shape)

mlx_result = mlx_left + mlx_right
np_result = np_left + np_right

print("Shapes same:", list(mlx_result.shape) == list(np_result.shape))
print("Result diff:",(np_result - mlx_result).max())
```
Prints:
Shapes same: True
Result diff: 1.0

 - OS Version:  MacOS 14.2.1
 - Version 0.0.11

angeloskath:
When dispatching with many collapsible dims we were passing the wrong number of dims to the kernel.

This fixes and closes #583 .


awni (approved):
thanks for the fix

Issue #297, opened 2023-12-27
array.shape should return tuple not list for maximal compatibility with numpy 


dastrobu:
I like the statement: 
> MLX has a Python API that closely follows NumPy.

One thing I noticed is that mx.array.shape returns a list:

```
mx.ones([2, 3]).shape
[2, 3]
```

while numpy returns a tuple: 
```
np.ones([2, 3]).shape
(2, 3)
```

Given that tuples are immutable containers in python, returning a tuple makes more sense from my point of view. 

I am not sure whether this was a design decision. But given that mlx is sill in an early phase, it might be worth considering changing the shape to a tuple type for maximal consistency with numpy.

awni:
To be honest this wasn't a very intentional design decision. Pybind11 by default translates std::vector to Python list so that's why it is the way it is.  I believe it's easy to change, but a likely breaking API change for downstream code..

I see some arguments in favor of one or the other. But I think it would be good to have some votes in favor of the pros and cons.

- List is nice because it can be used almost everywhere a tuple can be (with the exception of maps) and it is easy to modify. A common idiom is to get the shape from an array, modify it, and use it to do another operation, like a reshape.

-Tuple is nice for perfect compatibility with NumPy. Beyond that what advantages does it ihave?

dastrobu:
Thanks for your comment. 

I think it does not really matter as you can convert lists and tuples very easily and performance shouldn't really matter when dealing with shapes. 

Having compatibility with numpy is the biggest advantage from my point of view, as you can easily migrate existing code from numpy arrays to mx arrays. 

I can live with lists, I just thought it might have been overlooked as numpy compatibility seemed a design goal, which I really like, as this makes learning MLX easier. 

Implementation wise this could be as easy as `py::tuple(py::cast(a.shape()))`: 

```
      // TODO, this makes a deep copy of the shape
      // implement alternatives to use reference
      // https://pybind11.readthedocs.io/en/stable/advanced/cast/stl.html
      .def_property_readonly(
          "shape",
          [](const array& a) { return py::tuple(py::cast(a.shape())); },
```

I am not sure if `py::cast` addresses the TODO currently in the code, since I am not too familiar with the details of pybind11. The term `cast` sounds like not-copying to me, though. Details see: https://pybind11.readthedocs.io/en/stable/advanced/pycpp/object.html#casting-back-and-forth
Maybe this helps deciding on the best type to return. 

[...]

PR #591, opened 2024-01-30
Make shape a tuple

awni:
## Proposed changes

Switch to tuple shape in Python.

Companion PR in mlx-examples: https://github.com/ml-explore/mlx-examples/pull/393

### DIFF
--- a/python/mlx/nn/losses.py
+++ b/python/mlx/nn/losses.py
@@ -70,8 +70,9 @@ def cross_entropy(
     targets_as_probs = targets.ndim == logits.ndim
 
     def _drop_dim(shape, axis):
+        shape = list(shape)
         shape.pop(axis)
-        return shape
+        return tuple(shape)
 
     # Check shapes in two cases: targets as class indices and targets as probabilities
     if (targets_as_probs and targets.shape != logits.shape) or (
--- a/docs/src/dev/extensions.rst
+++ b/docs/src/dev/extensions.rst
@@ -929,7 +929,7 @@ We see some modest improvements right away!
 
 This operation is now good to be used to build other operations, 
 in :class:`mlx.nn.Module` calls, and also as a part of graph 
-transformations such as :meth:`grad` and :meth:`simplify`!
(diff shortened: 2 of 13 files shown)


jagrit06:
@awni Do we need to go and change the docs for ops that require shape as an argument to be tuples ?
I don't think it's needed at the moment, but thought I should ask

awni:
> Do we need to go and change the docs for ops that require shape as an argument to be tuples

Luckily I don't think so, for all of those ops the shape can be either a `tuple` or a `list`. That was already the case before this change. We may want to make it more clear in the docs but I'll leave that as a future update:

```
>>> mx.zeros([2, 2])
array([[0, 0],
       [0, 0]], dtype=float32)
>>> mx.zeros((2, 2))
array([[0, 0],
       [0, 0]], dtype=float32)
```

python/mlx/nn/losses.py
```diff
@@ -70,8 +70,9 @@ def cross_entropy(
     targets_as_probs = targets.ndim == logits.ndim
 
     def _drop_dim(shape, axis):
+        shape = list(shape)
```
angeloskath, line 73: Feel free to ignore but `return shape[:axis] + shape[axis+1:]`? Avoids the list/tuple conversion.

angeloskath (approved):
Perfect!

jagrit06 (approved):
LGTM

python/mlx/nn/losses.py
awni: Ah it's unfortunately a slightly more involved to get negative axes to work.

angeloskath: Makes sense :-)

awni:
Ok I'm merging this!! Something is almost certainly going to break so just keep an eye out! I checked most things in core and examples though so that should be ok.

dastrobu:
@awni this seems to close #297. 
It would be intriguing to learn the decisive factor that prompted the implementation of proposal #297.

PR #167, opened 2023-12-14
Transformer fix

Jyun1998:
## Proposed changes

Changes
- Changed order of layernorm and ffn according to tensorflow and pytorch implementations. (norm_first=False is default in pytroch as well as transformer)

- Added Encoder, Decoder with Dropout

Parts want to discuss

- Do you want to use Transformer with Dropout as main class?
- Do you like to add ffn as function or classes?
- Do you like to add norm_first parameter so that you could choose to do layer_norm first or later?
- Tensorflow do not do dropout with ffn while pytorch do. Which one should I refer to?

- I guess you did not add dropout since you would like to benchmark transformer without dropout for all jax, torch, tensorflow, mlx. I also had trouble adding dropout for jax. 

Testing
- All Encoder, Decoder (w and w/o dropout) are tested with mlx-exmaples/transformer_lm

## Checklist

Put an `x` in the boxes that apply.

- [x] I have read the [CONTRIBUTING](https://github.com/ml-explore/mlx/blob/main/CONTRIBUTING.md) document
- [x] I have run `pre-commit run --all-files` to format my code / installed pre-commit prior to committing changes
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have updated the necessary documentation (if needed)

PR #599, opened 2024-01-31
Change the transformer to norm_first by default

angeloskath:
This is a backwards incompatible change as it changes the default to `norm_first` however I think this should have always been the default and hopefully if not too many people relied on it we caught it early.

If not we can cancel this PR.

### DIFF
--- a/python/mlx/nn/layers/transformer.py
+++ b/python/mlx/nn/layers/transformer.py
@@ -116,7 +116,7 @@ class TransformerEncoderLayer(Module):
         mlp_dims: Optional[int] = None,
         dropout: float = 0.0,
         activation: Callable[[Any], Any] = relu,
-        norm_first: bool = False,
+        norm_first: bool = True,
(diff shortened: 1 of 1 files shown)


awni (approved):
Yea at some point I reviewed a PR which added a bunch of stuff to Transformer including this. I think I said we should just leave it to post norm since it's way more common now (?). 

But this gets us compatibility with PyTorch so I'm also on board switching. I don't think it should break too much. In most of our examples we reimplement the Transformer anyway 😅

angeloskath:
Just to be clear. This breaks compatibility with PT and moves to the more common option. As you said the only thing that is used in places is the attention implementation and I think this is more of a guide/example at this point.

awni:
Sorry I confused myself 😅 , I actually did the exact opposite of what I thought I did in this PR #167 , where it was switched to post norm by default.

I am fine with reverting that.

PR #601, opened 2024-02-01
Reduce vmap + some fixes

### LINKED ISSUE #442: [Feature] Support vmap for reductions
Me and @tedwards2412 are trying to use vmap in conjunction with the summation operation mx.sum, but we're getting some results that we don't understand. We'd expect both the mx.sum() and sum() operations below to produce the same result, but for some reason mx.sum() only outputs one number. 

```python
import mlx.core as mx
x_array = mx.ones((5,10))

def test_func(x):
    return mx.sum(x), sum(x)

print(mx.vmap(test_func)(x_array), "\n")
print([test_func(x) for x in x_array])
```

The output of this is 

```python
(array(1, dtype=float32), array([10, 10, 10, 10, 10], dtype=float32)) 
[(array(10, dtype=float32), array(10, dtype=float32)), (array(10, dtype=float32), array(10, dtype=float32)), (array(10, dtype=float32), array(10, dtype=float32)), (array(10, dtype=float32), array(10, dtype=float32)), (array(10, dtype=float32), array(10, dtype=float32))]
```

while we would expect the vmap (first line of the output) to give

```python
(array([10, 10, 10, 10, 10], dtype=float32)), array([10, 10, 10, 10, 10], dtype=float32)) 
```

This is what we get with mlx v0.0.7. After upgrading to v0.0.9, vmap stopped working altogether:

```python
    print(mx.vmap(test_func)(x_array), "\n")
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^
RuntimeError: Reduce::vmap not yet implemented.
```

Any idea what's going on here?

### LINKED ISSUE #600: Unexpected result from vmap broadcasting
**Describe the bug**
I am not sure whether this is a bug or not. I guess the expected result of the operation and its Vmapped operation must be the same.  In my case, the shape of the Vmapped operation and the actual operation is not the same.

**To Reproduce**

```
import mlx.core as mx

arr1_shape = [10,1]
arr2_shape = [1,1,1,5]
arr1 = mx.ones(arr1_shape)
arr2 = mx.ones(arr2_shape)


def my_adder_mlx(arr1, arr2):
    return arr1 + arr2

my_adder_mlx_vmap = mx.vmap(my_adder_mlx)
vmap_result = my_adder_mlx_vmap(arr1, arr2)
result = arr1+arr2

print("Vmap result shape:", vmap_result.shape)
print("Op result shape:",result.shape)
```
prints:

Vmap result shape: [10, 1, 1, 5]
Op result shape: [1, 1, 10, 5]

**Expected behavior**
Vmap result shape should be equal to op result shape

 - OS Version: MacOS 14.2.1
 - Version: 0.0.11

awni:
## Proposed changes

- Add vmap for reductions - closes #442
- Change reductions to explicitly squeeze unkept dims. This has two benefits: 1. Catch more merges for simplify, 2. Simplifies vmap code a lot
- Throw if inputs do not have same shape along vmap axes - closes #600

### DIFF
--- a/mlx/ops.cpp
+++ b/mlx/ops.cpp
@@ -17,8 +17,7 @@ namespace {
 
 std::pair<std::vector<int>, std::vector<int>> compute_reduce_shape(
     const std::vector<int>& axes,
-    const std::vector<int>& shape,
-    bool keepdims) {
+    const std::vector<int>& shape) {
   std::set<int> axes_set;
   auto ndim = shape.size();
   for (auto ax : axes) {
@@ -38,7 +37,7 @@ std::pair<std::vector<int>, std::vector<int>> compute_reduce_shape(
   for (int i = 0; i < ndim; ++i) {
     if (axes_set.count(i) == 0) {
       out_shape.push_back(shape[i]);
-    } else if (keepdims) {
+    } else {
       out_shape.push_back(1);
     }
   }
@@ -1217,13 +1216,16 @@ array all(
   if (axes.empty()) {
     return astype(a, bool_, s);
   }
-  auto [out_shape, sorted_axes] =
-      compute_reduce_shape(axes, a.shape(), keepdims);
-  return array(
+  auto [out_shape, sorted_axes] = compute_reduce_shape(axes, a.shape());
+  auto out = array(
       out_shape,
       bool_,
       std::make_unique<Reduce>(to_stream(s), Reduce::And, sorted_axes),
       {a});
+  if (!keepdims) {
+    out = squeeze(out, sorted_axes, s);
+  }
+  return out;
 }
 
 array all(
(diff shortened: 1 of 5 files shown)


mlx/ops.cpp
```diff
@@ -1217,13 +1216,16 @@ array all(
...
-  auto [out_shape, sorted_axes] =
-      compute_reduce_shape(axes, a.shape(), keepdims);
-  return array(
+  auto [out_shape, sorted_axes] = compute_reduce_shape(axes, a.shape());
+  auto out = array(
       out_shape,
       bool_,
       std::make_unique<Reduce>(to_stream(s), Reduce::And, sorted_axes),
       {a});
+  if (!keepdims) {
```
angeloskath, line 1225: Love this change!

angeloskath (approved):
This looks great! Love the extra tests, the vmap fix and the change for keepdims!

Issue #573, opened 2024-01-28
Wrong output was obtained when the num parameter of 'mlx.core.linspace' was set to 1.


Redmept1on:
**Describe the bug**
An error output nan was obtained when the num parameter of mlx.core.linspace was set to 1.

**To Reproduce**

```python
import mlx.core as mx
import numpy as np

start = 1
stop = 10
num = 1

linspace_mlx = mx.linspace(start, stop, num, dtype=mx.float32)
print("Linspace using mlx:", linspace_mlx)

linspace_np = np.linspace(start, stop, num, dtype=np.float32)
print("Linspace using NumPy:", linspace_np)
```
![image](https://github.com/ml-explore/mlx/assets/84373897/aad92347-fc52-497f-9a1f-b14fd6ce6e67)

**Expected behavior**
maybe same as numpy

**Desktop (please complete the following information):**
 - OS Version: MacOS 14.2.1
 - Version 0.10.0

AvikantSrivastava:
@Redmept1on @awni can I work on this?

awni:
That would be awesome, thank you @AvikantSrivastava

AvikantSrivastava:
@awni can you take at look at the PR #602

AvikantSrivastava:
@Redmept1on I guess we can close this issue

PR #602, opened 2024-02-01
fix: handle linspace function when num is 1

AvikantSrivastava:
## Proposed changes

The linspace function had a bug that return `nan` whenever the value of `num` is 1.
Issue: https://github.com/ml-explore/mlx/issues/573

## Checklist


AvikantSrivastava:
@awni @Redmept1on 

here is the output of the following code
```python
import mlx.core as mx
import numpy as np

start = 1
stop = 10
num = 1

linspace_mlx = mx.linspace(start, stop, num, dtype=mx.float32)
print("Linspace using mlx:", linspace_mlx)

linspace_np = np.linspace(start, stop, num, dtype=np.float32)
print("Linspace using NumPy:", linspace_np)
```

```
Linspace using mlx: array(1, dtype=float32)
Linspace using NumPy: [1.]
```

AvikantSrivastava:
@awni
I have also added new test case for this
is it okay? can you review the PR?

awni:
> I have also added new test case for this

That's great, thanks! I will review it tonight.

python/tests/test_ops.py
```diff
@@ -1659,6 +1659,11 @@ def test_linspace(self):
         expected = mx.array(np.linspace(0, 1, 10))
         self.assertEqualArray(d, expected)
 
+        # Test num equal to 1
+        d = mx.linspace(1, 10, 1)
+        expected = mx.array(np.linspace(0, 1, 10))
+        self.assertEqualArray(d, expected)
```
awni, line 1665: Does this test pass?

awni: Because it shouldn't 😄

AvikantSrivastava: oops my bad, I forgot to make changes here

AvikantSrivastava: done!

AvikantSrivastava:
requesting for a re review @awni

python/tests/test_ops.py
```diff
@@ -1659,6 +1659,12 @@ def test_linspace(self):
         expected = mx.array(np.linspace(0, 1, 10))
         self.assertEqualArray(d, expected)
 
+        # Test num equal to 1
+        d = mx.linspace(1, 10, 1)
+        expected = mx.array(np.linspace(1, 10, 1))
+        breakpoint()
```
awni, line 1665: `breakpoint`?

AvikantSrivastava:
@awni requesting for re-review

awni (approved):
Looks great, thanks!

PR #603, opened 2024-02-01
Fix for AdaDelta

davidkoski:
## Proposed changes

- state was being read from parameter "s"
- but being stored in parameter "u"

## Checklist

I don't have a good way to test this -- fix by inspection.


awni (approved):
Good catch!!

PR #612, opened 2024-02-02
Fix eval in trace bugs

### LINKED ISSUE #585: Checks are only performed in operator overloads but not their underlying equivalent
Operator overload `operator&&` checks if both arrays are Boolean valued but `logical_and` does not. Similar behavior is found with `operator||` and `logical_or`. Shouldn't both `operator&&` and `logical_and` perform the check?

Version: 0.0.11

### LINKED ISSUE #610: Wrong grad obtained from mlx.core.vjp
**Describe the bug**
The wrong gradient is obtained from mlx.core.vjp. Removing the .tolist() method resolves the issue, and Jax behaves as expected. This bug does not occur in version 0.0.11. I think something wrong in broadcast.
 
**To Reproduce**
In version 0.1.0
```python
import mlx.core as mx
arr = mx.ones((3,3))
out_grad = mx.ones((3,3,3))
def func(arr):
    arr += mx.ones((3,3,3))
    cond = arr < 1
    cond.tolist()
    return arr**2

_, grad = mx.vjp(func,[arr], [out_grad])
print("Arr grad:", grad)
```
Prints:
Arr grad: [array([[0, 0, 0],
       [0, 0, 0],
       [0, 0, 0]], dtype=float32)]

In version 0.0.11 same code prints:
Arr grad: [array([[12, 12, 12],
       [12, 12, 12],
       [12, 12, 12]], dtype=float32)]

Jax implementation
```python
import jax
arr = jax.numpy.ones((3,3))
out_grad = jax.numpy.ones((3,3,3))
def func(arr):
    arr += jax.numpy.ones((3,3,3))
    cond = arr < 1
    cond.tolist()
    return arr**2

_, vjp_fn = jax.vjp(func, arr)
grad =vjp_fn(out_grad)
print("Arr grad:", grad)
```
Prints:
Arr grad: (Array([[12., 12., 12.],
       [12., 12., 12.],
       [12., 12., 12.]], dtype=float32),)

**Expected behavior**
Same result with Jax

 - OS Version:  MacOS 14.2.1
 - Version: 0.1.0

awni:
## Proposed changes

Fix two bugs when doing an eval in a trace (both introduced since our 0.0.11 release by me 😢 )

Closes #610 

Also just close #585

### DIFF
--- a/mlx/array.cpp
+++ b/mlx/array.cpp
@@ -180,7 +180,7 @@ array::ArrayDesc::ArrayDesc(
       primitive(std::move(primitive)),
       inputs(inputs) {
   std::tie(size, strides) = cum_prod(this->shape);
-  for (auto& in : inputs) {
+  for (auto& in : this->inputs) {
     is_tracer |= in.is_tracer();
     depth = std::max(in.graph_depth(), depth);
   }
@@ -197,7 +197,7 @@ array::ArrayDesc::ArrayDesc(
       primitive(std::move(primitive)),
       inputs(std::move(inputs)) {
   std::tie(size, strides) = cum_prod(this->shape);
-  for (auto& in : inputs) {
+  for (auto& in : this->inputs) {
     is_tracer |= in.is_tracer();
     depth = std::max(in.graph_depth(), depth);
   }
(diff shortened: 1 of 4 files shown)


awni:
This might be good to put in a patch release...

angeloskath (approved):
👍

PR #613, opened 2024-02-02
Minor changes to extensions

awni:
## Proposed changes

- Require cmake 27 (list transform in cmake requires it)
- Remove includes for json/gguf and make private. Not sure why we were exposing those


angeloskath (approved):
👍

PR #614, opened 2024-02-03
Kernel generation

angeloskath:
Generated fused kernels on the fly for the Compiled primitive. This is currently on top of #571 so the diff is a bit misleading.

Main topics of discussion:

- Generation of the preamble is slightly icky. I am using CMake to run the preprocessor to create a file that returns the output of the preprocessor as a string.
- I am generating unique names for the kernels and 1 kernel per output shape. It may be smart to have a way to clean up the loaded libraries.


angeloskath:
The bulk of the PR is in `mlx/backend/metal/compiled.cpp` . It is not the prettiest code generation but at least the kernels are reusable and general to the number of dims. Compiling and benchmarking the activation functions we get the following table

```
Activation function | Uncompiled | Compiled | Max Abs Difference
--------------------+------------+----------+--------------------
celu                |   0.6604   |  0.0847  | 2.9802322387695312e-08
elu                 |   0.8528   |  0.0844  | 0.0
gelu                |   0.4807   |  0.0776  | 4.76837158203125e-07
gelu_approx         |   0.7040   |  0.0772  | 7.152557373046875e-07
gelu_fast_approx    |   0.3058   |  0.0795  | 7.152557373046875e-07
glu                 |   0.1582   |  0.0592  | 5.960464477539062e-07
hardswish           |   0.4822   |  0.0773  | 4.76837158203125e-07
leaky_relu          |   0.2065   |  0.0802  | 0.0
log_sigmoid         |   0.2615   |  0.0775  | 4.76837158203125e-07
log_softmax         |   0.4256   |  0.3614  | 0.0
mish                |   0.3106   |  0.1115  | 5.960464477539062e-07
relu                |   0.0867   |  0.0859  | 0.0
relu6               |   0.1758   |  0.0775  | 0.0
selu                |   0.9433   |  0.0846  | 0.0
silu                |   0.2041   |  0.0776  | 7.152557373046875e-07
softmax             |   0.0768   |  0.0855  | 0.0
softplus            |   0.0878   |  0.0775  | 4.76837158203125e-07
softshrink          |   0.9460   |  0.0775  | 0.0
softsign            |   0.3055   |  0.0776  | 5.960464477539063e-08
step                |   0.5054   |  0.0878  | 0
tanh                |   0.0818   |  0.0822  | 0.0
```

The script I ran is the following

```python
import time

import mlx.core as mx
import mlx.nn as nn


def bench(f, x):
    xs = []
    for i in range(100):
        xs.append(f(x))
    mx.eval(xs)

    start = time.time()
    xs = []
    for i in range(100):
        xs.append(f(x))
    mx.eval(xs)
    end = time.time()

    return end-start


x = mx.random.normal(shape=(100, 1000, 100)).swapaxes(0, 1)
mx.eval(x)

print("Activation function | Uncompiled | Compiled | Max Abs Difference")
print("--------------------+------------+----------+--------------------")
for name in dir(nn.activations):
    if name.lower() != name:
        continue

    if name.startswith("_"):
        continue

    if name in ["math", "mx", "prelu", "sigmoid"]:
        continue

    f = getattr(nn, name)
    cf = mx.compile(f)

    before = bench(f, x)
    after = bench(cf, x)
    print(f"{name:19} |   {before:.4f}   |  {after:.4f}  |", (f(x) - cf(x)).abs().max().item())
```

mlx/backend/metal/make_compiled_preamble.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
+
+OUTPUT_FILE=$1
+CC=$2
+SRCDIR=$3
+
+CONTENT=$($CC -I $SRCDIR -E $SRCDIR/mlx/backend/metal/kernels/compiled_preamble.h 2>/dev/null)
+
+cat << EOF > $OUTPUT_FILE
+// Copyright © 2023-24 Apple Inc.
```
awni, line 10: Lol, probably don't need this in the preamble, though technically we should have one at the top of this file.

awni:
🤤 those results are wild 🚀 

> The bulk of the PR is in mlx/backend/metal/compiled.cpp 

I reviewed everything except `mlx/backend/metal/compiled.cpp` 😆 , it all looks great. I liked the way you handled the preamble.

I will finish that today, and play around with it a bit and then we should land this!

awni:
> softmax             |   0.0768   |  0.0855  | 0.0

Isn't softmax just using the custom primitive? Any idea why it's slower?

awni:
Also anther random question:

```
gelu                |   0.4807   |  0.0776  | 4.76837158203125e-07
gelu_approx         |   0.7040   |  0.0772  | 7.152557373046875e-07
```

What is the purpose of `gelu_approx` if it's slower than `gelu`?

mlx/backend/metal/compiled.cpp
```diff
@@ -1,44 +1,473 @@
...
+void print_int_constant(std::ostream& os, const array& x) {
+  os << x.item<T>();
+}
+
+void print_constant(std::ostream& os, const array& x) {
+  switch (x.dtype()) {
+    case float16:
+      return print_float_constant<float16_t>(os, x);
+    case float32:
+      return print_float_constant<float>(os, x);
```
awni, line 71: Is it intentional the `bfloat` is missing from this list?

awni (approved):
This is incredible, really excited to try it! I'm pretty comfortable merging it since pretty self-contained.  If there are any changes or kinks we can always iron them out as we find them.

🚀 🔥

angeloskath:
> Also anther random question:
> 
> ```
> gelu                |   0.4807   |  0.0776  | 4.76837158203125e-07
> gelu_approx         |   0.7040   |  0.0772  | 7.152557373046875e-07
> ```
> 
> What is the purpose of `gelu_approx` if it's slower than `gelu`?

Literally no purpose. It is an artifact of porting things from GPUs without benchmarking them. Don't know whether we should do something about it. It might be less compute intensive when fused but obviously read/write bandwidth still dominates.

mlx/backend/metal/compiled.cpp
angeloskath: Nope. Interestingly no bfloat constants came up in my testing. I had forgotten to put it as a type and that crashed when compiling mistral inference but apparently I didn't use any constants :-).

mlx/backend/metal/compiled.cpp
```diff
@@ -1,44 +1,475 @@
...
+    }
+  }
+}
+
+template <typename T>
+void print_float_constant(std::ostream& os, const array& x) {
+  auto old_precision = os.precision();
+  os << std::setprecision(std::numeric_limits<float>::digits10 + 1)
+     << x.item<T>() << std::setprecision(old_precision);
+}
```
jagrit06, line 59: For metal `half` types, it's best to put `h` after literals (`2.0` vs `2.0h`) otherwise everything is promoted to `float`. That said I don't think that's a problem for the current paradigm. 
As far as I understand it, we aren't inlining any constants, in that it is always going to be 
`half a = ...; half b = 2.0; half c = Add()(a, b);` instead of `half a = ...; half c = Add()(a, 2.0);` so not much to worry about unless there is a place where we are inlining constants.

mlx/backend/metal/compiled.cpp
```diff
@@ -1,44 +1,475 @@
...
+    os << "    constant const int& ndim [[buffer(" << cnt++ << ")]],"
+       << std::endl;
+  }
+
+  // The thread index in the whole grid
+  os << "    uint3 pos [[thread_position_in_grid]]," << std::endl
+     << "    uint3 grid [[threads_per_grid]]) {" << std::endl
+     << "  uint index = pos.x + grid.x * (pos.y + grid.y * pos.z);"
+     << std::endl;
+
```
jagrit06, line 207: So if `cnt` ends up `>= 30` we will get the metal library complaining anyway - I'll leave it up to your discretion if you want to a clearer error message here for now

jagrit06 (approved):
Love it! 🚀

mlx/backend/metal/compiled.cpp
angeloskath: Yeah exactly, I am also always printing them in float precision and they will automatically cast to whatever. Good to know for the future though!

mlx/backend/metal/compiled.cpp
angeloskath: I think that is a good point. The error message from the compiler is pretty clear but people may prefer a suggestion like "Too many inputs fused in the Compile primitive". I 'll try to make a good error message that includes the kernel :-) .

awni: Maybe we should try to catch this upstream. I do not think it would be so difficult to limit the maximum number of inputs and outputs to a Compiled primitive

awni: It must be quite rare I think because I haven't run into it yet.. but would be nice for large / wide / fusable graphs to work rather than crash ..

angeloskath: Yeah I think we should try to catch it. But in this PR improving the error message is pretty much all we can/should do. If someone provides us a graph that wide, we simply cannot build a kernel for it.

awni: Right just checking if it's something I should add to the primitive fusion function.

PR #619, opened 2024-02-03
fixes for building and running on iOS

davidkoski:
## Proposed changes

Fixes for building and running on iOS:

- veclib imports are not available -- use Accelerate in those cases
- getting the default device requires a different API

## Checklist

Verified manually.

### DIFF
--- a/mlx/backend/common/conv.cpp
+++ b/mlx/backend/common/conv.cpp
@@ -3,7 +3,11 @@
 #include <cassert>
 
 #ifdef ACCELERATE_NEW_LAPACK
+#if __has_include(<vecLib/vecLib.h>)
 #include <vecLib/cblas_new.h>
+#else
+#include <Accelerate/Accelerate.h>
+#endif
 #else
 #include <cblas.h>
 #endif
--- a/mlx/backend/common/conv.cpp
+++ b/mlx/backend/common/conv.cpp
@@ -3,11 +3,7 @@
 #include <cassert>
 
 #ifdef ACCELERATE_NEW_LAPACK
-#if __has_include(<vecLib/vecLib.h>)
-#include <vecLib/cblas_new.h>
-#else
 #include <Accelerate/Accelerate.h>
-#endif
 #else
 #include <cblas.h>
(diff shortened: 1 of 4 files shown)


mlx/backend/common/conv.cpp
```diff
@@ -3,8 +3,12 @@
 #include <cassert>
 
 #ifdef ACCELERATE_NEW_LAPACK
+#if __has_include(<vecLib/vecLib.h>)
 #include <vecLib/cblas_new.h>
 #else
+#include <Accelerate/Accelerate.h>
+#endif
+#else
```
davidkoski, line 11: on iOS you get this via the Accelerate umbrella header

mlx/backend/metal/device.cpp
```diff
@@ -26,7 +26,8 @@ static constexpr const char* default_mtllib_path = METAL_PATH;
 
 auto load_device() {
   auto devices = MTL::CopyAllDevices();
-  auto device = static_cast<MTL::Device*>(devices->object(0));
+  auto device = static_cast<MTL::Device*>(devices->object(0))
+      ?: MTL::CreateSystemDefaultDevice();
```
davidkoski, line 30: The copyAllDevices is not available on iOS so use a fallback:

- https://developer.apple.com/documentation/metal/1433367-mtlcopyalldevices

mlx/backend/common/conv.cpp
awni: Does `Accelerate.h` provide all the headers even on MacOS? If so, we could just use that instead of `vecLib`...?

davidkoski: It looks like it does actually:

```
#ifndef __VECLIB__
#include "../Frameworks/vecLib.framework/Headers/vecLib.h"
#endif
```

I will test that out.

davidkoski: Tested it out and it works on both iOS and macOS -- updated the PR.

awni (approved):
Thank you!!

PR #622, opened 2024-02-04
expose numpy compatible strides attribute in python extension.

dastrobu:
## Proposed changes

expose numpy compatible strides attribute in python extension.

See https://numpy.org/doc/stable/reference/generated/numpy.ndarray.strides.html

## Checklist

### DIFF
--- a/docs/src/python/array.rst
+++ b/docs/src/python/array.rst
@@ -16,6 +16,7 @@ Array
     array.ndim
     array.shape
     array.size
+    array.strides
     Dtype
     array.abs
     array.all
--- a/python/src/array.cpp
+++ b/python/src/array.cpp
@@ -450,13 +450,22 @@ std::optional<std::string> buffer_format(const array& a) {
   }
 }
 
-std::vector<size_t> buffer_strides(const array& a) {
-  std::vector<size_t> py_strides;
-  py_strides.reserve(a.strides().size());
+/** The strides of the array in bytes */
+std::vector<size_t> byte_strides(const array& a) {
+  std::vector<size_t> byte_strides;
+  byte_strides.reserve(a.strides().size());
   for (const size_t stride : a.strides()) {
-    py_strides.push_back(stride * a.itemsize());
+    byte_strides.push_back(stride * a.itemsize());
   }
-  return py_strides;
+  return byte_strides;
+}
+
+array ensure_evaled(array& a) {
+  if (!a.is_evaled()) {
+    py::gil_scoped_release nogil;
+    a.eval();
+  }
+  return a;
 }
 
 ///////////////////////////////////////////////////////////////////////////////
@@ -649,11 +658,7 @@ void init_array(py::module_& m) {
 
   array_class
       .def_buffer([](array& a) {
-        // Eval if not already evaled
-        if (!a.is_evaled()) {
-          py::gil_scoped_release nogil;
-          a.eval();
-        }
+        ensure_evaled(a);
         return pybind11::buffer_info(
             a.data<void>(),
             a.itemsize(),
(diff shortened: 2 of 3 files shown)


awni:
I was just chatting with @angeloskath the other day about exposing a `strides` for use with something like `as_strided` as well.

Couple thoughts on this:

1. Eval. I do not think we should default eval if `strides` is called, that is a recipe for accidentally evaluating the graph which can be very bad in the wrong places. IMO a better option is to return the strides as if the array was row-contiguous if the array is unevaluated (or maybe throw?)

2. Bytes vs itemsize units: there may be an unfortunate difference between how we do strides and NumPy does strides. I believe our `as_strided` expects strides in `itemsize` untis. It seems like a bad idea to return bytes units if we expect `itemsize` units in there. Another option is to change `as_strided` to take bytes strides (at least in Python), and return bytes strides from `array.strides`. I might prefer that as it is more NumPy compatible.

angeloskath:
My thoughts on this is that it will be very easy to shoot oneself in the foot.

As strided is generally an advanced op and even in numpy it comes with warnings about misuse. What the op expects is the row contiguous strides so returning anything else is not useful.

Accessing the strides for other ops should be judged on a case by case basis I believe. In general we should not expect arrays to be materialized so they should not have properties like strides.

@dastrobu If you have a good use case for accessing the strides then we can discuss based on that.

dastrobu:
Thanks for the swift feedback!

> Eval. I do not think we should default eval if strides is called, that is a recipe for accidentally evaluating the
> graph which can be very bad in the wrong places. IMO a better option is to return the strides as if the array was
> row-contiguous if the array is unevaluated (or maybe throw?)

Returning strides that differ from what they will be after eval might lead to confusion. Throwing an error could be a
viable option. I would expect `a.strides == memoryview(a).strides` to always hold true.

> Bytes vs itemsize units: there may be an unfortunate difference between how we do strides and NumPy does strides. I
> believe our as_strided expects strides in itemsize untis.

I agree that having two conventions for strides can be confusing. Since in Python, people are accustomed to the bytes
convention (as seen in NumPy and memoryview), it would be beneficial to adhere to that convention. Considering
that [NumPy](https://numpy.org/doc/stable/reference/generated/numpy.lib.stride_tricks.as_strided.html) expects strides
in bytes, I strongly advocate for a NumPy-compatible implementation. If the signature matches, MLX should implement
functions in a NumPy-compatible way. If there are variations, a distinct function name could be chosen, to make users aware of the difference to numpy.

> In general we should not expect arrays to be materialized so they should not have properties like strides.

It's an interesting point, and I agree that it might be better not to expose strides at all than to return potentially
confusing information. It's strange that I can access the shape of an unevaluated array with `a.T.shape` but not the
strides. I would expect that `shape` and `strides` always go hand in hand. Understanding why `a.T.shape` works
without eval while `a.T.strides` doesn't would be insightful.

As stated above, `a.strides` would be a shortcut to `memoryview(a).strides` for me.

> @dastrobu If you have a good use case for accessing the strides then we can discuss based on that.

When looking at performance, I sometimes check the strides of arrays to verify if it is C or F contiguous for example.
Another use case is interoperability with other frameworks (`memoryview` again).
I did not have any use cases where I needed to use this during an actual computation inside a model.

I wasn't aware that there are so many aspects to this. I just wanted to add another missing numpy property to help MLX
getting more numpy compatible. If there are good reasons for not exposing strides, I would continue to
use `memoryview(a).strides`. My guess is, though, that not everyone who want to check on strides will immediately think
of this indirect approach.

angeloskath:
I think we can close this. @dastrobu let me know if you want to continue the discussion.

PR #629, opened 2024-02-05
Compile with capture

awni:
## Proposed changes

- Allow caller to specify captured inputs and outputs to `mx.compile`.
- Convert `OptimizerState` to a simple `dict`

The primary reason for this is that the syntax at the NN level is very nice. You can do the following to compile through a full optimization step with a `nn.Module`:

```python
state = [model, optimizer.state]
def step(x, y):
    loss, grads = nn.value_and_grad(model, loss_fn)(model, x, y)
    return loss

for x, y in data:
    loss = step(x, y)
    mx.eval(model, optimizer.state)
```

Note you can even add things to the `state` as you go (so we can keep lazy initialization of the optimizer). It will just trigger a new compile if you change the # inputs.

### Pure version

For comparison here is what a pure version would look like (not too bad IMO) but would need some changes on the optimizer side to get it to work (initialization and state update):

```python
opt = optim.SGD(learning_rate=learning_rate)
params = model.paramaters()

@mx.compile
def step(params, opt_state, x, y):
    def loss_fn(params, x, y):
        model.update(params)
        return mx.mean(nn.cross_entropy_loss(model(X), y))

    loss, grads = mx.value_and_grad(loss_fn)(params, x, y)

    # NB opt state has to be set
    opt.set_state(opt_state)
    params = opt.apply_gradients(grads, params)
    # NB opt state has to be returnedx
    return loss, params, opt.state

# NB opt has to be initialized (either explicitly or by passing params on
# construction)
opt.init(params)
params = model.parameters()

for x, y in data:
    loss, params, opt_state = step(params, opt_state, x, y)
    mx.eval(loss, params, opt_state)
```

### Notes

The way this works:
- The `inputs` are edited in place to be tracers on the trace pass to compile. The original inputs are then put back.
- The `outputs` are updated in place. 

This in place editing is both the magic and the risk. But since it's optional and really nice on the API, I'm tempted to keep it, but would be great to discuss the trade-offs.


python/mlx/optimizers.py
```diff
@@ -65,15 +62,38 @@ def apply_gradients(self, gradients: dict, model: dict):
...
+        """Set optimizer's state dictionary."""
+        self._state = state
+
+    @property
+    def learning_rate(self):
+        return self.state["learning_rate"]
+
+    @learning_rate.setter
+    def learning_rate(self, learning_rate: mx.array):
+        self.state["learning_rate"] = mx.array(learning_rate)
```
angeloskath, line 95: ❤️

python/mlx/optimizers.py
```diff
@@ -52,6 +30,25 @@ def update(self, model: "mlx.nn.Module", gradients: dict):
         """
         model.update(self.apply_gradients(gradients, model))
 
+    def init(self, model: dict):
+        """Initialize the optimizer's state
+
+        Args:
+            model (dict): A Python tree of parameters
```
angeloskath, line 37: It is unclear if this should be parameters or gradients especially given line 66.

python/mlx/optimizers.py
```diff
@@ -113,34 +133,33 @@ def __init__(
...
+    def apply_single(self, gradient: mx.array, parameter: mx.array, state: dict):
         """Performs the SGD parameter update and stores :math:`v` in the
         optimizer state."""
 
         if self.weight_decay != 0:
             gradient += self.weight_decay * parameter
 
         if self.momentum <= 0:
-            return parameter - self.learning_rate * gradient
+            return parameter - self.learning_rate.astype(gradient.dtype) * gradient
```
angeloskath, line 148: ❤️

awni:
I think this is ready for review. There are quite a few changes in here, but they are all related so... unless there is opposition I will keep it as is:

- Capturing inputs and outputs in compile (as above)
- Expose `model.state` and `optimizer.state` for capturing
- A purification of optimizers -> state is initialized in a separate function, though it can also be lazy as before.
- Learning rate is by default part of the optimizer state to allow for scheduling in compiled graphs
- Eval ignores non-arrays instead of craching (so we can `eval(model.state)`
- Expose `mx.random.state` to capture random state for compiled graphs

python/src/random.cpp
```diff
@@ -309,4 +358,7 @@ void init_random(py::module_& parent_module) {
         Returns:
             array: The ``shape``-sized output array with type ``uint32``.
       )pbdoc");
+  // Register static Python object cleanup before the interpreter exits
+  auto atexit = py::module_::import("atexit");
+  atexit.attr("register")(py::cpp_function([]() { default_key().release(); }));
```
angeloskath, line 363: release and dec ref perhaps?

awni:
Updated and tested on 7 examples using code gen + compile + capture in https://github.com/ml-explore/mlx-examples/pull/420

python/src/transforms.cpp
```diff
@@ -495,33 +553,77 @@ std::unordered_map<size_t, py::object>& tree_cache() {
...
 
-      tree_cache().insert({this->fun_id, py_outputs});
+      tree_cache().insert({fun_id, py_outputs});
+
+      num_outputs = outputs.size();
+      if (!py::isinstance<py::none>(captured_outputs)) {
+        auto flat_out_captures = tree_flatten(captured_outputs, false);
+        outputs.insert(
+            outputs.end(),
+            std::make_move_iterator(flat_out_captures.begin()),
```
angeloskath, line 599: ❤️

python/src/transforms.cpp
```diff
@@ -615,8 +719,8 @@ void init_transforms(py::module_& m) {
         Args:
             *args (arrays or trees of arrays): Each argument can be a single array
               or a tree of arrays. If a tree is given the nodes can be a Python
-              :class:`list`, :class:`tuple` or :class:`dict` but the leafs must all be
-              an :class:`array`.
+              :class:`list`, :class:`tuple` or :class:`dict`. Leafs which are not
```
angeloskath: Absolutely a nitpick but is it not `leaves` ?

angeloskath (approved):
Love it. Don't know if I had a single important comment to provide. I especially like the comprehensive tests added.

Maybe the only thing is the wording in the optimizer init which makes it unclear whether we pass the parameters or gradients.

python/src/random.cpp
awni: Oops!

python/src/random.cpp
```diff
@@ -62,8 +105,9 @@ void init_random(py::module_& parent_module) {
          const ScalarOrArray& high,
          const std::vector<int>& shape,
          std::optional<Dtype> type,
-         const std::optional<array>& key,
+         const std::optional<array>& key_,
          StreamOrDevice s) {
+        auto key = key_.value_or(default_key().next());
```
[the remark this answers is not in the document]
awni: That's a great point, I think it should not do that

python/src/random.cpp
awni: Actually, `release()` is a function I wrote that does the dec ref on the underlying Python object 😅

python/mlx/optimizers.py
awni: Well really it can be either, In fact it just needs to be a tree that has the same shape / type as the gradients.

But the typical usage would be `optimizer.init(model.trainable_parameters())`

awni: I will make it more clear in the doc

awni:
🚀  away

PR #630, opened 2024-02-05
added mlp notebook example

hinriksnaer:
## Proposed changes

I noticed that there are no jupyter notebook examples in the project. I decided to port the [MLP example](https://ml-explore.github.io/mlx/build/html/python/nn/module.html#mlx.nn.Module) into a simple jupyter notebook that can be run by any new user to get a sense of how mlx works. 

Feel free to make any suggestions for modifications if you like the idea. I am currently not sure if any modifications to the documentation are required.

## Checklist

### DIFF
--- a/benchmarks/python/rope_bench.py
+++ b/benchmarks/python/rope_bench.py
@@ -1,0 +1,35 @@
+# Copyright © 2023-2024 Apple Inc.
+
+import mlx.core as mx
+import mlx.nn as nn
+from time_utils import time_fn
+
+
+def time_rope():
+    rope = nn.RoPE(4096)
+
+    # vec
+    x = mx.random.uniform(shape=(1, 4096)).astype(mx.float16)
+    mx.eval(x)
+
+    def rope_vec(x):
+        for _ in range(32):
+            x = rope(x)
+        return x
+
+    time_fn(rope_vec, x)
+
+    # matrix
+    x = mx.random.uniform(shape=(1024, 4096)).astype(mx.float16)
+    mx.eval(x)
+
+    def rope_mat(x):
+        for _ in range(32):
+            x = rope(x)
+        return x
+
+    time_fn(rope_mat, x)
+
+
+if __name__ == "__main__":
+    time_rope()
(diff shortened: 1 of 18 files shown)


angeloskath:
Thanks for the contribution @hinriksnaer!

Personally I don't think this particular example adds much compared to the one in the docs https://ml-explore.github.io/mlx/build/html/examples/mlp.html or the script itself https://github.com/ml-explore/mlx-examples/tree/main/mnist . Maybe people feel differently @jagrit06 @awni ?

I think it may make more sense with a visual example like the normalizing flow one. However, given that this would also just duplicate an example, I don't know...

hinriksnaer:
I appreciate the feedback @angeloskath!

This is definitely a matter of preference so I understand if this does not fit within the current scope. Personally, when looking into new frameworks, I really appreciate being able to jumpy straight into a notebook and get a feel for the step-by-step process in a hands on manner.  

If you have any suggestions for alternative examples that may contribute more value, feel free to suggest them.

awni:
I can see how for notebook users having a quick way to get started with MLX might be useful, but it's also pretty easy to copy and paste something into a notebook right? Is there something we could do to make the example easier in that respect? MLX isn't really meant to run in colab so it would have to be local, and we couldn't really have like an `open in colab` button.

On the downside more surface area = more maintenance burden for us if we change anything. 

I would defer on adding this until we have a better strategy and sense of how useful it is. One way to get started is for you to make a repo and add some notebooks there. I would gladly point to people to it. Wdyt?

Stealeristaken:
> I can see how for notebook users having a quick way to get started with MLX might be useful, but it's also pretty easy to copy and paste something into a notebook right? Is there something we could do to make the example easier in that respect? MLX isn't really meant to run in colab so it would have to be local, and we couldn't really have like an `open in colab` button.
> 
> On the downside more surface area = more maintenance burden for us if we change anything.
> 
> I would defer on adding this until we have a better strategy and sense of how useful it is. One way to get started is for you to make a repo and add some notebooks there. I would gladly point to people to it. Wdyt?

I tried to create some notebooks with MLX, but it didn't go well. I can't remember the exact errors I faced, but even after creating an MLX array, the notebook kept returning an exception. On the other hand, I believe it would be beneficial to include notebook examples in MLX. I see many ML engineers or data scientists using silicon chips and encountering errors when they try to use traditional libraries.

awni:
@Stealeristaken sounds like you might have some issues with either your Jupyter setup or your MLX setup. The two are pretty independent so if they are both working they should work together. I have no problem doing MLX stuff in a notebook.

Feel free to file an issue if you would like any assistance debugging. But we'll need more information on the issues you were facing..

Issue #632, opened 2024-02-05
access out of bounds with matmul


davidkoski:
**Describe the bug**

Metal kernel accesses memory outside one of its inputs.

**To Reproduce**

Here is the repro program, `test.py`:

```python
import mlx.core as mx

a = mx.zeros([784, 256])
b = mx.zeros([256, 32])
c = mx.matmul(a, b)
mx.eval(c)
```

Run like this (this is equivalent to turning on Metal API Validation and Metal Shader Validation in Xcode.  I trimmed down the env variables quite a bit :-)

Oh, this assumes you have Xcode installed in /Applications/Xcode.app:

```
[...]

PR #634, opened 2024-02-05
Remove masks from BlockLoader and clear out load case for invalid thread

jagrit06:
## Proposed changes

Fixes out of bounds reads issue #632 
I was planning to change the masks of gemm loader anyway, there wasn't any significant regression on the M2 ultra 

## Checklist

### DIFF
--- a/mlx/backend/metal/kernels/steel/gemm/gemm.h
+++ b/mlx/backend/metal/kernels/steel/gemm/gemm.h
@@ -89,20 +89,9 @@ struct GEMMKernel {
     // Appease the compiler
     (void)l;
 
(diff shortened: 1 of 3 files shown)


awni (approved):
Tests clear 🙈 

Also just curious, was there an actual bug there, e.g loading and using invalid memory?

davidkoski:
I confirm this fixes the issue I saw on iOS + my test program on macOS.  Thanks!

jagrit06:
We were loading invalid memory but not using it - it was a case of the thread being entirely out of bounds and none of its accesses being in bounds - so the default load it did was incorrect. It would do a read but then discard it so it never popped up on the Macs I supposed

PR #635, opened 2024-02-06
Skip compile when transforming

awni:
## Proposed changes

As per title. This let's us simplify and fuse much more than if we transform over the compiled primitive directly.

Here are a few examples:

### Compile compiled functions
```python
@mx.compile
def fun(x):
    @mx.compile
    def inner(y):
        return mx.abs(mx.exp(y))

    return inner(inner(x))
mx.export_to_dot("graph.dot", fun(mx.array([1.0, 2.0])))
```

Now produces a single fused primitive:

Screenshot 2024-02-05 at 4 27 11 PM

### Grad compiled functions

```python
@mx.compile
def fun(x):
    return mx.exp(mx.exp(x))

x = mx.array(1.0)
out = mx.compile(mx.value_and_grad(fun))(x)
mx.export_to_dot("graph.dot", out)
```

Also now produces a single primitive with no duplicate work since the VJP reuses the exp's.

Screenshot 2024-02-05 at 4 26 39 PM

### DIFF
--- a/mlx/compile.cpp
+++ b/mlx/compile.cpp
@@ -86,20 +86,20 @@ std::vector<array> Compiled::vjp(
     const std::vector<array>& cotangents,
     const std::vector<int>& argnums,
     const std::vector<array>& outputs) {
-  throw std::runtime_error("[Compiled] Cannot VJP Compiled primitive.");
+  throw std::runtime_error("[Compiled] Cannot vjp primitive.");
 }
 
 std::vector<array> Compiled::jvp(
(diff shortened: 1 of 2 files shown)


angeloskath (approved):
This is awesome.

It will incidentally break some kernel-gen tests cause I was evaluating without releasing the primitive using the tracing flags. I will do it properly by using the primitive now :-)

awni:
Sorry for breaking your tests 😬

Issue #618, opened 2024-02-03
Question: Add mean and std parameters to normal distribution


noahfarr:
Hello,

I've observed that in core.random, when generating a normal distribution, we're limited to specifying just the shape, dtype, and PRNG key, without the option to set the mean and variance directly. While I understand we can adjust the generated samples by rescaling to achieve the desired mean and standard deviation, it seems it could be more convenient to include mean and std as optional parameters. Is there a specific reason for not incorporating these options?

awni:
I see [np.random.normal](https://numpy.org/doc/stable/reference/random/generated/numpy.random.normal.html) has `loc` and `scale` parameters.

On the other hand, Jax [random.normal](https://jax.readthedocs.io/en/latest/_autosummary/jax.random.normal.html) does not.

I don't feel too strongly about it one way or the other. You can see from the two styles below including the params really doesn't save that much from a syntax standpoint:

`std * mx.random.normal(shape=shape) + mean` and `mx.random.normal(shape=shape, mean=mean, std=std)`

But the former might not be as clear to everyone, so from that perspective including the mean and std params seems reasonable..

If no-one comments here that we shouldn't do it, I would say feel free to send a PR.

noahfarr:
Okay, thank you for the quick response. [torch.normal](https://pytorch.org/docs/stable/generated/torch.normal.html) and [tf.random.normal](https://www.tensorflow.org/api_docs/python/tf/random/normal) also accept a mean/loc and std parameter so I think for most people it would be expected to be able to do this.

noahfarr:
FYI: I created a pull request

angeloskath:
I am closing this issue since the PR is merged

PR #638, opened 2024-02-06
Add loc and scale to random.normal

noahfarr:
## Proposed changes

Currently, random.normal always returns results with mean 0 and std 1. To get values with some other mean or std you currently need to rescale them. I added an optional loc and an optional scale parameter to random.normal that does the rescaling directly like in numpy, torch and tf. 

Issue: #618 

## Checklist

### DIFF
--- a/mlx/random.cpp
+++ b/mlx/random.cpp
@@ -161,8 +161,12 @@ array normal(
   auto low = array(std::nextafter(-1.0f, 0.0f), dtype);
   auto high = array(1.0f, dtype);
   auto samples = uniform(low, high, shape, dtype, key, stream);
-  auto normal_samples = multiply(array(std::sqrt(2.0), dtype), erfinv(samples, stream), stream);
-  return add(multiply(normal_samples, array(scale, dtype), stream), array(loc, dtype), stream);
+  auto normal_samples =
+      multiply(array(std::sqrt(2.0), dtype), erfinv(samples, stream), stream);
+  return add(
+      multiply(normal_samples, array(scale, dtype), stream),
+      array(loc, dtype),
+      stream);
 }
 
 array randint(
--- a/mlx/random.cpp
+++ b/mlx/random.cpp
@@ -153,14 +153,16 @@ array uniform(
 array normal(
     const std::vector<int>& shape,
     Dtype dtype,
+    const float loc,
+    const float scale,
     const std::optional<array>& key /*= nullopt */,
     StreamOrDevice s /* = {} */) {
   auto stream = to_stream(s);
   auto low = array(std::nextafter(-1.0f, 0.0f), dtype);
   auto high = array(1.0f, dtype);
   auto samples = uniform(low, high, shape, dtype, key, stream);
-  return multiply(
-      array(std::sqrt(2.0), dtype), erfinv(samples, stream), stream);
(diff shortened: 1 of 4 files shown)


python/tests/test_random.py
```diff
@@ -80,6 +80,20 @@ def test_normal(self):
...
+        self.assertTrue(mx.all(mx.abs(a - b) < 1e-5).item())
+
+        # Generate with a given mean and standard deviation and dtype
+        a = mx.random.normal(shape=(3, 2), loc=1, scale=2.0, dtype=mx.float16)
+        self.assertEqual(a.dtype, mx.float16)
+        b = mx.array(
+            [[1.26758, 0.763184], [-1.0293, 2.33984], [0.598633, 2.9082]],
+            dtype=mx.float16,
+        )
+        self.assertTrue(mx.all(mx.abs(a - b) < 1e-5).item())
```
awni, line 95: I would change these tests:

1. Instead of hardcoding random samples, just check that if you sample once with `loc` and `scale` and once without them (but with the same key) and then apply the transform, you get the right result. That way if we ever change the RNG this will not be so brittle.
2. Use `mx.allclose` for comparing arrays: e.g. `self.assertTrue(mx.allclose(a, b))`

mlx/random.cpp
```diff
@@ -153,14 +153,20 @@ array uniform(
...
   auto high = array(1.0f, dtype);
   auto samples = uniform(low, high, shape, dtype, key, stream);
-  return multiply(
-      array(std::sqrt(2.0), dtype), erfinv(samples, stream), stream);
+  auto normal_samples =
+      multiply(array(std::sqrt(2.0), dtype), erfinv(samples, stream), stream);
+  return add(
+      multiply(normal_samples, array(scale, dtype), stream),
+      array(loc, dtype),
+      stream);
```
awni, line 169:
```suggestion
  auto samples =
      multiply(array(std::sqrt(2.0), dtype), erfinv(samples, stream), stream);
  if (scale != 1.0) {
    samples = multiply(samples, array(scale, dtype), stream);
  }
  if (loc != 0.0) {
    samples = add(samples, array(loc, dtype), stream);
  }
  return samples;
```

awni: Avoid extra ops in common case.

mlx/random.cpp
```diff
@@ -153,14 +153,20 @@ array uniform(
 array normal(
     const std::vector<int>& shape,
     Dtype dtype,
+    const float loc,
+    const float scale,
```
awni, line 157:
```suggestion
    const float loc /* = 0.0 */,
    const float scale /* = 1.0 */,
```

awni (changes requested):
Left some comments, also can you double check the C++ tests pass?

noahfarr:
C++ tests are passing now after modifying the function overloading. I also fixed the other things you mentioned

mlx/random.cpp
noahfarr: Done

python/tests/test_random.py
noahfarr: Done

mlx/random.cpp
noahfarr: Done

awni (approved):
Very nice, thanks!!

PR #639, opened 2024-02-06
Metadata support for safetensors

### LINKED ISSUE #633: [Feature request] Add metadata to safetensors saving and loading
Just found the model weight saved in SafeTensor by mlx cannot be loaded by Transformers. An error is thrown, stating that the format is not supported. This issue seems to caused from the metadata format being saved as `mlx` (https://github.com/ml-explore/mlx/blob/main/mlx/io/safetensor.cpp#L165). However, mlx has no trouble loading HF model weights. so, I'm curious about the differences in the saved weights between them. 
Thanks in advance.

dc-dc-dc:
## Proposed changes

Adds metadata support for safetensors, Closes #633 

## Checklist

### DIFF
--- a/mlx/io/safetensor.cpp
+++ b/mlx/io/safetensor.cpp
@@ -93,7 +93,7 @@ Dtype dtype_from_safetensor_str(std::string str) {
 }
 
 /** Load array from reader in safetensor format */
-SafeTensorsLoad load_safetensors(
+SafetensorsLoad load_safetensors(
     std::shared_ptr<io::Reader> in_stream,
     StreamOrDevice s) {
   ////////////////////////////////////////////////////////
@@ -144,7 +144,7 @@ SafeTensorsLoad load_safetensors(
   return {res, metadata_map};
 }
 
-SafeTensorsLoad load_safetensors(const std::string& file, StreamOrDevice s) {
+SafetensorsLoad load_safetensors(const std::string& file, StreamOrDevice s) {
   return load_safetensors(std::make_shared<io::FileReader>(file), s);
 }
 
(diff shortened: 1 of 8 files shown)


dc-dc-dc:
added some aliases so its a little more readable, but if not desired i can undo that commit

python/src/ops.cpp
```diff
@@ -3225,6 +3226,7 @@ void init_ops(py::module_& m) {
         Args:
             file (file, str): File in which the array is saved.
             arrays (dict(str, array)): The dictionary of names to arrays to be saved.
+            metadata (dict(str, str)): The dictionary of metadata to be saved.
```
awni, line 3229:
```suggestion
            metadata (dict(str, str), optional): The dictionary of metadata to be saved.
```

mlx/io/safetensor.cpp
```diff
@@ -204,7 +208,8 @@ void save_safetensors(
 
 void save_safetensors(
     const std::string& file_,
-    std::unordered_map<std::string, array> a) {
+    std::unordered_map<std::string, array> a,
+    std::unordered_map<std::string, std::string> metadata) {
```
awni, line 212:
```suggestion
    std::unordered_map<std::string, std::string> metadata /* = {} */) {
```

awni (commented):
This looks really great, thanks for the quick addition! Mostly cosmetic suggestions, let me know when done and we can merge it!

dc-dc-dc:
addressed the comments in the latest commit, not finding instances of `safe_tensors` what file is it in?

awni:
> not finding instances of safe_tensors what file is it in

Sorry I meant we don't treat is as two words (e.g. we don't do `safe_tensors` we do `safetensors`

awni:
@dc-dc-dc would you mind adding one more test in python to cover the binding?

awni (approved):
Looks great, thanks!!

PR #641, opened 2024-02-06
Remainder negative numerator bug fixed

### LINKED ISSUE #640: [BUG] Remainder does not work properly when the numerator is negative
**Describe the bug**
The remainder does not function properly when the numerator is negative.

**To Reproduce**

Include code snippet
```python
import mlx.core as mx
print("Result:", (-mx.ones(1))%5)
```
Prints:
"Result: array([-1], dtype=float32)"

```python
import numpy as np
print("Result:",(-np.ones(1))%5)
```
Prints:
"Result: [4.]"

**Expected behavior**
Same result with Numpy

 - OS Version MacOS 14.2.1
 - Version: 0.1.0

abdussamettrkr:
## Proposed changes
Fix  remainder negative numerator is negative bug fixes #640.

## Checklist

### DIFF
--- a/mlx/backend/accelerate/primitives.cpp
+++ b/mlx/backend/accelerate/primitives.cpp
@@ -298,13 +298,22 @@ struct RemainderFn {
   std::enable_if_t<!std::is_integral_v<T>, T> operator()(
       T numerator,
       T denominator) {
-    return std::fmod(numerator, denominator);
+    auto r = std::fmod(numerator, denominator);
+    if (r < 0 && denominator > 0)
+      r += denominator;
+    return r;
   }
 
   template <typename T>
   std::enable_if_t<std::is_integral_v<T>, T> operator()(
       T numerator,
       T denominator) {
+    auto r = numerator % denominator;
+    if (r < 0 && denominator > 0)
+      r += denominator;
+    return r;
+  }
+  complex64_t operator()(complex64_t numerator, complex64_t denominator) {
     return numerator % denominator;
   }
 };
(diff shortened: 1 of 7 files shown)


mlx/backend/metal/kernels/binary.metal
```diff
@@ -15,10 +15,33 @@ struct Divide {
...
+      r+= y;
+    return r; 
+  }
+  template <> complex64_t operator()(complex64_t x, complex64_t y) {
+    return x % y; 
+  }
+  template <> float operator()(float x, float y) {
+    float r = fmod(x, y);
+    if (r < 0 && y >0)
+      r+= y;
```
angeloskath, line 30: Spaces after `r` and `y >`. Possibly we also need an unsigned numbers specialization that doesn't do the check. Otoh the compiler should be able to optimize that away so perhaps we are good.

angeloskath (changes requested):
Thanks for fixing this. I left a couple of comments, almost exclusively stylistic.

Feel free to ignore the unsigned inputs specialization since the compiler can most probably handle that case.

mlx/backend/metal/kernels/binary.metal
abdussamettrkr: I added specialization for eval_cpu and eval, but i am not sure how to do it in metal.

mlx/backend/accelerate/primitives.cpp
```diff
@@ -298,13 +298,22 @@ struct RemainderFn {
   std::enable_if_t<!std::is_integral_v<T>, T> operator()(
       T numerator,
       T denominator) {
-    return std::fmod(numerator, denominator);
+    auto r = std::fmod(numerator, denominator);
```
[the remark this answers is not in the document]
abdussamettrkr: I suppose I did

abdussamettrkr:
Closes also #650

angeloskath (approved):
I added the Metal specialization, an extra test, rebased on main and am merging. Thanks for fixing this!

PR #664, opened 2024-02-11
Implement the 'where' primitive for conditional selection

### LINKED ISSUE #598: Wrong output was obtained when x and y both have inf in mlx.coer.where
**Describe the bug**
Wrong output was obtained when x and y both have inf in mlx.coer.where. Whether it's from x or y, it should get inf instead of nan.

**To Reproduce**

```python
import mlx.core as mx
import numpy as np
condition_mx = mx.array([True, False, True, False], dtype=mx.bool_)
x_mx = mx.array([float('inf'), 2, 3, 4], dtype=mx.float32)
y_mx = mx.array([float('inf'), 20, 30, 40], dtype=mx.float32)

condition_np = np.array([True, False, True, False])
x_np = np.array([float('inf'), 2, 3, 4], dtype=np.float32)
y_np = np.array([float('inf'), 20, 30, 40], dtype=np.float32)

result_mlx = mx.where(condition_mx, x_mx, y_mx)
print("Result on Mlx:", result_mlx)

result_np = np.where(condition_np, x_np, y_np)
print("Result using NumPy:", result_np)
```
image

**Expected behavior**
array([inf, 20, 3, 40], dtype=float32)

 - OS Version: MacOS 14.2.1
 - Version 0.0.10

Rifur13:
## Proposed changes

This PR implements a `Select` primitive for conditional selection that will replace the arithmetic in the `where` op. It now handles infinite values correctly and closes #598.

```
def fn(condition, a, b):
      if condition:
          return a
      else:
          return b
``` 

I propose breaking this ternary op into 2 binary ops and execute them sequentially.
```
def fn(condition, a):
	if condition:
		out = a

def fn(condition, b):
	if condition:
		out = b
```

It involves refactoring the existing binary op.
Concretely:
1. Delegate regular op() calls in scalar to scalar cases to a DefaultScalarScalar struct. This makes it easier to swap out the operations.
2. Fully templating the input types, T1 and T2.
3. Change how 

A generalized ternary op would be quite complicated, dealing with all the different scenarios of broadcasting with 3 inputs. I’m certainly up for the challenge but I’m not sure if it’s worth the effort *only* for this primitive. Are there other ternary ops?

Please let me know you think and if you have any concerns with this approach. Single ops tests show it to be 20-30% faster.

Before:
Timing where ... 0.96897 msec
Timing (scalar-vector) where ... 0.41628 msec
Timing (scalar-scalar broadcast) where ... 0.23243 msec

After:
Timing where ... 0.69184 msec
Timing (scalar-vector) where ... 0.29164 msec
Timing (scalar-scalar broadcast) where ... 0.14348 msec

## Checklist


awni:
@Rifur13 this looks really nicely done! Thanks for adding it. We will review soon!

mlx/ops.cpp
```diff
@@ -1152,13 +1150,23 @@ array isneginf(const array& a, StreamOrDevice s /* = {} */) {
...
+    const array& b,
+    const array& c,
     StreamOrDevice s /* = {} */) {
-  // TODO, fix this to handle the NaN case when x has infs
-  auto mask = astype(condition, bool_, s);
-  return add(multiply(x, mask, s), multiply(y, logical_not(mask, s), s), s);
+  auto condition = a;
+  if (a.dtype() != bool_) {
+    condition = not_equal(a, zeros_like(a), s);
+  }
```
awni, line 1160:
```suggestion
  auto condition = astype(a, bool_, s);
```

Should work right?

mlx/backend/common/select.h
```diff
@@ -0,0 +1,160 @@
...
+        } else {
+          out.copy_shared_buffer(b);
+        }
+      } else {
+        out.set_data(allocator::malloc_or_wait(out.nbytes()));
+      }
+      break;
+    case VectorScalar:
+      // The buffer of `condition` can't be donated because it's called twice in
+      // the 'select' op.
```
awni, line 77: 👍

angeloskath (changes requested):
@Rifur13 this looks really nice indeed! I am excited for this to be merged.

I have one suggestion that will require some edits but it's mostly moving things around and will simplify quite a few things. I like the fact that the actual computation is happening via a binary select op. However, I think that the `Select` primitive should really just implement this select op and take inverted as an argument.

Given the buffer donation I don't think there is any negative aspect in using the above instead of a ternary op that launches two binary ops.

Rifur13:
@angeloskath Thanks! Can you elaborate on what you're thinking? Are you suggesting to change the `Select` primitive to actually be a select binary op, and then the `where` op would be implemented as 2 calls to `Select`?

angeloskath:
@Rifur13  yeah this is exactly what I had in mind. On second thought though it might be a bit harder since it would have to rely on buffer donation which is not something we do so the other option is to combine with a 3rd op (eg a sum) which I think is not as good of an idea.

Still it feels like a lot of duplication that I am trying to avoid... I will give it another look and try to suggest sth more concrete.

Rifur13:
gentle ping

angeloskath:
@Rifur13 I know. You did an amazing PR and I said something relatively vague with a request for change. I 'll take a look over the weekend and either have something concrete or change to an approve. Thanks for your patience!

angeloskath:
Ok, so looking a bit closer to this, I think the reason I am hesitating is that there is a lot of code that feels redundant in this implementation because it performs the following refactoring but without performing the refactoring.

The binary and unary ops in MLX are in the forms of functions that take some inputs and return their output. As a result implementing an op that writes conditionally, is not possible without refactoring the ops to take the output as a pointer to write to.

This is the solution that is done for the CPU in this PR while for the GPU a new set of kernels is written just for select to perform the conditional writing.

I am open to suggestions but I see 3 ways forward:
1. Actually do the refactoring and change all ops to take the output as a pointer so they can perform conditional writing.
2. Implement `where` using 3 reads/writes by implementing a select that writes the selected or 0 and then summing the results.
3. Implement `where` with a ternary op.

Personally, I prefer 2 and 3 because I think 1 is a slightly large undertaking just for implementing `where`. From 2 and 3 I don't really have a preference, 2 will be extremely easy to write, slower initially but with compile probably the same as 3. 3 will require a bit more code but I think it will be clearer.

Rifur13:
Thanks for the comments!

Can you clarify 1)? The functions that the binary and unary ops use are wrapped in a struct with an `operator()`, like `DefaultScalarScalar`, and they do have a pointer to the output.

The reason that I’m hesitant on 3) is that there will be 8 TernaryOpTypes (ScalarScalarScalar, ScalarScalarVector, and so on…).  Implementing specialized ops for all case seems like a big undertaking, and there’s a lot of optimizations we would lose. Of course, if speed isn’t important a single general n-dim op will be easy to write.

What do you think of option 4) in my last comment. We will have a specialized binary op for **one** conditional selection (all other refactorings will be removed). I think it’s very similar to 2), but instead of writing “or 0”, the correct value will already be set.

angeloskath:
Sure, so adding `DefaultScalarScalar` does exactly that refactoring. We should do the same to the GPU if we are going down that road. But that would mean changing all the functors like `Add` or `Sqrt` to that format which I feel is a bit excessive.

As an aside, there is no need for `DefaultScalarScalar`. It was and still is the `op`. It is just a nice way of changing all the ops to the format of taking an output pointer without actually changing the ops. The rest are for optimization purposes so it shouldn't be too bad to start option 3 without all of them and go from there.

I don't think option 4 changes the problem. It is great since when combined with donation it may make this a single read/write but the problem is the conditional writing of the output which one would still have to do. Option 2 doesn't do conditional writing as it always writes something, either the value or 0.

It seems to me that the simplest option would be to go with option 2 if you want to avoid writing the ternary op (which to be fair isn't unreasonable).

Rifur13:
Thanks I understand see your point, but I still don't quite understand why the conditional writing in option 4 is an issue if we treat it as a special binary op and don't refactor any of the existing binary op.

Option 3 sounds good though. I'll implement the General case for now, and we can add optimizations as needed in the future. Stay tuned!

angeloskath (approved):
This was super fast and it looks fantastic to me. Thanks for changing to `enum class` we should probably be doing a bit more of that (even though more verbose probably unavoidable to have conflicts as the project grows).

I pushed a small diff to enable compilation of the select op and after the tests clear we can merge.

mlx/primitives.cpp
```diff
@@ -48,6 +48,54 @@ std::tuple<array, array, int> vmap_binary_op(
   return {a, b, to_ax};
 }
 
+std::tuple<array, array, array, int> vmap_ternary_op(
+    const std::vector<array>& inputs,
+    const std::vector<int>& axes,
+    const Stream& stream) {
+  assert(inputs.size() == 3);
+  assert(axes.size() == 3);
```
awni, line 56: ❤️  awesome thanks!

PR #676, opened 2024-02-12
Custom primitive + RoPE fat op

awni:
## Proposed changes

- [x] add RoPE kernel
- [x] test transforms of custom primitive
- [x] benchmarks

### DIFF
--- a/mlx/backend/metal/rope.cpp
+++ b/mlx/backend/metal/rope.cpp
@@ -1,7 +1,5 @@
 // Copyright © 2023-2024 Apple Inc.
 
-#include <iostream>
-
 #include "mlx/backend/metal/utils.h"
 #include "mlx/extensions.h"
 #include "mlx/primitives.h"
--- a/mlx/backend/metal/rope.cpp
+++ b/mlx/backend/metal/rope.cpp
@@ -1,6 +1,6 @@
 // Copyright © 2023-2024 Apple Inc.
 
-#include "mlx/ext.h"
+#include "mlx/extensions.h"
 #include "mlx/primitives.h"
 
 namespace mlx::core::ext {
--- a/mlx/backend/metal/rope.cpp
+++ b/mlx/backend/metal/rope.cpp
@@ -1,5 +1,7 @@
 // Copyright © 2023-2024 Apple Inc.
 
+#include <iostream>
+
 #include "mlx/backend/metal/utils.h"
 #include "mlx/extensions.h"
 #include "mlx/primitives.h"
--- a/mlx/backend/metal/rope.cpp
+++ b/mlx/backend/metal/rope.cpp
@@ -1,0 +1,14 @@
+// Copyright © 2023-2024 Apple Inc.
+
+#include "mlx/ext.h"
+#include "mlx/primitives.h"
+
+namespace mlx::core::ext {
+
+void RoPE::eval_gpu(
+    const std::vector<array>& inputs,
+    std::vector<array>& outputs) {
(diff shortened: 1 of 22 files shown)


angeloskath:
Some benchmarks of the kernel on my M2 air

Before
```
Timing rope_vec ... 3.99837 msec
Timing rope_mat ... 63.36270 msec
```

After
```
Timing rope_vec ... 0.61199 msec
Timing rope_mat ... 6.72898 msec
```

The tests fail on float16 and bfloat16 but due to numerical issues. Tomorrow I will do a quick check on the performance if we do all of the computation in float32 in the kernel since it probably doesn't matter at all performance wise.

mlx/backend/metal/rope.cpp
```diff
@@ -0,0 +1,56 @@
...
+  auto& out = outputs[0];
+
+  if (in.ndim() != 3) {
+    throw std::runtime_error(
+        "[RoPE] Only 3 dimensions are supported (batch x sequence x dims)");
+  }
+  if (dims_ != in.shape(-1)) {
+    throw std::runtime_error("[RoPE] Partial RoPE application not supported");
+  }
+  if (in.flags().row_contiguous && in.is_donatable()) {
```
awni, line 25: We need a contig and copy check before this right?

awni:
Wow, that's so fast! We can also increase the tolerance for the lower precision tests if that's simpler.

mlx/backend/metal/rope.cpp
angeloskath: Not sure what the copy check is. Also `row_contiguous` is stricter than `contiguous` is it not? ie all `row_contiguous` arrays are `contiguous` but not the other way around.

awni: I meant, if it's not contiguous, we should make a contiguous copy

awni: Actually I think I missed it, I was looking for elem to loc, but you hardcoded the strides.. so I should be ok

awni: Do we need to check here though that the input has the same size as the output? If it's broadcasted e.g. along the last axis it would be incorrect to donate right?

awni:
Make this a real PR since I think we are almost done.

mlx/backend/metal/rope.cpp
angeloskath: Yeah, I hardcoded the strides cause the grid is launched with half the last dimension and it can't be delegated to a simple `elem_to_loc`. I would have to do something like multiply `pos.x` by 2 and then pass to `elem_to_loc` etc. I think this is equally readable but I am open to suggestions :-)

Regarding broadcasting, a broadcasted array wouldn't be `row_contiguous` so this check should be fine donation-wise, right?

python/src/extensions.cpp
```diff
@@ -0,0 +1,56 @@
...
+         int offset,
+         const StreamOrDevice& s /* = {} */) {
+        return ext::rope(x, dims, traditional, base, scale, offset, s);
+      },
+      "x"_a,
+      "dims"_a,
+      "traditional"_a,
+      "base"_a,
+      "scale"_a,
+      "offset"_a,
```
angeloskath, line 36: Do you think we should make the above keyword only? It would be verbose but error free...

angeloskath (approved):
This looks great. I really like the `Custom` primitive.

python/src/extensions.cpp
awni: I think so, yes.

PR #677, opened 2024-02-13
Quantized matmul fix

### LINKED ISSUE #665: [BUG]  Quantized matrix-vector gives bad result with output < 8
Bug with small output in quantized matrix-vector multiply. I think these sizes should work (at least it does not throw when trying to quantize).

```
import mlx.core as mx

x = 1e-1 * mx.random.normal(shape=(4096,)).astype(mx.float16)
w = 1e-1 * mx.random.normal(shape=(4, 4096)).astype(mx.float16)

print(x @ w.T)

w, s, b = mx.quantize(w)
x = mx.quantized_matmul(
        x,
        w,
        scales=s,
        biases=b,
        transpose=True)

print(x)
```
First print:
```
array([-0.780273, -0.264648, -0.37793, -0.293945], dtype=float16)
```

Second one gives:
```
array([nan, nan, nan, nan], dtype=float16)
```

angeloskath:
Fixes issues with quantized matmuls and sizes that are unaligned. This should fix #665 .

Importantly this changes the way `qmv` loads the input vector such that the first simd group does all the loading. I didn't notice any performance regression from this change but also as we 've discussed these kernels are due for a performance improvement pass anyway.

### DIFF
--- a/mlx/backend/metal/kernels/quantized.metal
+++ b/mlx/backend/metal/kernels/quantized.metal
@@ -39,11 +39,12 @@ template <typename T, const int BM, const int BN, const int group_size, const in
 
   static_assert(BN == SIMD_SIZE, "qmv expects BN to be equal to SIMD_SIZE");
 
+  (void)lid;
+
   constexpr int bitmask = (1 << bits) - 1;
   constexpr int el_per_thread = 32 / bits;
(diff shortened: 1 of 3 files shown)


awni (approved):
Thanks !!!

PR #679, opened 2024-02-13
Adds device context manager

dc-dc-dc:
## Proposed changes

Utilize python contextlib to easily swap out what device is being used without having to keep track, updated fft tests to showcase effectiveness.

## Checklist


awni:
Very nice! I've been wanting to do something like this on the C++ side to simplify writing ops which take streams

dc-dc-dc:
Yeah its pretty handy when you only have a device specific implementation. In c++ could probably achieve something similar with a lambda + macro

angeloskath:
Love it. We have been putting off doing that for a while now.

If you plan on doing the same in C++ then I 'd suggest RAII instead of lambda. Basically a simple struct that keeps the old device and sets the desired device on construction and resets to the old device on destruction. Managing device scopes is literally done with C++ scopes after that. Also this should be able to take a stream instead of simply a device so we can use it in the ops or gradients to simplify things if we want to.

awni:
💯 I was about to make the exact same comment re stream and RAII on the C++ side.

I don't want to overly complexify things, but I also wonder if we should do it on the C++ side and bind it in core instead of putting it in `mlx.utils` 🤔 . Seems sufficiently useful.

dc-dc-dc:
Something like this ? (kindof annoying the signatures dont match as default_device returns a const ref)
```c++
struct StreamContextManager {
  public:
  StreamContextManager(StreamOrDevice s) {
    _stream = default_stream(default_device());
    set_default_stream(to_stream(s));
  }

  ~StreamContextManager() {
    set_default_stream(_stream);
  }

  private:
    Stream _stream;
};
```

dc-dc-dc:
Pushed some progress, though it seems like after calling `set_default_device` in c++ it doesn't seem to have an affect from python. I can confirm that the stream is cpu in context manager but the subsequent ops calls still use gpu. Will mess with this more, but just pushing in case this is something you guys ran into before

dc-dc-dc:
nvm, fixing it now

dc-dc-dc:
Updated and added docs, works nicely with the `with` keyword. 

Couldn't get the decorator part working as I'd like, but it's not essential so will just leave it as is for now. If anyone is interested in adding decorator support, essentially just need to add `__call__` def to the `PyStreamContextManager` class that takes in a `py::function`  argument and return a wrapper function that calls `enter` `func(*args, **kwargs)` `exit` in that order.

python/src/utils.cpp
```diff
@@ -0,0 +1,63 @@
...
+  StreamOrDevice _s;
+  StreamContextManager* _inner;
+};
+
+void init_utils(py::module_& m) {
+  py::class_<PyStreamContextManager>(m, "StreamContextManager", R"pbdoc(
+        A context manager for setting the current device and stream.
+    
+        This class is a Python wrapper around the C++ class
+        `mlx::core::StreamContextManager`. It is used to set the current device
```
angeloskath, line 39: I don't think there is a need for that you can directly say that it can change the default device and stream for the duration of the context as you have it. No need to mention that it wraps the equivalent to the C++ object.

python/src/utils.cpp
```diff
@@ -0,0 +1,63 @@
...
+        and stream for the duration of a Python context block.
+    
+        Example:
+        ```python
+        import mlx.utils as mlx
```
angeloskath, line 44:
```python
import mlx.core as mx
```

python/src/utils.cpp
```diff
@@ -0,0 +1,63 @@
...
+    
+        # Create a context manager for the current device and stream.
+        with mlx.StreamContextManager(mx.cpu):
```
angeloskath, line 47: Same typo as above.

python/mlx/utils.py
```diff
@@ -1,5 +1,4 @@
 # Copyright © 2023 Apple Inc.
-
```
angeloskath, line 2: Probably removed unnecessarily when removing the original Python implementation of the context manager.

angeloskath (changes requested):
Love it! Left a couple of comments for some typos but otherwise looks great!

mlx/utils.h
```diff
@@ -9,6 +9,27 @@
 
 namespace mlx::core {
 
+using StreamOrDevice = std::variant<std::monostate, Stream, Device>;
```
angeloskath, line 12: The header `<variant>` is needed for gcc to compile this. Also the same explicit header include should be removed from `ops.h` now unless it `std::variant` is used elsewhere.

(Didn't see the failed linux build when I did the initial review hence the separate comment)

python/src/utils.cpp
```diff
@@ -0,0 +1,59 @@
...
+    
+        Example:
+        ```python
+        import mlx.core as mx
+    
+        # Create a context manager for the current device and stream.
+        with mx.StreamContextManager(mx.cpu):
+            # Run some code that uses the current device and stream.
+            pass
+        ```
```
awni, line 46: Could you move this to below the `Args` section?

Also you need to use an RST code block here:

```
```suggestion
        Example:
        
        .. code-block:: python
            
            import mlx.core as mx
    
            # Create a context manager for the current device and stream.
            with mx.StreamContextManager(mx.cpu):
                # Run some code that uses the current device and stream.
                pass
```

python/src/utils.cpp
```diff
@@ -0,0 +1,59 @@
...
+#include <pybind11/pybind11.h>
+#include <pybind11/stl.h>
+#include <optional>
+
+namespace py = pybind11;
+using namespace py::literals;
+using namespace mlx::core;
+
+// Slightly different from the original, with python context on init we are not
+// in the context yet. Only create the inner context on enter then de   lete on
```
awni, line 12:
```suggestion
// in the context yet. Only create the inner context on enter then delete on
```

awni:
Yea right it's not so elegant on the implementation side. I don't love that, though I think it could certainly be done and without breaking any downstream code.

But let's at least consider alternative names. For example another option is lower case stream is the context manager:

```
with mx.stream(...):
   ....
```

Personally I like that a lot more than `mx.StreamConstextManager` even though it introduces a bit of ambiguity between `Stream` and `stream`.

Any other suggestions?

dc-dc-dc:
I think dropping the manager part and leaving it as `StreamContext` fits its purpose. 

```python
with mx.StreamContext(mx.cpu):
   ...
```

awni:
> I think dropping the manager part and leaving it as `StreamContext` fits its purpose.

I still like simple `mx.stream` more but that is certainly an improvement.

Just for some points of comparison:

- PyTorch uses `torch.cuda.device(..)` as the context manager
- Jax uses `jax.default_device(..)` as the context manager

awni:
> For streams torch uses a class called StreamContext they have a wrapper called stream that returns an instance of it

Lol so they do both of our suggestions. What's the point of having both, wouldn't one just always use `torch.stream`?

dc-dc-dc:
added the `stream` wrapper function so we support both options similar to torch

awni:
Wonderful, thanks!! That's really nice!

Two more comments, then I think we are good:

- Could you update the C++ to be `StreamContext` as well? Seems better to be consistent there.
- Update the tests and docs to use the `stream` wrapper. I think it would be good to have that as the preferred option.

docs/src/python/devices_and_streams.rst
```diff
@@ -9,6 +9,7 @@ Devices and Streams
   :toctree: _autosummary
 
    Device
+   StreamContext
```
awni, line 12: Add `stream` to do

python/src/utils.cpp
```diff
@@ -0,0 +1,71 @@
...
+          [](PyStreamContext& scm,
+             const std::optional<py::type>& exc_type,
+             const std::optional<py::object>& exc_value,
+             const std::optional<py::object>& traceback) { scm.exit(); });
+  m.def(
+      "stream",
+      [](StreamOrDevice s) { return PyStreamContext(s); },
+      "s"_a,
+      R"pbdoc(
+        Wrap around the Context-manager StreamContext that selects a given stream.
```
awni, line 63:
```suggestion
        A wrapper around :obj:`StreamContext` to select the given stream.
```

python/src/utils.cpp
```diff
@@ -0,0 +1,71 @@
...
+
+        Args:
+            s: The stream or device to set as the current device and stream.
```
awni, line 66:
```suggestion
            s: The stream or device to set as the default device and stream.
```

python/src/utils.cpp
```diff
@@ -0,0 +1,71 @@
...
+        
+        Returns:
+            A context manager for setting the current device and stream.
```
awni, line 69:
```suggestion
            A context manager for setting the default device and stream.
```

python/src/utils.cpp
```diff
@@ -0,0 +1,71 @@
...
+        
+        Args:
+            s: The stream or device to set as the current device and stream.
+
+        Example:
+        .. code-block::python
+          import mlx.core as mx
+      
+          # Create a context manager for the current device and stream.
+          with mx.StreamContext(mx.cpu):
```
awni, line 46: Use `mx.stream` here.

```suggestion
          with mx.stream(mx.cpu):
```

awni:
I pushed a few changes to the docs to fix some issues there.

Also I noticed that the following runs:

```
with mx.stream(None):
  pass
```

That should probably not run I think.

angeloskath (approved):
👍

Issue #31, opened 2023-12-06
Segfaults when running examples using GPU inside a VM


jerpeter1:
When running mlx-examples/mnist on MacOS 14.1.1 VM running via Parallels 19.1.1:

```
user@Users-Virtual-Machine mnist % python main.py --gpu
*** Terminating app due to uncaught exception 'NSInvalidArgumentException', reason: '-[AppleParavirtDevice newArgumentEncoderWithLayout:]: unrecognized selector sent to instance 0x149029e00'
*** First throw call stack:
(
	0   CoreFoundation                      0x000000018a682800 __exceptionPreprocess + 176
	1   libobjc.A.dylib                     0x000000018a179eb4 objc_exception_throw + 60
	2   CoreFoundation                      0x000000018a7343bc -[NSObject(NSObject) __retain_OA] + 0
	3   AppleParavirtGPUMetalIOGPUFamily    0x0000000101fe1118 doUncompressedBlit + 11160
	4   Metal                               0x0000000194729184 -[_MTLDevice newArgumentEncoderWithArguments:structType:] + 136
	5   libmlx.dylib                        0x00000001113f660c _ZN3mlx4core6Gather8eval_gpuERKNSt3__16vectorINS0_5arrayENS2_9allocatorIS4_EEEERS4_ + 1380
	6   libmlx.dylib                        0x00000001113fc54c _ZNSt3__110__function6__funcIZN3mlx4core5metal9make_taskERNS3_5arrayENS_6vectorINS_13shared_futureIvEENS_9allocatorIS9_EEEENS_10shared_ptrINS_7promiseIvEEEEbE3$_2NSA_ISH_EEFvvEEclEv + 148
	7   libmlx.dylib                        0x0000000110d5ff14 _ZN3mlx4core9scheduler12StreamThread9thread_fnEv + 500
	8   libmlx.dylib                        0x0000000110d600d0 _ZNSt3__114__thread_proxyB7v160006INS_5tupleIJNS_10unique_ptrINS_15__thread_structENS_14default_deleteIS3_EEEEMN3mlx4core9scheduler12StreamThreadEFvvEPSA_EEEEEPvSF_ + 72
	9   libsystem_pthread.dylib             0x000000018a531034 _pthread_start + 136
	10  libsystem_pthread.dylib             0x000000018a52be3c thread_start + 8
)
libc++abi: terminating due to uncaught exception of type NSException
zsh: abort      python main.py --gpu
```

[...]

PR #683, opened 2024-02-14
Update gather and scatter to not use Argument Encoder

jagrit06:
## Proposed changes

Virtual Apple silicon devices do not have support for argument encoders, which is a challenge for the CI pipeline. I just adjusted our gather and scatter kernels to not rely on  argument encoders while leaving the core function the same. No noticeable regression in performance was observed on  M3 Max 

Before:
```
====================
X (100, 64), Indices (1000000,)
MLX: 2.767ms
PyTorch: 10.588ms
====================
X (100, 1024), Indices (100000,)
MLX: 2.901ms
PyTorch: 16.767ms
====================
X (4, 1000000), Indices ()
MLX: 0.250ms
PyTorch: 0.622ms
```

After:
```
====================
X (100, 64), Indices (1000000,)
MLX: 2.214ms
PyTorch: 10.772ms
====================
X (100, 1024), Indices (100000,)
MLX: 2.893ms
PyTorch: 17.314ms
====================
X (4, 1000000), Indices ()
MLX: 0.211ms
PyTorch: 0.234ms
```

## Checklist


jagrit06:
Fixes issue #31

awni (commented):
LGTM once it passes the tests. Getting rid of those argument buffers is a pretty nice simplification!

jagrit06:
I don’t think we’d call it a bug - you can check the MTLDevice object to see if it has support for argument encoders and that call comes out false for virtual devices 

I suppose it has to with packing pointers in a buffer and making that redirection available requires something more (there are also some hardware mechanics involved for some versions / things you are able do with argument encoders)

mlx/backend/metal/indexing.cpp
```diff
@@ -63,91 +64,47 @@ void Gather::eval_gpu(const std::vector<array>& inputs, array& out) {
...
-  }
+        inputs[i + 1].shape().end());
 
-  // Allocate the argument buffer
-  auto arg_buf = allocator::malloc_or_wait(arg_enc->encodedLength());
-
-  // Register data with the encoder
-  arg_enc->setArgumentBuffer(static_cast<MTL::Buffer*>(arg_buf.ptr()), 0);
-  for (int i = 0; i < nidx; ++i) {
-    set_array_buffer(compute_encoder, arg_enc, inputs[i + 1], i);
```
awni, line 114: Nit: let's remove this function overload since we should never use it

awni: I think it's also worth removing the argument encoder support from device.cpp, wdyt?

awni: I just love deleting code 😛

jagrit06: Agree, it will just be confusing

Issue #530, opened 2024-01-23
Wrong output when inf is the step in 'mlx.core.arange'


Redmept1on:
**Describe the bug**
According to arange's calculation method (Generate numbers in the half open interval [start, stop)), when step is inf, the generated array should contain the start value, but mlx.core.arange did not achieve this.

**To Reproduce**

Include code snippet
```python
import mlx.core as mx
import numpy as np

start = 0.1
stop = 2
step = float('inf')

#mlx
range = mx.arange(start, stop, step)
print('mlx:', range)

#mlx
np_range = np.arange(start, stop, step)
print('numpy:', np_range)
```
![image](https://github.com/ml-explore/mlx/assets/84373897/449a6f5f-0c15-4fec-9c58-be42b8dc1ceb)

**Expected behavior**
array([0.1], dtype=float32)

**Desktop (please complete the following information):**
 - OS Version: MacOS 14.2.1
 - Version  0.10.0

nakranivaibhav:
@awni I can take this up

awni:
Great!

nakranivaibhav:
@awni Can you help me how to setup an editable version of mlx on my local machine.
I am getting errors(Tried many times). I want to test the code as I make changes along side.
Or is there a better way to test it?

nakranivaibhav:
alright thanks for pointing it out. I will try that

nakranivaibhav:
@awni What might be a good way to test the code changes.
I have built using the -e flag but the c code that I have changed does not reflect in the library where I have installed for testing.
Do I have to rebuild the entire library everytime I change the c or cpp code.
Or is there a better way.

P.S: Sorry to bug you so much. This is the first time I am working with binded python code.

awni:
Yes, you have to rebuild the C++ code if you change it :). Cmake usually does a good job of only building parts which are touched so it shouldn't take that long.

noahfarr:
I can pick this up

awni:
Thanks!! Assigned to you!

noahfarr:
Just to be clear on how the behavior should be:
If step is inf it should return an array with the start value. 
But what about when start or stop are inf? 
What about when step is -inf? 
I feel like it would be most intuitive to throw an error in any case since they're all basically undefined behaviors.

awni:
Could you check what NumPy does. Typically we just follow what they do for these random edge cases unless there is a reason not to (e.g. it makes it more complex or it's a lot more intuitive to do something different).

[...]

PR #686, opened 2024-02-14
Fix arange with inf step

noahfarr:
## Proposed changes

Arange currently returns an empty list when step is inf, but it should be a list that only contains the start value.
#530 

## Checklist

### DIFF
--- a/mlx/ops.cpp
+++ b/mlx/ops.cpp
@@ -83,6 +83,11 @@ array arange(
   if (std::isnan(start) || std::isnan(step) || std::isnan(stop)) {
     throw std::invalid_argument("[arange] Cannot compute length.");
   }
+  if (std::isinf(start) || std::isinf(stop)) {
+    throw std::invalid_argument("[arange] Cannot compute length.");
+  }
+  step = std::clamp(
+      step, static_cast<double>(INT_MIN), static_cast<double>(INT_MAX));
   double real_size = std::ceil((stop - start) / step);
   if (std::isnan(real_size)) {
     throw std::invalid_argument("[arange] Cannot compute length.");
--- a/python/tests/test_ops.py
+++ b/python/tests/test_ops.py
@@ -1044,6 +1044,10 @@ class TestOps(mlx_tests.MLXTestCase):
             a = mx.arange(0, float("inf"), float("inf"))
         with self.assertRaises(ValueError):
             a = mx.arange(float("inf"), 1, float("inf"))
+        with self.assertRaises(ValueError):
+            a = mx.arange(float("inf"), 1, 5)
+        with self.assertRaises(ValueError):
+            a = mx.arange(float("-inf"), 1, 5)
 
         a = mx.arange(5)
         expected = [0, 1, 2, 3, 4]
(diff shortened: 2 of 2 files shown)


awni:
@noahfarr looks like you are missing a header for the linux build. Can you fix it? (clang tends to include a lot more stuff by default than gcc)

awni:
@noahfarr can you check the failed test?

mlx/ops.cpp
```diff
@@ -83,6 +84,11 @@ array arange(
   if (std::isnan(start) || std::isnan(step) || std::isnan(stop)) {
     throw std::invalid_argument("[arange] Cannot compute length.");
   }
+  if (std::isinf(start) || std::isinf(stop)) {
+    throw std::invalid_argument("[arange] Cannot compute length.");
+  }
+  step = std::clamp(
+      step, static_cast<double>(INT_MIN), static_cast<double>(INT_MAX));
```
awni, line 91: What is the purpose of this clamp?

awni (changes requested):
See inline comment.

noahfarr:
@awni The logic got a little complicated because of all the edge cases. Im still thinking about a better solution, but I thought I get some input from you if you have any idea

noahfarr:
@awni you can review again if you have time

mlx/ops.cpp
```diff
@@ -83,10 +83,24 @@ array arange(
...
+  if (std::isinf(step) &&
+      (step > 0 && start < stop || step < 0 && start > stop)) {
+    return array({start}, dtype);
+  }
+
   double real_size = std::ceil((stop - start) / step);
-  if (std::isnan(real_size)) {
+
+  if (real_size < INT_MIN || real_size > INT_MAX) {
     throw std::invalid_argument("[arange] Cannot compute length.");
```
awni, line 101:
```suggestion
    throw std::invalid_argument("[arange] Maximum size exceeded.");
```

mlx/ops.cpp
```diff
@@ -83,10 +83,24 @@ array arange(
...
   }
```
awni, line 102:
```suggestion
  if (real_size > INT_MAX) {
    throw std::invalid_argument("[arange] Maximum size exceeded.");
  }
```

Shouldn't we let the `std::max` below handle the negative cases?

mlx/ops.cpp
```diff
@@ -83,10 +83,24 @@ array arange(
...
+  if (std::isinf(start) || std::isinf(stop)) {
+    throw std::invalid_argument("[arange] Cannot compute length.");
+  }
+
+  // Check if start and stop specify a valid range because if not, we have to
+  // return an empty array
+  if (std::isinf(step) &&
+      (step > 0 && start < stop || step < 0 && start > stop)) {
+    return array({start}, dtype);
+  }
```
awni, line 96: Another option that might simplify this is to do something like:
```suggestion
  if (std::isinf(step)) {
    step = (step < 0) * std::numeric_limits<double>::max();
  }
```

I'm not 100% certain, but I think the rest of the logic would flow unchanged after that.

awni (commented):
Hi @noahfarr this looks great. I left a couple comments. Please address if it makes sense then we can merge!

Thanks for the contribution!

mlx/ops.cpp
noahfarr: Yes, you're correct.

mlx/ops.cpp
noahfarr: But wouldn't this be 0 when step is +inf?

awni: Right sorry! That is bug it should be `sign(step)` not `step<0`.  Or `(step < 0 ? -1 : 1)`

awni:
Also can you check the failed linux build, looks like it needs an include.

mlx/ops.cpp
noahfarr: This produces NaNs for some reason for the input start = 0, stop = 10, step = inf:
Currently I do this:
if (std::isinf(step)) {
    step = step > 0 ? std::numeric_limits<double>::max()
                    : std::numeric_limits<double>::min();
  }
  
 and then real_size is 1 so it should output an array with only the start but I guess due to some handling in the Arange primitive  it can't handle step being the numerical_limit because when I put 10000000 and -10000000 it works correctly.

awni: Ok no worries, I think it's fine the way you have it. Thanks for testing that!

vovw:
adding should fix this and make it ready to merge
```#include <climits>```

awni (approved):
Thanks!

PR #687, opened 2024-02-14
Shapeless compilation for some graphs

awni:
## Proposed changes

- Introduces shapeless compilation into the API
- Allows some graphs to be compiled with `shapeless=True`
- Some results using this for LM generation in https://github.com/ml-explore/mlx-examples/pull/451
- PR also adds constants to compiled functions.

### Example:

```python
@partial(mx.compile, shapeless=True)
def gelu(x):
  ...
```

Should work on most graphs which have point-wise ops with reductions as long as the broadcasts are fused into the `Compiled` primitive.

### DIFF
--- a/mlx/compile.cpp
+++ b/mlx/compile.cpp
@@ -63,6 +63,13 @@ bool is_fusable(const Primitive& p) {
   return is_unary(p) || is_binary(p) || is_broadcast(p) || is_noop(p);
 }
 
+bool allows_shapeless(const Primitive& p) {
+  return typeid(p) == typeid(Compiled) || is_unary(p) || is_binary(p) ||
+      is_noop(p) || is_reduction(p) || typeid(p) == typeid(Softmax) ||
+      typeid(p) == typeid(Sort) || typeid(p) == typeid(ArgSort) ||
+      typeid(p) == typeid(ArgPartition) || typeid(p) == typeid(Partition);
+}
+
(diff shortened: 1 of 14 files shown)


mlx/compile.cpp
```diff
@@ -13,7 +13,7 @@
 
 namespace mlx::core {
 
-constexpr int max_compile_depth = 10;
+constexpr int max_compile_depth = 11;
```
awni, line 16: Need 11 to fuse all the broadcasts out of layer norm

awni:
Some benchmarks on GELU and LayerNorm. The non shapeless path is virtually unaffected so I'm just reporting benchmarks on shapeless compile vs regular compile when the input frequently varies shape:

Time in milliseconds

function | No compile | Compile Shaped | Compile Shapeless
-------| ------ | ----- |------
gelu | 2.53732 | 0.77570 | 0.63014 
layernorm | 9.12316 |  3.94165 | 2.79180
```

awni:
@angeloskath can you check the basic design here?

If you are cool with it, I can work on fleshing out additional shapeless defaults. I think for this PR I'd like to default shapeless compile functions which do not take python constants (e.g. and `axis` parameter). Then in a follow up deal with constants and then default compile those functions as well.

angeloskath (approved):
🚀

It looks great!

Issue #668, opened 2024-02-11
[Feature request]  Add support atleast_1d,2d and 3d


abdussamettrkr:
Add support atleast1d,2d and 3d primitives

awni:
I didn't know about those but I do see they are in NumPy, are they used much?  They are pretty easy to implement using existing ops:

```
def atleast_1d(a):
  return a[None] if a.ndim == 0 else a
```

abdussamettrkr:
I am using these functions in my projects, but I suppose they're not as crucial. As you mentioned, it's easy to implement. If anyone thinks it's necessary, I can work on it.

hinriksnaer:
I added a [push request](https://github.com/ml-explore/mlx/pull/689) for the feature request.

hinriksnaer:
I think it's safe to close this one.

PR #689, opened 2024-02-14
Added support for atleast_1d, atleast_2d, atleast_3d

hinriksnaer:
## Proposed changes

Feature addition for the following [issue](https://github.com/ml-explore/mlx/issues/668). The code is based on the numpy [implementation](https://github.com/numpy/numpy/blob/v1.26.0/numpy/core/shape_base.py#L81-L132). PyTorch provides [similar functionality](https://github.com/pytorch/pytorch/blob/3470ab42bb754a973e27f552faf0f800edea1d14/aten/src/ATen/native/TensorTransformations.cpp#L170) but is limited to only Tensors and implements it in the C++ backend. Perhaps this is a good place to start?

Any feedback regarding the documentation, tests and implementation would be much appreciated.

## Checklist

### DIFF
--- a/mlx/ops.cpp
+++ b/mlx/ops.cpp
@@ -622,6 +622,13 @@ std::vector<array> split(
 
 std::vector<array>
 split(const array& a, int num_splits, int axis, StreamOrDevice s /* = {} */) {
+  auto ax = axis < 0 ? axis + a.ndim() : axis;
+  if (ax < 0 || ax >= a.ndim()) {
+    std::ostringstream msg;
+    msg << "Invalid axis " << axis << " passed to split"
+        << " for array with shape " << a.shape() << ".";
+    throw std::invalid_argument(msg.str());
+  }
   auto q_and_r = std::ldiv(a.shape(axis), num_splits);
   if (q_and_r.rem) {
     std::ostringstream msg;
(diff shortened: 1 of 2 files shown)


docs/src/python/atleast_utils.rst
```diff
@@ -0,0 +1,15 @@
+.. _utils:
+
+At least Utils
```
awni, line 3: Don't need a separate page in the docs for this, I would just put them in the `ops.rst` list.

python/mlx/utils.py
```diff
@@ -143,3 +145,103 @@ def tree_unflatten(tree):
         return l
     else:
         return {k: tree_unflatten(v) for k, v in children.items()}
+
+
+def atleast_1d(*values):
```
awni, line 150: These should be implemented in C++ not Python and then bound to Python via Pybind (the same as all the rest of our ops).

awni (changes requested):
These should follow the implementation of all of our other ops (C++ and then bound to Python). Are you up for doing the C++ implementation?

hinriksnaer:
Thanks for the feedback! Yes, I'll get started on implementing the functionality using C++. Just to be sure, I presume that you want the functionality to be added [here](https://github.com/ml-explore/mlx/blob/main/mlx/ops.cpp), correct? 

Please let me know if there is anything else you would want me to know before getting started.

awni:
That's exactly right!

PR #694, opened 2024-02-15
Added support for atleast_1d, atleast_2d, atleast_3d

hinriksnaer:
## Proposed changes

Second go at the atleast functionality based on the feedback [here](https://github.com/ml-explore/mlx/pull/689). It is now implemented on the C++ side. This feature is based on the following [request](https://github.com/ml-explore/mlx/issues/668). 

The functionality is implemented as both a core and array function. I am still trying to get more familiar with the stream implementation along with when a custom primitive is needed. I am guessing that custom primitives are not necessary since the functionality includes no additional arithmetic and only utilizes existing transformations?

I am also not sure if the documentation will be automatically updated or if I should make some modifications on my own. 

Finally, the current functionality only works on a single mlx.core.array, if the feedback is good and if people are interested, I can extend it to allow for a list of mlx.core.array. 

Since I am still relatively new to the framework, especially the C++ side of things, please let me know if there is something I missed. This feature has been a great introduction to the framework.

## Checklist


abdussamettrkr:
I suppose your current implementation can only take an array, but I think it should be able to handle a list of arrays, similar to Jax, NumPy, and Torch.

hinriksnaer:
I agree, that would be a good next step. I decided to start small to get some initial feedback for the implementation. I am not sure if they will want to hold the merge until that functionality is available. In either case, I will start working on it once I have wrapped up any necessary changes that may arise from the initial feedback.

python/src/array.cpp
```diff
@@ -1506,5 +1506,23 @@ void init_array(py::module_& m) {
...
+          [](const array& a, StreamOrDevice s) { return atleast_2d(a, s); },
+          py::kw_only(),
+          "stream"_a = none,
+          "See :func:`atleast_2d`.")
+      .def(
+          "atleast_3d",
+          [](const array& a, StreamOrDevice s) { return atleast_3d(a, s); },
+          py::kw_only(),
+          "stream"_a = none,
+          "See :func:`atleast_3d`.");
```
awni, line 1527: Numpy does not have these methods so I don't think we need to add them either..

python/src/ops.cpp
```diff
@@ -3636,4 +3636,64 @@ void init_ops(py::module_& m) {
...
+      "atleast_3d",
+      &atleast_3d,
+      "a"_a,
+      py::pos_only(),
+      py::kw_only(),
+      "stream"_a = none,
+      R"pbdoc(
+        atleast_3d(a: array, stream: Union[None, Stream, Device] = None) -> array
+
+        Convert array to contain at least three dimensions.
```
awni, line 3689:
```suggestion
        Convert array to have at least three dimensions.
```

awni (commented):
I think this is very nicely done! Just some very minor comments, can you take a look? Otherwise LGTM. We can merge it as is or if you prefer to extend the API to support multiple arrays in this PR that is fine too!

awni:
> I am also not sure if the documentation will be automatically updated or if I should make some modifications on my own.

The only change is to add the new ops to the [list of ops](https://github.com/ml-explore/mlx/blob/main/docs/src/python/ops.rst) in the docs.

hinriksnaer:
Thanks for the great feedback @awni as usual, I believe that I have made all of the necessary changes for the current functionality. It would be great if you could merge the current changes, and I will include the array support as a separate push request.

awni (commented):
Awesome thanks! Just one more nit for the docs then we can merge!

docs/src/python/ops.rst
```diff
@@ -125,3 +125,6 @@ Operations
    where
    zeros
    zeros_like
+   atleast_1d
+   atleast_2d
+   atleast_3d
```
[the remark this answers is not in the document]
hinriksnaer: Fixed

angeloskath (approved):

Issue #566, opened 2024-01-27
[BUG] malloc error "Unable to allocate" 8446744063349813240 bytes


ivanfioravanti:
I get malloc error "Unable to allocate" 8446744063349813240 bytes when passing from numpy array to mx.array with intermediate range of items. smaller works, larger works, mid fails

Include code snippet
```python
import numpy as np
import mlx.core as mx

test_ok_1 = mx.array(np.arange(1, 20 * 10**8))
test_ok_2 = mx.array(np.arange(1, 50 * 10**8))
test_ko_3 = mx.array(np.arange(1, 30 * 10**8))

```

**Desktop (please complete the following information):**
 - OS Version: MacOS 14.3
 - Version 0.11
 - M3 Max 40GPU 128GB

awni:
It looks like the index shape is overflowing `int32`. I would use 2D or 3D if you can for now. The fix (if we do it) for this is a bit involved..

ivanfioravanti:
Thanks @awni it was just to raise the flag.

awni:
I will add a useful error message at least until we fix this

jmousseau:
Taking a crack at improving the error message ([early WIP](https://github.com/ml-explore/mlx/compare/main...jmousseau:mlx:check-shape-wip)).

PR #704, opened 2024-02-17
Ensure shape dimensions are within supported integer range (#566)

jmousseau:
## Proposed changes

In #566, the code below triggered an "unable to allocate" error because the shape's dimension overflowed `INT32_MAX`.

```python
mx.array(np.arange(1, 30 * 10**8))
```

Now a more specific error is thrown.

```python
>>> mx.array(np.arange(-1, 2147483647))
ValueError: Shape dimension falls outside supported `int` range.

>>> mx.array([0] * 2147483648)
ValueError: Shape dimension falls outside supported `int` range.
```

Note its still possible to produce shapes that trigger an allocation failure.

## Checklist

### DIFF
--- a/mlx/utils.h
+++ b/mlx/utils.h
@@ -68,6 +68,22 @@ std::vector<int> broadcast_shapes(
 
 bool is_same_shape(const std::vector<array>& arrays);
 
+/** Returns the shape dimension if it's within allowed range. */
+template <typename T>
+int check_shape_dim(const T dim) {
+  constexpr bool is_signed = std::numeric_limits<T>::is_signed;
+  using U = std::conditional_t<is_signed, ssize_t, size_t>;
+  constexpr U min = static_cast<U>(std::numeric_limits<int>::min());
+  constexpr U max = static_cast<U>(std::numeric_limits<int>::max());
+
+  if ((is_signed && dim < min) || dim > max) {
+    throw std::invalid_argument(
+        "Shape dimension falls outside supported `int` range.");
+  }
+
+  return static_cast<int>(dim);
+}
+
 /**
  * Returns the axis normalized to be in the range [0, ndim).
  * Based on numpy's normalize_axis_index. See
--- a/python/src/array.cpp
+++ b/python/src/array.cpp
@@ -208,7 +208,6 @@ void get_shape(T list, std::vector<int>& shape) {
       for (int i = 0; i < arr.ndim(); i++) {
         shape.push_back(check_shape_dim(arr.shape(i)));
       }
-      shape.insert(shape.end(), arr.shape().begin(), arr.shape().end());
       return;
     }
   }
--- a/python/src/array.cpp
+++ b/python/src/array.cpp
@@ -196,7 +196,7 @@ PyScalarT validate_shape(
 
 template <typename T>
 void get_shape(T list, std::vector<int>& shape) {
-  shape.push_back(nb::len(list));
+  shape.push_back(check_shape_dim(nb::len(list)));
   if (shape.back() > 0) {
     auto l = list.begin();
     if (nb::isinstance<nb::list>(*l)) {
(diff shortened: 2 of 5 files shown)


mlx/array.h
```diff
@@ -93,6 +95,21 @@ class array {
...
+    constexpr U min = static_cast<U>(std::numeric_limits<int>::min());
+    constexpr U max = static_cast<U>(std::numeric_limits<int>::max());
+
+    if ((is_signed && dim < min) || dim > max) {
+      throw std::invalid_argument(
+          "Shape dimension falls outside supported `int` range.");
+    }
+
+    return static_cast<int>(dim);
+  }
```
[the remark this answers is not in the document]
awni: Also I don't think we need this to be tepmlated (?). I would just have it take a `size_t` and simplify things a bit.

tests/array_tests.cpp
```diff
@@ -69,6 +69,25 @@ TEST_CASE("test array basics") {
...
+  CHECK_EQ(array::check_shape_dim(static_cast<size_t>(dim_max)), dim_max);
+  CHECK_THROWS_AS(
+      array::check_shape_dim(static_cast<ssize_t>(dim_min) - 1),
+      std::invalid_argument);
+  CHECK_THROWS_AS(
+      array::check_shape_dim(static_cast<ssize_t>(dim_max) + 1),
+      std::invalid_argument);
+  CHECK_THROWS_AS(
+      array::check_shape_dim(static_cast<size_t>(dim_max) + 1),
+      std::invalid_argument);
```
awni, line 89: Maybe make this a python test and more integrated, something that tests e.g.:

```
>>> a = np.empty(2**32, np.bool_)
>>> mx.array(a).shape
(0,)
```

awni: That's only 4GB so it's not too bad I think..

awni (changes requested):
Thanks for the contribution!! Left a couple comments, please address and then let me know when it's ready.

awni:
@jmousseau are you coming back to this?

jmousseau:
@awni Yes, I should have time to address comments this weekend. Apologies for the slowness/lack of response on my end.

mlx/array.h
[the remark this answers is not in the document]
jmousseau: Templated because `shape` returns `ssize_t` whereas `len` returns `size_t`. The reason I didn't place in `python/sr/array.cpp` was so that I could cover additional edge cases with C++ test cases.

tests/array_tests.cpp
jmousseau: I've been burned by large allocations before, so excluded such tests. If we move `check_shape_dim` to `python/src/array.cpp`, will try adding python tests.

mlx/array.h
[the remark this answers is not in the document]
awni: Ok fair enough. I prefer it not be in `array.h`. You could put it in `mlx/utils.h`.

As for the template, I think you mean `shape` return `int`? It will be an always positive int, so you could cast it, but I don't feel too strongly about it.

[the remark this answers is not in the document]
jmousseau: Moved to `mlx/utils.h` and kept template as `py::shape` returns `ssize_t` and `py::len` returns `size_t`.

tests/array_tests.cpp
jmousseau: Added python test case.

mlx/utils.cpp
```diff
@@ -1,6 +1,8 @@
 // Copyright © 2023 Apple Inc.
 
+#include <limits>
 #include <sstream>
+#include <type_traits>
```
awni, line 5: Do we need these added includes here?

jmousseau: You're correct, these are unnecessary and have been removed.

awni (approved):
Looks great, thanks! I will merge when tests clear.

PR #705, opened 2024-02-17
Faster softmax VJP

abdussamettrkr:
## Proposed changes

I have been experimenting with the Softmax VJP implementation, and the new version is 25% faster than the old one. I have included a simple gradient benchmark. This testing was conducted on my M2 Pro PC.

Old Implementation: 1.843306064605713
New Implementation: 1.4345970153808594

## Checklist

### DIFF
--- a/benchmarks/python/comparative/bench_mlx.py
+++ b/benchmarks/python/comparative/bench_mlx.py
@@ -161,6 +161,14 @@ def softmax_fused(axis, x):
     mx.eval(ys)
 
 
+def softmax_fused_grad(axis, x):
+    ys = []
+    for i in range(100):
+        y = mx.grad(lambda x, axis: mx.sum(mx.softmax(x, axis=axis)))(x, axis)
+        ys.append(y)
+    mx.eval(ys)
+
+
 def relu(x):
     y = x
     for i in range(100):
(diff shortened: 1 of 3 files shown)


angeloskath:
This is pretty cool! Unfortunately, I think the speedup is an artifact of the benchmark. When I first saw the change I thought you were somehow avoiding the `sv` multiplication but I don't think that this is possible. So the number of operations are exactly the same.

- `s * cotangents[0]` which means `N` multiplications
- `sv` reduction unchanged
- `multiply` or `subtract` with broadcasting on `sv` which means `N` multiplications or additions
- `subtract` or `multiply` which means `N` additions or multiplications

The reason you see a speedup is because when using the sum to take the gradient the cotangents is a matrix created with `ones()` which is basically a broadcasted scalar. This means that the subtraction is actually less than `N` additions (about half usually).

If instead you multiply with a random array before summing then the cotangent would be that random array and the speedup goes away. Btw I would love to be wrong here so if you have specific sizes that you tested where the speedup remains after changing the benchmark that would be amazing!

awni:
@abdussamettrkr I'm closing this. Please let us know if we missed anything here.

abdussamettrkr:
Sorry for late respond. I've looked into several cases. There's some improvement in Numpy, though I'm not sure why. However, everything seems to be working as expected in MLX. Thank you

PR #708, opened 2024-02-19
Refactor CPU compile preamble

awni:
## Proposed changes

Refactors CPU preamble to reduce code duplication.


angeloskath (approved):
I don't know why, maybe I am old but this is so much nicer than the lambdas :-) . I think they started simple but then they were becoming more and more complicated. Even if it weren't for compile, I think it's awesome.

PR #710, opened 2024-02-19
added atleast *args input support

hinriksnaer:
## Proposed changes

Following up with [atleast feature request](https://github.com/ml-explore/mlx/issues/668)

Swapped out `array` to `*args` for atleast_1d, atleast_2d, atleast_3d. 

The function now returns either an `array` or `list(array)` depending on the amount of inputs, similar to the Numpy and PyTorch implementation of the same function. The corresponding tests have also been updated.

The current implementation simply iterates over the argument list and calls the original `atleast_nd` c++ function. I struggled a bit figuring out how to get Pybind11 to output one of two different datatypes but this seems to do the trick. I also considered making a separate `std::vector<array>` output variation of the `atleast_nd` functions and have the Pybind11 simply call the correct variation depending on the input length. I wasn't able to find a previously implemented function to use as a reference so I settled on simply iterating over the arguments in the pybind11 definition using a lambda function. 

Please let me know if this breaks any conventions within the codebase or if an alternative approach is more desirable. 

## Checklist


awni (changes requested):
Thanks for the follow up!

In the interest of keeping our C++ and Python APIs in sync, could you make this change on the C++ side and then dispatch to the appropriate function from Python?

Concretely something like:
- Have overloads in C++ for single `array` or `std::vector<array>`
- Have the second one call the first in a loop
- Dispatch in Python to the appropriate function

hinriksnaer:
Sure thing! I will look right into that. In the interest of clarity, do you have anything in mind I could use for reference?

awni:
I wasn't referring to anything specific above. If you run into issues implementing just post here and I'm happy to help.

mlx/ops.cpp
```diff
@@ -3388,6 +3388,17 @@ array atleast_1d(const array& a, StreamOrDevice s /* = {} */) {
   return a;
 }
 
+std::vector<array> atleast_1d(
+    const std::vector<array>& arrays,
+    StreamOrDevice s) {
```
awni, line 3393:
```suggestion
    StreamOrDevice s /* = {} */) {
```

mlx/ops.cpp
```diff
@@ -3399,6 +3410,17 @@ array atleast_2d(const array& a, StreamOrDevice s /* = {} */) {
   }
 }
 
+std::vector<array> atleast_2d(
+    const std::vector<array>& arrays,
+    StreamOrDevice s) {
```
awni, line 3415:
```suggestion
    StreamOrDevice s /* = {} */) {
```

mlx/ops.cpp
```diff
@@ -3411,4 +3433,16 @@ array atleast_3d(const array& a, StreamOrDevice s /* = {} */) {
       return a;
   }
 }
+
+std::vector<array> atleast_3d(
+    const std::vector<array>& arrays,
+    StreamOrDevice s) {
```
awni, line 3439:
```suggestion
    StreamOrDevice s /* = {} */) {
```

python/src/ops.cpp
```diff
@@ -3638,62 +3638,96 @@ void init_ops(py::module_& m) {
...
-        atleast_1d(a: array, stream: Union[None, Stream, Device] = None) -> array
+        atleast_1d(*args: array, stream: Union[None, Stream, Device] = None) -> Union[array, List[array]]
 
-        Convert array to have at least one dimension.
+        Convert all arrays to have at least one dimension.
 
-        args:
-            a (array): Input array
+        Args:
+            args (array or list(array)): Input array or list of input arrays.
```
awni, line 3664:
```suggestion
            *args: Input arrays.
```

Based on https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html

awni: Also would you mind changing the name from `args` to `arys` for consistency with NumPy?

python/src/ops.cpp
```diff
@@ -3638,62 +3638,96 @@ void init_ops(py::module_& m) {
...
   m.def(
       "atleast_1d",
-      &atleast_1d,
-      "a"_a,
+      [](const py::args& args,
+         StreamOrDevice s) -> std::variant<py::object, py::list> {
+        if (args.size() == 1) {
+          py::object result;
+          result = py::cast(atleast_1d(args[0].cast<array>(), s));
+          return result;
```
awni, line 3646:
```suggestion
          return py::cast(atleast_1d(args[0].cast<array>(), s));
```

Does that work?

python/src/ops.cpp
```diff
@@ -3638,62 +3638,96 @@ void init_ops(py::module_& m) {
         )pbdoc");
   m.def(
       "atleast_1d",
-      &atleast_1d,
-      "a"_a,
+      [](const py::args& args,
+         StreamOrDevice s) -> std::variant<py::object, py::list> {
```
awni, line 3642: Do you need to declare the `variant` output type here? I might just have it all be `py::object`?

python/src/ops.cpp
```diff
@@ -3638,62 +3638,96 @@ void init_ops(py::module_& m) {
...
+          py::object result;
+          result = py::cast(atleast_1d(args[0].cast<array>(), s));
+          return result;
+        }
+
+        std::vector<array> arrays = args.cast<std::vector<array>>();
+        std::vector<array> result = atleast_1d(arrays, s);
+        py::list py_result = py::cast(result);
+
+        return py_result;
```
awni, line 3653:
```suggestion
       return py::cast(result);
```

Does that work?

awni (changes requested):
I left a few comments on `atleast_1d` which apply to the others as well. Could you check and address the ones that make sense?

python/src/ops.cpp
hinriksnaer: returning only py::object is sufficient, thanks for that. I thought a variant was necessary when returning a python list.

hinriksnaer:
Here is a set of changes based on your feedback @awni. 

- [x] added /* = {} */
- [x] refactored pybind11 defenition
- [x] fixed docstring

The only thing I am uncertain about is how to rename the resulting python function to contain *arys instead of *args using pybind11.

awni:
> The only thing I am uncertain about is how to rename the resulting python function to contain *arys instead of *args using pybind11.

I think you did that in the docstring + function signature already? That's the only change need as far as I can tell.

python/src/ops.cpp
```diff
@@ -3638,62 +3638,72 @@ void init_ops(py::module_& m) {
...
       "atleast_3d",
-      &atleast_3d,
-      "a"_a,
+      [](const py::args& arys, StreamOrDevice s) -> py::object {
+        if (arys.size() == 1) {
+          return py::cast(atleast_3d(arys[0].cast<array>(), s));
+        }
+        return py::cast(atleast_3d(arys.cast<std::vector<array>>(), s));
+      },
       py::pos_only(),
```
awni, line 3694: Nit: the pos_only isn't needed here.

python/src/ops.cpp
```diff
@@ -3638,62 +3638,72 @@ void init_ops(py::module_& m) {
...
       "atleast_2d",
-      &atleast_2d,
-      "a"_a,
+      [](const py::args& arys, StreamOrDevice s) -> py::object {
+        if (arys.size() == 1) {
+          return py::cast(atleast_2d(arys[0].cast<array>(), s));
+        }
+        return py::cast(atleast_2d(arys.cast<std::vector<array>>(), s));
+      },
       py::pos_only(),
```
awni, line 3670: Nit: the pos_only isn't needed here.

python/src/ops.cpp
```diff
@@ -3638,62 +3638,72 @@ void init_ops(py::module_& m) {
...
       "atleast_1d",
-      &atleast_1d,
-      "a"_a,
+      [](const py::args& arys, StreamOrDevice s) -> py::object {
+        if (arys.size() == 1) {
+          return py::cast(atleast_1d(arys[0].cast<array>(), s));
+        }
+        return py::cast(atleast_1d(arys.cast<std::vector<array>>(), s));
+      },
       py::pos_only(),
```
awni, line 3647: Nit: the pos_only isn't needed here.

hinriksnaer:
Pushed the changes [here](https://github.com/hinriksnaer/mlx/tree/array_atleast).

Hopefully they made it to the push request.

awni (approved):
LGTM, thanks!!

PR #711, opened 2024-02-20
Native pooling_1d

dc-dc-dc:
## Proposed changes

Adds native support for `max_pool_1d` `avg_pool_1d`, almost a 4x speedup over current implementation

Speed comparison to torch
```shell
python benchmarks/python/comparative/compare.py --filter max_pool_1d
0.7296902512542818      max_pool_1d --size 32x1024x3 --cpu
0.9970802109584517      max_pool_1d --size 32x1024x3

python benchmarks/python/comparative/compare.py --filter avg_pool_1d
0.4435240507031219      avg_pool_1d --size 32x1024x3 --cpu
0.9984100432800462      avg_pool_1d --size 32x1024x3
```

## Checklist


awni:
Wow, that looks awesome!

awni:
@dc-dc-dc before you go too deep with this PR I wanted to run something by you:

I am a bit worried about having a 1D custom primitive just for pooling:
- Adding a primitive is a lot of work in order to not break our contract of what a primitive should be able to do. E.g. this really should include VMAP, VJP and JVP at the minimum
- It implies that we will also have 2D, 3D specializations as well which will make things even more complex

I'm wondering if it's at all possible to repurpose what you've done into our existing `Reduce` implementation while preserving (most of) the speedup.

In particular our reductions are really slow when reducing over a small axis. If we could make that faster (possibly leveraging the kernel you wrote?) then we would avoid the need for a custom primitive. We could simply use `as_strided` -> `reduce`.

Let me know what you think about that.

dc-dc-dc:
Should’ve clarified that I also intend to add 2D pooling support(cpu is already finished), but wanted to get some initial feedback on what’s already here. I could also look at possible speed ups to the reduce kernel but in terms of pooling there is the case for max where users would also want indicies returned. I originally didn’t include that in this PR as I wanted to figure out VMAP, VJP, and JBP before first

dc-dc-dc:
Quick comparison for 2D pooling op

```shell
# average
MAX=0 PERF=1 python ./pool.py
torch 1.069 s
old mx 17.153 s (stride + reduce)
mx 1.826 s
# max
❯ MAX=1 PERF=1 python ./pool.py
torch 2.389 s
old mx 17.392 s (stride + reduce)
mx 2.318 s
```

awni:
> but in terms of pooling there is the case for max where users would also want indicies returned

I don't fully understand what you mean by that. Can't we use arg reductions for those?

awni:
@dc-dc-dc those speedups are massive. It would be great if you could investigate if they could be translated to the reduction instead of a custom primitive. We should go that route unless there is a reason it can't be done.

dc-dc-dc:
> but in terms of pooling there is the case for max where users would also want indicies returned

In torch if return_indicies is true then a tuple is returned with the indicies of the max element alongside the output. You could get the indicies through other means afterwards but a waste of compute since you can just combine it into the max_pool kernel

awni:
Indeed, though you could say the same about `max` vs `argmax` as the two often go hand in hand. One often needs the indices for the grad. That is another optimization we could consider adding but probably lower priority

dc-dc-dc:
Cool, ill investigate reduce speed and keep this pr open in the meantime. Will probably create a new PR with any changes to reduce vs updating this one since this is going in a different direction.

awni:
Closing this as it's a bit stale and we are opting to work on reductions instead.

Issue #367, opened 2024-01-04
[Feature Request] Array pickle support


dc-dc-dc:
Add pickle support to [array](https://github.com/ml-explore/mlx/blob/main/python/src/array.cpp) through pybind, docs can be viewed [here](https://pybind11.readthedocs.io/en/latest/advanced/classes.html#pickling-support). This is a pre-req for #300

awni:
Thanks for making the issue!

I don't see any reason we shouldn't let an array be pickle-able, so I support adding this.

There may be some thorny issues in terms of managing the state of the array (a lot of it is not currently publicly accessible). If we want to keep it that way, we probably need to do a contiguous copy first.

Also it probably should not be ok to pickle an array used in tracing the graph in a function transform.

awni:
Is anyone actively working on this?

chunyang-wen:
If no one is working on it, I can give it a shot

awni:
Sounds great!! I will assign to you!

TristanBilot:
Anyone working on this?

dc-dc-dc:
@chunyang-wen any progress?

chunyang-wen:
@dc-dc-dc Sorry. I am on holidays. Will work on it.

chunyang-wen:
I will remove myself as people are working on it. Sorry for the delay.

Issue #300, opened 2023-12-27
[Feature Request] Enable nn.Module with Python copy


0x1orz:
To reuse a  module copy,  like:
```python
from copy import deepcopy
import  mlx.nn as mlx_nn
repeat = lambda module, n: [deepcopy(module) for _ in range(n)]

mlx_layer = mlx_nn.Linear(2, 3)
mlx_layers = repeat(mlx_layer, 2)
```
that  raised `TypeError: cannot pickle 'mlx.core.array' object` in `deepcopy(module)`.

ekdnam:
i would like to work on this!

awni:
Great! If you have any questions about it, we can use this issue for discussion or better yet start a draft PR and use that!

ekdnam:
Yes, I will start working on it! Could you point to where should I look into the code to start?

NripeshN:
Hey @awni,
Could you please ensure contributors link the PR to the issue properly before merging(eg closes <issue>). This would automatically close issues.

awni:
Good point! Fixed in #335

awni:
Spoke too soon. This is not actually fixed (se #366 )

PR #702, opened 2024-02-17
Fix `array.tolist` for half types

awni:
## Proposed changes

Closes #701

PR #713, opened 2024-02-20
Implementation of pickle, copy and deepcopy for Python arrays (#300 & #367).

arn4:
## Proposed changes

Implementation of pickle, copy and deepcopy for Python arrays (#300 & #367).

## Checklist


python/tests/test_array.py
```diff
@@ -657,6 +657,58 @@ def test_array_iteration(self):
...
+            mx.int16,
+            mx.int32,
+            mx.int64,
+            mx.uint8,
+            mx.uint16,
+            mx.uint32,
+            mx.uint64,
+            # mx.float16,
+            mx.float32,
+            # mx.bfloat16,
```
awni, line 672: Why not float16 and bfloat16?

python/src/array.cpp
```diff
@@ -774,6 +774,25 @@ void init_array(py::module_& m) {
             return a.shape(0);
           })
       .def("__iter__", [](const array& a) { return ArrayPythonIterator(a); })
+      .def(py::pickle(
+          [](array& a) { // __getstate__
+            return py::make_tuple(
+                dtype_to_array_protocol(a.dtype()), tolist(a));
```
awni, line 780: I'm wondering about the use of `tolist` here. That will be a pretty inefficient way to pickle and un-pickle large arrays..

Did you consider any alternatives?

awni (commented):
Thanks for getting this started! Could you check my comments / questions inline?

python/tests/test_array.py
arn4: My bad, I forgot to uncomment after closing #702 . Fixing soon

python/src/array.cpp
awni: Thanks for the benchmarks that's really interesting.

> For sure I can check how they are doing that in Numpy and take inspiration. 

Very good idea.. 

Another possibility is to convert the array to numpy and pickle that? I think this is what Jax does: https://github.com/google/jax/discussions/8494#discussioncomment-1619664


> PS: I've benchmarked also deepcopy(), and this such fast that I'm worried that is not actually behaving as expected!

This is expected given the way you implemented it as it is actually a shallow copy. But that's actually ok as arrays are "immutable" hence it doesn't really make sense to have a true deep copy.

dc-dc-dc:
Btw numpy uses `__reduce__` instead of `__getstate__` see [here](https://github.com/numpy/numpy/blob/6aab0398d64da41df61487db32f28ca583ec9a48/numpy/_core/src/multiarray/descriptor.c#L2642). They include various attributes from the array including dtype, shape, elementsize and flags alongside the subarray.

arn4:
I've looked into NumPy implementation and how to use `__reduce__`, and I think it is a little too complicated for my level.
If you want to go directly with this option I think I could not help for now.

I am trying to get an implementation that directly uses the Numpy interface provided by `py11bind`. The code I tried for now is:
```c++
.def(py::pickle(
          [](array& a) { // __getstate__
            auto buffer_info = pybind11::buffer_info(
                  a.data<void>(),
                  a.itemsize(),
                  buffer_format(a).value_or("B"),
                  a.ndim(),
                  a.shape(),
                  buffer_strides(a));
            return py::make_tuple(
                dtype_to_array_protocol(a.dtype()), py::array(buffer_info));
          },
          [](py::tuple t) { // __setstate__
            auto dtype = dtype_from_array_protocol(t[0].cast<std::string>());
            auto np_array = t[1].cast<py::array_t<float_t>>();
            return np_array_to_mlx(np_array, dtype);
            }
));
```
If then I try to
```python
x = mx.arange(10)
y = pickle.load(pickle.dump(x))
```
I get `ValueError: Cannot convert numpy array of type int32 to mlx array.`, or the same of any other `dtype` I try.

What I can't understand is, if I replace the casting in `__setstate__`
```
t[1].cast<py::array_t<float_t>>()
```
Then the pickling works for all the types except for `mx.float32`! 

I'm sure I'm missing something in how Numpy and `py11bind` work, but maybe you have some insight on how to proceed so I can close the PR.

If you find it easier to solve & close it yourself instead, no worries!

awni:
- I am not sure about the type issues but you do need to do an eval before making the buffer (as in the `def_buffer` function)
- While I'm not sure why the cast helps you, I would definitely remove it ` t[1].cast<py::array_t<float_t>>();`
- Why do you return the dtype? Shouldn't it be in the np array?

You see that if the dtype of the numpy array is not recognized:
`ValueError: Cannot convert numpy array of type int32 to mlx array.`

That means somehow the numpy array has a messed up type. You can check it in python:

`a.__getstate__[1].dtype`

I would inspect the `a.__getstate__` and make sure it looks reasonable.

arn4:
> I am not sure about the type issues but you do need to do an eval before making the buffer

Yes right. I've added it.

> Why do you return the dtype? Shouldn't it be in the np array?

I was doing it because it seemed to me that some `dtype`s are not avalable in `numpy`, such as `bfloat16`, so I wanted to ensure that is transferred correctly. Probably, I'm wrong.


I tried to test with this code, where I've also added this line of code in the implementation of `np_array_to_mlx`
```c++
std::cout << " np_array_to_mlx type: " << type << std::endl;
```
Python script:
```python
x = mx.arange(3, step=1.5 , dtype=mx.float32)
xnp = np.arange(3, step=1.5, dtype=np.float32)
xb = mx.array(x.__getstate__()[1])
x3 = mx.array(xnp)

print('mx array: ', x.__getstate__()[1].dtype, x)
print('np array: ', xnp.dtype, xnp)
print('mx back array: ', xb.dtype, xb)
print('no pickle array: ', x3.dtype, x3)

pickle.loads(pickle.dumps(x)) # fails
```
Output:
```
 np_array_to_mlx type: float32
 np_array_to_mlx type: float32
mx array:  float32 array([0, 1.5], dtype=float32)
np array:  float32 [0.  1.5]
mx back array:  mlx.core.float32 array([0, 1.5], dtype=float32)
no pickle array:  mlx.core.float32 array([0, 1.5], dtype=float32)
 np_array_to_mlx type: float32
ValueError: Cannot convert numpy array of type float32 to mlx array.                                                                  
```

I'm honestly a little bit lost know. I don't see any reason why it should not work.  `np_array_to_mlx` recivies always rhe same `dtype`, but apparently it fails when the `np.array` comes from `__get_state__`.

arn4:
> you do need to do an eval before making the buffer

Yes, right. I've added it.

> Why do you return the dtype? Shouldn't it be in the np array?
It seemed to me that not all `dtype`s are available in numpy, like `bfloat16`.  I wanted to ensure that it is correctly converted back. See this code:
```python
x = mx.arange(3, step=1.5 , dtype=mx.bfloat16)
y = np.array(x)
```
It returns `RuntimeError: Item size 2 for PEP 3118 buffer format string B does not match the dtype B item size 1.`

I've added a line to the implementation of `np_array_to_mlx type` to check what type is passed every time: `std::cout << " np_array_to_mlx type: " << type << std::endl;`.
Then, I've run this script
```python
x = mx.arange(3, step=1.5 , dtype=mx.float32)
xnp = np.arange(3, step=1.5, dtype=np.float32)
xb = mx.array(x.__getstate__()[1])
x3 = mx.array(xnp)

print('mx array: ', x.__getstate__()[1].dtype, x)
print('np array: ', xnp.dtype, xnp)
print('mx back array: ', xb.dtype, xb)
print('no pickle array: ', x3.dtype, x3)

pickle.loads(pickle.dumps(x)) # fails
```
Output:
```
 np_array_to_mlx type: float32
 np_array_to_mlx type: float32
mx array:  float32 array([0, 1.5], dtype=float32)
np array:  float32 [0.  1.5]
mx back array:  mlx.core.float32 array([0, 1.5], dtype=float32)
no pickle array:  mlx.core.float32 array([0, 1.5], dtype=float32)
 np_array_to_mlx type: float32
ValueError: Cannot convert numpy array of type float32 to mlx array.
```

I'm a little bit lost now. `np_array_to_mlx type` seems to recivie always the same `dtype`, but it does not work when the `np.array` comes from `__get_state__`

TristanBilot:
Any updates here? 👀

awni:
My understanding is @arn4 is a bit stuck here. I would be happy to help but I won't be able to get to it for a while. If anyone else has cycles to dive in I'm assuming that would be appreciated.

awni:
@arn4 #763 might fix a problem you were having getting this all to work.

arn4:
Yes! I'm waiting that is merged then I give a final try on this. Thanks!

arn4:
Now it is working with NumPy! 

Basically I'm using the plain numpy array as state. This is much faster than the solution with `to_list`, but still 2.5x slower than `numpy` and `jax`. Here are some little tests:
```
mx time: 3.2999801635742188
np time: 1.2117352485656738
jnp time: 1.748607873916626
```

There is just one issue: `bfloat16`. Basically, when the original `dtype` is `bfloat16`, the reconstructed array has different type:
`AssertionError: mlx.core.uint8 != mlx.core.bfloat16 : dtype mismatch expected=mlx.core.bfloat16 got=mlx.core.uint8`

Even when forcing the `dtype` by passing it into the state, the reconstructed array does not work:
```
x = mx.array([1,2,3], dtype=mx.bfloat16)
y = pickle.loads(pickle.dumps(x))
print('mx array: ', x, y)
```
Output:
```
mx array:  array([1, 2, 3], dtype=bfloat16) array([128, 0, 64], dtype=bfloat16)
```

I'm looking for a solution, but I'm also posting here because maybe someone of you knows already where to look

awni:
> but still 2.5x slower than numpy and jax

That is much better! And acceptable. I think it's going to be difficult to avoid a copy entirely which I assume is why Jax and NumPy are so fast (?). We may be able to inspect the incoming buffer to determine if we can avoid the copy, but usually when reading from a file we will have to do a copy so I wouldn't worry about it for now.

You can see in `np_array_to_mlx` there is no support for `bfloat16`. Presumably when you get the buffer out it has already lossed the dtype info:

```
    case bfloat16:
      // not supported by python buffer protocol or numpy.
      // must be null according to
      // https://docs.python.org/3.10/c-api/buffer.html#c.PyBUF_FORMAT
      // which implies 'B'.
      return {};
```

What I would do for now is throw if `bfloat` if you get a bfloat array on the input or output of the pickle. Supporting it might require some big changes which can be done in a follow up.

awni:
I prefer we raise an error for now. Users can always cast if they want and mostly reproduce that behavior. I prefer to remove the implicit up cast because it doubles the I/O in a non transparent manner

arn4:
Okay, good point. I've reverted the commit and tweaked a few details. 

Finally, it should be ready to merge

python/src/array.cpp
```diff
@@ -773,6 +774,24 @@ void init_array(py::module_& m) {
...
+            if (not py::isinstance<py::array>(npa)) {
+              throw std::runtime_error("Invalid state!");
+            }
+            return np_array_to_mlx(npa, std::nullopt);
+          }))
+      .def("__copy__", [](const array& self) { return array(self); })
+      .def(
+          "__deepcopy__",
+          [](const array& self, py::dict) { return array(self); },
+          "memo"_a)
```
awni, line 794: I'm not certain about having a `__deeepcopy__`. Could you say more about why you added it?

It's not a true deep copy so it might lead to unexpected/ambiguous behaviors.

For example:

```python
def fun(x):
  y = deepcopy(x)
  return y

# What should that be? 0 or 1?
dfdx = mx.grad(fun)(mx.array(1.0))
```

python/tests/test_array.py
```diff
@@ -1,9 +1,10 @@
-# Copyright © 2023 Apple Inc.
+# Copyright © 2024 Apple Inc.
```
awni, line 1:
```suggestion
# Copyright © 2023-2024 Apple Inc.
```

python/src/array.cpp
```diff
@@ -773,6 +774,24 @@ void init_array(py::module_& m) {
             return a.shape(0);
           })
       .def("__iter__", [](const array& a) { return ArrayPythonIterator(a); })
+      .def(py::pickle(
+          [](array& a) { // __getstate__
+            if (a.dtype() == bfloat16) {
+              throw std::runtime_error("Pickle not supported for bfloat16.");
```
awni, line 780:
```suggestion
              throw std::runtime_error("[array.__getstate__] Not supported for bfloat16.");
```

awni (commented):
Looks really great!! Thanks for the contribution. Left a couple minor comments.

The only thing I'm not sure about is the addition of `__deepcopy__` (see inline comment). Was there a specific reason you added it?

python/src/array.cpp
arn4: I have added it because it speeds up the code (even if it is so fast that it should ring a bell, see my message https://github.com/ml-explore/mlx/pull/713#discussion_r1496130452), and I think that `copy` and `deepcopy` are the same thing for `mx.array`.

Regarding the behaviour in the example you suggested, here is how `jax` handles it compared to the current implementation
```python
def fun_deepcopy(x):
  y = deepcopy(x)
  return y
def fun_copy(x):
  y = copy(x)
  return y
print('mx copy: ', mx.grad(fun_copy)(mx.array(1.0)))
print('mx deepcopy: ', mx.grad(fun_deepcopy)(mx.array(1.0)))
print('jax copy: ', jgrad(fun_copy)(jnp.array(1.0)))
print('jax deepcopy: ', (fun_deepcopy)(jnp.array(1.0)))
```
Output:
```
mx copy:  array(1, dtype=float32)
mx deepcopy:  array(1, dtype=float32)
jax copy:  1.0
jax deepcopy:  1.0
```
If instead I remove the actual implementation of `__deepcopy__` (meaning that it is delegated to `__getstate__`+`__setstate__`), they do not match anymore
```
mx copy:  array(1, dtype=float32)
mx deepcopy:  array(0, dtype=float32)
jax copy:  1.0
jax deepcopy:  1.0
```

Now we can discuss what should be the right thing to return. Probably the `copy`s should refer to the whole object, computational graph included, so the correct answer is 1 in both the cases.  If one wants to copy just the values, it should use something like `detach()` (PyTorch) or `stop_gradient` (Jax), combined with some copy functions. 

See [this issue](https://github.com/google/jax/issues/2632) and the folliwing pull request to  see how they handled it in Jax.

awni: Thank you for the thoughtful response!! That is exactly what I was looking for!

awni:
@arn4 looks like the formatting is broken, can you run the pre-commit hooks?

arn4:
It is done now

awni (approved):
Nice addition, thanks!!

TristanBilot:
Many thanks @arn4 🥳

PR #721, opened 2024-02-21
Add linear warmup and schedule joining for use with existing schedules

chimezie:
## Proposed changes

Add linear_warmup to schedulers, which takes a scheduler function (cosine_decay, etc.), the number of steps for the warmup, the initial value (defaults to 0), and the final value

## Checklist


python/mlx/optimizers/schedulers.py
```diff
@@ -84,3 +85,33 @@ def scheduler(step):
...
+    :param init: start value (0 by default)
+    :param finish: final value
+    :return: callable takes a step and returns the schedulable
+    """
+
+    def schedule_step_fn(step):
+        if step <= length:
+            return step * ((finish - init) / length) + init
+        else:
+            offset_idx = step.item() - length
```
awni, line 114: Why call `step.item()` here? I think it's important to leave it as an `mx.array` and not eval the graph.

python/mlx/optimizers/schedulers.py
```diff
@@ -84,3 +85,33 @@ def scheduler(step):
...
+    >>> for _ in range(100): optimizer.update({}, {})
+    ...
+    >>> optimizer.learning_rate
+    array(0.1, dtype=float32)
+
+    :param schedule_fn: other schedule function, such as cosine_decay,step_decay, etc.
+    :param length: warmup length
+    :param init: start value (0 by default)
+    :param finish: final value
+    :return: callable takes a step and returns the schedulable
```
awni, line 107: Could you follow the same docstring format as other functions in this package?

awni (changes requested):
Thanks for the addition. I think it could be useful to have a linear warmup since it's quite common.

One thing that might be nice to make more general is the approach to combining schedules. I kind of like the way optax let's you [join schedules](https://optax.readthedocs.io/en/latest/api/optimizer_schedules.html#optax.join_schedules).

I also find it to be a little more intuitive as to what the step parameters mean when you explicitly join them.

python/mlx/optimizers/schedulers.py
chimezie: Yes, and I'll also try for a clean implementation of the schedule joining you mentioned.  Seems very helpful to have.

awni: Awesome, thanks!

python/mlx/optimizers/schedulers.py
chimezie: Ok. I'll remove that and make it:

```python
step - length
```

 I wasn't sure if I could assume using an arithmetic operation on both length and step would work whether they were scalars or mx.arrays.

awni: Yup that will work fine! You can always do ops with one `mx.array` and a Python scalar. Doing the `item()` also works in most cases but it would cause issues with compilation so that's one reason to not have it.

python/mlx/optimizers/schedulers.py
```diff
@@ -84,3 +85,87 @@ def scheduler(step):
...
+        array(1e-5, dtype=float32)
+
+    """
+
+    def __init__(self, schedules: List[Callable], boundaries: List[int]):
+        self.schedules = schedules
+        self.boundaries = boundaries
+
+    def __call__(self, step: int):
+        if step < self.boundaries[0] or not self.boundaries:
```
awni, line 123: We need to rethink this approach. We shouldn't be doing control flow based on the value of `step` which is an `mx.array` because it won't work well with function transformations..

awni (changes requested):
Could you take a look at how [Optax implemented the schedule joiner](https://github.com/google-deepmind/optax/blob/main/optax/schedules/_join.py#L25-L45).

It is both a lot simpler and will work with function transformations since there is no control flow.

Would you mind changing to be more like that and also to a function closure api like the rest of the schedules?

chimezie:
```python
def join_schedules(schedules, boundaries
  def schedule(step: chex.Numeric) -> chex.Numeric:
    output = schedules[0](step)
    for boundary, schedule in zip(boundaries, schedules[1:]):
      output = jnp.where(step < boundary, output, schedule(step - boundary))
    return output
  return schedule
```

I was trying to avoid lazily evaluating the schedules every time. But I have now updated/adapted per @awni

PR #735, opened 2024-02-23
Fast Inference SDPA op

bpkeene:
Implements Metal kernels backing the new compound op:

```python
o = mx.fast.scaled_dot_product_attention(queries, keys, values, scale=scale, mask=mask)
```

to replace current primitives for attention e.g. Mistral-7b in [mlx-examples/llms/mistral/mistral.py](https://github.com/ml-explore/mlx-examples/blob/main/llms/mistral/mistral.py):
```python
scores = (queries * self.scale) @ keys.transpose(0, 1, 3, 2)
if mask is not None:
    scores += mask
scores = mx.softmax(scores.astype(mx.float32), axis=-1).astype(scores.dtype)
output = (scores @ values)
```

### Context
This "fat op" design is inspired by the design of [torch.nn.functional.scaled_dot_product_attention](https://pytorch.org/docs/stable/generated/torch.nn.functional.scaled_dot_product_attention.html). This PR implements the Metal kernel for KV-cached SDPA inference following the lazy softmax principle from [1]. [2] follows the same principles for their CUDA implementation. Future PRs will extend coverage to self-attention (prompt encoding) and improve performance (un-tuned) as well as dispatch heuristics.

[1] https://arxiv.org/abs/2112.05682
[2] https://pytorch.org/blog/flash-decoding/

### Supported
- Supports `mx.float16` and `mx.float32` dtypes
- Supports [head_dim](https://github.com/ml-explore/mlx-examples/blob/main/llms/mistral/mistral.py#L58C44-L58C57)=128 (covers most 7b+ LLMs)
- New Metal kernel added by this PR covers KV-cached forward passes
- Standard prompt encoding still goes through MLX primitives under the hood.
- No backward pass implementation (inference-only kernel)

### Future Work
- Support for `head_dim=64` (Whisper). Other models need discussion (e.g. `head_dim=80` for phi)
- Extension to `query_sequence_length > 1` (non KV-cached inference, i.e. speculative decoding, prompt encoding, ViT, DiT etc.)
- Auto-tuning of `TILE_SIZE` and `N_SIMD_GROUPS`
- Profiling and performance tuning
- Refactor the second kernel (reduction across KV cache tiles) from naive implementation to shave overhead)
- Re-examine loads from K, V to maximize shared memory utilization
    - Esp. for tile size 128 with half, and tile size 64 with float32: can fit all of values in to shared memory
- Fast CPU implementation
- `mx.bfloat16` support

### Quick measurements
- We will share principled measurements on our Apple Silicon Mac test fleet (M1 through M3) before merging the PR
- First measurements to kick off discussions:
  - MacBook Pro M3 Max + macOS 23B2091 & MLX 972d9a3 (Feb 21) & `model = mx.compile(model)`
  - Up to ~2.5x higher tokens/second for Mistral-7b. for large sequence length ~8k tokens
  - Majority of the improvement is attributable to native GQA support (no tiling of KV cache, compresses KV cache by 4x for Mistral)

## Proposed changes
New `fast.scaled_dot_product_attention` op, with C API and Python bindings.

## Checklist


atiorh:
Quick measurements:
mlx_speedup_720

bpkeene:
Happy to move this in to a non-core-primitive namespace as discussed previously, please let me know initial feedback and I'll incorporate in an update to the PR!

bmorphism:
lgtm

awni:
> Happy to move this in to a non-core-primitive namespace as discussed previously, please let me know initial feedback and I'll incorporate in an update to the PR!

@bpkeene I do think this op makes sense to put in the `mlx::core::fast` namespace / `mlx.core.fast` subpackage. E.g. see the [fast rope function](https://github.com/ml-explore/mlx/blob/main/mlx/fast.h).  Wdyt? Should be relatively straightforward to move right?

fblissjr:
This is awesome!

angeloskath:
@bpkeene this looks really good, excited to do a review when I get back to a computer.

Otherwise, I was about to comment the same thing as Awni, the way `mlx::core::fast` is set up you provide an mlx function that will be called when gradients are needed so this can be used in more general situations and it will just be faster when doing inference. Rope is indeed a good example use case.

bpkeene:
absolutely!  I'll take a look at the rope related PR for the fast namespace, I think it should be straightforward

awni:
I also suggest removing `fast_inference` from the name, I think it will be self-evident if it's in `mx.fast`. Some ideas:
- Just `sdpa`
- `scaled_dot_product_attention` like PyTorch
- `multi_head_attention` (maps nicely to the `MultiHeadAttention` layer)

adhulipa:
> Quick measurements: mlx_speedup_720

@atiorh curious if argmax team is considering updating mlx mistral example using the new op?  And perhaps sharing the benchmark/test that generated this comparison?  It would be nice to have the community run on their variety of hardware and try to reproduce the results or its variances

Thinking out loud here, I suppose it's as straightforward as replacing
```python
        scores = (queries * self.scale) @ keys.transpose(0, 1, 3, 2)
        if mask is not None:
            scores += mask
        scores = mx.softmax(scores.astype(mx.float32), axis=-1).astype(scores.dtype)
```
with
```python
        scores = mx.fast.sdpa(q, k, v, self.scale, mask)
```
in mistral.py https://github.com/ml-explore/mlx-examples/blob/main/llms/mistral/mistral.py#L89-L92
?

bpkeene:
> > Quick measurements: mlx_speedup_720
> 
> @atiorh curious if argmax team is considering updating mlx mistral example using the new op? And perhaps sharing the benchmark/test that generated this comparison? It would be nice to have the community run on their variety of hardware and try to reproduce the results or its variances
> 
> Thinking out loud here, I suppose it's as straightforward as replacing
> 
> ```python
>         scores = (queries * self.scale) @ keys.transpose(0, 1, 3, 2)
>         if mask is not None:
>             scores += mask
>         scores = mx.softmax(scores.astype(mx.float32), axis=-1).astype(scores.dtype)
> ```
> 
> with
> 
> ```python
>         scores = mx.fast.sdpa(q, k, v, self.scale, mask) # or mx.core.fast_inference_sdpa or whatever the final op name is
> ```
> 
> in mistral.py https://github.com/ml-explore/mlx-examples/blob/main/llms/mistral/mistral.py#L89-L92 ?

great point!

just about, it's important as well to look a couple lines above in mistral.py and remove the tiling of the KV cache to get the performance benefits of GQA quoted above (the 'repeat()' function)

atiorh:
@adhulipa Preparing https://github.com/argmaxinc/mlx-bench at the moment for this purpose. We have internal CI machines (M1 through M3) for periodic benchmarking but this public benchmarking script should be useful for PRs like this 👍

atiorh:
Ready to use as a bench utility here: https://github.com/argmaxinc/mlx-bench/pull/1

bpkeene:
Converting this to a draft while we're investigating some M1 & M2 specific numerical issues.  M3 Max / 23B2091 has proper tokens generated and appears unaffected.  We'll continue addressing the other issues in tandem with the investigation

python/tests/test_autograd.py
```diff
@@ -3,6 +3,7 @@
 import unittest
 
 import mlx.core as mx
+
```
awni, line 6: Just random nit: any reason you edited all of these python test files?

python/tests/test_fast_inference_sdpa.py
```diff
@@ -0,0 +1,114 @@
+# Created by Brian Keene, 2024
```
awni, line 1: @bpkeene we are putting acknowledgements in the [ACKNOWLEDGEMENTS.md](https://github.com/ml-explore/mlx/blob/main/ACKNOWLEDGMENTS.md) file rather than have author / maintainer lists in individual source code files.

python/tests/test_fast_inference_sdpa.py
```diff
@@ -0,0 +1,114 @@
...
+
+        q_mlx = mx.array(q_npy)
+        k_mlx = mx.array(k_npy)
+        v_mlx = mx.array(v_npy)
+
+        reference = mlx_primitives_sdpa(q_mlx, k_mlx, v_mlx, scale)
+
+        o_mlx = mx.fast_inference_sdpa(q_mlx, k_mlx, v_mlx, scale, None)
+
+        mx.eval()
```
awni, line 74: This is a no-op

python/tests/test_fast_inference_sdpa.py
```diff
@@ -0,0 +1,114 @@
...
+
+        self.assertListEqual(list(reference.shape), list(o_mlx.shape))
+        self.assertTrue(np.allclose(o_mlx, reference, atol=1e-4))
```
awni, line 77: Nit: You can use `mx.allclose` .. no need to cast to `np`

python/tests/test_fast_inference_sdpa.py
```diff
@@ -0,0 +1,114 @@
...
+import numpy as np
+
+import mlx_tests
+
+
+def relerr(truth, other):
+    a = truth.flatten()
+    b = other.flatten()
+    return np.linalg.norm(a - b) / np.linalg.norm(a)
+
```
awni, line 17: Unused

python/tests/test_fast_inference_sdpa.py
```diff
@@ -0,0 +1,114 @@
+# Created by Brian Keene, 2024
+
+import math
+import unittest
+from itertools import permutations
```
awni, line 5: nit: unused import

python/src/ops.cpp
```diff
@@ -2280,6 +2280,38 @@ void init_ops(py::module_& m) {
...
       )pbdoc");
+  m.def(
+      "fast_inference_sdpa",
+      &fast_inference_sdpa,
+      "q"_a,
+      "k"_a,
+      "v"_a,
+      "scale"_a,
+      "mask"_a = none,
+      py::kw_only(),
```
awni, line 2291:
```suggestion
      py::kw_only(),
      "scale"_a,
      "mask"_a = none,
```

python/src/ops.cpp
```diff
@@ -2280,6 +2280,38 @@ void init_ops(py::module_& m) {
...
+
+          Args:
+              q (array): Input query vector.
+              k (array): Input keys matrix from KV cache.
+              v (array): Input values matrix from KV cache.
+              scale (float): Scale for queries (typically 1 / sqrt(d_k))
+              mask (array, optional): Mask for prompt encoding
+
+          Returns:
+              array: o (array), the output attention score
```
awni, line 2312: This is not the `score` right?  I is the weighted hidden state. Maybe change to:

```suggestion
              array: the output of the attention
```

Note also the name and type annotation aren't included in the return string

python/src/ops.cpp
```diff
@@ -2280,6 +2280,38 @@ void init_ops(py::module_& m) {
...
+      R"pbdoc(
+          fast_inference_sdpa(q: array, k: array, v: array,  /, *, stream: Union[None, Stream, Device] = None) -> array
+
+          A parallelized implementation of multi-head attention: O = softmax(Q @ K.T, dim=-1) @ V,
+          with a split along the KV sequence length.  Supports MHA, GQA, and MQA.
+
+          Perform the (possibly batched) matrix multiplication of two arrays. This function is an inference-only
+          kernel optimized specifically for KV-cached transformer decoder inference (query sequence length = 1) and
+          large KV-cached sequences.  It handles prompt encoding via MLX primitives, optimized metal kernel
+          is for decoding only.
```
awni, line 2302: The first sentence is out of place? I think if we do the fallback + custom primitive then we shouldn't need to caveat the application of this function for inference only. So you could probably remove this entire paragraph.

python/src/ops.cpp
```diff
@@ -2280,6 +2280,38 @@ void init_ops(py::module_& m) {
...
+      "v"_a,
+      "scale"_a,
+      "mask"_a = none,
+      py::kw_only(),
+      "stream"_a = none,
+      R"pbdoc(
+          fast_inference_sdpa(q: array, k: array, v: array,  /, *, stream: Union[None, Stream, Device] = None) -> array
+
+          A parallelized implementation of multi-head attention: O = softmax(Q @ K.T, dim=-1) @ V,
+          with a split along the KV sequence length.  Supports MHA, GQA, and MQA.
```
awni, line 2297:
```suggestion
          Scaled dot-product attention.
          
          Applies the function: ``O = softmax(Q @ K.T, dim=-1) @ V``
```

I didn't understand the comment `with a split along the KV sequence length`. That sounds like an implementation detail?

Also Regarding the different options `MHA`, `GQA` and `MQA`. I think those acronyms are not widely understood. Would be great to extend the docstring slightly to explain the constraints / supported cases.

mlx/primitives.h
```diff
@@ -856,6 +856,30 @@ class Floor : public UnaryPrimitive {
   void eval(const std::vector<array>& inputs, array& out);
 };
 
+class FastInferenceSDPA : public UnaryPrimitive {
```
awni, line 859: Nit: Based on the name we go for in the high-level op API, I think it makes sense to also change the name of this primitive to reflect that.

mlx/backend/metal/kernels/fast_inference_sdpa_params.h
```diff
@@ -0,0 +1,21 @@
+//
+//  MLXFastInferenceSDPAParams.h
+//  mlx
+//
+//  Copyright (C) 2024 Argmax, Inc.
```
awni, line 5: I don't know if we can have this in the source code. If you are cool to remove it that would be simplest. O/w will need to check internally that it's ok.

mlx/ops.cpp
```diff
@@ -2222,6 +2222,176 @@ array matmul(
...
+    const array& queries,
+    const array& keys,
+    const array& values,
+    const float scale,
+    const std::optional<array>& mask,
+    StreamOrDevice s) {
+  auto q = queries;
+  auto k = keys;
+  auto v = values;
+  auto m = mask;
```
awni, line 2237: I would just change the function names (and make them pass by values instead of ref to const) instead of doing the copy here.

mlx/ops.cpp
```diff
@@ -2222,6 +2222,176 @@ array matmul(
...
+
+  if (q.ndim() == 0 || k.ndim() == 0 || v.ndim() == 0) {
+    throw std::invalid_argument(
+        "[fast_inference_sdpa] Got 0 dimension input. Inputs must "
+        "have at least one dimension.");
+  }
```
awni, line 2243: It seems like you check the input dims are not 0 here but then check again that they are exactly 4 below? Maybe this check is superfluous?

mlx/ops.cpp
```diff
@@ -2222,6 +2222,176 @@ array matmul(
...
+  // Multihead attention, n_heads == n_kv_heads; Multi-query attention,
+  // n_kv_heads == 1; and for Grouped Query Attention, n_kv_heads < n_heads,
+  // n_kv_heads > 1, n_heads % n_kv_heads == 0. K, V tensors must have same
+  // number of heads.
+
+  // Q, K must have matching last dims (d_k);
+  if (q.shape(-1) != k.shape(-1)) {
+    std::ostringstream msg;
+    msg << "[fast_inference_sdpa] query, keys expected to have matching d_k; found query "
+        << q.shape() << " for keys " << k.shape() << ".";
```
awni, line 2282: Nit: the `d_k` here is obscure. Maybe replace with something like:

```suggestion
    msg << "[fast_inference_sdpa] query, keys expected to have matching last dimension, but got query shape "
        << q.shape() << " for keys shape " << k.shape() << ".";
```

mlx/ops.cpp
```diff
@@ -2222,6 +2222,176 @@ array matmul(
...
+    throw std::invalid_argument(msg.str());
+  }
+
+  // head_dim must be 128.
+  constexpr const int supported_head_dim = 128;
+  if (q.shape(-1) != supported_head_dim) {
+    std::ostringstream msg;
+    msg << "[fast_inference_sdpa] required: args.head_dim == 128 ; found "
```
awni, line 2290:
```suggestion
    msg << "[fast_inference_sdpa] query head dim must be 128 ; found "
```

mlx/ops.cpp
```diff
@@ -2222,6 +2222,176 @@ array matmul(
...
+        << q.shape(-1) << " .";
+    throw std::invalid_argument(msg.str());
+  }
+
+  // K, V must have matching head dims (n_kv_heads);
```
awni, line 2295:
```suggestion
  // K, V must have matching head number of heads (n_kv_heads);
```

mlx/ops.cpp
```diff
@@ -2222,6 +2222,176 @@ array matmul(
...
+    std::ostringstream msg;
+    msg << "[fast_inference_sdpa] query sequence length must be 1 at decode time, found  "
+        << q.shape(-2) << ".";
+    throw std::invalid_argument(msg.str());
+  }
+
+  // Type promotion: common type between q, k, v;
+  auto out_type = promote_types(q.dtype(), k.dtype());
+  auto out_type_kv = promote_types(k.dtype(), v.dtype());
+  auto final_type = promote_types(out_type, out_type_kv);
```
awni, line 2325:
```suggestion
  auto final_type = result_type({q, k , v});
```

mlx/ops.cpp
```diff
@@ -2222,6 +2222,176 @@ array matmul(
...
+  if (q.dtype() != out_type) {
+    q = astype(q, out_type, s);
+  }
+  if (k.dtype() != out_type) {
+    k = astype(k, out_type, s);
+  }
+
+  if (v.dtype() != final_type) {
+    v = astype(v, final_type, s);
+  }
```
awni, line 2336: Nit: you don't need these checks. `astype(...)` is a no-op if the type is correct.

mlx/ops.cpp
```diff
@@ -2222,6 +2222,176 @@ array matmul(
...
+      std::vector<int>({q.shape(0), q.shape(1), q.shape(2), v.shape(-1)});
+
+  auto prompt_encoding_with_or_without_tiling = [](const array& q_tensor,
+                                                   const array& k_tensor,
+                                                   const array& v_tensor,
+                                                   const array& mask_tensor,
+                                                   const float alpha,
+                                                   const Dtype out_dtype,
+                                                   StreamOrDevice s) {
+    auto q_scaled = multiply(array(alpha), q_tensor, s);
```
awni, line 2348: Use `array(alpha, q_tensor.dtype())` to avoid an implicit upcast if `q_tensor` is a half type.

mlx/ops.cpp
```diff
@@ -2222,6 +2222,176 @@ array matmul(
...
+  }
+
+  if (v.dtype() != final_type) {
+    v = astype(v, final_type, s);
+  }
+
+  auto out_shape =
+      std::vector<int>({q.shape(0), q.shape(1), q.shape(2), v.shape(-1)});
+
+  auto prompt_encoding_with_or_without_tiling = [](const array& q_tensor,
```
awni, line 2341: So you basically already have the fallback function here which is great. We can use this with the CustomPrimitive if you change the `SDPA` primitive to inherit from it. That will give us all function transformations out of the box!

awni (commented):
@bpkeene I took a look everything looks really great so far! 

Before I review the API further, I think it makes sense to change to the fast package / namespace and inherit from the `Custom` primitive. That is conceptually pretty straightforward, you have all the pieces in place but might be a big diff. It should also simplify some stuff (like you can remove the VJP/JVP/VMAP for `SDPA`).

mlx/backend/metal/fast_inference_sdpa.cpp
```diff
@@ -0,0 +1,236 @@
...
+        "[fast_inference_sdpa] q_transposed NYI for Metal backend.");
+  }
+  if (k_transposed) {
+    throw std::runtime_error(
+        "[fast_inference_sdpa] k_transposed NYI for Metal backend.");
+  }
+  if (v_transposed) {
+    throw std::runtime_error(
+        "[fast_inference_sdpa] v_transposed NYI for Metal backend.");
+  }
```
awni, line 184: I think (?) you could remove this by removing the case where we do the copy. So basically L158 gets removed and it will copy everything to be row-contiguous.

mlx/backend/metal/fast_inference_sdpa.cpp
```diff
@@ -0,0 +1,236 @@
...
+  // Keep a vector with copies to be cleared in the completed buffer to release
+  // the arrays
+  std::vector<array> temporaries;
+  auto check_transpose = [&temporaries, &s](const array& arr) {
+    auto stx = arr.strides()[arr.ndim() - 2];
+    auto sty = arr.strides()[arr.ndim() - 1];
+    if (stx == arr.shape(-1) && sty == 1) {
+      return std::make_tuple(false, stx, arr);
+    } else if (stx == 1 && sty == arr.shape(-2)) {
+      return std::make_tuple(true, sty, arr);
```
awni, line 159: Removing these lines, so you get a contiguous copy for transpose.

python/tests/test_autograd.py
bpkeene: some formatter did that, I'll see if I can revert and it'll leave them alone (might have run it before pulling the hooks or something? not sure)

mlx/backend/metal/kernels/fast_inference_sdpa_params.h
bpkeene: cool to remove

python/tests/test_autograd.py
awni: Yea that's possible. It's minor so don't worry if it's tedious. As long as it passes the pre-commit hooks it's ok. You can try running `pre-commit run --all-files` to see if it yells at you. (Our CI will run that as well once we run the tests for this)

mlx/backend/metal/fast_inference_sdpa.cpp
```diff
@@ -0,0 +1,236 @@
...
+    compute_encoder->setBytes(&params, sizeof(MLXFastInferenceSDPAParams), 3);
+    set_array_buffer(compute_encoder, out, 4);
+
+    MTL::Size grid_dims_reduce = MTL::Size(heads, 1, batch);
+    MTL::Size group_dims_reduce = MTL::Size(128, 1, 1);
+
+    compute_encoder->dispatchThreadgroups(grid_dims_reduce, group_dims_reduce);
+
+    d.get_command_buffer(s.index)->addCompletedHandler(
+        [&, temporaries](MTL::CommandBuffer*) mutable { temporaries.clear(); });
```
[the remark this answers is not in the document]
bpkeene: yes, removing.  & was leftover from some debugging.

thanks!

mlx/backend/metal/fast_inference_sdpa.cpp
bpkeene: i like that, nice

mlx/backend/metal/fast_inference_sdpa.cpp
bpkeene: done, yes, should be safe.  I also never ran in to the transposed case in practice.

mlx/ops.cpp
bpkeene: sorry, can you clarify?

I was mostly following that matmul() primitive in style:

```
array matmul(
    const array& in_a,
    const array& in_b,
  auto a = in_a;
  auto b = in_b;
```

mlx/ops.cpp
bpkeene: yes

mlx/ops.cpp
bpkeene: nice, thanks!

malfet:
@bpkeene would you be interested in porting this kernel to PyTorch? Will be happy to help with the port

bpkeene:
> @bpkeene would you be interested in porting this kernel to PyTorch? Will be happy to help with the port

@malfet we’d be happy to guide a port to pytorch if anyone would like to take point, if so please cc us and we’re here to help

mlx/ops.cpp
bpkeene: yes, can remove the zero case and leave the rank 4 check

mlx/ops.cpp
bpkeene: nice, thanks

mlx/primitives.h
bpkeene: yes, we're going to use 

scaled_dot_product_attention;  and 
ScaledDotProductAttention

and put in the mx.fast namespace (python, & c analogue etc., fleshing out the details locally)

python/src/ops.cpp
```diff
@@ -2280,6 +2280,38 @@ void init_ops(py::module_& m) {
         Returns:
             list(array): A list of split arrays.
       )pbdoc");
+  m.def(
+      "fast_inference_sdpa",
+      &fast_inference_sdpa,
+      "q"_a,
+      "k"_a,
+      "v"_a,
+      "scale"_a,
```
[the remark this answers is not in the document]
bpkeene: sgtm

python/src/ops.cpp
bpkeene: sgtm, will update and remove the comment regarding split along KV sequence length & expand MHA/GQA/MQA etc. with clarification on the tensor dimensions

python/src/ops.cpp
bpkeene: yes, can remove that.

the GPU shaders would remain supporting inference only though? unless there is a way to route to the MLX primitives sequence, as of now only the prompt encoding or CPU impl would go through mlx primitives?

python/tests/test_fast_inference_sdpa.py
bpkeene: sgtm! will move to ACKNOWLEDGEMENTS.md & add Argmax, Inc. :)

awni: Awesome! Please take liberty to add anyone who contributed to this on your team (e.g. Atila as well)

python/src/ops.cpp
awni: Yea that's right. So the way it would work is in the `sdpm` op itself you would route to the fast primitive if the conditions made sense (e.g. query was a single example, supported data type, etc) otherwise you call the fallback. Here's [an example](https://github.com/ml-explore/mlx/blob/main/mlx/fast.cpp#L111-L119).

It might be nice add a comment about when the actual fast op will be used, but I would put it in terms of constraints on the inputs (rather than "inference only"), we can update it if the conditions change.

bpkeene: ok nice!  yes that's a great example, perfect.

python/tests/test_fast_inference_sdpa.py
```diff
@@ -0,0 +1,114 @@
...
+                    rtol = 1e-5
+                    atol = 1e-1
+
+                    if SEQUENCE_LENGTH > 500:
+                        rtol = 1e-2
+
+                    if DTYPE == np.half:
+                        rtol = 1e-2
+
+                    self.assertTrue(np.allclose(o_mlx, reference, rtol=rtol, atol=atol))
```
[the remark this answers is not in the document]
bpkeene: done

python/tests/test_fast_inference_sdpa.py
bpkeene: done

python/tests/test_fast_inference_sdpa.py
bpkeene: removed

python/tests/test_fast_inference_sdpa.py
bpkeene: removed

python/tests/test_autograd.py
bpkeene: this should be good in the latest commit

mlx/ops.cpp
bpkeene: done

mlx/ops.cpp
bpkeene: done

mlx/ops.cpp
bpkeene: done

mlx/ops.cpp
bpkeene: done

mlx/ops.cpp
bpkeene: done, refactored to be compatible with Custom primitive.  

I believe in the current implementation, the dispatch rules should just directly call the lambda function within the mlx::fast::scaled_dot_product_attention(), rather than going through ScaledDotProductAttention::fallback_(), but please correct my understanding if things work differently

bpkeene:
running additional tests before removing draft status, but the op is now moved to mlx::fast:: (cpp) and mx.fast (python) namespaces using the Custom primitives, with a more thorough fallback_() lambda function backing it (prompt encoding & runs on CPU stream, with metal shaders exercised for GPU stream in the decoding use case)

ACKNOWLEDGMENTS.md
```diff
@@ -13,7 +13,7 @@ MLX was developed with contributions from the following individuals:
 - Diogo Da Cruz: Added `tri`, `tril`, `triu`, `tensordot`, `inner`, `outer`, `tile`, `StreamContext`, `stream` and safetensor support.
 - Gabrijel Boduljak: Added `mlx.core.linalg`, implemented `norm` method and `InstanceNorm` layer. Implemented ``MaxPool1d``, ``MaxPool2d``, ``AvgPool1d``, ``AvgPool2d``.
 - Hinrik Snær Guðmundsson: Added `atleast_1d`, `atleast_2d`, `atleast_3d` ops.
-
+- Brian Keene & Atila Orhon, with Argmax Inc.: Added `fast.scaled_dot_product_attention`
```
awni, line 16: ❤️

python/src/fast.cpp
```diff
@@ -56,4 +56,45 @@ void init_extensions(py::module_& parent_module) {
...
+      "scale"_a,
+      "mask"_a = none,
+      "stream"_a = none,
+      R"pbdoc(
+                  scaled_dot_product_attention(q: array, k: array, v: array, *, scale: float, /,  mask: Union[None, array] = None, stream: Union[None, Stream, Device] = None) -> array
+
+            A parallelized implementation of multi-head attention: O = softmax(Q @ K.T, dim=-1) @ V.
+            Supports Multi-Head Attention (see https://arxiv.org/abs/1706.03762),
+            Grouped Query Attention (https://arxiv.org/abs/2305.13245),
+            and Multi-Query Attention (https://arxiv.org/pdf/1911.02150.pdf).
```
awni, line 83:
```suggestion
            and Multi-Query Attention (https://arxiv.org/abs/1911.02150).
```

python/src/fast.cpp
```diff
@@ -56,4 +56,45 @@ void init_extensions(py::module_& parent_module) {
...
+
+            This function is an inference-focused kernel optimized specifically for KV-cached transformer
+            decoder inference (query sequence length = 1) and large KV-cached sequences.
+            It handles prompt encoding via MLX primitives.  The optimized metal kernel is for decoding only.
```
awni, line 87:
```suggestion
            This function will dispatch to an optimized kernel when the query sequence length is 1.
            It handles other cases with regular MLX operations.
```

python/src/fast.cpp
```diff
@@ -56,4 +56,45 @@ void init_extensions(py::module_& parent_module) {
...
+
+            Args:
+                q (array): Input query array.
+                k (array): Input keys array.
+                v (array): Input values array.
+                scale (float): Scale for queries (typically 1.0 / sqrt(q.shape(-1))
+                mask (array, optional): Mask for prompt encoding
```
awni, line 94:
```suggestion
                mask (array, optional): An additive mask to apply to the query-key scores.
```

python/src/fast.cpp
```diff
@@ -56,4 +56,45 @@ void init_extensions(py::module_& parent_module) {
...
+
+            This function is an inference-focused kernel optimized specifically for KV-cached transformer
+            decoder inference (query sequence length = 1) and large KV-cached sequences.
+            It handles prompt encoding via MLX primitives.  The optimized metal kernel is for decoding only.
+
+            Args:
+                q (array): Input query array.
+                k (array): Input keys array.
+                v (array): Input values array.
+                scale (float): Scale for queries (typically 1.0 / sqrt(q.shape(-1))
```
awni, line 93:
```suggestion
                scale (float): Scale for queries (typically ``1.0 / sqrt(q.shape(-1)``)
```

python/src/fast.cpp
```diff
@@ -56,4 +56,45 @@ void init_extensions(py::module_& parent_module) {
...
+      },
+      "q"_a,
+      "k"_a,
+      "v"_a,
+      py::kw_only(),
+      "scale"_a,
+      "mask"_a = none,
+      "stream"_a = none,
+      R"pbdoc(
+                  scaled_dot_product_attention(q: array, k: array, v: array, *, scale: float, /,  mask: Union[None, array] = None, stream: Union[None, Stream, Device] = None) -> array
```
awni, line 78:
```suggestion
                  scaled_dot_product_attention(q: array, k: array, v: array, *, scale: float,  mask: Union[None, array] = None, stream: Union[None, Stream, Device] = None) -> array
```

mlx/fast_primitives.h
```diff
@@ -65,4 +65,39 @@ class RoPE : public Custom {
...
+  void eval_gpu(const std::vector<array>& inputs, array& out);
+  bool is_equivalent(const Primitive& other) const override;
+
+  DEFINE_PRINT(ScaledDotProductAttention)
+
+ private:
+  std::function<std::vector<array>(std::vector<array>)> fallback_;
+  float scale_;
+  bool needs_mask_;
+  void eval(const std::vector<array>& inputs, array& out);
```
awni, line 100: I don't think you need that function.

mlx/fast_primitives.h
```diff
@@ -65,4 +65,39 @@ class RoPE : public Custom {
...
+  };
+
+  void eval_gpu(const std::vector<array>& inputs, std::vector<array>& outputs)
+      override {
+    return eval_gpu(inputs, outputs[0]);
+  };
+
+  void eval_cpu(const std::vector<array>& inputs, array& out_array) {
+    out_array = fallback_(inputs)[0];
+  };
```
awni, line 89: Nit: I would just put this line in the `eval_cpu` above directly. No need to have this function

mlx/fast_primitives.h
```diff
@@ -65,4 +65,39 @@ class RoPE : public Custom {
...
+  explicit ScaledDotProductAttention(
+      Stream stream,
+      std::function<std::vector<array>(std::vector<array>)> fallback,
+      const float scale,
+      const bool needs_mask)
+      : Custom(stream, fallback), scale_(scale), needs_mask_(needs_mask){};
+
+  void eval_cpu(const std::vector<array>& inputs, std::vector<array>& outputs)
+      override {
+    return eval_cpu(inputs, outputs[0]);
```
awni, line 79:
```suggestion
    outputs[0] = fallback_(inputs)[0];
```

(and delete the extra `eval_cpu` below

mlx/fast_primitives.h
```diff
@@ -65,4 +65,39 @@ class RoPE : public Custom {
...
+  };
+
+  void eval_gpu(const std::vector<array>& inputs, std::vector<array>& outputs)
+      override {
+    return eval_gpu(inputs, outputs[0]);
```
awni, line 84: these functions don't return anything:

```suggestion
    eval_gpu(inputs, outputs[0]);
```

python/src/fast.cpp
awni: Would you also mind adding something about the fact that the softmax is done in higher precision regardless of the precision of the inputs?

E.g.:

The ``softmax`` is always performed in ``float32`` and then cast back to the input dtype before multiplying with ``V``.

mlx/backend/metal/scaled_dot_product_attention.cpp
```diff
@@ -0,0 +1,222 @@
...
+  kname_partials << "fast_inference_sdpa_compute_partials_";
+
+  std::ostringstream kname_reduce;
+  std::string delimiter = "_";
+  kname_reduce << "fast_inference_sdpa_reduce_tiles" + delimiter;
+
+  for (const auto& arr : {k, v, out}) {
+    if (arr.dtype() != q.dtype()) {
+      throw std::runtime_error(
+          "[fast_inference_sdpa]: expected matching dtypes for q,k,v,o");
```
awni, line 49:
```suggestion
          "[scaled_dot_product_attention::eval_gpu]: expected matching dtypes for q,k,v,o");
```

mlx/backend/metal/scaled_dot_product_attention.cpp
```diff
@@ -0,0 +1,222 @@
...
+
+  if (q.dtype() == float32) {
+    kname_partials << "float" + delimiter;
+    kname_reduce << "float";
+  } else if (q.dtype() == float16) {
+    kname_partials << "half" + delimiter;
+    kname_reduce << "half";
+  } else {
+    throw std::runtime_error(
+        "[scaled_dot_product_attention::eval_gpu]: unexpected dtype found for queries: expected either float32 or float16.");
```
awni, line 61:
```suggestion
        "[ScaledDotProductAttention::eval_gpu]: unexpected dtype found for queries: expected either float32 or float16.");
```

mlx/fast.cpp
```diff
@@ -127,4 +127,168 @@ bool RoPE::is_equivalent(const Primitive& other) const {
...
+      throw std::invalid_argument(msg.str());
+    }
+  }
+
+  size_t batch_dim = queries.shape(0);
+  if (batch_dim != 1) {
+    std::ostringstream msg;
+    msg << "[scaled_dot_product_attention] Expected batch dimension of 1. ";
+    throw std::invalid_argument(msg.str());
+  }
```
awni, line 152: Correct me if I'm wrong, but it looks like the fallback will support batch > 1. If that is correct (which would be great), then maybe we use this in the dispatch condition rather than throwing here?

mlx/fast.cpp
```diff
@@ -127,4 +127,168 @@ bool RoPE::is_equivalent(const Primitive& other) const {
...
+  for (const auto& tensor : {queries, keys, values}) {
+    if (tensor.shape(0) != batch_dim) {
+      std::ostringstream msg;
+      msg << "[scaled_dot_product_attention] mismatching batch dimension for input with shape "
+          << tensor.shape() << ".";
```
awni, line 157:
```suggestion
          << tensor.shape() << " and batch size " << batch_dim << ".";
```

mlx/fast.cpp
```diff
@@ -127,4 +127,168 @@ bool RoPE::is_equivalent(const Primitive& other) const {
...
+  // Q, K must have matching last dims (d_k aka 'head_dim');
+  if (queries.shape(-1) != keys.shape(-1)) {
+    std::ostringstream msg;
+    msg << "[scaled_dot_product_attention] query, keys expected to have matching last dimension; found query shape "
+        << queries.shape() << " for keys shape " << keys.shape() << ".";
+    throw std::invalid_argument(msg.str());
+  }
+
+  // head_dim must be 128.
+  constexpr const int supported_head_dim = 128;
```
awni, line 171: Same comment here regarding the fallback. I think it can handle this, so maybe use it in the condition for dispatch instead?

mlx/fast.cpp
```diff
@@ -127,4 +127,168 @@ bool RoPE::is_equivalent(const Primitive& other) const {
...
+    throw std::invalid_argument(msg.str());
+  }
+
+  // q_seq_len == 1 for now in decoding (no mask) implementation
+  if (queries.shape(-2) > 1 && !mask.has_value()) {
+    std::ostringstream msg;
+    msg << "[scaled_dot_product_attention] query sequence length must be 1 at decode time, found  "
+        << queries.shape(-2) << ".";
+    throw std::invalid_argument(msg.str());
+  }
```
awni, line 205: Maybe the same comment here re using the fallback?

mlx/fast.cpp
```diff
@@ -127,4 +127,168 @@ bool RoPE::is_equivalent(const Primitive& other) const {
...
+        auto& k_tensor = inputs[1];
+        auto& v_tensor = inputs[2];
+        auto q_scaled = multiply(array(scale, q_tensor.dtype()), q_tensor, s);
+        int n_q_heads = q_tensor.shape()[1];
+        int n_kv_heads = k_tensor.shape()[1];
+
+        auto tile_if_needs_repeat = [](const array& arr,
+                                       int query_heads,
+                                       int kv_heads,
+                                       StreamOrDevice& s) -> array {
```
awni, line 236: nit: you could capture these in the lambda and type a little less in the function calls below.

mlx/fast.cpp
```diff
@@ -127,4 +127,168 @@ bool RoPE::is_equivalent(const Primitive& other) const {
...
+
+        // dim check on k, v; repeat if untiled, since naive matmul will have
+        // dim mismatch for GQA (MQA could make use of broadcast)
+        auto k_transposed = transpose(k_tensor_tiled, {0, 1, 3, 2}, s);
+        auto s_tensor = matmul(q_scaled, k_transposed, s);
+        if (needs_mask) {
+          auto mask_tensor = inputs[3];
+          s_tensor = add(s_tensor, mask_tensor, s);
+        }
+        auto p = softmax(s_tensor, std::vector<int>({-1}), s);
```
awni, line 259:
```suggestion
        auto p = softmax(s_tensor, std::vector<int>{-1}, s);
```

mlx/fast.cpp
```diff
@@ -127,4 +127,168 @@ bool RoPE::is_equivalent(const Primitive& other) const {
...
+        auto out_tensor = matmul(p, v_tensor_tiled, s);
+        out_tensor = astype(out_tensor, final_type, s);
```
awni, line 261: Interesting that you do the cast after the out projection here, was that intentional? In almost all cases we do it before the output projection. I think we should do that here as well.

awni (commented):
Really really nice! Can't wait to land this. Left a few more mostly cosmetic suggestions.

@jagrit06 is planning to check out the kernel shortly, but we can aim to get this landed very soon.

mlx/backend/metal/kernels/scaled_dot_product_attention.metal
```diff
@@ -0,0 +1,451 @@
...
+        float tgroupExpSum = simd_sum(sumExpLocal);
+
+        lse = log(tgroupExpSum);
+        float2 local_p_hat = expf_shifted / tgroupExpSum;
+        pvals[0].x = local_p_hat.x;
+        pvals[0].y = local_p_hat.y;
+        smemPtrFlt2[simd_lane_id] = float2(0.f);
+    }
+    constexpr const bool TILE_SIZE_LARGER_THAN_64 = TILE_SIZE_CONST > 64;
+    constexpr const int TILE_SIZE_ITERS_128 = (TILE_SIZE_CONST + 1) / 128;
```
jagrit06, line 146: I'm not sure I understood why we're doing `(TILE_SIZE_CONST + 1) / 128` here - we don't hit the loop below if tile size <= 64 and our tile sizes are multiples of 128 so we end up with the same results as `(TILE_SIZE_CONST) / 128` - did I miss why we need the + 1 ?

mlx/backend/metal/kernels/scaled_dot_product_attention.metal
```diff
@@ -0,0 +1,451 @@
...
+            
+            for(size_t tile_start = simd_group_id; tile_start < TILE_SIZE_CONST_DIV_8; tile_start += NSIMDGROUPS) {
+                simdgroup_matrix<T, 8, 8> tmp;
+                ulong simdgroup_matrix_offset = matrix_load_loop_iter * NSIMDGROUPS * SIMDGROUP_MATRIX_LOAD_FACTOR + simd_group_id * SIMDGROUP_MATRIX_LOAD_FACTOR;
+                ulong2 matrixOrigin = ulong2(col * SIMDGROUP_MATRIX_LOAD_FACTOR, simdgroup_matrix_offset);
+                simdgroup_load(tmp, baseV, DK, matrixOrigin, true);
+                const ulong2 matrixOriginSmem = ulong2(simdgroup_matrix_offset, 0);
+                const ulong elemsPerRowSmem = TILE_SIZE_CONST;
+                simdgroup_store(tmp, smemV, elemsPerRowSmem, matrixOriginSmem, false);
+                matrix_load_loop_iter++;
```
jagrit06, line 203: Unless I missed something, it looks like we are just loading in data from device memory to shared memory (and transposing it) - I do not understand why they need to be loaded in as simdgroup matrices and then stored as simdgroup matrices. 

`simdgroup_load` loads the 2d data in as simdgroup matrices - which divides it up into a specific format (simdgroup_matrix) among the warp / simdgroup that is needed to so the `simdgroup_matrix_multiply_accumulate`
Conversely, `simdgroup_store` takes that format and writes it back as regular 2d data. 
If we aren't doing a `simdgroup_matrix_multiply_accumulate`, then it would just add overhead as far as I understand it. 

Is there a particular reason we do this here ? Like maybe you have found transposing blocks as you read in from device to shared memory is faster doing it this way ? If not, might be best to not do it this way

mlx/backend/metal/kernels/scaled_dot_product_attention.metal
```diff
@@ -0,0 +1,451 @@
...
+            uint smem_col_index = simd_group_id * SIMDGROUP_MATRIX_LOAD_FACTOR;
+            int32_t tile_start;
+            for(tile_start = START_ROW + simd_group_id * SIMDGROUP_MATRIX_LOAD_FACTOR; tile_start < MAX_START_ROW; tile_start += NSIMDGROUPS * SIMDGROUP_MATRIX_LOAD_FACTOR) {
+                simdgroup_matrix<T, 8, 8> tmp;
+                ulong2 matrixOrigin = ulong2(col * SIMDGROUP_MATRIX_LOAD_FACTOR, tile_start);
+                simdgroup_load(tmp, baseVThisHead, DK, matrixOrigin, /* transpose */ true);
+                const ulong2 matrixOriginSmem = ulong2(smem_col_index, 0);
+                constexpr const ulong elemsPerRowSmem = TILE_SIZE_CONST;
+                simdgroup_store(tmp, smemV, elemsPerRowSmem, matrixOriginSmem, /* transpose */ false);
+                smem_col_index += NSIMDGROUPS * SIMDGROUP_MATRIX_LOAD_FACTOR;
```
jagrit06, line 266: Same comment as above

mlx/backend/metal/kernels/scaled_dot_product_attention.metal
```diff
@@ -0,0 +1,451 @@
...
+
+#include "mlx/backend/metal/kernels/scaled_dot_product_attention_params.h"
+using namespace metal;
+
+template<typename T, typename T2, typename T4, uint16_t TILE_SIZE_CONST, uint16_t NSIMDGROUPS>
+[[kernel]] void fast_inference_sdpa_compute_partials_template(const device T *Q [[buffer(0)]],
+                              const device T *K [[buffer(1)]],
+                              const device T *V [[buffer(2)]],
+                              const device uint64_t& L [[buffer(3)]],
+                              const device MLXScaledDotProductAttentionParams& params [[buffer(4)]],
```
jagrit06, line 12: Not actually a big deal, but this could probably be in the constant namespace

jagrit06 (commented):
I'd want to do a deeper dive and another pass over this kernel at some later date - but for now, since it works and looks to be fast, I won't block merging it in!

Just a couple questions popped up while I was going over the kernel I was hoping you could clarify

awni:
Also, looks like the linux build is just missing an include (`#include <optional>` in `fast.h`)

mlx/backend/metal/kernels/scaled_dot_product_attention.metal
bpkeene: I did it for the free transpose to shared memory, may very well be the case that it's a perf overhead (I didn't try it the other way).

Adding pre-transposed support for Values data would (should) then just be tripping a flag and swapping some indices

mlx/backend/metal/kernels/scaled_dot_product_attention.metal
bpkeene: that was probably just leftover from development, no need

mlx/fast.cpp
bpkeene: yes, let's do that

mlx/backend/metal/kernels/scaled_dot_product_attention.metal
bpkeene: yes.  params.KV_TILES would change only every (% tile_size) inferences.  followup change, or would you like it for this one?

mlx/fast.cpp
awni: @bpkeene looking at this again, I think the right thing to do is the `softmax` in higher presicion and then downcast that to the final type. See how we do it e.g. [here](https://github.com/ml-explore/mlx-examples/blob/main/llms/mlx_lm/models/llama.py#L100). In almost all cases (training and inference) you want the softmax to be done in fp32. (We could always expose it as an option in the future, but for now I think having that as the default is good).

bpkeene: sgtm! I'll add that to this change as well, along with downcasting afterwards for the output projection (in the event that we're using fp16)

mlx/backend/metal/kernels/scaled_dot_product_attention.metal
jagrit06: You can do it later!

python/src/fast.cpp
bpkeene: Yes, will do

bpkeene: looks like Atila covered this, marking as resolved.  thanks!

mlx/fast_primitives.h
bpkeene: removing, thanks

mlx/fast_primitives.h
bpkeene: sgtm

mlx/fast_primitives.h
bpkeene: ack, removing return

mlx/backend/metal/scaled_dot_product_attention.cpp
bpkeene: nice, thanks

mlx/backend/metal/scaled_dot_product_attention.cpp
bpkeene: sgtm

mlx/fast.cpp
bpkeene: yes

mlx/fast.cpp
bpkeene: yes

mlx/fast.cpp
bpkeene: done

awni (approved):
🚀  this is an exceptional contribution, thanks @bpkeene and @atiorh for making it happen!!

PR #737, opened 2024-02-24
feat: allow clang-format to format all metal files in repo

NripeshN:
## Proposed changes

This change allows metal files to be formatted

## Checklist

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,11 +1,8 @@
 repos:
--   repo: local
+-   repo: https://github.com/NripeshN/clang-format-metal
+    rev: 17.0.6
     hooks:
-    -   id: clang-format
-        name: clang-format
-        entry: clang-format -i
-        language: system
-        files: '\.(cpp|h|hpp|metal)$'
+    -   id: clang-format-metal
 # Using this mirror lets us use mypyc-compiled black, which is about 2x faster
 -   repo: https://github.com/psf/black-pre-commit-mirror
     rev: 24.2.0
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,8 +1,11 @@
 repos:
--   repo: https://github.com/pre-commit/mirrors-clang-format
-    rev: v17.0.6
+-   repo: local
     hooks:
     -   id: clang-format
+        name: clang-format
+        entry: clang-format -i
+        language: system
+        files: '\.(cpp|h|hpp|metal)$'
 # Using this mirror lets us use mypyc-compiled black, which is about 2x faster
 -   repo: https://github.com/psf/black-pre-commit-mirror
     rev: 24.2.0
--- a/mlx/backend/metal/kernels/steel/gemm/kernels/steel_gemm.metal
+++ b/mlx/backend/metal/kernels/steel/gemm/kernels/steel_gemm.metal
@@ -10,80 +10,108 @@ using namespace mlx::steel;
 // GEMM kernels
 ///////////////////////////////////////////////////////////////////////////////
 
-template <typename T,
-          int BM,
-          int BN,
-          int BK,
-          int WM,
-          int WN,
-          bool transpose_a, 
-          bool transpose_b,
-          bool MN_aligned,
-          bool K_aligned>
-[[kernel, max_total_threads_per_threadgroup(WM * WN * 32)]] void gemm(
-    const device T *A [[buffer(0)]],
-    const device T *B [[buffer(1)]],
-    device T *C [[buffer(2)]],
+template <
+    typename T,
+    int BM,
+    int BN,
+    int BK,
+    int WM,
+    int WN,
+    bool transpose_a,
+    bool transpose_b,
+    bool MN_aligned,
+    bool K_aligned>
+[[kernel, max_total_threads_per_threadgroup(WM* WN * 32)]] void gemm(
+    const device T* A [[buffer(0)]],
+    const device T* B [[buffer(1)]],
+    device T* C [[buffer(2)]],
     const constant GEMMParams* params [[buffer(3)]],
     uint simd_lane_id [[thread_index_in_simdgroup]],
     uint simd_group_id [[simdgroup_index_in_threadgroup]],
     uint3 tid [[threadgroup_position_in_grid]],
-    uint3 lid [[thread_position_in_threadgroup]]) { 
-    
-    using gemm_kernel = GEMMKernel<T, T, BM, BN, BK, WM, WN, transpose_a, transpose_b, MN_aligned, K_aligned>;
-    
-    threadgroup T As[gemm_kernel::tgp_mem_size_a];
-    threadgroup T Bs[gemm_kernel::tgp_mem_size_b];
+    uint3 lid [[thread_position_in_threadgroup]]) {
+  using gemm_kernel = GEMMKernel<
+      T,
+      T,
+      BM,
+      BN,
+      BK,
+      WM,
+      WN,
+      transpose_a,
+      transpose_b,
+      MN_aligned,
+      K_aligned>;
 
-    // Adjust for batch
-    A += params->batch_stride_a * tid.z;
-    B += params->batch_stride_b * tid.z;
-    C += params->batch_stride_c * tid.z;
+  threadgroup T As[gemm_kernel::tgp_mem_size_a];
+  threadgroup T Bs[gemm_kernel::tgp_mem_size_b];
 
-    gemm_kernel::run( 
-      A, B, C, 
-      params,
-      As, Bs,
-      simd_lane_id, simd_group_id, tid, lid
-    );
+  // Adjust for batch
+  A += params->batch_stride_a * tid.z;
+  B += params->batch_stride_b * tid.z;
+  C += params->batch_stride_c * tid.z;
+
+  gemm_kernel::run(
+      A, B, C, params, As, Bs, simd_lane_id, simd_group_id, tid, lid);
 }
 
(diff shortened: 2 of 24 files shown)


mlx/backend/metal/kernels/ternary.metal
```diff
@@ -89,96 +92,98 @@ template <typename T, typename Op>
...
+          instantiate_ternary_all(name, uint32, uint32_t, op)               \
+              instantiate_ternary_all(name, uint64, uint64_t, op)           \
+                  instantiate_ternary_all(name, int8, int8_t, op)           \
+                      instantiate_ternary_all(name, int16, int16_t, op)     \
+                          instantiate_ternary_all(name, int32, int32_t, op) \
+                              instantiate_ternary_all(                      \
+                                  name, int64, int64_t, op)                 \
+                                  instantiate_ternary_all(                  \
+                                      name, complex64, complex64_t, op)     \
+                                      instantiate_ternary_float(name, op)
```
awni, line 187: This looks a little strange.. maybe worth adding a no format here.

mlx/backend/metal/kernels/unary.metal
```diff
@@ -22,83 +22,118 @@ template <typename T, typename Op>
...
+  instantiate_unary_all(name, bool_, bool, op)                                \
+      instantiate_unary_all(name, uint8, uint8_t, op) instantiate_unary_all(  \
+          name, uint16, uint16_t, op)                                         \
+          instantiate_unary_all(name, uint32, uint32_t, op)                   \
+              instantiate_unary_all(name, uint64, uint64_t, op)               \
+                  instantiate_unary_all(name, int8, int8_t, op)               \
+                      instantiate_unary_all(name, int16, int16_t, op)         \
+                          instantiate_unary_all(name, int32, int32_t, op)     \
+                              instantiate_unary_all(name, int64, int64_t, op) \
+                                  instantiate_unary_float(name, op)
```
awni, line 59: And here.

mlx/backend/metal/kernels/unary.metal
```diff
@@ -22,83 +22,118 @@ template <typename T, typename Op>
...
+                                                                    Tanh)
+                                                                    instantiate_unary_all(
+                                                                        round,
+                                                                        complex64,
+                                                                        complex64_t,
+                                                                        Round)
 
-instantiate_unary_all(lnot, bool_, bool, LogicalNot)
+                                                                        instantiate_unary_all(
+                                                                            lnot,
```
awni, line 136: and here.

mlx/backend/metal/kernels/steel/gemm/kernels/steel_gemm.metal
```diff
@@ -10,78 +10,189 @@ using namespace mlx::steel;
...
+  instantiate_gemm_transpose_helper(                               \
+      iname, itype, oname, otype, 32, 32, 16, 2, 2)                \
+      instantiate_gemm_transpose_helper(                           \
+          iname, itype, oname, otype, 64, 64, 16, 2, 2)            \
+          instantiate_gemm_transpose_helper(                       \
+              iname, itype, oname, otype, 64, 32, 32, 2, 2)        \
+              instantiate_gemm_transpose_helper(                   \
+                  iname, itype, oname, otype, 64, 32, 16, 2, 2)    \
+                  instantiate_gemm_transpose_helper(               \
+                      iname, itype, oname, otype, 32, 64, 16, 2, 2)
```
awni, line 195: this looks bad to me, guard with no formatting

mlx/backend/metal/kernels/steel/gemm/kernels/steel_gemm_addmm.metal
```diff
@@ -10,243 +10,433 @@ using namespace mlx::steel;
...
+                  bn,                                                        \
+                  bk,                                                        \
+                  wm,                                                        \
+                  wn,                                                        \
+                  naligned,                                                  \
+                  false,                                                     \
+                  naligned,                                                  \
+                  false)
+
+#define instantiate_gemm_transpose_helper(                                     \
```
awni, line 407: Same here.

mlx/backend/metal/kernels/steel/gemm/kernels/steel_gemm_addmm.metal
```diff
@@ -10,243 +10,433 @@ using namespace mlx::steel;
...
+  instantiate_gemm_transpose_helper(                               \
+      iname, itype, oname, otype, 32, 32, 16, 2, 2)                \
+      instantiate_gemm_transpose_helper(                           \
+          iname, itype, oname, otype, 64, 64, 16, 2, 2)            \
+          instantiate_gemm_transpose_helper(                       \
+              iname, itype, oname, otype, 64, 32, 32, 2, 2)        \
+              instantiate_gemm_transpose_helper(                   \
+                  iname, itype, oname, otype, 64, 32, 16, 2, 2)    \
+                  instantiate_gemm_transpose_helper(               \
+                      iname, itype, oname, otype, 32, 64, 16, 2, 2)
```
awni, line 439: And here.

awni (changes requested):
I'd actually be in favor of adding formatting to metal code, so thanks for starting that. There are several places where we instantiate a bunch of kernels with the preprocessor and the formatted makes it very hard to read. Would you mind guarding those sections by turning the formatter off?

.pre-commit-config.yaml
```diff
@@ -1,8 +1,11 @@
 repos:
--   repo: https://github.com/pre-commit/mirrors-clang-format
-    rev: v17.0.6
+-   repo: local
```
awni, line 2: Why remove the repo/version here?

awni (commented):
Left a question, let me know. Otherwise I think it looks good, and we can merge shortly.

.pre-commit-config.yaml
NripeshN: Clang-format's [pre-commit hook](https://github.com/pre-commit/mirrors-clang-format/blob/main/.pre-commit-hooks.yaml) does not have support to add metal files. Hence using custom hook seems like the solution

awni: Could we keep the repo/version and instead specify the extensions in a `.clang-format` config?

I'm worried this is a bit brittle / relies on people having the same version of clang installed.

NripeshN: I have created a custom pre-commit that will format all metal files now. Should solve the above problem

awni: I don’t think that’s the right solution. In that case we have to rely on your hook to be updated and I’m not so comfortable having it as a dependency

NripeshN: Maybe you can fork my repo in your organization and use the hook from there.

NripeshN: The `.clang-format` configuration file is used to specify coding style preferences for `clang-format` but does not directly control which file extensions are processed by `clang-format`. The handling of specific file extensions and the invocation of `clang-format` on those files is typically managed outside of the `.clang-format` file, such as through script arguments or, in our case, pre-commit hook configurations.

Since the `mirrors-clang-format` pre-commit hook configuration we're using doesn't natively support `.metal` files through a simple configuration change in the `.pre-commit-config.yaml` or `.clang-format`, and you wish to avoid a `local` repo setup due to concerns about version consistency, we're in a bit of a bind. The pre-commit framework itself relies on matching files to hooks based on the configuration parameters provided in `.pre-commit-config.yaml`, not `.clang-format`.

NripeshN: I have created a [PR](https://github.com/pre-commit/mirrors-clang-format/pull/31) in their repo.

NripeshN: @awni they have closed the PR as the file I made changes to is being generated.
Given the context that the `.pre-commit-hooks.yaml` file is generated by a [GitHub Actions](https://github.com/pre-commit/mirrors-clang-format/blob/main/.github/workflows/main.yml) workflow, and the maintainers have indicated that my PR was closed because modifications should be made to the generation process rather than directly to the file, we'll need to modify the [GitHub Action](https://github.com/pre-commit/mirrors-clang-format/blob/main/.github/workflows/main.yml) itself to ensure `.metal` files are considered in the generated configuration.

The[ GitHub Action](https://github.com/pre-commit/mirrors-clang-format/blob/main/.github/workflows/main.yml) uses `pre-commit-mirror-maker` to automatically generate the pre-commit hook configuration for `clang-format`. The key part of the action where types are specified for the `clang-format` hook is in the `pre-commit-mirror` command. Since `.metal` files are not recognized as a type by pre-commit, we cannot directly add a new type for `.metal` files in the `--types-or` options.

However, considering the limitations, the goal to support Metal file formatting within the constraints of the existing system, we're somewhat at an impasse because the `pre-commit-mirror-maker` utility and the pre-commit framework itself rely on recognized file types.

awni: >  Since .metal files are not recognized as a type by pre-commit

Can that be updated? It seems like they recognize `cuda`, `metal` is pretty standard.

NripeshN: I had created [this](https://github.com/pre-commit/mirrors-clang-format/pull/32) PR in their repo and they are hesitant to merge it right away. Also I tried running the pre-commit hook using the commits SHA and it did not work. I think pre-commit itself does not support metal. Not entirely sure either.

awni: Nice that looks like the right solution to me. Maybe we wait until they merge that and then we can come back to this.

awni: @NripeshN should I leave this PR open until Metal is in the precommit? Are you planning to follow up there?

PR #745, opened 2024-02-26
Fix compile with non standard types

awni:
## Proposed changes

- Compile throws you call the function with anything other than a tree of arrays or constants
- Refactored C++ tree code into separate files


angeloskath (changes requested):
Much needed refactor! Thanks.

I added a test that showcases the problem. Granted flattening the pytrees is much better than the temporary stopgap that I had implemented but I do think we need to hash the pytree structure somehow otherwise we can't recognize different structures that result in the same flat constants.

I didn't push a solution as it is a matter of choice. We could allow only lists, tuples, and scalars or we could hash the structure. We can hash with 1 value or 1 per container or 2 per container.

My go-to solution would be to add constants that encode the structure like an int for list tuple etc (we do run the risk of collision but a big-ish and prime-ish number per container should not be too bad).

awni:
Right now, I kind of think the tree should be hashed. It's not an issue just limited to constants right now, we've been kind of avoiding it but it's also relevant for arrays as well. If you change the tree structure (without changing shapes or types) probably you expect the function to be recompiled.

awni:
@angeloskath I added a start to hash the tree structure. Can you look and see if we should add / change it? I didn't bother to encode the type yet.... not sure there. We can add it if it makes sense. There's definitely some edge cases / usage questions that have to be made there (like should changing from a list to a tuple trigger recompilation? My thought was no, but maybe it should be yes?).

angeloskath (approved):
That looks great! I am not sure why you keep two vectors and move one into the other instead of just one.

angeloskath:
Regarding the list to tuple. I can think of reasons why it should but I think it is ok to draw the line there and assume that we take the general structure of the arguments.

Another thing is that this is still a bit brittle as for instance it is not clear how one would differentiate between the following calls.

```python
x = mx.array(0)
y = 0

f(x, y)
f(y, x)
```

This is why I mentioned to encode the full structure.

angeloskath:
@awni I pushed the hack I mentioned above which encodes the full argument structure in the `constants` vector. I added the above test as well and it works fine. I ran the transformer example and saw no performance regression (I guess it makes sense as we have to go over the vector of arrays to check types and shapes anyway).

If you think it is too hacky feel free to revert otherwise we can merge it.

awni:
>  I am not sure why you keep two vectors and move one into the other instead of just one.

You're guess is as good as mine.. thanks for changing it.

I am good to go on this PR. I was going to do the doc fix here but I think it requires more work so we can do that as a follow up. Let me know if you are good we can merge.

PR #747, opened 2024-02-27
Fix cpu compile on older OS

awni:
Tested and works on OS 13.5 and OS 14

Ideally closes https://github.com/ml-explore/mlx-examples/issues/487


angeloskath (approved):
Thanks!

PR #749, opened 2024-02-27
Add back compiled function signatures and docstrings

awni:
## Proposed changes

This seems to work ok in most cases so maybe better than not doing it.

E.g. in the sphinx doc:

Screenshot 2024-02-27 at 7 17 03 AM

And
```shell
>>>help(nn.leaky_relu)
```
gives:
```shell
Help on built-in function leaky_relu:

leaky_relu(...) method of builtins.PyCapsule instance
    leaky_relu(x, negative_slope=0.01)
    
    
    Applies the Leaky Rectified Linear Unit.
    
    Simply ``mx.maximum(negative_slope * x, x)``.
```

### DIFF
--- a/python/mlx/nn/layers/activations.py
+++ b/python/mlx/nn/layers/activations.py
@@ -18,7 +18,8 @@ def _make_activation_module(f):
 
 @partial(mx.compile, shapeless=True)
 def sigmoid(x):
-    r"""Applies the sigmoid function.
+    r"""
+    Applies the sigmoid function.
 
     .. math::
         \text{Sigmoid}(x) = \sigma(x) = \frac{1}{1 + \exp(-x)}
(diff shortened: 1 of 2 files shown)


python/mlx/nn/layers/activations.py
```diff
@@ -18,7 +18,8 @@ def decorator(klass):
 
 @partial(mx.compile, shapeless=True)
 def sigmoid(x):
-    r"""Applies the sigmoid function.
```
angeloskath, line 21: I know it looks better when indented but I think that having it be in the same line is recommended. Is it causing any problem anywhere?

angeloskath (approved):

python/mlx/nn/layers/activations.py
awni: Yes 😞  it causes an issue in the interpretation of the docstring (it treats the first line on a different indentation level than the remainder which messes up how it renders. An alternative is to insert the proper amount of indentation in the first line so it matches the others before setting it on the wrapped function.

awni: I wanted to avoid dealing with a bunch of edge cases there.. if we consistently do not indentat the first line then it might be simple to patch it in the code itself.

awni: I added the indentation to the first line

PR #759, opened 2024-02-29
Avoid nested closures in module

awni:
## Proposed changes

- Avoid nested closure in `nn.Module` which was causing references to be kept for too long
- Also updated `Module.__setattr__` to not add everything to the dictionary. Reduces the `model.state` size from 1463
to 390 for a 32 layer Transformer.

With this change, can remove the calls to garbage collect in e.g. MLX LM (https://github.com/ml-explore/mlx-examples/pull/509)


angeloskath (approved):
Nice! I guess C++ optimizations are spoiling me a bit 😁

Issue #742, opened 2024-02-25
GPU Memory Management?


igm503:
In pytorch, if I delete a reference to a torch.tensor that has been assigned to the mps device and then call torch.mps.empty_cache(), the memory allocated for that tensor will be freed. 

In MLX, however, if I delete a reference to an mx.array that has been evaluated, there's no way to get the memory back. If I am very careful to delete references, I can ensure that the memory reserved by the python process doesn't exceed the maximum instantaneous memory requirement by an eval() call, but in some cases, it's difficult to delete the references, or deleting the references doesn't seem to make a difference to the MLX memory allocator. 

Consider the script below:

```
import mlx.core as mx


fn = lambda x, y: x @ y

a = mx.random.normal(shape=(10000, 10000))
b = mx.random.normal(shape=(10000, 10000))

mx.eval(fn(a, b))

# Memory: 1.13 GB

c = mx.random.normal(shape=(10000, 10000))
d = mx.random.normal(shape=(10000, 10000))

mx.eval(fn(c, d))

# Memory: 1.88 GB

a = None
b = None
c = None
d = None

e = mx.random.normal(shape=(10000, 10000))
f = mx.random.normal(shape=(10000, 10000))

mx.eval(fn(e, f))

# Memory: 1.88 GB

g = mx.random.normal(shape=(10000, 10000))
h = mx.random.normal(shape=(10000, 10000))

mx.eval(fn(g, h))

# Memory: 1.88 GB
```

Setting a, b, c, and d to None has the effect of preventing the e, f, g, and h tensors from increasing the memory usage but not of decreasing the memory usage. I assume that this is because the memory that was used by a, b, c, and d is used in turn by e, f, g, and h, but the memory allocator for the python process still held on to the chunk of memory throughout. 

This seems like it's fine for model training and inference loops, where references are repeatedly going to be recycled, but it causes problems elsewhere. 

Consider Tristan Bilot's [benchmarking repo](https://github.com/TristanBilot/mlx-benchmark), for example: to benchmark ~12 MLX ops with a few input sizes each, he has to create a separate process for each op and shape combination and then garbage collect once the process has joined in order to prevent memory usage from blowing up past 25 GB, despite the fact that the single biggest memory allocation required in the benchmark, from what I've tested, is about 7 GB. 

Is MLX designed in such a way that would make it possible to have something like torch.mps.empty_cache()? Also, am I missing anything important conceptually here?

igm503:
[...]

PR #761, opened 2024-02-29
bindings for memory info

awni:
## Proposed changes

Add a few functions to the Python and C++ API to help with memory management:

-  `get_active_memory`: Query the currently used memory
-  `get_peak_memory`: Query the peak memory used so far
- `get_cache_memory`: Query the free cache size 
-  `set_memory_limit(int, bool)`: Set the memory limit (particularly useful for iOS where the device recommendation doesn't match the OS throttling limit).
- `set_cache_limit(int)`: Set the cache limit. When the free cache is bigger than this size, it gets partially purged.

I also removed `{enable, disable}_cache` since they can be achieved by setting the cache size

Related to #742

### DIFF
--- a/mlx/backend/metal/allocator.cpp
+++ b/mlx/backend/metal/allocator.cpp
@@ -1,7 +1,6 @@
-// Copyright © 2023 Apple Inc.
+// Copyright © 2023-2024 Apple Inc.
 
 #include "mlx/backend/metal/allocator.h"
-#include <iostream> // TODO
 #include "mlx/backend/metal/metal.h"
 
 #include <mach/vm_page_size.h>
@@ -150,16 +149,20 @@ MetalAllocator::MetalAllocator()
     : device_(device(mlx::core::Device::gpu).mtl_device()),
       buffer_cache_(device_),
       block_limit_(1.5 * device_->recommendedMaxWorkingSetSize()),
-      gc_limit_(0.95 * device_->recommendedMaxWorkingSetSize()) {}
+      gc_limit_(0.95 * device_->recommendedMaxWorkingSetSize()),
+      max_pool_size_(block_limit_) {}
 
-size_t MetalAllocator::set_gc_limit(size_t limit) {
-  std::swap(limit, gc_limit_);
+size_t MetalAllocator::set_cache_limit(size_t limit) {
+  std::swap(limit, max_pool_size_);
   return limit;
 };
 
 size_t MetalAllocator::set_memory_limit(size_t limit, bool relaxed) {
   std::swap(limit, block_limit_);
   relaxed_ = relaxed;
+  gc_limit_ = std::min(
+      block_limit_,
+      static_cast<size_t>(0.95 * device_->recommendedMaxWorkingSetSize()));
   return limit;
 };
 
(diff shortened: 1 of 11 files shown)


awni:
CC @davidkoski

mlx/backend/metal/allocator.cpp
```diff
@@ -1,6 +1,7 @@
 // Copyright © 2023 Apple Inc.
 
 #include "mlx/backend/metal/allocator.h"
+#include <iostream> // TODO
```
awni, line 4: 🖨️

mlx/backend/metal/allocator.cpp
```diff
@@ -158,10 +149,20 @@ void BufferCache::remove_from_list(BufferCache::BufferHolder* to_remove) {
 MetalAllocator::MetalAllocator()
     : device_(device(mlx::core::Device::gpu).mtl_device()),
       buffer_cache_(device_),
-      peak_allocated_size_(0),
       block_limit_(1.5 * device_->recommendedMaxWorkingSetSize()),
       gc_limit_(0.95 * device_->recommendedMaxWorkingSetSize()) {}
 
+size_t MetalAllocator::set_gc_limit(size_t limit) {
```
davidkoski, line 155: So gc_limit of 0 would be like `set_cache_enabled(false)`?  The gc limit controls how much it keeps in the buffer cache?

awni: Sort of yes. But I'm having some doubts about this part of the API, I think I should remove it as it's confusing.

Let me explain what I am going for in case you have any ideas though:

The allocator has two limits:
1. The memory limit. When this limit is reached, we wait for outstanding tasks to complete before allocating any more memory.
2. The garbage collection limit. Usually the allocator works by trying to reuse from the buffer cache. If it doesn't fine a fit then we will allocate more memory (without clearing the cache). But if we are over the gc limit then we will clear memory from the cache until we are under the gc limit.

The APIs correspond exactly to those two limits but I don't think that is the right way to expose it anymore. In part because the gc limit should always be <= the memory limit (it's not really sensible otherwise) so changing the memory limit should also change the gc limit.

What you are suggesting (e.g. making the gc limit control the size of the cache) is actually a reasonable alternative. I think I like it more than what I have now.

davidkoski: I think that sounds usable -- you want to tell it how much "idle" memory you can tolerate.  More might mean increased efficiency because allocating memory isn't free.  Less would mean smaller footprint, possibly at the cost of more frequent allocations.

In that case I wonder if you would want an allocation rate?  Number of buffers (and/or bytes) allocated between calls or something like that.  This would let you observe the behavior of tuning this parameter (along with the run time of your code, of course).

davidkoski: For the rate, if you make it between calls you put the sampling / window on the caller.

davidkoski: Along with the allocation rate, you might want the buffer pool reuse rate.

awni: All sounds very reasonable. I may play with that but possibly for a follow up. Prefer not to introduce too much to the API at once.

awni:
Running gemma 7b:
With no cache limit you can see the cache holds about 13.5G after 1k tokens
```
Memory 16430.702, Cache 13611.450
Generation: 27.164 tokens-per-sec
```

With a cache limit of 0 (`mx.metal.set_cache_llimit(0)`), the cache is.. well 0. So you are only usin about the model size worth of memory even after 1k tokens:

```
Memory 16432.681, Cache 0.000
Generation: 26.941 tokens-per-sec
```

awni:
I think this is good to be reviewed. There should be no perf impact since the defaults are all the same, but I will run a couple benchmarks to be sure.

davidkoski:
> ```
> Memory 16432.681, Cache 0.000
> Generation: 26.941 tokens-per-sec
> ```

The hit for no cache is surprisingly small.  Is this just not getting a lot of cache hits (normally) because the cache size keeps changing?

awni:
> The hit for no cache is surprisingly small. Is this just not getting a lot of cache hits (normally) because the cache size keeps changing?

I only eyeballed it so far, but it looks like the hit rate is ~50% for the kv cache when the cache is on and 100% for the query vector. There are a lot of moving pieces here though. I think the most likely explanation is simply that the rest of the model is more expensive than the time we spend allocating the arrays in the kv cache.

But it's due for some more careful measurement.. that's the next thing on my todo list.

davidkoski (approved):
I like it

PR #768, opened 2024-03-02
Fix the top-k op

angeloskath:
- Fixes the top-k operation to actually return the top `k` elements.
- Fixes the tests as well.
- Checks for `k` being positive and removes the functionality that `-k` would return `N-k` top elements (we can add that back, I don't think it is very standard and for sure it wasn't documented (or working)).


awni (approved):
Thanks for the fix!

PR #776, opened 2024-03-03
Impl dot

dc-dc-dc:
## Proposed changes

Adds dot implementation

## Checklist

### DIFF
--- a/mlx/ops.cpp
+++ b/mlx/ops.cpp
@@ -3294,10 +3294,7 @@ array dot(const array& a, const array& b, StreamOrDevice s /* = {} */) {
   if (a.ndim() == 0 || b.ndim() == 0) {
     return multiply(a, b, s);
   }
-  if (a.ndim() == 1 && b.ndim() == 1) {
-    return inner(a, b, s);
-  }
-  if (a.ndim() == 2 && b.ndim() == 2) {
+  if (a.ndim() <= 2 && b.ndim() <= 2) {
     return matmul(a, b, s);
   }
   if (b.ndim() == 1) {
--- a/mlx/ops.cpp
+++ b/mlx/ops.cpp
@@ -3298,13 +3298,13 @@ array dot(const array& a, const array& b, StreamOrDevice s /* = {} */) {
     return matmul(a, b, s);
   }
   if (b.ndim() == 1) {
-    return tensordot(a, b, {{-1}, {-1}}, s);
+    return tensordot(a, b, {-1}, {-1}, s);
   }
   if (a.shape(-1) != b.shape(-2)) {
     throw std::invalid_argument(
         "[dot] the last dimension of a must match the second to last dimension of b.");
   }
-  return tensordot(a, b, {{-1}, {-2}}, s);
+  return tensordot(a, b, {-1}, {-2}, s);
 }
 
 array outer(const array& a, const array& b, StreamOrDevice s /* = {} */) {
--- a/mlx/ops.cpp
+++ b/mlx/ops.cpp
@@ -3294,11 +3294,8 @@ array dot(const array& a, const array& b, StreamOrDevice s /* = {} */) {
   if (a.ndim() == 0 || b.ndim() == 0) {
     return multiply(a, b, s);
   }
-  if (a.ndim() <= 2 && b.ndim() <= 2) {
+  if (a.ndim() <= 2 && b.ndim() <= 2 || b.ndim() == 1) {
     return matmul(a, b, s);
-  }
-  if (b.ndim() == 1) {
-    return tensordot(a, b, {-1}, {-1}, s);
   }
   if (a.shape(-1) != b.shape(-2)) {
     throw std::invalid_argument(
--- a/mlx/ops.cpp
+++ b/mlx/ops.cpp
@@ -3290,6 +3290,26 @@ array tensordot(
   return reshape(matmul(x, y, s), rshape, s);
 }
 
+array dot(const array& a, const array& b, StreamOrDevice s /* = {} */) {
+  if (a.ndim() == 0 || b.ndim() == 0) {
+    return multiply(a, b, s);
+  }
+  if (a.ndim() == 1 && b.ndim() == 1) {
+    return inner(a, b, s);
+  }
+  if (a.ndim() == 2 && b.ndim() == 2) {
+    return matmul(a, b, s);
+  }
+  if (b.ndim() == 1) {
+    return tensordot(a, b, {{-1}, {-1}}, s);
+  }
+  if (a.shape(-1) != b.shape(-2)) {
+    throw std::invalid_argument(
+        "[dot] the last dimension of a must match the second to last dimension of b.");
+  }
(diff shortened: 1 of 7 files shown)


python/src/ops.cpp
```diff
@@ -3811,4 +3811,25 @@ void init_ops(py::module_& m) {
         Returns:
             array or list(array): An array or list of arrays with at least three dimensions.
         )pbdoc");
+
+  m.def(
+      "dot",
+      &dot,
+      "a"_a,
+      "b"_a,
+      py::pos_only(),
```
awni, line 3820:
```suggestion
```

python/src/ops.cpp
```diff
@@ -3811,4 +3811,25 @@ void init_ops(py::module_& m) {
...
+      py::kw_only(),
+      "stream"_a = none,
+      R"pbdoc(
+      dot(a: array, b: array, /, *, stream: Union[None, Stream, Device] = None) -> array
```
awni, line 3824:
```suggestion
      dot(a: array, b: array, *, stream: Union[None, Stream, Device] = None) -> array
```

python/src/ops.cpp
```diff
@@ -3811,4 +3811,25 @@ void init_ops(py::module_& m) {
...
+
+      Dot product of two arrays.
+
```
awni, line 3827: Can you elaborate on the docstring here and/or provide an `Example:` section below?

mlx/ops.cpp
```diff
@@ -3273,6 +3273,26 @@ array tensordot(
...
+    return matmul(a, b, s);
+  }
+  if (b.ndim() == 1) {
+    return tensordot(a, b, {{-1}, {-1}}, s);
+  }
+  if (a.shape(-1) != b.shape(-2)) {
+    throw std::invalid_argument(
+        "[dot] the last dimension of a must match the second to last dimension of b.");
+  }
+  return tensordot(a, b, {{-1}, {-2}}, s);
```
awni, line 3293: Note, we change the C++ interface to tensor dot (to make it easier to bind to C). It now takes two vectors instead of a single pair of vectors.

mlx/ops.cpp
```diff
@@ -3273,6 +3273,26 @@ array tensordot(
...
+array dot(const array& a, const array& b, StreamOrDevice s /* = {} */) {
+  if (a.ndim() == 0 || b.ndim() == 0) {
+    return multiply(a, b, s);
+  }
+  if (a.ndim() == 1 && b.ndim() == 1) {
+    return inner(a, b, s);
+  }
+  if (a.ndim() == 2 && b.ndim() == 2) {
+    return matmul(a, b, s);
+  }
```
awni, line 3285:
```suggestion
  if (a.ndim() <= 2 && b.ndim() <= 2) {
    return matmul(a, b, s);
  }
```

mlx/ops.cpp
```diff
@@ -3273,6 +3273,26 @@ array tensordot(
...
+  if (b.ndim() == 1) {
+    return tensordot(a, b, {{-1}, {-1}}, s);
+  }
```
awni, line 3288: Do you want a shape check here as well like you have below?

awni:
Thanks for adding this! Left some comments inline. Please take a look and let me know when it's ready to go 🚀

tests/ops_tests.cpp
```diff
@@ -2878,3 +2878,21 @@ TEST_CASE("test topk") {
...
+  CHECK_EQ(dot(array(3), array(4)).item<int>(), 12);
+  auto x = array({1., 0., 0., 1.}, {2, 2});
+  auto y = array({4., 1., 2., 2.}, {2, 2});
+  auto z = dot(x, y);
+  auto expected = array({4., 1., 2., 2.}, {2, 2});
+  CHECK(array_equal(z, expected).item<bool>());
+  x = reshape(arange(10.), {5, 2});
+  y = reshape(arange(8.), {2, 2, 2});
+  z = dot(x, y);
+  MESSAGE("z: " << z);
```
awni, line 2892: Just wondering, what does that do?

mlx/ops.cpp
dc-dc-dc: tensordot does a shape check, can add it here or leave it to tensordot to check and remove the other check under dot before calling dot.

The only side effect is that the error message will show `[tensordot]` instead of dot if leaving it to tensordot to do the check

tests/ops_tests.cpp
dc-dc-dc: useful for debugging 😁, removing

Stealeristaken (approved):

angeloskath (approved):
This looks good to me. Thanks for the addition.

angeloskath:
@awni I 'll wait if you want to comment on the docs but otherwise I think it is ok.

python/src/ops.cpp
```diff
@@ -3811,4 +3811,41 @@ void init_ops(py::module_& m) {
...
+      "a"_a,
+      "b"_a,
+      py::kw_only(),
+      "stream"_a = none,
+      R"pbdoc(
+      dot(a: array, b: array, *, stream: Union[None, Stream, Device] = None) -> array
+
+      Computes the dot product of the two arrays.
+
+      Even though this function follows the numpy API and is defined for
```
awni, line 3827:
```suggestion
      Even though this function follows the NumPy API and is defined for
```

awni (approved):
Looks great! I think for the doc it might be helpful to say why we discourage using it? It left me a bit confused reading that line .. like why have the op if we don't want people to use it sort of thing

dc-dc-dc:
> Looks great! I think for the doc it might be helpful to say why we discourage using it? It left me a bit confused reading that line .. like why have the op if we don't want people to use it sort of thing

Going to remove that comment, this pr was created a while back and looking through it I'm not sure why I added that comment.

awni:
@angeloskath what if we just repurpose this to a `vecdot`/`vdot` and scrap the rest of the `dot` API? I'm really ok not including it as I agree with you the API is confusing and it's always better to use one of the alternatives.

Also I just noticed it's not in the [Python array API standard](https://data-apis.org/array-api/latest/API_specification/linear_algebra_functions.html) so from that perspective we can at least support the decision not to include it.

I don't mean to blow this up though, if you all think we should have it, let's do it.

angeloskath:
I definitely support moving this to `vdot`. Wdyt @dc-dc-dc ? Is there a specific need for `dot`? Some old codebase or something?

dc-dc-dc:
Somebody in the discord mentioned that they needed it for their project and I had this sitting in my local repo. I'm fine with closing this if it's not of interest

awni:
Regarding the need, I remember the discussion and that person was satisfied using `matmul` in the end.. so I don't think there is a strong need for `dot`?

How about just changing to `vdot` / `vecdot`? That one actually seems useful IMO. Should be a small(ish) diff from here.

awni:
@dc-dc-dc wdyt about this? Any plan to update or should we close?

awni:
Closing until we have a good vdot.

PR #777, opened 2024-03-03
Some fixes in cache / thread safety

awni:
## Proposed changes

- Change the cache limit so it's more closely respected. On `free` we don't add to the cache if we are already above the limit
- Moved locks from buffer to malloc / free are more thread safe. No noticeable regression as we were already acquiring the locks on anyway when querying the buffer cache.

Adding something like below to MLX LM generate script effectively halts the cache growth without any degradation to TPS. It's not the prettiest, but seems to be pretty effective.

```python
    def set_cache_limit():
        _, cache = model(mx.array([[0]]))
        cache_size = max_tokens * sum(c.nbytes for _, c in tree_flatten(cache))
        cache_size += 10 * 2**20  # buffer
        mx.metal.set_cache_limit(cache_size)

    set_cache_limit()
```

### Benchmarks:

No cache limit:

```
Generation: 59.588 tokens-per-sec
Memory Peak: 1822.406 Cache: 9105.971
```

Cache disabled:
```
Generation: 53.861 tokens-per-sec
Memory Peak: 1822.146 Cache: 0.000
```

Cache limited to max kv-cache size:

```
Generation: 58.017 tokens-per-sec
Memory Peak: 1821.762 Cache: 948.803
```

### DIFF
--- a/mlx/backend/metal/allocator.cpp
+++ b/mlx/backend/metal/allocator.cpp
@@ -170,7 +170,7 @@ Buffer MetalAllocator::malloc(size_t size, bool allow_swap /* = false */) {
   }
 
   // Try the cache
-  std::lock_guard<std::mutex> lk(mutex_);
+  std::unique_lock lk(mutex_);
   MTL::Buffer* buf = buffer_cache_.reuse_from_cache(size);
   if (!buf) {
     size_t mem_required = get_active_memory() + get_cache_memory() + size;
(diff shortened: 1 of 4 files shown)


mlx/backend/metal/allocator.cpp
```diff
@@ -190,36 +182,41 @@ Buffer MetalAllocator::malloc(size_t size, bool allow_swap /* = false */) {
...
 
 void MetalAllocator::free(Buffer buffer) {
   auto buf = static_cast<MTL::Buffer*>(buffer.ptr());
+  std::unique_lock lk(mutex_);
   active_memory_ -= buf->length();
-  if (max_pool_size_ > 0) {
+  if (get_cache_memory() < max_pool_size_) {
     buffer_cache_.recycle_to_cache(buf);
   } else {
+    lk.unlock();
```
awni, line 218: Interestingly - with the cache disabled, unlocking here makes a big difference in tokens per second. It seems that holding the mutex and releasing / allocating buffers from the device do not combine well.. 🤔 .

angeloskath (approved):
Interesting. So there is still some overhead from not letting the cache grow. I am not sure why... calling release I guess.

awni:
> So there is still some overhead from not letting the cache grow

Indeed there is a smidge. It merits further investigation, I don't know what causes it yet

awni:
I'm going to merge this since I think the race condition is causing our tests to fail from time to time. I also piggy backed another test fix for our optimizer to stop throwing the return warning.

PR #778, opened 2024-03-04
revision bump

awni:
## Proposed changes

This is a small revision to get the new memory controls out for swift to unblock iOS examples.


awni:
CC @davidkoski

angeloskath (approved):

PR #782, opened 2024-03-04
Nicer tensordot for mlx c

awni:
## Proposed changes

Make it two args instead of a pair of vectors.


andresy (approved):
LGTM, thank you!

PR #786, opened 2024-03-05
Improved quantized matrix vector product

angeloskath:
This PR changes the `qmv` kernel. Adds an aligned version and an unaligned version but both are overhauled and improved. 

Further improvements should be possible but for now the following table shows tokens per second when generating with Mistral 7B before and after this PR (keep in mind naive 2 bit is not viable for Mistral 7B):

```
                | Before | After
-------------------------------
float16         |  34.2  |  -
gs=32  bits=2   |  44.6  | 83.1
gs=32  bits=4   |  51.4  | 70.7
gs=64  bits=4   |  54.4  | 74.1
gs=32  bits=8   |  39.8  | 52.6
gs=64  bits=8   |  40.7  | 54.4
gs=128 bits=8   |  40.2  | 55.2

```

### DIFF
--- a/mlx/backend/metal/quantized.cpp
+++ b/mlx/backend/metal/quantized.cpp
@@ -41,8 +41,35 @@ void QuantizedMatmul::eval_gpu(const std::vector<array>& inputs, array& out) {
   int B = x.size() / D;
   int O = out.shape(-1);
   if (transpose_) {
+    // Route to the specialized qmv kernel
+    if (B < 6 && group_size_ == 32 && bits_ == 4 && O % 32 == 0) {
+      std::ostringstream kname;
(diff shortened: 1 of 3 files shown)


mlx/backend/metal/quantized.cpp
```diff
@@ -41,8 +41,35 @@ void QuantizedMatmul::eval_gpu(const std::vector<array>& inputs, array& out) {
   int B = x.size() / D;
   int O = out.shape(-1);
```
jagrit06, line 42: Do we want to add a check here (or in the op) to make sure that the input / output are `float` ? We can land up with an unclear error message otherwise

jagrit06 (approved):
Look great to me down to the profiling!

awni (approved):
🚀 🚀

PR #787, opened 2024-03-05
Improve testing of reduction ops

nicolov:
## Proposed changes

Parameterize the ops tests to test both cpu and gpu. This in preparation for other improvements to the CPU path.

## Checklist

### DIFF
--- a/tests/ops_tests.cpp
+++ b/tests/ops_tests.cpp
@@ -4,6 +4,7 @@
 
 #include "doctest/doctest.h"
 
+#include "mlx/device.h"
 #include "mlx/mlx.h"
 
 using namespace mlx::core;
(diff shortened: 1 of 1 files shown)


awni (changes requested):
I think it would be better to have a flag set outside the tests (or in the main test driver) to toggle cpu or gpu. It makes it easy to switch all the tests from the CPU to the GPU. For Python we have a `DEVICE={cpu, gpu}` environment variable that sets the default device. Do you want to do something similar for our C++ tests?

Also, in some cases we don't want to run on a given device (for example when testing on linux) so we have to be sure it's configurable.

awni (changes requested):
I think it would be better to have a flag set outside the tests (or in the main test driver) to toggle cpu or gpu. It makes it easy to switch all the tests from the CPU to the GPU. For Python we have a `DEVICE={cpu, gpu}` environment variable that sets the default device. Do you want to do something similar for our C++ tests?

Also, in some cases we don't want to run on a given device (for example when testing on linux) so we have to be sure it's configurable.

nicolov:
This is actually already done: https://github.com/ml-explore/mlx/blob/main/tests/tests.cpp so I'll close this PR.

PR #788, opened 2024-03-05
checkpoint module's __call__

awni:
## Proposed changes

Change `nn.utils` to checkpoint the module's `__call__` directly instead of an arbitrary function.

Let's us do stuff like:

```python
self.layers = [nn.utils.checkpoint(l) for l in self.layers]
```

- Added checkpointing to the docs
- Added a few tests to make sure the checkpoint module has the right attributes etc.

### DIFF
--- a/python/mlx/nn/utils.py
+++ b/python/mlx/nn/utils.py
@@ -37,25 +37,23 @@ def value_and_grad(model: Module, fn: Callable):
     return wrapped_value_grad_fn
 
 
-def checkpoint(module: Module, fn: Callable = None):
-    """Transform the passed callable to one that performs gradient
-    checkpointing with respect to the trainable parameters of the module (and
-    the callable's inputs).
+def checkpoint(module: Module):
+    """Transform the passed module to one that performs gradient
+    checkpointing.
+
+    The checkpointing is with respect to the module's trainable parameters and
+    inputs of the module's ``__call__`` function.
 
     Args:
         module (mlx.nn.Module): The module for whose parameters we will be
             performing gradient checkpointing.
-        fn (Callable, optional): The function to checkpoint. If not provided it
-            defaults to the provided module.
 
     Returns:
-        A callable that saves the inputs and outputs during the forward pass
+        The module that saves the inputs and outputs during the forward pass
         and recomputes all intermediate states during the backward pass.
     """
-    if fn is None:
-        # Capturing module instead of module.__call__ allows someone to
-        # monkey-patch __call__ later on and the correct method will be used
-        fn = module
+
+    fn = module.__call__
 
     def inner_fn(params, *args, **kwargs):
         module.update(params)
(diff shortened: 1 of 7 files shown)


awni:
Moving back to draft until we figure out what to do here. Making a new class at each call to checkpoint is way too slow.

awni:
The latest version works pretty well, it makes a new checkpointed class for every *unique* incoming class. It is not that elegant so I am ok not merging it and/or refining parts that are sticky. But just to show that this way can be done if we want to keep pushing.

Benchmarks:

### This PR
```
Training a transformer with 153.883 M parameters
Iter 10: Train loss 8.825, It/sec 4.382, Peak memory 4.240 GB
Iter 20: Train loss 8.139, It/sec 4.607, Peak memory 4.240 GB
Iter 30: Train loss 7.839, It/sec 4.611, Peak memory 4.240 GB
Iter 40: Train loss 7.672, It/sec 4.610, Peak memory 4.240 GB
```

### Old version

```shell
Training a transformer with 153.883 M parameters
Iter 10: Train loss 8.829, It/sec 4.383, Peak memory 4.240 GB
Iter 20: Train loss 8.265, It/sec 4.605, Peak memory 4.240 GB
Iter 30: Train loss 8.033, It/sec 4.610, Peak memory 4.240 GB
```

awni:
I'm kind of ok with this version (modulo comments).. we could mark the API as somewhat experimental, so we can iterate without worrying about back-comaptibility. I find it more usable than what we had before.

For completeness, other directions to consider:

- Decorating call instead and toggling with some kind of flag (as discussed with @angeloskath).
- Keep the original version and just use a different method name and/or arguments to module

python/mlx/nn/utils.py
```diff
@@ -37,34 +37,46 @@ def wrapped_value_grad_fn(*args, **kwargs):
     return wrapped_value_grad_fn
 
 
-def checkpoint(module: Module, fn: Callable = None):
```
angeloskath, line 40: There is a lot of python magic going on here but with a second look I think it is fine. Ie not that hidden/hard to reason about.

The only thing is that it is not entirely clear to me what happens when someone keeps around both modules and starts messing with them. Would we expect changes in one to be reflected to the other?

angeloskath (approved):
I think this is mergeable. I would maybe add a note as you suggested that this is slightly experimental so the API may change.

python/mlx/nn/utils.py
awni: > The only thing is that it is not entirely clear to me what happens when someone keeps around both modules and starts messing with them. Would we expect changes in one to be reflected to the other?

I think the behavior to get across is that it makes a *new* module and because of that you should not see changes reflected in the old one. Now I'm only 90% certain that's the case :), but if it's not it is a bug and I should add a test / fix it.

awni:
I'm closing this PR for now and working on a variation of functions in `nn.utils` instead.. the solution here is a bit too hacky / too much magic IMO

PR #791, opened 2024-03-06
Fix reshaping of empty arrays

### LINKED ISSUE #789: should not be able to reshape a zero size array
**Describe the bug**

```
>>> a = mx.zeros([1, 0, 32, 64])
>>> a.shape
(1, 0, 32, 64)
>>> a.reshape(1, 0, -1).shape
(1, 0, -1)
```

We should not end up with a -1 in any dimension -- we probably shouldn't be able to reshape a zero size array.  Numpy reports:

```
>>> b.reshape(1, 0, -1)
ValueError: cannot reshape array of size 0 into shape (1,0,newaxis)
```

for a similar case.

**To Reproduce**

See above.

**Expected behavior**

Probably an error thrown.

 - OS Version: [e.g. MacOS 14.1.2]
 - Version [e.g. 0.7.0] 0.5.0

**Additional context**

angeloskath:
Simply fixes #789 .

```python
import mlx.core as mx
import numpy as np

x = mx.random.normal((100, 0, 1024))
y = np.random.randn(100, 0, 1024)

x.reshape(2, 50, 0).shape  # (2, 50, 0)
y.reshape(2, 50, 0).shape  # (2, 50, 0)

x.reshape(2, 50, -1).shape  # (2, 50, 0)
y.reshape(2, 50, -1).shape  # (2, 50, 0)

x.reshape(2, 50, 0, -1).shape  # throws ValueError: Cannot infer the shape of an empty array
y.reshape(2, 50, 0, -1).shape  # throws ValueError: cannot reshape array of size 0 into shape (2,50,0,newaxis)
```

### DIFF
--- a/mlx/ops.cpp
+++ b/mlx/ops.cpp
@@ -300,19 +300,27 @@ array reshape(
       size *= shape[i];
     }
   }
+
+  // Infer the shape
   if (size > 0) {
     auto q_and_r = std::ldiv(a.size(), size);
     if (infer_idx >= 0) {
       shape[infer_idx] = q_and_r.quot;
       size *= q_and_r.quot;
     }
+  } else if (infer_idx >= 0) {
+    std::ostringstream msg;
+    throw std::invalid_argument("Cannot infer the shape of an empty array");
   }
+
+  // Check the the reshaping is valid
   if (a.size() != size) {
     std::ostringstream msg;
     msg << "Cannot reshape array of size " << a.size() << " into shape "
         << shape << ".";
     throw std::invalid_argument(msg.str());
   }
+
   return array(
       shape, a.dtype(), std::make_unique<Reshape>(to_stream(s), shape), {a});
 }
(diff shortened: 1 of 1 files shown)


mlx/ops.cpp
```diff
@@ -300,19 +300,27 @@ array reshape(
...
+  // Infer the shape
   if (size > 0) {
     auto q_and_r = std::ldiv(a.size(), size);
     if (infer_idx >= 0) {
       shape[infer_idx] = q_and_r.quot;
       size *= q_and_r.quot;
     }
+  } else if (infer_idx >= 0) {
+    std::ostringstream msg;
+    throw std::invalid_argument("Cannot infer the shape of an empty array");
```
awni, line 313: Nitty nit:

```suggestion
    throw std::invalid_argument("[reshape] Cannot infer the shape of an empty array");
```

awni (approved):
Thanks for the fix!!

mlx/ops.cpp
angeloskath: Yeah and also forgot the `ostringstream`. Let me fix all these...

PR #268, opened 2023-12-23
Implement RNN, GRU, LSTM

jdeschena:
## Proposed changes

Implement recurrent cells and layers (Elman RNN, GRU, LSTM) in Python. Ultimately, it would probably be more efficient to implement in metal for parallelization (esp. for multi-layer and bi-directional), but in the meantime I think it is worth having a simple, be it only for benchmarking.

I have tested the implementations on small character-level language models.

## Checklist
- [x] I have read the [CONTRIBUTING](https://github.com/ml-explore/mlx/blob/main/CONTRIBUTING.md) document
- [x] I have run `pre-commit run --all-files` to format my code / installed pre-commit prior to committing changes
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have updated the necessary documentation (if needed)

PR #792, opened 2024-03-06
Adding LSTMCell

Rifur13:
## Proposed changes

This PR adds LSTMCell. It represents a single cell of an LSTM layer, and processes one time step of the sequence data at a time.

## Checklist

### DIFF
--- a/python/mlx/nn/layers/__init__.py
+++ b/python/mlx/nn/layers/__init__.py
@@ -62,6 +62,7 @@ from mlx.nn.layers.normalization import (
 from mlx.nn.layers.pooling import AvgPool1d, AvgPool2d, MaxPool1d, MaxPool2d
 from mlx.nn.layers.positional_encoding import ALiBi, RoPE, SinusoidalPositionalEncoding
 from mlx.nn.layers.quantized import QuantizedLinear
+from mlx.nn.layers.recurrent import LSTMCell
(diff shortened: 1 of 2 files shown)


Rifur13:
I need this layer for a model I’m working and it would be nice to merge it here. I see the open PR to add LSTM, but it has a lot more functionality on top and won’t share code. 

Let me know what you think before I clean it up and start writing docs and tests.

awni:
We don't intend to have the concept of "Cell" for recurrent nets in general. One should be able to use the LSTM in #268 for the same purpose.  In fact #268 originally had cells like that and we decided to remove them for simplicity.

If you aren't able to use #268 that would be useful to understand why

Rifur13:
Ok I see thanks, let’s close this PR then.

In my case I can use a LSTM and just call it one time step at a time.

FWIW my reasons for adding this were:
- Better readability because I would expect LSTM to be applied for all time steps.
- The input to LSTM will have the shape (time_steps, input_size) or (batch, time_steps, input_size), so I would avoid reshaping from (batch, input_size) to (batch, 1, input_size)

awni:
Makes sense, thanks for the explanation. The LSTM can indeed be used on a per-time-step basis. The reshaping is maybe not so nice, but I think that's something we could fix in that layer.

PR #793, opened 2024-03-06
Remove code duplication in reduce ops

nicolov:
## Proposed changes

Switch the reducer operations to a template to avoid duplicating the loop logic for min/max/sum.

The change is covered by existing tests.

## Checklist

### DIFF
--- a/mlx/backend/accelerate/reduce.cpp
+++ b/mlx/backend/accelerate/reduce.cpp
@@ -10,78 +10,65 @@
 
 namespace mlx::core {
 
-template <typename T, typename VT, int N>
-void _vectorized_strided_sum(const T* x, T* accum, int size, size_t stride) {
-  for (int i = 0; i < size; i++) {
-    size_t s = stride;
-    T* a = accum;
-    while (s >= N) {
-      VT val = (*(VT*)x);
-      *(VT*)a += val;
-      x += N;
-      a += N;
-      s -= N;
-    }
-    while (s-- > 0) {
-      *a++ += *x++;
-    }
-  }
-}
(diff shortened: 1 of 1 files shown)


angeloskath (changes requested):
I think this is a good change but I have a proposal to make that will simplify more. Instead of passing the signature, just wrap it with an `Op` and pass that to a `StridedReduce` struct.

```cpp
template <typename T, typename VT>
struct Sum {
  T operator()(T a, T b) {
    return a + b;
  }
  VT operator()(VT a, VT b) {
    return a + b;
  }
}

template <typename T, typename Op>
struct StridedReduce {
  void operator()(const T* x, T * accum, int size, size_t stride) {
    Op op;
    ...
  }
}
```

I think it is way more readable and removes the `_vectorized_strided_max`, `*_min`, `*_sum` functions. Furthermore, if you feel like improving this further, then I would propose to do a `ContiguousReduce` as well (or in the same struct, whatever you prefer) and then we can implement vectorized reductions for many more types like ints etc. We can even have specializations for the operations with arm neon for 16 bit floats or even bfloats.

nicolov:
> I think this is a good change but I have a proposal to make that will simplify more.

Thank you very much! This is indeed much better!

angeloskath (approved):
I think we can merge this as an intermediate step, but I believe the compiler is doing a pretty good job of autovectorizing our default implementations (since we moved to Xcode >= 15) and I think this whole file can probably be removed.

On my M2 air I see no difference between float32 and int32 with the one calling into accelerate and the other not. I will do more checks on other chips as well.

PR #797, opened 2024-03-06
Add a maximum graph depth

awni:
## Proposed changes

Basically just gives a more helpful error message when graphs get too large. Previously they would stack overflow at around 105k, so I set the limit to 100k for now. We can revisit the limit if it makes sense.


angeloskath (approved):
Very nice!

PR #802, opened 2024-03-07
Dynamic number of elements computation for shapeless compile

angeloskath:
Puts the computation of the number of elements across some axes in the graph to allow for more shapeless compiles (basically shapeless `mean` and `var`).

This is a draft to discuss whether this kind of solution makes sense. I think instead we could have a "shape function" that would return a double that would be casted to whatever type is the scalar array. This would allow computing the number of elements or ddof correction factor in one go.


awni:
>  I think instead we could have a "shape function" that would return a double that would be casted to whatever type is the scalar array. This would allow computing the number of elements or ddof correction factor in one go.

Could you elaborate what you mean by that? Is it essentially the same as this but fixing the output type to double?

I think what you have is fine for now. It is pretty specific to this case.. but it's not exposed anywhere so we can update it if we need to. The more general solution doesn't feel like it would be better here tbh...I don't love the idea of blowing up the graph with all those tiny scalar ops, unless we always compile and they get fused.

Maybe a possibility to make the more general solution better is finding a clean way to do it but only if we are compiling the function and otherwise just using the shape directly..

mlx/ops.cpp
```diff
@@ -3563,4 +3558,28 @@ std::vector<array> atleast_3d(
...
+          << " from an array with " << a.ndim() << " dimensions.";
+      throw std::invalid_argument(msg.str());
+    }
+    ax = normal_axis;
+  }
+
+  return stop_gradient(array(
+      std::vector<int>{},
+      dtype,
+      std::make_unique<NumberOfElements>(to_stream(s), axes, inverted, dtype),
```
awni, line 3581: Minor but you set the class up nicely to `std::move(axes)` here and in the constructor as well. Might as well take advantage of it :)

mlx/primitives.h
```diff
@@ -1229,6 +1229,37 @@ class NotEqual : public UnaryPrimitive {
...
 
+class NumberOfElements : public UnaryPrimitive {
+ public:
+  explicit NumberOfElements(
+      Stream stream,
+      std::vector<int> axes,
+      bool inverted,
+      Dtype dtype)
+      : UnaryPrimitive(stream),
+        axes_(axes),
```
awni, line 1240:
```suggestion
        axes_(std::move(axes)),
```

awni (approved):
Looks good! Feel free to merge this and we can follow on fix the vmap (it's not so urgent IMO). I just saw that `-1` and it ocurred to me that it might break things in `vmap`

angeloskath:
@awni feel free to check the vmaps and I 'll merge after

awni:
Awesome, thanks for patching that up! vmap fixes look great. I pushed the test I had for the mean vmap that was failing. Feel free to merge whenever

PR #803, opened 2024-03-07
Slight speed improvement for LLM inference

youknow04:
## Proposed changes

This PR introduces modifications to the `qmv` kernel aimed at enhancing the inference speed of quantized LLM.
The improvement, albeit modest (approximately 0.1%)

### Considerations: Speed Versus Readability
Maintainer's decision is required.
By unrolling a key computational loop, we observe a slight acceleration in performance; however, this comes at the cost of increased code complexity.

### Personal Stance
I advocate for prioritizing performance in kernel code, refusing to compromise on speed.
To address potential concerns regarding code complexity, I have employed template metaprogramming.
This approach ensures that all unrolled codes are generated at compile time, no runtime overhead, while providing flexibility of code.

### Performance Benchmarks
Machine: M3MAX (128GB)
LLM: https://huggingface.co/TheBloke/Mistral-7B-Instruct-v0.2-GGUF
Client: https://github.com/ml-explore/mlx-examples/tree/main/llms/gguf_llm
Prompt: "Write a quicksort in Python"
Before: https://github.com/ml-explore/mlx/pull/786.
After: this PR

Token Processing Speed (tokens per second, higher is better, all values are median of 5 runs):
| Bits | Before | After |
|------|-----------|----------|
| 4    | 29.747    | 29.753   |
| 8    | 24.960    | 25.033   |

### Additional Notes
- `uint16_t` was uniformly applied across different bit configurations (Q2, Q4, Q8).
- All tests passed.
- Applying `clang-format` to `quantized.metal` resulted in significant formatting changes. To avoid overwhelming diffs, I opted to bypass the linting step for this file.

### DIFF
--- a/mlx/backend/metal/kernels/quantized.metal
+++ b/mlx/backend/metal/kernels/quantized.metal
@@ -31,12 +31,16 @@ inline U load_vector(const device T *x, thread U *x_thread) {
   U sum = 0;
 
   if (bits == 2) {
-    for (int i = 0; i < values_per_thread; i += 4) {
-      sum += x[i] + x[i+1] + x[i+2] + x[i+3];
+    for (int i = 0; i < values_per_thread; i += 8) {
+      sum += x[i] + x[i+1] + x[i+2] + x[i+3] + x[i+4] + x[i+5] + x[i+6] + x[i+7];
       x_thread[i] = x[i];
       x_thread[i+1] = x[i+1] / 4.0f;
       x_thread[i+2] = x[i+2] / 16.0f;
       x_thread[i+3] = x[i+3] / 64.0f;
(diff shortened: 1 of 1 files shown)


angeloskath:
I appreciate the attempt here. There is definitely room for improvement in our `qmv` but I am inclined to close this. The template unrolling is not necessary as the compiler will automatically unroll constant loops and readability takes a hit. The performance improvement (although I certainly appreciate the median of 5 runs) is so close to the noise ceiling that I can't be sure it is there.

Btw I really appreciate this and I hope you keep at it. I also really appreciate your template kung-fu it is very handy for reducing code reuse (but not for unrolling here).

PR #807, opened 2024-03-07
Version bump and os error

awni:
## Proposed changes

🚀

### DIFF
--- a/python/mlx/_os_warning.py
+++ b/python/mlx/_os_warning.py
@@ -1,0 +1,6 @@
+import platform
+
+if platform.system() == "Darwin":
+    major, minor, _ = map(int, platform.mac_ver()[0].split("."))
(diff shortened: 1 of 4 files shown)


python/mlx/_os_warning.py
```diff
@@ -0,0 +1,6 @@
+import platform
+
+if platform.system() == "Darwin":
+    major, minor, _ = map(int, platform.mac_ver()[0].split("."))
+    if (major, minor) < (13, 5):
+        raise OSError(f"Only macOS 13.5 and newer are supported, not {major}.{minor}")
```
angeloskath, line 6: Is `OSError` the right one here? I would use `ImportError`. Either way it's fine.

angeloskath (approved):
🚀

PR #809, opened 2024-03-08
Add SVD primitive

nicolov:
## Proposed changes

Add SVD op using Accelerate's LAPACK following
https://developer.apple.com/documentation/accelerate/ compressing_an_image_using_linear_algebra

## Checklist


python/src/linalg.cpp
```diff
@@ -210,4 +210,27 @@ void init_linalg(py::module_& parent_module) {
...
+  m.def(
+      "svd",
+      &svd,
+      "a"_a,
+      py::kw_only(),
+      "stream"_a = none,
+      R"pbdoc(
+        svd(a: array, *, stream: Union[None, Stream, Device] = None) -> (array, array, array)
+
+        The SVD decomposition of the input matrix.
```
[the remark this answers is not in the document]
nicolov: Done, thanks.

python/src/linalg.cpp
```diff
@@ -210,4 +210,28 @@ void init_linalg(py::module_& parent_module) {
...
+        has more than two dimensions, the function operates in "stacked mode" like
+        numpy.
+
+        Args:
+            a (array): Input array.
+            stream (Stream, optional): Stream or device. Defaults to ``None``
+              in which case the default stream of the default device is used.
+
+        Returns:
+            tuple(array, array, array): The ``U``, ```S```, and ``Vt`` matrices, such that
```
awni, line 234:
```suggestion
            tuple(array, array, array): The ``U``, ``S``, and ``Vt`` matrices, such that
```

python/src/linalg.cpp
```diff
@@ -210,4 +210,28 @@ void init_linalg(py::module_& parent_module) {
...
+              A = U @ diag(S) @ Vt
```
awni, line 235:
```suggestion
              ``A = U @ diag(S) @ Vt``
```

mlx/linalg.cpp
```diff
@@ -200,4 +200,46 @@ std::pair<array, array> qr(const array& a, StreamOrDevice s /* = {} */) {
   return std::make_pair(out[0], out[1]);
 }
 
+std::tuple<array, array, array> svd(
+    const array& a,
+    StreamOrDevice s /* = {} */) {
+  if (a.dtype() != float32) {
+    std::ostringstream msg;
+    msg << "[linalg::svd] Arrays must type float32. Received array "
```
awni, line 208:
```suggestion
    msg << "[linalg::svd] Input array must have type float32. Received array "
```

mlx/backend/metal/primitives.cpp
```diff
@@ -890,4 +890,10 @@ void QRF::eval_gpu(
   throw std::runtime_error("[QRF::eval_gpu] Metal QR factorization NYI.");
 }
 
+void SVD::eval_gpu(
+    const std::vector<array>& inputs,
+    std::vector<array>& outputs) {
+  throw std::runtime_error("[SVD::eval_gpu] Metal SVD decomposition NYI.");
```
awni, line 896:
```suggestion
  throw std::runtime_error("[SVD::eval_gpu] Metal SVD NYI.");
```

mlx/backend/common/svd.cpp
```diff
@@ -0,0 +1,149 @@
...
+        // According to the identity above, lapack will write Uᵀ as Vᵀ.
+        /* vt = */ u.data<float>() + M * M * i,
+        /* ldvt = */ &ldvt,
+        /* work = */ static_cast<float*>(scratch.raw_ptr()),
+        /* lwork = */ &lwork,
+        /* iwork = */ iwork.data(),
+        /* info = */ &info);
+
+    if (info != 0) {
+      throw std::runtime_error("svd_impl: sgesvdx_ failed.");
```
awni, line 128: I would also log the error code here, may be useful to help us debug issues.

mlx/backend/common/svd.cpp
```diff
@@ -0,0 +1,149 @@
...
+  const int ldu = N;
+  // Vᵀ of shape N x N. (M x M in lapack).
+  const int ldvt = M;
+
+  size_t num_matrices = a.size() / (M * N);
+
+  // lapack clobbers the input, so we have to make a copy.
+  array in(a.shape(), float32, nullptr, {});
+  in.set_data(allocator::malloc_or_wait(in.nbytes()));
+  copy_inplace(a, in, CopyType::GeneralGeneral);
```
awni, line 41: Maybe this should be a vector copy when `a` is row contiguous? That would be a lot faster.

You can check with `a.flags().row_contiguous`

mlx/linalg.h
```diff
@@ -62,4 +62,6 @@ norm(const array& a, int axis, bool keepdims = false, StreamOrDevice s = {}) {
 
 std::pair<array, array> qr(const array& a, StreamOrDevice s = {});
 
+std::tuple<array, array, array> svd(const array& a, StreamOrDevice s = {});
```
awni, line 65: I think this output type is going to cause issues for MLX C (CC @andresy ). Could you make it a `std::vector<array>` instead?

mlx/backend/common/svd.cpp
```diff
@@ -0,0 +1,149 @@
+// Copyright © 2023-2024 Apple Inc.
```
awni, line 1:
```suggestion
// Copyright © 2024 Apple Inc.
```

mlx/linalg.h
nicolov: Done, though it's not as ergonomic because unpacking stops working.

mlx/backend/common/svd.cpp
nicolov: Agree on the concept, but I think this should be handled in the framework? If you show me how, I can make the copy function itself smarter so that primitives don't need to worry about the input memory layout.

awni: It's a good point. We kind of drew the line in the copy API a little bit lower because we needed some more control over it in a lot of cases. There are some slight differences here and there in terms of needs for the type of copy.

Here is the [copy function](https://github.com/ml-explore/mlx/blob/main/mlx/backend/common/copy.cpp#L288) which dispatches based on copy type.

I think you want to use:

- `CopyType::Vector`  if `in.flags().row_contiguous`
- `CopyType::General` otherwise (not `GeneralGeneral` since the output of the copy is always contiguous.

And you could probably use `copy` instead of `copy_inplace` and skip dealing with the allocation.

mlx/backend/common/svd.cpp
```diff
@@ -0,0 +1,148 @@
...
+      /* info = */ &info);
+
+  if (info != 0) {
+    std::stringstream ss;
+    ss << "svd_impl: sgesvdx_ workspace calculation failed with code " << info;
+    throw std::runtime_error(ss.str());
+  }
+
+  const int lwork = workspace_dimension;
+  auto scratch = allocator::malloc_or_wait(sizeof(float) * lwork);
```
awni, line 93: A bit of a nit, but maybe a good idea to put this in a unique pointer so it doesn't leak in the event of an exception (will also make this code a bit more future proof).

mlx/linalg.cpp
```diff
@@ -200,4 +200,42 @@ std::pair<array, array> qr(const array& a, StreamOrDevice s /* = {} */) {
...
+    std::ostringstream msg;
+    msg << "[linalg::svd] Input array must have >= 2 dimensions. Received array "
+           "with "
+        << a.ndim() << " dimensions.";
+    throw std::invalid_argument(msg.str());
+  }
+
+  const auto m = a.shape(-2);
+  const auto n = a.shape(-1);
+  const auto rank = a.shape().size();
```
awni, line 220:
```suggestion
  const auto rank = a.ndim();
```

python/src/linalg.cpp
```diff
@@ -210,4 +210,28 @@ void init_linalg(py::module_& parent_module) {
...
+      py::kw_only(),
+      "stream"_a = none,
+      R"pbdoc(
+        svd(a: array, *, stream: Union[None, Stream, Device] = None) -> (array, array, array)
+
+        The Singular Value Decomposition (SVD) of the input matrix.
+
+        This function supports arrays with at least 2 dimensions. When the input
+        has more than two dimensions, the function operates in "stacked mode" like
+        numpy.
```
awni, line 226:
```suggestion
        NumPy.
```

awni: Could you elaborate here on what is meant by "stacked mode"? I think it would be clearer if it said it treats the last two dimensions as the matrix dimensions and computes the SVD for each matrix independently.

python/src/linalg.cpp
```diff
@@ -210,4 +210,28 @@ void init_linalg(py::module_& parent_module) {
...
+
+        Args:
+            a (array): Input array.
+            stream (Stream, optional): Stream or device. Defaults to ``None``
+              in which case the default stream of the default device is used.
+
+        Returns:
+            tuple(array, array, array): The ``U``, ``S``, and ``Vt`` matrices, such that
```
awni, line 234: Is it actually a `tuple` or is it a `list`?

awni: If it's a list (which I think is what pybind11 will cast the vector to),  I think it would be good to make it a `tuple` to stay consistent with NumPy.

awni, line 234: I notice you call it `Vt` instead of `Vh` as in the numpy docs. Do you know what the `t` (or `h`) is intended to stand for? Just curious..

mlx/backend/common/svd.cpp
```diff
@@ -0,0 +1,148 @@
...
+  static constexpr auto job_u = "V";
+  static constexpr auto job_vt = "V";
+  static constexpr auto range = "A";
+
+  // Will contain the number of singular values after the call has returned.
+  int ns = 0;
+  float workspace_dimension = 0;
+
+  std::vector<int> iwork;
+  iwork.resize(12 * K);
```
awni, line 51: Mind adding a comment on the 12 here?

awni (commented):
Looks really nice! I left a few more minor comments. After that I think this should be good to go!

mlx/backend/common/svd.cpp
nicolov: Yes, of course! I just copy-pasted from other code and I thought there was a reason behind it.

awni: I realize I gave you a bad suggestion. You need something with a custom deleter. Actually we have an object you could use for this: https://github.com/ml-explore/mlx/blob/main/mlx/array.h#L203-L214

I think if you do:
```suggestion
  auto scratch = array::Data{allocator::malloc_or_wait(sizeof(float) * lwork)};
```

And then add a `.buffer` before casting the pointer below it should be good.

> I just copy-pasted from other code and I thought there was a reason behind it.

No reason other than negligence ;). If you feel like cleaning that up elsewhere that would be welcome.

nicolov: Yes, I was about to say that `Buffer` and `Allocator` don't play nice with `unique_ptr`. I was gonna create a `ScopedBuffer` with a destructor, but this is also good.

python/src/linalg.cpp
nicolov: I used the same term as the NumPy docs:

> When a is higher-dimensional, SVD is applied in stacked mode as explained below.

I rephrased it anyway.

python/src/linalg.cpp
nicolov: Yes sorry, I overlooked this when changing the C++ signature from `tuple` to `vector`.

nicolov: Both lapack and NumPy return the last term in the decomposition (`V`) as its transpose. To be precise, it's not just the transpose, but the hermitian conjugate (or conjugate transpose, see https://en.wikipedia.org/wiki/Conjugate_transpose). The difference only matters for complex numbers, as for real numbers transpose and conjugate transpose are identical.

Since we only support float32 for now, I've used `t` to avoid confusion (and match lapack signatures).

awni (approved):
Awesome, thanks for the contribution! I will merge once the tests clear.

PR #812, opened 2024-03-08
Compile stride bug

awni:
## Proposed changes

- Fix bug in compile with strides + donation
- Reverts #805

### DIFF
--- a/mlx/backend/metal/compiled.cpp
+++ b/mlx/backend/metal/compiled.cpp
@@ -329,7 +329,9 @@ void Compiled::eval_gpu(
       if (in.flags().row_contiguous && in.nbytes() == outputs[o].nbytes() &&
           in.is_donatable() &&
           constant_ids_.find(inputs_[i].id()) == constant_ids_.end()) {
-        outputs[o++].move_shared_buffer(in);
+        outputs[o].move_shared_buffer(
+            in, outputs[o].strides(), in.flags(), in.data_size());
+        o++;
       }
     }
     for (; o < outputs.size(); ++o) {
--- a/mlx/array.cpp
+++ b/mlx/array.cpp
@@ -164,12 +164,23 @@ void array::copy_shared_buffer(const array& other) {
   copy_shared_buffer(other, other.strides(), other.flags(), other.data_size());
 }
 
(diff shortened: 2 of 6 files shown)


mlx/backend/metal/compiled.cpp
```diff
@@ -329,7 +329,9 @@ void Compiled::eval_gpu(
       if (in.flags().row_contiguous && in.nbytes() == outputs[o].nbytes() &&
           in.is_donatable() &&
           constant_ids_.find(inputs_[i].id()) == constant_ids_.end()) {
-        outputs[o++].move_shared_buffer(in);
+        outputs[o].move_shared_buffer(
+            in, outputs[o].strides(), in.flags(), in.data_size());
```
angeloskath, line 333: Just to verify my understanding. The only way `output[o].strides()` is not the same as `in.strides()` is if the output has a bunch of unsqueezed dimensions in the beginning right? Otherwise the input would not be `row_contiguous` or `in.nbytes() != outputs[o].nbytes()`.

angeloskath (approved):
This looks like an extremely difficult bug to track down...

mlx/backend/metal/compiled.cpp
awni: Yea that's right. I'm pretty sure it's the only case.

python/tests/test_compile.py
```diff
@@ -605,6 +605,14 @@ def fun(x, y):
         with self.assertRaises(ValueError):
             out = fun(mx.array(0.0), y=MyClass())
 
+    def test_compile_create_list(self):
+        @mx.compile
+        def fun():
+            return [0.1 * mx.zeros((2,)), 0.1 * mx.zeros((2,))]
+
+        out = fun()
+        mx.eval(out)
```
awni, line 614: @angeloskath piggy backing another bug fix here if you want to see. This is actually an issue with simplify merging outputs when it shouldn't be (so the parents map gets invalid which then causes fusion to fuse in cases that it shouldn't). Another difficult one to track down 😅 .

angeloskath: Wow! Very nice!

PR #818, opened 2024-03-12
Reshape improvement

angeloskath:
Change reshape so it only copies when necessary. For example the following series of ops doesn't result in a copy

```python
x = mx.random.normal(shape=(10, 20, 30))
x = x.reshape(10, 20, 6, 5).transpose(0, 2, 1, 3).reshape(10, 2, 3, 20, 5)
```

Simple algorithm, basically use the collapse contiguous dims (that function keeps on giving :-) ) to collapse the input and then try to fit the output dims in the collapsed dims. If they don't fit exactly then it needs a copy. It also checks for contiguity to reset the flags.


jagrit06 (approved):
LGTM!

awni:
I'd like to run a couple regression tests with this before we merge just to be sure. Its probably right but we may have been assuming elsewhere that reshape does a copy so useful to stress test a bit IMO.

awni:
- [x] Mnist CPU and GPU
- [x] Transformer GPU
- [x] Cifar CPU/GPU
- [x] Lora with Mistral (200 iters)
- [x] Generation with Mistral

awni:
Some benchmarks:

Bench | Post | Pre
----|----|----
MNIST 2 CPU | 0.333 (s) | 0.322 (s)
MNIST 2 GPU | 0.596 (s) | 0.596 (s)
MNIST 10 CPU | 0.642 (s) | 0.634 (s)
MNIST 10 GPU | 1.117 (s) | 1.102 (s)
MNIST 100 CPU |  3.930 (s) | 3.895 (s)
MNIST 100 GPU |  6.806 (s) | 6.847 (s)
Prompt | 504.449 | 498.683
Generatrion | 31.423 | 29.004
Prompt q | 447.283  | 355.295
Generation q | 74.230 | 74.877
Transformer 2 | It/sec 5.703 | It/sec 5.75
Transformer 4 | It/sec 3.053 | It/sec 3.06
kwt1 | 4873.74 samples/second |4820.78 samples/second
kwt2 | 2747.42 samples/second | 2750.43 samples/second
kwt3 | 1978.75 samples/second | 1975.19 samples/second

awni (approved):
Nice change, thanks!!

PR #819, opened 2024-03-12
Fix modules with dict

awni:
## Proposed changes

As title. Closes #817


angeloskath (approved):
Damn... slipped through.

PR #821, opened 2024-03-12
fix scatter

awni:
## Proposed changes

As title. Should also close https://github.com/ml-explore/mlx-examples/issues/557


angeloskath (approved):
👍

PR #822, opened 2024-03-12
Add matrix inversion primitive

nicolov:
## Proposed changes

Add `mx.linalg.inv` for matrix inversion.

## Checklist

### DIFF
--- a/mlx/backend/common/inverse.cpp
+++ b/mlx/backend/common/inverse.cpp
@@ -1,0 +1,95 @@
+// Copyright © 2023-2024 Apple Inc.
+
+#include "mlx/allocator.h"
+#include "mlx/backend/common/copy.h"
+#include "mlx/primitives.h"
+
+#ifdef ACCELERATE_NEW_LAPACK
+#include <Accelerate/Accelerate.h>
+#else
+#include <lapack.h>
+#endif
+
+namespace mlx::core {
+
+void inverse_impl(const array& a, array& inv) {
+  // Lapack uses the column-major convention. We take advantage of the following
+  // identity to avoid transposing (see
+  // https://math.stackexchange.com/a/340234):
+  //   (A⁻¹)ᵀ = (Aᵀ)⁻¹
+
+  // The inverse is computed in place, so just copy the input to the output.
+  copy(a, inv, a.flags().row_contiguous ? CopyType::Vector : CopyType::General);
+
+  const int N = a.shape(-1);
+  const size_t num_matrices = a.size() / (N * N);
+
+  int info;
+  auto ipiv = array::Data{allocator::malloc_or_wait(sizeof(int) * N)};
+
+  for (int i = 0; i < num_matrices; i++) {
+    // Compute LU factorization.
+    sgetrf_(
+        /* m = */ &N,
+        /* n = */ &N,
+        /* a = */ inv.data<float>() + N * N * i,
+        /* lda = */ &N,
+        /* ipiv = */ static_cast<int*>(ipiv.buffer.raw_ptr()),
+        /* info = */ &info);
+
+    if (info != 0) {
+      std::stringstream ss;
+      ss << "inverse_impl: LU factorization failed with error code " << info;
+      throw std::runtime_error(ss.str());
+    }
+
+    static const int lwork_query = -1;
+    float workspace_size = 0;
+
+    // Compute workspace size.
+    sgetri_(
+        /* m = */ &N,
+        /* a = */ nullptr,
+        /* lda = */ &N,
+        /* ipiv = */ nullptr,
+        /* work = */ &workspace_size,
+        /* lwork = */ &lwork_query,
+        /* info = */ &info);
+
+    if (info != 0) {
+      std::stringstream ss;
+      ss << "inverse_impl: LU workspace calculation failed with error code "
+         << info;
+      throw std::runtime_error(ss.str());
+    }
+
+    const int lwork = workspace_size;
+    auto scratch =
+        array::Data{allocator::malloc_or_wait(sizeof(float) * lwork)};
+
+    // Compute inverse.
+    sgetri_(
+        /* m = */ &N,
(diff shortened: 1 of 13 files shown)


python/src/linalg.cpp
```diff
@@ -241,4 +241,28 @@ void init_linalg(py::module_& parent_module) {
...
+  m.def(
+      "inv",
+      &inv,
+      "a"_a,
+      py::kw_only(),
+      "stream"_a = none,
+      R"pbdoc(
+        inv(a: array, *, stream: Union[None, Stream, Device] = None) -> array
+
+        Compute the (multiplicative) inverse of a (square) matrix.
```
awni, line 253:
```suggestion
        Compute the inverse of a square matrix.
```

Maybe just me but the multiplicative seems quite clear.. you could add it back if you disagree.

python/src/linalg.cpp
```diff
@@ -241,4 +241,28 @@ void init_linalg(py::module_& parent_module) {
...
+
+        This function supports arrays with at least 2 dimensions. When the input
+        has more than two dimensions, the function iterates over all indices of the first
+        a.ndim - 2 dimensions and the inverse is calculated for the matrix described
+        by the last two indices.
```
awni, line 258:
```suggestion
        This function supports arrays with at least 2 dimensions. When the input
        has more than two dimensions, the inverse is computed for each matrix
        in the last two dimensions of ``a``.
```

python/src/linalg.cpp
```diff
@@ -241,4 +241,28 @@ void init_linalg(py::module_& parent_module) {
...
+
+        Args:
+            a (array): Input array.
+            stream (Stream, optional): Stream or device. Defaults to ``None``
+              in which case the default stream of the default device is used.
+
+        Returns:
+            array: ```ainv``` such that ```dot(a, ainv) = dot(ainv, a) = eye(a.shape[0])```
```
awni, line 266:
```suggestion
            array: ``ainv`` such that ``dot(a, ainv) = dot(ainv, a) = eye(a.shape[0])``
```

mlx/primitives.h
```diff
@@ -1856,4 +1856,18 @@ class SVD : public Primitive {
   void eval(const std::vector<array>& inputs, std::vector<array>& outputs);
 };
 
+/* Matrix inversion primitive. */
+class Inverse : public UnaryPrimitive {
+ public:
+  explicit Inverse(Stream stream) : UnaryPrimitive(stream){};
+
+  void eval_cpu(const std::vector<array>& inputs, array& output) override;
+  void eval_gpu(const std::vector<array>& inputs, array& output) override;
```
awni, line 1865: Btw I feel like we should be implementing vmap for these primitives.. I didn't catch it for QR.. but it's actually not too bad to implement. You could move the vmapped axis to any axis other than the last two.

mlx/backend/common/inverse.cpp
```diff
@@ -0,0 +1,95 @@
...
+#include <lapack.h>
+#endif
+
+namespace mlx::core {
+
+void inverse_impl(const array& a, array& inv) {
+  // Lapack uses the column-major convention. We take advantage of the following
+  // identity to avoid transposing (see
+  // https://math.stackexchange.com/a/340234):
+  //   (A⁻¹)ᵀ = (Aᵀ)⁻¹
```
awni, line 19: 👍

mlx/backend/common/inverse.cpp
```diff
@@ -0,0 +1,95 @@
...
+
+  // The inverse is computed in place, so just copy the input to the output.
+  copy(a, inv, a.flags().row_contiguous ? CopyType::Vector : CopyType::General);
+
+  const int N = a.shape(-1);
+  const size_t num_matrices = a.size() / (N * N);
+
+  int info;
+  std::vector<int> ipiv;
+  ipiv.resize(N);
```
awni, line 29: I noticed you did this in SVD and on second thought, I think it would be better to not use a std::vector for this pattern. It kind of subverts our allocator and so we wouldn't get a good read on e.g. peak memory use (or able to cache the allocation for repeat applications).

awni: Maybe instead use a scratch allocation

mlx/backend/common/inverse.cpp
```diff
@@ -0,0 +1,95 @@
...
+    if (info != 0) {
+      std::stringstream ss;
+      ss << "inverse_impl: inversion failed with error code " << info;
+      throw std::runtime_error(ss.str());
+    }
+  }
+}
+
+void Inverse::eval(const std::vector<array>& inputs, array& output) {
+  if (!(inputs[0].dtype() == float32)) {
```
awni, line 89:
```suggestion
  if (inputs[0].dtype() != float32) {
```

awni (commented):
Looks really good! Thanks for the addition. I left a few minor comments please take a look. I will also run the tests for an initial read.

mlx/primitives.h
nicolov: Can I do it for both on a follow-up?

awni: Of course!

mlx/backend/common/inverse.cpp
nicolov: Yes, of course! I just did like it this for RAII, but now that you showed `Buffer::Data`, I'll just use that instead.

mlx/backend/common/svd.cpp
```diff
@@ -49,8 +49,7 @@ void svd_impl(const array& a, array& u, array& s, array& vt) {
 
   // Will contain the indices of eigenvectors that failed to converge (not used
   // here but required by lapack).
-  std::vector<int> iwork;
-  iwork.resize(12 * K);
+  auto iwork = array::Data{allocator::malloc_or_wait(sizeof(int) * 12 * K)};
```
awni, line 52: Thanks for updating it here as well!

awni (approved):
Looks really nice, thank you!!

PR #829, opened 2024-03-13
Make reshape faster for row_contiguous cases

angeloskath:
Remove unnecessary computation for row contiguous reshapes.


awni (approved):
🚀🚀

PR #838, opened 2024-03-15
No reshape rope

awni:
## Proposed changes

Gives  a small bump on TPS to avoid the two reshapes:

On M2 Ultra (1)

| | Prompt | Generation | Memory
--- | --- | --- | ---
Pre | 430.603 | 73.586 | 4.108
Post | 430.470 | 75.072 | 4.108


angeloskath (approved):
Looks great. Simpler from python and faster.

awni:
I was going to try and squeeze a proper ND implementation but for the sake of not letting this stagnate I will land it and follow up with that.

PR #839, opened 2024-03-16
Switch to nanobind

awni:
- [x] clear regular test
- [x] clear regression tets
- [x] benchmarks
- [x] fix signatures
- [x] fix stubs
- [x] update build develop docs
- [ ] update extension + docs


awni:
Mnist benchmark (no compilation)
```
Task | Pre | Post
2 MNIST CPU | 0.446 | 0.418 
2 MNIST GPU | 0.919 | 0.802
10 MNIST CPU | 0.839 | 0.800
10 MNIST GPU | 1.967 | 1.866
100 MNSIT CPU | 5.508 | 5.063
100 MNIST GPU | 12.811 | 12.208
```

wjakob:
A quick thought: try adding `NOMINSIZE` to the `nanobind_add_module` call to see if this impacts the benchmarks you posted above.

awni:
Indeed it is slightly faster with `NOMINSIZE` and hardly effects our overall binary size so well worth it IMO. Thanks for the tip!

CMakeLists.txt
```diff
@@ -146,8 +146,12 @@ target_include_directories(
 
 if (MLX_BUILD_PYTHON_BINDINGS)
   message(STATUS "Building Python bindings.")
-  find_package(Python COMPONENTS Interpreter Development)
-  find_package(pybind11 CONFIG REQUIRED)
+  find_package(Python 3.8 COMPONENTS Interpreter Development.Module REQUIRED)
```
jagrit06, line 149: Maybe the python version should be a range here ?

python/src/buffer.h
```diff
@@ -0,0 +1,122 @@
...
+
+  view->obj = obj;
+  view->ndim = a.ndim();
+  view->internal = info;
+  view->buf = a.data<void>();
+  view->itemsize = a.itemsize();
+  view->len = a.size();
+  view->readonly = false;
+  if ((flags & PyBUF_FORMAT) == PyBUF_FORMAT) {
+    view->format = const_cast<char*>(info->format.c_str());
```
angeloskath, line 110: Since we are not relying on pybind now to fill out the CPython buffer struct, shouldn't that be a null pointer? Is the null terminated string the better way of doing it?

angeloskath: This is the relevant docs that I assume you also used https://docs.python.org/3/c-api/buffer.html#c.Py_buffer.format . Btw using "B" seems to be an equivalent and equally safe choice.

python/src/transforms.cpp
```diff
@@ -793,57 +797,52 @@ void init_transforms(py::module_& m) {
...
-            doc << ' ';
-          }
-          doc << dstr;
-        }
-        auto doc_str = doc.str();
-        return py::cpp_function(
-            PyCompiledFun{fun, inputs, outputs, shapeless},
-            py::name(name.c_str()),
-            py::doc(doc_str.c_str()));
+        /*        std::ostringstream doc;
```
angeloskath, line 808: I guess this means we leave the docs for later?

angeloskath (approved):
This is awesome! Very tedious work, thanks for doing it!

CMakeLists.txt
awni: Good point, I didn't realize this but I checked and it is actually a range (>= 3.8).

python/src/transforms.cpp
awni: No, I should have a fix for that shortly, seems to mostly work just debugging some strange behavior

python/src/array.cpp
```diff
@@ -769,35 +597,34 @@ void init_array(py::module_& m) {
...
             }
-            return np_array_to_mlx(npa, std::nullopt);
-          }))
+            return mlx_to_np_array(a);
+          })
+      .def(
+          "__setstate__",
+          [](array& arr,
+             const nb::ndarray<nb::ro, nb::c_contig, nb::device::cpu>& state) {
+            new (&arr) array(nd_array_to_mlx(state, std::nullopt));
```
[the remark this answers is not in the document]
awni: Yea, in fact it's not really an option, you have to do it this way in nanobind for `__setstate__` and `__init__` if you don't bind a constructor directly.

python/src/buffer.h
awni: I'm not 100pc sure I follow your question, but we have to set the format string correctly for the different types: https://github.com/python/cpython/blob/f6cdc6b4a191b75027de342aa8b5d344fb31313e/Include/typeslots.h#L2-L3

The only case I'm not sure about is bfloat.. I have it as `\0` right now but in fact maybe it should be `B` or `nullptr`.

angeloskath: In Pybind11 we weren't filling the `Py_buffer` struct ourselves which meant we had to pass a `std::string` for pybind to use.

My interpretation of the docs is that we should either set it to `nullptr` (which will be interpreted as `"B"`) or set it to `"B"`.

My question is basically, where did the `"\0"` come from? Is that what we should be doing?

Oh and reading back all 3 messages, your confusion makes sense. All this should be in the context of bfloat16 which I failed to mention anywhere.

awni: > My interpretation of the docs is that we should either set it to nullptr (which will be interpreted as "B") or set it to "B".

Yes I think you're right about that. I will update it.

jagrit06 (approved):
Look great! Thanks for taking the time!

wjakob:
One last thing: I recommend setting an `NB_DOMAIN` (e.g. "apple" or "mlx") in the call to `nanobind_add_module`. That isolates you from other nanobind-based extensions, which is probably a safe default.

awni:
Thanks for the tip, it's already set as `mlx`.

wjakob:
Great, I missed that 👍

PR #841, opened 2024-03-16
Some C++ code are not needed

zcbenz:
## Proposed changes

1. Anonymous namespace means internal linkage, static keyword is not needed.
2. The default constructor of std::shared_ptr initializes the pointer to nullptr, you don't need to explicitly set it.

## Checklist


awni (approved):
Thanks!

PR #836, opened 2024-03-14
vmap matmul and admm

awni:
As title. Closes #834

PR #849, opened 2024-03-19
Add vmap for SVD and inverse and fix some other ops

nicolov:
## Proposed changes

Add vmap for SVD and inverse, following https://github.com/ml-explore/mlx/pull/836

Also fix a bug in the `vmap` implementation of some binary ops, that was causing an issue on Linux (but not on Mac, probably because of different initial sizing and/or initialization of `std::vector`).

## Checklist

### DIFF
--- a/mlx/backend/common/inverse.cpp
+++ b/mlx/backend/common/inverse.cpp
@@ -2,6 +2,7 @@
 
 #include "mlx/allocator.h"
 #include "mlx/backend/common/copy.h"
+#include "mlx/linalg.h"
 #include "mlx/primitives.h"
 
 #ifdef ACCELERATE_NEW_LAPACK
@@ -90,6 +91,14 @@ void Inverse::eval(const std::vector<array>& inputs, array& output) {
     throw std::runtime_error("[Inverse::eval] only supports float32.");
   }
   inverse_impl(inputs[0], output);
+}
+
+std::pair<std::vector<array>, std::vector<int>> Inverse::vmap(
+    const std::vector<array>& inputs,
+    const std::vector<int>& axes) {
+  auto ax = axes[0] >= 0 ? 0 : -1;
+  auto a = axes[0] > 0 ? moveaxis(inputs[0], axes[0], 0, stream()) : inputs[0];
+  return {{linalg::inv(a, stream())}, {ax}};
 }
 
 } // namespace mlx::core
--- a/mlx/backend/common/svd.cpp
+++ b/mlx/backend/common/svd.cpp
@@ -3,6 +3,7 @@
 #include "mlx/allocator.h"
 #include "mlx/backend/common/copy.h"
 #include "mlx/backend/common/lapack_helper.h"
+#include "mlx/linalg.h"
 #include "mlx/primitives.h"
 
 namespace mlx::core {
@@ -142,6 +143,14 @@ void SVD::eval(const std::vector<array>& inputs, std::vector<array>& outputs) {
     throw std::runtime_error("[SVD::eval] only supports float32.");
   }
   svd_impl(inputs[0], outputs[0], outputs[1], outputs[2]);
+}
+
+std::pair<std::vector<array>, std::vector<int>> SVD::vmap(
+    const std::vector<array>& inputs,
+    const std::vector<int>& axes) {
+  auto ax = axes[0] >= 0 ? 0 : -1;
+  auto a = axes[0] > 0 ? moveaxis(inputs[0], axes[0], 0, stream()) : inputs[0];
+  return {{linalg::svd(a, stream())}, {ax, ax, ax}};
 }
 
 } // namespace mlx::core
--- a/mlx/primitives.cpp
+++ b/mlx/primitives.cpp
@@ -1127,7 +1127,7 @@ std::pair<std::vector<array>, std::vector<int>> Equal::vmap(
     const std::vector<array>& inputs,
     const std::vector<int>& axes) {
   auto [a, b, to_ax] = vmap_binary_op(inputs, axes, stream());
-  return {{equal(a, b, stream())}, axes};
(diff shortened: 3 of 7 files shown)


mlx/backend/common/inverse.cpp
```diff
@@ -92,4 +93,14 @@ void Inverse::eval(const std::vector<array>& inputs, array& output) {
   inverse_impl(inputs[0], output);
 }
 
+std::pair<std::vector<array>, std::vector<int>> Inverse::vmap(
+    const std::vector<array>& inputs,
+    const std::vector<int>& axes) {
+  auto maybe_move_ax = [this](auto& arr, auto ax) {
+    return ax > 0 ? moveaxis(arr, ax, 0, stream()) : arr;
+  };
+  auto a = maybe_move_ax(inputs[0], axes[0]);
```
awni, line 102: Nit: I don't think you need a lambda for this, just use the ternary directly?

mlx/backend/common/svd.cpp
```diff
@@ -144,4 +145,14 @@ void SVD::eval(const std::vector<array>& inputs, std::vector<array>& outputs) {
   svd_impl(inputs[0], outputs[0], outputs[1], outputs[2]);
 }
 
+std::pair<std::vector<array>, std::vector<int>> SVD::vmap(
+    const std::vector<array>& inputs,
+    const std::vector<int>& axes) {
+  auto maybe_move_ax = [this](auto& arr, auto ax) {
+    return ax > 0 ? moveaxis(arr, ax, 0, stream()) : arr;
+  };
+  auto a = maybe_move_ax(inputs[0], axes[0]);
```
awni, line 154: Same comment as above.

awni (commented):
It looks great! Minor nits inline.

awni:
@nicolov the tests are failing.. could you take a look?

mlx/backend/common/inverse.cpp
```diff
@@ -92,4 +93,11 @@ void Inverse::eval(const std::vector<array>& inputs, array& output) {
   inverse_impl(inputs[0], output);
 }
 
+std::pair<std::vector<array>, std::vector<int>> Inverse::vmap(
+    const std::vector<array>& inputs,
+    const std::vector<int>& axes) {
+  auto a = axes[0] > 0 ? moveaxis(inputs[0], axes[0], 0, stream()) : inputs[0];
+  return {{linalg::inv(a, stream())}, {0}};
```
angeloskath, line 100: For inputs that are not vmapped (`axes[0] == -1`) this should also return `{-1}` for the vectorized axis. Something like the following should take care of it.

```cpp
auto ax = axes[0] >= 0 ? 0 : -1;
auto a = axes[0] > 0 ? moveaxis(inputs[0], axes[0], 0, stream()) : inputs[0];
return {{linalg::inv(a, stream())}, {ax}};
```

mlx/backend/common/svd.cpp
```diff
@@ -144,4 +145,11 @@ void SVD::eval(const std::vector<array>& inputs, std::vector<array>& outputs) {
   svd_impl(inputs[0], outputs[0], outputs[1], outputs[2]);
 }
 
+std::pair<std::vector<array>, std::vector<int>> SVD::vmap(
+    const std::vector<array>& inputs,
+    const std::vector<int>& axes) {
+  auto a = axes[0] > 0 ? moveaxis(inputs[0], axes[0], 0, stream()) : inputs[0];
+  return {{linalg::svd(a, stream())}, {0}};
```
angeloskath, line 152: Same comment as above.

angeloskath (changes requested):
This looks great!

I left a comment regarding handling non vectorized axes. The linux test seems to fail on the vmap test and the Mac test seems unrelated and possibly a fluke. I launched both tests again but it is unlikely that the linux one will clear.

mlx/backend/common/inverse.cpp
awni: Good catch!

nicolov:
I'm looking into it. Linux (not Mac) calls `moveaxis` with a big axis number in `mlx::core::detail::vmap_replace`. @angeloskath's suggestion does not fix the issue sadly.

angeloskath:
Yeah indeed I also saw that (that's weird and I may look into it later if you haven't fixed it by then). My suggestion has to do with the general correctness for cases where the input is not vmapped. So it wouldn't fix that...

nicolov:
> My suggestion has to do with the general correctness for cases where the input is not vmapped.

But in those cases `vmap` wouldn't be called at all, no?

mlx/primitives.cpp
```diff
@@ -1127,7 +1127,7 @@ std::pair<std::vector<array>, std::vector<int>> Equal::vmap(
     const std::vector<array>& inputs,
     const std::vector<int>& axes) {
   auto [a, b, to_ax] = vmap_binary_op(inputs, axes, stream());
-  return {{equal(a, b, stream())}, axes};
```
awni, line 1130: Thanks for debugging that!

mlx/transforms.cpp
```diff
@@ -655,6 +656,11 @@ std::vector<array> vmap_replace(
     }
 
     auto [v_outputs, v_out_axes] = a.primitive().vmap(v_inputs, v_axes);
+
+    assert(
+        "Expected the number of outputs and vmapped axes to match. Please fix "
+        "the vmap implementation for this primitive.");
```
awni, line 662: I'm not sure what is the purpose of this loc, maybe remove?

angeloskath: I think this was `assert(v_outputs.size() == v_out_axes.size(), "...")` but I kind of agree that we should remove it.

angeloskath (approved):

angeloskath (changes requested):
Let's merge after removing the assert.

PR #850, opened 2024-03-19
Add a SliceUpdate op and primitive

jagrit06:
## Proposed changes

* Add a slice update op and primitive to copy over strided slices without needing to go to scatter 

## Checklist


jagrit06:
Also adds the option for copy kernels on the cpu and GPU to work with negative strides making flipping axes much faster

Some benchmarks for operation `y = x[..., ::-1]`

Before:
```
Shape: (4, 4096, 4096), tpose: (0, 1, 2), Time(s): 16.811, Bandwidth(GB/s):  237.942
```

After:
```
Shape: (4, 4096, 4096), tpose: (0, 1, 2), Time(s): 4.856, Bandwidth(GB/s):  823.713
```

mlx/backend/common/primitives.cpp
```diff
@@ -694,7 +691,87 @@ void Slice::eval(const std::vector<array>& inputs, array& out) {
...
+  if (upd.size() == 0) {
+    out.copy_shared_buffer(in);
+    return;
+  }
+
+  // Check if materialization is needed
+  auto ctype = in.flags().contiguous && in.size() == in.data_size()
+      ? CopyType::Vector
+      : CopyType::General;
+  copy(in, out, in.data_size() == 1 ? CopyType::Scalar : ctype);
```
angeloskath, line 759: Should there be a donation check here? It is very likely that we can avoid this copy.

mlx/backend/metal/primitives.cpp
```diff
@@ -865,7 +865,73 @@ void Sqrt::eval_gpu(const std::vector<array>& inputs, array& out) {
...
+  auto& in = inputs[0];
+  auto& upd = inputs[1];
+
+  if (upd.size() == 0) {
+    out.copy_shared_buffer(in);
+    return;
+  }
+
+  // Check if materialization is needed
+  auto ctype = in.flags().contiguous && in.size() == in.data_size()
```
angeloskath, line 915: Same as above. This copy can probably be avoided if the input is donatable.

mlx/primitives.cpp
```diff
@@ -2849,6 +2849,86 @@ bool Slice::is_equivalent(const Primitive& other) const {
       end_indices_ == s_other.end_indices_ && strides_ == s_other.strides_);
 }
 
+std::pair<std::vector<array>, std::vector<int>> SliceUpdate::vmap(
```
angeloskath, line 2852: This is great but it assumes that only `inputs[0]` is vmapped. For instance the following pseudocode should work but it won't in this case

```
def foo(x, y):
    x[:5] = y
    return x

x = mx.zeros((2, 10))
y = mx.ones((2, 5))
x = mx.vmap(foo)(x, y)
print(x)
array([[1, 1, 1, 1, 1, 0, 0, 0, 0, 0],
       [1, 1, 1, 1, 1, 0, 0, 0, 0, 0]])
```

It is also not very clear to me what is the correct behavior here when only 1 of the inputs is vmapped. Your implementation is assuming that the update was defined for one input and broadcasts it appropriately. I think this is not correct, I would just do nothing to the other input if it was not broadcasted.

Also if both are vmapped then we need to make sure that the vectorized dimensions are in the correct place so I would transpose the update vectorized axis to `ax` if they are different.

wdyt?

angeloskath (changes requested):
This looks fantastic! I like the refactoring a lot and the new capability!

I left a few comments regarding donation which I think would be good to add and a comment on the `vmap` implementation. Let me know what you think.

mlx/backend/common/primitives.cpp
jagrit06: Copy does the donation check the we pass the type as vector - that's why the check doesn't look for row or column contiguity when directing to the vector copy

mlx/backend/metal/primitives.cpp
jagrit06: Same thing, copy_gpu handles the donation when copy type vector is passed

mlx/utils.cpp
```diff
@@ -329,4 +329,13 @@ std::ostream& operator<<(std::ostream& os, const std::vector<size_t>& v) {
   return os;
 }
 
+std::ostream& operator<<(std::ostream& os, const std::vector<int64_t>& v) {
```
awni, line 332: Probably time to templatize these.. though I'm not really sure we need this function here.. I guess you added it for debugging?

mlx/ops.cpp
```diff
@@ -582,6 +536,56 @@ array slice(
...
+
+/** Update a slice from the source array with stride 1 in each dimension */
+array slice_update(
+    const array& src,
+    const array& update,
+    std::vector<int> start,
+    std::vector<int> stop,
+    StreamOrDevice s /* = {} */) {
+  auto strides = std::vector<int>(src.ndim(), 1);
+  return slice_update(src, update, start, stop, strides, s);
```
awni, line 586: I think you want to use `std::move` here on `start`, `stop`, `strides`

mlx/ops.cpp
```diff
@@ -445,127 +445,81 @@ array expand_dims(
...
-    std::vector<int> strides,
-    StreamOrDevice s /* = {} */) {
-  if (start.size() != a.ndim() || stop.size() != a.ndim() ||
-      strides.size() != a.ndim()) {
-    std::ostringstream msg;
-    msg << "[slice] Invalid number of indices or strides for "
-        << "array with dimension " << a.ndim() << ".";
-    throw std::invalid_argument(msg.str());
-  }
+    std::vector<int> strides) {
```
awni, line 455: Why not make these references and update them in place. I think they are already copied into the `slice` and `slice_update`.

mlx/ops.cpp
```diff
@@ -445,127 +445,81 @@ array expand_dims(
...
-            std::move(start),
-            std::move(stop),
-            std::move(strides)),
-        {a});
-
-    std::vector<array> indices;
-    std::vector<int> slice_sizes = out.shape();
-    std::vector<int> t_axes(out.ndim(), -1);
-    std::vector<int> out_reshape(out.ndim(), -1);
+      has_neg_strides |= true;
```
awni, line 473: nit: `has_neg_strides |= (strides[i] < 0)`

awni, line 473: nit `has_neg_strides = true`

mlx/backend/metal/kernels/copy.metal
```diff
@@ -1,91 +1,91 @@
-// Copyright © 2023 Apple Inc.
+// Copyright © 2023-2024 Apple Inc.
 
 #include "mlx/backend/metal/kernels/bf16.h"
 #include "mlx/backend/metal/kernels/utils.h"
 
 template <typename T, typename U>
 [[kernel]] void copy_s(
-    device const T* src,
```
awni, line 8: Thanks for adding the buffer argnumber here 👍

mlx/utils.cpp
jagrit06: Yes, I was printing them to make sure things went around correctly

mlx/ops.cpp
jagrit06: Good point

angeloskath (approved):
🚀

PR #853, opened 2024-03-20
Pass shape and inputs by value in array's constructor

zcbenz:
## Proposed changes

Since the shape and inputs are always saved as copy in ArrayDesc, we can unify array's constructors to just take the arguments by value.

There are 2 cases:
1. When shape is a lvalue, it will be copied into array's constructor and then moved into ArrayDesc's member. So only 1 copy happens.
2. When shape is a rvalue, it will be moved into array's constructor and then moved into ArrayDesc's member. So no copy happens.

So having 1 constructor that takes by value is equivalent to having 2 constructors that take const reference and rvalue separately.

## Checklist

### DIFF
--- a/mlx/array.cpp
+++ b/mlx/array.cpp
@@ -37,21 +37,10 @@ array::array(const std::complex<float>& val, Dtype dtype /* = complex64 */)
 }
 
 array::array(
-    const std::vector<int>& shape,
-    Dtype dtype,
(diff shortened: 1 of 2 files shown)


awni:
This is a nice change, thanks! I am letting the tests clear, then we can merge it.

awni (approved):
This is really great.. now we just need to start `std::move`ing stuff into array more consistently :)

PR #859, opened 2024-03-20
Add minimum for cosine decay function

chimezie:
## Proposed changes

The Cosine decay schedule is more commonly used to decay the LR to a proportion of a targeted LR rather than to 0.  This simply adds an (optional) argument for the final decayed value.

## Checklist

### DIFF
--- a/python/mlx/optimizers/schedulers.py
+++ b/python/mlx/optimizers/schedulers.py
@@ -65,7 +65,7 @@ def cosine_decay(init: float, decay_steps: int, minimum: float = 0.0) -> Callabl
         init (float): Initial value.
         decay_steps (int): Number of steps to decay over. The decayed
             value is constant for steps beyond ``decay_steps``.
-        minimum (float): Minimal value to decay to (defaults to 0)
+        minimum (float, optional): Minimal value to decay to. Default: ``0``.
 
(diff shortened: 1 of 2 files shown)


python/mlx/optimizers/schedulers.py
```diff
@@ -58,13 +58,14 @@ def schedule(step):
...
 
-def cosine_decay(init: float, decay_steps: int) -> Callable:
+def cosine_decay(init: float, decay_steps: int, minimum: float = 0.0) -> Callable:
     r"""Make a cosine decay scheduler.
 
     Args:
         init (float): Initial value.
         decay_steps (int): Number of steps to decay over. The decayed
             value is constant for steps beyond ``decay_steps``.
+        minimum (float): Minimal value to decay to (defaults to 0)
```
awni, line 68:
```suggestion
        minimum (float, optional): Minimal value to decay to. Default: ``0``.
```

awni (changes requested):
Looks great, just a couple minor comments then we can merge!

awni (approved):

awni:
Thanks for the addition!

chimezie:
NP.  My pleasure

PR #862, opened 2024-03-20
Fast RMS Norm

awni:
- Added looped kernel
- Buffer donation in softmax

Without compile TPS on Mistral 4bit M2 Ultra 1 goes from `79.413` to `85.172`.


awni:
This is ready for review I think.

angeloskath (approved):
I tested with Qwen 72B and underflow was happening so I changed the kernel to perform the squaring in float32 which took care of it with an imperceptible performance drop. 84.1 to 83.8 on the M2 Ultra 0.

I think this looks great and ready to go.

PR #863, opened 2024-03-20
Do not use static constexpr in header

zcbenz:
## Proposed changes

Doing so results in each compilation unit (.cpp file) having its own copy of the variable, while inline constexpr makes sure there is only one copy.
## Checklist


awni (approved):
Nice, thanks!

PR #864, opened 2024-03-21
Add missing && in eval

zcbenz:
## Proposed changes

Without the && args would be copied and perfect forwarding won't work.

To avoid eval calling itself recursively, the const ref version of eval is changed to take by value instead, which will also save a copy of array when a rvalue is passed.

## Checklist

### DIFF
--- a/mlx/transforms.cpp
+++ b/mlx/transforms.cpp
@@ -37,7 +37,7 @@ class Synchronizer : public Primitive {
 // are currently under a function transformation.
 int detail::InTracing::tracing_counter{0};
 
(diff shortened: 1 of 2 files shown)


awni:
Looks like this broke the tests..

zcbenz:
It turns out `eval` ended up calling itself recursively, because the rvalue version takes precedence over the const ref version.

I solved it by changing the const ref version of `eval` to take argument by value instead, as the bonus a copy of array will be saved when a rvalue is passed to `eval`.

awni (approved):
🙏  thanks!!

PR #867, opened 2024-03-21
Use reshape and transpose for non-overlapping pooling windows

angeloskath:
For non overlapping pooling windows route to simple reshape and transpose instead of `as_strided` to have gradients that do not resort to scatter.

Training a LeNet on MNIST:

Before
```
Test accuracy 0.983, Time 2.924 (s)
```

After
```
Test accuracy 0.982, Time 2.810 (s)
```

### DIFF
--- a/python/mlx/nn/layers/pooling.py
+++ b/python/mlx/nn/layers/pooling.py
@@ -20,6 +20,22 @@ def _value_or_list(x, n, msg):
     return [x] * n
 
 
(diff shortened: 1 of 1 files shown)


awni:
Nice bump! It might make sense to push all the pooling logic into a `pool_general` op in C++, that way we can get these gains in Swift and likely avoid any python overhead as well.

awni (approved):

PR #870, opened 2024-03-21
Same scope as mx.fast.rms_norm this adds mx.fast.layer_norm

angeloskath:
Starting the PR for review and will be adding some benchmarks as I am running them.

### DIFF
--- a/mlx/fast.cpp
+++ b/mlx/fast.cpp
@@ -87,7 +87,7 @@ array rms_norm(
   if (s.device == Device::gpu) {
     return array(
         x.shape(),
-        x.dtype(),
+        out_type,
         std::make_unique<RMSNorm>(s, fallback, eps),
         {astype(x, out_type, s), astype(weight, out_type, s)});
   }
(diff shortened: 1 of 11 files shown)


mlx/fast.cpp
```diff
@@ -87,7 +87,7 @@ array rms_norm(
   if (s.device == Device::gpu) {
     return array(
         x.shape(),
-        x.dtype(),
+        out_type,
```
awni, line 90: that looks like a bug.. 😬

angeloskath: Looks like a bug now or before? Cause I thought I was fixing a bug :-)

awni: Yea I meant before. Good catch!

awni:
`python -m mlx_lm.generate --model  mlx-community/starcoder2-7b-4bit --prompt "Write a quick sort in C++" --max-tokens 256 --temp 0.0`

Pre
```
Generation: 67.713 tokens-per-sec
```

Post
```
Generation: 75.267 tokens-per-sec
```

awni (approved):
🚀

angeloskath:
Waiting on the tests and merging it.

PR #871, opened 2024-03-22
Reduce a little overhead

awni:
A few minor changes here and there to reduce some overheads in graph creation.

Graph only Mistral 4-bit from 969.3 -> 1450.453
TPS on Mistral 4-bit from 87.7 -> 90.3

I kind of found some of this just doing a little digging into overheads. Doesn't really move the needle much but also adds basically no code complexity so I thought I would add it in.


angeloskath (approved):
I am amazed at the impact of these small changes. Great job!

jagrit06 (approved):
🚀

PR #874, opened 2024-03-22
Reduce implicit copies in make_array

zcbenz:
## Proposed changes

1. Move shapes into outputs instead of copying them.
2. Pass primitive by const ref as it is always copied into outputs, which removes a copy when calling make_array.

## Checklist


awni (approved):
🙏  thanks!

PR #875, opened 2024-03-22
Pinv

adhulipa:
## Proposed changes

Add Moore-Penrose Pseudo Inverse function.   Inspired by the recent PRs from @nicolov in adding [svd](https://github.com/ml-explore/mlx/pull/809) and [inv](https://github.com/ml-explore/mlx/pull/822), this PR adds the `pinv` primitive

## Tests
Ran some tests locally and included them in PR
```
>>> import mlx.core as mx; mx.set_default_device(mx.cpu)
... A = mx.array([[1, 2, 1, 1, 9], [3, 4, 2, 2, 8], [2, 2, 1, 1, 4], [5, 6, 7, 2, 3.0] ])
... A_pinv = mx.linalg.pinv(A)
... A @ A_pinv @ A
>>>
>>>  A @ A_pinv @ A
array([[1, 2, 1, 1, 9],
       [3, 4, 2, 2, 8],
       [2, 2, 1, 0.999999, 4],
       [5, 6, 7, 2, 3]], dtype=float32)
```

Re-Tested, and everything looks good

```

>>>
... import mlx.core as mx; mx.set_default_device(mx.cpu)
... A = mx.array([[1, 2], [3, 4], [2, 2], [5, 3.0] ])
... A_pinv = mx.linalg.pinv(A)
... print(A.shape, ",\tallclose? ", mx.allclose(A, A @ A_pinv @ A))
...
... import mlx.core as mx; mx.set_default_device(mx.cpu)
... A = mx.array([[1, 2, 1, 1, 9], [3, 4, 2, 2, 8], [2, 2, 1, 1, 4], [5, 6, 7, 2, 3.0] ])
... A_pinv = mx.linalg.pinv(A)
... print(A.shape, ",\tallclose? ", mx.allclose(A, A @ A_pinv @ A))
...
... import mlx.core as mx; mx.set_default_device(mx.cpu)
... A = mx.array([[1.0, 2], [3, 4.0]])
... A_pinv = mx.linalg.pinv(A)
... print(A.shape, ",\tallclose? ", mx.allclose(A, A @ A_pinv @ A))
(4, 2) ,	allclose?  array(True, dtype=bool)
(4, 5) ,	allclose?  array(True, dtype=bool)
(2, 2) ,	allclose?  array(True, dtype=bool)
>>>
>>>
```

## Checklist

### DIFF
--- a/mlx/linalg.cpp
+++ b/mlx/linalg.cpp
@@ -320,26 +320,33 @@ array pinv(const array& a, StreamOrDevice s /* = {} */) {
     throw std::invalid_argument(msg.str());
   }
 
-  const auto m = a.shape(-2);
-  const auto n = a.shape(-1);
-  const auto k = std::min(m, n);
-  const auto outs = linalg::svd(a, s);
-  const auto U = outs[0];
-  const auto S = outs[1];
-  const auto Vt = outs[2];
+  int m = a.shape(-2);
+  int n = a.shape(-1);
+  int k = std::min(m, n);
+  auto outs = linalg::svd(a, s);
+  array U = outs[0];
+  array S = outs[1];
+  array V = outs[2];
 
   std::vector<int> starts(a.ndim(), 0);
   std::vector<int> ends = a.shape();
-  const int i = a.ndim() - 2;
-  const int j = a.ndim() - 1;
+  int i = a.ndim() - 2;
+  int j = a.ndim() - 1;
+
+  // Prepare U
   ends[i] = m;
   ends[j] = k;
-  const auto U_slice = slice(U, starts, ends, s);
+  U = swapaxes(slice(U, starts, ends, s), -1, -2, s);
+
+  // Prepare V
   ends[i] = k;
   ends[j] = n;
-  const auto V_slice = slice(Vt, starts, ends, s);
-  return swapaxes(
-      matmul(divide(U_slice, expand_dims(S, -2, s), s), V_slice, s), -1, -2, s);
+  V = swapaxes(slice(V, starts, ends, s), -1, -2, s);
+
+  // Prepare S
+  S = expand_dims(S, -2, s);
+
+  return matmul(divide(V, S, s), U);
 }
 
 array cholesky_inv(
--- a/mlx/linalg.cpp
+++ b/mlx/linalg.cpp
@@ -306,6 +306,42 @@ array cholesky(
       {a});
 }
 
+array pinv(const array& a, StreamOrDevice s /* = {} */) {
+  if (a.dtype() != float32) {
+    std::ostringstream msg;
+    msg << "[linalg::pinv] Arrays must type float32. Received array "
+        << "with type " << a.dtype() << ".";
+    throw std::invalid_argument(msg.str());
+  }
+  if (a.ndim() < 2) {
+    std::ostringstream msg;
+    msg << "[linalg::pinv] Arrays must have >= 2 dimensions. Received array "
+        << "with " << a.ndim() << " dimensions.";
+    throw std::invalid_argument(msg.str());
+  }
+
+  const auto m = a.shape(-2);
+  const auto n = a.shape(-1);
+  const auto k = std::min(m, n);
+  const auto outs = linalg::svd(a, s);
+  const auto U = outs[0];
+  const auto S = outs[1];
+  const auto Vt = outs[2];
+
+  std::vector<int> starts(a.ndim(), 0);
+  std::vector<int> ends = a.shape();
+  const int i = a.ndim() - 2;
+  const int j = a.ndim() - 1;
+  ends[i] = m;
+  ends[j] = k;
+  const auto U_slice = slice(U, starts, ends, s);
+  ends[i] = k;
+  ends[j] = n;
+  const auto V_slice = slice(Vt, starts, ends, s);
+  return swapaxes(
+      matmul(divide(U_slice, expand_dims(S, -2, s), s), V_slice, s), -1, -2, s);
+}
+
 array cholesky_inv(
     const array& L,
     bool upper /* = false */,
--- a/tests/linalg_tests.cpp
+++ b/tests/linalg_tests.cpp
@@ -348,3 +348,45 @@ TEST_CASE("test matrix cholesky") {
   CHECK(allclose(matmul(transpose(U), U), A, /* rtol = */ 0, /* atol = */ 1e-6)
             .item<bool>());
 }
+
+TEST_CASE("test matrix pseudo-inverse") {
+  // 0D and 1D throw
+  CHECK_THROWS(linalg::pinv(array(0.0), Device::cpu));
+  CHECK_THROWS(linalg::pinv(array({0.0, 1.0}), Device::cpu));
+
+  // Unsupported types throw
+  CHECK_THROWS(linalg::pinv(array({0, 1}, {1, 2}), Device::cpu));
+
+  { // Square m == n
+    const auto A = array({1.0, 2.0, 3.0, 4.0}, {2, 2});
+    const auto A_pinv = linalg::pinv(A, Device::cpu);
+    const auto A_again = matmul(matmul(A, A_pinv), A);
+    CHECK(allclose(A_again, A).item<bool>());
+    const auto A_pinv_again = matmul(matmul(A_pinv, A), A_pinv);
+    CHECK(allclose(A_pinv_again, A_pinv).item<bool>());
+  }
+  { // Rectangular matrix m < n
+    const auto prng_key = random::key(42);
+    const auto A = random::normal({4, 5}, prng_key);
+    const auto A_pinv = linalg::pinv(A, Device::cpu);
+    const auto zeros = zeros_like(A_pinv, Device::cpu);
+    CHECK_FALSE(allclose(zeros, A_pinv, /* rtol = */ 0, /* atol = */ 1e-6)
+                    .item<bool>());
+    const auto A_again = matmul(matmul(A, A_pinv), A);
+    CHECK(allclose(A_again, A).item<bool>());
+    const auto A_pinv_again = matmul(matmul(A_pinv, A), A_pinv);
+    CHECK(allclose(A_pinv_again, A_pinv).item<bool>());
+  }
+  { // Rectangular matrix m > n
+    const auto prng_key = random::key(10);
+    const auto A = random::normal({6, 5}, prng_key);
+    const auto A_pinv = linalg::pinv(A, Device::cpu);
+    const auto zeros2 = zeros_like(A_pinv, Device::cpu);
+    CHECK_FALSE(allclose(zeros2, A_pinv, /* rtol = */ 0, /* atol = */ 1e-6)
+                    .item<bool>());
+    const auto A_again = matmul(matmul(A, A_pinv), A);
+    CHECK(allclose(A_again, A).item<bool>());
+    const auto A_pinv_again = matmul(matmul(A_pinv, A), A_pinv);
+    CHECK(allclose(A_pinv_again, A_pinv).item<bool>());
+  }
+}
--- a/mlx/linalg.h
+++ b/mlx/linalg.h
@@ -70,6 +70,8 @@ array tri_inv(const array& a, bool upper = false, StreamOrDevice s = {});
 
 array cholesky(const array& a, bool upper = false, StreamOrDevice s = {});
 
+array pinv(const array& a, StreamOrDevice s = {});
+
 array cholesky_inv(const array& a, bool upper = false, StreamOrDevice s = {});
 
 } // namespace mlx::core::linalg
--- a/python/src/linalg.cpp
+++ b/python/src/linalg.cpp
@@ -353,4 +353,28 @@ void init_linalg(nb::module_& parent_module) {
         Returns:
           array: :math:`\mathbf{A^{-1}}` where :math:`\mathbf{A} = \mathbf{L}\mathbf{L}^T`.
       )pbdoc");
+  m.def(
+      "pinv",
+      &pinv,
+      "a"_a,
+      nb::kw_only(),
+      "stream"_a = nb::none(),
+      nb::sig(
+          "def pinv(a: array, *, stream: Union[None, Stream, Device] = None) -> array"),
+      R"pbdoc(
+        Compute the (Moore-Penrose) pseudo-inverse of a matrix.
+
+        This function calculates a generalized inverse of a matrix using its
+        singular-value decomposition. This function supports arrays with at least 2 dimensions.
+        When the input has more than two dimensions, the inverse is computed for each
+        matrix in the last two dimensions of ``a``.
+
+        Args:
+            a (array): Input array.
+            stream (Stream, optional): Stream or device. Defaults to ``None``
+              in which case the default stream of the default device is used.
+
+        Returns:
+            array: ``aplus`` such that ``a @ aplus @ a = a``
+      )pbdoc");
 }
--- a/python/tests/test_linalg.py
+++ b/python/tests/test_linalg.py
@@ -181,6 +181,18 @@ class TestLinalg(mlx_tests.MLXTestCase):
         for M, L in zip(AB, Ls):
             self.assertTrue(mx.allclose(L @ L.T, M, rtol=1e-5, atol=1e-7))
 
+    def test_pseudo_inverse(self):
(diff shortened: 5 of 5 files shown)


tests/linalg_tests.cpp
```diff
@@ -322,3 +322,25 @@ TEST_CASE("test matrix inversion") {
...
+  CHECK_THROWS(linalg::pinv(array({0, 1}, {1, 2}), Device::cpu));
+
+  // Non-square doesn't throw.
+  CHECK_NOTHROW(
+      linalg::pinv(array({1.0, 2.0, 3.0, 4.0, 5.0, 6.0}, {2, 3}), Device::cpu));
+
+  const auto prng_key = random::key(42);
+  const auto A = random::normal({4, 5}, prng_key);
+  const auto A_plus = transpose(linalg::pinv(A, Device::cpu));
+  const auto identity = eye(A.shape(0));
```
adhulipa, line 341:
```suggestion
```

angeloskath:
@adhulipa I think this implementation is the wrong way to go about it. Using SVD to compute the pseudo inverse means we don't need a primitive and kernels etc. It is just an op that can reside in the `mlx::core::linalg` namespace.

Basically something like the following (of the top of my head so ymmv)

```python
def pinv(x):
     U, S, V = mx.linalg.svd(x)
     return (V[:len(x)].T * 1/S) @ U.T
```

adhulipa:
Ahh I see!  I didn’t think about that.  Thanks for the review @angeloskath 

I suppose we could modify this PR to merge in a Python form one as a first step and then investigate whether a custom kernel is necessary. 

Would you recommend such a direction?

awni:
The op should be in C++ and then do a binding (we try to keep the C++ and Python APIs reasonably consistent). I think the Python impl from @angeloskath is just intended as pseudo-code for that.

adhulipa:
Ah yes that makes sense.  I should add the Python api that matches the cpp api for pinv().  I haven’t gotten around to it. Thank you for taking a look folks!

adhulipa:
I made a few updates.  Still gotta figure out how to fix the cpp op where svd(A) returns u, s, vt where u has same dims as A (when rectangular). This makes the matmul incompatible. 

I have a path to green where I need to tweak u to match expected end-shape. 

(I’m positive the SVD approach works accurately because I validated it in Python api mlx; and few other langs such as matlab to be certain)

Also can use the PyTorch impl as a reference https://github.com/pytorch/pytorch/blob/2ffab6e663b9c6951048b8c8ba82d2cc5ca5c2fc/aten/src/ATen/native/LinearAlgebra.cpp#L532

just need to get around to it in due time

mlx/backend/common/pseudoinverse.cpp
```diff
@@ -0,0 +1,58 @@
...
+  auto ut = transpose(u);
+  // s_plus 4x4
+  // u      4x5
+  // ut     5x4
+  // v      5x5
+  // ut*s   5x4
+  // v*     5x4
+  auto inner = transpose(matmul(s_plus, u));
+  auto result = matmul(v, inner);
+  copy(result, pinv, a.flags().row_contiguous ? CopyType::Vector : CopyType::General);
```
adhulipa, line 51: there's bug here (or it's call site where the `pinv` array is allocated) where the size/shape of the pinv array is not correctly setup
```
terminated by signal SIGSEGV (Address boundary error)
```

awni:
@adhulipa are you planning to come back to this?

awni:
@adhulipa are you planning to return to this one?

adhulipa:
Hi @awni yes I will update this one.  I am running into an issue where I haven't figured out how to allocate the rectangular array for the output/result array before passing it off to the PINV function.

adhulipa:
@awni do you think it’s better to close this PR and reopen against a newer mainline commit?  Happy to do so if it helps keep your PR todo list clean

awni:
Its more up to you. If you plan to work on it in the near future then you can keep it open (or start a new one if you prefer). If not, I would close it.

adhulipa:
I'll keep it open for now.  I'll close and re-open if it gets too far behind significantly -- for now these changes are additive; so that's not a risk. It just needs a bit of polish/bugfixing.

adhulipa:
Made some progress.  Need to fix a few more things.

adhulipa:
I am suspecting there's something I need to figure out with how im using `mx.linalg.svd(A)` or rather in cpp `svd_imp()` and then the matmuls for getting the pinv

Im seeing 
```
>>> A = mx.array([[1.0, 2, 1, 1, 9], [3, 4, 2, 2, 8], [2, 2, 1, 1, 4], [5, 6, 7, 2, 3] ])
>>> U, S, Vt = mx.linalg.svd(A)
>>> U @ mx.diag(S) @ Vt
ValueError: [matmul] Last dimension of first input with shape (4,4) must match second to last dimension of second input with shape (5,5).
```

Which seems contradictory to the [MLX svd doc](https://ml-explore.github.io/mlx/build/html/python/_autosummary/mlx.core.linalg.svd.html#mlx.core.linalg.svd)
>  Returns
> The U, S, and Vt matrices, such that A = U @ diag(S) @ Vt


Of course, I ack MLX mimics the NumPy API and NumPy indeed also produces a similar result.  But it looks like they have support for a `full_matrices: Bool = True` kwarg; which I suppose was designed to help for these types of cases

```
A = np.array([[1, 2, 1, 1, 9], [3, 4, 2, 2, 8], [2, 2, 1, 1, 4], [5, 6, 7, 2, 3] ])
U, S, Vt = np.linalg.svd(A, full_matrices=False)
U @ np.diag(S) @ Vt

>>> U @ np.diag(S) @ Vt
array([[1., 2., 1., 1., 9.],
       [3., 4., 2., 2., 8.],
       [2., 2., 1., 1., 4.],
       [5., 6., 7., 2., 3.]])

>>> np.allclose(A, U @ np.diag(S) @ Vt)
True
```

(Fwiw, without full_matrics=False, error is same as mlx)
```
>>> U, S, Vt = np.linalg.svd(A)
>>> U @ np.diag(S) @ Vt
ValueError: matmul: Input operand 1 has a mismatch in its core dimension 0, with gufunc signature (n?,k),(k,m?)->(n?,m?) (size 5 is different from 4)
```

awni:
I think you can just do something like this:

```python
U, S, V = mx.linalg.svd(A)
K = min(A.shape[0], A.shape[1])
Atilde = (U[:, :K] * S) @ V[:K, :]
```

We could add the slicing as an option like Numpy if it's useful.


Also I would recommend you rebase before making further progress to make it easier to resolve conflicts.

adhulipa:
Ah thanks Awni! Will use that
ack on rebase.

adhulipa:
Small update: Got a local build that correctly computes pinv in most of the tests.   Cleaning up some things and polishing up the code. 

Turns out I was incorrectly relying on the computation array graph API instead of computing the _actual_ matrix products (D'oh!).  Now I have some code locally using lapack's mm func (such as sgemm) to compute the final pinv product. 

Will update this PR soon

adhulipa:
Perhaps there's one more thing to check (on my part) in the python tests.  Will look into it.  In the meantime, this PR is in a good enough shape for a review from @awni and other MLX folks.  Thanks!

adhulipa:
Drats. I have another bug to fix.  I updated the tests to catch it. Will look into and fix.  Essentially, long rectangular matrices have a matmul dim mismatch -- which means I have made an error in the m, n, k calculations and/or slice selections or U/Vt

adhulipa:
Fixed the bug for rectangular matrices where M > N 🎉 
Will publish commit soon

adhulipa:
This PR is ready for a review from Awni, Angelos and other MLX folks. Thanks!

angeloskath:
Hi @adhulipa . I think there shouldn't be a primitive for this operation. It can really just be an op in the linalg namespace.

adhulipa:
Hi @angeloskath ohh I see. I think I may have misinterpreted something in the thread here then.  Particularly what @awni shared after you (@angeloskath) shared that comment earlier. 

> The op should be in C++ and then do a binding

Is it accurate to say that you meant this should in linalg.cpp where we add something relatively simple

```
auto outs = linalg.svd(x);
array U = outs[0]; 
array S = outs[1];
...///  etc.. 

return (V[:len(x)].T * 1/S) @ U.T // of course, in the cpp variant instead of the python-esque here
```

adhulipa:
Actually, @angeloskath do you mean to say that we don't need a primitive; but all the logic of calling `linalg::svd()' ensuring the svd() call is eval()'d and then passing to lapack's `sgemm()` function calls all should be in the linalg.cpp file (and namespace)? 

It seems like the recommendation here is to keep the core logic intact but just not make this a primitive.  Am I understanding that correctly?

adhulipa:
@angeloskath @awni -- quick question: do you folks feel like this is in a good shape for a review?  Of course, no rush from my pov; just thought I'd check.

adhulipa:
hi @angeloskath @awni -- im curious if you think this PR is good for merge or review? Or perhaps there's another way to build this functionality? I am thankful for all the feedback so far and pointers on the road forward : )

mlx/linalg.cpp
```diff
@@ -292,4 +292,41 @@ array cholesky(
...
+    std::ostringstream msg;
+    msg << "[linalg::pinv] Arrays must type float32. Received array "
+        << "with type " << a.dtype() << ".";
+    throw std::invalid_argument(msg.str());
+  }
+  if (a.ndim() < 2) {
+    std::ostringstream msg;
+    msg << "[linalg::pinv] Arrays must have >= 2 dimensions. Received array "
+           "with "
+        << a.ndim() << " dimensions.";
```
angeloskath, line 306: I would change the above to 

```
<< "with " << a.ndim() << " dimensions.";
```

mlx/linalg.cpp
```diff
@@ -292,4 +292,41 @@ array cholesky(
...
+    throw std::invalid_argument(msg.str());
+  }
+
+  const auto m = a.shape(-2);
+  const auto n = a.shape(-1);
+  const auto k = std::min(m, n);
+  const auto rank = a.ndim();
```
angeloskath, line 313: `rank` is not used anywhere .

mlx/linalg.cpp
```diff
@@ -292,4 +292,41 @@ array cholesky(
...
+  std::vector<int> pinv_shape;
+  if (m <= n) {
+    pinv_shape = {n, k};
+  } else {
+    pinv_shape = {k, m};
+  }
+  array pinv(pinv_shape, a.dtype(), nullptr, {});
+  const auto U_slice = slice(U, {0, 0}, {m, k});
+  const auto Vt_slice = slice(Vt, {0, 0}, {k, n});
+  return matmul(matmul(transpose(Vt_slice), diag(1.0 / S)), transpose(U_slice));
```
angeloskath, line 329: No need for two matmuls and a `diag` this can be done with a division. Also the streams should be appropriately used.

```cpp
std::vector<int> starts(a.ndim(), 0);
std::vector<int> ends = a.shape();
int i = a.ndim()-2;
int j = a.ndim()-1;
ends[i] = n;
ends[j] = k;
array U_slice = slice(U, starts, ends, s);
ends[i] = k;
ends[j] = m;
array V_slice = slice(Vt, starts, ends, s);
return swapaxes(matmul(divide(U_slice, expand_dims(S, -2, s), s), V_slice, s), -1, -2, s);
```

mlx/linalg.cpp
```diff
@@ -292,4 +292,41 @@ array cholesky(
...
+  auto S = outs[1];
+  auto Vt = outs[2];
+
+  std::vector<int> pinv_shape;
+  if (m <= n) {
+    pinv_shape = {n, k};
+  } else {
+    pinv_shape = {k, m};
+  }
+  array pinv(pinv_shape, a.dtype(), nullptr, {});
```
angeloskath, line 326: There is no need to initialize a `pinv` array at all this array is also not used anywhere.

angeloskath (changes requested):
Hi! Sorry for taking some time to review this, it is close but it still needs some changes before merging.

adhulipa:
Thank you Angelos! No need to apologize re: delay : ) -- I appreciate the review + feedback.  I will address and update the PR this week.

mlx/linalg.cpp
adhulipa: Done ✅ - in new commit 587a30f6; [diff here](https://github.com/ml-explore/mlx/pull/875/commits/587a30f6bd32a12a0f58ed660fc7a5144820c3b1#diff-9f87ac95c2fab89844d98cd9bd7727e2e7a872880ea49db8baf529bb43bdfcddR305)

mlx/linalg.cpp
adhulipa: Removed ✅ - in new commit 587a30f6; [diff here](https://github.com/ml-explore/mlx/pull/875/commits/587a30f6bd32a12a0f58ed660fc7a5144820c3b1#diff-9f87ac95c2fab89844d98cd9bd7727e2e7a872880ea49db8baf529bb43bdfcddL313)

mlx/linalg.cpp
adhulipa: Updated ✅ - in new commit 587a30f6; [diff here](https://github.com/ml-explore/mlx/pull/875/commits/587a30f6bd32a12a0f58ed660fc7a5144820c3b1#diff-9f87ac95c2fab89844d98cd9bd7727e2e7a872880ea49db8baf529bb43bdfcddR328)

Btw, I think your message above mixed up the `ends[]` dims for `U_slice` vs `V_slice` leading to their dims being `{n, k}` and `{k, m}` respectively.  But IIUC they should instead be `U_slice -> {m, k}` and `V_slice -> {k, n}`. The _exact_ code you shared above caused a test error and after investigating I updated the code to use ensure dims of `U_slice` is `{m, k}` and `V_slice` is `{k, n}`.  I verified the results comparing to numpy/matlab. But please double check my changes. Thanks!

mlx/linalg.cpp
adhulipa: Removed ✅ - in new commit 587a30f6; [diff here](https://github.com/ml-explore/mlx/pull/875/commits/587a30f6bd32a12a0f58ed660fc7a5144820c3b1#diff-9f87ac95c2fab89844d98cd9bd7727e2e7a872880ea49db8baf529bb43bdfcddL326)

adhulipa:
Looks like a test failed. I will investigate

```
FAIL [0.006s]: test_pseudo_inverse (test_linalg.TestLinalg)
  File "/Users/distiller/project/python/tests/test_linalg.py", line 173, in test_pseudo_inverse
    self.assertTrue(mx.allclose(A @ A_plus @ A, A, rtol=0, atol=1e-6))
```

adhulipa:
It looks like the failure is a matter of precision

For the square matrix case, it looks like 1e-5 is a passing tolerance limit whereas the existing 1e-6 level is causing a failure
```
>>>         A = mx.array([[1, 2, 3], [6, -5, 4], [-9, 8, 7]], dtype=mx.float32)
...         A_plus = mx.linalg.pinv(A, stream=mx.cpu)
>>> mx.allclose(A @ A_plus @ A, A, rtol=0, atol=1e-5)
array(True, dtype=bool)

>>> mx.allclose(A @ A_plus @ A, A, rtol=0, atol=1e-6)
array(False, dtype=bool)
```

Similarly, for a 2x3x3 matrix, it seems to be within 1e-3 but not within 1e-6
```
>>>         B = A - 100
...         AB = mx.stack([A, B])
...         pinvs = mx.linalg.pinv(AB, stream=mx.cpu)
>>>
>>> pinvs
array([[[0.313084, -0.0467289, -0.107477],
        [0.364486, -0.158878, -0.0654206],
        [-0.0140187, 0.121495, 0.0794393]],
       [[0.0915965, -0.0186531, -0.0762813],
        [0.169056, -0.134106, -0.0378953],
        [-0.274593, 0.154526, 0.11614]]], dtype=float32)

>>> for M, M_plus in zip(AB, pinvs):
...     print(M @ M_plus @ M)
...     print(mx.allclose(M @ M_plus @ M, M, rtol=0, atol=1e-3))
array([[1, 2, 3],
       [6, -5, 4],
       [-9, 8, 7]], dtype=float32)
array(True, dtype=bool)
array([[-98.9998, -97.9998, -96.9998],
       [-93.9997, -105, -95.9998],
       [-109, -91.9998, -92.9998]], dtype=float32)
array(True, dtype=bool)
```

adhulipa:
@angeloskath hey there Angelos, curious if you or others on MLX could come back to this PR sometime? Thanks!

angeloskath (approved):

PR #876, opened 2024-03-22
Change unary/binary vmap to accept vector as axis

zcbenz:
## Proposed changes

It is hard to tell what the code does without looking at the source code when you call something like `vmap(fun, 1, 2)`.

Refactored the code of vmap to make use of move semantics.

Reordered the code so the order of declaration matches the definition.

## Checklist

### DIFF
--- a/mlx/transforms.cpp
+++ b/mlx/transforms.cpp
@@ -689,10 +689,10 @@ std::vector<array> vmap_replace(
 } // namespace detail
 
 std::function<std::vector<array>(const std::vector<array>&)> vmap(
-    const std::function<std::vector<array>(const std::vector<array>&)>& fun,
-    const std::vector<int>& in_axes /* = {} */,
-    const std::vector<int>& out_axes /* = {} */) {
-  auto infer_axes = [](auto axes) {
+    std::function<std::vector<array>(const std::vector<array>&)> fun,
+    std::vector<int> in_axes /* = {} */,
+    std::vector<int> out_axes /* = {} */) {
+  auto infer_axes = [](const auto& axes) {
     return !axes.empty() &&
         std::all_of(axes.begin(), axes.end(), [](int ax) { return ax < 0; });
   };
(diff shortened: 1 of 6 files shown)


awni:
Looks like we are using these in the tests... you could update the tests to use the vector version. I don't mind removing these overloads, they are mostly there to make it easy to use vmap in C++. On the other hand, I don't really mind keeping them either..

awni:
PS probably you did not run the C++ tests which is why you did not notice we use them. I recommend running those in addition to the Python tests, especially if you are doing a lot of C++ changes.

zcbenz:
Sorry I forgot the C++ tests.

It is actually nice for the APIs to be able to accepts unary/binary functions, I have refactored the code to make unary/binary `vmap` accept vector instead of `int`s, so all three versions of `vmap` have the same APIs.

(It is also possible to write a template version of `vmap` that accepts a function with arbitrary numbers of arrays, but without C++20 it would be too hard to deduce the types of function parameters from lambda.)

awni:
The axes parameters are intended to match the number of arrays which makes it easy to call these functions from C++ otherwise partly defeats the purpose.

awni:
Probably we should close this one as I don't really think it makes to change it to how you have it. Removing some overloads might be nice, but again isn't super important since we basically never use those functions except in the tests.

zcbenz:
Yeah I feel I'm spending too much efforts on something not important.

zcbenz:
@awni By the way, is it possible to reuse the build dir for building both python tests and C++ tests? Running `python3 setup.py build_ext --inplace && cmake --build build` results in cmake error.

awni:
I never tried what you are doing exactly but I build in the same dir usually using:

```
env CMAKE_BUILD_PARALLEL_LEVEL="" python setup.py develop
```

And for C++ in the build dir (though `cmake --build build -- -j` should be fine)

```
make -j
```

angeloskath:
This is my goto for building both

```
CMAKE_ARGS='-DMLX_BUILD_TESTS=ON' python setup.py build_ext --inplace -j 8
```

Issue #502, opened 2024-01-19
[Feature] Add support mx.random.multivariate_normal()


tedwards2412:
@NNSSA and I are working on a sampling package for mlx (https://github.com/tedwards2412/samplex) and it would be extremely useful to have these two functions to do more generic sampling. The latter will involve adding more functionality to the core.linalg sub-package. Is this likely to come in a future update? Happy to help if needed!

awni:
Let's do these as two issues as the first is much easier than the second. I assume for multivariate normal you need a non-diagonal covariance?

awni:
Leaving this issue for `mx.random.multivariate_normal` and created #503 for `diag`

awni:
FYI: for multivariate normal we probably 🤔 need matrix inversion e.g. `mx.linalg.inv`. Which will also probably help with other things.

tedwards2412:
Great, thanks! And yes, non-diagonal covariance would be essential for this.

awni:
Cool package by the way! You should add a little quick start/usage guide (when it's ready for it).

NNSSA:
Thanks! Will definitely do :-) 
Related, I think numpy uses a singular value decomposition to compute a multivariate normal.

awni:
We have a PR out for QR #310. I think SVD and Cholesky would go similarly. The main issue is there are no Metal implementations for most of Lapack so a lot of this will be CPU only until we can get some kernels implemented.

tedwards2412:
@awni We finally got round to adding the quick start you recommended on https://github.com/tedwards2412/samplex. The sampling seems to work well with mlx so far. Looking forward to working on this more in the future!

awni:
That's awesome!! Out of curiosity, could you tell me a bit more about (some) intended uses for the package? I would love to point people to it if you are ok with that and I understand a bit more in what cases you are targeting.

tedwards2412:
Overall the goal is to see if we can allow people to quickly run fairly large scale MCMC sampling locally rather than having to run on a cluster. But it's also just a research project for us to see if there are particular sampling algorithms that are substantially better when you can switch between running on CPU and GPU without any overhead; I don't think has been explored at all before.

awni:
This was closed a while ago.

PR #877, opened 2024-03-22
Implementation of mlx.random.multivariate_normal (#502)

arn4:
## Proposed changes

I am implementing `mlx.random.multivariate_normal`, as requested in Issue #502.

Right now, there is no check if the provided covariance is positive semi-definite, but something [like `numpy`](https://numpy.org/doc/stable/reference/random/generated/numpy.random.multivariate_normal.html) can be implemented.

All the generated samples must have the same `mean` and `cov`, but maybe we can handle them [like `jax`does](https://jax.readthedocs.io/en/latest/_autosummary/jax.random.multivariate_normal.html).

Let me know any comment/suggestion you have!

## Checklist


python/src/random.cpp
```diff
@@ -173,6 +173,39 @@ void init_random(nb::module_& parent_module) {
...
+      "key"_a = nb::none(),
+      "stream"_a = nb::none(),
+      R"pbdoc(
+        Generate N-dimensional jointly normal distributed random numbers with  given mean and covariance. Note that ``cov`` is cast to double for the computation.
+
+        ``cov`` must be a positive semi-definite matrix. If the matrix is not positive semi-definite, the behavior is undefined.
+
+        Args:
+            mean (array): N-dimensional array, the mean of the distribution.
+            cov (array): N x N-dimensional array, the covariance matrix of the distribution.
```
awni, line 200: I think the use of `N`-dimensional is a bit confusing here. Usually we use `dimension` to represent `a.ndim` which should be `1` (I think?).

python/src/random.cpp
```diff
@@ -173,6 +173,39 @@ void init_random(nb::module_& parent_module) {
...
+            shape (list(int), optional): Shape of the output. Default is ``()``.
```
awni, line 201:
```suggestion
            shape (list(int), optional): Shape of the output. Default: ``()``.
```

python/src/random.cpp
```diff
@@ -173,6 +173,39 @@ void init_random(nb::module_& parent_module) {
...
+            dtype (Dtype, optional): Type of the output. Default is ``float32``.
```
awni, line 202:
```suggestion
            dtype (Dtype, optional): The output type. Default: ``float32``.
```

python/src/random.cpp
```diff
@@ -173,6 +173,39 @@ void init_random(nb::module_& parent_module) {
...
+            key (array, optional): A PRNG key. Default: None.
```
awni, line 203:
```suggestion
            key (array, optional): A PRNG key. Default: ``None``.
```

python/src/random.cpp
```diff
@@ -173,6 +173,39 @@ void init_random(nb::module_& parent_module) {
...
+            mean, cov, shape, type.value_or(float32), key, s);
+      },
+      "mean"_a,
+      "cov"_a,
+      "shape"_a = std::vector<int>{},
+      "dtype"_a.none() = float32,
+      "key"_a = nb::none(),
+      "stream"_a = nb::none(),
+      R"pbdoc(
+        Generate N-dimensional jointly normal distributed random numbers with  given mean and covariance. Note that ``cov`` is cast to double for the computation.
```
awni, line 194: > cast to double

Do you mean `float32`, we don't have a double type? Just curious what's the rational there.

mlx/random.cpp
```diff
@@ -172,6 +173,48 @@ array normal(
...
 }
 
+array multivariate_normal(
+    const array& mean,
+    const array& cov,
+    const std::vector<int>& shape,
+    Dtype dtype,
+    const std::optional<array>& key,
+    StreamOrDevice s) {
+  auto stream = to_stream(s);
```
awni, line 183: Probably I would just require everything to be fp32 here for now (throw otherwise). It really only works for fp32 and I think it's better to be explicit.

awni (changes requested):
Thanks for adding this!! It looks nice so far! I left a few initial comment.

mlx/random.cpp
arn4: Do you mean the `dtype` passed to the function? In that case you are right, since all the operations are working just for `float32`. I would keep the possibility of having `mean` and `variance` of other types though.

python/src/random.cpp
arn4: Yes, sorry, my bad, sometimes I get the names mixed up in my head.

python/src/random.cpp
arn4: Good point. I can switch to something like `shape = (N,)`. If then we decide that is better to have the arguments like in jax, I can use the same explanation of the arguments

awni:
@arn4 could you please rebase, there is a conflict?

python/src/random.cpp
arn4: I have replicated jax in the end, let me know if now it works

arn4:
In the end I have implemented the same signature of jax, since it is more complete than NumPy

mlx/random.cpp
```diff
@@ -192,6 +193,62 @@ array normal(
   return samples;
 }
 
+array multivariate_normal(
+    const array& mean,
+    const array& cov,
+    const std::vector<int>& shape,
+    Dtype dtype,
+    const std::optional<array>& key,
```
awni, line 201:
```suggestion
    const std::optional<array>& key /* = nullopt */,
```

python/src/random.cpp
```diff
@@ -179,6 +179,39 @@ void init_random(nb::module_& parent_module) {
...
+            mean, cov, shape, type.value_or(float32), key, s);
+      },
+      "mean"_a,
+      "cov"_a,
+      "shape"_a = std::vector<int>{},
+      "dtype"_a.none() = float32,
+      "key"_a = nb::none(),
+      "stream"_a = nb::none(),
+      R"pbdoc(
+        Generate jointly-normal random samples, given mean and covariance.
```
awni, line 200:
```suggestion
        Generate jointly-normal random samples, given a mean and covariance.
```

python/src/random.cpp
```diff
@@ -179,6 +179,39 @@ void init_random(nb::module_& parent_module) {
...
+
+        ``cov`` must be a positive semi-definite matrix. If the matrix is not positive semi-definite, the behavior is undefined. The only supported ``dtype`` is ``float32``.
+
+        Args:
+            mean (array): array of shape ``(..,n)``, the mean of the distribution.
+            cov (array): array  of shape ``(..,n,n)``, the covariance matrix of the distribution. The batch shape ``...`` must be broadcast-compatible with that of mean.
```
awni, line 206:
```suggestion
            mean (array): array of shape ``(...,n)``, the mean of the distribution.
            cov (array): array  of shape ``(...,n,n)``, the covariance matrix of the distribution. The batch shape ``...`` must be broadcast-compatible with that of mean. 
```

mlx/random.cpp
```diff
@@ -192,6 +193,62 @@ array normal(
...
+    StreamOrDevice s) {
+  auto stream = to_stream(s);
+
+  if (dtype != float32) {
+    throw std::invalid_argument("[multivariate_normal] dtype must be float32.");
+  }
+
+  auto n = *(mean.shape().end() - 1);
```
awni, line 209:
```suggestion
  auto n = mean.shape(-1);
```

mlx/random.cpp
```diff
@@ -192,6 +193,62 @@ array normal(
...
+
+  // Check shapes comatibility of mean and cov
+  if (n != *(cov.shape().end() - 2)) {
```
awni, line 212:
```suggestion
  if (n != cov.shape(-1)) {
```

mlx/random.cpp
```diff
@@ -192,6 +193,62 @@ array normal(
   return samples;
 }
 
+array multivariate_normal(
+    const array& mean,
+    const array& cov,
```
awni, line 198: Could you add a check that the `mean.ndim() >= 1` and `cov.ndim() >= 2`? O/w it will crash in the shape access.

mlx/random.cpp
```diff
@@ -192,6 +193,62 @@ array normal(
...
+  if (dtype != float32) {
+    throw std::invalid_argument("[multivariate_normal] dtype must be float32.");
+  }
+
+  auto n = *(mean.shape().end() - 1);
+
+  // Check shapes comatibility of mean and cov
+  if (n != *(cov.shape().end() - 2)) {
+    throw std::invalid_argument(
+        "[multivariate_normal] last two dimensions of cov must be equal.");
```
awni, line 214: This error message doesn't match the condition. Do you mean to check `cov.shape(-1) == cov.shape(-2)`?

mlx/random.cpp
```diff
@@ -192,6 +193,62 @@ array normal(
...
+  }
+  if (*(mean.shape().end() - 1) != *(cov.shape().end() - 1)) {
```
awni, line 216:
```suggestion
  if (mean.shape(-1) != cov.shape(-1)) {
```

mlx/random.cpp
```diff
@@ -192,6 +193,62 @@ array normal(
...
+
+  // Compute output shape
+  std::vector<int> truncated_output_shape;
+  try {
+    auto truncated_mean_shape =
+        std::vector<int>(mean.shape().begin(), mean.shape().end() - 1);
+    auto truncated_cov_shape =
+        std::vector<int>(cov.shape().begin(), cov.shape().end() - 2);
+    auto trucated_inferred_shape =
+        broadcast_shapes(truncated_cov_shape, truncated_cov_shape);
```
awni, line 229: Is a bug here? You pass the cov shape twice.

mlx/random.cpp
```diff
@@ -192,6 +193,62 @@ array normal(
...
+    truncated_output_shape = broadcast_shapes(trucated_inferred_shape, shape);
+  } catch (const std::invalid_argument& e) {
```
awni, line 231: No need to put this in a try catch. I would just let the broad throw directly.

mlx/random.cpp
```diff
@@ -192,6 +193,62 @@ array normal(
...
+    throw std::invalid_argument(
+        "[multivariate_normal] mean, cov and shape must have compatible shapes.");
+  }
+  std::vector<int> output_shape = {n};
+  output_shape.insert(
+      output_shape.begin(),
+      truncated_output_shape.begin(),
+      truncated_output_shape.end());
```
awni, line 239: I don't think you need a new shape vector here. Just push `n` back on the `truncated_output_shape` (and maybe rename it to just `output_shape`.

mlx/random.cpp
```diff
@@ -192,6 +193,62 @@ array normal(
...
+
+  // Compute the square-root of the covariance matrix, using the SVD
+  auto covariance = astype(cov, float32, stream);
+  auto SVD = linalg::svd(covariance, stream);
+  auto std =
+      astype(matmul(SVD[0] * sqrt(SVD[1]), SVD[2], stream), dtype, stream);
```
awni, line 245: missing stream on the `sqrt` and the `*` (use `multiply` instead).

awni (changes requested):
Looks really useful! I left a few comments.. should be straightforward changes, then we can merge this. Apologies for the delayed review!

mlx/random.cpp
arn4: Yep, I meant ``truncated_mean_shape``

mlx/random.cpp
arn4: Yes!

awni (approved):
Looks great, thanks for the updates! Will merge when tests clear.

PR #880, opened 2024-03-22
Increase test tolerance for fast.layer_norm

angeloskath:
I wrote a comparison with numpy using double to do the computation. The error of the reference implementation and the kernel is pretty much the same with the kernel winning ~50% of the time. For float32 it is well below 1e-6 for small vectors so I think it is fine to just increase the tolerance a bit.


awni (approved):
Thanks!!

PR #881, opened 2024-03-22
Fix copy donation and add partial rope

angeloskath:
As the title says, it makes sure that when the input can be donated and the types are the same no kernel is ran for copy. Also adds partial application of RoPE which would go to the fallback implementation. Finally, fixes the vjp of scatter which I randomly saw that it was doing a bunch of things wrongly (multiply with mask but also used where, and didn't use streams correctly).

Phi inference which uses partial RoPE application goes from 56 tps to >70 for 256 tokens.

### DIFF
--- a/benchmarks/python/rope_bench.py
+++ b/benchmarks/python/rope_bench.py
@@ -6,21 +6,21 @@ from time_utils import time_fn
 
 
 def time_rope():
-    rope = nn.RoPE(4096)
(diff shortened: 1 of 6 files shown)


jagrit06 (approved):
👍 Good find!

awni (approved):
🔥

awni:
A solid 14 TPS on phi-2, that's amazing!

PR #883, opened 2024-03-24
Implement vjps for some primitives in the fast namespace

angeloskath:
RoPE is conveniently a small change because we can define its gradient as a "weird" RoPE and call the same kernel. RMSNorm and LayerNorm are more complicated and require separate primitives and kernels. To avoid atomics which are horribly slow I am allocating an intermediate array for the gradient wrt `w` of the same size as the output. I also exposed the metal reduction functions so we can call them similarly to `copy_gpu` which has been very convenient.

- [x] RoPE
- [x] RMSNorm
- [x] LayerNorm

#### RoPE micro benchmark

```
RoPE grad before: 24.5 ms
RoPE grad now:    3.7 ms
```

#### RMSNorm micro benchmark

```
Expanded implementation:          164.99797 msec
Expanded implementation compiled: 108.32838 msec
mx.fast:                          26.47019 msec
mx.fast compiled:                 26.22546 msec
```

#### LayerNorm micro benchmark

```
Expanded implementation:          235.18537 msec
Expanded implementation compiled: 170.20382 msec
mx.fast:                          30.50168 msec
mx.fast compiled:                 30.29384 msec
```

### DIFF
--- a/mlx/fast.cpp
+++ b/mlx/fast.cpp
@@ -131,12 +131,10 @@ std::vector<array> RMSNorm::vjp(
     std::vector<int> axes(g.ndim() - 1);
     std::iota(axes.begin(), axes.end(), 0);
     vjps.push_back(
-        sum(multiply(g, multiply(x, n, s), s), axes, /* keepdims= */ true, s));
+        sum(multiply(g, multiply(x, n, s), s), axes, /* keepdims= */ false, s));
 
     return vjps;
   };
-
-  // TODO: Check that primals, output and cotangents have the same dtype.
 
   auto vjps = array::make_arrays(
       {primals[0].shape(), primals[1].shape()},
--- a/mlx/fast.cpp
+++ b/mlx/fast.cpp
@@ -239,8 +239,87 @@ array layer_norm(
   return fallback({x, passed_weight, passed_bias})[0];
 }
 
+std::vector<array> LayerNorm::vjp(
+    const std::vector<array>& primals,
+    const std::vector<array>& cotangents,
+    const std::vector<int>& argnums,
+    const std::vector<array>& outputs) {
+  assert(primals.size() == 3);
+  assert(outputs.size() == 1);
+  assert(cotangents.size() == 1);
+
+  auto s = stream();
+  auto fallback = [eps = eps_, s](const std::vector<array>& inputs) {
+    auto& x = inputs[0];
+    auto& w = inputs[1];
+    auto& b = inputs[2];
+    auto& g = inputs[3];
+
+    std::vector<array> vjps;
+
+    auto norm = number_of_elements(x, {-1}, true, x.dtype(), s);
+    auto sumx = sum(x, /* axis= */ -1, /* keepdims= */ true, s);
+    auto sumx2 = sum(square(x, s), /* axis= */ -1, /* keepdims= */ true, s);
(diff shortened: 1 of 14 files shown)


mlx/fast.cpp
```diff
@@ -94,11 +97,71 @@ array rms_norm(
...
+    const std::vector<array>& primals,
+    const std::vector<array>& cotangents,
+    const std::vector<int>& argnums,
+    const std::vector<array>& outputs) {
+  assert(primals.size() == 2);
+  assert(outputs.size() == 1);
+  assert(cotangents.size() == 1);
+
+  auto s = stream();
+  auto fallback = [eps = eps_, s](const std::vector<array>& inputs) {
```
awni, line 110: This is a genuine question because it's bending my brain a bit. But do you need to check the `argnums` in the fallback as well and only return the requested grads?

angeloskath: Don't think so. The fallback implements the Primitive so it should return all of the outputs given the inputs.

Say for instance we want second order gradients. Then we 'll run the vjp of LayerNormVJP and get the gradients wrt all inputs (including the cotangent). Then we will select the one that was requested in argnums.

awni, line 110: Unrelated question, do you think we should implement these in terms of the VJP of the original fallback?

angeloskath: Yeah I did it for efficiency purposes indeed. I first tested with just the fallback and then implemented the kernel. It was the natural order of things anyway since I first did the vjp by hand then wrote it in normal code, tested that it is correct (iterate these two steps 50 times, damn it had been a while since I computed complicated gradients) then wrote it in a kernel.

I was thinking what would be a good way to automatically test those. Perhaps vmap a singleton dimension in the beginning?

awni:
Some benchmarks for (Q)LoRA  training:

Model | Pre Toks/sec | Post Toks/sec | Pre Memory | Post Memory
----- | ------ | ------ | ---- | -----
Phi-2 | 10086 |1208.227 | 6.786 | 6.824
4-bit Phi-2 |855.994 |918.688 | 3.289 |3.321
Mistral | 502.786 | 519.968 | 15.581 | 15.611
4-bit Mistral | 442.8 | 454.4 | 6.093 | 6.122
4-bit Mixtral | 132.243 | 135.444 | 30.341 | 30.341
4-bit Qwen1.5 72B | 48.046 | 48.137 | 41.851 | 41.851

mlx/fast.cpp
awni: >I was thinking what would be a good way to automatically test those. Perhaps vmap a singleton dimension in the beginning?

CPU?

awni: Maybe we do test them already when we run the CPU tests actually 🤔

awni: Yea a vmap over the VJP might work. Or just a vjp of the vjp.

angeloskath: Yeah exactly. Vmap causes problems cause scatter lacks a vmap (which we should fix).

awni: Hmm that reminds me of a discussion I forgot about 😬

awni (approved):
LGTM, thanks for adding these. The benchmarks I ran all had good losses after 200 iterations so I think we are good.

Issue #866, opened 2024-03-21
comparing python list with `mx.array`


amirhossein-razlighi:
Right now, comparing an `mx.array` instance with a python list, encounters error. There's a need for fixing this to pursue the common convention of numpy and PyTorch, so converting from list to `mx.array` is done implicitly (and is not needed to be done explicitly each time).
If this is a valid requirement, please let me know. I'd be happy to help fix this and make a contribution.
Thanks!

awni:
In general NumPy let's you use lists in many places which implicitly convert to `np.ndarray`.

```
>>> np.array([1, 2]) == [1,3]
array([ True, False])
```

In Torch that's more like an object comparison:

```
>>> torch.tensor([1, 2]) == [1,3]
False
```

And in MLX it just throws. I don't think we are going to implicitly convert from list to `mx.array` in our ops. Given that, the `torch` version probably makes the most sense: `__eq__` should not crash if it receives a non`mx.array` it should just return `False`.

PR #886, opened 2024-03-24
Comparing python objects such as list and tuple with `mlx.core.array` (#866)

amirhossein-razlighi:
## Proposed changes

This pull request is related to issue #866 . When trying to compare python list or tuple with an `mlx.core.array`, mlx throws exception. Right now, regarding the comments in the related issue, mlx returns False (following the way it is done in PyTorch). An example is as follows:
```
import mlx.core as mx

a = mx.array([1, 2, 3])
a == [1, 2, 3] # False
a == mx.array([1,2,3]) # True
a != [1, 2, 3] # True
a == (1, 2) # False
```

## Checklist

### DIFF
--- a/tests/compile_tests.cpp
+++ b/tests/compile_tests.cpp
@@ -214,9 +214,10 @@ TEST_CASE("test simplify multi output") {
   // is respected in simplification
   {
     auto e = compile(multi_three)({})[0];
+    CHECK_EQ(e.inputs().size(), 4);
+    CHECK_EQ(e.inputs().at(0).id(), e.inputs().at(2).id());
+    CHECK_EQ(e.inputs().at(1).id(), e.inputs().at(3).id());
     CHECK(array_equal(e, array({0.0f, 1.0f, 0.0f, 1.0f})).item<bool>());
-    CHECK_EQ(e.inputs()[0].id(), e.inputs()[2].id());
-    CHECK_EQ(e.inputs()[1].id(), e.inputs()[3].id());
   }
   set_compile_mode(CompileMode::enabled);
 }


python/src/array.cpp
```diff
@@ -770,37 +770,90 @@ void init_array(nb::module_& m) {
...
+                return false;
+              }
+            }
             return less(a, to_array(v, a.dtype()));
           },
           "other"_a)
       .def(
           "__le__",
-          [](const array& a, const ScalarOrArray v) {
+          [](const array& a,
```
awni, line 803: Let's not enable comparison to arbitrary objects for all of these overloads, it does not make sense to do `a < "hello"`, that is fine if it raises an exception.

python/src/array.cpp
```diff
@@ -770,37 +770,90 @@ void init_array(nb::module_& m) {
...
+             const ScalarOrArray v) -> std::variant<array, bool> {
+            auto check_for_obj = std::get_if<nb::object>(&v);
+            if (check_for_obj) {
+              // return true in case of object inequlity which is not mlx array
+              if (!isMlxCoreArray(*check_for_obj)) {
+                return true;
+              }
+            }
```
awni, line 856: This seems broken because it will no longer allow `==` between arrays and integers or floats.

awni (changes requested):
Thanks for getting this started. Left some comments inline, please address.

python/src/utils.h
```diff
@@ -44,6 +44,11 @@ inline array to_array_with_accessor(nb::object obj) {
   }
 }
 
+inline bool isMlxCoreArray(const nb::object& obj) {
+  std::string type_name = nb::type_name(obj.type()).c_str();
+  return type_name.find("mlx.core.array") != std::string::npos;
+}
```
awni, line 50: use `isinstance` for this, not a string comparison.

python/src/array.cpp
```diff
@@ -770,37 +770,90 @@ void init_array(nb::module_& m) {
...
           "__ne__",
-          [](const array& a, const ScalarOrArray v) {
+          [](const array& a,
+             const ScalarOrArray v) -> std::variant<array, bool> {
+            auto check_for_obj = std::get_if<nb::object>(&v);
+            if (check_for_obj) {
+              // return true in case of object inequlity which is not mlx array
+              if (!isMlxCoreArray(*check_for_obj)) {
+                return true;
+              }
```
awni, line 855: It might be easier with something like `to_array_or_bool` that returns a variant 🤔

python/src/array.cpp
amirhossein-razlighi: Sure. But isn't it better to raise a more meaningful exception? Right now, the exception occurs in `to_array` function, while trying to do `nb::cast`. The exception is something like this:
`Cannot cast type <?> to mlx array`
I will change the exception to a more meaningful one, showing that the comparison is wrong

python/src/array.cpp
amirhossein-razlighi: No. The equality between an array and a scalar still works. There's a test that I wrote and covers this aspect:
```
    def test_array_eq_scalar(self):
        a = mx.array([1, 2, 3])
        b = 1
        c = 4

        self.assertTrue(mx.any(a == b))
        self.assertFalse(mx.all(a == c))
```
These lines of code, check for if the argument is of type `object`, and if so, checks if it's of type `mlx.core.array`. If not, so it's a list, tuple or an instance of a class, compared to an array.

python/src/array.cpp
amirhossein-razlighi: I think I did not quite understand what you meant. I wrote a function `to_array_or_bool` in `utils.h` as you can see:
```
inline std::variant<array, bool> to_array_or_bool(
    const ScalarOrArray& v,
    std::optional<Dtype> dtype = std::nullopt) {
  // try to convert to array, if it fails, return false
  try {
    return to_array(v, dtype);
  } catch (const std::exception& e) {
    return false;
  }
}
```
Do you mean that we replace this, instead of the check that we are doing with `isMlxCoreArray`?

python/src/utils.h
amirhossein-razlighi: Done ✅

python/src/array.cpp
amirhossein-razlighi: I updated the code and throw this exception in these cases:
```
ValueError: Object comparison is not a valid operation on an mlx array.
```
I also added some tests to make sure it works properly.

awni: That's much nicer! I might go a step further and include the type name in the exception error message, that might be helfpul when debugging these.

python/src/array.cpp
```diff
@@ -770,37 +770,94 @@ void init_array(nb::module_& m) {
...
+            }
             return less_equal(a, to_array(v, a.dtype()));
           },
           "other"_a)
       .def(
           "__gt__",
-          [](const array& a, const ScalarOrArray v) {
+          [](const array& a,
+             const ScalarOrArray v) -> std::variant<array, bool> {
```
awni, line 821: Please revert these to simply returning `array` not a variant.

python/src/array.cpp
awni: Yes exactly.

python/src/utils.h
```diff
@@ -44,6 +44,10 @@ inline array to_array_with_accessor(nb::object obj) {
   }
 }
 
+inline bool isMlxCoreArray(const nb::object& obj) {
+  return nb::hasattr(obj, "__mlx_array__") || nb::isinstance<array>(obj);
+}
```
awni, line 49: Remove.

python/src/array.cpp
awni: I mean it doesn't really make sense but I do see that you are right. I would have thought scalars were considered `nb::object` as well, that is surprising..

awni: Evidently its not so difficult to check if something is convertible to an `array` so I might actually do the check explicitly rather than use the variant return type. That may be a premature optimization.

But I would prefer you have a standalone function in `utils.h` like this:

```
bool is_convertible_to_array(const ScalarOrArray& v) {
 // Do the check here and return true/false
}
```

python/src/array.cpp
amirhossein-razlighi: I see. but isn't it a little bit confusing while debugging? Because suppose one is doing this operation:
```
mlx.array([1, 2]) > [1,2]
```
and receives:
```
ValueError: Invalid type  list received in array initialization.
```
while we are not actually initialising anything here! So, the error message should somehow represent that this operation is not valid. Don't you agree?

awni: Indeed I see your point. But its an issue for all of our ops which take `ScalarOrArray` (which are many) so it doesn't make sense to special case these. We can try to find a way to give more useful error messaging for all cases as a follow up / separate issue.

amirhossein-razlighi: I think we need a more meaningful exception for all __operations__ which are invalid on an `mlx.core.array`. E.g:
```
mx.array([0.5, 0.75]) % [1, 2]
# OR
mx.array([0.5, 0.75]) * (1, 2)
# OR ...
```
Do you think this is a valid concern? and if so, how do you propose to pursue this?

awni: Maybe a minor concern.. I would open an issue for it so we can track / discuss, but it's beyond the scope of this PR so let's move it to an issue. We can also discuss possible fixes there.

python/src/array.cpp
amirhossein-razlighi: Done ✅

python/src/array.cpp
amirhossein-razlighi: Sure, great!

python/src/array.cpp
```diff
@@ -770,37 +770,47 @@ void init_array(nb::module_& m) {
...
             return greater_equal(a, to_array(v, a.dtype()));
           },
           "other"_a)
       .def(
           "__ne__",
-          [](const array& a, const ScalarOrArray v) {
+          [](const array& a,
+             const ScalarOrArray v) -> std::variant<array, bool> {
+            bool is_convertable = is_convertable_to_array(v);
+            if (!is_convertable) {
```
awni, line 811:
```suggestion
            if (!is_convertable_to_array(v)) {
```

python/src/utils.h
```diff
@@ -44,6 +44,23 @@ inline array to_array_with_accessor(nb::object obj) {
...
+    return true;
+  } else if (auto pv = std::get_if<nb::int_>(&v); pv) {
+    return true;
+  } else if (auto pv = std::get_if<nb::float_>(&v); pv) {
+    return true;
+  } else if (auto pv = std::get_if<std::complex<float>>(&v); pv) {
+    return true;
+  } else if (auto pv = std::get_if<nb::object>(&v); pv) {
+    return nb::hasattr(*pv, "__mlx_array__") || nb::isinstance<array>(*pv);
+  }
```
awni, line 60: You had a different version earlier which just pulled out an object if it existed. So if it's not an `nb::oject` then it's ok (int, bool, float, complex) and if it is then check if its an array or has the attribute? Maybe we should do that instead?

awni:
Looks really nice, thank you! I left a couple comments. Could you please take a look and then we can run tests and merge?

python/src/utils.h
amirhossein-razlighi: It's the same line that is present here:
```
else if (auto pv = std::get_if<nb::object>(&v); pv) {
    return nb::hasattr(*pv, "__mlx_array__") || nb::isinstance<array>(*pv);
  }
```
The function that we had earlier (`isMlxCoreArray`) checked this line. I think right now it's better and we may use this utility function further in more operations (where we need to check if the operands are valid and can be converted to `mlx.core.array`).

awni: Ok just thinking about it from an efficiency standpoint. Can you not do?

```
if (auto pv = std::get_if<nb::object>(&v); pv) {
    return nb::isinstance<array>(*pv) || nb::hasattr(*pv, "__mlx_array__");
} else {
    return true;
}
```

Also it might be minor but I would swap the `hasattr` and the `isinstance` since the former will be much much more common (and also probably a lot less expensive).

amirhossein-razlighi: I swapped the `hasattr` and `isinstance`. Also, why should we add `else { return true; }`? Because if it's not each of the options mentioned above (bool/int/float/complex/array) it is not convertible to array and thus, we should simply return `False`.

amirhossein-razlighi: please check the lates commit and let me know about what you think should be changed

amirhossein-razlighi: Also, I think we need a better name for this function. Because, instances like `list` and `tuple` are actually convertible to `mlx.core.array`. They are not __comparable__ to arrays. Don't you agree? This name might be a little misleading in further developments

PR #889, opened 2024-03-24
post nanobind docs fixes and some updates

### LINKED ISSUE #884: [BUG] No auto-generated stub files for `mlx.core` for `0.8.0`
**Describe the bug**
When I install mlx from pip, the latest version `0.8.0` does not have auto-generated stub files for `mlx.core`, thus there is no type hint in VSCode. When I install version `0.7.0`, there are stub files. It's probably a bug introduced when mlx switched to nanobind.

A quick GitHub search shows that there is indeed a command that generates stub files, but perhaps there is a bug in the config somewhere that the stub files are not included in the pip package.

https://github.com/ml-explore/mlx/blob/be98f4ab6be2e466e2ebf59622e5f4866b6527e2/setup.py#L136-L139
https://github.com/ml-explore/mlx/blob/be98f4ab6be2e466e2ebf59622e5f4866b6527e2/.circleci/config.yml#L43-L47

**To Reproduce**
NA

**Expected behavior**
There are stub files.

 - OS Version: 14.2
 - Version 0.8.0

**Additional context**
NA

awni:
- Fix an issue with docs incorrectly rendering nanobind class methods
- Add signatures for random functions
- Add a few missing ops
- Fix stubgen (closes #884)
- Reduce `arange` to just two overloads, mostly for simplicity of docs. There is no way to disable signatures in nanobind.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -44,7 +44,7 @@ jobs:
           name: Generate package stubs
           command: |
             echo "stubs"
-            python -m nanobind.stubgen -m mlx.core -r -O python
+            python setup.py generate_stubs 
       - run:
           name: Run Python tests
           command: |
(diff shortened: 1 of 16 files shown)


angeloskath (approved):
Thanks for fixing these!

PR #890, opened 2024-03-24
Adding ConvLSTM Implementation

tomooga9:
## Proposed changes

Adding ConvLSTM layer to mlx.nn.layers

## Checklist


awni:
This is very cool! I think it's a bit too niche at the moment to include as a standlone layer in `mlx.nn`. I encourage you to use it and/or publish a package with something like this.

PR #891, opened 2024-03-25
Remove unnecessary string copies

zcbenz:
## Proposed changes

1. Use string_view instead of string when there is no need for copy.
2. Otherwise move string when possible.

## Checklist

### DIFF
--- a/mlx/io.h
+++ b/mlx/io.h
@@ -23,13 +23,13 @@ using SafetensorsLoad = std::pair<
 void save(std::shared_ptr<io::Writer> out_stream, array a);
 
 /** Save array to file in .npy format */
-void save(const std::string& file, array a);
+void save(std::string file, array a);
 
 /** Load array from reader in .npy format */
 array load(std::shared_ptr<io::Reader> in_stream, StreamOrDevice s = {});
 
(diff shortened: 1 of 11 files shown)


mlx/io.h
```diff
@@ -23,13 +23,13 @@ using SafetensorsLoad = std::pair<
...
 /** Save array to file in .npy format */
-void save(const std::string& file, array a);
+void save(std::string file, array a);
 
 /** Load array from reader in .npy format */
 array load(std::shared_ptr<io::Reader> in_stream, StreamOrDevice s = {});
 
 /** Load array from file in .npy format */
-array load(const std::string& file, StreamOrDevice s = {});
+array load(std::string file, StreamOrDevice s = {});
```
dc-dc-dc, line 32: How come these remain strings instead of swapping to string_view in other spots?

zcbenz: The definition of `load` in `mlx/io/load.cpp` passes the `file` to `io::Reader`, which takes it by value. So making `load` take a string by value can save a copy if rvalue is passed.

angeloskath (approved):
This looks great! Thanks.

Issue #885, opened 2024-03-24
[FEATURE] `mlx.nn.Module` to return `self` in some methods to support chaining


gau-nernst:
Currently many "self-modification" methods of `mlx.nn.Module` return None. If we return `self` instead, we can do chaining to improve user experience.

Some methods that would be useful to return self
- `.freeze()`, `.unfreeze()`
- `.train()`, `.eval()`

Some examples of better UX

```python
# create a model and put it in eval mode
model = nn.Transformer().eval()

# only train biases (example from MLX doc)
model.freeze().unfreeze(keys="bias")
```

awni:
This seems like a simple and nice usage improvement to me. I will mark it as an enhancement.

gau-nernst:
I can submit a PR for this if that's ok with you.

PR #897, opened 2024-03-25
Support Chaining for some of functionalities of `nn.Module` (#885)

amirhossein-razlighi:
## Proposed changes

Closes issue #885 . This PR adds chaining support for some of functionalities in `nn.Module`. In this PR, functions like `.freeze()`, `.update()`, `train()`, etc. return the instance of that Module, instead of returning None. So, operations like these are possible:
```
import mlx.core as mx
import mlx.nn as nn

m = nn.Sequential(nn.Linear(2, 2), nn.ReLU(), nn.Linear(2, 1))
m.freeze().unfreeze().train()
```

## Checklist

### DIFF
--- a/python/mlx/nn/layers/base.py
+++ b/python/mlx/nn/layers/base.py
@@ -566,7 +566,7 @@ class Module(dict):
                 Default: ``False``.
 
         Returns:
-            The module instance with unfreezed parameters.
+            The module instance with unfrozen parameters.
         """
 
         def _unfreeze_impl(_, m):
--- a/python/mlx/nn/layers/base.py
+++ b/python/mlx/nn/layers/base.py
@@ -146,7 +146,7 @@ class Module(dict):
         self,
         file_or_weights: Union[str, List[Tuple[str, mx.array]]],
         strict: bool = True,
-    ) -> self:
+    ):
         """
         Update the model's weights from a ``.npz``, a ``.safetensors`` file, or a list.
 
@@ -318,7 +318,7 @@ class Module(dict):
 
         return self.filter_and_map(self.valid_child_filter, is_leaf_fn=_is_leaf_module)
 
-    def update(self, parameters: dict) -> self:
+    def update(self, parameters: dict):
         """Replace the parameters of this Module with the provided ones in the
         dict of dicts and lists.
 
@@ -367,7 +367,7 @@ class Module(dict):
         self,
         map_fn: Callable[[mx.array], mx.array],
         filter_fn: Optional[Callable[["mlx.nn.Module", str, Any], bool]] = None,
-    ) -> self:
+    ):
         """Map all the parameters using the provided ``map_fn`` and immediately
         update the module with the mapped parameters.
 
@@ -386,7 +386,7 @@ class Module(dict):
         self.update(self.filter_and_map(filter_fn, map_fn))
         return self
 
-    def update_modules(self, modules: dict) -> self:
+    def update_modules(self, modules: dict):
         """Replace the child modules of this :class:`Module` instance with the
         provided ones in the dict of dicts and lists.
 
@@ -427,7 +427,7 @@ class Module(dict):
         apply(self, modules)
         return self
 
-    def apply_to_modules(self, apply_fn: Callable[[str, "mlx.nn.Module"], Any]) -> self:
+    def apply_to_modules(self, apply_fn: Callable[[str, "mlx.nn.Module"], Any]):
         """Apply a function to all the modules in this instance (including this
         instance).
 
@@ -482,7 +482,7 @@ class Module(dict):
         recurse: bool = True,
         keys: Optional[Union[str, List[str]]] = None,
         strict: bool = False,
-    ) -> self:
+    ):
         """Freeze the Module's parameters or some of them. Freezing a parameter means not
         computing gradients for it.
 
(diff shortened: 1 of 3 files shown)


python/mlx/nn/layers/base.py
```diff
@@ -146,7 +146,7 @@ def load_weights(
         self,
         file_or_weights: Union[str, List[Tuple[str, mx.array]]],
         strict: bool = True,
-    ):
```
awni, line 149: Why quote Module here? Can you not just do `-> Module` directly?

python/mlx/nn/layers/base.py
```diff
@@ -540,6 +564,9 @@ def unfreeze(
                 ``module.unfreeze(keys="bias")``.
             strict (bool, optional): If set to ``True`` validate that the passed keys exist.
                 Default: ``False``.
+
+        Returns:
+            The module instance with unfreezed parameters.
```
awni, line 569:
```suggestion
            The module instance with unfrozen parameters.
```

python/mlx/nn/layers/base.py
```diff
@@ -488,6 +508,9 @@ def freeze(
                 ``module.freeze(keys="bias")``.
             strict (bool, optional): If set to ``True`` validate that the passed keys exist.
                 Default: ``False``.
+
+        Returns:
+            The module instance with freezed parameters.
```
awni, line 513:
```suggestion
            The module instance with frozen parameters.
```

awni (commented):
Do you think this change introduces some ambiguity which might cause confusion?

Like:

`m1 = m.freeze()`

Does one expect that `m1` refers to the same module as `m`?

python/mlx/nn/layers/base.py
amirhossein-razlighi: No. If I do `-> Module` or `-> self`, since the class is not defined completely yet, it gives error `Module is not defined`. So, we should use python's "forward referencing" that is mentioning the class name inside quotes

amirhossein-razlighi:
I do not think this change, causes ambiguity. Since one may expect that every valid operation on an instance of `nn.Module` has the ability to support chaining. And some operations like `.freeze()` which return `None` before, right now return the module instance itself, so no conflicts are caused. And yes, I think in case of `m1 = m.freeze()`, one may expect `m1` refers to `m` module, but with frozen parameters.

amirhossein-razlighi:
> Do you think this change introduces some ambiguity which might cause confusion?
> 
> Like:
> 
> `m1 = m.freeze()`
> 
> Does one expect that `m1` refers to the same module as `m`?

I do not think this change, causes ambiguity. Since one may expect that every valid operation on an instance of `nn.Module` has the ability to support chaining. And some operations like `.freeze()` which return `None` before, right now return the module instance itself, so no conflicts are caused. And yes, I think in case of `m1 = m.freeze()`, one may expect `m1` refers to `m` module, but with frozen parameters.

python/mlx/nn/layers/base.py
awni: I see, that makes sense.

amirhossein-razlighi:
> > Do you think this change introduces some ambiguity which might cause confusion?
> > Like:
> > `m1 = m.freeze()`
> > Does one expect that `m1` refers to the same module as `m`?
> 
> I do not think this change, causes ambiguity. Since one may expect that every valid operation on an instance of `nn.Module` has the ability to support chaining. And some operations like `.freeze()` which return `None` before, right now return the module instance itself, so no conflicts are caused. And yes, I think in case of `m1 = m.freeze()`, one may expect `m1` refers to `m` module, but with frozen parameters.

Do you think it's okay? In this case, I think this PR is almost finished and ready to go 🚀

amirhossein-razlighi:
> > Do you think this change introduces some ambiguity which might cause confusion?
> > Like:
> > `m1 = m.freeze()`
> > Does one expect that `m1` refers to the same module as `m`?
> 
> I do not think this change, causes ambiguity. Since one may expect that every valid operation on an instance of `nn.Module` has the ability to support chaining. And some operations like `.freeze()` which return `None` before, right now return the module instance itself, so no conflicts are caused. And yes, I think in case of `m1 = m.freeze()`, one may expect `m1` refers to `m` module, but with frozen parameters.

@awni 
Do you think it's okay? In this case, I think this PR is almost finished and ready to go 🚀

python/mlx/nn/layers/base.py
```diff
@@ -159,6 +159,9 @@ def load_weights(
               only the weights actually contained in the model are loaded and
               shapes are not checked. Default: ``True``.
 
+        Returns:
+            The model with the updated weights.
```
angeloskath, line 163: To further reduce the risk of confusion perhaps we can reword this to "The module instance after updating the weights", just to reinforce that it is the same module being returned.

angeloskath (approved):
I think this looks good. I would reword the return docs but in general I don't think there will be much confusion.

awni:
I like @angeloskath suggestion, would you mind making the change @amirhossein-razlighi ? Then we can run the tests and merge.

python/mlx/nn/layers/base.py
amirhossein-razlighi: Sure. Thanks for the comment!

amirhossein-razlighi:
> I like @angeloskath suggestion, would you mind making the change @amirhossein-razlighi ? Then we can run the tests and merge.

Done ✅

angeloskath:
Sorry to be a stickler but can you also change the rest of the docs. ie "The module instance after updating the training mode".

amirhossein-razlighi:
> Sorry to be a stickler but can you also change the rest of the docs. ie "The module instance after updating the training mode".

Done ✅

awni:
Thanks for adding this!

Issue #898, opened 2024-03-25
Unsqueeze with None not working


abdussamettrkr:
Unsqueeze array with None is not working in some conditions.

**To Reproduce**

```python
import mlx.core as mx
print(mx.ones((5,5,5))[:,None].shape)
```
prints: (5, 5, 5)

In Numpy

```python
import numpy as np
print(np.ones((5,5,5))[:, None].shape)
```
prints: (5, 1, 5, 5)

**Expected behavior**
Same shape with Numpy.

 - OS Version: MacOS 14.3.1
 - Version: 0.8.1

awni:
Looks like a bug. @jagrit06 can you check it? I guess related to https://github.com/ml-explore/mlx/pull/879 as it was working in 0.8.0 but broken on main. Probably good to add a test case as well.

abdussamettrkr:
Can I solve this bug? I found the problem, and solved. Pushing the code

awni:
Of course!

PR #899, opened 2024-03-25
Fix unsqueeze with None

abdussamettrkr:
## Proposed changes
 Fix bug unsqueeze with None #898.

## Checklist

### DIFF
--- a/python/src/indexing.cpp
+++ b/python/src/indexing.cpp
@@ -367,6 +367,7 @@ array mlx_get_item_nd(array src, const nb::tuple& entries) {
   }
 
   bool squeeze_needed = false;
+  bool unsqueeze_needed = false;
 
   // Slice handling
   {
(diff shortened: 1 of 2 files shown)


python/src/indexing.cpp
```diff
@@ -400,6 +400,12 @@ array mlx_get_item_nd(array src, const nb::tuple& entries) {
     src = slice(src, starts, ends, strides);
   }
 
+  bool none_exists = false;
+  for (auto idx : remaining_indices) {
+    if (idx.is_none())
+      none_exists = true;
+  }
```
jagrit06, line 407: Instead of having this as a separate loop, you can integrate it into the previous loop to check for Nones

python/src/indexing.cpp
```diff
@@ -418,6 +424,15 @@ array mlx_get_item_nd(array src, const nb::tuple& entries) {
...
     src = reshape(src, out_shape);
+  } else if (none_exists) {
+    std::vector<int> out_shape = src.shape();
+    int shifted = 0;
+    for (size_t i = 0; i < remaining_indices.size(); i++) {
+      if (remaining_indices[i].is_none()) {
+        out_shape.insert(out_shape.begin() + (shifted++) + i, 1);
+      }
+    }
+    src = reshape(src, out_shape);
```
jagrit06, line 435: We can just enter the reshape loop above without the need for another branch 
What caused the bug is that is previously the expansion of ellipsis would also pad the end of the array with none slices (hence remaining indices would cover all dimensions for the src making the unsqueezed check be `remaining_indices.size() > src.ndim()`)

You can juts check if a unsqueeze is needed in the loop that prepares the slices and then enter the reshape branch above

jagrit06 (changes requested):

jagrit06:
You can just apply this quick patch! Thanks for finding the issue and working on it!
[a.patch](https://github.com/ml-explore/mlx/files/14746982/a.patch)

abdussamettrkr:
> You can just apply this quick patch! Thanks for finding the issue and working on it! [a.patch](https://github.com/ml-explore/mlx/files/14746982/a.patch)

Oh I see, thats cool.

jagrit06 (approved):
Thanks!

awni:
Could you please run the formatting hooks, looks like they failed the test

awni:
Closes #898

PR #901, opened 2024-03-26
Make MLX build on x64 macOS

zcbenz:
## Proposed changes

The arm64 macbook pros are heavy and I usually use my intel one for mobile, it would be nice if I can play with MLX on it.

To build with x64, user must pass `MLX_ENABLE_X64_MAC` to cmake:
`CMAKE_ARGS='-DMLX_ENABLE_X64_MAC=ON' python setup.py`

## Checklist

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -15,6 +15,7 @@ option(MLX_BUILD_EXAMPLES "Build examples for mlx" ON)
 option(MLX_BUILD_BENCHMARKS "Build benchmarks for mlx" OFF)
 option(MLX_BUILD_PYTHON_BINDINGS "Build python bindings for mlx" OFF)
 option(MLX_BUILD_METAL "Build metal backend" ON)
+option(MLX_ENABLE_X64_MAC "Enable building for x64 macOS" OFF)
 option(BUILD_SHARED_LIBS "Build mlx as a shared library" OFF)
 
 if(NOT MLX_VERSION)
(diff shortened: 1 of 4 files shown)


python/tests/test_ops.py
```diff
@@ -1762,8 +1762,8 @@ def test_repeat(self):
         )
 
     def test_tensordot(self):
-        # No fp16 matmuls on linux
-        if self.is_linux:
+        # No fp16 matmuls on cpu
+        if mx.default_device() == mx.cpu:
```
awni, line 1766: I don't think that's right actually, we can do fp16 with BNNS on the CPU 🤔 , we shouldn't stop testing that.

It might be good to check the platform instead.

python/tests/test_fast_sdpa.py
```diff
@@ -62,7 +62,7 @@ def test_fast_sdpa(self):
         B = 1
         H = 32
         dtypes = [np.float32]
-        if not self.is_linux:
+        if mx.default_device() == mx.gpu:
             dtypes.append(np.half)
```
awni, line 66: Same comment as below.

CMakeLists.txt
```diff
@@ -23,23 +23,23 @@ endif()
...
+    if(${CMAKE_HOST_SYSTEM_PROCESSOR} MATCHES "arm64" AND ${CMAKE_HOST_APPLE})
+      message(FATAL_ERROR
+        "Building for x86_64 on arm64 macOS is not supported."
+        " If you are on an Apple silicon system, check the build"
+        " documentation for possible fixes: "
+        "https://ml-explore.github.io/mlx/build/html/install.html#build-from-source")
+    else()
+      message(WARNING "Building for x86_64 arch is not officially supported.")
+    endif()
+    set(MLX_BUILD_METAL OFF)
```
awni, line 41: I'm cool enabling an x86 build if it's useful for you. That said, I think we have to make it explicitly opt-in. Too many people were accidentally building for x86 (e.g. through rosetta) before and the warning was not sufficient to alert them to the fact that their environment was bad.

What we could do is have a flag that you have to set with cmake asking for x86 `MLX_ENABLE_X86` or something..

awni (changes requested):
Have you tried running the tests on an x86 mac?  I am not sure they will pass.. 

Maybe you could check and see if it actually works. If it works out of the box with this change that would be nice.. o/w we probably won't to spend time supporting old x86 arches.

python/tests/test_ops.py
zcbenz: This is where throws:
https://github.com/ml-explore/mlx/blob/28fcd2b519f0fabcd681f2c33e14d71983cad819/mlx/backend/common/default_primitives.cpp#L169-L176

CMakeLists.txt
zcbenz: If there are people running cmake with rosetta, then the `CMAKE_HOST_SYSTEM_PROCESSOR` check won't work. ☹️ 

I have added a `MLX_ENABLE_X64_MAC` switch.

zcbenz:
> Have you tried running the tests on an x86 mac? I am not sure they will pass..

The python and C++ tests pass on my intel macbook pro.

python/tests/test_ops.py
awni: Oh I know that on the common backend there is no fp16 but on the accelerate backend for the CPU (which is what is used on Apple silicon devices) we support fp16 and that test should run on the CPU:

https://github.com/ml-explore/mlx/blob/main/mlx/backend/accelerate/matmul.cpp#L201-L206

awni: You could set a variable:

```
@property
def is_apple_silicon(self):
    return platform.machine() == "arm64" and platform.system() == "Darwin"
```

zcbenz: Ah I forgot the Accelerate backend, I have updated the code.

awni (approved):
Looks great, thanks!

PR #903, opened 2024-03-26
Fix nan and improve speed for qvm

### LINKED ISSUE #896: [BUG] quantized_matmul gives non reproducible NaNs
**Describe the bug**
I have a model using quantized_matmuls. It works fine in inference. And it works in when I fine-tune the lora weights in that model in the first iteration, but starting from the second iteration, loss becomes NaN. 

I cache intermediate values and set a breakpoint after the optimizer update. I found the output of the first quantuzed_matmul is NaN (the second operation in my model). Now I start trying different stuff in this pdb session:
1. I feed the same inputs (including parameters in the QuantizedLinear module) to mx.quantized_matmul using cpu stream, it works fine. 
2. I dump the inputs to file and load bad, and run it on gpu, it works fine as well. 
3. If I still use the same inputs and run quantized_linear on gpu again, NaNs still.

**To Reproduce**

It's hard to reproduce since the project is a bit complicated.. And as mentioned previously, it cannot be reproduced with standalone mx.quantized_matmul with saved tensors. However, it is reproducible consistently with my training script.

I will appreciate any general suggestions or tips to further debug this issue.

For example, this symptom seems to suggest that there is some state/cache in the gpu pipeline? If so, maybe there is a way to "force reset" any possible state?

Thank you in advance!

**Expected behavior**
Identical or similar result as the cpu quantized_matmul.

 - OS Version: [MacOS 14.1]
 - Version [0.8.0]

**Additional context**
I'm using float16 for inputs and non-quantized parameters.

angeloskath:
This fixes #896 and while I was at it I simplified the `qvm` kernel quite a bit as a first optimization pass. Currently some cases are quite a bit faster while others are a bit slower than before. I think this can be merged and I will do some more optimization to make it purely faster than before.

The bug, btw, had to do with accumulating in `float16`. Just changing that to `float` would fix the bug.

#### Before
```
python benchmarks/python/comparative/bench_mlx.py quant_matmul_64_4 --size 1x1024 --size 1024x128 --size 1024x16 --size 1024x16 --dtype float16 --dtype uint32 --dtype float16 --dtype float16
0.08292818069458008
python benchmarks/python/comparative/bench_mlx.py quant_matmul_64_4 --size 1x4096 --size 4096x512 --size 1024x64 --size 1024x64 --dtype float16 --dtype uint32 --dtype float16 --dtype float16
0.24095416069030762
```

#### After
```
python benchmarks/python/comparative/bench_mlx.py quant_matmul_64_4 --size 1x1024 --size 1024x128 --size 1024x16 --size 1024x16 --dtype float16 --dtype uint32 --dtype float16 --dtype float16
0.04205632209777832
python benchmarks/python/comparative/bench_mlx.py quant_matmul_64_4 --size 1x4096 --size 4096x512 --size 1024x64 --size 1024x64 --dtype float16 --dtype uint32 --dtype float16 --dtype float16
0.09795498847961426
```

### DIFF
--- a/mlx/backend/metal/kernels/quantized.metal
+++ b/mlx/backend/metal/kernels/quantized.metal
@@ -96,6 +96,38 @@ inline U qdot(const device uint8_t* w, const thread U *x_thread, U scale, U bias
   return scale * accum + sum * bias;
 }
 
+template <typename U, int values_per_thread, int bits>
+inline void qouter(const thread uint8_t* w, U x, U scale, U bias, thread U* result) {
+  static_assert(bits == 2 || bits == 4 || bits == 8, "Template undefined for bits not in {2, 4, 8}");
+
+  if (bits == 2) {
+    U s[4] = {scale, scale / 4.0f, scale / 16.0f, scale / 64.0f};
+    for (int i = 0; i < (values_per_thread / 4); i++) {
+      result[4*i] += x * (s[0] * (w[i] & 0x03) + bias);
+      result[4*i+1] += x * (s[1] * (w[i] & 0x0c) + bias);
+      result[4*i+2] += x * (s[2] * (w[i] & 0x30) + bias);
+      result[4*i+3] += x * (s[3] * (w[i] & 0xc0) + bias);
+    }
+  }
+
+  else if (bits == 4) {
+    const thread uint16_t* ws = (const thread uint16_t*)w;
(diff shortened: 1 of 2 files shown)


awni:
What's going on with Circle.. all of a sudden a missing import 🤔 .. I wonder if it's due to a cmake upgrade.

awni (approved):
Thanks for the fix!

tyeah:
Thank you folks for the fast fix!

Issue #642, opened 2024-02-06
[BUG] mx.random.normal can produce -inf


davidkoski:
**Describe the bug**
mx.random.normal can produce -inf

**To Reproduce**

Include code snippet
```python
import mlx.core as mx

mx.random.seed(61635)
a = mx.random.normal([4, 3], dtype=mx.float16)

print(a)
```

Output is:

```
array([[-0.967285, -0.724121, 0.654785],
       [1.31836, 1.0498, -0.641113],
       [-0.123352, 0.152344, 0.524902],
       [-inf, 0.279053, -1.46973]], dtype=float16)
```

Notice the item in the lower left.

**Expected behavior**

Expected: no `-inf` in the result.

**Desktop (please complete the following information):**
 - OS Version: 14.2.1
 - Version 16750f3c51921e0a632071ff30532d9d6afa7fb4

**Additional context**

davidkoski:
Another seed that produces it:

```
MLXRandom.seed(53978)
```

awni:
Should be fixed in #904

PR #904, opened 2024-03-26
Fixing random.normal for half-precision dtype #642

arn4:
## Proposed changes

Fixing issue #642. The problem was this line:
```c++
auto low = array(std::nextafter(-1.0f, 0.0f), dtype);
```
```nextafter``` works on float, but when creating the array with half-precision the value is casted to `low=-1`, that ultimately leads to `-inf` when applying `erfinv`.

Note that this updates is probably breaking the seed reproducibility for `random.normal`, because now the uniform interval is shorter, so the generated values are slightly different. Here is the output of the code in issue #642 after the change:
```
array([[-0.96582, -0.723145, 0.655273],
       [1.31934, 1.0498, -0.640625],
       [-0.122742, 0.152954, 0.525391],
       [-3.48633, 0.279541, -1.46973]], dtype=float16)
``` 
while before it was
```
array([[-0.967285, -0.724121, 0.654785],
       [1.31836, 1.0498, -0.641113],
       [-0.123352, 0.152344, 0.524902],
       [-inf, 0.279053, -1.46973]], dtype=float16)
```
## Checklist

### DIFF
--- a/python/tests/test_random.py
+++ b/python/tests/test_random.py
@@ -96,6 +96,11 @@ class TestRandom(mlx_tests.MLXTestCase):
 
         self.assertEqual(mx.random.normal().dtype, mx.random.normal(dtype=None).dtype)
 
+        # Test not getting -inf or inf with half precison
+        for hp in [mx.float16, mx.bfloat16]:
+            a = abs(mx.random.normal(shape=(10000,), loc=0, scale=1, dtype=hp))
+            self.assertTrue(mx.all(a < mx.inf).item())
+
     def test_randint(self):
         a = mx.random.randint(0, 1, [])
         self.assertEqual(a.shape, ())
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: python/tests/test_random.py [hunk 1]
```diff
@@ -96,6 +96,11 @@ def test_normal(self):
 
         self.assertEqual(mx.random.normal().dtype, mx.random.normal(dtype=None).dtype)
 
+        # Test not getting -inf or inf with half precison
+        for hp in [mx.float16, mx.bfloat16]:
+            a = abs(mx.random.normal(shape=(10000,), loc=0, scale=1, dtype=hp))
+            self.assertTrue(mx.all(a < mx.inf).item())
```
awni, line 102:
```suggestion
            self.assertTrue(mx.all(a < mx.inf))
```

awni (approved):
Thank you!

PR #905, opened 2024-03-26
Pick up previously set attributes

awni:
Closes https://github.com/ml-explore/mlx-examples/issues/625


angeloskath (approved):
Thanks for the fix.

This is getting tricky, I think we won't be able to avoid doing a custom `nn.Module` that does't subclass dict 🤔.

Issue #882, opened 2024-03-23
[BUG] GPU Sorting Failure for Arrays Exceeding 32768 Elements on any dimension


jj-curious:
**Describe the bug**
I've encountered this strange bug where sorting an array using `mx.sort` or `mx.argsort` is broken if:
- any of the axis has more than 2^15 = 32768 elements (e.g. `shape=(32769, )`)
- **and** device is set to GPU.

This issue does not occur for arrays with 32768 or fewer elements on any of its axis or when using the CPU instead of the GPU.

After numerous tests, it appears the problem is not related to the specific values within the array. Instead, the dimensions of the array and the chosen device are the primary factors influencing this issue.

**To Reproduce**

```python
import numpy as np
import mlx.core as mx


def compare_sort(shape: tuple[int, ...], device: mx.DeviceType) -> None:
    """
    Generate random samples of `shape` on `device` and compare sorting the array
    with mlx and numpy.    
    """
    mx.set_default_device(device)

    a = mx.random.normal(shape=shape)

    # verify samples do not contain nan or inf
    if not np.isfinite(a).all():
        raise ValueError("array has inf or nan value!")

    sorted_mx = mx.sort(a)
    sorted_np = mx.array(np.sort(a))

    all_equal = (sorted_mx == sorted_np).all().tolist()

    print(f"device:    {mx.default_device().type.name}")
    print(f"Shape:     {a.shape}")

    if not all_equal:
        print("-> mlx sorting differs from numpy! Last 5 elements:")
        print(f"mlx:       {sorted_mx[-5: ]}")
        print(f"numpy:     {sorted_np[-5: ]}")
    else:
        print("-> everything fine!")
```

Using GPU and a 1-dim array of size <= 2^15 works fine: 
```python
compare_sort(shape=(2**15,), device=mx.gpu)
```

device:    gpu
Shape:     (32768,)
-> everything fine!

But if the array size is > 2^15 sorting is broken if using GPU:
```python
compare_sort(shape=(2**15 + 1,), device=mx.gpu)
```

device:    gpu
Shape:     (32769,)
-> mlx sorting differs from numpy! Last 5 elements:
mlx:       array([3.50236, 3.55312, 3.59164, 3.81999, 1.85534], dtype=float32)
numpy:     array([3.45098, 3.50236, 3.55312, 3.59164, 3.81999], dtype=float32)

If switching to CPU everything is ok:
```python
compare_sort(shape=(2**15 + 1,), device=mx.cpu)
[...]

PR #906, opened 2024-03-26
Fix multiblock sort limits

jagrit06:
## Proposed changes

Fixes issue pointed out in #882 and added a similar test
The multi-block merge sort would exit one step early if the number of blocks were not a power of 2. This should fix that 

## Checklist


awni (approved):
🚀

Issue #833, opened 2024-03-14
[BUG] Convolution cannot work with all padding sizes


abdussamettrkr:
**Describe the bug**
I am attempting to use mx.conv2d with a padding size larger than the kernel size, but it's raising the error message '[conv] Padding sizes must be non-negative' in the vjp.

**To Reproduce**

Include code snippet
```python
import mlx.core as mx

inp = mx.random.normal((5, 11, 14, 2))
kernel = mx.random.normal((3, 3, 3, 2))
grad = mx.random.normal((5, 17, 20, 3))

def func(inp,kernel):
    return mx.conv2d(inp, kernel, stride=(1,1), padding=(4,4), dilation=1, groups=1)

x = mx.vjp(func, [inp, kernel], [grad])
```

Raises:
ValueError: [conv] Padding sizes must be non-negative. Got padding (-2,-2) | (-2,-2).

**Desktop (please complete the following information):**
 - OS Version:  MacOS 14.3.1
 - Version  0.6.0

PR #907, opened 2024-03-26
Implement negative padding in conv with slicing

jagrit06:
## Proposed changes

Fixes issue in #833
I think we can do the negative strides in the kernels, but for now I'm being careful and handling it through a slice

## Checklist

### DIFF
--- a/mlx/ops.cpp
+++ b/mlx/ops.cpp
@@ -2971,6 +2971,35 @@ array conv_general(
     input_dilation = std::vector<int>(spatial_dims, input_dilation_int);
   }
 
+  // Check for negative padding
+  bool has_neg_padding = false;
+  for (auto& pd : padding_lo) {
+    has_neg_padding = (pd < 0);
+  }
+  for (auto& pd : padding_hi) {
+    has_neg_padding = (pd < 0);
+  }
+
+  // Handle negative padding
+  if (has_neg_padding) {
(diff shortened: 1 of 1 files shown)


mlx/ops.cpp
```diff
@@ -2971,6 +2971,35 @@ array conv_general(
...
+        padding_lo[i] = 0;
+      }
+
+      if (padding_hi[i] < 0) {
+        stops[i + 1] += padding_hi[i];
+        padding_hi[i] = 0;
+      }
+    }
+
+    in = slice(in, starts, stops, s);
```
awni, line 3000:
```suggestion
    in = slice(in, std::move(starts), std::move(stops), s);
```

awni (approved):
Nice, just a minor comment since I'm on an overhead bender.

mlx/ops.cpp
```diff
@@ -2971,6 +2971,35 @@ array conv_general(
     input_dilation = std::vector<int>(spatial_dims, input_dilation_int);
   }
 
+  // Check for negative padding
+  bool has_neg_padding = false;
+  for (auto& pd : padding_lo) {
+    has_neg_padding = (pd < 0);
```
mlaves, line 2977: If the first padding value is negative, but any subsequent value is positive, `has_neg_padding` will be set to `false` again and the negative padding won't be handled.

PR #911, opened 2024-03-26
Fix race in multi-stream eval

awni:
Use the 0th output to get the id for dependencies rather than the primitive's id. The input's primitive may be deleted before the output array is scheduled


angeloskath (approved):
Looks good.

Issue #399, opened 2024-01-08
[Feature] Add Metal support for FFT


awni:
Add support for Metal backend with FFT primitive as mentioned here https://github.com/ml-explore/mlx-examples/issues/249

aneeshk1412:
I would like to take up this issue @awni. Please let me know if I can take this implementation up.

awni:
Do you have experience with GPU programming? This is not a trivial one so I would recommend starting with something simpler if not.

aneeshk1412:
I have some experience from my intern at Amazon HPC and a GPU programming course. I am not completely familiar with Metal, but I am looking at its documentation. I'm familiar with the basics of FFT. Would that be enough to start on this?

awni:
It's hard for me to answer. I would recommend you take a look at parallel implementations of FFT. There is also this [code](https://github.com/philipturner/metal-fft) which does FFT or Metal. Maybe a good place to start is to benchmark that code against a CPU implementation and just see how usable it might be?

We can use this thread to discuss your findings.

Rifur13:
What's the status on this? It's blocking some audio processing stuff I'm trying to do. It looks like there's MPS [implementation](https://developer.apple.com/documentation/metalperformanceshadersgraph/filtering_images_with_mpsgraph_fft_operations) - any interests in adding this?

A tangential question to this, why doesn't the codebase leverage MPS more? IIUC it handles broadcasting between matrices correctly and has optimized ops for lower dimensions.

AndreSlavescu:
Just to chime in, if this operator is implemented (when I eventually implement conv3d) I can accelerate the operation for larger input tensor shapes. FFT conv tends to outperform winowgrad conv for larger. Would be great to have as a feature to dispatch the appropriate conv kernel depending on shape.

adonath:
@AndreSlavescu See also: https://github.com/ml-explore/mlx/issues/811#issuecomment-1988934827

avinashahuja:
Could you reuse the code written for PyTorch? They added MPS support for FFT. https://github.com/pytorch/pytorch/pull/119670

awni:
It's not a great fit for us as we don't wrap MPSGraph and likely wouldn't make an exception for FFTs.

PR #915, opened 2024-03-27
Metal FFT for powers of 2 up to 2048

barronalex:
## Proposed changes

#399

Add a GPU FFT algorithm for the powers of 2 from 4 -> 2048.

Only supports 1D, forward, complex to complex transforms at the moment but planning to follow up with more features shortly.

## Performance

On my M1 Max:
fft_plot

For 64 <= N <= 512, we're doing ~360Gb/s with a large batch size which isn't far off from the maximum memory bandwidth of an M1 Max (~400Gb/s). The other sizes are slightly slower but can be addressed in a follow up PR.

The GPU implementation is 20 to 35 times faster than the CPU implementation for the FFT sizes it implements.

## Checklist

### DIFF
--- a/benchmarks/python/fft_bench.py
+++ b/benchmarks/python/fft_bench.py
@@ -1,4 +1,4 @@
-# Copyright © -2024 Apple Inc.
+# Copyright © 2024 Apple Inc.
 
 import matplotlib
 import mlx.core as mx
--- a/benchmarks/python/fft_bench.py
+++ b/benchmarks/python/fft_bench.py
@@ -1,0 +1,57 @@
+# Copyright © -2024 Apple Inc.
+
+import matplotlib
+import mlx.core as mx
+import numpy as np
+from time_utils import measure_runtime
+
+matplotlib.use("Agg")
+import matplotlib.pyplot as plt
+
+
+def bandwidth_gb(runtime_ms, system_size):
+    bytes_per_fft = np.dtype(np.complex64).itemsize * 2
+    bytes_per_gb = 1e9
+    ms_per_s = 1e3
+    return system_size * bytes_per_fft / runtime_ms * ms_per_s / bytes_per_gb
+
+
+def run_bench(system_size):
+    def fft(x):
+        out = mx.fft.fft(x)
+        mx.eval(out)
+        return out
+
+    bandwidths = []
+    for k in range(4, 12):
+        n = 2**k
+        x = mx.random.uniform(shape=(system_size // n, n)).astype(mx.float32)
+        x = x.astype(mx.complex64)
+        mx.eval(x)
+        runtime_ms = measure_runtime(fft, x=x)
+        bandwidths.append(bandwidth_gb(runtime_ms, system_size))
+
+    return bandwidths
+
+
+def time_fft():
+
+    with mx.stream(mx.cpu):
+        cpu_bandwidths = run_bench(system_size=int(2**22))
+
+    with mx.stream(mx.gpu):
+        gpu_bandwidths = run_bench(system_size=int(2**29))
+
+    # plot bandwidths
+    x = [2**k for k in range(4, 12)]
+    plt.scatter(x, gpu_bandwidths, color="green", label="GPU")
+    plt.scatter(x, cpu_bandwidths, color="red", label="CPU")
+    plt.title("MLX FFT Benchmark")
+    plt.xlabel("N")
+    plt.ylabel("Bandwidth (GB/s)")
+    plt.legend()
+    plt.savefig("fft_plot.png")
+
+
+if __name__ == "__main__":
+    time_fft()
--- a/mlx/backend/metal/fft.cpp
+++ b/mlx/backend/metal/fft.cpp
@@ -92,8 +92,8 @@ void FFT::eval_gpu(const std::vector<array>& inputs, array& out) {
 
     bool donated = in.data_shared_ptr() == nullptr;
     compute_encoder->setComputePipelineState(kernel);
-    set_array_buffer(compute_encoder, in_contiguous, 0);
-    set_array_buffer(compute_encoder, out, 1);
+    compute_encoder.set_input_array(in_contiguous, 0);
+    compute_encoder.set_output_array(compute_encoder, out, 1);
 
     auto group_dims = MTL::Size(1, m, 1);
     auto grid_dims = MTL::Size(batch, m, 1);
--- a/mlx/backend/metal/fft.cpp
+++ b/mlx/backend/metal/fft.cpp
@@ -84,7 +84,7 @@ void FFT::eval_gpu(const std::vector<array>& inputs, array& out) {
   size_t m = n / 4;
   size_t batch = in.size() / in.shape(axes_[0]);
 
-  auto compute_encoder = d.get_command_encoder(s.index);
+  auto& compute_encoder = d.get_command_encoder(s.index);
   {
     std::ostringstream kname;
     kname << "fft_" << n;
--- a/mlx/backend/metal/fft.cpp
+++ b/mlx/backend/metal/fft.cpp
@@ -1,5 +1,4 @@
 // Copyright © 2023 Apple Inc.
-
 #include "mlx/backend/metal/copy.h"
 #include "mlx/backend/metal/utils.h"
 #include "mlx/mlx.h"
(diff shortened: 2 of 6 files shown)


python/tests/test_fft.py
```diff
@@ -85,7 +77,33 @@ def test_fftn(self):
...
+    @unittest.skipIf(not mx.metal.is_available(), "Metal is not available")
+    def test_fft_gpu(self):
+        shape = (16, 4, 8)
+        # np.fft.fft always uses double precision complex128
+        # mx.fft.fft only supports single precision complex64
+        # hence the fairly tolerant equality checks.
+        atol = 1e-4
+        rtol = 1e-4
+        np.random.seed(7)
+        with mx.stream(mx.gpu):
```
awni, line 93: I would just run these tests on both devices rather than skipping them for the CPU and specifying the context manager. The goal is to remove the context manager for the CPU above once all the ops can run on the GPU.

awni: (usually we don't specify the device in the tests but just run the full test suite twice with the CPU and GPU as default).

benchmarks/python/fft_bench.py
```diff
@@ -0,0 +1,57 @@
+# Copyright © -2024 Apple Inc.
```
awni, line 1:
```suggestion
# Copyright © 2024 Apple Inc.
```

mlx/backend/metal/fft.cpp
```diff
@@ -1,12 +1,113 @@
...
+      in_contiguous.strides(),
+      in_contiguous.flags());
+
+  // Prod all the non-FFT dims to get the batch size
+  size_t batch = 1;
+  for (int axis = 0; axis < in.ndim(); axis++) {
+    if (axis != axes_[0]) {
+      batch *= in.shape(axis);
+    }
+  }
```
awni, line 87: Maybe just:

```suggestion
  size_t batch = in.size() / axes_[0];
```

mlx/backend/metal/fft.cpp
```diff
@@ -1,12 +1,113 @@
...
+  if (axes_.size() == 0 || axes_.size() > 1 || inverse_ ||
+      in.dtype() != complex64 || out.dtype() != complex64) {
+    // Could also fallback to CPU implementation here.
+    throw std::runtime_error(
+        "GPU FFT is only implemented for 1D, forward, complex FFTs.");
+  }
+
+  size_t n = in.shape(axes_[0]);
+
+  if (!is_power_of_2(n) || n > 2048 || n < 4) {
```
awni, line 25: Just curious, how difficult would it be to support non powers of 2? (We could easily pad with zeros.. but maybe there is a clean way to do it in the kernel itself?).

mlx/backend/metal/fft.cpp
```diff
@@ -1,12 +1,113 @@
...
+      flags.col_contiguous = true;
+      flags.row_contiguous = true;
+      for (int i = 0, ri = x.ndim() - 1; i < x.ndim(); ++i, --ri) {
+        flags.col_contiguous &= (strides[i] == f_stride || x.shape(i) == 1);
+        f_stride *= x.shape(i);
+        flags.row_contiguous &= (strides[ri] == b_stride || x.shape(ri) == 1);
+        b_stride *= x.shape(ri);
+      }
+      // FFT dim has stride 1 so there are no gaps
+      flags.contiguous = true;
```
awni, line 63: That doesn't seem quite right. Even if the FFT dim has stride 1, you could have gaps in due to another dimensions having a larger stride?

mlx/backend/metal/fft.cpp
```diff
@@ -1,12 +1,113 @@
...
+
+      x_copy.set_data(
+          allocator::malloc_or_wait(x.nbytes()), x.data_size(), strides, flags);
+      copy_gpu_inplace(x, x_copy, CopyType::GeneralGeneral, s);
+      copies.push_back(x_copy);
+      return x_copy;
+    }
+  };
+  const array& in_contiguous = check_input(inputs[0]);
```
awni, line 72: This array, is it actually contiguous or just contiguous in the FFT dim? If it's not truly row_contiguous, then we presumably need to specify the strides to the kernel?

barronalex:
If this goes in, I have a [follow up working](https://github.com/ml-explore/mlx/compare/main...barronalex:mlx:ab-metal-rfft?expand=1) that implements `ifft`, `rfft` and `rifft` with similar performance characteristics.

mlx/backend/metal/fft.cpp
djphoenix: @barronalex FWIW, it is indeed place where nextup operations on FFT result (like `.abs()`) cause fatal error. Replacing `in_contiguous.flags()` to empty is exact that makes my experimental code working. BTW big thank you for your work on Metal FFT.

barronalex: Thanks for catching this! Will push a fix shortly.

yury:
I wonder how pocketfft compared to Accelerate.framework vDSP_fft in terms of performance.
Or Accelerate framework uses pocketfft inside?

mlx/backend/metal/fft.cpp
```diff
@@ -1,12 +1,104 @@
...
+    throw std::runtime_error(
+        "GPU FFT is only implemented for the powers of 2 from 4 -> 2048");
+  }
+
+  // Make sure that the array is contiguous and has stride 1 in the FFT dim
+  std::vector<array> copies;
+  auto check_input = [this, &copies, &s](const array& x) {
+    // TODO: Pass the strides to the kernel so
+    // we can avoid the copy when x is not contiguous.
+    bool no_copy = x.strides()[axes_[0]] == 1 && x.flags().contiguous;
```
awni, line 34: I think you want `x.falgs().row_contiguous || x.flags().col_contiguous` here instead of `x.flags().contiguous` (which has a different meaning).

mlx/backend/metal/fft.cpp
awni: @barronalex would you mind adding a test or two with non-contiguous arrays (e.g. output of transpose or broadcast). The logic here is a bit subtle so some tests would be good.

awni:
This looks great and I think we can merge it soon! Just left a couple more comments. Let me know when you've addressed and will re-run tests.

barronalex:
@awni Thanks for the comments! I added some contiguity tests and changed some of the logic. Let me know if they look reasonable.

mlx/backend/metal/fft.cpp
```diff
@@ -1,12 +1,106 @@
...
+  auto compute_encoder = d.get_command_encoder(s.index);
+  {
+    std::ostringstream kname;
+    kname << "fft_" << n;
+    auto kernel = d.get_kernel(kname.str());
+
+    bool donated = in.data_shared_ptr() == nullptr;
+    compute_encoder->setComputePipelineState(kernel);
+    set_array_buffer(compute_encoder, in_contiguous, 0);
+    set_array_buffer(compute_encoder, out, 1);
```
awni, line 96:
```suggestion
    compute_encoder.set_input_array(in_contiguous, 0);
    compute_encoder.set_output_array(compute_encoder, out, 1);
```

awni: Sorry we changed that under your feet. That's the way to add input and output arrays to command encoders moving forward (we added a wrapper in MLX so we can control dependencies).

mlx/backend/metal/fft.cpp
```diff
@@ -1,12 +1,106 @@
...
+  auto compute_encoder = d.get_command_encoder(s.index);
+  {
+    std::ostringstream kname;
+    kname << "fft_" << n;
+    auto kernel = d.get_kernel(kname.str());
+
+    bool donated = in.data_shared_ptr() == nullptr;
+    compute_encoder->setComputePipelineState(kernel);
+    compute_encoder.set_input_array(in_contiguous, 0);
+    compute_encoder.set_output_array(compute_encoder, out, 1);
```
awni, line 96:
```suggestion
    compute_encoder.set_input_array(in_contiguous, 0);
    compute_encoder.set_output_array(out, 1);
```

awni (approved):
This is awesome. Let's get it landed and focus on additions in #981

mlx/backend/metal/fft.cpp
```diff
@@ -1,12 +1,106 @@
...
-
+#include "mlx/backend/metal/copy.h"
+#include "mlx/backend/metal/utils.h"
+#include "mlx/mlx.h"
 #include "mlx/primitives.h"
 
 namespace mlx::core {
 
 void FFT::eval_gpu(const std::vector<array>& inputs, array& out) {
+  auto& s = out.primitive().stream();
```
awni, line 10: `stream()` works here ;)

awni (approved):
I'm cool to land this and continue the discussion in #981

barronalex:
Sounds good to me!

PR #916, opened 2024-03-27
Use uintptr_t instead of size_t to store funtion id

zcbenz:
## Proposed changes

The `uintptr_t` is guaranteed to be able to hold a pointer but `size_t` is not. Also when I see a variable with `uintptr_t` type I know it means to be an ID, so code becomes easier to read.

Also does some small cleanup of the compile cache code.

## Checklist

### DIFF
--- a/mlx/compile.cpp
+++ b/mlx/compile.cpp
@@ -162,7 +162,6 @@ CompileMode& compile_mode() {
   return compile_mode_;
 }
 
-using CompileFn = std::function<std::vector<array>(const std::vector<array>&)>;
 using ParentsMap =
     std::unordered_map<std::uintptr_t, std::vector<std::pair<array, int>>>;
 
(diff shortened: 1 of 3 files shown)


mlx/compile.cpp
```diff
@@ -211,20 +211,20 @@ struct CompilerCache {
...
-    // Try to find the entry
-    auto [entry_it, inserted] = cache_.insert({fun_id, {}});
-    auto& entries = entry_it->second;
-    auto is_match = [shapeless](
-                        const std::vector<array>& in1,
-                        const std::vector<array>& in2) {
+    // Find the cache entries for |fun_id|.
+    std::vector<CacheEntry>& entries = cache_[fun_id];
+    // Compare if 2 arrays have same shape and dtype.
+    auto has_same_shape_and_dtype = [shapeless](
```
awni, line 221: Do you think this is the right name given that `shapeless` can be `True` in which case the shapes need not match?

zcbenz: It is not 100% correct, but easy to understand I think, like a `compare_string` with a `insensitive` parameter. I'm fine renaming it to something else.

awni (approved):
LGTM, thanks!

PR #917, opened 2024-03-28
Fix OOB read in qmv when non-divisible by blocksize

angeloskath:
As title suggests, it fixes out of bounds reads for `qmv` when the vector size is not divisible by 256 which is the block size for the small kernel. It adds ~4% in the micro-benchmark but I didn't really see any overhead in actual generation.

It also adds some code duplication which I hate but I think it is not too bad for now.

### DIFF
--- a/mlx/backend/metal/kernels/quantized.metal
+++ b/mlx/backend/metal/kernels/quantized.metal
@@ -15,14 +15,6 @@ using namespace metal;
 
 MLX_MTL_CONST int SIMD_SIZE = 32;
 
-template <typename T> struct AccT {
-  typedef T acc_t;
-};
-
-template <> struct AccT<bfloat16_t> {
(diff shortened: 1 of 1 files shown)


awni:
Closes https://github.com/ml-explore/mlx-examples/issues/452

awni:
Maybe worth adding a test for this / testing more unusual shapes since the dispatch is getting a bit complex? Or do you prefer to do that in some longer running CI thing?

awni (approved):
Thanks for the quick fix 🚀

angeloskath:
Yeah definitely longer running CI / fuzzing. It had to do a loop of 20 with sizes 2k x 5k and some additions etc otherwise it would not show up and even then it was dependent on the rest of the program ie running that test alone would not fail.

PR #919, opened 2024-03-28
Fixes #918 bug in compile_tests

### LINKED ISSUE #918: [BUG] Possible access of freed memory in compile_tests.cpp
**Describe the bug**

The `compile_tests.cpp` test passes but it is actually accessing freed memory.

**To Reproduce**

Add `CHECK_EQ(e.inputs().size(), 4);` to:

https://github.com/ml-explore/mlx/blob/d6112515024f930131134d0c0e91d50493349a0e/tests/compile_tests.cpp#L215-L220

Or just change `e.inputs()[0]` to `e.inputs().at(0)`.

And the test fails.

Crashes on macOS 14, both intel and apple silicon. And also Linux.

**Additional context**

It happens on latest main, and also 16546c70d8983e9860f7456c0dfc63176fc95568, didn't test further.

I'm not familiar with this part of code so I can't say whether it is a bug of the test or the library.

angeloskath:
Fixes #918 .

### DIFF
--- a/tests/compile_tests.cpp
+++ b/tests/compile_tests.cpp
@@ -214,9 +214,10 @@ TEST_CASE("test simplify multi output") {
   // is respected in simplification
   {
     auto e = compile(multi_three)({})[0];
+    CHECK_EQ(e.inputs().size(), 4);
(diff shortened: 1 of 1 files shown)


awni (approved):
🙏

PR #920, opened 2024-03-28
Comparing python objects (such as list/tuple) with `mlx.core.array`

amirhossein-razlighi:
## Proposed changes

This closes issue #866. When trying to compare python list or tuple with an mlx.core.array, mlx throws exception. Right now, regarding the comments in the related issue, mlx returns False (following the way it is done in PyTorch). An example is as follows:
```
import mlx.core as mx

a = mx.array([1, 2, 3])
a == [1, 2, 3] # False
a == mx.array([1,2,3]) # True
a != [1, 2, 3] # True
a == (1, 2) # False
```

## Checklist

### DIFF
--- a/ACKNOWLEDGMENTS.md
+++ b/ACKNOWLEDGMENTS.md
@@ -15,7 +15,7 @@ MLX was developed with contributions from the following individuals:
 - Hinrik Snær Guðmundsson: Added `atleast_1d`, `atleast_2d`, `atleast_3d` ops.
 - Luca Arnaboldi: Added `Ceil` and `Floor` ops; implemented pickling, copy and deepcopy for mlx arrays.
 - Brian Keene & Atila Orhon, with Argmax Inc.: Added `fast.scaled_dot_product_attention`
-- AmirHossein Razlighi: Added chaining support for some of the ops in `nn.Module`, Some Debug and Enhancements in `mlx.core.array`
+- AmirHossein Razlighi: Added chaining support for some of the ops in `nn.Module`. Comparison works for non array objects in `mlx.core.array`.
 <a href="https://github.com/ml-explore/mlx/graphs/contributors">
   <img class="dark-light" src="https://contrib.rocks/image?repo=ml-explore/mlx&anon=0&columns=20&max=100&r=true" />
 </a>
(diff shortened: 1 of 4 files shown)


amirhossein-razlighi:
I'm sorry. I accidentally closed [previous PR](https://github.com/ml-explore/mlx/pull/886#issue-2204299594) regarding to a hard reset of my main branch, and then I couldn't update that PR anymore.

amirhossein-razlighi:
In continue of your [latest review](https://github.com/ml-explore/mlx/pull/886#discussion_r1541844727):

I swapped the hasattr and isinstance. Also, why should we add else { return true; }? Because if it's not each of the options mentioned above (bool/int/float/complex/array) it is not convertible to array and thus, we should simply return False.
Also, I think we need a better name for this function. Because, instances like list and tuple are actually convertible to mlx.core.array. They are not comparable to arrays. Don't you agree? This name might be a little misleading in further developments

python/src/utils.h
```diff
@@ -44,6 +44,23 @@ inline array to_array_with_accessor(nb::object obj) {
...
+    return true;
+  } else if (auto pv = std::get_if<nb::int_>(&v); pv) {
+    return true;
+  } else if (auto pv = std::get_if<nb::float_>(&v); pv) {
+    return true;
+  } else if (auto pv = std::get_if<std::complex<float>>(&v); pv) {
+    return true;
+  } else if (auto pv = std::get_if<nb::object>(&v); pv) {
+    return nb::isinstance<array>(*pv) || nb::hasattr(*pv, "__mlx_array__");
+  }
```
awni, line 60: What I mean is that you replace this entire function with:
```c++
if (auto pv = std::get_if<nb::object>(&v); pv) {
    return nb::isinstance<array>(*pv) || nb::hasattr(*pv, "__mlx_array__");
} else {
    return true;
}
```

The only time it's not an object is when it's a python bool, int or complex.

amirhossein-razlighi: Done ✅

ACKNOWLEDGMENTS.md
```diff
@@ -15,7 +15,7 @@ MLX was developed with contributions from the following individuals:
 - Hinrik Snær Guðmundsson: Added `atleast_1d`, `atleast_2d`, `atleast_3d` ops.
 - Luca Arnaboldi: Added `Ceil` and `Floor` ops; implemented pickling, copy and deepcopy for mlx arrays.
 - Brian Keene & Atila Orhon, with Argmax Inc.: Added `fast.scaled_dot_product_attention`
-- AmirHossein Razlighi: Added chaining support for some of the ops in `nn.Module`.
+- AmirHossein Razlighi: Added chaining support for some of the ops in `nn.Module`, Some Debug and Enhancements in `mlx.core.array`
```
awni, line 18:
```suggestion
- AmirHossein Razlighi: Added chaining support for some of the ops in `nn.Module`. Comparisons works for non array objects in `mlx.core.array`.
```

awni (approved):
LGTM, thanks!

PR #921, opened 2024-03-28
[Test] Small vector optimization for array inputs

zcbenz:
This PR means to be used as an evaluation of the possible performance gain instead of something to be merged. Read this discussion to get the background: https://github.com/ml-explore/mlx/pull/895#issuecomment-2021606855.

In summary, using `std::vector` to store array has 2 problems:
1. `std::vector` does heap allocation for even 1 element.
2. aggregate constructor `std::vector{a1, a2, a3}` copies args.

I wrote a very trivial `small_vector` implementation, which stores elements on stack when aggregate constructor is used, and converts to `std::vector` when needed.

In this way, when using the aggregate constructor there is no heap allocation happened and the args are moved instead of copied.

There is no need to convert the whole codebase to use `small_vector` all at once, it is designed to work with existing code with minimal overhead, so we can replace the `std::vector<array>` with `small_vector<array>` piece by piece.

The overhead should be invisible: the `small_vector` is a thin wrapper of `std::variant<std::array, std::vector>`, and there is no penalty for accessing elements. When calling eval/vjp/jvp the `small_vector` is converted to `std::vector` by copying values from stack storage to `std::vector`, which is expensive but does not make things worse, because the copy would be done anyway when using the aggregate constructor of `std::vector`. And we should be able to remove most of the stack to vector copies eventually.

Currently `small_vector` is designed for computing graphs, but it should also be possible to be used for storing shapes.

There are some large changes in `array.h`, because to use `small_vector` the `ArrayDesc` must know the size of `array` so it can no longer be a sub-class in `array`, and since `ArrayDesc` becomes opaque to `array`, the `array`'s methods can no longer be defined inside the class definition and must be moved out.

### DIFF
--- a/mlx/array.cpp
+++ b/mlx/array.cpp
@@ -30,7 +30,7 @@ array::array(
     std::vector<int> shape,
     Dtype dtype,
     std::shared_ptr<Primitive> primitive,
-    std::vector<array> inputs)
+    small_vector<array> inputs)
     : array_desc_(std::make_shared<ArrayDesc>(
           std::move(shape),
           dtype,
@@ -161,7 +161,7 @@ void array::move_shared_buffer(array other) {
   move_shared_buffer(other, other.strides(), other.flags(), other.data_size());
 }
 
-void array::ArrayDesc::init() {
+void ArrayDesc::init() {
   strides.resize(shape.size());
   size = 1;
   for (int i = shape.size() - 1; i >= 0; --i) {
@@ -173,16 +173,16 @@ void array::ArrayDesc::init() {
   }
 }
 
-array::ArrayDesc::ArrayDesc(std::vector<int> shape, Dtype dtype)
+ArrayDesc::ArrayDesc(std::vector<int> shape, Dtype dtype)
     : shape(std::move(shape)), dtype(dtype) {
   init();
 }
 
-array::ArrayDesc::ArrayDesc(
+ArrayDesc::ArrayDesc(
     std::vector<int> shape,
     Dtype dtype,
     std::shared_ptr<Primitive> primitive,
-    std::vector<array> inputs)
+    small_vector<array> inputs)
     : shape(std::move(shape)),
       dtype(dtype),
       primitive(std::move(primitive)),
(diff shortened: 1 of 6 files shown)


awni:
@zcbenz this is interesting.. and I could see us moving in this direction. But I think we need some evidence that it is actually helpful before we can commit / start moving in this direction. 

I know it's a bit difficult to provide that evidence while also making the PRs in manageable, but I think in this case it might be worth expanding the scope of this PR. Maybe in some of the common ops where we construct arrays with primitives you can change that to use the smalle vector type (if it hasn't been updated already) and we can measure the benefit?

zcbenz:
This PR makes `array` store `inputs` with `small_vector`, so for ops that support move semantics (https://github.com/ml-explore/mlx/pull/895), you can observe that the extra array copy and heap allocation from `std::vector` in calls like `array(shape, dtype, {astype(std::move(a))})` are eliminated.

The `ops.cpp` does not use `std::vector` explicitly to construct inputs, and thanks to it they are all implicitly using `small_vector` in this PR, and they should have 1 less copy for each array passed as inputs, which was introduced by `initializer_list` (I haven't tested this though).

`eval` and `compile` should have no change of performance, as the inputs are converted to `std::vector` eventually in this PR, but we should test performance is at least not getting worse.

awni:
I made a gist here for a benchmark of just the graph construction for a Mistral 7B model. On an M2 Ultra this is the time I get:

```
Time 1516.782 (tps)
Time 0.659 (ms)
```

I will test it with this PR to see if it makes a difference.

awni:
I ran that same benchmark and it appears to be almost 3x slower 🤔 ..  do you want to look into that @zcbenz?

```
Time 585.678 (tps)
Time 1.707 (ms)
```

zcbenz:
Thanks for the benchmark code, I guess being 3x slower should make it easier to debug...

zcbenz:
So using `std::array` as stack container is a bad idea, it makes sure the default constructors of remaining elements are called on construction, which slows down things a lot.

zcbenz:
I'm going to be offline for a few days, closing this for now.

PR #923, opened 2024-03-28
Fix qmm_t for unaligned cases

angeloskath:
Typo in the check in the kernel. Only showed for specific sizes which made it elude our tests.


angeloskath:
Should address the issues at https://github.com/ml-explore/mlx-examples/pull/628#issue-2211237317 for the quantized gates.

awni (approved):
🚀

Issue #910, opened 2024-03-26
Better Exceptions in case of invalid operations on `mlx.core.array`


amirhossein-razlighi:
Regarding a conversation on [this PR](https://github.com/ml-explore/mlx/pull/886#issue-2204299594), there is a requirement for better exception (and error messages) in case of invalid operations on mlx array. For example:
```
import mlx.core as mx

mx.array([1, 2]) > [1,2] # ValueError: Invalid type  list received in array initialization.

# same thing for '<', '<=', '>=', '%', '*', etc.
```

while we are not actually initialising anything here! So, the error message should somehow represent that this operation is not valid. I think this will help for better debugging and more meaningful error messages.
Any suggestions?

amirhossein-razlighi:
@awni 
Any suggestions on what we throw in these cases? Maybe something like:
```
Value Error: You should convert operands to mlx.core.array before comparing to (or operating on) an array.
```

awni:
Maybe something like:

```
Value Error: Cannot <op_name> on an mlx.core.array and <type>.
```

I don't think we should suggest to cast to array because it's not essential.

PR #926, opened 2024-03-29
Better exceptions in case of invalid operations on `mlx.core.array` (#910)

amirhossein-razlighi:
## Proposed changes

This closes issue #910. In case of invalid operations on an `mlx.core.array`, we throw exceptions that are more meaningful. See the examples below:
```
import mlx.core as mx

a = mx.array([1, 2.0, 3])
str_ = "hello"
tpl_ = (1, 2.1, 3)
lst_ = [1, 2.5, 3.25]

a * str_  # ValueError: Cannot perform multiplication on an mlx.core.array and str
a /= tpl_ # ValueError: Cannot perform inplace division on an mlx.core.array and tuple
a // lst_ # ValueError: Cannot perform floor division on an mlx.core.array and list
# and same for other operations like <+ / - / ** / *= / += / -= / etc.>
```

## Checklist


amirhossein-razlighi:
Please note that I made this branch from (this PR)[https://github.com/ml-explore/mlx/pull/920#issue-2212567287] because the function `is_comparable_with_array` was needed here.
Please let me know if there is something needed to change and any suggestions you have.

awni:
I merged #920, you can rebase on main to clean this PR up.

amirhossein-razlighi:
> I merged #920, you can rebase on main to clean this PR up.

Done ✅

amirhossein-razlighi:
@awni 
Can you please review this PR whenever it's possible for you?
THX

awni (commented):
Could you change the name `is_comparable_with_array` to `is_convertible_to_array`? 

O/w I think this looks good.. it doesn't seem to add noticeable overhead.

amirhossein-razlighi:
> Could you change the name `is_comparable_with_array` to `is_convertible_to_array`?

Sure, I can do that. But I think it's not a good idea, since if you look into the logic of this function in `utils.h`, it is not actually checking if the arg is __convirtable__ to an array, but it's checking if it's __comparable__ to an array. (for example, for a python list, it returns `False`. But if it was checking for convertibility, it should've returned `True`)

awni:
Ok fair enough.. comparable is also not the best fit since it's more like `is_operable_with_array` but it's ok for now.

amirhossein-razlighi:
> Ok fair enough.. comparable is also not the best fit since it's more like `is_operable_with_array` but it's ok for now.

Sure, great.

awni (approved):
Thanks for adding this!!

Issue #908, opened 2024-03-26
Indexing an `mlx.array` with list


amirhossein-razlighi:
If we index an `mlx.core.array` type with a list, an exception occurs. While common convention in numpy or PyTorch is to  return the given elements correctly. E.g:
```
import mlx.core as mx
import numpy as np
import torch

a = mlx.array([1,2,3])
a[[0,1]] # ValueError: Cannot index mlx array using the given type.

b  = np.array([1,2,3])
b[[0,1]] # array([1, 2])

c = torch.tensor([1,2,3])
c[[0,1]] # tensor([1, 2])
```

If it's a valid issue and we want to follow these convention in mlx, please let me know. I'll be happy to work on it!

amirhossein-razlighi:
Sure, great. I will start working on that

PR #932, opened 2024-03-30
Support for list indexing (#908)

amirhossein-razlighi:
## Proposed changes

This closes issue #908. Indexing using lists is now supported for `mlx.core.array` type. Examples are as follows:
```
a = mx.array([1, 2, 3, 4, 5])
a[[ [0,2,4] ]] # array([1, 3, 5], dtype=int32)

a = mx.arange(10).reshape(5, 2)
a[[ [0,2,4] ]] # array([[0, 1],
                       #           [4, 5],
                       #           [8, 9]], dtype=int32)


 a[[ [0,1], [1] ]] # array([1, 3], dtype=int32)

```

## Checklist

### DIFF
--- a/python/src/indexing.cpp
+++ b/python/src/indexing.cpp
@@ -466,7 +466,9 @@ array mlx_get_item_list(array src, const nb::list& entries) {
     throw std::invalid_argument(msg.str());
   }
 
-  std::vector<array> gathered;
+  std::vector<array> gather_indices_vec;
+  std::vector<int> axes_vec;
+
   // Go through indices and gather the arrays
   int axis = 0;
   for (auto& idx : indices) {
@@ -486,23 +488,20 @@ array mlx_get_item_list(array src, const nb::list& entries) {
           gather_indices.begin(),
           {static_cast<int>(gather_indices.size())},
           uint32);
-      std::vector<int> slice_sizes = src.shape();
-      std::fill(slice_sizes.begin(), slice_sizes.end(), 1);
-      gathered.push_back(gather(src, arr, {axis}, slice_sizes));
+      gather_indices_vec.push_back(arr);
+      axes_vec.push_back(axis);
       axis++;
     } else {
-      gathered.push_back(mlx_get_item_int(src, nb::cast<nb::int_>(idx)));
+      gather_indices_vec.push_back(
+          mlx_get_item_int(src, nb::cast<nb::int_>(idx)));
     }
   }
-  printf("gathered size: %d\n", gathered.size());
-  // print each gathered array
-  for (int i = 0; i < gathered.size(); i++) {
-    printf("gathered[%d]:\n", i);
-    std::ostringstream os;
-    os << gathered[i];
-    printf("%s\n", os.str().c_str());
+  if (!gather_indices_vec.empty()) {
+    std::vector<int> slice_sizes(src.ndim(), 1);
+    std::fill(slice_sizes.begin(), slice_sizes.end(), 1);
+    src = gather(src, gather_indices_vec, axes_vec, slice_sizes);
+    src = squeeze(src);
   }
-  src = stack(gathered, 0);
   return src;
 }
 
--- a/python/src/indexing.cpp
+++ b/python/src/indexing.cpp
@@ -191,79 +191,64 @@ array mlx_gather_nd(
 
 auto mlx_expand_ellipsis(
     const std::vector<int>& shape,
-    std::variant<
-        std::reference_wrapper<const nb::tuple>,
-        std::reference_wrapper<const nb::list>> entries_variant) {
+    const nb::tuple& entries) {
   std::vector<nb::object> indices;
 
-  auto process_entries = [&](const auto& entries) {
-    // Go over all entries and note the position of ellipsis
-    int non_none_indices_before = 0;
-    int non_none_indices_after = 0;
-    std::vector<nb::object> r_indices;
-    int i = 0;
-    bool has_ellipsis = false;
+  // Go over all entries and note the position of ellipsis
+  int non_none_indices_before = 0;
+  int non_none_indices_after = 0;
+  std::vector<nb::object> r_indices;
+  int i = 0;
+  bool has_ellipsis = false;
 
-    // Start from dimension 0 till we hit an ellipsis
-    for (; i < entries.size(); i++) {
-      auto idx = entries[i];
-      if (!is_valid_index_type(idx)) {
-        throw std::invalid_argument(
-            "Cannot index mlx array using the given type yet");
-      }
-      if (!nb::ellipsis().is(idx)) {
-        indices.push_back(idx);
-        non_none_indices_before += !idx.is_none();
-      } else {
-        has_ellipsis = true;
-        break;
-      }
+  // Start from dimension 0 till we hit an ellipsis
+  for (; i < entries.size(); i++) {
+    auto idx = entries[i];
+    if (!is_valid_index_type(idx)) {
+      throw std::invalid_argument(
+          "Cannot index mlx array using the given type yet");
     }
-
-    // If we do hit an ellipsis, collect indices from the back
-    for (int j = entries.size() - 1; j > i; j--) {
-      auto idx = entries[j];
-      if (!is_valid_index_type(idx)) {
-        throw std::invalid_argument(
-            "Cannot index mlx array using the given type yet");
-      }
-      if (nb::ellipsis().is(idx)) {
-        throw std::invalid_argument(
-            "An index can only have a single ellipsis (...)");
-      }
-      r_indices.push_back(idx);
-      non_none_indices_after += !idx.is_none();
+    if (!nb::ellipsis().is(idx)) {
+      indices.push_back(idx);
+      non_none_indices_before += !idx.is_none();
+    } else {
+      has_ellipsis = true;
+      break;
     }
-
-    // Count up the number of non none indices
-    int non_none_indices = non_none_indices_before + non_none_indices_after;
-
-    // Expand ellipsis
-    if (has_ellipsis) {
-      for (int axis = non_none_indices_before;
-           axis < shape.size() - non_none_indices_after;
-           axis++) {
-        indices.push_back(nb::slice(0, shape[axis], 1));
-        non_none_indices++;
-      }
-    }
-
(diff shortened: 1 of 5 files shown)


amirhossein-razlighi:
Right now, list indexing does not support ellipsis, newaxis and slices. It support list of lists and integers. Since the additions are a lot, I wanted you to review these changes before adding more support.
(Also, if the changes are acceptable by you, I. will edit related documentations correspondingly)

angeloskath:
@amirhossein-razlighi I think this should be much simpler than that. Basically we should just make all lists into arrays and index as usual.

The simplest implementation would be to add a function that gets the python tuple of indices and returns a vector of indices with lists cast to mlx arrays. This can be called before `expand_ellipsis` for both get and set item.

Keep in mind that indexing many dimensions at once is always through a tuple of indices and not a list of indices ie `x[tuple([0, 1, 2])]` is very different than `x[[0, 1, 2]]`. In the first case we are indexing 3 dimensions while in the latter we are indexing the 0-th dimension.

In python pseudocode the implementation should be like the following

```python
def __getitem__(self, indices):
    if isinstance(indices, tuple):
        indices = tuple(mx.array(x) if isinstance(x, list) else x for x in indices)
    return self.old_getitem(indices)
```

amirhossein-razlighi:
> @amirhossein-razlighi I think this should be much simpler than that. Basically we should just make all lists into arrays and index as usual.
> 
> The simplest implementation would be to add a function that gets the python tuple of indices and returns a vector of indices with lists cast to mlx arrays. This can be called before `expand_ellipsis` for both get and set item.
> 
> Keep in mind that indexing many dimensions at once is always through a tuple of indices and not a list of indices ie `x[tuple([0, 1, 2])]` is very different than `x[[0, 1, 2]]`. In the first case we are indexing 3 dimensions while in the latter we are indexing the 0-th dimension.
> 
> In python pseudocode the implementation should be like the following
> 
> ```python
> def __getitem__(self, indices):
>     if isinstance(indices, tuple):
>         indices = tuple(mx.array(x) if isinstance(x, list) else x for x in indices)
>     return self.old_getitem(indices)
> ```

Oh, so simply converting the list that is being indexed to `mlx.core.array` is enough? i.e:
```
a = mx.array([[1,2], [3,4]])
a[ mx.array([0,1]) ] # Is it the same behaviour as a [ [0,1] ]?
```

awni:
>Oh, so simply converting the list that is being indexed to mlx.core.array is enough? 

Yes 💯 that's the right way to do it.

amirhossein-razlighi:
> > Oh, so simply converting the list that is being indexed to mlx.core.array is enough?
> 
> Yes 💯 that's the right way to do it.

I see your point. But there's something that is worth noting. Suppose this example:
```
import mlx.core as mx

a =  mx.array([[1,2], [3,4]])

# using mlx.array to index
a[ mx.array([[0,1], 0]) ] # This results in:
array([[[1, 2],
        [3, 4]],
       [[1, 2],
        [1, 2]]], dtype=int32)

# using current list indexing
a[ [[0,1], 0] ] # This results in:
array([1, 3], dtype=int32)
```

or even doing something like this:
```
a[ mx.array([[0,1], [0]]) ]
```
results in an exception `ValueError: Initialization encountered non-uniform length.`. Although with current implementation, these indexing result in the correct array and no exceptions (equivalent to PyTorch results).
So, maybe implicitly converting to an `mlx.core.arra` does not fulfil all the requirements needed. However, the current implementation supports all of these cases.

angeloskath:
@amirhossein-razlighi this points to a bug in converting lists to arrays. `mx.array([[0, 1], [0]])` is not a meaningful array as it doesn't have a meaningful shape.

Regarding indexing with a list, it should exactly be casting the list to an array so `a[ [[0,1], 0] ]` is an invalid operation as the provided list cannot be cast to an array.

amirhossein-razlighi:
> @amirhossein-razlighi this points to a bug in converting lists to arrays. `mx.array([[0, 1], [0]])` is not a meaningful array as it doesn't have a meaningful shape.
> 
> Regarding indexing with a list, it should exactly be casting the list to an array so `a[ [[0,1], 0] ]` is an invalid operation as the provided list cannot be cast to an array.

I know it cannot be converted to an array because of shape mismatch, but isn't it a known convention for array indexing (ie. in numpy, PyTorch, tensorflow and etc.)? I mean in each of these frameworks, `tensor[ [[0,1], [0]] ]` means selecting 0th and 1st rows and 0th columns and returning the intersection. For example in this case:
```
a = torch.tensor([[ [1,2,3], [4,5,6], [7,8,9] ]])
```
Indexing `a[ [[0,1], [0]] ]` returns `tensor([1,4])` which is the expected behaviour among all of frameworks. So, maybe one also expects the same behaviour in mlx. right?

angeloskath:
I think it is a confusing choice from PyTorch's side. TF doesn't allow list indexing, Jax follows numpy, numpy does the sensible thing here (not always but I do believe this is the more intuitive choice in this scenario) which is to only allow tuples for multidimensional indexing.

Given the above I believe we should only allow tuples for multidimensional indexing and casting to array is the expected behavior. You can see that in numpy the error is actually the exact same as doing `np.array([[0, 1],0])`.

Finally, another weirdness from PyTorch, doing `a[[0, 1], [0, 0]]` is equal to doing `a[[[0, 1], [0, 0]]]` which is a very weird choice for anyone familiar with numpy indexing as these are clearly two very different things. The first is two indices one for the 1st and one for the 2nd dimension while the latter is one index for the 1st dimension of the array.

amirhossein-razlighi:
@angeloskath 
You are absolutely right! I also think it's better to go with NumPY's convention. I will change the code accordingly.

awni:
What's the latest with this PR @amirhossein-razlighi ? Are you still planning to work on it?

amirhossein-razlighi:
> What's the latest with this PR @amirhossein-razlighi ? Are you still planning to work on it?

Yes. will complete that this weekend. Sorry for the delay

python/src/indexing.cpp
```diff
@@ -435,6 +463,8 @@ array mlx_get_item(const array& src, const nb::object& obj) {
     return mlx_get_item_int(src, nb::cast<nb::int_>(obj));
   } else if (nb::isinstance<nb::tuple>(obj)) {
     return mlx_get_item_nd(src, nb::cast<nb::tuple>(obj));
+  } else if (nb::isinstance<nb::list>(obj)) {
```
angeloskath, line 466: That's great!

python/src/utils.h
```diff
@@ -90,6 +90,17 @@ inline array to_array(
   }
 }
 
+inline void throw_if_invalid_index(const array& src, int sub_, int axis) {
```
angeloskath, line 93: Not used anywhere.

angeloskath (changes requested):
I think this requires quite a bit more work. As I mentioned in a previous comment, the way to go about it is to normalize the tuple of indices in `mlx_get_item` and `mlx_set_item`.

The decision to be made is whether we would make another tuple or vector of `nb::object` or a vector of `std::variant`. After the `mlx_get/set_item` there should be no new logic added and no lists anywhere.

To be completely precise, the normalization should also happen to `mlx_compute_scatter_args` and maybe `mlx_slice_update` but in general the logic added should be just dealing with normalizing the passed indices to contain `ints, arrays, slices, ellipses and Nones`.

python/src/indexing.cpp
amirhossein-razlighi: So, you want me to change the code in a way that in `mlx_get_item`, we first convert the elements to be one of `ints, arrays, slices, ellipses and Nones` (and thus, lists are not an option from here. If we have a list, we covert it to `array`) and then, proceed with the logic provided in `mlx_get_item` as it is? How is it different from what we are currently doing? (Just checking if the object `is instance<nb::list>` and if so, we do the conversion (or normalisation) inside `mlx_get_item_list` function.
OR maybe I misunderstood your point

python/src/utils.h
amirhossein-razlighi: Pushed new code related to this one. Previously, if we have something like this:
```
a = mx.array([0,1,2,3])
a[[0, 10]]
```
It will not throw; but it obviously should (because the index is out of bounds for dimension 0).
I now added the required checks, using `throw_if_invalid_index` to throw appropriate exception

python/src/indexing.cpp
amirhossein-razlighi: I think you are referring to this pseudo-code:
```python
def __getitem__(self, indices):
    if isinstance(indices, tuple):
        indices = tuple(mx.array(x) if isinstance(x, list) else x for x in indices)
    return self.old_getitem(indices)
```

But I do not quite understand it. You mean, for list indexing, we put the lists inside a tuple and then do the indexing in such way? But why should we go in this way instead of full converting to an `mlx.array`?

angeloskath: I think there may be a misunderstanding here. Multidimensional indexing is only performed using tuples.

When calling `x[a, b, c, d]` this really calls `x.__getitem__((a, b, c, d))` . Notice the double parentheses. The index passed to `__getitem__` will be a tuple. Passing a list does **not** perform multidimensional indexing. It indexes the 0-th dimension only!

So in the code-snippet above. `if not isinstance(indices, tuple)` then we only have one index for the 0-th dimension.

To give you an idea of the type of solution that we are looking for, the following implements the above python snippet and is a correct implementation of indexing using lists. Ideally we want this but without having to cast back and forth from C++ types to python types and without code duplication so that it covers all the setters as well.

```cpp
      .def("__getitem__", [](const array& src, nb::object indices) {
            if (nb::isinstance<nb::list>(indices)) {
              indices = nb::cast(array_from_list(indices, {}));
            } else if (nb::isinstance<nb::tuple>(indices)) {
              std::vector<nb::handle> new_indices;
              for (auto arg : nb::cast<nb::tuple>(indices)) {
                if (nb::isinstance<nb::list>(arg)) {
                  new_indices.push_back(nb::cast(array_from_list(arg, {})));
                } else {
                  new_indices.push_back(arg);
                }
              }
              indices = nb::tuple(nb::cast(new_indices));
            }
            return mlx_get_item(src, indices);
          }, nb::arg().none())

```

awni:
@amirhossein-razlighi what's the status of this PR? Have comments been addressed and is it ready for another review?

PR #933, opened 2024-03-30
Donation bug

awni:
- Fix bug with donation in softmax, rmsnorm+vjp, layernorm+vjp
- Fix a couple other bugs in softmax + vjps

### DIFF
--- a/mlx/backend/metal/normalization.cpp
+++ b/mlx/backend/metal/normalization.cpp
@@ -26,15 +26,14 @@ void RMSNorm::eval_gpu(
       no_copy &= (s == 0 || s == x.shape().back());
     }
     if (no_copy) {
-      return x;
+      return false;
     } else {
-      array x_copy(x.shape(), x.dtype(), nullptr, {});
-      copy_gpu(x, x_copy, CopyType::General, s);
-      copies.push_back(x_copy);
-      return x_copy;
+      copies.push_back(array(x.shape(), x.dtype(), nullptr, {}));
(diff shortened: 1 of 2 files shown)


angeloskath (approved):
Very nice!

My understanding is that we would copy and then not donate because `copies.back()` and `x_copy` were copies of the same array. Wouldn't returning `copies.back()` have the same result? Either way, great fix.

awni:
> My understanding is that we would copy and then not donate because copies.back() and x_copy were copies of the same array. 

I think it's a little more subtle than that. Even in the no copy case we did not donate before (the use count became 2 after calling `check_input`).  I think C++ extends the lifetime of the `return x` since we hold a `const & ` reference to it. So the use count gets incremented and does not decrement until that reference is gone.

angeloskath:
Interesting. What are the chances the return type is inferred to be `array` instead of `array&` or `cost array&`?

awni:
Great point, this works and is nicer actually
```
  auto check_input = [&copies, &s](const array& x) -> const array& {
    bool no_copy = x.strides()[x.ndim() - 1] == 1;
    if (x.ndim() > 1) {
      auto s = x.strides()[x.ndim() - 2];
      no_copy &= (s == 0 || s == x.shape().back());
    }
    if (no_copy) {
      return x;
    } else {
      copies.push_back(array(x.shape(), x.dtype(), nullptr, {}));
      copy_gpu(x, copies.back(), CopyType::General, s);
      return copies.back();
    }
  };
```

PR #934, opened 2024-03-30
Fix cpu compile

awni:
- Fix file name length bug for CPU compile
- Add test for another bug causing a segfault when all inputs are scalars

Simplified the contiguity stuff for the CPU it was too complicated and I think buggy, it just uses the exact same logic as the GPU eval now. We may want to improve both the CPU and GPU contiguity conditions at some point, but I think it needs more care.

Transformer training runs now with CPU compile.

### DIFF
--- a/mlx/backend/common/compiled.cpp
+++ b/mlx/backend/common/compiled.cpp
@@ -169,11 +169,12 @@ void compiled_allocate_outputs(
     for (int i = 0; i < inputs.size() && o < outputs.size(); ++i) {
       auto& in = inputs[i];
       // Conditions for donation
-      // - Contiguous
-      // - Donatable
       // - Correct size
+      // - Not a scalar
+      // - Donatable
       // - Not a constant
-      if (in.flags().contiguous && !is_scalar(in) && in.is_donatable() &&
+      if (in.itemsize() == outputs[o].itemsize() && !is_scalar(in) &&
+          in.is_donatable() &&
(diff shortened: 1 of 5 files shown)


angeloskath (approved):
Looks great!

awni:
@angeloskath I changed this a little to keep the contiguity check more generous but refactor the logic to clean it up (it's shared between CPU and GPU now). I think it's right as it is...lmk if you want to look again.

mlx/backend/common/compiled.cpp
```diff
@@ -126,4 +126,101 @@ std::string build_lib_name(
...
+    array::Flags flags;
+    for (int i = 0; i < inputs.size() && o < outputs.size(); ++i) {
+      auto& in = inputs[i];
+      // Conditions for donation
+      // - Contiguous
+      // - Donatable
+      // - Correct size
+      // - Not a constant
+      if (in.flags().contiguous && !is_scalar(in) && in.is_donatable() &&
+          constant_ids_.find(inputs_[i].id()) == constant_ids_.end()) {
```
angeloskath, line 177: I think we still need to check `nbytes()` to make sure the types are the same or `itemsize()` to be more explicit. Also presumably the `in.flags().contiguous` check is not needed as it is covered by `is_scalar(in)`.

angeloskath (approved):
I left a comment for what I think is an edge-case. The refactoring was so needed, thanks!

PR #942, opened 2024-04-01
Fix array initialization from list

### LINKED ISSUE #939: [BUG] `mx.array([[0, 1], 0])` should throw an error
**Describe the bug**
Inhomogeneous lists should not be castable to arrays.

**To Reproduce**

```python
import mlx.core as mx

mx.array([[0, 1], 0])
```

angeloskath:
Fixes #939 by checking that when an item is a scalar we do not have dimensions remaining in the shape.


awni (approved):
🙏

PR #944, opened 2024-04-01
Properly handle negative axes in python vmap

### LINKED ISSUE #922: [BUG] mapping confusion when in_axes is a negative index  in `vmap`
**Describe the bug**
in_axes' mapping confusion when in_axes is a negative index  in `vmap`

**To Reproduce**

Include code snippet
```python
import mlx 
import mlx.nn as nn
input_size=5
hidden_size=12
x = mx.random.normal((2, 25, 12))
hidden_layer = nn.Sequential(nn.Linear(input_dims=hidden_size, output_dims=hidden_size, bias=True), nn.Tanh())
all_h = []
for idx in range(x.shape[-2]):
    h = hidden_layer(x[...,idx,:])
    all_h.append(h)
all_h = mx.stack(all_h, -2)

vmap_pos_h = mx.vmap(hidden_layer, in_axes=1, out_axes=1)(x)
mx.allclose(all_h, vmap_pos_h)                              # array(True, dtype=bool)
```
that is all right, 
while 
```python
vmap_neg_h = mx.vmap(hidden_layer, in_axes=-2, out_axes=-2)(x)    # [vmap] Axis -2 invalid for input with 3 dimensions.
mx.allclose(all_h, vmap_neg_h) 
```
but 
```python
vmap_neg_h = mx.vmap(hidden_layer, in_axes=-1, out_axes=-2)(x)    
mx.allclose(all_h, vmap_neg_h)                            # array(True, dtype=bool)
```

 - OS Version: MacOS 14.4
 - Version: mlx=0.8.1

angeloskath:
Fixes #922 by allowing negative input/output axes in python.

I also made a multi-tree tree_visit to avoid the multiple traversals of the arguments. There is some repetition but I wanted to avoid creating all those python objects just to discard them in the end. Let me know if you think it isn't worth it.

### DIFF
--- a/python/src/transforms.cpp
+++ b/python/src/transforms.cpp
@@ -269,9 +269,11 @@ auto py_vmap(
                                 const nb::object& axes,
                                 bool output_axes) {
       std::vector<int> flat_axes;
+      bool encountered_tuple = false;
       tree_visit(
           {tree, axes},
-          [&flat_axes, output_axes](const std::vector<nb::object>& inputs) {
+          [&flat_axes, &encountered_tuple, output_axes](
+              const std::vector<nb::object>& inputs) {
             if (nb::isinstance<array>(inputs[0])) {
               if (inputs[1].is_none()) {
                 flat_axes.push_back(-1);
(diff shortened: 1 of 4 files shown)


angeloskath:
The test error has to do with the following

```python
def f(a, b):
    return a + b

g = mx.vmap(f, in_axes=(0, 0), out_axes=(0,))
# The output is a single array but the output axes are define in a tuple.
```

My first thought is that a single element tuple should match a single array and vice versa so I should fix that. Do you think we should be strict enough to have that be a failure?

angeloskath:
So this changes the behavior in a few ways:

This is now valid
```python
def f(x):
    return x["a"].sum()
mx.vmap(f, in_axes={"a": 1})({"a": mx.ones((2, 2))})
```

This is also valid
```python
def f(x):
    return x.sum()
mx.vmap(f, in_axes=(1,), out_axes=(0,))(mx.ones((2, 2)))
```

python/src/transforms.cpp
```diff
@@ -265,26 +266,70 @@ auto py_vmap(
...
+                  msg << "[vmap] Invalid" << (output_axes ? " output " : " ")
+                      << "vectorization axis " << axis
+                      << " for array with shape " << x.shape();
+                  throw std::invalid_argument(msg.str());
+                }
+                flat_axes.push_back(axis);
+              } else if (nb::isinstance<nb::tuple>(inputs[1])) {
+                encountered_tuple = true;
+                auto l = nb::cast<nb::tuple>(inputs[1]);
+                if (l.size() == 1 && nb::isinstance<nb::int_>(l[0])) {
```
awni, line 297: Is `None` allowed in this case?

awni (approved):
Looks great, thanks for fixing this!!

PR #949, opened 2024-04-01
Slightly dynamic ops per buffer

awni:
Speeds up generation w/out losing out much on other tasks. Not the most elegant but take a look and lmk if you think it's worth it. Will share some bench below.

### DIFF
--- a/mlx/backend/metal/device.cpp
+++ b/mlx/backend/metal/device.cpp
@@ -154,14 +154,16 @@ void Device::new_queue(int index) {
   queue_map_.insert({index, q});
 }
 
-int Device::get_command_buffer_ops(int index) {
+std::pair<int, int> Device::get_command_buffer_ops(int index) {
   auto bit = buffer_map_.find(index);
   return bit->second.first;
 }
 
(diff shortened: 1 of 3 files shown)


awni:
Bench | Old | New
----- | --- | ----
4-bit Mistral, 512 | 89.146 (tps) |92.247 (tps)
MNIST 2 | 0.626 (s) | Time 0.582 (s)
MNIST 10 | 1.050 (s) | Time 1.040 (s)
MNIST 100 | Time 5.982 (s) |  Time 4.678 (s)
Transformer 2, 12 | It/sec 6.025 | It/sec 6.069
Transformer 2, 24 | It/sec 3.116 | It/sec 3.120
LeNet |2.799 (s) | Time 2.809 (s)
ResNet20 CIFAR | 6434.22 samples/second | 6006.08 samples/second
ResNet44 CIFAR | 2854.87 samples/second | 2656.62 samples/second

awni:
I'm not giving up on a dynamic approach for this.. but I think it needs a cleaner solution.

Issue #946, opened 2024-04-01
[BUG] Segfault in compile with state


awni:
Following segfaults.. it should probably not:

```python
from functools import partial
import mlx.core as mx
            
w = mx.ones((10, 16))
b = mx.zeros((10,))                                                            
        
params = [w, b]
state = [w, b]  
                
x = mx.random.normal(shape=(1, 16))
            
def train_step(params, x):                                                     
    w, b = params
    state[0] = w
    state[1] = b
    return (x @ w.T + b).sum() #mx.addmm(b, x, w.T).sum()                      
    
@partial(mx.compile, inputs=state, outputs=state)                              
def step(inp):
    train_step_fn = mx.value_and_grad(train_step)                              
    loss, grads = train_step_fn(params, inp)                                   
    return loss
                                                                               
# Warmup                                                                       
loss = step(x)
mx.eval(loss)
```

PR #950, opened 2024-04-02
Fix compile fusion for multi-output edge cases

angeloskath:
Fixes 2 bugs in compile:
- When a Compiled primitive has many outputs the merge call to swap them in the graph was incorrectly using the same output again and again
- It was possible to fuse through an output that had different shape than the output of the primitive.

### DIFF
--- a/mlx/compile.cpp
+++ b/mlx/compile.cpp
@@ -165,25 +165,32 @@ CompileMode& compile_mode() {
 using ParentsMap =
     std::unordered_map<std::uintptr_t, std::vector<std::pair<array, int>>>;
 
+// Helper like below but only merges the two provided arrays. If the src has
+// siblings then these won't be merged to the dst.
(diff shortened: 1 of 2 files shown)


angeloskath:
Fixes #946 .

mlx/compile.cpp
```diff
@@ -676,7 +694,7 @@ void compile_fuse(
     // - Update outputs parents to point to compiled outputs
     // - Update any overall graph outputs to be compiled outputs
     for (int o = 0; o < old_outputs.size(); ++o) {
-      merge(compiled_outputs[o], old_outputs[o], parents_map);
+      merge_one(compiled_outputs[o], old_outputs[o], parents_map);
```
awni, line 697: That's a subtle one, good catch!

awni (approved):
Awesome!!

PR #953, opened 2024-04-02
Option for precise softmax

awni:
- Option to allow softmax to accumulate in float for half-type inputs

## Generation

Command `python -m mlx_lm.generate --model mlx-community/NeuralBeagle14-7B-4bit-mlx --prompt "Write a story about Einstein" --temp 0.0 --max-tokens 512`

|| Pre | Post
----- | ---- | ----
Prompt | 211.785  | 213.709 
Generation | 90.733 | 92.392
Iter 70: Train loss 1.117, Learning Rate 1.000e-05, It/sec 1.375, Tokens/sec 591.958, Trained Tokens 27487, Peak mem 16.213 GB

## LoRA

Slight memory saving with LoRA, no real diff in speed (as expected).

Command `python -m mlx_lm.lora --model mistralai/Mistral-7B-Instruct-v0.1 --data ../lora/data --train --iters 100 --steps-per-report 10`

| | Pre | Post
| ---- | --- | ---
Tokens/sec | 591.013 | 591.958,
Peak mem (GB) | 16.460 | 16.213

### DIFF
--- a/mlx/primitives.cpp
+++ b/mlx/primitives.cpp
@@ -3005,6 +3005,11 @@ std::vector<array> Softmax::jvp(
       multiply(s, sum(sv, std::vector<int>{-1}, true, stream()), stream()))};
 }
 
+bool Softmax::is_equivalent(const Primitive& other) const {
+  const Softmax& s_other = static_cast<const Softmax&>(other);
+  return precise_ == s_other.precise_;
+}
+
 std::pair<std::vector<array>, std::vector<int>> Sort::vmap(
     const std::vector<array>& inputs,
     const std::vector<int>& axes) {
--- a/mlx/primitives.cpp
+++ b/mlx/primitives.cpp
@@ -2975,7 +2975,7 @@ std::pair<std::vector<array>, std::vector<int>> Softmax::vmap(
   } else {
     softmax_axes.push_back(-2);
   }
-  return {{softmax(inputs[0], softmax_axes, stream())}, axes};
+  return {{softmax(inputs[0], softmax_axes, precise_, stream())}, axes};
 }
 
 std::vector<array> Softmax::vjp(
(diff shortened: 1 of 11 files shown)


angeloskath (approved):
Looks great!

I am not sure why we don't always perform the accumulation in float. It is exactly the same speed and would remove a couple of kernels. Given that we don't have the precise option implemented in the CPU side then we can simply remove it from the primitive altogether and just implement the logic in the op.

awni:
Yea I thought about making it the default / fixed behavior.

I'm not sure. The reason I didn't do it is two fold:
1. It feels somewhat inconsistent with our other accumulations (mean, std, ...) 
2. On the CPU side I think it will in fact be slower to always accumulate in high precision since we either have to make a temporary workspace (to avoid the second exp) or do two exps. If you have a clean solution for the CPU then I would be more inclined to make it the default since I agree it probably makes no diff on the GPU.

mlx/primitives.cpp
```diff
@@ -3005,6 +3005,11 @@ std::vector<array> Softmax::jvp(
       multiply(s, sum(sv, std::vector<int>{-1}, true, stream()), stream()))};
 }
 
+bool Softmax::is_equivalent(const Primitive& other) const {
+  const Softmax& s_other = static_cast<const Softmax&>(other);
+  return precise_ == s_other.precise_;
+}
```
awni, line 3011: Thanks.. I meant to update this and forgot..

angeloskath:
Hm but that 's not what I said. Quite the opposite actually :-) .

Just the GPU kernel always accumulates in float. The CPU "kernel" accumulates at the precision of the input. The op has an optional parameter `precise` which when true does exactly what you do now.

It is kind of a leaky abstraction as the op will have to check the device so the op needs to know the internals of the primitive but then it already does.

awni:
Ah I see, I didn't understand what you meant. Do you think it's ok that the CPU and GPU have different behavior when precise is `false`? It does remove a couple of kernels which is nice.

angeloskath:
Yeah I think it is totally fine. There is a reason to ask for precise, I can't imagine a reason to ask for imprecise.

awni:
Ok updated CPU backends to do precise as well.

Issue #935, opened 2024-03-30
[Feature] Implementation of `mx.meshgrid`


amirhossein-razlighi:
The equivalent of `torch.meshgrid` is needed in mlx for easier and more convenient way of creating 2D maps from 1D tensors. Please refer to [here](https://pytorch.org/docs/stable/generated/torch.meshgrid.html) for Pytorch's documentation on `torch.meshgrid`

amirhossein-razlighi:
Sure! I will start working on that

abeleinin:
@awni Feature merged. This issue can be closed.

PR #961, opened 2024-04-05
Add mx.meshgrid

abeleinin:
## Adds mx.meshgrid support

Implements `mx.meshgrid` (#935) support in C++ and Python.

## Checklist


mlx/ops.cpp
```diff
@@ -696,6 +696,54 @@ split(const array& a, int num_splits, StreamOrDevice s /* = {} */) {
   return split(a, num_splits, 0, to_stream(s));
 }
 
+std::vector<array> meshgrid(
+    std::vector<array>& arrays,
+    const bool copy /* = true */,
+    const bool sparse /* = false */,
+    const std::string& indexing /* = "xy" */,
```
angeloskath, line 703: A better choice would be `std::string_view indexing`.

mlx/ops.cpp
```diff
@@ -696,6 +696,54 @@ split(const array& a, int num_splits, StreamOrDevice s /* = {} */) {
...
+    StreamOrDevice s /* = {} */) {
+  std::vector<std::string> valid = {"xy", "ij"};
+  if (std::find(valid.begin(), valid.end(), indexing) == valid.end()) {
```
angeloskath, line 706: This should be `if (indexing != "xy" && indexing != "ij")`, no need for vector creation etc.

mlx/ops.cpp
```diff
@@ -696,6 +696,54 @@ split(const array& a, int num_splits, StreamOrDevice s /* = {} */) {
   return split(a, num_splits, 0, to_stream(s));
 }
 
+std::vector<array> meshgrid(
+    std::vector<array>& arrays,
+    const bool copy /* = true */,
```
angeloskath, line 701: This doesn't really make sense in MLX since operations are really always returning a new array. For instance you can't have a "view" of one array that is then updated and the view is updated as well (which is kind of the point for this argument).

mlx/ops.cpp
```diff
@@ -696,6 +696,54 @@ split(const array& a, int num_splits, StreamOrDevice s /* = {} */) {
...
+    output = arrays;
+  }
+
+  for (int i = 0; i < ndim; ++i) {
+    std::vector<int> shape(ndim, 1);
+    shape[i] = -1;
+    output[i] = reshape(output[i], shape, s);
+  }
+
+  if (indexing == "xy" and ndim > 1) {
```
angeloskath, line 728: The following can be simplified as

```cpp
// The shape vector from above that can be reused throughout.
// You really only need to create this once.
std::vector<int> shape(ndim, 1);

shape[1] = arrays[0].size();
outputs[0] = reshape(arrays[0], shape);
shape[1] = 1;
shape[0] = arrays[1].size();
outputs[1] = reshape(arrays[1], shape);
```

python/tests/test_ops.py
```diff
@@ -1737,7 +1787,7 @@ def test_linspace(self):
         expected = mx.array(np.linspace(0, 1))
         self.assertEqualArray(a, expected)
 
-        # Test int32 dtype
+        # Test int64 dtype
```
angeloskath, line 1790: Thanks!

python/src/ops.cpp
```diff
@@ -2515,6 +2515,40 @@ void init_ops(nb::module_& m) {
         Returns:
             array: The resulting stacked array.
       )pbdoc");
+  m.def(
+      "meshgrid",
+      [](nb::args arrays_,
+         bool copy,
```
angeloskath, line 2521: Should be removed, as mentioned above.

mlx/ops.cpp
```diff
@@ -696,6 +696,54 @@ split(const array& a, int num_splits, StreamOrDevice s /* = {} */) {
   return split(a, num_splits, 0, to_stream(s));
 }
 
+std::vector<array> meshgrid(
+    std::vector<array>& arrays,
```
angeloskath, line 700: `const std::vector<array>&`

angeloskath (changes requested):
Sorry for the slew of comments. This is a very nice PR, thanks for adding that!

The main thing is the removal of `copy` which should also simplify a lot of things in the op.

abeleinin:
Thanks for the comments. Let me know if anything else needs to be adjusted.

mlx/ops.cpp
```diff
@@ -696,6 +696,41 @@ split(const array& a, int num_splits, StreamOrDevice s /* = {} */) {
...
+    throw std::invalid_argument(
+        "[meshgrid] Invalid indexing value. Valid values are 'xy' and 'ij'.");
+  }
+
+  auto ndim = arrays.size();
+  std::vector<array> outputs;
+  for (int i = 0; i < ndim; ++i) {
+    std::vector<int> shape(ndim, 1);
+    shape[i] = -1;
+    outputs.push_back(reshape(arrays[i], shape, s));
```
awni, line 714:
```suggestion
    outputs.push_back(reshape(arrays[i], std::move(shape), s));
```

mlx/ops.cpp
```diff
@@ -696,6 +696,41 @@ split(const array& a, int num_splits, StreamOrDevice s /* = {} */) {
...
+  }
+
+  if (indexing == "xy" and ndim > 1) {
+    std::vector<int> shape(ndim, 1);
+
+    shape[1] = arrays[0].size();
+    outputs[0] = reshape(arrays[0], shape);
+    shape[1] = 1;
+    shape[0] = arrays[1].size();
+    outputs[1] = reshape(arrays[1], shape);
```
awni, line 724:
```suggestion
    outputs[1] = reshape(arrays[1], std::move(shape));
```

mlx/ops.cpp
```diff
@@ -696,6 +696,41 @@ split(const array& a, int num_splits, StreamOrDevice s /* = {} */) {
   return split(a, num_splits, 0, to_stream(s));
 }
 
+std::vector<array> meshgrid(
+    const std::vector<array>& arrays,
+    bool sparse /* = false */,
+    std::string_view indexing /* = "xy" */,
```
awni, line 702: Let's use a `std::string` here solely for the sake of MLX C (to save binding a new type).

awni: It might work with `string_view` (@andresy )?

mlx/ops.cpp
awni, line 724: Reshapes need to be done on the correct stream.

mlx/ops.cpp
awni: @abeleinin do you mind using `std::string` here?

awni: At @abeleinin we should use a `std::string` here, do you mind updating it?

awni (approved):
Thanks! LGTM! We can merge once @angeloskath approves and tests clear.

abeleinin:
Sorry for any delays. Glad to get this merged!

angeloskath (approved):

PR #964, opened 2024-04-06
Metal shaders for memory efficient self attention on large sequences

bpkeene:
Implements metal shaders for:

o = mx.fast.scaled_dot_product_attention(queries, keys, values, scale=scale, mask=None)

Supports fp16, fp32 dtypes; flexible hidden dimension, currently templated for 64 and 128.

Causal masking for prompt encoding not yet implemented, this shader is focused at present on full self-attention.

### Context
We're continuing the work started with https://github.com/ml-explore/mlx/pull/735 and following up with memory savings for the self-attention use case common in diffusion transformer workflows such as [Stable Diffusion 3](https://arxiv.org/pdf/2403.03206).  In particular, we observe memory savings for the 8-b use case of > 5GB for float32.  Current shaders are implemented using Steel-like GEMM primitives in MLX style, with potential for performance tuning to improve latency while retaining the memory savings.

### Supported
- Supports `mx.float16` and `mx.float32` dtypes
- Supports [head_dim](https://github.com/ml-explore/mlx-examples/blob/main/llms/mistral/mistral.py#L58C44-L58C57)=64,128 (covers most 7b+ LLMs)
- MHA supported (no MQA, no GQA)
- Unsupported use cases still go through MLX primitives under the hood.
- No backward pass implementation (inference-only kernel)

## Checklist

### DIFF
--- a/mlx/backend/metal/scaled_dot_product_attention.cpp
+++ b/mlx/backend/metal/scaled_dot_product_attention.cpp
@@ -29,9 +29,28 @@ void sdpa_full_self_attention_metal(
     array& out,
     std::vector<array>& temporaries) {
   std::ostringstream kname_self_attention;
-  kname_self_attention << "fast_inference_self_attention_";
+  kname_self_attention << "steel_gemm_attention_";
+
+  constexpr const int bm = 16;
+  constexpr const int bn = 16;
+  const int bk = q.shape(-1); // already forced to be 64 or 128
+
+  if (bk != 64 && bk != 128) {
+    throw std::runtime_error(
+        "[ScaledDotProductAttention::eval_gpu]: hidden dim: expected either 64, 128");
+  }
+
+  constexpr const int wm = 2;
+  constexpr const int wn = 2;
 
   std::string delimiter = "_";
+
+  kname_self_attention << "bm_" + std::to_string(bm) + delimiter;
+  kname_self_attention << "bn_" + std::to_string(bn) + delimiter;
+  kname_self_attention << "bk_" + std::to_string(bk) + delimiter;
+
+  /*   template [[host_name("steel_gemm_attention_bm_" #bm        \
+   "_bn_" #bn "_bk_" #bk "_itype_" #itype) */
 
   for (const auto& arr : {k, v, out}) {
     if (arr.dtype() != q.dtype()) {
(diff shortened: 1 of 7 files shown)


bpkeene:
Marking as draft, currently working through some numerical issues via separate workflow, and will add CPU side bindings + dispatch, test & docs - sharing a current status and will update this PR

bpkeene:
Working on benchmarks

mlx/backend/metal/scaled_dot_product_attention.cpp
```diff
@@ -19,6 +19,97 @@
...
+  const uint n_kv_heads = k.shape(1);
+
+  MLXScaledDotProductAttentionParams params{
+      query_sequence_length,
+      n_q_heads,
+      n_kv_heads,
+      1 /* self attention: unused */,
+      alpha};
+
+  set_array_buffer(compute_encoder, q, 0);
```
awni, line 97: These changed to `compute_encoder.set_input_array` (for inputs) and `compute_encoder.set_output_array` for outputs of the shader.

mlx/backend/metal/scaled_dot_product_attention.cpp
```diff
@@ -19,6 +19,97 @@
...
+  } else {
+    throw std::runtime_error(
+        "[ScaledDotProductAttention::eval_gpu]: NYI: sequence length > 16384");
+  }
+
+  constexpr const uint nsimd = 16;
+  constexpr const uint rows_per_tgroup = 16;
+  const int tgp_y_indices = ((int(qseq) - 1) / rows_per_tgroup) + 1;
+
+  auto compute_encoder = d.get_command_encoder(s.index);
```
awni, line 77: This changes to `auto& compute_encoder ...`

bpkeene:
Hi folks, 
Latency - Self Attention SDPA
Memory - Self Attention SDPA

Attaching some graphs for measured latency on M3 Max and some estimated memory savings per attention block (empirically observed at several data points, graph here obtained via formulas)

bpkeene:
Some room for improvement on larger sequences re: latency, with a divergence after ~2300 sequence length, though the memory savings exceeds 1GB ~2k, and is approaching 5GB at 4250 sequence length (SD3 8B use case).  

All measurements were with batch size 2, heads = 38, hidden dim = 64, and float32 on M3 Max / 48GB.

python/tests/test_fast_sdpa.py
```diff
@@ -32,9 +32,82 @@ def repeat(a):
...
+
+        Dk = 64
+
+        if self.is_apple_silicon:
+            dtypes.append(np.half)
+
+        for SEQUENCE_LENGTH in [63, 129, 400, 1178, 2202]:
+            for DTYPE in dtypes:
+                if DTYPE == np.float32 and SEQUENCE_LENGTH > 8192:
+                    continue
```
awni, line 73: There is no `8192` as far as I can tell

python/tests/test_fast_sdpa.py
```diff
@@ -32,9 +32,82 @@ def repeat(a):
...
+
+        dtypes = [np.float32]
+
+        Dk = 64
+
+        if self.is_apple_silicon:
+            dtypes.append(np.half)
+
+        for SEQUENCE_LENGTH in [63, 129, 400, 1178, 2202]:
+            for DTYPE in dtypes:
```
awni, line 71: We might want to make these test cases a bit smaller and then move a more comprehensive set to our longer running test suite (which is not public). Our CI test machines are not that big..

mlx/fast.cpp
```diff
@@ -559,15 +562,25 @@ array scaled_dot_product_attention(
...
+      !mask.has_value() && supported_head_dim_self_attn &&
+      n_q_heads == n_kv_heads && final_type != bfloat16 &&
       stream.device == Device::gpu;
-  // TODO, update routing conditions post further tuning
+
+  // fast decoding gpu shader
+  bool implementation_supports_use_case = batch_dim == 1 &&
+      query_sequence_length == 1 && !mask.has_value() && supported_head_dim &&
+      final_type != bfloat16 && stream.device == Device::gpu;
   implementation_supports_use_case &= false;
```
awni, line 581: Let's default to off until we improve perf regression over the fallback option.

mlx/fast.cpp
```diff
@@ -529,9 +529,12 @@ array scaled_dot_product_attention(
    * * CPU implementation
    * * batch size > 1
    * * query sequence length > 1
```
awni, line 531: No longer true right?

mlx/fast.cpp
```diff
@@ -529,9 +529,12 @@ array scaled_dot_product_attention(
...
+   * * query sequence length < 16 (e.g. for speculative decoding)
+   * * query sequence length > 16 && non-null mask (causal attention)
```
awni, line 533:
```suggestion
   * * query sequence length > 16 (causal attention)
```

`non-null mask` is mentioned below.

mlx/fast.cpp
awni, line 581: Could you change the condition names slightly. Something like:

- `supports_sdpa`
- `supports_full_self_attention`

Then `implementation_supports_use_case = supports_sdpa | supports_full_self_attention`.

I think it will be a bit more clear what's going on there.

awni:
@bpkeene left a few minor comments. Could you address? Once updated we can run the tests and get this merged.

bpkeene:
Updated with the requested changes, thank you for the prompt review!

awni (approved):
🚀 🚀 🚀

PR #967, opened 2024-04-07
[feat] upsample with bicubic interpolation

gathierry:
## Proposed changes

- Enable `mode="cubic"` for `nn.Upsample` with bicubic interpolation
- Refactor the function `upsample_linear()` a little to reuse part of the code
- Remove `indices = mx.clip(indices, 0, N - 1)` in `_scaled_indices ()` to align the result with PyTorch. Add another two lines in `_linear_indices()` to make sure the linear interpolation result won't change.

## Not covered in this PR

- The numerical results are aligned to PyTorch only when the `input_size * scale_factor` is an integer (tested in `test_upsample.py`. Otherwise the indices selections of the two frameworks are different. The same thing also happens to the existing `linear` and `nearest` interpolations.

## Checklist

### DIFF
--- a/python/mlx/nn/layers/upsample.py
+++ b/python/mlx/nn/layers/upsample.py
@@ -181,7 +181,7 @@ class Upsample(Module):
             Otherwise, the number of scale factors provided must match the
             number of spatial dimensions.
         mode (str, optional): The upsampling algorithm, either ``"nearest"``,
-            ``"linear"`` or ``cubic``. Default: ``"nearest"``.
+            ``"linear"`` or ``"cubic"``. Default: ``"nearest"``.
         align_corners (bool, optional): Changes the way the corners are treated
             during ``"linear"`` upsampling.  See the note above and the
             examples below for more details.  Default: ``False``.
--- a/python/mlx/nn/layers/upsample.py
+++ b/python/mlx/nn/layers/upsample.py
@@ -164,13 +164,14 @@ class Upsample(Module):
     For example, an audio signal would be 3D with 1 spatial dimension, an image
     4D with 2 and so on and so forth.
 
-    There are two upsampling algorithms implemented nearest neighbor upsampling
-    and linear interpolation. Both can be applied to any number of spatial
-    dimensions and the linear interpolation will be bilinear, trilinear etc
-    when applied to more than one spatial dimension.
+    There are three upsampling algorithms implemented nearest neighbor upsampling,
+    linear interpolation, and cubic interpolation. All can be applied to any number
+    of spatial dimensions. The linear interpolation will be bilinear, trilinear etc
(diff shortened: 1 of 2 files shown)


python/mlx/nn/layers/upsample.py
```diff
@@ -154,6 +218,18 @@ class Upsample(Module):
                [1.66667, 2, 2.33333, 2.66667],
                [2.33333, 2.66667, 3, 3.33333],
                [3, 3.33333, 3.66667, 4]], dtype=float32)
+        >>> b = nn.Upsample(scale_factor=2, mode='cubic')
```
angeloskath, line 221: I would remove the extra examples and add the new mode in the docs above.

angeloskath (commented):
This looks great! I left some comments. I also did a small benchmark and unfortunately doing 16 gathers has its toll. This is ~5x slower than PTs.

Given that it should most probably never be used in the hot path it is not too bad to merge it I believe.

python/mlx/nn/layers/upsample.py
```diff
@@ -35,12 +35,56 @@ def _linear_indices(N, scale, align_corners, dim, ndims):
     weight = indices - indices_l
     weight = mx.expand_dims(weight, -1)
 
+    # padding with border value
+    indices_l = mx.clip(indices_l, a_min=0, a_max=N - 1)
+    indices_r = mx.clip(indices_r, a_min=0, a_max=N - 1)
```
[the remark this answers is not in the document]
gathierry: Thanks for pointing out. For linear interpolation, we can definitely merge it into one line. But for bicubic, we have to do this clip after computing the `weight`. So I move it down for `bilinear` as well to keep the consistency. I'll move it to line 32 as one single line.

python/mlx/nn/layers/upsample.py
gathierry: nice catch. I forgot to update the docstring

python/mlx/nn/layers/upsample.py
```diff
@@ -121,8 +180,8 @@ class Upsample(Module):
             If a ``float`` is provided, it is the multiplier for all spatial dimensions.
             Otherwise, the number of scale factors provided must match the
             number of spatial dimensions.
-        mode (str, optional): The upsampling algorithm, either ``"nearest"`` or
-            ``"linear"``. Default: ``"nearest"``.
+        mode (str, optional): The upsampling algorithm, either ``"nearest"``,
```
angeloskath, line 183: Perfect. Would it be too much to ask you to change the paragraph above as well?

gathierry: Not at all, I should have updated that all at once. Sorry for overlooking the other part of the doc. I just have a new commit.

angeloskath (approved):
Looks great thanks!

PR #969, opened 2024-04-07
Improve profiling with gpu tracing

awni:
A few things:

- Python bindings for the metal tracing function so we can trace from Python
- Improve the docs a bit
- Change the behavior of putting one command per buffer in "MLX_METAL_DEBUG" mode because it messes with the performance characteristics of the program.
- Refactor `metal.h` into a front-end API and put the implementation stuff in `metal_impl.h`


jagrit06 (approved):
LGTM 🚀

PR #972, opened 2024-04-07
Async eval

awni:
Exposes an async evaluation for pipelining graph construction with compute.

Used like so in MLX LM:

```python
    y, prob = _step(y)

    while True:
        sync = mx.async_eval(y)
        next_out = _step(y)
        sync.wait()
        yield y.item(), prob
        y, prob = next_out
```

Pre: `95.905 tokens-per-sec`
Post: `99.682 tokens-per-sec`

### DIFF
--- a/mlx/transforms.cpp
+++ b/mlx/transforms.cpp
@@ -39,6 +39,12 @@ class Synchronizer : public Primitive {
 int detail::InTracing::tracing_counter{0};
 
 std::shared_future<void> async_eval(std::vector<array> outputs) {
+  static std::shared_future<void> global_synchronizer;
+  // Catch up with previous async eval if needed
+  if (global_synchronizer.valid()) {
+    global_synchronizer.wait();
+  }
+
   std::function<void(const array&)> recurse;
   std::queue<array> tape;
   std::unordered_set<std::uintptr_t> cache;
@@ -152,7 +158,8 @@ std::shared_future<void> async_eval(std::vector<array> outputs) {
       scheduler::enqueue(stream, std::move(task));
     }
   }
-  return deps[synchronizer.id()];
+  global_synchronizer = std::move(deps[synchronizer.id()]);
+  return global_synchronizer;
 }
 
 void eval(std::vector<array> outputs) {
--- a/mlx/transforms.cpp
+++ b/mlx/transforms.cpp
@@ -38,7 +38,7 @@ class Synchronizer : public Primitive {
 // are currently under a function transformation.
 int detail::InTracing::tracing_counter{0};
 
-void eval(std::vector<array> outputs) {
+std::shared_future<void> async_eval(std::vector<array> outputs) {
   std::function<void(const array&)> recurse;
   std::queue<array> tape;
   std::unordered_set<std::uintptr_t> cache;
@@ -152,8 +152,11 @@ void eval(std::vector<array> outputs) {
       scheduler::enqueue(stream, std::move(task));
     }
   }
+  return deps[synchronizer.id()];
+}
 
-  deps[synchronizer.id()].wait();
+void eval(std::vector<array> outputs) {
+  async_eval(std::move(outputs)).wait();
 }
 
 std::pair<std::vector<array>, std::vector<array>> vjp(
--- a/mlx/array.cpp
+++ b/mlx/array.cpp
@@ -93,7 +93,9 @@ void array::detach() {
 }
 
 void array::eval() {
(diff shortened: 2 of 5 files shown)


mustafaaljadery:
One line 45, after setting the unordered deps map, you should reverse size to save overhead: 
```cpp
std::unordered_map<std::uintptr_t, std::shared_future<void>> deps;
deps.reserve(outputs.size());
```

mustafaaljadery:
https://github.com/ml-explore/mlx/blob/a36a4db3b3f1ad4669ea698c0aad09a2b11e8d80/mlx/transforms.cpp#L272

Here on line 272, you could use emplace which is faster than insert as it doesn't copy to memory. 

```cpp
cotan_map.emplace(o.id(), astype(cotans[cotan_idx], o.dtype(), s));
```

In general this should be faster than most inserts.

mustafaaljadery:
https://github.com/ml-explore/mlx/blob/a36a4db3b3f1ad4669ea698c0aad09a2b11e8d80/mlx/transforms.cpp#L318

On line 318, I think you can reserve based on size of primals right after (before the loop). 

```cpp
vjps.reserve(primals_.size());
```

mustafaaljadery:
https://github.com/ml-explore/mlx/blob/a36a4db3b3f1ad4669ea698c0aad09a2b11e8d80/mlx/transforms.cpp#L322

move here might avoid unnecessary copying: 

```cpp
vjps.push_back(std::move(cotan_it->second));
```

awni:
@mustafaaljadery send a PR for some of these if you like? They aren't so related to this diff, so I won't include them but some of them could be good suggestions!

mustafaaljadery:
> @mustafaaljadery send a PR for some of these if you like? They aren't so related to this PR, so I won't include them but some of them could be good suggestions!

Ah that's my bad. Totally misunderstood the PR, I thought it was optimizations to be made to transforms.cpp so I was going line by line trying to see if there were memory allocation optimizations.

angeloskath (approved):
🔥🔥🔥

mlx/transforms.cpp
```diff
@@ -38,7 +38,13 @@ class Synchronizer : public Primitive {
 // are currently under a function transformation.
 int detail::InTracing::tracing_counter{0};
 
-void eval(std::vector<array> outputs) {
+std::shared_future<void> async_eval(std::vector<array> outputs) {
+  static std::shared_future<void> global_synchronizer;
+  // Catch up with previous async eval if needed
+  if (global_synchronizer.valid()) {
+    global_synchronizer.wait();
+  }
```
awni, line 46: @angeloskath do you prefer it with something like that? It's definitely safer this way.. maybe good to include it for now and revisit if it limits us.

awni:
100+ tps out of the box 🚀 

```
python -m mlx_lm.generate --model mlx-community/NeuralBeagle14-7B-4bit-mlx --prompt "Write a quick sort in python" --temp 0.0 --max-tokens 256
Fetching 6 files: 100%|██████████████████████████████████████████████████████████████████████████████████████████████████████| 6/6 [00:00<00:00, 68200.07it/s]
==========
Prompt: <|im_start|>user
Write a quick sort in python<|im_end|>
<|im_start|>assistant

Here's a quick implementation of the Quick Sort algorithm in Python:

```python
def partition(array, low, high):
    """
    This function takes an array, a low index and a high index, and returns the index at which the array is partitioned.
    """
    pivot = array[high]  # Choose the rightmost element as the pivot
    i = low - 1  # Index of the smaller element

    for j in range(low, high):
        if array[j] <= pivot:
            i += 1
            array[i], array[j] = array[j], array[i]  # Swap elements

    array[i + 1], array[high] = array[high], array[i + 1]  # Swap pivot with the element at i + 1
    return i + 1


def quick_sort(array, low, high):
    """
    This function takes an array, a low index and a high index, and recursively sorts the array using the Quick Sort algorithm.
    """
    if low < high:
        #
==========
Prompt: 246.650 tokens-per-sec
Generation: 101.639 tokens-per-sec
```

mlx/transforms.cpp
angeloskath: Yes! It looks great! The example I was using locally was

```python
x = mx.random.normal(shape=(100, 100, 100))
mx.eval(x)
for i in range(10):
    y = 2 * x
    h = mx.async_eval(y)
    z = 2 * y
    t = mx.async_eval(z)
    print(y)
```

it segfaults without the above global synchronizer.

PR #974, opened 2024-04-08
enable bfloat scan

awni:
As title.. works fine. Not sure why it was not instantiated.


angeloskath (approved):
Nice. Probably had to do with the older OS support that we had initially.

PR #977, opened 2024-04-10
Explicit barriers with concurrent dispatch

awni:
```
python -m mlx_lm.generate --model mlx-community/NeuralBeagle14-7B-4bit-mlx --prompt "Write a story about Einstein" --temp 0.0 --max-tokens 256
```

Pre: `101.414 tokens-per-sec`
Post: `103.578 tokens-per-sec`

This is complementary to MLX_MAX_OPS_PER_BUFFER. I get `106.619 tokens-per-sec` if I set it to 50 using the same command.

### DIFF
--- a/mlx/backend/accelerate/softmax.cpp
+++ b/mlx/backend/accelerate/softmax.cpp
@@ -355,6 +355,7 @@ void Softmax::eval_cpu(const std::vector<array>& inputs, array& out) {
       break;
     case float16:
       if (precise_) {
+        std::cout << "PRECISE SMAX? " << std::endl;
         softmax<
             float16_t,
             float,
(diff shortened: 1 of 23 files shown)


awni:
Made two new methods: `CommandEncoder::set_input_array` and `CommandEncoder::set_output_array` and use them.

angeloskath:
It looks great! I took a stab as well. Let me know. I did it with RAII but I am not sure I like it too much. Maybe just making a setter like `set_concurrent(bool)` would be a better option in which case the programmer needs to remember to unset the flag.

The concept however is

```cpp
auto compute_encode = ...

// Sync with any output

{
  auto concurrency_ctx = compute_encoder.start_concurrent();
  // Don't sync with outputs in this block only previous outputs
}

// Now sync with all outputs including these from the concurrent block
```

The names I put suck a bit so feel free to change them or of course revert the commit.

awni:
Very nice! Let's go with it! I'm good on this diff (can't approve it though).

angeloskath (approved):

PR #978, opened 2024-04-10
Masked mm

jagrit06:
## Proposed changes

Adds a tile level masked matrix multiplication primitive 

## Checklist


angeloskath:
This looks fantastic! Can't wait for the benchmarks! On a very crude benchmark on my laptop doing 1024x1024 matmuls with ~half the tiles is 25% faster which seems great!

I will go through it in detail later.

python/src/ops.cpp
```diff
@@ -3645,6 +3645,42 @@ void init_ops(nb::module_& m) {
         Returns:
             array: ``alpha * (a @ b)  + beta * c``
       )pbdoc");
+  m.def(
+      "tile_masked_mm",
```
awni, line 3649: Sorry if this is a nuisance.. but I think the word `block` is more familiar than `tile`. I might prefer the name `block_sparse_mm`  over `tile_masked_mm`. Wdyt?

mlx/ops.h
```diff
@@ -1185,6 +1185,38 @@ array addmm(
...
+/** Compute matrix product with tile-level masking */
+inline array tile_masked_mm(
+    array a,
+    array b,
+    int tile_size,
+    array mask_lhs,
+    array mask_rhs,
+    StreamOrDevice s = {}) {
+  return tile_masked_mm(a, b, tile_size, std::nullopt, mask_lhs, mask_rhs, s);
+}
```
awni, line 1218: I don't really think we need these overloads since they are unused and probably won't be used for a long time. Let's add them if/when they are useful.

python/src/ops.cpp
jagrit06: I could change the names to `block_masked_mm`

jagrit06:
Some benchmark numbers run on a M3 Max laptop (almost half of the output tiles are randomly masked, compared against same matmul)

```
  B,     M,     N,     K,   dtype,  t, block_size,  time, time_masked, diff% = 100. * (1.0 - time_mlx_masked / time_mlx)
 16,  1024,  1024,  1024, float32, nn,         64, 1.496,       0.901, +39.78%
  4,  1024,  1024,  4096, float32, nn,         64, 1.417,       0.945, +33.31%
  4,  1024,  4096,  1024, float32, nn,         64, 1.416,       0.817, +42.31%
  1,  4096,  4096,  4096, float32, nn,         64, 5.742,       3.769, +34.37%
  B,     M,     N,     K,   dtype,  t, block_size,  time, time_masked, diff% = 100. * (1.0 - time_mlx_masked / time_mlx)
 16,  1024,  1024,  1024, float32, nt,         64, 1.387,       0.869, +37.30%
  4,  1024,  1024,  4096, float32, nt,         64, 1.392,       0.941, +32.39%
  4,  1024,  4096,  1024, float32, nt,         64, 1.395,       0.798, +42.82%
  1,  4096,  4096,  4096, float32, nt,         64, 5.670,       3.824, +32.55%
  B,     M,     N,     K,   dtype,  t, block_size,  time, time_masked, diff% = 100. * (1.0 - time_mlx_masked / time_mlx)
 16,  1024,  1024,  1024, float16, nn,         64, 1.073,       0.550, +48.74%
  4,  1024,  1024,  4096, float16, nn,         64, 1.072,       0.585, +45.48%
  4,  1024,  4096,  1024, float16, nn,         64, 1.080,       0.589, +45.41%
  1,  4096,  4096,  4096, float16, nn,         64, 4.283,       2.460, +42.55%
  B,     M,     N,     K,   dtype,  t, block_size,  time, time_masked, diff% = 100. * (1.0 - time_mlx_masked / time_mlx)
 16,  1024,  1024,  1024, float16, nt,         64, 1.126,       0.624, +44.60%
  4,  1024,  1024,  4096, float16, nt,         64, 1.110,       0.623, +43.91%
  4,  1024,  4096,  1024, float16, nt,         64, 1.111,       0.612, +44.95%
  1,  4096,  4096,  4096, float16, nt,         64, 4.600,       2.520, +45.22%
```

mlx/backend/common/default_primitives.cpp
```diff
@@ -190,4 +224,140 @@ void AddMM::eval_cpu(const std::vector<array>& inputs, array& out) {
   return matmul_common_general(inputs[0], inputs[1], out, alpha_, beta_);
 }
 
+void BlockMaskedMM::eval_cpu(const std::vector<array>& inputs, array& out) {
```
awni, line 227: Maybe I'm missing something, but is this identical to the `eval_cpu` for the accelerate backend? If so I would refactor to just use this one. You can delete that on and move the implementation to a C++ file which is not conditionally included.

mlx/backend/common/default_primitives.cpp
```diff
@@ -190,4 +224,140 @@ void AddMM::eval_cpu(const std::vector<array>& inputs, array& out) {
...
+    throw std::runtime_error(
+        "[BlockMaskedMM::eval_cpu] Currently only supports float32.");
+  }
+  out.set_data(allocator::malloc_or_wait(out.nbytes()));
+
+  auto& a_pre = inputs[0];
+  auto& b_pre = inputs[1];
+  auto& out_mask = inputs[2];
+
+  auto check_transpose = [](const array& arr, bool do_copy) {
```
awni, line 238: The `do_copy` flag is unused unless I miss something. Otherwise this function is the same as the `check_transpose` in `matmul_common_general`.  Probably worth refactoring it to a standalone helper since it doesn't save that much as a closure.

python/src/ops.cpp
```diff
@@ -3645,6 +3645,42 @@ void init_ops(nb::module_& m) {
...
       )pbdoc");
+  m.def(
+      "block_masked_mm",
+      &block_masked_mm,
+      nb::arg(),
+      nb::arg(),
+      "tile_size"_a = 64,
+      "mask_out"_a = nb::none(),
+      "mask_lhs"_a = nb::none(),
+      "mask_rhs"_a = nb::none(),
```
awni, line 3656: I might suggest we make these keyword only. But I don't feel too strongly about it. Forces the calling code more readable though.

python/src/ops.cpp
```diff
@@ -3645,6 +3645,42 @@ void init_ops(nb::module_& m) {
         Returns:
             array: ``alpha * (a @ b)  + beta * c``
       )pbdoc");
+  m.def(
+      "block_masked_mm",
+      &block_masked_mm,
+      nb::arg(),
+      nb::arg(),
+      "tile_size"_a = 64,
```
awni, line 3653: Could you change this to `block_size` to be inline with `block` in the name?

mlx/ops.cpp
```diff
@@ -3572,6 +3572,160 @@ array addmm(
...
+  }
+
+  bool has_out_mask = !!mask_out;
+  bool has_operand_mask = !(!mask_out && !mask_rhs);
+
+  // Check valid tile sizes
+  // TODO: Add support for 16x16 tile
+  if (tile_size != 32 && tile_size != 64) {
+    std::ostringstream msg;
+    msg << "[tile_masked_mm] Only tile_sizes 32, 64 are supported."
```
awni, line 3596:
```suggestion
    msg << "[block_masked_mm] Only tile_sizes 32, 64 are supported. "
```

mlx/ops.cpp
```diff
@@ -3572,6 +3572,160 @@ array addmm(
...
+    throw std::invalid_argument(msg.str());
+  }
+
+  // Do shape checks for operands
+  int in_a_ndim = a.ndim();
+  int in_b_ndim = b.ndim();
+
+  if (a.ndim() == 0 || b.ndim() == 0) {
+    throw std::invalid_argument(
+        "[addmm] Got 0 dimension input. Inputs must "
```
awni, line 3607:
```suggestion
        "[block_masked_mm] Got 0 dimension input. Inputs must "
```

mlx/ops.cpp
```diff
@@ -3572,6 +3572,160 @@ array addmm(
...
+    a = reshape(a, {1, -1}, s);
+  }
+  if (b.ndim() == 1) {
+    // Insert a singleton dim at the end
+    b = reshape(b, {-1, 1}, s);
+  }
+
+  if (a.shape(-1) != b.shape(-2)) {
+    std::ostringstream msg;
+    msg << "[tile_masked_mm] Last dimension of first input with shape "
```
awni, line 3622:
```suggestion
    msg << "[block_masked_mm] Last dimension of first input with shape "
```

mlx/ops.cpp
```diff
@@ -3572,6 +3572,160 @@ array addmm(
...
+        << a.shape() << " must match second to last dimension of"
+        << " second input with shape " << b.shape() << ".";
+    throw std::invalid_argument(msg.str());
+  }
+
+  // Type promotion
+  auto out_type = result_type(a, b);
+  if (!issubdtype(out_type, floating)) {
+    std::ostringstream msg;
+    msg << "[tile_masked_mm] Only real floating point types are supported but "
```
awni, line 3632:
```suggestion
    msg << "[block_masked_mm] Only real floating point types are supported but "
```

mlx/ops.cpp
```diff
@@ -3572,6 +3572,160 @@ array addmm(
...
+    // RHS mask
+    array mask_rhs_p = mask_rhs.value_or(array({true}));
+    mask_rhs_p = broadcast_mask(mask_rhs_p, bsx_shape, tk, tn, s);
+
+    mask_list.push_back(mask_lhs_p);
+    mask_list.push_back(mask_rhs_p);
+  }
+
+  std::vector<array> inputs = {a, b};
+  inputs.insert(inputs.end(), mask_list.begin(), mask_list.end());
```
awni, line 3709: I would just make the `inputs` above and push the masks onto it directly. Saves making and copying an extra `vector<array>`. These kinds of little overheads can start to add up so worth avoiding when it doesn't add much complexity.

mlx/ops.cpp
```diff
@@ -3572,6 +3572,160 @@ array addmm(
...
+
+  // Caculate array
+  auto out = array(
+      out_shape,
+      out_type,
+      std::make_shared<BlockMaskedMM>(to_stream(s), tile_size),
+      inputs);
```
awni, line 3716:
```suggestion
      std::move(inputs));
```

mlx/ops.h
```diff
@@ -1185,6 +1185,16 @@ array addmm(
     const float& beta = 1.f,
     StreamOrDevice s = {});
 
+/** Compute matrix product with block masking */
+array block_masked_mm(
+    array a,
+    array b,
+    int tile_size,
```
awni, line 1192: Could you change that to `block_size` as well? (Here and in `ops.cpp`)

mlx/backend/common/default_primitives.cpp
jagrit06: Oh yes, I can do that for now, but later so that we can do `float16` and `bfloat16`, the same logic would be used for for the BNNS calls - so I ended up making copies

mlx/backend/common/default_primitives.cpp
jagrit06: So the `do_copy` flag plays a role when we have operand masks 
In which case a copy is always forced so that we can mask out the inputs inlace before sending them into the blas germ

awni: Ok.. but I don't see the call to `check_transpose` with `do_copy = true` anywhere. Is it missing or am I missing it?

awni: 👍  Sorry I missed that

mlx/backend/common/default_primitives.cpp
awni: Fair enough. Maybe we should have this be in a `BlockMaskedMM::eval` function to avoid duplicating all that code and dispatch to it for fp32?

mlx/backend/common/default_primitives.cpp
jagrit06: It's in there 
```python
  bool has_op_mask = inputs.size() > 3;
  auto [a_transposed, lda, a] = check_transpose(a_pre, has_op_mask);
  auto [b_transposed, ldb, b] = check_transpose(b_pre, has_op_mask);
```

python/src/ops.cpp
```diff
@@ -3645,6 +3645,44 @@ void init_ops(nb::module_& m) {
...
+
+        Note: Only ``block_size=64`` and ``block_size=32`` are currently supported
+
+        Args:
+            a (array): Input array or scalar.
+            b (array): Input array or scalar.
+            block_size (int): Size of blocks to be masked. Must be ``32`` or ``64`` (default: ``64``)
+            mask_out (array, optional): Boolean mask for output (default: ``None``)
+            mask_lhs (array, optional): Boolean mask for a (default: ``None``)
+            mask_rhs (array, optional): Boolean mask for b (default: ``None``)
```
awni, line 3683:
```suggestion
            mask_lhs (array, optional): Boolean mask for ``a`` (default: ``None``)
            mask_rhs (array, optional): Boolean mask for ``b`` (default: ``None``)
```

awni (approved):
🔥

Issue #811, opened 2024-03-08
[Feature] Support fft based convolution


adonath:
It would be nice to have FFT based convolution supported in mlx. FFT bases convolution shows much better performance for large images / arrays and kernels.  The FFT building blocks are already supported in mlx, so it is mostly a matter of combining them to a convolution operation.

awni:
One challenge here is that FFT is not yet supported on the GPU (in Metal). So you could use it but on the CPU it would almost certainly be much slower than our GPU convolution. 

Also I think FFT-based convolution is more of an implementation detail. If there are some sizes that are slow for you, please share any benchmarks. We can then figure out the best way to make them faster (which may or may not require an FFT-based convolution).

adonath:
Thanks @awni and @sebblanchet! I did a quick implementation of a FFT based convolution in MLX:

```python
def _centered(arr, newshape):
    newshape = mx.array(newshape)
    currshape = mx.array(arr.shape)

    startind = (currshape - newshape) // 2
    endind = startind + newshape
    myslice = [slice(startind[k].item(), endind[k].item()) for k in range(len(endind))]
    return arr[tuple(myslice)]


def convolve_fft(image, kernel, stream):
    """Convolve FFT for torch tensors"""
    image_2d, kernel_2d = image[0, 0], kernel[0, 0]

    shape = [image_2d.shape[i] + kernel_2d.shape[i] - 1 for i in range(image_2d.ndim)]

    image_ft = mx.fft.rfft2(image, s=shape, stream=stream)
    kernel_ft = mx.fft.rfft2(kernel, s=shape, stream=stream)
    result = mx.fft.irfft2(image_ft * kernel_ft, s=shape, stream=stream)
    return _centered(result, image.shape)
```

I also did a simple benchmark. It uses a random image of size 1024x1024 and varying kernel sizes. It compares `mx.conv2d` on the GPU and CPU respectively, the FFT based algorithm from above and for comparison Scipy's FFTI convolution implementation. The result is the following:

![mlx-conv-mini-benchmark](https://github.com/ml-explore/mlx/assets/3715928/09cf8c64-af1b-4d61-916e-1beab58076cb)

I think it follows exactly the expectation:
- gpu is faster than cpu for native convolution 
- native convolution is faster for small kernel sizes
- for large kernel sizes FFT clearly wins, because of the Nlog(N) scaling instead of N^2
- Scipy's FFT convolve is faster, than mine (not too surprising)

[...]

PR #981, opened 2024-04-11
Fast Metal FFT for all N

### LINKED ISSUE #399: [Feature] Add Metal support for FFT

barronalex:
## Proposed changes

A feature-complete Metal FFT that's faster than both CPU and PyTorch MPS in the majority of 1D cases.

Fully functional, but still needs some clean up.

Resolves #399.

## Supports
- All values of N (tested up to 2^20)
- Real and Inverse FFTs: `fft`, `ifft`, `rfft`, `irfft`
- ND FFTs: `fft2`, `ifft2`, `rfft2`, `irfft2`, `fftn`, `ifftn`, `rfftn`, `irfftn`

## Performance

For `N < 1024`, 1D FFTs on my M1 Max:
- Faster than PyTorch MPS for ~90% of FFT sizes
- ~1.5X higher average throughput than PyTorch MPS
- ~13X higher average throughput than CPU

All

We're only behind MPS on some multiples of 7 and all multiples of 11 and 13:
Radix 2-13
Our Bluestein's implementation is significantly more efficient for N < 1024:

Bluestein's

Note: For the sake of time, I ran at a slightly lower batch size than is required to max out the bandwidth for the powers of 2. I'll run a full one shortly, but in my experiments so far the relative speeds seem to hold.

## Implementation Details

For `N <= 2048` whose prime factors are all `<= 7`:
- Stockham's algorithm in threadgroup memory with a mixed-radix, out of place FFT
- Codelets for radix-2,3,4,5,7
- Kernels are specialized at runtime for each N with Metal function constants
- Threadgroup batching for small N to improve performance

For all other `N <= 1024`:
- A fused [Bluestein's algorithm](https://en.wikipedia.org/wiki/Chirp_Z-transform) implementation
- Bluestein twiddles are computed on CPU in float64 to maintain acceptable precision for the overall algorithm.

For `N > 1024`:
- The [four step FFT](https://en.wikipedia.org/wiki/Bailey%27s_FFT_algorithm) algorithm
- If N has prime factors `> 1024`, we use a manual version of Bluestein's implemented with MLX ops

RFFT:
- We implement a custom kernel for real FFTs that uses a trick to perform two at a time, doubling the bandwidth

### Codelet optimizations and additions
The radix codelets are extremely naive currently and could be replaced with hand-tuned or compiled ones that perform fewer than O(N^2) operations. We should also add radix11 and radix13 codelets to match MPS and VkFFT.
### Performance on ND and four step FFT cases
These have quite a few unnecessary copies currently. A fused implementation incorporating the transpose and twiddle factors would bring us closer to the max bandwidth.
### Accuracy
Accuracy is comparable to MPS' implementation but about an order of magnitude behind `pocketfft`. More careful twiddle factor computation inspired by `pocketfft` could help here. Precision also suffers a bit on very large N. Computing the twiddle factors in float64 as we do with Bluestein's would help.
### IRFFT
`irfft` on GPU currently only works for outputs of `rfft` (there are a couple exceptions in the tests to account for this).
### Convolution theorem
The fused Bluestein's implementation contains a convolution implemented with FFTs via the [convolution theorem](https://en.wikipedia.org/wiki/Convolution_theorem). For larger kernel sizes we might want to adapt this and add it to the main convolution implementation as suggested in #811.

## Checklist


awni:
Do you mind rebasing this @barronalex ?

PR #982, opened 2024-04-11
Add ArrayDesc destructor to avoid possible stack overflow

angeloskath:
The following script failed before this PR

```python
import gc

import mlx.core as mx

x = mx.ones(1)
for i in range(300_000):
    x = 2 * x
del x
gc.collect()
```

Weirdly, it failed with segmentation fault rather than stack overflow error but it may be an artifact of it being in the destructor. The stack trace was >260k levels deep so I think in general this is not too bad to do.

I measured no slowdown for Mistral inference or transformer training in comparison to main.

### DIFF
--- a/mlx/array.cpp
+++ b/mlx/array.cpp
@@ -203,7 +203,7 @@ array::ArrayDesc::~ArrayDesc() {
   std::vector<std::shared_ptr<ArrayDesc>> for_deletion;
 
   for (array& a : inputs) {
-    if (a.array_desc_ != nullptr && a.array_desc_.use_count() == 1) {
+    if (a.array_desc_.use_count() == 1) {
       for_deletion.push_back(std::move(a.array_desc_));
     }
   }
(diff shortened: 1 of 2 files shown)


mlx/array.cpp
```diff
@@ -192,6 +192,36 @@ array::ArrayDesc::ArrayDesc(
...
+  // that may also destory their corresponding descriptions and so on and so
+  // forth.
+  //
+  // This calls recursively the destructor and can result in stack overflow, we
+  // instead put them in a vector and destroy them one at a time resulting in a
+  // max stack depth of 2.
+  std::vector<std::shared_ptr<ArrayDesc>> for_deletion;
+
+  for (array& a : inputs) {
+    if (a.array_desc_ != nullptr && a.array_desc_.use_count() == 1) {
```
awni, line 206: If you want you can rely on the fact that `use_count == 0` if it's a `nullptr`, but maybe a little too slippery

```suggestion
    if (a.array_desc_.use_count() == 1) {
```

awni (approved):
Perfect! Thanks!

awni:
CC @davidkoski maybe you can try this diff as well to make sure stack issues are resolved on your end.

davidkoski:
> Weirdly, it failed with segmentation fault rather than stack overflow error but it may be an artifact of it being in the destructor. The stack trace was >260k levels deep so I think in general this is not too bad to do.

Segmentation fault is what I would expect on the C side -- it goes past the end of the stack pages and hits unavailable memory.

I confirm this fixes my use case, thank you!

mlx/array.cpp
angeloskath: Done. Very good point.

PR #986, opened 2024-04-11
No copy command encoder

awni:
Won't compile if you do:

```
auto compute_encoder = d.get_command_encoder(s.index);
```

which previously would silently work and cause nasty synchronization issues.

That will now give:
```
error: call to deleted constructor of 'CommandEncoder'
  auto compute_encoder = d.get_command_encoder(s.index);
```


awni:
CC @barronalex

angeloskath (approved):
👍

PR #989, opened 2024-04-12
gelu tanh approx

gathierry:
## Proposed changes

add tanh approximation for gelu function to match the implementation in torch and jax

## Checklist


awni:
This is nicely done @gathierry but as discussed I don't think we need yet another gelu approximation.

If you are able to show some numerical differences from your pretrained model usign the different gelus that would be very interesting. If it is in fact noticeably smaller for the tanh approximation perhaps we could replace the fast approximation with it.

awni (approved):
Thanks!!

PR #994, opened 2024-04-14
Quantize embedding

awni:
The main change is a `QuantizedEmbedding` layer but it also includes some changes to the quantize API and embedding API which make it much easier to use from e.g. MLX LM (corresponding PR https://github.com/ml-explore/mlx-examples/pull/680)

- Add a `call_as_linear` to the `Embedding` / `QuantizeEmbedding` layers.
- Move `quantize` to a top level function in `mlx.nn` (since it doesn't make sense a class method anymore).
- Add a `tree_map_with_path` and use it in the `quantize` call. This makes it much easier to selectively quantize modules (really simplifies model loading logic in MLX LM).

Will post some more detailed benchmarks on Gemma 2B. Seeing nice perf improvement so far. Should check and see if the dequantize is hurting us anywhere before landing this since models will by default have quantized embedding layers.

Pre: `102.395 tokens-per-sec`
Post: `125.979 tokens-per-sec`

I have it running at 160+ TPS locally with async + using the streaming detokenization  + more ops in the buffer. We should push for that after this lands.

### DIFF
--- a/python/mlx/nn/layers/quantized.py
+++ b/python/mlx/nn/layers/quantized.py
@@ -104,9 +104,9 @@ class QuantizedEmbedding(Module):
         )
         return out.reshape(*s, -1)
 
-    def call_as_linear(self, x):
+    def as_linear(self, x):
         """
-        Treat the quantized embedding layer as a quantized linear layer.
+        Call the quantized embedding layer as a quantized linear layer.
 
         Use this for example when input embedding and output projection
         weights are tied.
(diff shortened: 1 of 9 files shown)


python/mlx/nn/layers/quantized.py
```diff
@@ -1,11 +1,140 @@
...
+            self["weight"],
+            scales=self["scales"],
+            biases=self["biases"],
+            transpose=True,
+            group_size=self.group_size,
+            bits=self.bits,
+        )
+
+    def _extra_repr(self):
+        return f"{self.num_embeddings}, {self.dims}"
```
angeloskath, line 125:
```suggestion
        return  f"{self.num_embeddings}, {self.dims}, group_size={self.group_size}, bits={self.bits}"
```

angeloskath (approved):
Looks great!

PR #998, opened 2024-04-15
Shared events for synchronization + async eval

awni:
- Use Metal shared event for synchronization between the CPU and GPU
- Enable `async_eval` without returning a future
- Attach events to `array` in order to synchronize on possibly async evaluated arrays

For example:

```python
a = mx.array(1.0)
b = a + 1
mx.async_eval(b)
print(b) # No need to wait here.. handled automatically
```

Non-async benchmarks on M1 Max so far not much changed:

Bench | Pre | Post
----  | --- | ---
MNIST 2 CPU |  Time 0.345 (s) | Time 0.329 (s) |
MNIST 10 CPU | Time 0.628 (s)| Time 0.608 (s) |
MNIST 10 GPU | Time 1.278 (s) |  Time 1.276 (s) |
MNIST 100 GPU | Time 7.422 (s) | Time 7.448 (s) |
Transformer GPU | It/sec 2.670 | It/sec 2.671 |

Generating with Gemma on M1 Max:

```
MLX_MAX_OPS_PER_BUFFER=50 python -m mlx_lm.generate --model mlx_model --prompt "Write a story about Einstein" --temp 0.0 --max-tokens 256
```

Pre: `128.018 tokens-per-sec`
Post: `140.112 tokens-per-sec`

Pre + Max Ops 80: `143.687 tokens-per-sec`
Post + Max Ops 80: `164.307 tokens-per-sec`

### DIFF
--- a/cmake/metal.14.0.diff
+++ b/cmake/metal.14.0.diff
@@ -1,0 +1,36 @@
+diff -ur Metal/MTLEvent.hpp MetalNew/MTLEvent.hpp
+--- Metal/MTLEvent.hpp	2023-06-01 12:18:26
++++ MetalNew/MTLEvent.hpp	2024-04-15 07:36:59
+@@ -62,6 +62,7 @@
+ 
+     uint64_t                 signaledValue() const;
+     void                     setSignaledValue(uint64_t signaledValue);
++    bool                     waitUntilSignaledValue(uint64_t signaledValue, uint64_t timeoutMS);
+ };
+ 
+ class SharedEventHandle : public NS::SecureCoding<SharedEventHandle>
+@@ -138,6 +139,11 @@
+ _MTL_INLINE void MTL::SharedEvent::setSignaledValue(uint64_t signaledValue)
+ {
+     Object::sendMessage<void>(this, _MTL_PRIVATE_SEL(setSignaledValue_), signaledValue);
++}
++
++// method: waitUntilSignaledValue
++_MTL_INLINE bool MTL::SharedEvent::waitUntilSignaledValue(uint64_t signaledValue, uint64_t timeoutMS) {
++    return Object::sendMessage<bool>(this, _MTL_PRIVATE_SEL(waitUntilSignaledValue_timeoutMS_), signaledValue, timeoutMS);
+ }
+ 
+ // static method: alloc
+diff -ur Metal/MTLHeaderBridge.hpp MetalNew/MTLHeaderBridge.hpp
+--- Metal/MTLHeaderBridge.hpp	2023-06-01 12:18:26
++++ MetalNew/MTLHeaderBridge.hpp	2024-04-15 07:37:29
+@@ -1906,6 +1906,9 @@
+     "setShouldMaximizeConcurrentCompilation:");
+ _MTL_PRIVATE_DEF_SEL(setSignaledValue_,
+     "setSignaledValue:");
++_MTL_PRIVATE_DEF_SEL(
++    waitUntilSignaledValue_timeoutMS_,
++    "waitUntilSignaledValue:timeoutMS:");
+ _MTL_PRIVATE_DEF_SEL(setSize_,
+     "setSize:");
+ _MTL_PRIVATE_DEF_SEL(setSlice_,
--- a/mlx/backend/metal/metal.cpp
+++ b/mlx/backend/metal/metal.cpp
@@ -60,11 +60,12 @@ std::function<void()> make_task(array arr, bool signal) {
     auto pool = new_scoped_memory_pool();
     auto s = arr.primitive().stream();
     auto command_buffer = increment_command_buffer(s);
-    for (auto& i : arr.inputs()) {
-      if (i.event().valid() && i.event().stream() != arr.primitive().stream()) {
+    for (auto& input : arr.inputs()) {
+      if (input.event().valid() &&
+          input.event().stream() != arr.primitive().stream()) {
         // TODO, consider committing the buffer and encoding a wait in the new
         // buffer rather than on the task thread
-        i.event().wait();
+        input.event().wait();
       }
     }
 
@@ -92,8 +93,6 @@ std::function<void()> make_task(array arr, bool signal) {
     }
 
     if (signal) {
-      // Make sure this buffer had at least one command encoder on it.
-      auto id = arr.id();
       metal::device(s.device).end_encoding(s.index);
       command_buffer->encodeSignalEvent(
           static_cast<MTL::Event*>(arr.event().raw_event().get()),
@@ -107,9 +106,8 @@ std::function<void()> make_task(array arr, bool signal) {
           });
       metal::device(s.device).commit_command_buffer(s.index);
     } else {
-      auto id = arr.id();
       command_buffer->addCompletedHandler(
-          [s, buffers = std::move(buffers), id](MTL::CommandBuffer* cbuf) {
+          [s, buffers = std::move(buffers)](MTL::CommandBuffer* cbuf) {
             check_error(cbuf);
           });
     }
--- a/mlx/backend/metal/metal.cpp
+++ b/mlx/backend/metal/metal.cpp
@@ -55,17 +55,19 @@ inline void check_error(MTL::CommandBuffer* cbuf) {
   }
 }
 
-std::function<void()> make_task(
-    array& arr,
-    std::vector<std::shared_future<void>> deps,
-    std::shared_ptr<std::promise<void>> p) {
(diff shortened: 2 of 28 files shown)


mlx/backend/metal/metal.cpp
```diff
@@ -55,17 +55,19 @@ inline void check_error(MTL::CommandBuffer* cbuf) {
...
-  auto task = [arr, deps = std::move(deps), p = std::move(p)]() mutable {
+std::function<void()> make_task(array arr, bool signal) {
+  auto task = [arr = std::move(arr), signal]() mutable {
     auto pool = new_scoped_memory_pool();
-    for (auto& d : deps) {
-      d.wait();
-    }
     auto s = arr.primitive().stream();
     auto command_buffer = increment_command_buffer(s);
+    for (auto& i : arr.inputs()) {
```
angeloskath, line 63: Nitpick: `for (auto& arr_i : arr.inputs())` or `for (auto& input : arr.inputs())` etc

mlx/backend/metal/metal.cpp
```diff
@@ -55,17 +55,19 @@ inline void check_error(MTL::CommandBuffer* cbuf) {
...
+      if (i.event().valid() && i.event().stream() != arr.primitive().stream()) {
+        // TODO, consider committing the buffer and encoding a wait in the new
+        // buffer rather than on the task thread
+        i.event().wait();
```
angeloskath, line 67: This is quite interesting as it would allow us to continue encoding things in the GPU that will run immediately when the CPU signals the dependencies.

Otoh if the CPU takes too much time the GPU may run too far ahead taking up all the memory and so on. I think this is a good first choice but I see us implementing a compromise in the future. Something like the `malloc_or_wait`...

mlx/backend/metal/metal.cpp
```diff
@@ -88,20 +90,26 @@ std::function<void()> make_task(
     if (!arr.is_tracer()) {
       arr.detach();
     }
-    if (p) {
+
+    if (signal) {
+      // Make sure this buffer had at least one command encoder on it.
```
angeloskath, line 95: Is there something I am missing for this comment?

mlx/backend/metal/metal.cpp
```diff
@@ -88,20 +90,26 @@ std::function<void()> make_task(
...
+      auto id = arr.id();
```
angeloskath, line 96: Not used anywhere.

mlx/backend/metal/metal.cpp
```diff
@@ -88,20 +90,26 @@ std::function<void()> make_task(
...
-            p->set_value();
             scheduler::notify_task_completion(s);
             check_error(cbuf);
           });
       metal::device(s.device).commit_command_buffer(s.index);
     } else {
+      auto id = arr.id();
       command_buffer->addCompletedHandler(
-          [s, buffers = std::move(buffers)](MTL::CommandBuffer* cbuf) {
+          [s, buffers = std::move(buffers), id](MTL::CommandBuffer* cbuf) {
```
angeloskath, line 112: Same. The id is not used anywhere.

mlx/backend/metal/metal_impl.h
```diff
@@ -2,21 +2,17 @@
...
 void new_stream(Stream stream);
-std::shared_ptr<void> new_scoped_memory_pool();
 
-std::function<void()> make_task(
-    array& arr,
-    std::vector<std::shared_future<void>> deps,
-    std::shared_ptr<std::promise<void>> p);
+std::unique_ptr<void, std::function<void(void*)>> new_scoped_memory_pool();
+
+std::function<void()> make_task(array arr, bool flush);
```
angeloskath, line 16: `flush` vs `signal` I guess it started as `flush` and ended up as signal.

mlx/backend/no_metal/metal.cpp
```diff
@@ -12,14 +12,12 @@ bool is_available() {
...
+
+std::unique_ptr<void, std::function<void(void*)>> new_scoped_memory_pool() {
   return nullptr;
 }
 
-std::function<void()> make_task(
-    array& arr,
-    std::vector<std::shared_future<void>> deps,
-    std::shared_ptr<std::promise<void>> p) {
+std::function<void()> make_task(array arr, bool flush) {
```
angeloskath, line 20: Same story.

mlx/transforms.cpp
```diff
@@ -93,71 +105,82 @@ std::shared_future<void> async_eval(std::vector<array> outputs) {
...
-          stream, metal::make_task(arr, std::move(arr_deps), std::move(p)));
+      scheduler::enqueue(stream, metal::make_task(std::move(arr), signal));
     } else {
-      auto task = [arr,
-                   stream,
-                   deps = std::move(arr_deps),
-                   p = std::move(p)]() mutable {
-        for (auto& d : deps) {
-          d.wait();
+      bool flush = needs_signal.find(arr.id()) != needs_signal.end();
```
angeloskath, line 152: Unused.

mlx/transforms.cpp
```diff
@@ -93,71 +105,82 @@ std::shared_future<void> async_eval(std::vector<array> outputs) {
...
+      auto task = [arr = std::move(arr), stream, signal]() mutable {
+        for (auto& i : arr.inputs()) {
```
angeloskath, line 154: Same nitpick as in the metal task.

angeloskath (approved):
This is a great change! Opens up the possibility for much better interop between CPU/GPU and network produced arrays in the future.

Something that may be interesting to know is how does `encodeSignalEvent` work with concurrent dispatch command encoders. Could it be that the higher signal is actually emitted before the smaller one if we encode concurrently a slow running GPU command and a fast one?

mlx/backend/metal/metal.cpp
awni: No, sorry from a previous bug

awni:
Thanks for the comments.. sorry I left a lot of litter in there!

awni:
> Something that may be interesting to know is how does encodeSignalEvent work with concurrent dispatch command encoders. Could it be that the higher signal is actually emitted before the smaller one if we encode concurrently a slow running GPU command and a fast one?

Very good question. Short answer is I'm not certain there isn't the possibility for a race 😓 . 

On the GPU stream it should NOT be possible. `encodeSignalEvent` is on the command buffer not the command encoder.  So the possibility for race is between command buffers on the same stream. But for that we rely on hazard tracking (in particular there are no waits on events for tasks on the same stream). (FWIW even two concurrent command encoders in the same command buffer (which doesn't happen right now but maybe we want at some point) will be hazard tracked so there won't be a race between them.)

The other question is if it's possible to race between two streams (e.g. CPU and GPU) and that depends on the command queue semantics about which I am less certain. I know that command buffers in a queue are guaranteed to be dispatched in the same order and to end in the same order.  But I don't know if the events signaled at the end of the command buffer passes will respect the queue order. I tried to expose this in some tests (which I added here) but I can't get them to fail.

angeloskath:
I see. Then if my understanding is correct, if the events are signaled at the end of the command buffer the order doesn't matter since all previous buffers will have also finished and there can't be a race condition.

awni:
> Then if my understanding is correct, if the events are signaled at the end of the command buffer the order doesn't matter since all previous buffers will have also finished and there can't be a race condition.

Yes, and I confirmed this behavior to be sure, so I think we are good there.

awni:
Ok, moving this out of draft. @angeloskath would be good to double check the new status flag.

I'll share some benchmarks on Mistral 4-bit.

awni:
Command: `python -m mlx_lm.generate --model mlx-community/NeuralBeagle14-7B-4bit-mlx --prompt "Write a story about Einstein" --temp 0.0 --max-tokens 256`

Pre: 103.5
Post: 107.350

cmake/metal.14.0.diff
```diff
@@ -0,0 +1,36 @@
+diff -ur Metal/MTLEvent.hpp MetalNew/MTLEvent.hpp
+--- Metal/MTLEvent.hpp	2023-06-01 12:18:26
++++ MetalNew/MTLEvent.hpp	2024-04-15 07:36:59
```
awni, line 3: Just a comment on this patch for posterity. The `waitUntilSignaledValue` was retro-acitvely enabled recently (like a couple months ago) at least as far back as OS's that MLX supports. It has not percolated into metal-cpp yet. So I added the bindings here. However, we should remove these patches once it lands in metal-cpp.

angeloskath (approved):
Status looks good :-)

PR #1006, opened 2024-04-18
Add synchronize function

awni:
- Add `mx.synchronize(stream=None)` for synchronizing with a given stream.
- Updated `mx.metal.start_capture` function to throw on failure and print the error message rather than silently do nothing


angeloskath (approved):
That's pretty cool.

PR #1010, opened 2024-04-19
Fix buffer protocol buffer size designation

angeloskath:
As title and test show. The length for the buffer should be in bytes not elements.

Thanks to @nastya236 for finding the bug.


awni (approved):
Nice catch! Thanks for the fix @nastya236 @angeloskath

Issue #1008, opened 2024-04-18
[Feature Request] mlx.core.radians and mlx.core.degrees


Crear12:
Hi,
Will MLX support similar operations like np.radians and np.degrees? I'm trying to lazy-migrate my current code to MLX by ```import mlx.core as np```. In geo-information, I used a lot of np.radians and np.degrees, which haven't been implemented in MLX yet. 

I checked developer documentation but it's a little bit too complicated for me to contribute (I'm civil engineering student and I barely have any experience in C++). The python version should be something like this:
```
import mlx.core as mx

def radians(degrees_array: mx.array) -> mx.array:
    return degrees_array*(mx.pi/180)

def degrees(radians_array: mx.array) -> mx.array:
    return radians_array*(180.0 / mx.pi)
```

Following the documentation, C++ version should be like this (but pi needs to be specified or called somewhere):
```
array radians(
    const array& degrees, // Input array degrees
    StreamOrDevice s /* = {} */ // Stream on which to schedule the operation
) {
    return multiply(array(pi/180), degrees, s);
   
}
```
and
```
array degrees(
    const array& radians, // Input array radians
    StreamOrDevice s /* = {} */ // Stream on which to schedule the operation
) {
    return multiply(array(180/pi), radians, s);
   
}
```

Starting from Primitives, I'm totally lost...I'd appreciate it if someone can help integrate radians and degrees to MLX.

Thanks in advance!

awni:
You can easily patch MLX with whatever Python function you want:

```Python
import mlx.core as np

def degrees(radians_array: np.array) -> np.array:
    return radians_array*(180.0 / np.pi)

np.degrees = degrees

print(np.degrees(np.array(1.0)))
```

I'm not opposed to add this to MLX since NumPy supports it. You basically implemented it in your issue already.. just need the python bindings and it's done. Do you want to add it yourself?

Crear12:
> You can easily patch MLX with whatever Python function you want:
> 
> ```python
> import mlx.core as np
> 
> def degrees(radians_array: np.array) -> np.array:
>     return radians_array*(180.0 / np.pi)
> 
> np.degrees = degrees
> 
> print(np.degrees(np.array(1.0)))
> ```
> 
> I'm not opposed to add this to MLX since NumPy supports it. You basically implemented it in your issue already.. just need the python bindings and it's done. Do you want to add it yourself?

[...]

PR #1011, opened 2024-04-19
Adds mlx.core.radians and mlx.core.degrees and python bindings

aneeshk1412:
## Proposed changes

Adds `mlx.core.radians` and `mlx.core.degrees` and python bindings, which will complete issue (#1008).

## Checklist

### DIFF
--- a/python/tests/test_vmap.py
+++ b/python/tests/test_vmap.py
@@ -49,8 +49,10 @@ class TestVmap(mlx_tests.MLXTestCase):
             "sin",
             "sqrt",
             "square",
+            "degrees",
+            "radians",
         ]
-        ops = ["erfinv"]
+        ops += ["erfinv"]
         for opname in ops:
             with self.subTest(op=opname):
                 op = getattr(mx, opname)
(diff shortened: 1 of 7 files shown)


python/tests/test_vmap.py
```diff
@@ -49,8 +49,10 @@ def test_unary(self):
             "sin",
             "sqrt",
             "square",
+            "degrees",
+            "radians",
         ]
-        ops = ["erfinv"]
+        ops += ["erfinv"]
```
angeloskath, line 55: Damn thanks for catching that. You could also do it as follows though
```suggestion
                "erfinv"
        ]
```

angeloskath (commented):
This looks pretty straightforward and great to me, thanks!

Would you mind rebasing on main so that we can run the tests and then merge?

python/tests/test_vmap.py
aneeshk1412: Apparently `"erfinv"` is already there in the previous list, so I think we can just remove it from this line.

aneeshk1412:
Rebased to current main branch.

angeloskath (approved):

PR #1016, opened 2024-04-19
Make sure 0 is represented in the quantization

angeloskath:
Fixes our quantization such that 0 is always representable exactly. That in turn fixes https://github.com/ml-explore/mlx-examples/issues/692 .

### DIFF
--- a/mlx/ops.cpp
+++ b/mlx/ops.cpp
@@ -3268,6 +3268,9 @@ std::tuple<array, array, array> quantize(
   array scales = squeeze(delta, -1, s);
   array biases = squeeze(w_min, -1, s);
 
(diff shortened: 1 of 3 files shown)


awni (approved):
Amazing

mzbac:
You guys are legends. I can confirm that the fix works and the 4-bit can now generate correct output.

lin72h:
I'm so lucky to be late, everything works for me on the first try! Thank you team!

on my M3 Max 128GB

the q4 group 32 gives me:
==========
Prompt: 17.868 tokens-per-sec
Generation: 8.208 tokens-per-sec

the q4 group 64 gives me slightly better results:
==========
Prompt: 20.681 tokens-per-sec
Generation: 9.012 tokens-per-sec

PR #1019, opened 2024-04-20
Fixes teardown of program while GPU is running

angeloskath:
Guards the `Data` deleter and the `notify_task_completion` with `try ... catch` blocks because they are called from the command buffer completion handlers which may be called after the destruction of the main thread (now that we have `async_eval`).


angeloskath:
Closing because of fix in the synchronize PR.

PR #1028, opened 2024-04-23
Fix quantization of all 0s

angeloskath:
Set the minimum scale to 1e-7 so that the bias doesn't become nan if the scale is 0.


awni (approved):
Thanks!!

PR #1032, opened 2024-04-24
Expose function to clear memory cache

awni:
As title.


docs/src/python/ops.rst
```diff
@@ -92,6 +92,7 @@ Operations
    moveaxis
    multiply
    negative
+   not_equal
```
angeloskath, line 95: ❤️

angeloskath (approved):
Thanks!

PR #1037, opened 2024-04-26
Add bitwise ops

awni:
Add the numpy bitwise ops:

- bitwise_and, or, xor
- left_shift, right_shift

NB I did these all in a single `BitwiseBinary` primitive rather than a primitive for each. Uses a lot less code.. and also since the transform semantics are the same seemed ok. But it's a slight departure from our other binary ops so I'm open to changing it.

Also changed `^` to be the `bitwise_xor` rather than `power` which is I think more consistent with Python / the C++ meaning of the actual operator.


angeloskath (approved):
Looks great to me!

I think it's ok to have one Primitive for all these as the main thing that is facilitated by multiple primitives is the definition of the gradients and in this case there aren't any.

Issue #1031, opened 2024-04-24
[BUG] Using `make test` (e.g. ctest) causes segfaults


awni:
Build the C++ back-end and run the tests using ctest:

```
cmake .. && make -j
make test
```

Sample output:

```
 23/188 Test  #23: test reshape and transpose grads .............   Passed    0.05 sec
        Start  24: test copy grads
 24/188 Test  #24: test copy grads ..............................***Exception: SegFault  0.05 sec
        Start  25: test matmul vjp
 25/188 Test  #25: test matmul vjp ..............................   Passed    0.05 sec
```

PR #1039, opened 2024-04-26
Fix order device -> scheduler

awni:


awni:
This also closes #1031

PR #1043, opened 2024-04-27
feat: implement `clip_grad_norm`

### LINKED ISSUE #1040: [Feature] Add a `clip_grad_norm` to `mlx.optimizers`
It's relatively simple to implement, but maybe worth adding since it's also quite common. Here's a possible implementation:

```python
import mlx.core as mx

def clip_grad_norm(grads, max_norm):
    norm = mx.array(0.0)
    def accumulate(g):
        nonlocal norm
        norm += g.square().sum()

    tree_map(accumulate, grads)
    norm = mx.sqrt(norm)

    clip = lambda g: mx.where(norm < max_norm, g, g * max_norm / (norm + 1e-6))
    grads = tree_map(clip, grads)
    return grads, norm
```

We could also add a `tree_reduce` which would simplify it a bit.

NripeshN:
## Proposed changes

Closes #1040
Closes https://github.com/ml-explore/mlx/issues/1045

## Checklist

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -94,7 +94,7 @@ elseif (MLX_BUILD_METAL)
   FetchContent_Declare(
     metal_cpp
     URL ${METAL_CPP_URL}
-    PATCH_COMMAND patch -N -i ${METAL_CPP_PATCH} || true
+    PATCH_COMMAND /usr/bin/patch -N -i ${METAL_CPP_PATCH} || true
   )
 
   FetchContent_MakeAvailable(metal_cpp)
--- a/python/mlx/optimizers/optimizers.py
+++ b/python/mlx/optimizers/optimizers.py
@@ -739,25 +739,25 @@ class Adafactor(Optimizer):
 
 
 def clip_grad_norm(grads, max_norm):
-    """Clips the gradients by scaling them when their norm exceeds a maximum
-    threshold.
+    """Clips the global norm of the gradients.
 
-    This function ensures that the norm of the gradients does not exceed
-    `max_norm` to prevent issues like exploding gradients in neural networks.
-    It scales down the gradients proportionally if their norm is greater than
-    `max_norm`.
+    This function ensures that the global norm of the gradients does not exceed
+    ``max_norm``. It scales down the gradients proportionally if their norm is
+    greater than ``max_norm``.
 
     Example:
         >>> grads = {"w1": mx.array([2, 3]), "w2": mx.array([1])}
-        >>> max_norm = 2.0
-        >>> clipped_grads, total_norm = clip_grad_norm(grads, max_norm)
+        >>> clipped_grads, total_norm = clip_grad_norm(grads, max_norm=2.0)
         >>> print(clipped_grads)
         {"w1": mx.array([...]), "w2": mx.array([...])}
 
     Args:
-        grads (dict): A dictionary containing the gradient tensors for each
-            parameter of the model.
-        max_norm (float): The maximum norm value for the gradients.
+        grads (dict): A dictionary containing the gradient arrays.
+        max_norm (float): The maximum allowed global norm of the gradients.
+
+    Returns:
+        (dict, float): The possibly rescaled gradients and the original
+        gradient norm.
     """
     norm_squared = tree_reduce(lambda acc, g: acc + g.square().sum(), grads, 0.0)
     total_norm = mx.sqrt(norm_squared)
--- a/python/mlx/optimizers/optimizers.py
+++ b/python/mlx/optimizers/optimizers.py
@@ -742,7 +742,7 @@ def clip_grad_norm(grads, max_norm):
     """Clips the gradients by scaling them when their norm exceeds a maximum
     threshold.
 
-    This function ensures that the norm of the gradients does not exceed 
+    This function ensures that the norm of the gradients does not exceed
     `max_norm` to prevent issues like exploding gradients in neural networks.
     It scales down the gradients proportionally if their norm is greater than
     `max_norm`.
@@ -755,7 +755,7 @@ def clip_grad_norm(grads, max_norm):
         {"w1": mx.array([...]), "w2": mx.array([...])}
 
     Args:
-        grads (dict): A dictionary containing the gradient tensors for each 
+        grads (dict): A dictionary containing the gradient tensors for each
             parameter of the model.
         max_norm (float): The maximum norm value for the gradients.
     """
--- a/python/mlx/optimizers/optimizers.py
+++ b/python/mlx/optimizers/optimizers.py
@@ -739,10 +739,13 @@ class Adafactor(Optimizer):
 
 
 def clip_grad_norm(grads, max_norm):
-    """
-    Clips the gradients by scaling them when their norm exceeds a maximum threshold.
+    """Clips the gradients by scaling them when their norm exceeds a maximum
+    threshold.
 
-    This function ensures that the norm of the gradients does not exceed `max_norm` to prevent issues like exploding gradients in neural networks. It scales down the gradients proportionally if their norm is greater than `max_norm`.
+    This function ensures that the norm of the gradients does not exceed 
+    `max_norm` to prevent issues like exploding gradients in neural networks.
+    It scales down the gradients proportionally if their norm is greater than
+    `max_norm`.
 
     Example:
         >>> grads = {"w1": mx.array([2, 3]), "w2": mx.array([1])}
@@ -752,7 +755,8 @@ def clip_grad_norm(grads, max_norm):
         {"w1": mx.array([...]), "w2": mx.array([...])}
 
     Args:
-        grads (dict): A dictionary containing the gradient tensors for each parameter of the model.
+        grads (dict): A dictionary containing the gradient tensors for each 
+            parameter of the model.
         max_norm (float): The maximum norm value for the gradients.
     """
     norm_squared = tree_reduce(lambda acc, g: acc + g.square().sum(), grads, 0.0)
--- a/python/mlx/optimizers/optimizers.py
+++ b/python/mlx/optimizers/optimizers.py
@@ -739,9 +739,23 @@ class Adafactor(Optimizer):
 
 
 def clip_grad_norm(grads, max_norm):
-    norm_squared = tree_reduce(
-        lambda acc, g: acc + g.square().sum(), grads, mx.array(0.0)
-    )
+    """
+    Clips the gradients by scaling them when their norm exceeds a maximum threshold.
+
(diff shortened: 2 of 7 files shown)


NripeshN:
@awni 
Tests pass locally, ready to merge

python/mlx/utils.py
```diff
@@ -191,3 +191,46 @@ def tree_unflatten(tree):
         return l
     else:
         return {k: tree_unflatten(v) for k, v in children.items()}
+
+
+def tree_reduce(fn, tree, initializer=None, is_leaf=None):
```
angeloskath, line 196: I think it is best to simplify this function a bit. The following is exactly the same logic but without keeping an extra `first` flag which is implicitly kept by the accumulator being None or not.

```python
def tree_reduce(fn, tree, initializer=None, is_leaf=None):
    if is_leaf is not None and is_leaf(tree):
        return tree if initializer is None else fn(initializer, tree)

    accumulator = initializer

    if isinstance(tree, (list, tuple)):
        for item in tree:
            accumulator = tree_reduce(fn, item, accumulator, is_leaf)
    elif isinstance(tree, dict):
        for item in tree.values():
            accumulator = tree_reduce(fn, item, accumulator, is_leaf)
    else:
        return tree if accumulator is None else fn(accumulator, tree)

    return accumulator
```

As an aside, the following two-liner also does the same thing.

```python
def tree_reduce(fn, tree, initializer=None, is_leaf=None):
    values = (v for p, v in tree_flatten(tree, is_leaf=is_leaf))
    return functors.reduce(fn, values, initializer)
```

In terms of speed the two-liner is 10%-15% slower for a 128 layer transformer which I think is not too bad but the simplified version of reduce is also quite straightforward so let's keep that.

Finally, I would format the docstring as is in the other functions.

python/mlx/optimizers/optimizers.py
```diff
@@ -736,3 +736,16 @@ def apply_single(self, gradient: mx.array, parameter: mx.array, state: dict):
         if self.weight_decay != 0:
             parameter += parameter * (-self.weight_decay * learning_rate)
         return parameter - update
+
+
+def clip_grad_norm(grads, max_norm):
+    norm_squared = tree_reduce(
+        lambda acc, g: acc + g.square().sum(), grads, mx.array(0.0)
+    )
+    total_norm = mx.sqrt(norm_squared)
```
angeloskath, line 745: Slightly nitpicky but I would compute the normalizer here instead of in the clipper function, ie

```python
normalizer = max_norm / (total_norm + 1e-6)
```

CMakeLists.txt
```diff
@@ -94,7 +94,7 @@ elseif (MLX_BUILD_METAL)
   FetchContent_Declare(
     metal_cpp
     URL ${METAL_CPP_URL}
-    PATCH_COMMAND patch -N -i ${METAL_CPP_PATCH} || true
+    PATCH_COMMAND /usr/bin/patch -N -i ${METAL_CPP_PATCH} || true
```
angeloskath, line 97: Why this change? Was it not working without it?

angeloskath (changes requested):
Thanks this looks great! I left a few comments the main one being the simplification of `tree_reduce`.

CMakeLists.txt
NripeshN: > Why this change? Was it not working without it?

As discussed in this https://github.com/ml-explore/mlx/issues/1045 issue, it does not work without adding this.

awni: @NripeshN your default patch wasn't working. Could you do `patch -v` and tell me the output? I tried with both the macOS default patch and gpatch and they both work fine. So I'm wondering which version of patch you have that didn't work.

angeloskath (commented):
Looks great! I know it is a nitpick but do you mind text-wrapping the docstring to the same columns as the rest of the functions in the module?

After that I think this is a great addition especially the `tree_reduce(...)`!

python/mlx/optimizers/optimizers.py
```diff
@@ -736,3 +736,17 @@ def apply_single(self, gradient: mx.array, parameter: mx.array, state: dict):
         if self.weight_decay != 0:
             parameter += parameter * (-self.weight_decay * learning_rate)
         return parameter - update
+
+
+def clip_grad_norm(grads, max_norm):
```
awni, line 741: Could you add a docstring for this as well?

python/mlx/optimizers/optimizers.py
```diff
@@ -736,3 +736,17 @@ def apply_single(self, gradient: mx.array, parameter: mx.array, state: dict):
...
+    norm_squared = tree_reduce(
+        lambda acc, g: acc + g.square().sum(), grads, mx.array(0.0)
```
awni, line 743: I would use `0.0` here instead of `mx.array(0.0)` then it will be weakly typed so we don't need to worry about inadvertent up-casting.

python/mlx/utils.py
```diff
@@ -191,3 +191,45 @@ def tree_unflatten(tree):
...
+    """
+    Recursively applies a reducer function `fn` to the leaves of the Python tree `tree` and aggregates them into a single value.
```
awni, line 198: The word "recursive" here is slightly confusing. I would shorten this. There's enough detail on usage below.

```suggestion
    Applies a reduction to the leaves of a Python tree.
```

python/mlx/utils.py
```diff
@@ -191,3 +191,45 @@ def tree_unflatten(tree):
...
+
+    This function simplifies complex tree structures into a single accumulated result by applying a provided function `fn` from the leaves upward. It is similar to the built-in `reduce` function but works over nested, tree-like data structures.
```
awni, line 200:
```suggestion
    This function reduces Python trees into an accumulated result by applying the provided function `fn` to the leaves of the tree. It is similar to the built-in `reduce` function but works over nested, tree-like data structures.
```

python/mlx/utils.py
```diff
@@ -191,3 +191,45 @@ def tree_unflatten(tree):
...
+
+    Example:
+        >>> from mlx.utils import tree_reduce
+        >>> tree = {"a": [1, 2, 3], "b": [4, 5]}
+        >>> tree_reduce(lambda acc, x: acc + x, tree, 0)
+        15
+
+    Args:
+        fn (callable): The reducer function that takes two arguments (accumulator, current_value) and returns the updated accumulator.
```
awni, line 209:
```suggestion
        fn (callable): The reducer function that takes two arguments (accumulator, current value) and returns the updated accumulator.
```

python/mlx/utils.py
```diff
@@ -191,3 +191,45 @@ def tree_unflatten(tree):
...
+        tree (Any): The Python tree to reduce. It can be any nested combination of lists, tuples, or dictionaries.
+        initializer (Any, optional): The initial value to start the reduction. If not provided, the first leaf value is used.
+        is_leaf (callable, optional): A function to determine if an object is a leaf, returning True for leaf nodes and False otherwise.
```
awni, line 212:
```suggestion
        is_leaf (callable, optional): A function to determine if an object is a leaf, returning ``True`` for leaf nodes and ``False`` otherwise.
```

python/mlx/utils.py
```diff
@@ -191,3 +191,45 @@ def tree_unflatten(tree):
...
+
+    Returns:
+        Any: The aggregated value after reduction.
```
awni, line 215:
```suggestion
        Any: The accumulated value.
```

python/mlx/optimizers/optimizers.py
awni: I guess you could also leave it unspecified and use the first leaf?

python/mlx/utils.py
```diff
@@ -191,3 +191,45 @@ def tree_unflatten(tree):
...
+
+    .. note::
+        If `initializer` is not provided, the function will use the first leaf value as the initial accumulator. This may lead to unexpected results if the first leaf is not representative or compatible with the intended reduction.
```
awni, line 218:
```suggestion
        If ``initializer`` is not provided, the function will use the first leaf value as the initial accumulator. 
```

awni (changes requested):
Looks nice, thanks for the addition! I left a few comments mostly on docs.

CMakeLists.txt
NripeshN:
```(venv) nripeshniketan@Nripeshs-MBP ~/D/p/mlx (clip_grad_norm) [0|2]> patch -v
GNU patch 2.7.6
Copyright (C) 2003, 2009-2012 Free Software Foundation, Inc.
Copyright (C) 1988 Larry Wall

License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>.
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.

Written by Larry Wall and Paul Eggert```

File: python/mlx/optimizers/optimizers.py [hunk 12]
```diff
@@ -736,3 +736,31 @@ def apply_single(self, gradient: mx.array, parameter: mx.array, state: dict):
         if self.weight_decay != 0:
             parameter += parameter * (-self.weight_decay * learning_rate)
         return parameter - update
+
+
+def clip_grad_norm(grads, max_norm):
+    """
+    Clips the gradients by scaling them when their norm exceeds a maximum threshold.
+
+    This function ensures that the norm of the gradients does not exceed `max_norm` to prevent issues like exploding gradients in neural networks. It scales down the gradients proportionally if their norm is greater than `max_norm`.
```
angeloskath, line 745: I know it is a nitpick but since the rest of the docstrings are wrapped at something like 80 chars could you please wrap these docstrings as well?

Also the first sentence should start in the same line as `"""`.

python/mlx/utils.py
```diff
@@ -191,3 +191,40 @@ def tree_unflatten(tree):
         return l
     else:
         return {k: tree_unflatten(v) for k, v in children.items()}
+
+
+def tree_reduce(fn, tree, initializer=None, is_leaf=None):
+    """Applies a reduction to the leaves of a Python tree.
+
+    This function reduces Python trees into an accumulated result by applying the provided function `fn` to the leaves of the tree. It is similar to the built-in `reduce` function but works over nested, tree-like data structures.
```
angeloskath, line 199: Same as above. Could you please wrap the docstrings (including the args section).

angeloskath (changes requested):
I left a comment about text-wrapping the docstrings.

I know it feels like a lame comment when we have 4k 30'' monitors but since all other docstrings are wrapped could you textwrap these as well for consistency?

angeloskath (approved):
Looks good to me! Thanks!

awni (approved):
Thanks!!

PR #1058, opened 2024-04-30
Block sparse mm

jagrit06:
## Proposed changes

Adds operation and primitive to gather matrices before matmul on the fly 

## Checklist


python/src/ops.cpp
```diff
@@ -3716,6 +3716,39 @@ void init_ops(nb::module_& m) {
...
+      nb::arg(),
+      nb::arg(),
+      "lhs_indices"_a = nb::none(),
+      "rhs_indices"_a = nb::none(),
+      nb::kw_only(),
+      "stream"_a = nb::none(),
+      nb::sig(
+          "def block_sparse_mm(a: array, b: array, /, lhs_indices: array, rhs_indices: array, *, stream: Union[None, Stream, Device] = None) -> array"),
+      R"pbdoc(
+        Matrix multiplication with matrix-level gather
```
awni, line 3732:
```suggestion
        Matrix multiplication with matrix-level gather.
```

python/src/ops.cpp
```diff
@@ -3716,6 +3716,39 @@ void init_ops(nb::module_& m) {
...
+
+        Perform the (possibly batched) matrix multiplication of two arrays and with a gather
+        along the provided indices on the operands.
```
awni, line 3735:
```suggestion
        Perform the (possibly batched) matrix multiplication of two arrays with a gather
        along the provided indices on the operands.
```

python/src/ops.cpp
```diff
@@ -3716,6 +3716,39 @@ void init_ops(nb::module_& m) {
...
+
+        ``lhs_indices`` and ``rhs_indices`` contain flat indices along the batch dimension 
+        for ``a`` and ``b`` respectively
```
awni, line 3738:
```suggestion
        Performs a gather of the operands with the given indices followed by a (possibly batched) matrix multiplication of two arrays. This operation is more efficient than explicitly applying a :func:``take`` followed by a :func:``matmul``.
        
        The indices ``lhs_indices`` and ``rhs_indices`` contain flat indices along the batch dimensions (i.e. all but the last two dimensions) of ``a`` and ``b`` respectively.
```

python/src/ops.cpp
```diff
@@ -3716,6 +3716,39 @@ void init_ops(nb::module_& m) {
...
+
+        For ``a`` with shape ``(A1, A2, ..., AS, M, K)``, 
+        ``lhs_indices`` contains indices from the range ``[0, A1 * A2 * ... * AS)``
+
+        For ``b`` with shape ``(B1, B2, ..., BS, M, K)``, 
+        ``rhs_indices`` contains indices from the range ``[0, B1 * B2 * ... * BS)``
+
+        Args:
+            a (array): Input array or scalar.
+            b (array): Input array or scalar.
```
awni, line 3748:
```suggestion
            a (array): Input array which must be at least 2D.
            b (array): Input array which must be at least 2D. 
```

mlx/ops.cpp
```diff
@@ -3777,6 +3777,123 @@ array block_masked_mm(
...
+    n_batch_b *= i;
+
+  // Pull and broadcast indices
+  array lhs_indices =
+      lhs_indices_.value_or(reshape(arange(n_batch_a, uint32, s), bsx_a, s));
+  array rhs_indices =
+      rhs_indices_.value_or(reshape(arange(n_batch_b, uint32, s), bsx_b, s));
+
+  if (issubdtype(lhs_indices.dtype(), inexact) ||
+      lhs_indices.dtype() == bool_) {
```
awni, line 3851: Nit:

```suggestion
  if (!issubdtype(lhs_indices.dtype(), integer))) {
```

mlx/ops.cpp
```diff
@@ -3777,6 +3777,123 @@ array block_masked_mm(
...
+    throw std::invalid_argument(
+        "[block_sparse_mm] Got lhs_indices with invalid dtype. Indices must be integral.");
+  }
+
+  if (issubdtype(rhs_indices.dtype(), inexact) ||
+      rhs_indices.dtype() == bool_) {
```
awni, line 3857:
```suggestion
  if (!issubdtype(rhs_indices.dtype(), integer)) {
```

mlx/ops.cpp
```diff
@@ -3777,6 +3777,123 @@ array block_masked_mm(
...
+  for (auto& i : bsx_a)
+    n_batch_a *= i;
+  for (auto& i : bsx_b)
+    n_batch_b *= i;
+
+  // Pull and broadcast indices
+  array lhs_indices =
+      lhs_indices_.value_or(reshape(arange(n_batch_a, uint32, s), bsx_a, s));
+  array rhs_indices =
+      rhs_indices_.value_or(reshape(arange(n_batch_b, uint32, s), bsx_b, s));
```
awni, line 3848: I'm not a big fan of this style with optionals because I think it will always evaluate the operand which can be kind of inefficient to construct all those arrays and never use them if you specify `lhs_indices`.  Maybe cleaner to do it with a lambda like so:
```c++
auto indices_or_default = [](const std::optional<array>& indices, std::vector<int> bsx_shape) {
  if (indices.has_value()) {
    return indices.value();
  } else {
     int n_batch = 1;
     for (auto& i : bsx_shape)
       n_batch *= i;
    return reshape(arange(n_batch, uint32, s),  bsx, s);
  }
}
```

awni:
Very cool!! Can't wait to try this in an MOE!

awni:
Some MOE benchmarks:

### Generation
```
python -m mlx_lm.generate --model Qwen/Qwen1.5-MoE-A2.7B-Chat  --prompt "Write a story about Einstein" --max-tokens 256 --temp 0.0
```

### LoRA

```
python -m mlx_lm.lora --train --iters 50 --model Qwen/Qwen1.5-MoE-A2.7B-Chat --data ../lora/data
```

awni (approved):
🚀

PR #1059, opened 2024-04-30
Fix leak for multi-output primitives which are never detached

awni:
Closes: https://github.com/ml-explore/mlx-examples/issues/738

No issues destroying very deep graphs, the following still works:

```python
a = mx.array(1.0)
for _ in range(1000000):
    a = mx.abs(a)
```

### DIFF
--- a/mlx/array.cpp
+++ b/mlx/array.cpp
@@ -171,12 +171,16 @@ array::~array() {
     return;
   }
 
-  // Break circular reference for non-detached arrays with siblings
+  // Ignore arrays that will be detached
   if (status() != array::Status::unscheduled) {
     return;
   }
+  // Break circular reference for non-detached arrays with siblings
   if (auto n = siblings().size(); n > 0) {
     bool do_detach = true;
+    // If all siblings have siblings.size() references except
+    // the one we are currently destroying (which has siblings.size() + 1)
+    // then there are no more external references
     do_detach &= (array_desc_.use_count() == (n + 1));
     for (auto& s : siblings()) {
       do_detach &= (s.array_desc_.use_count() == n);
--- a/mlx/array.cpp
+++ b/mlx/array.cpp
@@ -1,5 +1,4 @@
 // Copyright © 2023-2024 Apple Inc.
-
 #include <functional>
 
 #include "mlx/array.h"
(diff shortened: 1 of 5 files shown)


mlx/array.cpp
```diff
@@ -167,6 +166,32 @@ void array::move_shared_buffer(array other) {
...
+      if (!do_detach) {
+        break;
+      }
+    }
+    if (do_detach) {
+      for (auto& s : siblings()) {
+        for (auto& ss : s.siblings()) {
+          ss.array_desc_ = nullptr;
+        }
+        s.array_desc_->siblings.clear();
```
angeloskath, line 189: I may be particularly slow today but I think the above code has a bug. The siblings of siblings will set the `array_desc_` to `nullptr` and then we 'll try to dereference that to clear the siblings.

```
Let our arrays be A1, A2, A3 all being siblings and the corresponding siblings vectors being S1, S2, S3.

S1 = [A2, A3]
S2 = [A1, A3]
S3 = [A1, A2]

Suppose A1 is the array being destroyed and A2, A3 have no external references the following lines are the two loops unrolled.

s = A2
ss = A1
ss.array_desc_ = nullptr;  // A1.array_desc_ = nullptr;
ss = A3;
ss.array_desc_ = nullptr;  // A3.array_desc_ = nullptr;
s.array_desc_->siblings.clear();  // Clear S2
s = A3;
// segfault when fetching ss cause s.array_desc is nullptr (assuming it didn't)
ss = A1;
ss.array_desc_ = nullptr;  // A1.array_desc_ = nullptr;
ss = A2;
ss.array_desc_ = nullptr;  // A2.array_desc_ = nullptr;
// segfault again when clearing siblings for s
```

awni: This works fine:
```
a, b, c = mx.split(mx.zeros(9), 3)
del a, b, c
```

Maybe I'm misunderstanding your comment, but I think it's ok because each array has it's own `array_desc_`?

awni: However, there is a bug elsewhere as the C++ tests aren't passing.. so let me track that down. Then I can look into simplifying this function so it's not so confusing.

awni:
@angeloskath I added some comments in the destructor. I think it's working now.  And I don't have any great ideas to simplify it. Lmk what you think.

angeloskath (approved):
Let's merge it. It still gives me a headache but I am not sure there is much to be done to simplify it.

PR #1064, opened 2024-05-02
Change initial memory limits and add memory size to device info

awni:
1. Add a `memory_size` field to the `device_info` dictionary
2. Change the allocator memory limits to be more sensible. Previously on devices with large machines the memory limit is set to well over the total physical memory which is not a very sensible default.

LLM generation / transformer training timings did not change on M2 Ultra.


angeloskath (approved):
👌

PR #1070, opened 2024-05-03
Fix conversion to array

### LINKED ISSUE #1066: [BUG] arithmetic operations with numpy arrays are not commutative
**Describe the bug**
mx array being first operand in arithmetic op is being converted complex numbers

**To Reproduce**

Include code snippet
```python
import numpy as np
import mlx.core as mx

mx.array(1) + np.array(2)
x = mx.array(1)
y = np.array(2)
x + y 
# >>> array(3+0j, dtype=complex64)
x * y
# >>> array(2+0j, dtype=complex64)
y + x
# >>> 3
type(y + x)
# >>> <class 'numpy.int64'>
type(x + y)
# >>> <class 'mlx.core.array'>
```

**Expected behavior**
Conversion to complex should not happen. Arithmetic ops should be commutative (ignoring type like ndarray or mx).

 - OS Version: [e.g. MacOS 14.1.2] MacOS 14.1.1
 - Version [e.g. 0.7.0] 0.11

**Additional context**

awni:
Closes #1066

This is kind of a nanobind issue but one that they have refused to fix in the past.

- Basically a `np.ndarray` is implicitly convertible to a `std::complex<float>` so our `to_array` was allowing that. 
- Rather than disallowing ops with `np.ndarrays` instead I added it as a member in the variant.
- That however has side effects because now `mx.array` gets implicitly converted to `nb::ndarray` because of the buffer protocol. 
- So to fix that I also had to enable `array` as a member of the variant.

### DIFF
--- a/python/src/CMakeLists.txt
+++ b/python/src/CMakeLists.txt
@@ -18,6 +18,7 @@ nanobind_add_module(
   ${CMAKE_CURRENT_SOURCE_DIR}/linalg.cpp
   ${CMAKE_CURRENT_SOURCE_DIR}/constants.cpp
   ${CMAKE_CURRENT_SOURCE_DIR}/trees.cpp
+  ${CMAKE_CURRENT_SOURCE_DIR}/utils.cpp
 )
 
 if (NOT MLX_PYTHON_BINDINGS_OUTPUT_DIRECTORY)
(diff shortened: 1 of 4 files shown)


angeloskath (approved):
That's quite nice! Implicit conversion from all frameworks apparently cause `nb::ndarray` works with numpy, PT etc.

PR #1074, opened 2024-05-03
Reset peak memory

awni:
Add a reset function to make memory profiling easier.


angeloskath (approved):
That may be the fastest feature request implementation cycle ever 😅

Issue #1065, opened 2024-05-02
[Feature] arctan2


yrahul3910:
I want to eventually contribute to [this Keras issue](https://github.com/keras-team/keras/issues/19571), and since this would be my first contribution, I want to start small and implement `arctan2`. If nobody is currently working on this, I would like to contribute this feature. Specifically, my implementation plan, based on looking at `arctan`, is as follows:

* Add an `ArcTan2` class to `mlx/primitives.h`, and its corresponding `vjp`, `jvp`, and `vmap` functions in `mlx/primitives.cpp`
* Add an `ArcTan2` struct to `mlx/backend/common/ops.h`, and overload the function call operator.
* Add an `ArcTan2` struct to `mlx/backend/metal/unary.h`, and add `instantitate_unary_float` calls in `mlx/backend/metal/unary.metal`. The necessary `metal::precise::atan2` function seems to already be implemented.
* Add tests, and update docs.

Please let me know if I can proceed.

awni:
Your description is right modulo the fact that it is a binary op not a unary op.

I would look at another binary op like `Add` as an example. Most of the implementation should be boiler plate copying of another binary primitive. The core change will be in the op struct which gets passed to the CPU / GPU binary op kernel.

awni:
And yes please add this, that would be great!

yrahul3910:
I think I did it :) When running `make`, I get warnings about the use of `binary(...)` (but it successfully builds and tests pass). This is because the `binary` function also tests `bool`, so `std::atan2` warns about precision loss. I'm not entirely sure what to do there--maybe for future binary (but not bool) ops we have a different function called `binary_fp`, like we have for unary ops?

PR #1079, opened 2024-05-05
Added ArcTan2 operation

yrahul3910:
## Proposed changes

This PR adds the arctan2 operation, as discussed in #1065.

## Checklist


mlx/backend/accelerate/primitives.cpp
```diff
@@ -193,6 +193,25 @@ void ArcTan::eval_cpu(const std::vector<array>& inputs, array& out) {
...
+    } else if (b.is_donatable() && b.itemsize() == out.itemsize()) {
+      out.copy_shared_buffer(b);
+    } else {
+      out.set_data(allocator::malloc_or_wait(out.nbytes()));
+    }
+    int size = a.data_size();
+    vvatan2f(out.data<float>(), a.data<float>(), b.data<float>(), &size);
+  } else {
+    eval(inputs, out);
+  }
```
awni, line 212: This is not correct. Could you follow the example of `Add` (or any other binary primitive)?

mlx/backend/common/primitives.cpp
```diff
@@ -94,6 +94,19 @@ void ArcTan::eval(const std::vector<array>& inputs, array& out) {
...
+  const auto& in1 = inputs[0];
+  const auto& in2 = inputs[1];
+  if (issubdtype(out.dtype(), inexact)) {
+    binary(in1, in2, out, detail::ArcTan2());
+  } else {
+    throw std::invalid_argument(
+        "[arctan2] Cannot compute inverse tangent of elements in array"
+        " with non floating point type.");
+  }
+}
```
awni, line 108: Move this to `binary.cpp` and follow the example of `LogAddExp` to only compile for floating point types: https://github.com/ml-explore/mlx/blob/main/mlx/backend/common/binary.cpp#L178-L197

mlx/primitives.cpp
```diff
@@ -5,6 +5,7 @@
 #include <numeric>
 #include <sstream>
 #include <stdexcept>
+#include <utility>
```
awni, line 8: Was this added intentionally?

mlx/primitives.h
```diff
@@ -314,6 +314,23 @@ class ArcTan : public UnaryPrimitive {
   void eval(const std::vector<array>& inputs, array& out);
 };
 
+class ArcTan2 : public UnaryPrimitive {
```
awni, line 317: Could you add it to this list as well: https://github.com/ml-explore/mlx/blob/main/mlx/compile.cpp#L38

mlx/ops.h
```diff
@@ -828,6 +828,9 @@ array arccos(const array& a, StreamOrDevice s = {});
 /** Arc Tangent of the elements of an array */
 array arctan(const array& a, StreamOrDevice s = {});
 
+/** Arc Tangent of elements of two arrays */
```
awni, line 831: Clarify this docstring please: `Inverse tangent of the ratio of two arrays`. Also please use `a` and `b` here to follow the Python API.

awni (changes requested):
Thanks for adding this. I left a few comments inline

Also you'll need to add the primitive to this list to enable compilation in CPU only mode: https://github.com/ml-explore/mlx/blob/main/mlx/backend/no_metal/primitives.cpp

mlx/ops.cpp
```diff
@@ -3,6 +3,7 @@
 #include <algorithm>
 #include <climits>
 #include <cmath>
+#include <cwchar>
```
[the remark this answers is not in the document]
yrahul3910: Whoops, seems there were a couple of these. I'll get rid of them.

mlx/backend/accelerate/primitives.cpp
yrahul3910: Ah, gotcha. I was following `Power`. I'll update it now :)

awni: Actually following `Power` might be better given there are no scalar equivalents. However, you would need to use `row_contiguous` here and not `contiguous` (which is a bug).

yrahul3910: Ah, I was stuck on that and just about to ask you :) I'll re-run tests/lint and push the commit to the PR.

mlx/backend/accelerate/primitives.cpp
```diff
@@ -193,6 +193,26 @@ void ArcTan::eval_cpu(const std::vector<array>& inputs, array& out) {
...
 
+void ArcTan2::eval_cpu(const std::vector<array>& inputs, array& out) {
+  assert(inputs.size() == 2);
+  auto& a = inputs[0];
+  auto& b = inputs[1];
+  if (out.dtype() == float32 && a.flags().row_contiguous &&
+      b.flags().row_contiguous) {
+    if (a.is_donatable() && a.itemsize() == out.itemsize()) {
+      out.copy_shared_buffer(a);
+    } else if (b.is_donatable() && b.itemsize() == out.itemsize()) {
```
awni, line 204: I don't think we need to check the `itemsize` here. `a` and `b` have to have the same type as `out` at this point.

mlx/primitives.cpp
```diff
@@ -402,6 +402,37 @@ std::pair<std::vector<array>, std::vector<int>> ArcTan::vmap(
...
+  return jvp(primals, cotangents, argnums);
+}
+
+std::vector<array> ArcTan2::jvp(
+    const std::vector<array>& primals,
+    const std::vector<array>& tangents,
+    const std::vector<int>& argnums) {
+  assert(primals.size() == 2);
+  assert(argnums.size() == 2);
+  array one = array(1., primals[0].dtype());
```
awni, line 419: Unused?

mlx/backend/common/binary.cpp
```diff
@@ -293,4 +293,25 @@ void BitwiseBinary::eval_cpu(const std::vector<array>& inputs, array& out) {
...
+  if (out.dtype() == float32) {
+    binary_op<float>(a, b, out, detail::ArcTan2());
+  } else if (out.dtype() == float16) {
+    binary_op<float16_t>(a, b, out, detail::ArcTan2());
+  } else if (out.dtype() == bfloat16) {
+    binary_op<bfloat16_t>(a, b, out, detail::ArcTan2());
+  } else if (issubdtype(out.dtype(), inexact)) {
+    std::ostringstream err;
+    err << "[arctan2] Does not support " << out.dtype();
+    throw std::invalid_argument(err.str());
```
awni, line 309: Did you try for `complex`? Does `std::atan2` not support it?

awni (commented):
This looks great! I left just a couple more comments then should be good to go.

mlx/primitives.cpp
yrahul3910: I wasn't very sure--I added it since `jvp` and `vjp` are `virtual` in `Primitive`

awni: I meant the `array one` not the function itself which is should be added.

yrahul3910: Ooh, yes, nice catch, thanks!

mlx/backend/common/binary.cpp
yrahul3910: I don't think so: the [docs for atan2](https://en.cppreference.com/w/cpp/numeric/math/atan2) don't seem to have an overload for it, and the [complex class has their own inverse trig functions, but not atan2](https://en.cppreference.com/w/cpp/numeric/complex/atan). Perhaps we could roll our own?

yrahul3910:
Thanks! Is that linux build supposed to pass? It seems like it's mainly having trouble finding the ArcTan2 symbol and I'm not sure what I missed.

awni:
Yes, linux tests should pass. I think you just need to add  to the list here: https://github.com/ml-explore/mlx/blob/main/mlx/backend/common/default_primitives.cpp#L36

That is just a way of saying that we should just use the default `eval` when accelerate is not available (e.g. on linux).

awni (approved):
Looks great, thanks!!

PR #1091, opened 2024-05-08
JIT compile option for binary minimization

awni:
- Adds a build flag `MLX_METAL_JIT` to reduce the Metal library size by using runtime compilation.
- Big refactor of unary, binary, ternary, copy, scatter, gather to allow JIT compilation
- Current MTL library size `15M mlx.metallib`


mlx/backend/metal/unary.cpp
```diff
@@ -0,0 +1,233 @@
...
+#include "mlx/backend/metal/kernels/defines.h"
+#include "mlx/backend/metal/utils.h"
+#include "mlx/primitives.h"
+
+namespace mlx::core {
+
+constexpr std::string_view unary_kernels = R"(
+[[kernel]] void {0}_v(
+    device const {1}* in,
+    device {2}* out,
```
awni, line 18: I put the kernels here (binary kernels in the `binary.cpp`, and so on). Maybe better to put them in a different file in `kernels/` that gets included? Not sure if either of you have a preference there @angeloskath @jagrit06

angeloskath: I kinda like it like this. If we do want to do this for more complicated kernels maybe we need a solution like the preamble but for unary, binary, ternary this is pretty great imho.

awni: I think it will likely be a combination of the two. Anything that needs to be formatted at runtime will likely be like this. And we can try to keep that bit to a minimum by having it be just the instantiations essentially.

The other stuff I will probably put in preambles. But I don't think it makes sense to do it as one giant preamble since that won't scale. So likely I will change the preamble / include stuff to be a little more modular

awni:
Benchmarks:

No degradation in token generation:

```
python -m mlx_lm.generate --model mlx-community/NeuralBeagle14-7B-4bit-mlx --prompt "Write a story about Albert Einstein" --temp 0.0 --max-tokens 256
```

```
Pre:
Prompt: 219.423 tokens-per-sec
Generation: 107.316 tokens-per-sec

Post:
Prompt: 219.580 tokens-per-sec
Generation: 107.562 tokens-per-sec
```

Transformer training:
```
Pre: Iter 30: Train loss 7.943, It/sec 5.911, Peak memory 5.534 (GB)
Post: Iter 30: Train loss 7.923, It/sec 5.912, Peak memory 5.534 (GB)
```

LeNet training:

```
Pre: Test accuracy 0.982, Time 2.792 (s)
Post: Test accuracy 0.983, Time 2.798 (s)
```

MNIST:

```
Pre: Test accuracy 0.937, Time 0.639 (s)
Post: Test accuracy 0.929, Time 0.638 (s)
```

mlx/backend/metal/ternary.cpp
```diff
@@ -0,0 +1,136 @@
...
+    kname << "_" << op << type_to_name(b);
+    kernel_name = kname.str();
+  }
+
+  auto& s = out.primitive().stream();
+  auto& d = metal::device(s.device);
+
+  MTL::ComputePipelineState* kernel;
+
+  if constexpr (mlx_metal_jit()) {
```
awni, line 61: @jagrit06 @angeloskath  this is an example using the `constexpr` to figure out if we are JITing or not. It's not as messy as I thought it would be provided the right helper utilities.

mlx/backend/metal/ternary.cpp
```diff
@@ -0,0 +1,136 @@
+// Copyright © 2024 Apple Inc.
+
+#include "mlx/backend/common/ternary.h"
+#include "mlx/backend/metal/device.h"
+#include "mlx/backend/metal/utils.h"
+#include "mlx/primitives.h"
+
+#ifndef MLX_METAL_JIT
```
awni, line 8: This is a minor downside of going the constexpr route. But the only place we really need to use the preprocessor.

awni:
@jagrit06 @angeloskath I think this is ready for review.

.circleci/config.yml
```diff
@@ -117,7 +117,13 @@ jobs:
           name: Run CPP tests
           command: |
             DEVICE=gpu METAL_DEVICE_WRAPPER_TYPE=1 METAL_DEBUG_ERROR_MODE=0 ./build/tests/tests
-            DEVICE=cpu ./build/tests/tests
+      - run:
+          name: Build small binary
+          command: |
+            source env/bin/activate
+            cd build/
+            cmake .. -DCMAKE_BUILD_TYPE=MinSizeRel -DBUILD_SHARED_LIBS=ON -DMLX_BUILD_CPU=ON -DMLX_BUILD_SAFETENSORS=OFF -DMLX_BUILD_GGUF=OFF -DMLX_METAL_JIT=ON
```
andresy, line 125: Curious about `MinSizeRel` (which I believe favors size to code speed), vs `Release` in terms of size. Also, have you tried strip?
https://stackoverflow.com/questions/38675403/how-to-config-cmake-for-strip-file/38676023

awni: It doesn't make a difference for what I tried for the GPU back-end. It might for the CPU, I haven't checked. But it's really meant as an option for deploying mostly on the GPU (when you want a really small CPU binary).

I haven't checked strip.. let me try and see if it reduces the size anymore.

mlx/backend/metal/CMakeLists.txt
```diff
@@ -1,33 +1,80 @@
...
 )
+make_jit_source(binary_ops)
+make_jit_source(ternary_ops)
+make_jit_source(
+  reduction
+  kernels/atomic.h
+  kernels/reduction/ops.h
+)
+make_jit_source(scatter)
+make_jit_source(gather)
```
angeloskath, line 52: Gather, scatter always jitted I guess. Honestly, it seems fitting. Do you foresee speedups as well for these going down this route?

mlx/backend/metal/kernels/reduction/ops.h
```diff
@@ -5,9 +5,11 @@
 #include <metal_atomic>
 #include <metal_simdgroup>
 
+#ifndef MLX_METAL_JIT
```
angeloskath, line 8: That's a bit annoying. I think we 'll waste a fair amount of time forgetting to do that. I don't have anything better to propose... just commenting.

angeloskath (approved):
It looks great!

In MLX tradition one can take the simple route and not provide a JIT option for a new kernel which means no change to the usage at all. That means that even for us, we can add kernels as we 're used to and add a JIT option later.

mlx/backend/metal/CMakeLists.txt
awni: Yes.. I kept them in the JIT for a few reasons:
- Kernels are pretty simple
- They take up a disproportionate amount of space for kernels which are mostly never used (e.g nidx > 2). I think 30Mb of the metal library.
- With the JIT we don't need to worry about a hard ceiling on the number of arguments (up to the buffer limit) which is kind of nice albeit perhaps not that useful.
- There is negligible additional change to cold start time JITing these.
- I didn't love the way we build these kernels in the preprocessor, it was kind of hard to follow and I didn't feel like adding it back ;)

> Do you foresee speedups as well for these going down this route?

If I understand your question correctly - do we plan to improve these kernels and will JITing them make it hard to do? I guess maybe.. if we get to that point and having them in the JIT is too annoying we can always revisit. (It's also pretty easy to add an instantiation which does not get included in the Metal library but is useful just for compile-time compilation.)

mlx/backend/metal/kernels/reduction/ops.h
awni: Yea it's annoying. Getting the includes right is annoying in general. There is probably a way to avoid this but it requires some care in what you include where and in what order (which is quite brittle).

awni: I was able to remove a few of these and I think I can remove the rest in #1132.

mlx/backend/metal/CMakeLists.txt
angeloskath: > If I understand your question correctly 

No I meant the exact opposite. These kernels feel very natural to be jitted. We could even imagine taking advantage of the fact that we are jitting to make on the fly kernels specific to index shapes for instance.

awni: Oh yes very good point! In fact I had considered doing specializations in the past for different index layouts but it was too unwieldy from a combinatorial perspective

PR #1092, opened 2024-05-09
IO thread-pool and device

angeloskath:
At this point this PR is more of a cautionary tale and a lesson for the future, I don't think this should ever be merged.

The interesting bits:

- This implements a thread pool that synchronizes within itself and could be used for interop parallelism on the CPU.
- Refactors a tiny bit the cpu synchronization task and computation task similar to metal which I think looks nicer.

The ugly bits:

- Most IO happens on a few files (often just one) which needs to be serialized with a mutex anyway.
- Dealing with synchronization between threads is actually complicated and we can't rely on array events as they can be out of order (and they aren't even signaled for within a stream). The solution is not the ugliest but it does use a lot of locks that can slow things down in latency critical situations (especially if there is contention).


angeloskath:
I will just close this but keep the branch around in case we want to extract the `cpu::make_task` refactoring later.

PR #1093, opened 2024-05-09
Allow unary ops to accept array like

awni:
Allows unary ops to accept array like (e.g. scalars, numpy arrays, etc). 

More consistent with binary ops and numpy. But let me know if you think there is a good reason not to allow it.


angeloskath (approved):
👍👍👍

awni:
Closes #1094

PR #1097, opened 2024-05-09
Comms

angeloskath:
Beginning of communication namespace (perhaps it should be named comms instead of dist). This is mostly to get feedback while implementing the rest of the primitives and figuring out how to package this in the distribution.

Interesting bits:

- `mlx::core::dist` defines a bunch of functions that are optionally implemented by a communication backend. Currently mpi.
- This defines a `Stream communication_stream()` and all communication operations go in that CPU stream.
- Primitives have transformations defined as expected which means we can write model parallel code with minimal fuss. Whenever sth needs to be communicated just communicate and gradients will flow accordingly. (I have to fix the gradient for all reduce sum but when everything is done it should be easy to use).


awni:
I might prefer the name `comm` or `comms` over `dist`. 

I also think `distributed` is fine.. and perhaps better since it is what every other package uses. One can always make a short name ..

mlx/dist/mpi/mpi.cpp
```diff
@@ -0,0 +1,114 @@
...
+bool is_available() {
+  return true;
+}
+
+Stream communication_stream() {
+  static std::shared_ptr<Stream> comm_stream = nullptr;
+  if (comm_stream == nullptr) {
+    comm_stream = std::make_shared<Stream>(
+        scheduler::scheduler().new_stream(Device::cpu));
+  }
```
awni, line 84: nit: maybe a unique_ptr is better here

awni: Nit: Also m use the free function `new_stream(Device::cpu)` rather than call the scheduler directly

mlx/dist/primitives.cpp
```diff
@@ -0,0 +1,76 @@
...
+
+  auto ensure_row_contiguous = [](const array& arr) {
+    if (arr.flags().row_contiguous) {
+      return arr;
+    } else {
+      array arr_copy(arr.shape(), arr.dtype(), nullptr, {});
+      copy(arr, arr_copy, CopyType::General);
+      return arr_copy;
+    }
+  };
```
awni, line 28: I wonder if there is a better strategy for this. The CPU copy could be kind of slow / might be better to use a GPU copy prior to the comm.

I wonder if we should consider adding a `ensure_contiguous(inputs)` op (which is meant for internal use only), but actually puts the copy in the graph if its needed.

mlx/dist/dist.h
```diff
@@ -0,0 +1,40 @@
...
+
+#include "mlx/array.h"
+
+namespace mlx::core::dist {
+
+/* Check if a communication backend is available */
+bool is_available();
+
+/* Return the communication stream. */
+Stream communication_stream();
```
awni, line 15: Maybe this should be an implementation detail for now rather than in the public API

awni:
This is very nice and simple! Looks great!

mlx/dist/mpi/mpi.cpp
angeloskath: The only reason I did the heap allocation was to avoid potential pitfalls with static initializations. I can do the above instead and reconsider if we encounter issues. I am also wondering what would happen if I used an optional initialized empty 🤔

awni: I think it should be ok.. either way it won't be initialized until this function is called the first time which will determine destruction order.

fishelegs:
Is there any example?
How to use it to train/finetune or inference?
Thanks!

mlx/dist/dist.h
```diff
@@ -0,0 +1,40 @@
...
+/**
+ * A dist::Group represents a group of independent mlx processes that can
+ * communicate. We must also be able to create sub-groups from a group in order
+ * to define more granular communication.
+ */
+struct Group {
+  virtual int rank() = 0;
+  virtual int size() = 0;
+  virtual std::shared_ptr<Group> split(int n) = 0;
+};
```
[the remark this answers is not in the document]
angeloskath: Not sure why that way would be better 🤷‍♂️ but I implemented it. It is pretty much the same thing as it just forces us to write the `MPIGroup` in mpi.cpp (as we would) and hide it behind a `std::shared_ptr<void>`. It is still pretty clean if not a little bit more cryptic and hard to follow and also forces us to have only one type of group at any given point.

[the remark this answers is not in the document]
awni: I guess I'm not a big fan of the pattern of virtualization to change behavior at compile-time, feels like the wrong tool for the job. Maybe there is another option that is cleaner than the payload (which is indeed a bit cryptic).

Also it's meant as a suggestion.. if you think it's a lot more readable in the previous version feel free to revert it.

.gitignore
```diff
@@ -82,3 +82,6 @@ build/
 
 # Jetbrains
 .cache
+
+# Negate mlx/dist
+!mlx/dist
```
awni, line 87: Nit: don't need that anymore

CMakeLists.txt
```diff
@@ -167,6 +167,11 @@ else()
   set(MLX_BUILD_ACCELERATE OFF)
 endif()
 
+find_package(MPI)
```
awni, line 170: Is this still needed now that you do it dynamically?

angeloskath: Well it depends :-) We can skip it and then I 'd add an mpi.h defining all the functions that I am using. It is basically finding the header.

angeloskath: Functions are fine actually, the types need defining.

awni: I see, let's leave it then! It should still build without MPI installed which is good.

angeloskath:
Maan that was a very nice suggestion. It feels so much better now with `no_cpu.cpp` removed and the copy moved to the distributed implementation.

awni (approved):
🚀

lin72h:
This is huge, wish someone could write a tutorial of how to connect 2 Macs use MLX

awni:
Usage docs coming soon!

sck-at-ucy:
I can wait to try this out!!

jami3f:
> Usage docs coming soon!

Any updates on docs for this? Can't wait to give it a go!

PR #1100, opened 2024-05-10
Add conjugate operator

barronalex:
## Proposed changes

This is part of the FFT PR but I thought I'd separate it out to make it easier to review.

## Checklist


python/src/ops.cpp
```diff
@@ -3027,6 +3027,22 @@ void init_ops(nb::module_& m) {
           inclusive (bool): The i-th element of the output includes the i-th
             element of the input.
       )pbdoc");
+  m.def(
+      "conjugate",
```
barronalex, line 3031: NumPy has both `np.conj` and `np.conjugate` which are aliases. Not sure if we want to do the same thing.

mlx/backend/common/primitives.cpp
```diff
@@ -203,6 +203,17 @@ void Concatenate::eval(const std::vector<array>& inputs, array& out) {
...
 
+void Conjugate::eval(const std::vector<array>& inputs, array& out) {
+  assert(inputs.size() == 1);
+  const auto& in = inputs[0];
+  if (out.dtype() == complex64) {
+    unary_fp(in, out, detail::Conjugate());
+  } else {
+    throw std::invalid_argument(
+        "[conjugate] conjugate must be called on complex input.");
+  }
```
awni, line 214: Numpy allows any type. Maybe we should do the same for consistency. The conjugate of a non complex type is just a buffer copy I think (see e.g. abs on an unsigned type)

mlx/backend/metal/primitives.cpp
```diff
@@ -588,6 +588,17 @@ void Concatenate::eval_gpu(const std::vector<array>& inputs, array& out) {
...
 
+void Conjugate::eval_gpu(const std::vector<array>& inputs, array& out) {
+  assert(inputs.size() == 1);
+  const auto& in = inputs[0];
+  if (out.dtype() == complex64) {
+    unary_op(inputs, out, "conj");
+  } else {
+    throw std::invalid_argument(
+        "[conjugate] conjugate must be called on complex input.");
+  }
```
awni, line 599: Same comment re type support.

mlx/ops.cpp
```diff
@@ -4101,6 +4101,15 @@ array number_of_elements(
       {a}));
 }
 
+array conjugate(const array& a, StreamOrDevice s /* = {} */) {
+  // Mirror NumPy's behaviour for real input
+  if (a.dtype() != complex64) {
+    return a;
+  }
```
awni, line 4108: Ah I see this now, you can disregard previous comments. I think this is ok instead of doing the buffer copy.

python/src/ops.cpp
awni: I would say yes

awni (approved):
Very nice! Could you add `conj` and `conjugate` to `docs/srs/ops.rst`? 

Also should we add `array.conj`?

awni (commented):
LGTM!

PR #1101, opened 2024-05-10
Improve error messaging in eval

### LINKED ISSUE #1083: [BUG] give better diagnostic message when calling compiled code with an eval in it -- currently "Attempting to eval an array without a primitive"
**Describe the bug**

From https://github.com/ml-explore/mlx-swift/issues/82 -- compiled code produces

```
ValueError: [eval] Attempting to eval an array without a primitive
```

**To Reproduce**

```python
import mlx.core as mx

def foo(x):
	y = mx.softmax(x, -1)
	logits = y[0, 0, 0].item()
	return mx.full([1, 100], mx.array(logits))


compiled = mx.compile(foo)

# without compile it works
# compiled = foo

for _ in range(10):
	x = mx.random.randint(0, 10, [1, 32, 50000]).astype(mx.float32)
	y = compiled(x)
	value = y[0, -1].item()
	print(value)

```

produces:

```
ValueError: [eval] Attempting to eval an array without a primitive
```

If run without the compile it gives a list of random numbers as expected.

**Expected behavior**

Compiled version should run the same way the non compiled code runs.

 - OS Version: [e.g. MacOS 14.1.2] macOS 14.3
 - Version [e.g. 0.7.0] hash 09f17778969381e5582c87bd1084846ebbcf5ceb

awni:
Closes #1083

### DIFF
--- a/mlx/transforms.cpp
+++ b/mlx/transforms.cpp
@@ -82,8 +82,15 @@ array eval_impl(std::vector<array> outputs, bool async) {
                 "[async_eval] Not allowed inside a graph transformation.");
           }
           if (!in.has_primitive()) {
-            throw std::invalid_argument(
-                "[eval] Attempting to eval an array without a primitive.");
+            if (in.is_tracer()) {
+              throw std::invalid_argument(
+                  "[eval] Attempting to eval an array during function"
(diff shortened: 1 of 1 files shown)


angeloskath (approved):
Nice!

PR #1102, opened 2024-05-10
Feature complete Metal FFT

barronalex:
## Proposed changes

A feature complete GPU FFT implementation in Metal.

## Supports

- All `n < 2^20`
- Real and Inverse FFTs: `fft`, `ifft`, `rfft`, `irfft`
- ND FFTs: `fft2`, `ifft2`, `rfft2`, `irfft2`, `fftn`, `ifftn`, `rfftn`, `irfftn`

## Algorithms
- A mixed radix out of place Stockham FFT for `n` where all prime factors `p` have `2 =< p <= 13`.
- [Rader's Algorithm](https://en.wikipedia.org/wiki/Rader%27s_FFT_algorithm) for `n` with one prime factor `p > 13` where `p-1` can be computed via Stockholm.
- [Bluestein's Algorithm](https://en.wikipedia.org/wiki/Chirp_Z-transform) for all other `n`.
- [Four Step FFT](https://en.wikipedia.org/wiki/Bailey%27s_FFT_algorithm) for `n > 4096` when the FFT can no longer be done purely in GPU shared memory.

## Performance

For `2 <= n < 512`, 1D complex to complex FFTs on my M1 Max, the average bandwidths are:
```
MLX GPU: 162.9 GB/s
MPS GPU: 69.3 GB/s
MLX CPU: 5.9 GB/s
```
So this implementation is about 2.3x faster than MPS on average and about 27x faster than CPU MLX which uses `pocketfft`.

This implementation does specialize for different values of n with Metal function constants so it will have more overhead than MPS on the first call for new Stockham/Rader sizes.

Radix 2-13
Bluestein's


mlx/backend/accelerate/primitives.cpp
```diff
@@ -255,6 +257,96 @@ void AsType::eval_cpu(const std::vector<array>& inputs, array& out) {
   eval(inputs, out);
 }
```
barronalex, line 259: Feels a bit wrong having this as a primitive but I wasn't sure if there's a better way to it.

mlx/backend/metal/device.cpp
```diff
@@ -357,7 +357,6 @@ MTL::Function* Device::get_function_(
   }
 
   mtl_func_consts->release();
```
barronalex, line 359: @jagrit06 I was getting segfaults caused by this release when using function constants, but couldn't figure out the best place in the code to move it to. Any idea where it should fit in?

awni: I think it's a bug to release that, and deleting it is correct. https://github.com/bkaradzic/metal-cpp/blob/metal-cpp_macOS14.2_iOS17.2/README.md#memory-allocation-policy

awni:
Very impressive perf!

Regarding the design, there is a big style difference from other MLX ops which we should change if possible. Basically you do the dispatch at the op-level rather than the Primitive level. I see how this might be easier as you have access to all the ops you need for different FFT algorithms, but I don't think we should do it this way. The compute graph should be more independent of the implementation details. Also, I don't think it makes sense for the FFT plans themselves should not be part of the compute graph (implementation detail).

This redesign may require some changes to our existing backend to make it workable for you to use the requisite back-end ops from the FFT primitive's `eval_gpu`.

barronalex:
That makes sense to me, it did feel a little anti-pattern bloating out the graph but the MLX api is just really convenient!
Let me give the re-write a go today, I don't think it'll be too bad.

awni:
> That makes sense to me, it did feel a little anti-pattern bloating out the graph but the MLX api is just really convenient!
Let me give the re-write a go today, I don't think it'll be too bad.

We have really bad support for doing stuff on arrays inside primitives (MLX wasn't really designed with that in mind 😓 ). But I think we can improve it a lot if needed.

mlx/backend/accelerate/primitives.cpp
```diff
@@ -255,6 +257,96 @@ void AsType::eval_cpu(const std::vector<array>& inputs, array& out) {
...
+  // in double precision for the overall numerical stability
+  // of Bluestein's FFT algorithm to be acceptable.
+  //
+  // MLX doesn't currently support float64, so instead we
+  // manually implement the required operations using accelerate.
+  //
+  // In numpy:
+  // w_k = np.exp(-1j * np.pi / N * (np.arange(-N + 1, N) ** 2))
+  // w_q = np.fft.fft(1/w_k)
+  // return w_k, w_q
```
angeloskath, line 273: What do you think of section IV.E of https://mc.stanford.edu/cgi-bin/images/7/75/SC08_FFT_on_GPUs.pdf . Would it solve our problem here to avoid double precision arithmetic?

angeloskath: Fig 6 is very promising :-)

angeloskath (commented):
Awesome perf and generally very nice work! Kudos!

I left a comment on BluesteinFFTSetup to maybe avoid the double precision math. I think it should be doable, let me know if I am missing something or if it feels too experimental.

barronalex:
OK that took a little while but I think the FFTs are in a reasonable state now:
- All the GPU planning/running logic has been moved to `metal/fft.cpp` so we're not bloating the graph at all
- Added a no transpose four step FFT implementation so big powers of two are fast now (~100-140GB/s on M1 Max)
- Added FFT to the JIT
- Refactored the reading/writing so we now support RFFT/IRFFT for Stockham/Rader/Bluestein/4 Step directly in the kernel

mlx/backend/accelerate/primitives.cpp
barronalex: This looks nice! I simplified the double precision part a bit so I think I'm going to keep it for now since it's not really an accuracy or performance bottleneck. Happy to revisit in the future though.

mlx/fft.cpp
```diff
@@ -81,7 +81,9 @@ array fft_impl(
   if (any_greater) {
     // Pad with zeros
     auto tmp = zeros(in_shape, a.dtype(), s);
-    in = scatter(tmp, std::vector<array>{}, in, std::vector<int>{}, s);
+    // GPU scatter for complex64 is NYI
+    in =
+        scatter(tmp, std::vector<array>{}, in, std::vector<int>{}, Device::cpu);
```
awni, line 86: Can we do that with a `slice_update` instead?

barronalex: That sounds nicer -- I'll update it

mlx/backend/metal/kernels/fft.h
```diff
@@ -0,0 +1,488 @@
...
+// - VkFFT (https://github.com/DTolm/VkFFT)
+// - Eric Bainville's excellent page (http://www.bealto.com/gpu-fft.html)
+
+#include <metal_common>
+
+#include "mlx/backend/metal/kernels/defines.h"
+#include "mlx/backend/metal/kernels/fft/radix.h"
+#include "mlx/backend/metal/kernels/fft/readwrite.h"
+#include "mlx/backend/metal/kernels/steel/defines.h"
+#include "mlx/backend/metal/kernels/utils.h"
```
awni, line 15: So this is why you don't need to use the `utils()` in the JIT, because its already included here by the preprocessor.

To keep the JIT source small, it would be better to move the includes that we already have in the JIT out of this file (e.g. `kernels/utils.h`) and use the `utils()` when constructing the JIT source.

You can include `kernels/utils.h` in `fft.metal` before you include `fft.h`. I would just turn off clang formatting for that whole file and it won't mess with the include order.

mlx/backend/metal/kernels/fft/radix.h
```diff
@@ -0,0 +1,307 @@
...
+for n=2,3,4,5,6,7,8,10,11,12,13.
+
+For n=2,3,4,5,6 we hand write the codelets.
+For n=8,10,12 we combine smaller codelets.
+For n=7,11,13 we use Rader's algorithm which decomposes
+them into (n-1)=6,10,12 codelets. */
+
+#include <metal_common>
+
+#include "mlx/backend/metal/kernels/utils.h"
```
awni, line 13: Note you should also remove the include here.

mlx/backend/metal/kernels/utils.h
```diff
@@ -339,3 +339,20 @@ inline int64_t simd_shuffle_down(int64_t data, uint16_t delta) {
...
+METAL_FUNC float2 complex_mul_conj(float2 a, float2 b) {
+  return float2(a.x * b.x - a.y * b.y, -a.x * b.y - a.y * b.x);
+}
+
+// Compute an FFT twiddle factor
+METAL_FUNC float2 get_twiddle(int k, int p) {
+  float theta = -2.0f * k * M_PI_F / p;
+
+  float2 twiddle = {metal::fast::cos(theta), metal::fast::sin(theta)};
+  return twiddle;
```
awni, line 357: If the only reason you are using `utils.h` is for these, it might be cleaner to just put those in `fft.h` instead. I think they also just fit better in fft.h if it works.. we have the complex64_t which should be used in general for complex muls.

barronalex: Agreed it's definitely a bit confusing otherwise. I've removed the `utils.h` import.

awni (approved):
🚀 🚀

PR #1105, opened 2024-05-12
No CPU option for binary minimization

awni:
## Proposed changes

Adds the option to disable building the CPU backend for GPU only programs.

One can produce a much C++ binary with the following:

```
cmake .. \
  -DCMAKE_BUILD_TYPE=MinSizeRel \
  -DBUILD_SHARED_LIBS=ON \
  -DMLX_BUILD_CPU=ON \
  -DMLX_BUILD_SAFETENSORS=OFF \
  -DMLX_BUILD_GGUF=OFF
```

Produces `libmlx.dylib` which is just 2.0 MB, compared to the default `libmlx.a` which is 17.0 MB


angeloskath (approved):
Nice!

PR #1111, opened 2024-05-14
Pass missing stream arg in array.flatten

zcbenz:
## Proposed changes

The `s` arg is missing in the `array.flatten` implementation.

## Checklist

### DIFF
--- a/python/src/array.cpp
+++ b/python/src/array.cpp
@@ -1139,7 +1139,7 @@ void init_array(nb::module_& m) {
              int start_axis,
              int end_axis,
              const StreamOrDevice& s) {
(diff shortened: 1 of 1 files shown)


python/src/array.cpp
```diff
@@ -1139,7 +1139,7 @@ void init_array(nb::module_& m) {
              int start_axis,
              int end_axis,
              const StreamOrDevice& s) {
-            return flatten(a, start_axis, end_axis);
+            return flatten(a, start_axis, end_axis, s);
```
awni, line 1142: Another good catch, thanks!!

awni (approved):

PR #1112, opened 2024-05-14
Remove unused code in mlx_scatter_args_slice

zcbenz:
## Proposed changes

The `up` is not used anywhere in the code below.

## Checklist

### DIFF
--- a/python/src/indexing.cpp
+++ b/python/src/indexing.cpp
@@ -526,14 +526,6 @@ std::tuple<std::vector<array>, array, std::vector<int>> mlx_scatter_args_slice(
 
   // If simple stride
   if (stride == 1) {
-    // Squeeze out singleton dims from the start of update
-    int s = 0;
-    for (; s < update.ndim() && update.shape(s) == 1; s++)
-      ;
-    auto up_shape =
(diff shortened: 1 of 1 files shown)


python/src/indexing.cpp
```diff
@@ -526,14 +526,6 @@ std::tuple<std::vector<array>, array, std::vector<int>> mlx_scatter_args_slice(
...
   // If simple stride
   if (stride == 1) {
-    // Squeeze out singleton dims from the start of update
-    int s = 0;
-    for (; s < update.ndim() && update.shape(s) == 1; s++)
-      ;
-    auto up_shape =
-        std::vector<int>(update.shape().begin() + s, update.shape().end());
-    auto up = reshape(update, up_shape);
-
```
awni, line 536: Why do you say its unused?

zcbenz: `up` is overwritten by `up = broadcast_to(update, up_shape_broadcast);` below, there is no code making use of the `up` in this line.

awni:
This was a bug. I added a fix in #1122. Closing this.

Issue #1026, opened 2024-04-23
[Feature] Cholesky decomposition


awni:
Add for the CPU using Lapack.

For the GPU [MPS has a Cholesky](https://developer.apple.com/documentation/metalperformanceshaders/mpsmatrixdecompositioncholesky?language=objc) which could be a good option to start with (following how we used to do bind MPS matmul):

barronalex:
I've been using the CPU Cholesky recently and unfortunately it's quite slow for large matrices.

A hybrid CPU/GPU Cholesky with MLX ops is about 2-3x faster than the pure CPU version for `N >= 4096`:

```python
def cholesky(A: mx.array, block_size: int = 512):
    N = A.shape[-1]
    L = mx.zeros_like(A)
    A = mx.array(A)

    # For numerical stability
    amax = A.abs().max(axis=range(1, A.ndim), keepdims=True)
    A /= amax

    for k in range(0, N, block_size):
        end = min(k + block_size, N)

        L[..., k:end, k:end] = mx.linalg.cholesky(A[..., k:end, k:end], stream=mx.cpu)

        if end < N:
            L_inv = mx.linalg.tri_inv(mx.swapaxes(L[..., k:end, k:end], -1, -2), upper=True, stream=mx.cpu)
            L[..., end:N, k:end] = A[..., end:N, k:end] @ L_inv
            A[..., end:N, end:N] -= mx.matmul(L[..., end:N, k:end], mx.swapaxes(L[..., end:N, k:end], -1, -2))

    L *= mx.sqrt(amax)
    return L
```

`N = 2048: 8ms -> 10ms speed up: 0.8x`
`N = 4096: 81ms -> 28ms speed up: 2.9x`
`N = 8192: 693ms -> 197ms speed up: 3.5x`
`N = 16384: 6010ms -> 2411ms speed up: 2.5x`

It seems to be similarly numerical accurate, if anything slightly better than the CPU version when you check $A = LL^{T}$ in `float64`.

We don't really have a great pattern for ops that force some computation on the CPU and some on the GPU but maybe it's worth merging anyway?

It could be quite hard to write a more performant GPU only kernel for a single matrix since the unblocked `cholesky` that runs on the CPU above can't easily be parallelized. The batched version could probably be quite a bit faster though.

awni:
That's pretty awesome that it's faster. A rare example of mixing CPU / GPU speeding things up!

[...]

PR #1119, opened 2024-05-14
Implemented Cholesky on CPU #1026

arn4:
## Proposed changes

Implemented Cholesky Decomposition for real matrix on CPU, as requested in #1026. 

There is nothing done for the GPU.

## Checklist

### DIFF
--- a/mlx/backend/common/cholesky.cpp
+++ b/mlx/backend/common/cholesky.cpp
@@ -13,6 +13,35 @@
 
 namespace mlx::core {
 
+namespace {
+
+// Delegate to the Cholesky factorization taking into account differences in
+// LAPACK implementations (basically how to pass the 'uplo' string to fortran).
+int spotrf_wrapper(char uplo, float* matrix, int N) {
+  int info;
+
+#ifdef LAPACK_FORTRAN_STRLEN_END
+  spotrf_(
+      /* uplo = */ &uplo,
+      /* n = */ &N,
+      /* a = */ matrix,
+      /* lda = */ &N,
+      /* info = */ &info,
+      /* uplo_len = */ static_cast<size_t>(1));
+#else
+  spotrf_(
+      /* uplo = */ &uplo,
+      /* n = */ &N,
+      /* a = */ matrix,
+      /* lda = */ &N,
+      /* info = */ &info);
+#endif
+
+  return info;
+}
+
+} // namespace
+
 void cholesky_impl(const array& a, array& factor, bool upper) {
   // Lapack uses the column-major convention. We take advantage of the fact that
   // the matrix should be symmetric:
@@ -33,17 +62,11 @@ void cholesky_impl(const array& a, array& factor, bool upper) {
   const int N = a.shape(-1);
   const size_t num_matrices = a.size() / (N * N);
 
-  int info;
   float* matrix = factor.data<float>();
 
   for (int i = 0; i < num_matrices; i++) {
     // Compute Cholesky factorization.
-    spotrf_(
-        /* uplo = */ &uplo,
-        /* n = */ &N,
-        /* a = */ matrix,
-        /* lda = */ &N,
-        /* info = */ &info);
+    int info = spotrf_wrapper(uplo, matrix, N);
 
     // TODO: We do nothing when the matrix is not positive semi-definite
     // because throwing an error would result in a crash. If we figure out how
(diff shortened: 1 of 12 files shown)


mlx/backend/common/cholesky.cpp
```diff
@@ -0,0 +1,86 @@
...
+
+#ifdef ACCELERATE_NEW_LAPACK
+#include <Accelerate/Accelerate.h>
+#else
+#include <lapack.h>
+#endif
+
+namespace mlx::core {
+
+void cholesky_impl(const array& a, array& T, bool upper) {
```
angeloskath, line 16: Nit: `T` is kind of a bad name for a variable such as this. `out` or `factor` or something along these lines is probably better.

mlx/backend/common/cholesky.cpp
```diff
@@ -0,0 +1,86 @@
...
+  char uplo;
+  if (upper) {
+    uplo = 'L';
+  } else {
+    uplo = 'U';
+  }
+
+  // The decomposition is computed in place, so just copy the input to the
+  // output.
+  copy(a, T, a.flags().row_contiguous ? CopyType::Vector : CopyType::General);
```
angeloskath, line 33: That's great and should also handle donation. Very nice.

python/src/linalg.cpp
```diff
@@ -260,4 +260,33 @@ void init_linalg(nb::module_& parent_module) {
...
+      nb::sig(
+          "def cholesky(a: array, upper: bool = False, *, stream: Union[None, Stream, Device] = None) -> array"),
+      R"pbdoc(
+        Compute the Cholesky decomposition of a real symmetric positive semi-definite matrix.
+
+        This function supports arrays with at least 2 dimensions. When the input
+        has more than two dimensions, the Cholesky decomposition is computed for each matrix
+        in the last two dimensions of ``a``.
+
+        If the input matrix is not symmetric positive semi-definite, behaviour is undefined.
```
angeloskath, line 279: It currently crashes. Maybe undefined is indeed better. We could just not throw in that case.

tests/linalg_tests.cpp
```diff
@@ -322,3 +322,29 @@ TEST_CASE("test matrix inversion") {
   CHECK(allclose(matmul(A_inv, A), identity, /* rtol = */ 0, /* atol = */ 1e-6)
             .item<bool>());
 }
+
+TEST_CASE("test matrix cholensky") {
```
angeloskath, line 326: Nit: typo cholensky -> cholesky

angeloskath (changes requested):
This looks fantastic! Left some tiny nitpicky comments for small typos and such.

mlx/backend/common/cholesky.cpp
arn4: If I understood correctly the donation, yes! It's done through ```CopyType::Vector```

python/src/linalg.cpp
arn4: Yes right, it indeed crashes when matrix is not positive semi-definite. The behaviour I wanted was:

- raise when the matrix is not positive semidefinite, since it is directly detected by Lapack
- undefined if the matrix is not symmetric, since adding a check will come an unnecessary overhead.

Maybe I should specify it in the doc, or do you still prefer not raising at all?

angeloskath: I changed it to do nothing and added a comment to add some error checking if we can throw from the implementation in the future. Otherwise, potentially crashing the program in an unrecoverable way based on the values of the input matrix seems an extremely bad idea to me.

mlx/backend/common/cholesky.cpp
```diff
@@ -0,0 +1,86 @@
...
+      throw std::runtime_error(msg.str());
+    }
+
+    // Zero out the upper/lower triangle while advancing the pointer to the
+    // next matrix at the same time.
+    for (int row = 0; row < N; row++) {
+      if (upper) {
+        std::fill(matrix, matrix + row, 0);
+      } else {
+        std::fill(matrix + row + 1, matrix + N, 0);
```
angeloskath, line 64: I missed it in the first review but here you were zeroing column-wise which is really inefficient especially for large matrices. I also use `std::fill` which may be faster if the compiler couldn't figure out that the previous loop was vectorizable.

angeloskath (approved):

PR #1120, opened 2024-05-14
Allow conversion to dlpack

### LINKED ISSUE #1080: Is dlpack supported?
We are looking to support mlx arrays in the ONNX in memory IR: https://onnxscript.ai/intermediate_representation/tensors.html

To create interoperability support we implement the `__dlpack__` protocol (https://dmlc.github.io/dlpack/latest/) on the tensor adapters (https://onnxscript.ai/intermediate_representation/tensors.html#computation-with-different-frameworks). 

I read https://ml-explore.github.io/mlx/build/html/usage/numpy.html and did a quick search but it doesn't seem mlx supports dlpack. Just wondering if there are any plans for it?

awni:
Closes #1080

### DIFF
--- a/python/src/array.cpp
+++ b/python/src/array.cpp
@@ -669,19 +669,14 @@ void init_array(nb::module_& m) {
             return a.shape(0);
           })
       .def("__iter__", [](const array& a) { return ArrayPythonIterator(a); })
-      .def(
-          "__getstate__",
-          [](const array& a) {
-            if (a.dtype() == bfloat16) {
-            }
-            return mlx_to_np_array(a);
-          })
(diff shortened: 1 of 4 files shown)


angeloskath (approved):

justinchuby:
Is __dlpack_device__() needed?

awni:
Is it needed by you? Array's don't have devices in MLX they live in shared memory accessible to both CPU/GPU.. so it doesn't really make sense to have it. 

But for interop purposes we could add it and return like the default device (usually `METAL`) if that's useful.

justinchuby:
I remember seeing frameworks raising errors when the method was not implemented.

PR #1124, opened 2024-05-15
Block sparse qmm

angeloskath:
The `block_sparse_mm` equivalent for quantized matmuls. A big diff but mostly boilerplate code and refactoring.


mlx/ops.cpp
```diff
@@ -3430,6 +3469,51 @@ array dequantize(
...
+  lhs_indices = broadcast_to(lhs_indices, out_bsx_shape, s);
+  rhs_indices = broadcast_to(rhs_indices, out_bsx_shape, s);
+
+  // Compute the full output shape
+  auto out_shape = out_bsx_shape;
+  out_shape.push_back(x.shape(-2));
+  out_shape.push_back(w_outer_dims);
+
+  // and output type
+  auto out_type = result_type(x, scales, biases);
```
jagrit06, line 3505: Do we need to add a check or astype here to make sure all 3 are the same dtype ?

mlx/primitives.cpp
```diff
@@ -2375,6 +2375,76 @@ bool QuantizedMatmul::is_equivalent(const Primitive& other) const {
...
+              stream()),
+          x.shape(),
+          stream()));
+    }
+
+    // gradient wrt to w_q, scales or biases
+    else {
+      throw std::runtime_error(
+          "BlockSparseQMM::vjp no gradient wrt the quantized matrix yet.");
+    }
```
jagrit06, line 2431: Maybe we can add a separate error message if gradients w.r.t to the indices are requested

mlx/primitives.cpp
```diff
@@ -2375,6 +2375,76 @@ bool QuantizedMatmul::is_equivalent(const Primitive& other) const {
...
+std::vector<array> BlockSparseQMM::jvp(
+    const std::vector<array>& primals,
+    const std::vector<array>& tangents,
+    const std::vector<int>& argnums) {
+  throw std::runtime_error("BlockSparseQMM::jvp NYI");
+}
+
+bool BlockSparseQMM::is_equivalent(const Primitive& other) const {
+  const BlockSparseQMM& qm_other = static_cast<const BlockSparseQMM&>(other);
+  return group_size_ == qm_other.group_size_ && bits_ == qm_other.bits_;
```
jagrit06, line 2445: Should we also check `transpose_` here ?

jagrit06 (approved):
Looks great! Just a few small comments here and there

mlx/ops.cpp
angeloskath: Good catch I forgot to use it below to cast `x`, `scales` and `biases` to out_type.

angeloskath:
@awni and @jagrit06 I made two changes that are not directly related to the `block_sparse_qmm` but general quantization. So I 'd appreciate a review. Also if you think that they should be on a different PR (or maybe just the 2nd change) let me know and I can merge this before these commits and make another PR for this.

The changes are the following:
1. `quantize` and `dequantize` now support arbitrary shapes instead of only 2D matrices
2. I added a `to_quantized` method to `Linear` and `Embedding` and changed `nn.quantized` to check for modules with the `to_quantized` method implemented. This allows arbitrary modules to support quantization using `nn.quantize` without MLX knowing anything about them.

awni (commented):
LGTM 🚀

awni (approved):

PR #1126, opened 2024-05-15
Fix extensions

awni:
I also enabled the extension test in Circle since a few people are using it and it will be good not to break it moving forward.

### DIFF
--- a/docs/src/dev/extensions.rst
+++ b/docs/src/dev/extensions.rst
@@ -504,10 +504,10 @@ below.
 
         // Encode input arrays to kernel
         compute_encoder.set_input_array(x, 0);
-        compute_encoder.set_output_array(y, 1);
+        compute_encoder.set_input_array(y, 1);
 
         // Encode output arrays to kernel
-        compute_encoder.set_output_buffer(out, 2);
+        compute_encoder.set_output_array(out, 2);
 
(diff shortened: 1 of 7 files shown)


docs/src/dev/extensions.rst
```diff
@@ -503,11 +503,11 @@ below.
         size_t nelem = out.size();
 
         // Encode input arrays to kernel
-        set_array_buffer(compute_encoder, x, 0);
-        set_array_buffer(compute_encoder, y, 1);
+        compute_encoder.set_input_array(x, 0);
+        compute_encoder.set_output_array(y, 1);
```
objorkman, line 507: `set_input_array` as well to match

docs/src/dev/extensions.rst
```diff
@@ -503,11 +503,11 @@ below.
...
 
         // Encode output arrays to kernel
-        set_array_buffer(compute_encoder, out, 2);
+        compute_encoder.set_output_buffer(out, 2);
```
objorkman, line 510: `set_output_array` instead

docs/src/dev/extensions.rst
awni: Thanks!

angeloskath (approved):
Nice!

PR #1132, opened 2024-05-17
More jitting

awni:
Jits:
- Sort
- Scan
- Arange
- Softmax
- Reduce

Down to 10MB.

This is built on #1091. So will rebase on main once that land and the diff will not be so big.


awni:
Ok this one is ready for review. I removed the remaining include guards (i.e.  `#ifdef MLX_METAL_JIT`), so that is a bit cleaner.

angeloskath (approved):
Looks great!

PR #1139, opened 2024-05-19
Option to JIT steel gemm / conv

awni:
Metal library down to 3.2 MB when enabling `MLX_METAL_JIT=ON`.

This is built on #1091 and #1132 so the diff looks much bigger than it actually is.


angeloskath (approved):
🚀

Issue #1143, opened 2024-05-21
Implement trace analogical to numpy.trace


JoeJoe1313:
Is it possible to have `trace` explicitly implemented as in `numpy.trace`?

awni:
We could add this. I do not think it would be too difficult to implement in terms of `diag` and `sum`. I will mark as enhancement if someone wants to work on it.

PR #1147, opened 2024-05-22
add mx.trace (#1143)

abeleinin:
## Adds mx.trace support

Implements `mx.trace` (#1143) in C++ and Python.

## Checklist


python/src/ops.cpp
```diff
@@ -4052,6 +4052,45 @@ void init_ops(nb::module_& m) {
...
+      },
+      nb::arg(),
+      "offset"_a = 0,
+      "axis1"_a = 0,
+      "axis2"_a = 1,
+      "dtype"_a = nb::none(),
+      nb::kw_only(),
+      "stream"_a = nb::none(),
+      nb::sig(
+          "def trace(a: array, offset: int = 0, axis1: int = 0, axis2: int = 1, dtype = Optional[Dtype] = None, stream: Union[None, Stream, Device] = None) -> array"),
```
awni, line 4076:
```suggestion
          "def trace(a: array, /, offset: int = 0, axis1: int = 0, axis2: int = 1, dtype = Optional[Dtype] = None, *, stream: Union[None, Stream, Device] = None) -> array"),
```

awni (commented):
This is really nicely done, thanks! I left two minor comments. Could you address then we can merge?

mlx/ops.cpp
```diff
@@ -4114,6 +4114,59 @@ array diag(const array& a, int k /* = 0 */, StreamOrDevice s /* = {} */) {
...
+    throw std::out_of_range(msg.str());
+  }
+
+  if (ax1 == ax2) {
+    throw std::invalid_argument(
+        "[trace] axis1 and axis2 cannot be the same axis");
+  }
+
+  return astype(
+      sum(diagonal(a, offset, axis1, axis2), -1), dtype, to_stream(s));
```
awni, line 4154: One more thing, I would do the `astype` before the `sum`. Part of the purpose of the `dtype` is to enable the reduction in higher precision if needed. So if you take the trace of an int8 matrix and you want the reduction in `int32` so it doesn't overflow.

awni (approved):
Thanks!!

mlx/ops.cpp
```diff
@@ -4113,6 +4113,63 @@ array diag(const array& a, int k /* = 0 */, StreamOrDevice s /* = {} */) {
...
+    int offset,
+    int axis1,
+    int axis2,
+    StreamOrDevice s /* = {} */) {
+  auto dtype = a.dtype();
+  return trace(a, offset, axis1, axis2, dtype, to_stream(s));
+}
+array trace(const array& a, StreamOrDevice s /* = {} */) {
+  auto dtype = a.dtype();
+  return trace(a, 0, 0, 1, dtype, to_stream(s));
```
awni, line 4170: Actually, one last thing. You don't need to call `to_stream` here every time. You should be able to simply pass in the stream as `s`. Will be a bit more readable.  Do you mind updating that?

abeleinin: sure thing!

Issue #1160, opened 2024-05-27
[BUG] Compiled mx.eval(model.state) raises “Attempting to eval an array without a primitive” with mlx.optimizers.Adam


igor-pechersky:
**Describe the bug**
Compiled mx.eval(model.state) raises “Attempting to eval an array without a primitive” in MNIST example with mlx.optimizers.Adam 

**To Reproduce**

in the code of https://github.com/ml-explore/mlx-examples/tree/main/mnist
replace in  https://github.com/ml-explore/mlx-examples/blob/ca7ce60c911c107d0ffa34554f5ca41a81e0fdc6/mnist/main.py#L65C5-L65C55

```python
optimizer = optim.SGD(learning_rate=learning_rate)
```
by
```python
optimizer = optim.Adam(learning_rate=learning_rate)
```

**Expected behavior**
Error raised
```
mx.eval(model.state)
RuntimeError: [eval] Attempting to eval an array without a primitive. This may be a bug, please file an issue here:  https://github.com/ml-explore/mlx/issues.
```

**Desktop (please complete the following information):**
 - OS Version: MacOS 14.5, Apple M3 Max (64 GB)
 - MLX Version 0.14.0

**Additional context**
1. Error does **not** occur if MLX compilation is disabled (`mx.disable_compile()`)
2. Error occurs with MLX versions 0.10.0 - 0.14.0
3. Error occurs on the **last** minibatch (for MNIST with batchsize=256 - it occurs with minibatch 234)
4. Error occurs when optimizer is set to  mlx.optimizers.{Adam, AdamW, RMSprop, Adagrad} , but not mlx.optimizers.SGD
5. Error does **not** occur if each epoch runs as single batch
```python
    for e in range(num_epochs):
        tic = time.perf_counter()
        loss=step(train_images, train_labels)
        mx.eval(model.state)
        accuracy = eval_fn(test_images, test_labels)
        toc = time.perf_counter()
        print(
            f"Epoch {e}: Test accuracy {accuracy.item():.3f},"
            f" Time {toc - tic:.3f} (s)"
        )
```

awni:
In this case it's not a bug..though I don't know why the error message makes it seem so. That's something to fix.

For optimizers with state, the state has to be captured in the inputs/outputs of `compile` as well, like so:


```python
    state = [model.state, optimizer.state]
    
    @partial(mx.compile, inputs=state, outputs=state)
    def step(X, y):
      ...
```

And then evaluate the full state after each iteration:
[...]

PR #1161, opened 2024-05-28
Fix a couple bugs

### LINKED ISSUE #1063: [BUG] libc++abi crash when using recurrent layer and transformer
**Describe the bug**

```
(libc++abi: terminating due to uncaught exception of type std::runtime_error: 
[compile] Too many inputs/outputs fused in the Metal Compiled primitive which 
exhausted the available argument buffers for the kernel. Please file an issue with 
the function that results in this error. The name of the kernel is
'Nf4MultiplyABOf4AddEFPf4AddOGQf4AddPHRf4AddQISf4MultiplyRJTf4AddDSUf4MultiplyCTVf4SquareRWf4MultiplyVLXf4AddKWYf4SqrtXZf4AddYMAAf4DivideUZABf4SubtractNAA_VVVVVVVVVVVVV_f4f4f4f4f4f4f4f4f4f4f4f4f4_11160318154034397263_strided_dynamic')
```

In:

```python
def __call__(self, x):
        L = x.shape[1]
        mask = nn.MultiHeadAttention.create_additive_causal_mask(L)
        x = self.embed(x)
        x = x + self.pe(mx.arange(L))
        x = self.transformer(x, mask)
        x = self.context_recurrent(x) + x
        x = self.transformer2(x, mask)
        x = self.out_proj(x)
        return x
```

If self.context_recurrent is any of nn.LSTM, RNN, or GRU, the above crash happens.

**To Reproduce**

Insert  `x = nn.RNN(x)` between two transformer layers causes the crash.

Complete code <https://github.com/domschl/mlx-poet/blob/cedac548256a1bd2a1bb33362cf9d99f22a360c7/mlx_poet_bug.py>
(requires `pip install ml-indie-tools`)

**Expected behavior**

No crash, and if necessary clear error message. I've checked that there is no tensor-shape problem.

 - OS Version: 14.4.1 (23E224)
 - Version 0.12.0

**Additional context**

### LINKED ISSUE #1158: [BUG] JIT compile mode does not work with LoRA
Not entirely sure what's going on there but I'll investigate this one.

```
libc++abi: terminating due to uncaught exception of type std::runtime_error: [metal::Device] Unable to build metal library from source
program_source:1222:34: error: use of undeclared identifier 'RMSNormVJP'
all_reduce<float16_t, float16_t, RMSNormVJP<float16_t>>(
                                 ^
```

awni:
- Fix RMS VJP reductions with JIT - closes #1158 
- Fix # buffers in compiled primitive - closes #1063 
- Better eval error message as per #1160
- Fix bug compiling with `inf` and half precision types

### DIFF
--- a/mlx/backend/common/make_compiled_preamble.sh
+++ b/mlx/backend/common/make_compiled_preamble.sh
@@ -28,6 +28,7 @@ const char* get_kernel_preamble() {
 return R"preamble(
 $INCLUDES
 $CONTENT
+using namespace mlx::core;
 using namespace mlx::core::detail;
 )preamble";
 }
--- a/mlx/backend/metal/compiled.cpp
+++ b/mlx/backend/metal/compiled.cpp
@@ -56,10 +56,13 @@ inline void build_kernel(
     } else {
       add_indices = true;
       os << "    device const " << get_type_string(x.dtype()) << "* " << xname
-         << " [[buffer(" << cnt++ << ")]]," << std::endl
-         << "    constant const size_t* " << xname << "_strides [[buffer("
-         << cnt++ << ")]]," << std::endl;
+         << " [[buffer(" << cnt++ << ")]]," << std::endl;
     }
+  }
+
+  if (add_indices) {
+    os << "    constant const size_t* in_strides [[buffer(" << cnt++
+       << ")]],\n";
   }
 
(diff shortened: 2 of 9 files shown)


angeloskath (approved):
Love the change in compiled!!!

PR #1167, opened 2024-05-29
Stable cumprod grad at 0

awni:
Fix cumprod at 0.

I'm not sure this is the best solution. It's linear but quite involved. Sending a PR so others can think about this


angeloskath (approved):
I love the tests!

PR #1168, opened 2024-05-29
Fix matvec vector stride bug

### LINKED ISSUE #1163: [BUG]  Wrong result for sliced matmul on GPU
**Describe the bug**

matmul behavior change between GPU and CPU as seen in https://github.com/ml-explore/mlx-swift/issues/94

**To Reproduce**

Include code snippet
```python
import mlx.core as mx

mArray = [
    [0.943755, 0.162902, -0.287733, -0.241071],
    [0.0669876, -0.946367, -0.316074, 0.150359],
    [-0.32379, 0.279022, -0.90405, -0.54078],
    [0, 0, 0, 1]
]

m = mx.array(mArray)

r = m[0:3, 0:3]
t = m[0:3, 3:4]

print(mx.matmul(r, t, stream=mx.cpu))
print(mx.matmul(r, t, stream=mx.gpu))
```

**Expected behavior**

Expected that the two prints should produce identical (or nearly so) results.  Actual:

```
array([[-0.0474179],
       [0.0124829],
       [0.608902]], dtype=float32)
array([[0.0557015],
       [0.219578],
       [0.952311]], dtype=float32)
```

 - OS Version: MacOS 14.4
 - Version a87ef5bfc181afdb5b8630dff118c8ea82eefea7

jagrit06:
## Proposed changes

Fixes #1163 and added similar tests
The issue comes down to the fact that a vector of shape [K, 1] must have stride 1 where shape is K but the matmul is more flexible so the check to make copies had to be updated

## Checklist

### DIFF
--- a/mlx/backend/metal/matmul.cpp
+++ b/mlx/backend/metal/matmul.cpp
@@ -710,15 +710,19 @@ void Matmul::eval_gpu(const std::vector<array>& inputs, array& out) {
   /////////////////////////////////////////////////////////////////////////////
   // Init checks and prep
 
+  int M = a_pre.shape(-2);
+  int N = b_pre.shape(-1);
+  int K = a_pre.shape(-1);
+
   // Keep a vector with copies to be cleared in the completed buffer to release
   // the arrays
   std::vector<array> copies;
-  auto check_transpose = [&copies, &s](const array& arr) {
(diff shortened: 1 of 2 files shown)


awni (approved):
🚀 thanks for the quick fix!!

PR #1174, opened 2024-06-02
Doc error for default for scale in SinusoidalPositionalEncoding

domschl:
Code uses the inverse, which is correct.

## Proposed changes

Documentation fix, made consistent with code for scale = sqrt(2/dims) default.

## Checklist


angeloskath (approved):
Thanks for the fix!

PR #1177, opened 2024-06-03
Add some internal GPU apis

barronalex:
Add some internal apis for calling `unary`/`binary`/`ternary` gpu ops when building primitives.
Also add `slice/concatenate/pad`.

These are used in the FFT PR but hopefully will be useful for future ops.

## Proposed changes

## Checklist


awni (approved):
Looks great!

I think there might be some challenges with the JIT and using those in place ops. The JIT expects the output to have a valid primitive in some cases. We probably will need to change that and just use the passed in op string instead.

PR #1178, opened 2024-06-03
In place all-reduce and forgiving init

angeloskath:
This PR adds an in-place all reduce if the array is donatable. It also makes init return a singleton group if the distributed backend wasn't available (except if otherwise requested).

The latter makes for simpler code as it can be written as if distributed is always available, e.g.

```python
if mx.distributed.is_available():
    foo = mx.distributed.all_reduce(foo)
```

becomes simply

```python
foo = mx.distributed.all_reduce(foo)
```

### DIFF
--- a/mlx/distributed/mpi/mpi.cpp
+++ b/mlx/distributed/mpi/mpi.cpp
@@ -236,7 +236,7 @@ bool is_available() {
   return mpi().is_available();
 }
 
-Group init(bool strict) {
+Group init(bool strict /* = false */) {
   static std::shared_ptr<MPIGroupImpl> global_group = nullptr;
 
   if (global_group == nullptr) {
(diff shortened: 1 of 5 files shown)


mlx/distributed/mpi/mpi.cpp
```diff
@@ -235,14 +236,18 @@ bool is_available() {
   return mpi().is_available();
 }
 
-Group init() {
+Group init(bool strict) {
```
awni, line 239: Nit:

```suggestion
Group init(bool strict /* = false */) {
```

mlx/distributed/no_distributed.cpp
```diff
@@ -20,7 +20,7 @@ bool is_available() {
   return false;
 }
 
-Group init() {
+Group init(bool strict) {
```
awni, line 23:
```suggestion
Group init(bool strict /* = false */) {
```

awni (approved):
Looks good!

PR #1179, opened 2024-06-04
Add view op

### LINKED ISSUE #1042: [Feature Request] Support `numpy.ndarray.view`
A perfect implementation is complex, but even just being able to change to another dtype of the same size would make it easier to do bitwise advanced processing in mlx.
```py
np.array(3.14, dtype=np.float32).view(dtype=np.int32)
# array(1078523331, dtype=int32)
# -> 4048f5c3 .. raw
```

awni:
Adds a view op and primitive. Closes #1042

### DIFF
--- a/tests/ops_tests.cpp
+++ b/tests/ops_tests.cpp
@@ -1,5 +1,6 @@
 // Copyright © 2023-2024 Apple Inc.
 #include <cmath>
+#include <iostream> // TODO
 #include <numeric>
 
 #include "doctest/doctest.h"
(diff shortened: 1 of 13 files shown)


tests/ops_tests.cpp
```diff
@@ -1,5 +1,6 @@
 // Copyright © 2023-2024 Apple Inc.
 #include <cmath>
+#include <iostream> // TODO
```
angeloskath: Forgotten.

angeloskath (approved):
Perfect! 🤩

I also wanted to implement that op when I saw the issue but never got around to it.

PR #1180, opened 2024-06-04
add softmin, hardshrink, hardtanh

starskyreverie:
## Proposed changes

hey :) just added three more obscure act functions, the more the merrier

can find docs for each in pytorch:

[softmin](https://pytorch.org/docs/stable/generated/torch.nn.Softmin.html)
[hardshrink](https://pytorch.org/docs/stable/generated/torch.nn.Hardshrink.html)
[hardtanh](https://pytorch.org/docs/stable/generated/torch.nn.Hardtanh.html)

## Checklist


angeloskath (approved):
Looks great, thanks!

I would be interested to know if any model actually uses these functions. I will merge those since they are quite well written and tested but it would really help in the future if we have a reason to add more activation functions. Something like all "blah" models use `hard_shrink` so it would be useful to not implement it separately in each model.

starskyreverie:
Sounds good, and thank you :)

Definitely agree, in this case: 

`hard_tanh` is used for clipping values within a specific range, generally useful in reinforcement learning and for RNNs. For example:

```python
class RNN:
    def __init__(self, input_size, hidden_size):
        self.input_size = input_size
        self.hidden_size = hidden_size
        self.Wxh = random_matrix(hidden_size, input_size)  # weight matrix for input
        self.Whh = random_matrix(hidden_size, hidden_size)  # weight matrix for hidden state
        self.bh = random_vector(hidden_size)  # bias for hidden state
        self.Why = random_matrix(output_size, hidden_size)  # weight matrix for output
        self.by = random_vector(output_size)  # bias for output

    def hard_tanh(self, x, min_val=-1.0, max_val=1.0):
        # clips values within the range [-1, 1], preventing exploding gradients
        return max(min(x, max_val), min_val)

    def forward(self, inputs):
        h = zero_vector(self.hidden_size) 
        for t in range(len(inputs)):
            x = inputs[t]
            # hard_tanh clips values and stabilizes training
            h = self.hard_tanh(self.Wxh @ x + self.Whh @ h + self.bh)
        output = self.Why @ h + self.by
        return output
```

`hard_shrink` is generally good for enforcing sparsity, e.g. helpful in sparse auto-encoders and denoising applications. 

`softmin` probably has a decent case for being removed, it's sometimes used in ranking models where lower values should have higher probabilities but honestly not seen much, just since the specific case (ranking lower values higher) is rare.

PR #1184, opened 2024-06-04
Add docs for the distributed namespace

angeloskath:
As title.

### DIFF
--- a/docs/src/python/distributed.rst
+++ b/docs/src/python/distributed.rst
@@ -15,5 +15,5 @@ made available.
     Group
     is_available
     init
-    all_reduce_sum
+    all_sum
     all_gather
--- a/docs/src/python/distributed.rst
+++ b/docs/src/python/distributed.rst
@@ -1,0 +1,19 @@
+.. _distributed:
+
+.. currentmodule:: mlx.core.distributed
+
+Distributed Communication
+==========================
+
+MLX provides a distributed communication package using MPI. The MPI library is
+loaded at runtime; if MPI is available then distributed communication is also
+made available.
+
+.. autosummary::
+   :toctree: _autosummary
+
+    Group
+    is_available
+    init
+    all_reduce_sum
+    all_gather
--- a/docs/src/usage/distributed.rst
+++ b/docs/src/usage/distributed.rst
@@ -23,13 +23,14 @@ machine. The minimal distributed program in MLX is as simple as:
 
 .. code:: python
 
-   import mlx.core as mx
+    import mlx.core as mx
 
-   x = mx.distributed.all_reduce_sum(mx.ones(10))
-   print(x)
+    world = mx.distributed.init()
+    x = mx.distributed.all_reduce_sum(mx.ones(10))
+    print(world.rank(), x)
 
-The program above computes sums the array ``mx.ones(10)`` across all
-distributed processes. If simply launched with ``python``, however, only one
+The program above sums the array ``mx.ones(10)`` across all
+distributed processes. If simply run with ``python``, however, only one
 process is launched and no distributed communication takes place.
 
 To launch the program in distributed mode we need to use ``mpirun`` or
@@ -39,8 +40,8 @@ following:
 .. code:: shell
 
     $ mpirun -np 2 python test.py
-    array([2, 2, 2, ..., 2, 2, 2], dtype=float32)
-    array([2, 2, 2, ..., 2, 2, 2], dtype=float32)
+    1 array([2, 2, 2, ..., 2, 2, 2], dtype=float32)
+    0 array([2, 2, 2, ..., 2, 2, 2], dtype=float32)
 
 The above launches two processes on the same (local) machine and we can see
 both standard output streams. The processes send the array of 1s to each other
@@ -50,15 +51,15 @@ print 4 etc.
 Installing MPI
 ---------------
 
-MPI can be installed with homebrew, using the anaconda package manager or
+MPI can be installed with Homebrew, using the Anaconda package manager or
 compiled from source. Most of our testing is done using ``openmpi`` installed
-with the anaconda package manager as follows:
+with the Anaconda package manager as follows:
 
 .. code:: shell
 
     $ conda install openmpi
 
-Installing with homebrew may require specifying the location of ``libmpi.dyld``
+Installing with Homebrew may require specifying the location of ``libmpi.dyld``
 so that MLX can find it and load it at runtime. This can simply be achieved by
 passing the ``DYLD_LIBRARY_PATH`` environment variable to ``mpirun``.
 
@@ -76,14 +77,18 @@ debug connectivity issues is the following:
 * ``ssh hostname`` works from all machines to all machines without asking for
   password or host confirmation
 * ``mpirun`` is accessible on all machines. You can call ``mpirun`` using its
-  full path to force a specific prefix on all machines.
+  full path to force all machines to use a specific path.
 * Ensure that the ``hostname`` used by MPI is the one that you have configured
-  in the ``.ssh/config`` files on all machines. MPI could be using ``foo`` from
-  a ``hostname`` defined as ``foo.bar.com`` if the ``hostname`` of the current
-  machine is ``baz.bar.com``.
+  in the ``.ssh/config`` files on all machines.
+
+.. note::
+  For an example hostname ``foo.bar.com`` MPI could be using only ``foo`` as
+  the hostname passed to ssh, especially if the current host's name is also
+  ``*.bar.com``.
 
 An easy way to pass the host names to MPI is using a host file. A host file
-looks like the following:
+looks like the following, where ``host1`` and ``host2`` should be the fully
+qualified domain names or IPs for these hosts.
 
 .. code::
 
@@ -91,8 +96,8 @@ looks like the following:
     host2 slots=1
 
 When using MLX, it is very likely that you want to use 1 slot per host, ie one
-process per host.  Moreover, the hostfile needs to also contain the current
-host if you wlso ant to run on the local host. Passing the host file to
+process per host.  The hostfile also needs to contain the current
+host if you want to run on the local host. Passing the host file to
 ``mpirun`` is simply done using the ``--hostfile`` command line argument.
 
 Training Example
--- a/docs/src/usage/distributed.rst
+++ b/docs/src/usage/distributed.rst
@@ -82,9 +82,8 @@ debug connectivity issues is the following:
   in the ``.ssh/config`` files on all machines.
 
 .. note::
-  For an example hostname ``foo.bar.com`` MPI could be using only ``foo`` as
-  the hostname passed to ssh, especially if the current host's name is also
-  ``*.bar.com``.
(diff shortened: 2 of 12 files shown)


docs/src/python/distributed.rst
```diff
@@ -0,0 +1,19 @@
...
+loaded at runtime; if MPI is available then distributed communication is also
+made available.
+
+.. autosummary::
+   :toctree: _autosummary
+
+    Group
+    is_available
+    init
+    all_reduce_sum
```
awni, line 18: Sorry for the random comment, but the thought just ocurred to me, isn't it more common to just use `all_reduce` as the op name? Presumably the intention with the `sum` suffix is to preserve flexibility (if one wanted a max or prod)?

docs/src/usage/distributed.rst
```diff
@@ -0,0 +1,162 @@
...
+machine. The minimal distributed program in MLX is as simple as:
+
+.. code:: python
+
+   import mlx.core as mx
+
+   x = mx.distributed.all_reduce_sum(mx.ones(10))
+   print(x)
+
+The program above computes sums the array ``mx.ones(10)`` across all
```
awni, line 31:
```suggestion
The program above sums the array ``mx.ones(10)`` across all
```

docs/src/usage/distributed.rst
```diff
@@ -0,0 +1,162 @@
...
+distributed processes. If simply launched with ``python``, however, only one
+process is launched and no distributed communication takes place.
```
awni, line 33: Nit the double launched sounded a bit odd:

```suggestion
distributed processes. If simply run with ``python``, however, only one
process is launched and no distributed communication takes place.
```

docs/src/usage/distributed.rst
```diff
@@ -0,0 +1,162 @@
...
+
+MLX already comes with the ability to "talk" to MPI if it is installed on the
+machine. The minimal distributed program in MLX is as simple as:
+
+.. code:: python
+
+   import mlx.core as mx
+
+   x = mx.distributed.all_reduce_sum(mx.ones(10))
+   print(x)
```
awni, line 29: What do you think about adding the rank in the print? That will make it more clear what's going on in the output and add a nice feature to the minimal example

docs/src/usage/distributed.rst
```diff
@@ -0,0 +1,162 @@
...
+
+The above launches two processes on the same (local) machine and we can see
+both standard output streams. The processes send the array of 1s to each other
+and compute the sum which is printed. Launching with ``mpirun -np 4 ...`` would
+print 4 etc.
+
+Installing MPI
+---------------
+
+MPI can be installed with homebrew, using the anaconda package manager or
```
awni, line 53:
```suggestion
MPI can be installed with Homebrew, using the Anaconda package manager or
```

docs/src/usage/distributed.rst
```diff
@@ -0,0 +1,162 @@
...
+compiled from source. Most of our testing is done using ``openmpi`` installed
+with the anaconda package manager as follows:
```
awni, line 55:
```suggestion
with the Anaconda package manager as follows:
```

docs/src/python/distributed.rst
angeloskath: It actually isn't a random comment at all. The common approach is `all_reduce(x, op[, group])`. I kinda went with `all_reduce_<op>(x[, group])` cause it felt a bit better. Kind of in the spirit of having `sum`, `prod`, ... etc instead of `reduce(x, op)`.

docs/src/usage/distributed.rst
```diff
@@ -0,0 +1,162 @@
...
+
+.. code:: shell
+
+    $ conda install openmpi
+
+Installing with homebrew may require specifying the location of ``libmpi.dyld``
```
awni, line 61:
```suggestion
Installing with Homebrew may require specifying the location of ``libmpi.dyld``
```

docs/src/python/distributed.rst
angeloskath: Perhaps we should remove the reduce... not sure.

docs/src/usage/distributed.rst
```diff
@@ -0,0 +1,162 @@
...
+looks like the following:
+
+.. code::
+
+    host1 slots=1
+    host2 slots=1
+
+When using MLX, it is very likely that you want to use 1 slot per host, ie one
+process per host.  Moreover, the hostfile needs to also contain the current
+host if you wlso ant to run on the local host. Passing the host file to
```
awni, line 95:
```suggestion
process per host.  The hostfile also needs to contain the current
host if you want to run on the local host. Passing the host file to
```

docs/src/usage/distributed.rst
```diff
@@ -0,0 +1,162 @@
...
+Tuning All Reduce
+-----------------
+
+We are working on improving the performance of all reduce on MLX but for now
+the two main things one can do to extract the most out of distributed training with MLX are:
+
+1. Perform a few large reductions instead of many small ones to improve
+   bandwidth and latency
+2. Pass ``--mca btl_tcp_links 4`` to ``mpirun`` to configure it to use 4 tcp
+   connections between each host to improve bandwidth
```
awni, line 162: This requires 4 physical connections right?

docs/src/usage/distributed.rst
```diff
@@ -0,0 +1,162 @@
...
+-----------------------
+
+MPI can automatically connect to remote hosts and set up the communication over
+the network if the remote hosts can be accessed via ssh. A good checklist to
+debug connectivity issues is the following:
+
+* ``ssh hostname`` works from all machines to all machines without asking for
+  password or host confirmation
+* ``mpirun`` is accessible on all machines. You can call ``mpirun`` using its
+  full path to force a specific prefix on all machines.
```
awni, line 79: >  force a specific prefix on all machines.

I think what that means is to use a mpi at a certain location e.g. `/my/installed/mpirun`?

docs/src/usage/distributed.rst
```diff
@@ -0,0 +1,162 @@
...
+  a ``hostname`` defined as ``foo.bar.com`` if the ``hostname`` of the current
+  machine is ``baz.bar.com``.
+
+An easy way to pass the host names to MPI is using a host file. A host file
+looks like the following:
+
+.. code::
+
+    host1 slots=1
+    host2 slots=1
```
awni, line 91: I'm slightly confused by the setup description here. Is `host1` supposed to be e.g. `foo` and `host2` supposed to be `baz` in this example? 

Maybe that third bullet could be simpler or switched to be an example of how to specify the host file and use it with MPI?

docs/src/python/distributed.rst
awni: Yes the thought occurred to me as well, e.g. `all_sum`, `all_max`, ....

In a vacuum I think I prefer that as it's both fewer characters and a bit more consistent with the naming conventinos in core (e.g. `sum`, `max`, ... and not `reduce_sum`, `reduce_max`, ...). But I'm not sure if removing the reduce adds some confusion in this case?

awni: Yea you're right, we should keep the `all` in there.

awni: So I guess the question is should it be:

`distributed.all_sum` (for all-to-all) and `distributed.sum` for reducing to a single process

or keep it as is and include the `reduce` as a prefix?

angeloskath: I am down to change it to `all_sum`, it will also look better as I always use it as `mx.distributed.all_reduce_sum` never as a standalone function.

docs/src/usage/distributed.rst
angeloskath: This opens four sockets between each host pair. The reason is that the driver doesn't manage to saturate the bandwidth with a single socket so it is better to have many. 2 gets 90% of the way and above 4 didn't improve further in my testing.

docs/src/usage/distributed.rst
angeloskath: Yeah exactly. I changed it to "You can call ``mpirun`` using its full path to force all machines to use that path."

docs/src/usage/distributed.rst
angeloskath: That particular peculiarity cost me at least a day in debugging.

The host name in the host file can be whatever, ideally its FQDN. However, when the hosts share a part of the domain then MPI is not using the FQDN but just the distinct subset.

To make it concrete, if we have the following 2 hosts `foo.bar.com` and `baz.bar.com`, MPI will (may) not try to ssh to `foo.bar.com` but instead simply to `foo`. The problem that arises is that we may not have configured ssh to `foo` but only to `foo.bar.com`.

I removed it from the bullet and made it a note and added a few more clarifications on the hostfile example.

docs/src/python/distributed.rst
awni, line 18: That one needs to be changed as well.

docs/src/usage/distributed.rst
```diff
@@ -0,0 +1,167 @@
...
+        mx.eval(loss, model.parameters())
+
+All we have to do to average the gradients across machines is perform an
+:func:`all_reduce_sum` and divide by the size of the :class:`Group`. Namely we
+have to :func:`mlx.utils.tree_map` the gradients with following function.
+
+.. code:: python
+
+    def all_reduce_avg(x):
+        return mx.distributed.all_reduce_sum(x) / mx.distributed.init().size()
```
awni, line 135: Should that be the same function as `all_reduce_grads` below?

docs/src/usage/distributed.rst
```diff
@@ -0,0 +1,167 @@
...
+  password or host confirmation
+* ``mpirun`` is accessible on all machines. You can call ``mpirun`` using its
+  full path to force all machines to use a specific path.
+* Ensure that the ``hostname`` used by MPI is the one that you have configured
+  in the ``.ssh/config`` files on all machines.
+
+.. note::
+  For an example hostname ``foo.bar.com`` MPI could be using only ``foo`` as
+  the hostname passed to ssh, especially if the current host's name is also
+  ``*.bar.com``.
```
awni, line 87: Sorry this is a nit: the word "especially" makes it sound like there are other cases. But presumably it only works if the current hostname has the same domain suffix?

Maybe rephrase like:
```suggestion
  For an example hostname ``foo.bar.com`` MPI can use only ``foo`` as
  the hostname passed to ssh if the current hostname matches ``*.bar.com``.
```

awni (approved):
This is awesome! I left a few more nits in the docs, but o/w g2g! 🚀

docs/src/usage/distributed.rst
angeloskath: It isn't really a nit, you are correct. I put it to mean that there may be other cases. I couldn't find this anywhere in some docs or specs I just inferred it. Basically MPI was using just ``foo`` and couldn't connect to the remote hosts.

I think just using "can use only" or "may use only" is cleaner and conveys the same meaning so I 'll use it verbatim :-)

PR #1190, opened 2024-06-06
Fix jit scan when output doesn't have primitive

awni:
I think in general we might want to move away from relying on the output to have the correct primitive as it makes it hard to use the JIT functions inside other primitives. This one is just for scan but will probably update the others as needed.


angeloskath (approved):
Nice! 🙏

PR #1194, opened 2024-06-10
minor fixes

csukuangfj:
## Proposed changes

1. Use `std::lock_guard` to replace `std::unique_lock` when ownership of the mutex does not need to be transferred.
2. Remove trailing `;`

## Checklist


mlx/event.h
```diff
@@ -10,46 +10,46 @@ namespace mlx::core {
...
     return event_;
-  };
+  }
 
  private:
   // Default constructed stream should never be used
   // since the event is not yet valid
   Stream stream_{0, Device::cpu};
-  std::shared_ptr<void> event_{nullptr};
+  std::shared_ptr<void> event_;
```
awni, line 52: Why remove that?

csukuangfj: shared_ptr is default initialized to nullptr, right?

awni: It looks like it, yes!

awni:
@csukuangfj can you check the tests, looks like the code did not compile.

csukuangfj:
> @csukuangfj can you check the tests, looks like the code did not compile.

Sorry, fixed.

awni (approved):
LGTM, thanks!

Issue #909, opened 2024-03-26
Implement VMAP for Scatter


awni:
As title.

amirhossein-razlighi:
I am willing to look into that and start working on it. Can you give a littlebit more info on what is needed and where to get started please?

pharringtonp19:
@awni Is this something I could help implement?

awni:
Yes.. @amirhossein-razlighi let us know if you are working on it already.

amirhossein-razlighi:
No I have not started it yet.

pharringtonp19:
@awni I can dive into this at the end of the week. What's the background needed to develop for this library? Python & C++?

PR #1200, opened 2024-06-10
Add vmap to scatter

nicolov:
## Proposed changes

Add `vmap` support to `scatter`.

Closes https://github.com/ml-explore/mlx/issues/909

## Checklist

### DIFF
--- a/mlx/primitives.cpp
+++ b/mlx/primitives.cpp
@@ -3026,9 +3026,15 @@ std::pair<std::vector<array>, std::vector<int>> Scatter::vmap(
     // Clone updates along the vmap dimension so they can be applied to each
     // source tensor in the vmap.
     auto& updates = inputs.back();
-    updates =
-        expand_dims(updates, {0, static_cast<int>(inputs[1].ndim())}, stream());
-    updates = repeat(updates, vmap_size, 0, stream());
+    if (vmap_axes.back() < 0) {
+      updates = expand_dims(
+          updates, {0, static_cast<int>(inputs[1].ndim())}, stream());
+      updates = repeat(updates, vmap_size, 0, stream());
+    } else {
+      updates =
+          expand_dims(updates, static_cast<int>(inputs[1].ndim()), stream());
+      updates = moveaxis(updates, vmap_axes.back(), 0, stream());
+    }
   }
 
   auto out = array(
--- a/mlx/primitives.cpp
+++ b/mlx/primitives.cpp
@@ -3037,9 +3037,11 @@ std::pair<std::vector<array>, std::vector<int>> Scatter::vmap(
     }
   }
 
+  auto& shape = inputs[0].shape();
+  auto dtype = inputs[0].dtype();
   auto out = array(
-      inputs[0].shape(),
-      inputs[0].dtype(),
+      shape,
+      dtype,
       std::make_shared<Scatter>(stream(), reduce_type_, scatter_axes),
       std::move(inputs));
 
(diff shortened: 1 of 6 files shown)


mlx/primitives.cpp
```diff
@@ -2926,6 +2927,68 @@ std::vector<array> Scatter::jvp(
...
+std::pair<std::vector<array>, std::vector<int>> Scatter::vmap(
+    const std::vector<array>& inputs,
+    const std::vector<int>& vmap_axes) {
+  assert(inputs.size() >= 2);
+  assert(inputs.size() == vmap_axes.size());
+
+  for (size_t i = 1; i < vmap_axes.size(); i++) {
+    if (vmap_axes[i] != -1) {
+      throw std::runtime_error(
+          "[scatter]: vmap is only supported on the src input.");
```
awni, line 2939: I'm curious about this restriction? Why not allow the update or the indices to be vmapped?

mlx/primitives.cpp
```diff
@@ -2926,6 +2927,68 @@ std::vector<array> Scatter::jvp(
...
+  const auto& src = inputs.front();
+  std::vector<array> indices(inputs.begin() + 1, inputs.end() - 1);
+  const auto& updates = inputs.back();
+
+  auto src_ax = vmap_axes.front();
+  auto upd_ax = vmap_axes.back();
+  auto scatter_axes = axes_;
+
+  auto out_ax = *std::find_if(
+      vmap_axes.begin(), vmap_axes.end(), [](int a) { return a >= 0; });
```
awni, line 2952: Does this work if all the axes are `-1`?

awni:
@nicolov are you planning to come back to this?

mlx/primitives.cpp
nicolov: For `updates`: If we take one axis of `updates` as vmap dimension, it means scattering the remaining tensor and overwriting the same locations of `src` multiple times (as many as `updates.shape[updates_ax]`). It didn't seem very useful to me, plus we would have to define the semantics (the "last" update wins"?).

`indices` doesn't have tensor semantics, and it's also matched to `axes` (which can't have a vmap dimension because it's just a `vector<int>`. I struggled to imagine how to make the resulting operation clear. I guess we could slice off the `vmap` dimension off `indices`, then check that `axes` covers the remaining `indices`.

mlx/primitives.cpp
nicolov: How could this happen? I added a test case with the no-op vmap (all `in_axes` equal to `-1`) and then `Primitive::vmap` is not called at all. An empty `in_axes` defaults to `{0}` so I can't see how `vmap_axes` full of `-1` could happen. I added an assertion for good measure.

mlx/primitives.cpp
```diff
@@ -2926,6 +2927,68 @@ std::vector<array> Scatter::jvp(
...
+    return {
+        {scatter(src, indices, stacked_updates, scatter_axes, stream())},
+        // The updated scatter leaves the vmapped axis in the same place, so we
+        // can use src_ax here.
+        {src_ax},
+    };
+  }
+
+  // No vmapping needed.
+  return {{scatter(src, indices, updates, axes_, stream())}, {-1, -1, -1}};
```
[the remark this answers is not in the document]
nicolov: Yes, sorry, copy/paste mistake. If I understood correctly, it should return a single `-1` matching the single output that doesn't need any axis to be swapped back, right?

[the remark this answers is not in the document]
awni: Yes exactly.

awni (approved):
This is good to go! Thanks for the addition!

PR #1204, opened 2024-06-11
Add Quantized Ops to the JIT

barronalex:
Also add some utilities that reduce the boilerplate when defining Metal kernels (with and without JIT).

Instead of rewriting the whole function template again when we instantiate it, we use:

``` template [[host_name(name)]] [[kernel]] decltype(func<a, b, c>) func<a, b, c>;```

This makes the preprocessor instantiations in the `.metal` files more succinct.

Also when we're adding a JIT for an op, we can now use a generic function to get the template instantiation:

```get_template_definition(name, func, ...template_args)```

This removes the need to make instantiation strings in `metal/jit/<op>.h`.

Let me know if this obfuscates things too much.
I converted the FFTs to this pattern as a test run and I think it came out a bit cleaner.

## Benchmarks (M1 Max)
I ran a quantized LLM with the JIT on and off and it seemed like the initial prompt time was very similar between the two (note: there was quite a lot of variance between the runs).
```
JIT
Prompt: 138.470 tokens-per-sec
Generation: 55.092 tokens-per-sec
```
```
No JIT
Prompt: 136.970 tokens-per-sec
Generation: 55.145 tokens-per-sec
```
I also manually measured the run time of fetching the quantized kernels on the cold start and it was around 0.2 milliseconds which is probably pretty acceptable as a one time cost?

### CODE REVIEW

File: mlx/backend/metal/kernels.h [hunk 1]
```diff
@@ -159,11 +161,34 @@ MTL::ComputePipelineState* get_fft_kernel(
...
+get_template_definition(std::string name, std::string func, Args... args) {
+  std::ostringstream s;
+  s << func << "<";
+  bool first = true;
+  auto add_arg = [&s, &first](const auto& arg) {
+    if (!first) {
+      s << ", ";
+    }
+    first = false;
+    s << fmt::format("{}", arg);
```
awni, line 184: Does that need a format string or can you just do `s << arg`?

mlx/backend/metal/nojit_kernels.cpp
```diff
@@ -195,13 +195,16 @@ MTL::ComputePipelineState* get_fft_kernel(
...
-    const metal::MTLFCList& func_consts) {
+    const metal::MTLFCList& func_consts,
+    const std::string& template_def) {
   return d.get_kernel(kernel_name, "mlx", hash_name, func_consts);
 }
 
+MTL::ComputePipelineState* get_quantized_kernel(
+    metal::Device& d,
+    const std::string& kernel_name,
+    const std::string& template_def) {
```
awni, line 206: nit: delete the name so we don't get unused warnings.

mlx/backend/metal/nojit_kernels.cpp
```diff
@@ -195,13 +195,16 @@ MTL::ComputePipelineState* get_fft_kernel(
...
     const std::string& kernel_name,
     const std::string& hash_name,
-    const int tg_mem_size,
-    const std::string& in_type,
-    const std::string& out_type,
-    int step,
-    bool real,
-    const metal::MTLFCList& func_consts) {
+    const metal::MTLFCList& func_consts,
+    const std::string& template_def) {
```
awni, line 199: nit: delete the name so we don't get unused warnings.

mlx/backend/metal/kernels/quantized.h
```diff
@@ -0,0 +1,1461 @@
...
+
+#include <metal_simdgroup>
+#include <metal_stdlib>
+
+#include "mlx/backend/metal/kernels/bf16.h"
+#include "mlx/backend/metal/kernels/defines.h"
+#include "mlx/backend/metal/kernels/utils.h"
+
+#include "mlx/backend/metal/kernels/steel/gemm/gemm.h"
+
```
awni, line 11: So the way I've been trying to do these includes is to put them in the `.metal` file instead. For the no JIT case it should work as long as they are above the `quantized.h` include. Then for the JIT case what you do is use the JIT include strings when building the source string to compile. These should all be available already. That way we can avoid duplicating all that source code for the JIT case.

awni:
The way you do the instantiations and JIT is a really nice improvement! I don't think its too obfuscated.

mlx/backend/metal/kernels/quantized.h
barronalex: Makes total sense -- I've updated it

mlx/backend/metal/kernels/quantized.metal
```diff
@@ -1,1516 +1,50 @@
 // Copyright © 2023-2024 Apple Inc.
 
-#include <metal_simdgroup>
-#include <metal_stdlib>
-
 #include "mlx/backend/metal/kernels/bf16.h"
 #include "mlx/backend/metal/kernels/defines.h"
+#include "mlx/backend/metal/kernels/quantized.h"
```
awni, line 5: Tests are failing probably because these are in the wrong order. You can put a clang format off at the top of this file to keep the pre-commit hooks from changing them. I would just turn it off at the top and turn it back on at the bottom (the clang formatting is usually not so useful for these preprocessor files anyway and often messes it up)

PR #1205, opened 2024-06-12
Fix kernel deps to reduce build times

awni:
Refactored the metal/kernels/CmakeLists.txt so that kernels (for the most part) only depend on headers which they actually use. This way if you change e.g. `binary.h` you don't have to recompile every kernel.


jagrit06 (approved):
Looks good, tests clear on my system

PR #1206, opened 2024-06-12
Refactor JIT for unary/binary/ternary ops

barronalex:
Refactor the JIT for unary/binary/ternary ops so that we don't use the output array primitive to get the op type. Also switch them over to the new pattern introduced in #1204.

This fixes a bug for multi-upload bluestein FFTs when the JIT is enabled.


mlx/primitives.h
```diff
@@ -119,6 +119,13 @@ class Primitive {
   Primitive& operator=(const Primitive& other) = delete;
   Primitive& operator=(Primitive&& other) = delete;
 
+  /** Return the Primitive name as a string */
+  inline std::string op_name() {
+    std::ostringstream op_t;
+    print(op_t);
+    return op_t.str();
+  };
```
awni, line 127: If we make this a method, I would change the name to be a bit more general as its basically a wrapper around print and in many cases includes more than just the op name.

- One option is to make it an overload of `print` which returns a string and takes no input.
- Another option is to give it a name like `to_string` or `repr` or something. 
- A third option (which is a bit safer since it doesn't change the API) is to not have this as a method and just make it a utility in the metal back-end which takes the primitive and returns the string representation

barronalex: I had option 3 originally so I'm happy to go back to that and keep the API stable. If we find a need for this in the CPU implementation then maybe we switch it into the class.

awni (approved):
🚀

PR #1209, opened 2024-06-13
Only allow building with OpenMPI

awni:
Sometimes people have MPICH (or another MPI installed) and the build breaks. So this emits an explicit warning about it and builds without MPI.

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -1,7 +1,7 @@
 FetchContent_Declare(
   doctest
   GIT_REPOSITORY "https://github.com/onqtam/doctest"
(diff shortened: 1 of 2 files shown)


tests/CMakeLists.txt
```diff
@@ -1,7 +1,7 @@
 FetchContent_Declare(
   doctest
   GIT_REPOSITORY "https://github.com/onqtam/doctest"
```
barronalex, line 3: This just upgrades to the latest version?

barronalex (approved):
LGTM

tests/CMakeLists.txt
awni: Yea sorry, I kind of snuck that in. I was hoping it would resolve a cmake warning but it seems thats not in the latest doctest release yet.

PR #1216, opened 2024-06-18
Add GPU Cholesky using MPS

nicolov:
## Proposed changes

Add a GPU implementation of Cholesky decomposition based on the MPS kernel. Completes https://github.com/ml-explore/mlx/issues/1026.

Sadly the kernel is slower than the CPU, maybe it's still useful (times in ms):

```
size       cpu     gpu
1000      0.99    7.67
5000     24.44   61.94
10000    79.13  162.09
```

Many of the bindings in `mlx/backend/metal/mps/gemm.h` (eg `Matrix`) could be moved to a standalone header. I can do that in a follow-up if wanted.

## Checklist

### DIFF
--- a/mlx/backend/metal/cholesky.cpp
+++ b/mlx/backend/metal/cholesky.cpp
@@ -1,0 +1,91 @@
+// Copyright © 2024 Apple Inc.
+
+#include "mlx/backend/metal/mps/cholesky.h"
+#include "mlx/backend/metal/copy.h"
+#include "mlx/backend/metal/device.h"
+#include "mlx/backend/metal/kernels.h"
+#include "mlx/backend/metal/kernels/defines.h"
+#include "mlx/backend/metal/kernels/steel/gemm/params.h"
+#include "mlx/backend/metal/matmul.h"
+#include "mlx/backend/metal/mps/gemm.h"
+#include "mlx/backend/metal/utils.h"
+#include "mlx/primitives.h"
+#include "mlx/utils.h"
+
+namespace mlx::core {
+
+void Cholesky::eval_gpu(const std::vector<array>& inputs, array& out) {
+  auto& s = stream();
+  auto& d = metal::device(s.device);
+
+  const auto& a = inputs.at(0);
+  if (a.dtype() != float32) {
+    throw std::runtime_error("[Cholesky::eval_gpu] only supports float32.");
+  }
+
+  out.set_data(allocator::malloc_or_wait(out.nbytes()));
+
+  MPS::DataType mps_dtype = MPS::DataTypeFloat32;
+
+  const int num_rows = a.shape(-2);
+  const int row_nbytes = a.shape(-1) * sizeof(float);
+  const int num_matrices = a.size() / (num_rows * num_rows);
+
+  const auto mat_desc = MPS::MatrixDescriptor::matrixDescriptor(
+      /* rows = */ num_rows,
+      /* columns = */ num_rows,
+      /* rowBytes = */ row_nbytes,
+      /* dataType = */ mps_dtype);
+
+  d.end_encoding(s.index);
+
+  // Ensure that the input is contiguous.
+  auto input = a;
+  if (!a.flags().contiguous) {
+    copy(
+        a,
+        input,
+        a.flags().row_contiguous ? CopyType::Vector : CopyType::General);
+  }
+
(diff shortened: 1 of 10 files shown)


mlx/backend/metal/cholesky.cpp
```diff
@@ -0,0 +1,84 @@
...
+
+namespace mlx::core {
+
+void Cholesky::eval_gpu(const std::vector<array>& inputs, array& out) {
+  auto& s = stream();
+  auto& d = metal::device(s.device);
+
+  const auto& a = inputs.at(0);
+  if (a.dtype() != float32) {
+    throw std::runtime_error("Cholesky::eval_gpu only supports float32.");
```
awni, line 23: Nit:
```suggestion
    throw std::runtime_error("[Cholesky::eval_gpu] only supports float32.");
```

mlx/backend/metal/cholesky.cpp
```diff
@@ -0,0 +1,84 @@
...
+        d.mtl_device(), !upper_, num_rows);
+
+    auto command_buffer = d.get_command_buffer(s.index);
+    kernel->encodeToCommandBuffer(command_buffer, a_mat, out_mat, nullptr);
+    command_buffer->addCompletedHandler(
+        [a_mat, out_mat, kernel](MTL::CommandBuffer*) mutable {
+          a_mat->release();
+          out_mat->release();
+          kernel->release();
+        });
```
awni, line 61: I don't know the answer to this.. but it's not clear to me that we need to keep these data structures around until the kernel is complete. It might be the case that its safe to release them immediately (as long as the underlying memory is retained which it should be)?

mlx/backend/metal/cholesky.cpp
```diff
@@ -0,0 +1,84 @@
...
+      /* rowBytes = */ row_nbytes,
+      /* dataType = */ mps_dtype);
+
+  d.end_encoding(s.index);
+
+  // The kernel does not support batching, so we have to loop explicitly.
+  for (int i = 0; i < num_matrices; i++) {
+    size_t offset = i * num_rows * row_nbytes;
+    auto a_buf = static_cast<const MTL::Buffer*>(a.buffer().ptr());
+    auto a_mat = MPS::Matrix::alloc()->init(a_buf, offset, mat_desc);
```
awni, line 46: I think you are assuming the input is contiguous here? Mind adding test cases for that + a fix?

mlx/backend/metal/cholesky.cpp
```diff
@@ -0,0 +1,84 @@
...
+    kernel->encodeToCommandBuffer(command_buffer, a_mat, out_mat, nullptr);
+    command_buffer->addCompletedHandler(
+        [a_mat, out_mat, kernel](MTL::CommandBuffer*) mutable {
+          a_mat->release();
+          out_mat->release();
+          kernel->release();
+        });
+  }
+
+  // Zero out the unwanted part of the output.
```
awni, line 64: That's interesting that it's needed. Is it because the memory is uninitialized or does the MPS kernel actually write some stuff to the 0 triangle? If it's the former, probably simpler to just zero the memory initially.

awni:
Thanks for the addition, that's awesome. I left a few comments. Also any idea how fast it is?

mlx/backend/metal/cholesky.cpp
nicolov: I had just copy-pasted from other ops, but I applied your suggestion and it seems to still work.

mlx/backend/metal/cholesky.cpp
nicolov: It's used as scratch space.

mlx/backend/metal/cholesky.cpp
nicolov: I've added a branch to make a copy, is that the nicest way to do it?

nicolov:
Addressed your comments and added a benchmark. Sadly it's not very fast.. (on M1 Max at least). Do you still think it's worth to merge it?

awni:
Wow.. that's much slower... the kernel does not support batching (as you said in your comment). I an not sure it's worth merging. The default behavior would be to use it and it's 7x slower for 1k matrices..

nicolov:
Yes, I agree.

PR #1221, opened 2024-06-22
Fix NumPy 2.0 pickle test

barronalex:
NumPy updated to 2.0 in CircleCI which broke one of our tests.

`np.array(x, copy=False)` now errors instead of silently copying the array if the no-copy path is not possible.
This means we get a different error when pickling a `bfloat16` array so our `with self.assertRaises(TypeError)` doesn't trigger.

We fix by explicitly raising the type error in the `bfloat16` template.


awni (approved):
LGTM!

Issue #1223, opened 2024-06-22
[BUG] `mx.inner` treat scalar array and array with one element differently


minghuaw:
**Describe the bug**

`mx.inner` has no problem broadcasting the scalar array to the shape of the other argument but fails when the array contains only one element.

**To Reproduce**

Include code snippet
```python
import mlx.core as mx

x = mx.eye(2)
y1 = mx.array(7)
y2 = mx.array([7])

# This works fine as expected
mx.inner(x, y1)

# This would throw an exception "[inner] a and b must have the same last dimension."
mx.inner(x, y2)
```

**Expected behavior**

`mx.inner(x, y2)` is expected to yield the same result as `mx.inner(x, y1)`

**Desktop (please complete the following information):**
 - OS Version: MacOS 14.5
 - Version: mlx 0.15

**Additional context**

The same problem is observed in the c binding, so the problem is probably rooted in the c++ code

minghuaw:
Closing because the same behavior is observed in numpy and the one-element array should probably be treated as a singleton matrix instead of a scalar

PR #1224, opened 2024-06-22
Fix `inner()` throws when one argument is an one element array

minghuaw:
## Proposed changes

This PR addresses the issue #1223 by checking if `a` or `b` is a one-element array by adding the following checks to the if statement `a.shape() == std::vector({1}) || b.shape() == std::vector({1})` (Is there a better way for comparing vectors in cpp?). A new test case is also added to the end of `ops_tests.cpp/TEST_CASE("inner")`.

## Checklist


minghuaw:
Closing because the same behavior is observed in numpy and the one-element array should probably be treated as a singleton matrix instead of a scalar

Issue #1217, opened 2024-06-18
[BUG] iOS builds fail with float16 issue


davidkoski:
**Describe the bug**

**To Reproduce**

Build mlx for iOS (I am using Xcode 15.3) and you will run into something like this:

```
/Users/dkoski/Developer/mlx-swift/Source/Cmlx/mlx-conditional/../mlx/mlx/backend/accelerate/softmax.cpp:61:22: error: non-constant-expression cannot be narrowed from type 'float' to 'float16_t' (aka '__fp16') in initializer list [-Wc++11-narrowing]
    x = vmulq_f16(x, vdupq_n_f16(1.442695)); // multiply with log_2(e)
                     ^~~~~~~~~~~~~~~~~~~~~
In file included from /Users/dkoski/Developer/mlx-swift/Source/Cmlx/mlx-conditional/accelerate-softmax.cpp:7:
In file included from /Users/dkoski/Developer/mlx-swift/Source/Cmlx/mlx-conditional/../mlx/mlx/backend/accelerate/softmax.cpp:6:
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/lib/clang/15.0.0/include/arm_neon.h:6501:26: note: expanded from macro 'vdupq_n_f16'
  __ret = (float16x8_t) {__s0, __s0, __s0, __s0, __s0, __s0, __s0, __s0}; 
```

This is on lines like this:

```
    x = vmulq_f16(x, vdupq_n_f16(1.442695F16)); // multiply with log_2(e)
```

The macro expands like:

```
x = vmulq_f16(x, __extension__ ({
    float16x8_t __ret;
    float16_t __s0 = 1.442695;
    __ret = (float16x8_t) {__s0, __s0, __s0, __s0, __s0, __s0, __s0, __s0};
    __ret;
}));
```

which looks ok, but the problem is that `float16_t` isn't the type we are looking for:

```
#define ADD_HALF_BINOPS
#include "mlx/types/fp16.h"
namespace mlx::core {
typedef struct _MLX_Float16 float16_t;
} // namespace mlx::core
```

You can see this by changing the macro to use `::float16_t` or add this to the top of softmax.cpp:

```
#define __ARM_FEATURE_FP16_SCALAR_ARITHMETIC 1
```

I am not sure all the intricacies here, but this is interfering with the float16 code in softmax.cpp on iOS.

Currently blocking https://github.com/ml-explore/mlx-swift/pull/101

[...]

PR #1227, opened 2024-06-24
Accelerate import updates for iOS

jagrit06:
## Proposed changes

Fixes issues from #1217

## Checklist

### DIFF
--- a/mlx/backend/accelerate/softmax.cpp
+++ b/mlx/backend/accelerate/softmax.cpp
@@ -376,12 +376,7 @@ void Softmax::eval_cpu(const std::vector<array>& inputs, array& out) {
             NeonFp16SimdOps<float16_t, float16x8_t>,
             8>(in, out);
 #else // __ARM_FEATURE_FP16_VECTOR_ARITHMETIC
-        softmax<
-            float16_t,
-            float,
-            simd_float16,
-            AccelerateSimdOps<float, simd_float16>,
-            16>(in, out);
+        eval(inputs, out); // Redirect to common backend for consistency
 #endif // __ARM_FEATURE_FP16_VECTOR_ARITHMETIC
       }
       break;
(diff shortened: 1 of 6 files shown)


awni (approved):
Looks great, thanks for fixing that! @davidkoski can you check that it compiles now with your MLX Swift branch?

davidkoski:
Yes, will do

davidkoski:
I still get the error -- it looks like the `float16_t` is still resolving to `_MLX_Float16`

awni (changes requested):
Could you update this to use https://github.com/ml-explore/mlx/blob/main/mlx/backend/common/softmax.cpp#L110 so that we have the same numerical behavior across platforms?

mlx/backend/accelerate/softmax.cpp
```diff
@@ -53,25 +56,26 @@ inline simd_float16 simd_fast_exp(simd_float16 x) {
...
 /**
  * The ARM neon equivalent of the fast exp above.
  */
 inline float16x8_t neon_fast_exp(float16x8_t x) {
-  x = vmulq_f16(x, vdupq_n_f16(1.442695)); // multiply with log_2(e)
-  x = vmaxq_f16(x, vdupq_n_f16(-14)); // clamp under with -14
-  x = vminq_f16(x, vdupq_n_f16(14)); // clamp over with 14
+  x = vmulq_f16(x, vdupq_n_f16(float16_t(1.442695f))); // multiply with log_2(e)
+  x = vmaxq_f16(x, vdupq_n_f16(float16_t(-14.f))); // clamp under with -14
+  x = vminq_f16(x, vdupq_n_f16(float16_t(14.f))); // clamp over with 14
```
davidkoski, line 66: Maybe these changes should be reverted?  They don't fix it and maybe just make it less clear?

jagrit06: So they do fix things on some versions of iOS, so I will leave them there to be safe

awni (approved):
Thanks for the fix!!

PR #1235, opened 2024-06-25
Fixes segfault when compiling checkpointed functions

### LINKED ISSUE #1162: [BUG] compile + checkpoint segfaults
```python
import mlx.core as mx
import mlx.nn as nn
from mlx.nn.utils import checkpoint

dims = 32

class Linear(nn.Module):
    def __init__(self, dims):
        super().__init__()
        self.proj = nn.Linear(dims, dims)

    def __call__(self, x):
        l = checkpoint(self.proj) 
        x = l(x)
        return x

model = Linear(dims)
mx.eval(model.parameters())

@mx.compile
def step(inputs):
    return model(inputs)

inputs = mx.ones((2, dims), dtype=mx.float32)
mx.eval(inputs)

out = step(inputs)
mx.eval(out)
```

Above segfaults with checkpointing but not without.

angeloskath:
Closes #1162.

See title. It exposes a `clear_compiler_cache` method in C++ and Python which is also registered `atexit`.

### DIFF
--- a/mlx/compile.cpp
+++ b/mlx/compile.cpp
@@ -266,8 +266,8 @@ class CompilerCache {
     cache_.erase(fun_id);
   }
 
-  std::unordered_map<std::uintptr_t, std::vector<CacheEntry>>& entries() {
-    return cache_;
+  void clear() {
+    cache_.clear();
   }
 
  private:
@@ -863,6 +863,10 @@ void compile_erase(std::uintptr_t fun_id) {
   detail::compiler_cache().erase(fun_id);
 }
 
+void compile_clear_cache() {
+  detail::compiler_cache().clear();
+}
+
 } // namespace detail
 
 std::function<std::vector<array>(const std::vector<array>&)> compile(
@@ -885,10 +889,6 @@ void enable_compile() {
 
 void set_compile_mode(CompileMode mode) {
   detail::compile_mode() = mode;
-}
-
-void clear_compiler_cache() {
-  detail::compiler_cache().entries().clear();
 }
 
 } // namespace mlx::core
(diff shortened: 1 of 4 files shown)


python/src/transforms.cpp
```diff
@@ -961,12 +960,22 @@ void init_transforms(nb::module_& m) {
         Globally enable compilation. This will override the environment
         variable ``MLX_DISABLE_COMPILE`` if set.
       )pbdoc");
+  m.def(
+      "clear_compiler_cache",
+      &clear_compiler_cache,
+      R"pbdoc(
+        Clear the compiler cache causing all compiled function to recompile if
+        they are called again.
+      )pbdoc");
```
awni, line 969: Do we need a binding for that / what is the intended use for it?

mlx/compile.cpp
```diff
@@ -266,6 +266,10 @@ class CompilerCache {
     cache_.erase(fun_id);
   }
 
+  std::unordered_map<std::uintptr_t, std::vector<CacheEntry>>& entries() {
+    return cache_;
+  }
```
awni, line 271: Do you think it makes sense to make this a `clear` function instead?

awni (approved):
Awesome! Thanks for the fix! Left a couple minor comments re the API. But I'm good to merge this as is.

python/src/transforms.cpp
angeloskath: I 'll remove it. I added it for testing and then thought that maybe someone wanted to clean things up for long running programs.

mlx/compile.cpp
angeloskath: Same as above initially added because I had more functionality for testing and didn't limit enough the API in detail. Will change it to clear.

angeloskath:
@awni let me know if this is better. I put the clear in `detail::` and changed the name to `detail::compile_clear_cache()`.

awni:
Looks perfect!

PR #1236, opened 2024-06-26
Fix strided sort bug

### LINKED ISSUE #1230: `sorted` and `argSort` don't work for split arrays
When I sort an 'ordinary' array it works just fine. 
If, however, I pass an array that was produced as a result of `split`, both `sorted` and `argSort` output garbage.

Here is the code that reproduces the issue (tried M1, A15, both CPU and GPU):
```swift
let arr = randInt((Int32(0)..<10), [5,2])

let s = split(arr, axis: -1)
let s0 = MLXArray(s.0.asArray(Int32.self), s.0.shape)

print(s.0, s0)   // these two are identical

let idSplit =  argSort(s.0, axis: 0)
let idCopied = argSort(s0, axis: 0)

print(idSplit)  // this one is messed up
print(idCopied) // this one is correct
```

barronalex:
Resolves #1230

Turns out we weren't using the output strides in the CPU and single block GPU implementations.


awni:
Thanks for the quick fix.

> Turns out we weren't using the output strides in the CPU and single block GPU implementations.

Isn't the output always contiguous? Is the problem that we were using the input strides to index the output?

awni:
I'm wondering if there is a bug in this line: https://github.com/ml-explore/mlx/pull/1236/files#diff-ce10f0dc97a4bca59973f06340d5a09ca4d4a3568cf79547f5cba1d902a4b585L109-L110

The input being `contiguous` does not imply that it is row or col contiguous. If the input is contiguous but not row/col contig then we do a vector copy. Then the output will also not be row/col contig. And is that safe? Does the sort work if the output is not contiguous along the sorted dimension?

awni:
Here's an example that seems to still be broken on the GPU (though it works on the CPU).

```python
a = mx.array([1, 0, 2, 1, 3, 0, 4, 0])
b = mx.broadcast_to(a, (16, 8))
mx.eval(b)
d = mx.sort(b, axis=1)
print(d)
```

barronalex:
Good catch on the zero stride case! Turns out there was a typo in the GPU implementation that I've fixed.

I think it's OK if the sorted axis stride is zero or greater than 1 in the CPU case (and it seems to pass the tests). We'll overwrite the same memory multiple times but that shouldn't be an issue.

awni:
Makes perfect sense , thanks for fixing this and improving this JIT path as well! Very nice!

awni (approved):

PR #1237, opened 2024-06-26
Fix donation condition for compilation

### LINKED ISSUE #1234: [BUG] NaN when using max reduction + compile
The following should not have NaNs but it does:

```python
import mlx
import mlx.core as mx

def fun(x):
    pool_size = 2
    stride = 2
    batch, height, width, channels = x.shape
    out_height = (height - pool_size) // stride + 1
    out_width = (width - pool_size) // stride + 1
    x = x.reshape(batch, out_height, stride, out_width, stride, channels)
    return mx.max(x, axis=(2, 4)).sum()

grad_fn = mx.value_and_grad(fun)

@mx.compile
def step(x):
    loss, grads = grad_fn(x)
    return grads

x_batch = mx.random.uniform(shape=(50, 28, 28, 64))

grads = step(x_batch)
print(mx.any(mx.isnan(grads)))
```

awni:
As title. Closes #1234

### DIFF
--- a/mlx/backend/common/compiled.cpp
+++ b/mlx/backend/common/compiled.cpp
@@ -205,8 +205,8 @@ void compiled_allocate_outputs(
       // - Donatable
       // - Correct size
       // - Not a constant
-      if (in.flags().row_contiguous && in.nbytes() == outputs[o].nbytes() &&
-          in.is_donatable() &&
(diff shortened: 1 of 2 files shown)


angeloskath (approved):
👍

PR #1242, opened 2024-06-28
CPU binary reduction + Nits

### LINKED ISSUE #1241: [BUG] Typo: traditinoal -> traditional
https://github.com/ml-explore/mlx/blob/d6383a1c6a2d663d6912bf20f3f57c727cd606d5/python/src/fast.cpp#L86

I use https://marketplace.visualstudio.com/items?itemName=streetsidesoftware.code-spell-checker to avoid issues like this, which sometimes aren't caught by regular code listing (such as when the reference is inside a string, which is the case here). Just a suggestion of course!

I believe this typo causes this parameter to have no effect: 

https://github.com/ml-explore/mlx/blob/d6383a1c6a2d663d6912bf20f3f57c727cd606d5/python/mlx/nn/layers/positional_encoding.py#L50

as the generated code (I think it's generated as I couldn't find it in the repo) for the rope function is:

```
def rope(a: array, dims: int, *, traditinoal: bool, base: float, scale: float, offset: int, stream: Union[None, Stream, Device] = None) -> array:
    """
    Apply rotary positional encoding to the input.

    Args:
        a (array): Input array.
        dims (int): The feature dimensions to be rotated. If the input feature
            is larger than dims then the rest is left unchanged.
        traditional (bool): If set to ``True`` choose the traditional
            implementation which rotates consecutive dimensions.
        base (float): The base used to compute angular frequency for
            each dimension in the positional encodings.
        scale (float): The scale used to scale the positions.
        offset (int): The position offset to start at.

    Returns:
        array: The output array.
    """
```

Note the typo in `traditinoal`.

awni:
- Shrink CPU binary a little by avoiding unused instantiations
 
- Closes #1241

### DIFF
--- a/mlx/backend/accelerate/primitives.cpp
+++ b/mlx/backend/accelerate/primitives.cpp
@@ -101,7 +101,7 @@ void Add::eval_cpu(const std::vector<array>& inputs, array& out) {
   auto& b = inputs[1];
 
   if (a.dtype() == float32) {
-    binary(
+    binary_op<float, float>(
         a,
         b,
         out,
@@ -116,7 +116,7 @@ void Add::eval_cpu(const std::vector<array>& inputs, array& out) {
           vDSP_vadd((const float*)a, 1, (const float*)b, 1, (float*)o, 1, n);
         });
   } else if (a.dtype() == int32) {
-    binary(
+    binary_op<int, int>(
         a,
         b,
         out,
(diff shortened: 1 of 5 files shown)


angeloskath (approved):
Awesome!

awni:
Saves about 0.5 MB on the minimal binary build. Nothing too extravagant.. but a free win.

PR #1245, opened 2024-07-01
Transposed Convolution

mlaves:
## Proposed changes

This implementation extends `conv_general` and introduces a bool flag `transpose` to make use of the existing code path.

* Includes tests and benchmarks.
* Only supports groups=1 now.
* Does not support output padding, yet.
* torch uses (in_channels, kernel_size, ..., out_channels) weight shape for conv_transpose, this implementation uses (out_channels, kernel_size, ..., in_channels). This is up to discussion.

## Checklist


mlx/primitives.h
```diff
@@ -677,6 +677,47 @@ class Convolution : public UnaryPrimitive {
   void eval(const std::vector<array>& inputs, array& out);
 };
 
+class ConvolutionTranspose : public UnaryPrimitive {
+ public:
+  explicit ConvolutionTranspose(
```
awni, line 682: What's the rationale for making a new primitve for this? Why not just use the Convolution primitive as is?

awni:
@mlaves this looks really nice!! I think the main question is what the purpose of the `ConvolutionTranspose` primitive is? I didn't look to carefully, but it looks like we should just use the `Convolution` primitive and send it the right arguments.

Other than that I'll review the rest of the API in some detail shortly but looks great so far.

mlx/primitives.h
mlaves: I did that to implement the different `vjp` method, but realized that leads to some code duplication. I can extend the `Convolution` primitive to avoid that.

awni:
@mlaves you are planning to come back to this? Would be great to land it.

mlaves:
> @mlaves you are planning to come back to this? Would be great to land it.

Sorry for the delay, I'll finish that in the next couple of days.

mlx/primitives.h
mlaves: Done.

mlaves:
@awni I added a `transpose` flag to the existing `Convolution` primitive and removed the initially introduced primitive that is now obsolete. I also rebased onto main.

benchmarks/python/conv_transpose_bench.py
```diff
@@ -0,0 +1,138 @@
...
+import math
+import os
+import subprocess
+import time
+
+import mlx.core as mx
+import numpy as np
+import torch
+
+device_name = subprocess.check_output(["sysctl", "-n", "machdep.cpu.brand_string"])
```
awni, line 11: unused?

docs/src/python/ops.rst
```diff
@@ -45,6 +45,10 @@ Operations
    conv1d
    conv2d
    conv3d
+   conv3d
```
awni, line 48: remove as its a duplicate

mlx/ops.cpp
```diff
@@ -3305,9 +3380,10 @@ array conv_general(
     std::vector<int> input_dilation /* = {} */,
     int groups /* = 1 */,
     bool flip /* = false */,
+    bool transpose /* = false */,
     StreamOrDevice s /* = {} */) {
   // Run checks
-  if (groups != 1 && in.ndim() != 3 && in.ndim() != 4) {
+  if (groups != 1 && in.ndim() != 3 && in.ndim() != 4 && in.ndim() != 4) {
```
awni, line 3386: duplicate condition

awni:
I remove the tranpose flag and attempted to resolve the incorrect gradient w.r.t. the `weight` when the `flip` argument is `True`. I don't think all the gradients are correct yet / our tests are not comprehensive. But this is a good improvement over where we were.

We should add more comprehensive tests for backward particularly with dilations / flip / groups.

mlaves:
> I remove the tranpose flag and attempted to resolve the incorrect gradient w.r.t. the `weight` when the `flip` argument is `True`. I don't think all the gradients are correct yet / our tests are not comprehensive. But this is a good improvement over where we were.
> 
> We should add more comprehensive tests for backward particularly with dilations / flip / groups. But we can leave that for a follow on PR (along with any potential fixes).

Cool, that looks like a nice simplification over my initial implementation. This PR looks good to me now. I can follow up with some more tests as mentioned.

awni (approved):
Thanks for the addition @mlaves !

Issue #1090, opened 2024-05-08
[Feature] Expose something like custom VJP in Python


awni:
Consider allowing a function to have a custom VJP function attached to it.

angeloskath:
Oops I thought I had exposed that. It might be worth having a high level API for that as well (which is probably why I didn't expose it), maybe implemented in python using the `custom_vjp` transform? Either of the following two feels more natural to use imho

Option a:
```python
import mlx.core as mx

@mx.custom_function
def my_addition(a, b):
    return a + b

@my_addition.vjp
def my_addition_vjp(primals, outputs, cotangents):
    a, b = primals
    cotan = cotangents[0]
    return cotan * b, cotan * a
```

Option b:
```python
import mlx.core as mx

@mx.custom_function
def my_addition(a, b):
    def my_vjp(primals, outputs, cotangents):
        a, b = primals
        cotan = cotangents[0]
        return cotan * b, cotan * a
    return a + b, my_vjp
```

I lean towards the first one. Obviously names and details need some changing.

kaeru-shigure:
I think the interface your propose is good, but if it becomes pytorch compatible, it might be useful when migrating.  
Anyway, I am eagerly waiting for this feature to be implemented.  
  
Here is an example of a [torch.autograd.Function](https://pytorch.org/docs/stable/autograd.html#torch.autograd.Function) compatible interface:
```py
import mlx.core as mx

class MxCustomVjp():
    @classmethod
    def apply(cls, *args, **kwargs):
        ctx = cls()
        ctx.save_for_backward(*args)
        r = ctx.forward(ctx, *args, **kwargs)
        return r if type(r) == tuple else (r,), ctx
    def save_for_backward(self, *args):
        self.saved_tensors = args
    def vjp(self, grad_output):
        return self.__class__.backward(self, *grad_output)
    @staticmethod
    def forward(self, *args, **kwargs):
        raise NotImplementedError
    @staticmethod
    def backward(self, *args, **kwargs):
        raise NotImplementedError

class CustomLinear(MxCustomVjp):
    @staticmethod
    def forward(ctx, x, weight):
        ctx.save_for_backward(x, weight)
        return x @ weight
    @staticmethod
    def backward(ctx, grad_output):
        x, weight = ctx.saved_tensors
        grad_x = grad_output @ weight.T
        grad_weight = x.T @ grad_output
        return grad_x, grad_weight

pair2x = lambda a, b: (a * 2, b * 2,)
[...]

PR #1246, opened 2024-07-02
Custom transforms

angeloskath:
Exposes `custom_vjp` in python and addresses the eval-in-vjp problem that will occur with custom_vjp usage.

This PR is in draft mode to discuss the solution for the above and if we like it and after the following list is checked it can probably be merged:

- [x] Ensure no perf regression from the added flag
- [x] Write tests and docs
- [ ] Maybe write a high-level grad inspection helper


angeloskath:
Since I realized I didn't add an example of how this looks I 'll add a couple below:

Extremely simple custom binary function:
```python
@mx.custom_vjp
def my_mul(a, b):
    return 2 * a * b

@my_mul.vjp
def my_mul(primals, cotangent, output):
    a, b = primals
    return cotangent * b * 2, cotangent * a * 2
```

The following is a generic straight through estimator (literally just passes the gradient along).
```python
def straight_through(f):
    cf = mx.custom_vjp(f)

    @cf.vjp
    def cf_vjp(primals, cotangents, outputs):
        return cotangents

    return cf

bernoulli = straight_through(mx.random.bernoulli)

# Example usage
print(mx.grad(lambda x: bernoulli(x).sum())(mx.array([0.1, 0.9])))  # prints [1, 1]
print(mx.grad(lambda x: mx.random.bernoulli(x).sum())(mx.array([0.1, 0.9])))  # prints [0, 0]
```

And the following is an example of gradient inspection
```python
def inspect_grad(cb, x):
    @mx.custom_vjp
    def identity(x):
        return x

    @identity.vjp
    def identity(x, dx, _):
        return cb(dx) or dx

    return identity(x)


def f(x):
    return 2 * x**2 + 2 * x + 1


def g(x):
    return f( 2 * inspect_grad(print, f(x)))**2

print(mx.grad(mx.grad(g))(mx.array(8)))
# array(784494088, dtype=int32)  1st order intermediate gradient
# array(550900096, dtype=int32)  2nd order intermediate gradient
# array(393743136, dtype=int32)  final total gradient
```

python/src/transforms.cpp
```diff
@@ -593,7 +593,122 @@ class PyCheckpointedFun {
...
+      nb::gil_scoped_acquire gil;
+
+      vjp_fun_.release().dec_ref();
+      input_structure_.release().dec_ref();
+      if (output_structure_.use_count() == 1) {
+        output_structure_->release().dec_ref();
+      }
+    }
+
+    std::vector<array> operator()(
```
awni, line 630: I'm slightly confused on the purpose of this `operator()` overload. Is it used for anything?

angeloskath: Yeah, this is the `InnerFunction` implementation. Basically an old school lambda to pass to `custom_vjp` (C++ side) as the vjp function. The reason for its existence (same as in `checkpoint`) is to control the destruction of the captured (in this case wrapped) python variables.

Basically this exists only for its destructor, otherwise it would be a simple lambda passed in line 686.

awni:
One thing I'm wondering about is the high level API for functions with customized transformations (related conversation in #1090).

If we envision a future of allowing other transforms to be customized.. it might make sense to design the API for that (even if the implementation is more-or-less the same).

For example should the API look more like:

```python
    @mx.custom
    def fun(x):
        return x

    @fun.vjp
    def fun_vjp(x, dx, _):
        return ...

    @fun.jvp
    def fun_jvp(x, dx):
        return ...
```

Or would it be simpler to keep them separate, in which case the current API probably doesn't need to change.

```python
    def fun(x):
        return x

    fun = mx.custom_vjp(fun)

    @fun.vjp
    def fun_vjp(x, dx, _):
        return ...

    fun = mx.custom_jvp(fun)

    @fun.jvp
    def fun_jvp(x, dx):
        return ...
```

angeloskath:
Yeah you are definitely right on that one. I don't like wrapping the wrapped function too much. It will work but it feels wrong.

I think I should just make the necessary changes in C++ as well so that we can do `vjp`, `jvp` and `vmap`. I guess I was being lazy implementing what we currently have and not what we want :-)

angeloskath:
Ok @awni, @jagrit06, @barronalex, I think this is ready for review. It now supports all transforms and has tests and a bit of docs. I think a high-level gradient inspection helper can go in another PR.

angeloskath:
Btw I ran both generation and training tests and no performance regression is observed by the extra flag.

awni:
This is very cool. I will try and review more carefully today. But overall it looks really great so far!

python/src/transforms.cpp
```diff
@@ -593,7 +593,451 @@ class PyCheckpointedFun {
...
+      "custom_function",
+      R"pbdoc(
+      Set up a function for custom gradient and vmap definitions.
+
+      This class is meant to be used as a function decorator. Instances are
+      callables that behave identically to the wrapped function. However, when
+      a function transformation is used (e.g. computing gradients using
+      :func:`value_and_grad`) then the functions defined via :method:`vjp`,
+      :method:`jvp` and :method:`vmap` are used instead of the default
+      transformation.
```
awni, line 939: Maybe worth a small comment about how defining the transforms is optional. Something like: `Note, not all transformations need to be defined. Undefined transformations fall back to the default behavior.`

python/src/transforms.cpp
```diff
@@ -888,8 +1332,10 @@ void init_transforms(nb::module_& m) {
          const nb::object& outputs,
          bool shapeless) {
         //  Try to get the name
-        auto n = fun.attr("__name__");
-        auto name = n.is_none() ? "compiled" : nb::cast<std::string>(n);
+        auto n =
+            nb::hasattr(fun, "__name__") ? fun.attr("__name__") : nb::none();
+        auto name = n.is_none() ? "compiled"
+                                : nb::cast<std::string>(fun.attr("__name__"));
```
awni, line 1338: Nice fix. Just leaving this as a todo for later. Probably should add this to the the custom function / checkpointed function at some point (?). Maybe we can refactor it a bit.

awni (approved):
Beautiful, thanks for adding this!

One thing I'm uncertain about if we want to deal with or not..we have the `Custom` primitive that we use for `mx.fast` transforms. We also have this `CustomTransforms` primitive. I wonder if there is enough overlapping functionality that it makes sense to merge them into some common thing 🤔 . I looked at it a bit.. and it's not obviously a good idea.. but just curious if you thought about that at all.

Issue #1187, opened 2024-06-05
[Feature] Add nan_to_num Function


abdussamettrkr:
Hello, In libraries such as Numpy, Jax, and Torch, we use a nan_to_num function to handle non-numeric values like NaNs and infinities. This function is currently not implemented in the MLX library. I believe adding this function would be useful.

awni:
For now, you can do this with where (maybe a couple of where's depending on how you want to handle inf).

```python
mx.where(mx.isnan(x), val, x)
```

abdussamettrkr:
Sorry, I meant to  'nan_to_num'. Thanks, I will use the 'where' function for now.

Commit2Cosmos:
Hi @awni, can I take it as my first contribution please? I have coded it up, but quite unsure about what the types of `nan`, `posinf` and `neginf` should be (right now it I've set it to `ScalarOrArray` and it seems to work; using `float` or `double` fails tests, because `to_array` converts them to `complex64`). I understand their types should be related to that of the input array, but cannot quite figure out how. I'll make a DPL for you to look at. Also, please let me know of any other tests you'd like to see.

awni:
I left comments in the code about this.

PR #1247, opened 2024-07-02
[Issue #1187] Add nan_to_num function initial attempt

Commit2Cosmos:
## Proposed changes

Adding a nan_to_num function, following issue #1187 

## Checklist

### DIFF
--- a/mlx/ops.cpp
+++ b/mlx/ops.cpp
@@ -1344,13 +1344,39 @@ array where(
 
 array nan_to_num(
     const array& a,
-    const array& nan,
-    const array& posinf,
-    const array& neginf,
+    const float& nan,
+    const float& posinf,
+    const float& neginf,
     StreamOrDevice s /* = {} */) {
-  auto out = where(isnan(a, s), nan, a, s);
-  out = where(isposinf(a, s), posinf, out, s);
-  out = where(isneginf(a, s), neginf, out, s);
+  Dtype dtype = a.dtype();
+
+  if (!issubdtype(dtype, inexact)) {
+    return a;
+  }
+
+  nan_ = array(nan, dtype);
+  posinf_ = array(posinf, dtype);
+  neginf_ = array(neginf, dtype);
+
+  // How to find the max value of a the same dtype as a?
+
+  // Checking if input values are too large for float16 and bfloat16
+  if (nan_ == std::numeric_limits<dtype>::infinity() ||
+      nan_ == -std::numeric_limits<dtype>::infinity()) {
+    nan_ = array(0.0, dtype);
+  }
+  if (posinf_ == std::numeric_limits<dtype>::infinity() ||
+      posinf_ == -std::numeric_limits<dtype>::infinity()) {
+    posinf_ = array(std::numeric_limits<dtype>::max());
+  }
+  if (neginf_ == std::numeric_limits<dtype>::infinity() ||
+      neginf_ == -std::numeric_limits<dtype>::infinity()) {
+    neginf_ = array(-std::numeric_limits<dtype>::max());
+  }
+
+  auto out = where(isnan(a, s), nan_, a, s);
+  out = where(isposinf(a, s), posinf_, out, s);
+  out = where(isneginf(a, s), neginf_, out, s);
 
   return out;
 }
--- a/mlx/ops.cpp
+++ b/mlx/ops.cpp
@@ -2,7 +2,6 @@
 #include <algorithm>
 #include <climits>
 #include <cmath>
-#include <iostream> // TODO
 #include <numeric>
 #include <set>
 #include <sstream>
--- a/mlx/ops.cpp
+++ b/mlx/ops.cpp
@@ -1,8 +1,8 @@
 // Copyright © 2023-2024 Apple Inc.
-
 #include <algorithm>
 #include <climits>
 #include <cmath>
+#include <iostream> // TODO
 #include <numeric>
 #include <set>
 #include <sstream>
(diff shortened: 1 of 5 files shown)


python/src/ops.cpp
```diff
@@ -3587,6 +3587,36 @@ void init_ops(nb::module_& m) {
...
+         const ScalarOrArray& posinf,
+         const ScalarOrArray& neginf,
+         StreamOrDevice s) {
+        return nan_to_num(
+            to_array(a), to_array(nan), to_array(posinf), to_array(neginf), s);
+      },
+      nb::arg(),
+      "nan"_a = 0,
+      "posinf"_a = std::numeric_limits<float>::max(),
+      "neginf"_a = -std::numeric_limits<float>::max(),
```
awni, line 3603: I think these should default to `None` and be chosen based on the dtype of `a`. We will presumably need to use different values for `fp16` and `bf16`.

python/tests/test_ops.py
```diff
@@ -1640,6 +1640,8 @@ def test_pad(self):
         self.assertTrue(mx.allclose(a_bwd[4:-2, 2:-4], df[0]).item())
 
     def test_where(self):
+        # print(mx.where(True, mx.array([[1, 2], [3, 4]]), 1))
+        # print(mx.where(True, 1, mx.array([[1, 2], [3, 4]])))
```
awni, line 1644: Remove comment?

mlx/ops.cpp
```diff
@@ -1330,6 +1330,19 @@ array where(
       inputs);
 }
 
+array nan_to_num(
+    const array& a,
+    const array& nan,
+    const array& posinf,
+    const array& neginf,
+    StreamOrDevice s /* = {} */) {
+  auto out = where(isnan(a, s), nan, a, s);
```
awni, line 1339: Probably should check the dtype of `a` here to see if it is inexact first. If it's not then just return `a` unmodified. Also worth adding a test case for that to make sure it works (e.g. test it when `a` has dtype `int32`.

python/tests/test_ops.py
```diff
@@ -1652,6 +1654,16 @@ def test_where(self):
             np.where,
         )
 
+    def test_nan_to_num(self):
+        self.assertCmpNumpy(
+            [mx.array([6, float("nan"), 2, 0])], mx.nan_to_num, np.nan_to_num
+        )
```
awni, line 1660: Could you add a test case for `mx.float16` as well?

awni (commented):
This looks great so far! There are a few modifications we need to make (left comments inline) but otherwise very nice.

python/src/ops.cpp
```diff
@@ -3587,6 +3587,36 @@ void init_ops(nb::module_& m) {
             array: The output containing elements selected from
             ``x`` and ``y``.
       )pbdoc");
+  m.def(
+      "nan_to_num",
+      [](const ScalarOrArray& a,
+         const ScalarOrArray& nan,
+         const ScalarOrArray& posinf,
+         const ScalarOrArray& neginf,
```
awni, line 3595: I think the type should be something like a variant of monostate (None), Python float, or Python int.

mlx/ops.h
```diff
@@ -400,6 +400,14 @@ array where(
     const array& y,
     StreamOrDevice s = {});
 
+/** TODO */
+array nan_to_num(
+    const array& a,
+    const array& nan,
+    const array& posinf,
+    const array& neginf,
```
awni, line 408: I would also make these types just `float` for now. If we want to follow the Python API they could be `optional<float>`. Maybe that is better.. though slightly more complicated.

python/src/ops.cpp
awni: Concretely I think the type `std::variant<std::monostate, int, float>` is what we want here. Then if it's the monostate (None) you can set the defaults accordingly inside the function.

Commit2Cosmos:
Thank you very much for your comments and patience! I've tried to implement those, but can't seem to figure out how to get the maximum value of the same dtype as the array. Do I have to cycle through all the posibilities, or is there some utility function that would help me do that?

mlx/ops.h
```diff
@@ -400,6 +400,14 @@ array where(
     const array& y,
     StreamOrDevice s = {});
 
+/** Replace NaN and infinities with finite numbers. */
+array nan_to_num(
+    const array& a,
+    const float& nan,
+    const float& posinf,
+    const float& neginf,
```
awni, line 408: I think these should be `std::optional<float>` so that we can have the default behavior in the C++ API.

python/src/ops.cpp
```diff
@@ -3587,6 +3587,52 @@ void init_ops(nb::module_& m) {
             array: The output containing elements selected from
             ``x`` and ``y``.
       )pbdoc");
+  m.def(
+      "nan_to_num",
+      [](const ScalarOrArray& a,
+         std::variant<std::monostate, int, float>& nan_,
+         std::variant<std::monostate, int, float>& posinf_,
+         std::variant<std::monostate, int, float>& neginf_,
```
awni, line 3595: On second thought these can be `std::optional<float>` and I believe `int` will properly cast to `float`.

Then you can pass them directly to the C++ API and handle the type-based max selection there.

mlx/ops.cpp
```diff
@@ -1330,6 +1330,45 @@ array where(
...
+    nan_ = array(0.0, dtype);
+  }
+  if (posinf_ == std::numeric_limits<dtype>::infinity() ||
+      posinf_ == -std::numeric_limits<dtype>::infinity()) {
+    posinf_ = array(std::numeric_limits<dtype>::max());
+  }
+  if (neginf_ == std::numeric_limits<dtype>::infinity() ||
+      neginf_ == -std::numeric_limits<dtype>::infinity()) {
+    neginf_ = array(-std::numeric_limits<dtype>::max());
+  }
```
awni, line 1363: Instead of this, I would:

- Make the inputs limits optional (e.g. `const std::optional<float>& nan` 
- check if the input has a value. Use it if so.
- If it doesn't have a value set it correctly kind of like how you are already doing. But you'll need to have a branch because `numeric_limits<dtype>` is not valid.

So something like:
```c++
if (dtype == float32) {
 posinf = std::numeric_limits<float>::max();
} else if (dtype == float16) {
...
} else if (dtype == bfloat16) {
...
}
```

Commit2Cosmos: Thank you for your comments! Everything is looking good apart from the issue below.
Currently part of the branch looks like this:
```cpp
if (posinf.has_value()) {
    posinf_ = posinf.value();
  } else if (dtype == float32) {
    posinf_ = std::numeric_limits<float>::max();
  } else if (dtype == float16) {
    posinf_ = std::numeric_limits<float16_t>::max();
  } else if (dtype == bfloat16) {
    posinf_ = std::numeric_limits<bfloat16_t>::max();
  }
  ```
It compiles fine, but when tested:
- `std::numeric_limits<float16_t>::max();` returns `0` (or `0.0F` to be precise). (In fact `min()` and `infinity()` all do for `float16_t`. Is this intended?)
-  `std::numeric_limits<bfloat16_t>::max();` throws `RuntimeError: Item size 2 for PEP 3118 buffer format string B does not match the dtype B item size 1.`

awni (approved):
Thanks for the addition!

Commit2Cosmos:
Thank you for the example solution! Sorry for making it more problematic than it had to be..

PR #1249, opened 2024-07-03
Fast Hadamard Transform

barronalex:
Add a fast [Hadamard transform](https://en.wikipedia.org/wiki/Hadamard_transform) in Metal.

Supports `n = m*2^k` where `m in (1, 12, 20, 28)`.  (e.g. Llama 3 70B has a hidden size of `28672 = 28*1024`).

Due to shared memory limits we support `2^k <= 8192` for FP32 and `2^k <= 16384` for FP16/BF16.

Planning to use this to enable low-bit, online quantization of the KV cache similar to [Quarot](https://arxiv.org/abs/2404.00456)/[SpinQuant](https://arxiv.org/abs/2405.16406).

## Benchmarks

We get close to full bandwidth for `2^k` and half bandwidth for `m*2^k` (since we do those in two uploads). This is much faster than manually doing `hadamard(N) @ x` for non-trivial batch size.

bench_float32
bench_float16


awni:
Very nice!!

From an API perspective, I'm wondering if this should live in the fast namespace?

The primitive does not have any transforms implemented (which is fine). I guess the question is really if we intend to implement them eventually. If yes, then maybe it makes sense to keep it as is. But if no, I would consider putting it in the fast package for now and maybe do a transformable fallback (`hadmard(n) @ x`) if it's not too tedious.

barronalex:
Agreed! The transforms are pretty simple so I added `vjp`/`jvp`/`vmap`.

mlx/primitives.h
```diff
@@ -1063,6 +1063,26 @@ class GreaterEqual : public UnaryPrimitive {
...
+  explicit Hadamard(Stream stream, float scale)
+      : UnaryPrimitive(stream), scale_(scale) {}
+
+  void eval_cpu(const std::vector<array>& inputs, array& out) override;
+  void eval_gpu(const std::vector<array>& inputs, array& out) override;
+
+  DEFINE_VMAP()
+  DEFINE_GRADS()
+  DEFINE_PRINT(Hadamard)
+  DEFINE_DEFAULT_IS_EQUIVALENT()
```
angeloskath, line 1077: Unfortunately the default `is_equivalent` is incorrect since we also need to check the scales. In general this is best left undefined unless certainly correct as it can cause quite hard to debug errors.

docs/src/python/ops.rst
```diff
@@ -72,6 +72,7 @@ Operations
    gather_qmm
    greater
    greater_equal
+   hadamard
```
angeloskath, line 75: Typo `hadamard_transform` .

angeloskath (changes requested):
Looks fantastic!

I left a few nitpicks, I think we can merge after that.

P.S.: Do we think we want a CPU implementation from the get go? Or we 'll simply add it later?

mlx/primitives.cpp
```diff
@@ -3950,4 +3950,37 @@ bool View::is_equivalent(const Primitive& other) const {
   return (dtype_ == a_other.dtype_);
 }
 
+std::pair<std::vector<array>, std::vector<int>> Hadamard::vmap(
```
angeloskath, line 3953:
```suggestion
bool Hadamard::is_equivalent(const Primitive& other) const {
  const Hadamard& h_other = static_cast<const Hadamard&>(other);
  return scale_ == h_other.scale_;
}

std::pair<std::vector<array>, std::vector<int>> Hadamard::vmap(
```

angeloskath (approved):
Left a suggestion for the `is_equivalent`. Otherwise looks awesome!

mlx/primitives.cpp
angeloskath: Also needs the declaration in `primitives.h` of course.

barronalex: Thanks! Added this in the latest commit.

PR #1253, opened 2024-07-06
Fix reshape copy bug

angeloskath:
Fixes https://github.com/ml-explore/mlx-swift/issues/104 and potentially https://github.com/ml-explore/mlx-examples/issues/642 .

The two first commits in the PR have two different fixes. The later one is what I consider a better choice. The bug is because the code in `Reshape` assumed that `copy` copies arrays of same total size but different shapes. That is not true, it is not checked but copy assumes `in.shape() == out.shape()`. I did not add this as an assertion but we could... let me know.

The fix in the previous PR was just allowing to copy from any shape to any shape assuming that the total number of elements is the same.

The current fix just fixes reshape to allocate and copy in place passing the correct output strides. I also updated the cpu copy to use `collapse_contiguous_dims` so we should get signficantly faster many-dimensional copies on the cpu.

### DIFF
--- a/mlx/backend/common/copy.cpp
+++ b/mlx/backend/common/copy.cpp
@@ -4,6 +4,7 @@
 
 #include "mlx/allocator.h"
 #include "mlx/backend/common/copy.h"
+#include "mlx/backend/common/utils.h"
 
 namespace mlx::core {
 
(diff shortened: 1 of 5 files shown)


awni (approved):
🚀  thanks for the fix!!

Issue #1213, opened 2024-06-15
Grid Sample functionality in MLX backend


sachinraja13:
I’m trying to implement Deformable DeTr in MLX and I’m stuck at the grid sampling part of the code. Can you help me replicate the torch.nn.functional.grid_sample function in MLX please?

sachinraja13:
`def grid_sample(input, grid, mode='bilinear', padding_mode='zeros'):
    N, C, H_in, W_in = input.shape
    N, H_out, W_out, _ = grid.shape

    # Normalize the grid to [0, H_in-1] and [0, W_in-1]
    grid_x = (grid[..., 0] + 1) * (W_in - 1) / 2
    grid_y = (grid[..., 1] + 1) * (H_in - 1) / 2
 
    if mode == 'bilinear':
        x0 = mx.floor(grid_x).astype(mx.int32)
        x1 = x0 + 1
        y0 = mx.floor(grid_y).astype(mx.int32)
        y1 = y0 + 1

        # Clip to the range [0, H_in-1] and [0, W_in-1]
        x0 = mx.clip(x0, 0, W_in - 1)
        x1 = mx.clip(x1, 0, W_in - 1)
        y0 = mx.clip(y0, 0, H_in - 1)
        y1 = mx.clip(y1, 0, H_in - 1)

        Ia = input[mx.arange(N)[:, None, None], :, y0, x0]
        Ib = input[mx.arange(N)[:, None, None], :, y1, x0]
        Ic = input[mx.arange(N)[:, None, None], :, y0, x1]
        Id = input[mx.arange(N)[:, None, None], :, y1, x1]
        

        wa = (x1 - grid_x) * (y1 - grid_y)
        wb = (x1 - grid_x) * (grid_y - y0)
        wc = (grid_x - x0) * (y1 - grid_y)
        wd = (grid_x - x0) * (grid_y - y0)

        output = wa[..., None] * Ia + wb[..., None] * Ib + wc[..., None] * Ic + wd[..., None] * Id
    elif mode == 'nearest':
        x = mx.round(grid_x).astype(mx.int32)
        y = mx.round(grid_y).astype(mx.int32)

        # Clip to the range [0, H_in-1] and [0, W_in-1]
        x = mx.clip(x, 0, W_in - 1)
        y = mx.clip(y, 0, H_in - 1)

        output = input[..., y, x]

    else:
        raise ValueError(f"Unsupported mode: {mode}")

    if padding_mode == 'zeros':
[...]

PR #1256, opened 2024-07-08
Add zero VJPs for bitwise ops and gather w.r.t. index

barronalex:
Motivated by implementing `torch.nn.functional.grid_sample` in MLX since it requires differentiating through indexing and bounds checks.

(see #1213)

### DIFF
--- a/mlx/primitives.cpp
+++ b/mlx/primitives.cpp
@@ -523,7 +523,7 @@ std::vector<array> AsType::vjp(
     const std::vector<array>&) {
   if (cotangents[0].dtype() != dtype_) {
     throw std::invalid_argument(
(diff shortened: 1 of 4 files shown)


mlx/primitives.cpp
```diff
@@ -523,7 +523,7 @@ std::vector<array> AsType::vjp(
     const std::vector<array>&) {
   if (cotangents[0].dtype() != dtype_) {
     throw std::invalid_argument(
-        "[astype] Type of cotangentsgent does not much primal output type.");
```
awni, line 526: :smile

awni (approved):
🚀  looks great, thanks!

Issue #1265, opened 2024-07-12
Adding option to follow PyTorch API in GELU initalization


AtakanTekparmak:
In [mlx/nn/layers/activations.py](https://github.com/ml-explore/mlx/blob/main/python/mlx/nn/layers/activations.py), the approximation options denoted by the `approx` argument are `'none'`(default), `precise` and `fast`. In PyTorch API the `precise` approximation is denoted by `tanh`, 
$\textrm{GELUApprox}(x) = 0.5x \left(1 + \tanh\left(\sqrt{\frac{2}{\pi}}(x + 0.044715x^3)\right)\right)$
We could change [line 555](https://github.com/ml-explore/mlx/blob/main/python/mlx/nn/layers/activations.py#L555) to:
```python
elif approx == "precise" or approx == "tanh":
```
[line 561](https://github.com/ml-explore/mlx/blob/main/python/mlx/nn/layers/activations.py#L561) to:
```python
f"The approximation should be in ['none', 'precise'/'tanh', 'fast'] but '{approx}' was given"
```
and [line 547](https://github.com/ml-explore/mlx/blob/main/python/mlx/nn/layers/activations.py#L547) to:
```python
approx ('none' | 'precise'/'fast' | 'fast'): Which approximation to gelu to use if any.
```
(and add tests where relevant) to allow for matching the torch API and easier code migration, if that is within the scope of MLX. I can add tests and open a PR if that's something we want to do?

awni:
I don't have a problem with adding this. If  you send a PR we can take a look.

PR #1268, opened 2024-07-15
feat: Added "tanh" option to GELU approximation

AtakanTekparmak:
## Proposed changes

This PR closes issue #1265. In the initialisation of GELU, the approximation options are `precise` and `fast`. The `precise` implementation is equivalent to PyTorch's `tanh`. I've added the option to initialise with `tanh` as well as `precise`.

## Checklist

### DIFF
--- a/python/mlx/nn/layers/activations.py
+++ b/python/mlx/nn/layers/activations.py
@@ -532,19 +532,21 @@ class GELU(Module):
 
     where :math:`\Phi(x)` is the Gaussian CDF.
 
-    However, if ``approx`` is set to 'precise'/'tanh' or 'fast' it applies
+    However, if ``approx`` is set to 'precise' or 'fast' it applies
 
     .. math::
         \textrm{GELUApprox}(x) &= 0.5 * x * \left(1 + \text{Tanh}\left((\sqrt{2 / \pi} * \left(x + 0.044715 * x^3\right)\right)\right) \\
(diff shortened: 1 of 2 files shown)


angeloskath (commented):
This looks great but I think the slash is confusing. I would change it to another option and in the docs clearly say that `tanh` and `precise` are the same approximation. For instance

```
However, if ``approx`` is set to 'precise' or 'tanh' it applies
...

and if it is set to 'fast' it applies
...
```

Or leave it as it were and add a `Note, that 'tanh' is an alias for precise`.

Sorry if this is a bit pedantic but I think especially in the `Args` and error message it is very confusing to have `'precise'/'tanh'` as a single value.

AtakanTekparmak:
> This looks great but I think the slash is confusing. I would change it to another option and in the docs clearly say that `tanh` and `precise` are the same approximation. For instance
> 
> ```
> However, if ``approx`` is set to 'precise' or 'tanh' it applies
> ...
> 
> and if it is set to 'fast' it applies
> ...
> ```
> 
> Or leave it as it were and add a `Note, that 'tanh' is an alias for precise`.
> 
> Sorry if this is a bit pedantic but I think especially in the `Args` and error message it is very confusing to have `'precise'/'tanh'` as a single value.

Changed the docstring to follow the comments you've made @angeloskath, sorry for the late reply, I thought this was merged already 😁. Let me know if there's anything else

angeloskath (approved):

PR #1269, opened 2024-07-15
Einsum

awni:
A working and reasonably optimized Einsum. Uses a greedy path algorithm following opt_einsum to find the contraction order.

Some benchmarks. The path only times are just computing the contraction order using `einsum_path`. The attention time does the full computation. When comparing to NumPy it's also using the same greedy algorithm to find the path.

```text
Timing little einsum path...
MLX ... 0.003 ms
NumPy... 0.040 ms
Timing big einsum path...
MLX ... 0.188 ms
NumPy... 0.385 ms
Timing einsum attention...
Regular ... 6.200 ms
Einsum ... 6.104 ms
```


angeloskath:
Looks great! Absolutely unbelievable given the extensive test cases but the first thing I tried actually resulted in an error. I will take a close look tomorrow to maybe fix it or figure out what is going on...

```python
a = mx.ones((3, 4, 5))
mx.einsum("ijk,lmk,ijf->lf", a, a, a) # IndexError
np.einsum("ijk,lmk,ijf->lf", a, a, a) # works
```

mlx/einsum.cpp
```diff
@@ -0,0 +1,825 @@
...
+    std::sort(out_str.begin(), out_str.end(), [&dim_dict](auto x, auto y) {
+      return dim_dict[x] < dim_dict[y];
+    });
+    Subscript new_output(std::move(out_str), std::move(best.output));
+
+    // Add the chosen contraction to the path
+    {
+      std::vector<Subscript> in_terms;
+      in_terms.push_back(std::move(inputs[best.y]));
+      in_terms.push_back(std::move(inputs[best.x]));
```
angeloskath, line 259: It took a while but I tracked it down :-)

The two lines above need to be swapped because the order Is wrong in `node.inputs`.

angeloskath (changes requested):
Found the bug. It still looks great tbh. I think besides that we need to add some more tests. I can do that when I get some time. If you prefer you can also just merge it as is (after the fix) and we do a PR for more tests later.

mlx/einsum.cpp
awni: Thanks for debugging that (sorry I forgot to reply to you, that was not necessary to debug my mistakes!)

I'll fix it and add a few more tests.

angeloskath: Nah it was a great break in between talks ;-)

angeloskath (approved):
🎉

awni:
I'm still planning to add a few more tests then I will merge this (assuming they pass). Sorry for the delay.

awni:
@angeloskath I added a bunch more tests which caught a few more issues. All the new and old tests clear. I'll merge it tomorrow if I don't hear from you in the interim.

PR #1270, opened 2024-07-15
Distributed layers

angeloskath:
Adds linear layers that allow training and inference of a model sharded across several devices. The main things added are

- float16/bfloat16 reductions for MPI
- AllToShardedLinear and its quantized sibling
- ShardedToAllLinear and its quantized sibling

simply changing linear layers to the above results in a model that works out of the box with distributed inference and training.

I am starting it as a draft so that we can iterate a bit on the design. The negative aspects of the above design are that we have yet another linear layer to think about when implementing LoRA and friends or weird new quantizations for instance. Perhaps it would be better to make the above layers with an internal linear layer so model surgery that swaps linear layers would still work out of the box.

### DIFF
--- a/python/mlx/nn/layers/distributed.py
+++ b/python/mlx/nn/layers/distributed.py
@@ -1,5 +1,6 @@
 # Copyright © 2024 Apple Inc.
 
+import math
 from functools import lru_cache
 from typing import Optional
 
@@ -168,7 +169,7 @@ class ShardedToAllLinear(Module):
         if self.group.size() > 1:
             # Perform the local projection and aggregate the results
             x = x @ self["weight"].T
-            x = mx.distributed.all_sum(x, group=group)
+            x = mx.distributed.all_sum(x, group=self.group)
 
             # Add the bias if we have one
             if "bias" in self:
@@ -316,9 +317,9 @@ class QuantizedAllToShardedLinear(Module):
             bits=quantized_linear_layer.bits,
             group=group,
         )
-        sl.weight = quantized_linear_layer.weight[r : step : (r + 1) * step] * 1
-        sl.scales = quantized_linear_layer.scales[r : step : (r + 1) * step] * 1
-        sl.biases = quantized_linear_layer.biases[r : step : (r + 1) * step] * 1
+        sl.weight = quantized_linear_layer.weight[r * step : (r + 1) * step] * 1
+        sl.scales = quantized_linear_layer.scales[r * step : (r + 1) * step] * 1
+        sl.biases = quantized_linear_layer.biases[r * step : (r + 1) * step] * 1
         if "bias" in quantized_linear_layer:
             sl.bias = quantized_linear_layer.bias[r * step : (r + 1) * step] * 1
 
@@ -413,7 +414,7 @@ class QuantizedShardedToAllLinear(Module):
             bits=self.bits,
         )
         if self.group.size() > 1:
-            x = mx.distributed.sum_all(x, group=group)
+            x = mx.distributed.all_sum(x, group=self.group)
         if "bias" in self:
             x = x + self["bias"]
         return x
@@ -428,6 +429,8 @@ class QuantizedShardedToAllLinear(Module):
         N = group.size()
         r = group.rank()
         output_dims, input_dims = quantized_linear_layer.weight.shape
+        step = input_dims // N
+        step_grouped = quantized_linear_layer.scales.shape[1] // N
         input_dims *= (32 // quantized_linear_layer.bits) * N
 
         sl = cls(
@@ -438,9 +441,15 @@ class QuantizedShardedToAllLinear(Module):
             bits=quantized_linear_layer.bits,
             group=group,
         )
-        sl.weight = quantized_linear_layer.weight[r : step : (r + 1) * step] * 1
-        sl.scales = quantized_linear_layer.scales[r : step : (r + 1) * step] * 1
-        sl.biases = quantized_linear_layer.biases[r : step : (r + 1) * step] * 1
+        sl.weight = quantized_linear_layer.weight[:, r * step : (r + 1) * step] * 1
+        sl.scales = (
+            quantized_linear_layer.scales[:, r * step_grouped : (r + 1) * step_grouped]
+            * 1
+        )
+        sl.biases = (
+            quantized_linear_layer.biases[:, r * step_grouped : (r + 1) * step_grouped]
+            * 1
+        )
         if "bias" in quantized_linear_layer:
             sl.bias = quantized_linear_layer.bias
 
--- a/python/mlx/nn/layers/distributed.py
+++ b/python/mlx/nn/layers/distributed.py
@@ -71,8 +71,10 @@ class AllToShardedLinear(Module):
             )
 
     def _extra_repr(self) -> str:
+        out_dims, in_dims = self.weight.shape
         N = self.group.size()
-        return f"input_dims={self.weight.shape[1]}, output_dims={N * self.weight.shape[0]}, bias={'bias' in self}"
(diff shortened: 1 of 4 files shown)


python/mlx/nn/layers/distributed.py
```diff
@@ -0,0 +1,456 @@
...
+        group = group or mx.distributed.init()
+        N = group.size()
+        r = group.rank()
+        output_dims, input_dims = linear_layer.weight.shape
+        step = output_dims // N
+
+        sl = cls(input_dims, output_dims, False, group)
+        # The multiplication with 1.0 forces a copy, perhaps change to
+        # something better when available.
+        sl.weight = linear_layer.weight[r * step : (r + 1) * step] * 1
```
awni, line 105: Is it possible the input buffer could be donated so we'd still hold on to the memory?

If so, maybe another option is to do `sl.weight[ ... ] = ...` that will force the copy since it's a slice update?

awni:
I kind of like this design. I like that it's all quite simple and easy to follow and we have a lot of control over how to shard the model (as in https://github.com/ml-explore/mlx-examples/pull/890).  We could possibly find a way to reduce the code needed for adding a new custom linear-like layer.. but the simplicity is nice, I wouldn't want to give that up.

python/mlx/nn/layers/distributed.py
angeloskath: Nice! That does sound better actually!

angeloskath: I am forcing contiguous now which now makes sure we are doing a copy.

angeloskath:
I am marking this ready for review. The main things that are new since I started the branch:

Exposing `mx.contiguous`. This ensures both that the array is contiguous and that it occupies at most `x.size() * x.itemsize() + 16384` bytes. Mainly a contiguous slice is still going to be copied.

`shard_linear` convenience function and `shard_inplace`. The first one just creates the appropriate linear layer quantized or not. The second actually shards the parameters in place which allows us to shard any layer and apply the collective operations as we see fit. It is used for instance to shard the single stream transformer blocks in FLUX but only perform one communication (https://github.com/ml-explore/mlx-examples/pull/1325). 

The sharding functions now also take a `groups` argument. This assumes the linear layer is a fused one and splits it according to the groups argument (evenly or percentage wise). I think the argument name may need improving here.

python/mlx/nn/layers/distributed.py
```diff
@@ -0,0 +1,553 @@
...
+    parameters: dict,
+    groups: Union[int, list] = 1,
+    group: Optional[mx.distributed.Group] = None,
+):
+    group = group or mx.distributed.init()
+    N = group.size()
+    r = group.rank()
+
+    # The multiplication with 1 forces a copy, perhaps change to
+    # something better when available.
```
awni, line 48: Nit remove comment?

python/mlx/nn/layers/distributed.py
```diff
@@ -0,0 +1,553 @@
...
+    for k in parameters:
+        if not isinstance(parameters[k], mx.array):
+            continue
+
+        axis = max(parameters[k].ndim - 2, 0)
```
awni, line 53: Should that just be always `0`? Maybe would work for conv in that case as well?

python/mlx/nn/layers/distributed.py
```diff
@@ -0,0 +1,553 @@
...
+    sharding: str,
+    *,
+    groups: Union[int, list] = 1,
+    group: Optional[mx.distributed.Group] = None,
+):
+    _check_sharding(sharding)
+    shard_function = (
+        _all_to_sharded if sharding == "all-to-sharded" else _sharded_to_all
+    )
+    module.update(shard_function(module.parameters(), groups=groups, group=group))
```
awni, line 115: Would it make sense to have the this take a callable which returns a sharding based on a key? It would be more like `nn.quantize` and capable of one-shot sharding a Module in place with a given policy.

awni:
> The sharding functions now also take a groups argument. This assumes the linear layer is a fused one and splits it according to the groups argument (evenly or percentage wise)

The purpose there is to allow uneven shardings? I think it would be good to think on a name that is more different from `group`.

python/mlx/nn/layers/distributed.py
angeloskath: Oops

python/mlx/nn/layers/distributed.py
angeloskath: Well it assumes linear layers (as in fully connected layers). It isn't 0 so that it can work with Switch layers and their variants. Perhaps combining with your comment below we could make this more general and support both.

python/mlx/nn/layers/distributed.py
angeloskath: Possibly you are right. Ideally we can keep the same interface but provide this as an extra. Returning a tuple (or named tuple) with `axis`, `groups` (renamed) given `path` and `weight` would a nice interface I think.

angeloskath:
> The purpose there is to allow uneven shardings?

Totally agree that we should name it something different. It isn't for uneven shadings in the sense that one node can take 70% of the computation. This isn't supported in this API. It is for weights that comprise several concatenated weights. In this case for the sharded linear to be valid we need to split, shard and concatenate. Otherwise one node will get all the queries and no keys and so on.

awni:
> Otherwise one node will get all the queries and no keys and so on.

Ah that makes sense now. Some suggestions on alternative names:

- shards
- segments
- sections
- splits

Maybe it makes sense to prefix `sub` with one of those like `sub_shards`?

angeloskath:
This should be ready for review again. I added a bunch of tests and changed the API to support a sharding function so we can shard layers in one go.

I don't think a solution like the `nn.quantize` one is very doable because the functionality changes based on the sharding type. This means that a linear layer maps to (at least) two sharded linear layers. As a result we can't have a `to_sharded` function cause it depends on the layer holding the linear and so on.

python/tests/mlx_distributed_tests.py
```diff
@@ -0,0 +1,250 @@
...
+import unittest
+
+import mlx.core as mx
+import mlx.nn as nn
+import mlx_tests
+from mlx.nn.layers.distributed import shard_inplace, shard_linear
+from mlx.nn.utils import average_gradients
+
+
+class MLXDistributedCommonTestCase(mlx_tests.MLXTestCase):
```
awni, line 12: Nice refactor 👍

python/tests/mlx_distributed_tests.py
```diff
@@ -0,0 +1,250 @@
...
+        mx.metal.reset_peak_memory()
+        scale = mx.array(2.0)
+        y = mx.distributed.all_sum(x)
+        mx.eval(y)
+        mx.synchronize(mx.default_stream(mx.default_device()))
+        all_sum_only = mx.metal.get_peak_memory()
+        y = mx.distributed.all_sum(x) * scale
+        mx.eval(y)
+        mx.synchronize(mx.default_stream(mx.default_device()))
+        all_sum_with_binary = mx.metal.get_peak_memory()
```
awni, line 80: Nit: update those to get rid of the runtime warning now.

awni (approved):
Looks awesome! 🚀

python/tests/mlx_distributed_tests.py
angeloskath: Ah good point :-)

angeloskath:
I am merging as is and when we use it a bit more and it settles I will add the layers to the docs (maybe with a dedicated page of their own ... not sure) and also expose `shard_linear` and `shard_inplace` to `mlx.nn` instead of `mlx.nn.layers.distributed`.

PR #1273, opened 2024-07-17
fix gguf

awni:
Needed to initialize weights to 0

### DIFF
--- a/mlx/io/gguf_quants.cpp
+++ b/mlx/io/gguf_quants.cpp
@@ -9,9 +9,7 @@
 namespace mlx::core {
 
 void unpack_32_4(uint8_t* data, int8_t* dst) {
-  for (int j = 0; j < 16; ++j) {
-    dst[j] = 0;
-  }
+  std::fill_n(dst, 16, 0);
   for (int j = 0; j < 16; ++j) {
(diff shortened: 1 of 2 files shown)


xnorai:
Tested the change with my C++ code - the issue no longer reproduces.

mlx/io/gguf_quants.cpp
```diff
@@ -9,15 +9,18 @@
 namespace mlx::core {
 
 void unpack_32_4(uint8_t* data, int8_t* dst) {
-  for (int64_t j = 0; j < 16; ++j) {
+  for (int j = 0; j < 16; ++j) {
+    dst[j] = 0;
+  }
```
angeloskath, line 14: Out of curiosity, why not `std::fill_n(dst, 16, 0);`?

Possible answers
a) it fits much better with the rest of the function
b) why include a header instead of 3 lines
c) compiles faster
d) all of the above 😅

angeloskath (approved):
Perfect!

mlx/io/gguf_quants.cpp
awni: e) didn't think of it

PR #1274, opened 2024-07-17
Fix leak with multi-output primitives

awni:
The test case shows how this was happening. I think the fix is safe since the only case where we could have a race is in the asynchronous eval in which case:

- No eval is allowed during transforms so the array will be detached during eval
- Hence, we don't go into the multi-output destructor since it will be handled by the detach.

### DIFF
--- a/mlx/array.cpp
+++ b/mlx/array.cpp
@@ -175,7 +175,7 @@ array::~array() {
     return;
   }
 
-  // Ignore arrays that will be detached during async eval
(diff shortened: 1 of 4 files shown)


angeloskath (approved):
I think that makes a lot of sense! Do we also want to add the `RetainGraph` change here or another PR?

awni:
Let's merge this and I will do a follow on with the retain graph fix.

PR #1275, opened 2024-07-19
[WIP] Adding `mlx.distributions`

ariG23498:
This PR aims to propose the `mlx.distributions` API similar to that of `torch.distributions`.

The PyTorch API seems to have the two main things:
- A [`Constraints`](https://github.com/pytorch/pytorch/blob/main/torch/distributions/constraints.py) API, that helps in enforcing constraints to the random variables.
- The [`Distribution`](https://github.com/pytorch/pytorch/blob/main/torch/distributions/distribution.py) API, which serves as the base template for all the distributions there are.

At the current stage, I have copied the `Normal` distribution implementation from [mlx-examples repository](https://github.com/ml-explore/mlx-examples/blob/main/normalizing_flow/distributions.py). The idea would be to be able to add more distributions to mlx and align them as closely as possible to the PyTorch API.

What do you folks think? And what would be the best way to approach this?

CC: [Kashif](https://github.com/kashif)

### DIFF
--- a/python/mlx/distributions/normal.py
+++ b/python/mlx/distributions/normal.py
@@ -1,0 +1,29 @@
+import math
+from typing import Optional, Tuple, Union
+
+import mlx.core as mx
+
+
+class Normal:
+    def __init__(self, mu: mx.array, sigma: mx.array):
+        super().__init__()
+        self.mu = mu
+        self.sigma = sigma
+
+    def sample(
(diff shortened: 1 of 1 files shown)


awni:
This looks cool! 

My suggestions is to develop this a third party package as there is no clear reason to put it in MLX core. I haven't seen anyone asking for this..so I don't think we should add it to MLX core until there are more active use cases for it.

PR #1279, opened 2024-07-22
Implement sampling from laplace distribution.

grananqvist:
## Proposed changes

Implements sampling from a Laplace distribution. Commonly used in Differential Privacy, e.g. in our framework `pfl-research` https://github.com/apple/pfl-research/blob/develop/pfl/internal/ops/mlx_ops.py#L91. 

## Checklist


mlx/random.h
```diff
@@ -224,4 +224,34 @@ array categorical(
     const std::optional<array>& key = std::nullopt,
     StreamOrDevice s = {});
 
+/** Generate samples from the laplace distribution. */
```
grananqvist, line 227: Using same interface as `normal`

tests/random_tests.cpp
```diff
@@ -640,3 +640,74 @@ TEST_CASE("test categorical") {
   CHECK_EQ(categorical(logits, -2, 7).shape(), std::vector<int>{5, 3, 7});
   CHECK_EQ(categorical(logits, -3, 7).shape(), std::vector<int>{4, 3, 7});
 }
+
+TEST_CASE("test laplace") {
```
grananqvist, line 644: All kinds of tests as is done in normal, but also add some statistical tests about the distribution

angeloskath (approved):
That looks great! Thanks especially for the C++ tests :-)

PR #1281, opened 2024-07-23
Fix tests on main + some minor improvements

awni:
- Fix expm1 tolerances and building
- Improve MPI build for case when MPI is found but not available (e.g. installed in a different python environment)
- Typo in pad op


barronalex (approved):
Thanks for the quick fix!

PR #1282, opened 2024-07-23
Fused Affine Quantize/Dequantize ops

barronalex:
Add fused ops for affine quantization and dequantization.

Gives about a 1% tps improvement on Mistral 7B on my M1 Max due to the embedding `dequantize` call:

Mistral 7B -- M1 Max
Before:
```
==========
Prompt: 88.394 tokens-per-sec
Generation: 54.417 tokens-per-sec
```

After:
```
==========
Prompt: 90.860 tokens-per-sec
Generation: 54.976 tokens-per-sec
```
Having quantize as a fused op should significantly reduce the overhead once we have a quantized KV cache (coming soon).


mlx/fast.h
```diff
@@ -39,4 +39,26 @@ array scaled_dot_product_attention(
...
+    int bits = 4,
+    StreamOrDevice s = {});
+
+array affine_quantize_with_params(
+    const array& w,
+    const array& scales,
+    const array& biases,
+    int group_size = 64,
+    int bits = 4,
+    StreamOrDevice s = {});
```
awni, line 54: Could you explain what the purpose of that one is?

barronalex: Yes it's useful for something like GPTQ where you calculate the quantization scales and biases for a group and then apply them to individual columns of the weight matrix which get updated as you iterate through them (e.g. [here](https://github.com/IST-DASLab/gptq/blob/2d65066eeb06a5c9ff5184d8cebdf33662c67faf/gptq.py#L130)).

I'm not sure if this is the best naming for it (or maybe it should be an overload of `affine_quantize`?).

angeloskath (approved):
Looks fantastic!

I was a bit torn about implementing both the scales, biases computation and just affine quantization. I think it is fine though as it saves ~25% on the quantization of big models and will also enable fancy stuff if needed like quantizing KV caches etc.

Quantizing gemma2 27B

```
before: 29s
after:  22s
```

the bottleneck of course is still writing to disk. ie it takes minutes to dequantize the same model.

mlx/fast.h
awni: > I'm not sure if this is the best naming for it (or maybe it should be an overload of affine_quantize?).

Indeed `with_params` is a bit vague. Maybe an overload is better.

python/src/fast.cpp
```diff
@@ -138,4 +139,138 @@ void init_fast(nb::module_& parent_module) {
         Returns:
             array: The output array.
       )pbdoc");
+
+  m.def(
```
awni, line 143: I'm wondering if we should have these functions in the Python API? Should we not encourage people to use `mx.quantize` and leave the implementation details hidden? It may be confusing to have two functions to do the same thing..

Or maybe we expose them later if we see that it's useful?

angeloskath:
Hmm I think it is the `*_with_params` that is the most confusing part here. If we don't consider that for a moment, we have the top-level `quantize` op (which in the future will receive a `mode` argument to support "affine", "nf4" or a new thing) and then we have the `fast::affine_quantize` which is a kernel implementation of `quantize` with mode "affine".

Now I do think that providing our own "params" is important. Arguably it is more important than computing them in one kernel call. So if I had to simplify, I would only implement a kernel for `affine_dequantize` and `affine_quantize` which both would accept the scales and biases.

Having said that, since Alex did all the work to have a fast kernel that also computes scales and biases in one go then it 'd be nice to keep it.

Let me know what you guys think.

barronalex:
So in that case would we also change the python API so that `mx.quantize` optionally accepts scales and biases? Then I think we could get rid of the python endpoints for `fast::affine_quantize` which I agree are a bit confusing.

Whether it switches to a special kernel for computing the scales/biases would then be an implementation detail.

awni:
> So in that case would we also change the python API so that mx.quantize optionally accepts scales and biases?

Or we could expose `mx.fast.affine_quantize` in python which accepts them in order to keep the `mx.quantize` op API from getting too big. When you've precomputed the scales/biases you can use that.

So to summarize one possible option:

- `mx.quantize` API stays the same for now (eventually add mode kwarg)
- `mx.quantize` / `mx.dequantize` implementations uses fast Primitives where available (implementation detail)
- Add `mx.fast.affine_quantize` which accepts `scales` / `biases`. Does it make sense to have them be optional or just require them for now?


The other option is:

- `mx.quantize` API adds optional scales and biases
- `mx.quantize` / `mx.dequantize` implementations uses fast Primitives where available (implementation detail)
- Don't add fast ops to Python API for now.

barronalex:
I like the first option since it sounds nice not to touch the `mx.quantize` API.
I think `affine_quantize` can require `scales` / `biases` for now.

I'll push a change doing that shortly unless @angeloskath disagrees.

angeloskath:
Nope. I think option 1 is the best!

PR #1283, opened 2024-07-23
Fix bfloat16 Hadamard

barronalex:
Fix Hadamard transform for `bfloat16`.

Also change the default scale to $\frac{1}{\sqrt{N}}$ so that by default the Hadamard matrix used in the transform is orthonormal.

### DIFF
--- a/mlx/ops.cpp
+++ b/mlx/ops.cpp
@@ -453,8 +453,10 @@ array flatten(const array& a, StreamOrDevice s /* = {} */) {
 
 array hadamard_transform(
     const array& a,
-    float scale /* = 1.0 */,
+    std::optional<float> scale_ /* = std::nullopt */,
     StreamOrDevice s /* = {} */) {
+  // Default to an orthonormal Hadamard matrix scaled by 1/sqrt(N)
+  float scale = scale_.value_or(1.0f / std::sqrt(a.shape(-1)));
   auto dtype = issubdtype(a.dtype(), floating) ? a.dtype() : float32;
   return array(
       a.shape(),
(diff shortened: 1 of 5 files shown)


mlx/ops.cpp
```diff
@@ -453,8 +453,10 @@ array flatten(const array& a, StreamOrDevice s /* = {} */) {
 
 array hadamard_transform(
     const array& a,
-    float scale /* = 1.0 */,
+    std::optional<float> scale_ /* = std::nullopt */,
     StreamOrDevice s /* = {} */) {
+  // Default to an orthonormal Hadamard matrix scaled by 1/sqrt(N)
+  float scale = scale_.value_or(1.0f / std::sqrt(a.shape(-1)));
```
awni, line 459: Nitty nit: ternaries are nicere here because only one of the sides is evaluated:

```
float scale = scale_.has_value ? *scale_ : std::sqrt(a.shape(-1)))
```

python/src/ops.cpp
```diff
@@ -4401,6 +4401,7 @@ void init_ops(nb::module_& m) {
         Args:
             a (array): Input array or scalar.
             scale (float): Scale the output by this factor.
+              Defaults to `1/sqrt(a.shape[-1])` so that H is orthonormal.
```
awni, line 4404:
```suggestion
              Defaults to ``1/sqrt(a.shape[-1])`` so that the Hadamard matrix is orthonormal.
```

I think `H` doesn't have context here.

awni (approved):
Nice, thanks! Left a couple minor comments. Feel free to merge after!

barronalex:
Thanks for the quick review! Addressed the comments

PR #1285, opened 2024-07-24
Fix GPU sort for large arrays

### LINKED ISSUE #1254: [Feature] Enable Metal argsort, sort for > 2M elements along an axis
```python
import mlx.core as mx

a = mx.zeros((3000000,))
mx.sort(a)
```

jagrit06:
## Proposed changes

Fixes #1254 - GPU sort should now support arbitrarily large arrays (as long as their shape is contained in `int`)
Added relevant test 

## Checklist

### DIFF
--- a/mlx/ops.cpp
+++ b/mlx/ops.cpp
@@ -1785,15 +1785,6 @@ array sort(const array& a, int axis, StreamOrDevice s /* = {} */) {
     throw std::invalid_argument(msg.str());
   }
 
-  // TODO: Fix GPU kernel
-  if (a.shape(axis) >= (1u << 21) && to_stream(s).device.type == Device::gpu) {
-    std::ostringstream msg;
-    msg << "[sort] GPU sort cannot handle sort axis of >= 2M elements,"
-        << " got array with sort axis size " << a.shape(axis) << "."
-        << " Please place this operation on the CPU instead.";
-    throw std::runtime_error(msg.str());
-  }
-
   return array(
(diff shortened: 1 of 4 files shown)


python/tests/test_ops.py
```diff
@@ -1840,6 +1840,15 @@ def test_sort(self):
             self.assertTrue(np.array_equal(c_np, c_mx))
             self.assertEqual(b_mx.dtype, c_mx.dtype)
 
+        # Test very large array
+        if mx.default_device() == mx.gpu:
```
barronalex, line 1844: Is it just too slow for CI on CPU?

barronalex (approved):
Thanks for looking into this!

mlx/ops.cpp
```diff
@@ -1785,15 +1785,6 @@ array sort(const array& a, int axis, StreamOrDevice s /* = {} */) {
     throw std::invalid_argument(msg.str());
   }
 
-  // TODO: Fix GPU kernel
```
barronalex, line 1788: Is worth a warning here for sorts greater than maxint?

jagrit06: `array::shape` is made of ints, so each of those will be `<= INT_MAX`

barronalex: Ah good point -- nice!

awni:
Thanks for the quick fix 🚀

awni:
That looks like an oversight. Will fix for the next release.

awni:
(Fixed in #1289 )

Issue #48, opened 2023-12-07
[RFC] Implement the Python array API standard


lucascolley:
[The Python array API standard](https://data-apis.org/array-api/2022.12/) standardises common functionality across Python array/tensor libraries. NumPy, PyTorch and CuPy are planning to have full implementations, and Dask and JAX also have implementations in progress. You could implement this in your main namespace or a separate namespace.

Why should you do this? As well as making it easier for users to convert existing NumPy/PyTorch/CuPy code to MLX, there is potential for interoperability with other libraries. For example, from the NumPy ecosystem, [SciPy](https://github.com/scipy/scipy/issues/18286) and [scikit-learn](https://github.com/scikit-learn/scikit-learn/issues/22352) have partial experimental support for arrays which comply with the standard.

If you are interested in this, the consortium would love to hear feedback over at https://github.com/data-apis/consortium-feedback/. Some potential pain points, such as missing `float64` support, have already been discussed very briefly in https://github.com/data-apis/array-api/issues/719.

lucascolley:
quoting @awni in gh-12:

> We had the luxury of picking the best from all the frameworks we've used and worked on in the past and combining them into something new.

In terms of API design (implementation aside), this is the same process as what the consortium did in creation of the standard - I imagine it would be fruitful to discuss where the differences are and why one might be preferable.

arpan-dhatt:
+1
Implementing this standard would help libraries like einops "just work" when dealing with MLX arrays. That particular library has been cropping up a lot for array reshape/transpose/stack, etc in the HyenaDNA model I'm trying to port.

It can also help clear up ambiguities in API design such as #113 and create a "checklist" of basic ops that should be implemented but are not yet, like linspace.

awni:
Are there any differences between the Python array API standard and NumPy or is the standard basically a subset of NumPy? If it's the latter, then I would say we are already on track to implement the standard.

Either way though, we will definitely take it into consideration as we continue to update the API!

lucascolley:
> Are there any differences between the Python array API standard and NumPy or is the standard basically a subset of NumPy?

Here is [the tracking issue for support in the main NumPy namespace](https://github.com/numpy/numpy/issues/25076) (making the NumPy API a superset of the standard). Some decisions were made to differ from NumPy where other libraries seemed to have improved upon NumPy, but it is quite similar. @mtsokol and @rgommers are working on a proposal to continue the work into making NumPy a superset.

rgommers:
[...]

PR #1289, opened 2024-07-26
Array api

awni:
Some updates to match the array api and remove deprecated constants.

These updates allow `mx.array` to be used with einops:

```python
import mlx.core as mx
from einops.array_api import rearrange, reduce, repeat, pack, unpack

# rearrange elements according to the pattern
input_tensor = mx.zeros((2,3,4))
output_tensor = rearrange(input_tensor, 't b c -> b c t')
print(output_tensor.shape)

# combine rearrangement and reduction
input_tensor = mx.zeros((2, 3, 2*3, 2*4))
output_tensor = reduce(input_tensor, 'b c (h h2) (w w2) -> b h w c', 'mean', h2=2, w2=2)
print(output_tensor.shape)

# copy along a new axis
input_tensor = mx.zeros((2, 3))
output_tensor = repeat(input_tensor, 'h w -> h w c', c=3)
print(output_tensor.shape)

# pack and unpack allow reversibly 'packing' multiple tensors into one.
# Packed tensors may be of different dimensionality:
h, w = 100, 200
image_rgb = mx.zeros([h, w, 3])
image_depth = mx.zeros([h, w])
# but we can stack them
image_rgbd, ps = pack([image_rgb, image_depth], 'h w *')
print(image_rgbd.shape)
print(ps)

unpacked_rgb, unpacked_depth = unpack(image_rgbd, ps, 'h w *')
print(unpacked_rgb.shape, unpacked_depth.shape)                                                             
```

### DIFF
--- a/mlx/ops.cpp
+++ b/mlx/ops.cpp
@@ -1839,15 +1839,6 @@ array argsort(const array& a, int axis, StreamOrDevice s /* = {} */) {
     throw std::invalid_argument(msg.str());
   }
 
-  // TODO: Fix GPU kernel
-  if (a.shape(axis) >= (1u << 21) && to_stream(s).device.type == Device::gpu) {
-    std::ostringstream msg;
-    msg << "[argsort] GPU sort cannot handle sort axis of >= 2M elements,"
-        << " got array with sort axis size " << a.shape(axis) << "."
-        << " Please place this operation on the CPU instead.";
-    throw std::runtime_error(msg.str());
-  }
-
   return array(
       a.shape(), uint32, std::make_shared<ArgSort>(to_stream(s), axis), {a});
 }
--- a/python/src/array.cpp
+++ b/python/src/array.cpp
@@ -305,9 +305,12 @@ void init_array(nb::module_& m) {
           },
           "api_version"_a = nb::none(),
           R"pbdoc(
-            Used to apply updates at the given indices.
+            Returns an object that has all the array API functions on it.
 
-             Args:
+            See the `Python array API <https://data-apis.org/array-api/latest/index.html>`_
+            for more information.
+
+            Args:
                 api_version (str, optional): String representing the version
                   of the array API spec to return. Default: ``None``.
 
--- a/python/src/array.cpp
+++ b/python/src/array.cpp
@@ -294,6 +294,26 @@ void init_array(nb::module_& m) {
             Returns:
                 array: The array with type ``dtype``.
           )pbdoc")
+      .def(
+          "__array_namespace__",
+          [](const array& a, const std::optional<std::string>& api_version) {
+            if (api_version) {
+              throw std::invalid_argument(
+                  "Explicitly specifying api_version is not yet implemented.");
+            }
+            return nb::module_::import_("mlx.core");
+          },
+          "api_version"_a = nb::none(),
+          R"pbdoc(
+            Used to apply updates at the given indices.
+
+             Args:
+                api_version (str, optional): String representing the version
+                  of the array API spec to return. Default: ``None``.
+
+            Returns:
+                out (Any): An object representing the array API namespace.
+          )pbdoc")
       .def("__getitem__", mlx_get_item, nb::arg().none())
       .def("__setitem__", mlx_set_item, nb::arg().none(), nb::arg())
       .def_prop_ro(
--- a/python/src/constants.cpp
+++ b/python/src/constants.cpp
@@ -6,18 +6,9 @@
 namespace nb = nanobind;
 
 void init_constants(nb::module_& m) {
-  m.attr("Inf") = std::numeric_limits<double>::infinity();
-  m.attr("Infinity") = std::numeric_limits<double>::infinity();
-  m.attr("NAN") = NAN;
-  m.attr("NINF") = -std::numeric_limits<double>::infinity();
-  m.attr("NZERO") = -0.0;
-  m.attr("NaN") = NAN;
-  m.attr("PINF") = std::numeric_limits<double>::infinity();
-  m.attr("PZERO") = 0.0;
   m.attr("e") = 2.71828182845904523536028747135266249775724709369995;
   m.attr("euler_gamma") = 0.5772156649015328606065120900824024310421;
   m.attr("inf") = std::numeric_limits<double>::infinity();
-  m.attr("infty") = std::numeric_limits<double>::infinity();
   m.attr("nan") = NAN;
   m.attr("newaxis") = nb::none();
   m.attr("pi") = 3.1415926535897932384626433;
--- a/python/src/ops.cpp
+++ b/python/src/ops.cpp
@@ -2061,7 +2061,7 @@ void init_ops(nb::module_& m) {
          const std::optional<std::vector<int>>& axes,
          StreamOrDevice s) {
         if (axes.has_value()) {
-          return transpose(a, get_reduce_axes(axes.value(), a.ndim()), s);
+          return transpose(a, *axes, s);
         } else {
           return transpose(a, s);
         }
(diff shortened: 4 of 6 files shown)


python/src/array.cpp
```diff
@@ -294,6 +294,26 @@ void init_array(nb::module_& m) {
...
+          [](const array& a, const std::optional<std::string>& api_version) {
+            if (api_version) {
+              throw std::invalid_argument(
+                  "Explicitly specifying api_version is not yet implemented.");
+            }
+            return nb::module_::import_("mlx.core");
+          },
+          "api_version"_a = nb::none(),
+          R"pbdoc(
+            Used to apply updates at the given indices.
```
barronalex, line 308: Nit: doc string? Took me a sec to figure out this is how einops knows to call `mx.permute_dims` etc

mlx/ops.cpp
```diff
@@ -1839,15 +1839,6 @@ array argsort(const array& a, int axis, StreamOrDevice s /* = {} */) {
     throw std::invalid_argument(msg.str());
   }
 
-  // TODO: Fix GPU kernel
```
barronalex, line 1842: Thanks for fixing this!

barronalex (approved):
This is so nice! Excited to use it. 🚀

python/src/array.cpp
awni: Ah that's a copy typo, good catch.

lucascolley:
great to see progress towards gh-48! FYI, there has been interest in adding a fully-compliant wrapper to array-api-compat in https://github.com/data-apis/array-api-compat/issues/162.

awni:
> FYI, there has been interest in adding a fully-compliant wrapper to array-api-compat in data-apis/array-api-compat#162

Cool, thanks for sharing! If there's anything we can do to help on the MLX side let us know.

asmeurer:
The main thing that would help is if you could let us know what your plans are for MLX and the array API. If you're planning to have full compliance in MLX itself, then there's not much that we need to do in array-api-compat, other than documentation and possibly adding [helper functions](https://data-apis.org/array-api-compat/helper-functions.html). This is similar to how JAX and sparse work. 

On the other hand, if there are some functionalities that you won't be able to support for backwards compatibility, we should add wrappers to array-api-compat. (if there are functions that you can't support because they simply aren't supported by the hardware, then there's no way to wrap that and we just have to document it)

awni:
Looking through the array API I think for 95+% of it we already have an implementation or would like to have one.

There are a few exceptions that I noticed:

- Attributes and methods related to devices, like [device](https://data-apis.org/array-api/latest/API_specification/generated/array_api.array.device.html#array_api.array.device) and [to_device](https://data-apis.org/array-api/latest/API_specification/generated/array_api.array.to_device.html#array_api.array.to_device). Arrays in MLX live in unified memory and don't belong to a specific device and can't be moved to a device
- The [empty](https://data-apis.org/array-api/latest/API_specification/generated/array_api.empty.html#array_api.empty) creation function. There are no true "in-place" ops in MLX. We manage memory buffer sharing under the hood. So something like an `empty` doesn't really make sense. 
- Operations with output shapes which depend on input data (e.g. boolean indexing and similar ops). Those are hard to do in MLX and we may not be able to support them at least in their typical form.

Let me know what you think about that.

lucascolley:
I believe there are also `float64` and `complex128` dtypes which don't exist in MLX.

awni:
Good point I missed that. And we don’t plan to add them as they are not natively supported for Apple GPUs.

betatim:
The lack of `float64` shouldn't be a problem. For example PyTorch with the MPS backend (their way of using the GPU in modern Apple laptops) also doesn't support `float64` but PyTorch is still considered compatible.

I think how to deal with the fact that mlx is designed with a "native" unified memory experience (I don't know a better short way to describe it :D) will be interesting to figure out. On the face of it I think mlx could do what Numpy does, have a `device` attribute that is always the same. In Numpy's case it is always the string `"cpu"`. More interesting might be how to deal with the fact that in mlx operations can be placed on a device. This is an idea that doesn't exist in any(?) other array library and also doesn't exist in the array API standard.

Maybe at the level of array API standard compatibility it isn't such a big deal, but as a "array consumer" in scikit-learn I am currently still scratching my head how we'd write code that works with mlx and (say) PyTorch inputs. Without a bunch of `if` statements to conditionally add `stream=mx.gpu` as argument to operations that I'd like to have executed on the GPU.

awni:
> Without a bunch of if statements to conditionally add stream=mx.gpu as argument to operations that I'd like to have executed on the GPU.

The default device is the GPU so you wouldn't need to place any operations explicitly unless you wanted to change the default.

To avoid needing a bunch of if statements, one possibility is to use a stream context manager e.g.

```python
with mx.stream(mx.cpu):
   # call the function / do the work
```

For the most part though running everything on the GPU works pretty well. I'm not sure how you feel about using that as the default though..

awni:
> You might want to use the test suite to see what is missing in more detail

Indeed I wasn't aware of it. Good idea.

> Are you saying that things like += don't work at all?

No they work as expected. They just don't give any guarantees about the underlying memory buffer being unchanged.

> If you don't have a real "empty" function you could always alias empty = zeros

Seems reasonable!

> It also sounds like an is_mlx_array helper in array-api-compat would be useful since MLX has some features that are specific to it (like the per-device computation features) that consuming code might want to take advantage of.

👍

awni:
I was trying to use MLX with scikit learn through the array api back-end. It failed because MLX doesn't have `float64` dtype. 

It would be super cool if we could get MLX to work with Scikit-learn.. but I'm not sure if there is a workaround for the `float64` issue. What do you all think about that?

Here's the example I tried for reference:

```python
from sklearn.datasets import make_classification
from sklearn import config_context
from sklearn.discriminant_analysis import LinearDiscriminantAnalysis
import mlx.core as mx

X_np, y_np = make_classification(random_state=0)
X_mx = mx.array(X_np)
y_mx = mx.array(y_np)

with config_context(array_api_dispatch=True):
    lda = LinearDiscriminantAnalysis()
    X_trans = lda.fit_transform(X_mx, y_mx)

print(type(X_trans))
```

lucascolley:
Do you just get an error message like "`mx` has no attribute `float64`"?

If so, and I'm not saying this is a good solution, things should run if you wrap MLX such that `xp.float64` is actually the `float32` dtype, but 'pretends' it is `float64`? (I may be missing something though)

awni:
Yea exactly.  And yes that's a good suggestion for a patch.. though long term it might be worth thinking about how to deal with that in the array API / compatibility layer.

lucascolley:
The functionality exists to query which dtypes a library supports: https://data-apis.org/array-api/latest/API_specification/generated/array_api.info.__array_namespace_info__.html

Indeed, based on https://github.com/data-apis/array-api/issues/640#issuecomment-1713475644, it sounds like the spec will be relaxed at some point to accommodate for missing dtypes. Consumers being able to tell which dtypes it can use and act accordingly can be done via this introspection method, but we're going to want a helper where we can say:

- If it exists, use {this dtype}
- If not, fallback to {one of these dtypes}
- Else, raise an exception

Probably something that should belong in array-api-compat, however I can't immediately see how we would want this to work. Most dtype selections happen through `xp.asarray` and `xp.astype`, but it's not clear to me whether we would want to:

- wrap these functions themselves
- wrap the input dtypes to these functions
- or something else

It could get quite complex if there are varying policies for different parts of a library, like if we are perfectly happy to fallback from float64 to float32 in some places, but not others. I suppose this case in itself is not too bad, but it would be pretty complex to generalize for weirder *(based on the standards of today...)* sets of implemented dtypes.

There is also the case of testing. It doesn't seem ideal to have to add a line to skip based on whether the dtype is supported in every test which is parameterized by standard dtypes, but I suppose that would work.

awni:
> could you open an issue on the scikit-learn repository with your code snippet and a few words of context (for those who haven't followed this thread)

Yes, done!

PR #1299, opened 2024-07-30
Fixes for large arrays with a few ops

awni:
Fixes for arrays with more than `UINT32_MAX` elements. (Corresponding tests in long running nightly tests clear).

- softmax
- RMS norm + VJP
- Layer norm + VJP
- Unary
- Binary and binary two
- Ternary
- Copy

### DIFF
--- a/mlx/backend/metal/binary.cpp
+++ b/mlx/backend/metal/binary.cpp
@@ -21,10 +21,43 @@ namespace mlx::core {
 
 constexpr int MAX_BINARY_SPECIALIZED_DIMS = 5;
 
+std::string get_kernel_name(
+    BinaryOpType bopt,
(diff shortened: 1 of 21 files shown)


angeloskath (approved):
Awesome!

Just as an aside, there is still the case that `data_size / axis_size > UINT32_MAX` which will fail as well as the `n_rows * threadgroup_size > UINT32_MAX` which will also fail. Obviously these are quite a bit rarer cases so we should merge this as is.

awni:
Indeed.. it's far from bullet proof. I think also the non-contiguous versions may still be broken in some cases. For example when we collapse contiguous dimension if the resulting dimension is `> UINT_MAX`.

PR #1300, opened 2024-07-31
Fix scatter, reduce

awni:
- Fixes scatter for large sizes
- Probably will be part of a fix for some reductions for large sizes 

TODO: measure impact on runtime to see if we need to specialize this.


angeloskath:
Ok I think this is ready for a review at least the reduction part (or I am just tired of improving it). I will add some benchmarks and then go ahead and benchmark the scatter that is also in this PR.

It is a bit big and there are several changes on various parts so sorry about that. I will also write a long running test and a fuzzer to make sure that every size is properly checked in our nightly tests.

All reduce
----------

When the array is big everything practically gets the memory bandwidth so it's all the same. But there is a difference for int8.

all_reduce

Row reduce
------------
row_reduce
row_reduce_pre

Col reduce
-----------

col_reduce
col_reduce_pre

awni:
🤤

awni:
I don't know what kind of magic you did here..but those graphs are pretty amazing.

mlx/backend/metal/kernels/reduce.metal
```diff
@@ -28,7 +28,8 @@
 
 #define instantiate_reduce_helper_64b(inst_f, name, op) \
   inst_f(name, int64, int64_t, op)                      \
-  inst_f(name, uint64, uint64_t, op)
+  inst_f(name, uint64, uint64_t, op)                    \
+  inst_f(name, complex64, complex64_t, op)
```
awni, line 32: Thanks for adding support for `complex64`! It might be good to add tests here (or in the fuzzer is fine too if that's your plan) since we haven't tested that yet..

mlx/backend/metal/kernels/utils.h
```diff
@@ -64,6 +64,14 @@ struct Limits<bool> {
   static constexpr constant bool min = false;
 };
 
+template <>
+struct Limits<complex64_t> {
+  static constexpr constant complex64_t max =
+      complex64_t(metal::numeric_limits<float>::infinity(), 0);
+  static constexpr constant complex64_t min =
+      complex64_t(-metal::numeric_limits<float>::infinity(), 0);
```
awni, line 72: Should those have `inf` and `-inf` in the complex spot as well?

mlx/backend/metal/kernels/utils.h
```diff
@@ -288,12 +324,110 @@ METAL_FUNC uint3 elem_to_loc_3_nd(
...
+      offset_t idx,
+      const constant int* shape,
+      const constant size_t* strides,
+      int ndim) {
+    (void)idx; // appease the compiler
+    (void)shape;
+    (void)strides;
+    (void)ndim;
+    return offset;
+  }
```
awni, line 400: Nit: you can do that more simply by leaving out the names, e.g.

```suggestion
  offset_t location(
      offset_t idx,
      const constant int*,
      const constant size_t*,
      int) {
    return offset;
  }
```

angeloskath: Awesome! Thanks, I hated having to write those `(void)` things.

angeloskath (approved):

PR #1301, opened 2024-07-31
fix rope and random

awni:
Fixes Random generation and RoPE for large arrays.

For RoPE:
- More work per thread to amortize the cost of computing indices in `size_t`
- Added a specialization which handles the inference case which is slightly faster than before o/w it would have been a bit slower (37ms using the general kernel). I'm not wedded to adding a new kernel esp since it might not matter too much..so feel free to push back on that.

Pre:

```
Timing rope_vec ... 0.32577 msec
Timing rope_mat ... 0.74966 msec
```

Post:

```
Timing rope_vec ... 0.31401 msec
Timing rope_mat ... 0.68538 msec
```

### DIFF
--- a/mlx/backend/metal/kernels/rope.metal
+++ b/mlx/backend/metal/kernels/rope.metal
@@ -25,8 +25,8 @@ template <typename T, bool traditional, bool forward>
   float sintheta = metal::fast::sin(theta);
 
   // Compute the input and output indices
-  size_t in_index_1, in_index_2;
-  size_t out_index_1, out_index_2;
+  uint in_index_1, in_index_2;
+  uint out_index_1, out_index_2;
   if (traditional) {
     out_index_1 = 2 * pos.x + pos.y * stride;
     out_index_2 = out_index_1 + 1;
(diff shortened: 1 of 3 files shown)


awni:
Not having specialized kernel does make a slight difference (106.8 -> 108.5 toks/s for 4-bit Mistral 7B on an M2 Ultra). Basically need it in order to not lose any time with the update RoPE which handles size_t.

mlx/backend/metal/kernels/rope.metal
```diff
@@ -6,43 +6,39 @@
...
+  float L = scale * static_cast<float>(offset);
   float d = static_cast<float>(pos.x) / static_cast<float>(grid.x);
 
   // Compute costheta, sintheta
   float theta = L * metal::exp2(-d * base);
   float costheta = metal::fast::cos(theta);
   float sintheta = metal::fast::sin(theta);
 
+  // Compute the input and output indices
+  size_t in_index_1, in_index_2;
```
angeloskath, line 28: No need for `size_t` here (in the single kernel) since it is a single dimension which cannot be bigger than INT_MAX.

angeloskath (approved):
Looks great.

I think you can remove the size_t from rope single but either way I like it. It is in the fast namespace, it better be fast 😊

mlx/backend/metal/kernels/rope.metal
awni: Oh good call. Even faster with `uint`: 0.305 msec

PR #1304, opened 2024-08-01
faster group norm

awni:
Use `mx.fast.layer_norm` to do a faster group norm. Also gets us higher precision reduction by default which is useful for inference.

Benchmarks on M2 Ultra:

`shape = (4, 720, 1280, 128)`

Pre: 42.69064903259277 ms
Post: 33.24143886566162 ms


barronalex (approved):
Nice!

Issue #1296, opened 2024-07-29
Cannot create tensor from raw bytes with dtype bfloat16


Narsil:
It is not possible at the moment create/manipulate tensors containing `bfloat16` outside of MLX.

```python
x = mx.array(jnp.ones((2, 2), dtype=jnp.bfloat16))
```

This is because it seems everything using `memoryview` and consorts fail on `bf16`.
This would be important in `safetensors` in order to implement in-memory loading (`weights = safetensors.mlx.load(f.read())` for instance) and also for lazy loading certain tensors.

`mx.load("file.safetensors")` works great.

As far as I can tell `memoryview` object loose the actual dtype anyway.

Any API that would get a `bytes` + shape + dtype would work super generically I feel (with or without copying depending on constraints).
This would allow me to correctly implement all supported dtypes on MLX within safetensors itself.
(And others to do advance stuff like loading files from network sockets directly)

Thanks a lot for this work.

Narsil:
Quick comment re-reading this:

The feature is not to fix the memory view for JAX -> MLX, but really for a way to create tensors from raw bytes instead. (The memoryview just show cases the issue why it's necessary, but we cannot expect jax/tf to be existant for this to work).

awni:
@Narsil I'm still not fully understanding what API you are looking for / what's missing? Right now you can create an array from a Python memoryview object which should be pretty flexible:

```python
a = np.array([1,2,3])

buffer = memoryview(a)
a_mx = mx.array(buffer)
```

Does that work for you or are you looking for something different? Maybe one thing that's missing is the lack of support for bfloat16?

zcbenz:
Closing since there had been a fix and we didn't get further responses.

PR #1305, opened 2024-08-01
Fix creating array from bf16 tensors in jax / torch

awni:
Addresses part of #1296 

This works now:

```python
import mlx.core as mx
import jax.numpy as jnp
a = jnp.array([1.0,2.0], dtype=jnp.bfloat16)
b = mx.array(a)
```


angeloskath (approved):
Awesome!

PR #1309, opened 2024-08-06
Add "edge" mode to mx.pad

### LINKED ISSUE #1284: [Feature Request] mx.pad supports the "edge" padding mode.
The function `pad` in [numpy](https://numpy.org/doc/stable/reference/generated/numpy.pad.html) supports "edge" padding mode which will pads with the edge values of array.

barronalex:
Resolves #1284

### DIFF
--- a/mlx/ops.h
+++ b/mlx/ops.h
@@ -256,6 +256,8 @@ array moveaxis(
     int destination,
     StreamOrDevice s = {});
 
+enum PaddingMode { Constant, Edge };
+
 /** Pad an array with a constant value */
 array pad(
     const array& a,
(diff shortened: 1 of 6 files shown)


mlx/ops.h
```diff
@@ -256,30 +256,36 @@ array moveaxis(
     int destination,
     StreamOrDevice s = {});
 
+enum PaddingMode { Constant, Edge };
```
awni, line 259: Maybe use an `enum class` here instead of a `enum`

awni: On the other hand we tend to match types from Python API to C++ API (I'm not sure that's the best choice tbh), but for the sake of consistency maybe simpler to have it be a string at the `ops.h` level and scope the enum to the Primitive itself.

I think (?) it also makes it easier for code binding ops.h (i.e. MLX C) to not have to add the new enum type.

E.g. like `indexing` for `meshgrid`, `ord` for `norm`.

barronalex: Sounds good! I changed it to just be a string.

awni (approved):
Looks great to me, thanks for adding that!

PR #1310, opened 2024-08-07
Add gemv masked to JIT plus some fixes

awni:
- Added masked gemvs to the JIT, reduces the library size from 5.9 to 1.7M
- Fixed some issues with JITing the new large vectorized kernels
- Some random code reorgs with a small attempt to reduce CPU binary footprint


angeloskath (approved):
Looks great!

awni:
Currently the MLX light build command (e.g .without the CPU back-end) gives you a C++ binary of 2.8MB and a MTL Lib of 1.7 MB (all in 4.5MB).

If you include the CPU back-end the C++ binary goes up to 8. MB, so all in 9.7 MB.

PR #1314, opened 2024-08-08
Revert variance to be numerically stable

### LINKED ISSUE #1302: [BUG] mx.std returns NaN
**Describe the bug**
`mx.std()` returns NaN instead of a value.

**To Reproduce**
Run the snippet and check value of `x_std`:

```python
import os
os.environ['ENABLE_PJRT_COMPATIBILITY'] = '1'
import mlx.core as mx

x = mx.array([ -0.8212978, -0.8214609 ])

x_std = mx.std(x) # NaN !

print(x_std)
```

**Expected behavior**
Should return value similar to numpy, i.e. `8.155000000004131e-05`.

 - OS Version: MacOS 14.4.1
 - Version 0.16.1
 - M3 Max

**Additional context**
```python
import mlx.core as mx
import numpy as np

# MLX version
print("MLX version:", mx.__version__)

# MLX calculation
x_mlx = mx.array([-0.8212978, -0.8214609])
x_std_mlx = mx.std(x_mlx)
print("MLX std:", x_std_mlx)

# Numpy calculation for comparison
x_np = np.array([-0.8212978, -0.8214609])
x_std_np = np.std(x_np)
print("Numpy std:", x_std_np)

# Manual calculation using MLX
mean = mx.mean(x_mlx)
variance = mx.mean((x_mlx - mean)**2)
x_std_manual = mx.sqrt(variance)
print("MLX manual std:", x_std_manual)

# Check if the result is very small instead of NaN
epsilon = 1e-10
is_small = mx.abs(x_std_manual) < epsilon
print("Is result very small?", is_small)
```
will print:
```
MLX version: 0.16.1
MLX std: array(nan, dtype=float32)
Numpy std: 8.155000000004131e-05
MLX manual std: array(8.15392e-05, dtype=float32)
Is result very small? array(False, dtype=bool)
```

angeloskath:
Closes #1302 .

Fixes the numerical stability of variance. When compiled it has the same speed as the previous "fast" numerically unstable version.

### DIFF
--- a/mlx/ops.cpp
+++ b/mlx/ops.cpp
@@ -1640,17 +1640,21 @@ array var(
     int ddof /* = 0*/,
     StreamOrDevice s /* = {}*/) {
   auto dtype = at_least_float(a.dtype());
-  auto mu2 = square(mean(a, axes, keepdims, s), s);
-  auto a2 = mean(square(a, s), axes, keepdims, s);
-  auto v = subtract(a2, mu2, s);
+  auto mu = mean(a, axes, /* keepdims= */ true, s);
+  auto v = sum(square(subtract(a, mu, s), s), axes, keepdims, s);
 
   if (ddof != 0) {
-    auto nelements = number_of_elements(a, axes, false, dtype, s);
-    auto factor = divide(
-        nelements,
-        maximum(subtract(nelements, array(ddof, dtype), s), array(0, dtype), s),
+    auto normalizer = maximum(
+        subtract(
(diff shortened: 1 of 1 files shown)


awni (approved):
Thanks!!

angeloskath:
I am not sure what is up with the test tolerance. I will merge this and if it fails on `main` I will do another PR bumping the tolerance. I can't make it to fail locally so I guess we were unlucky somehow...

Issue #1321, opened 2024-08-12
[BUG] Dropout not preserving `dtype`


vgoklani:
```python

d = nn.Dropout(p=0.1)

x = mx.array([1, 2, 3, 4, 5], dtype=mx.bfloat16)

assert d(x).dtype == mx.bfloat16 # fails

```
The output of the dropout function should preserve the `dtype` of the input array

bhargavyagnik:
I experimented two approaches for
 https://github.com/ml-explore/mlx/blob/1086dc4db070d4f7d211f2bb760f560b98f969f7/python/mlx/nn/layers/dropout.py#L35

- `return (1 / self._p_1) * mask * x  # Return dtype is mx.float32` 

but when I replace that with 

- `return mask * x * ( 1 / self._p_1) # Return dtype is mx.bfloat16`

 
This problem is related to type promotion and similar usecase exist in Numpy and numerous libraries.  


Since python executes from left to right; So in the first case, when weakly typed `float` (1 / self._p_1) is multiplied with `mx.bool_`, (mask) the resulting dtype is `mx.float32` and thus entire equation is calculated in `mx.float32`.

When we first multiply `mx.bool_` with `mx.bfloat16`, the weakly typed `float` is promoted to just `mx.bfloat16`thus preserving the dtype

vgoklani:
```python
class Dropout(nn.Module):
    def __init__(self, p: Optional[float] = 0.5):
        super().__init__()
        self.p = p

    def __call__(self, x: mx.array, key: Optional[mx.array] = None):
        if not self.training or self.p == 0:
            return x
        else:
            mask = mx.random.uniform(shape=x.shape, dtype=x.dtype, key=key) > self.p
            return mask * x / (1.0 - self.p)
```

vgoklani:
Just a heads up, the `tree_flaten` method drops the Dropout layers from the network, so when we serialized the optimizer:

`mx.save_safetensors(file=optimizer_state_path, arrays=dict(optimizer_state))`

and decided to continue training from a previously saved checkpoint, we got a bunch of key errors...

awni:
I'm not following. The `Dropout` layers shouldn't be saved in `tree_flatten` since they don't have parameters. Could you share some code which reproduces the errors you are getting?

vgoklani:
@awni created a separate issue for this: https://github.com/ml-explore/mlx/issues/1328

PR #1323, opened 2024-08-13
Fix: Preserve input dtype in Dropout layer output

bhargavyagnik:
- Modified Dropout implementation to ensure that the output dtype matches the input dtype.
- This resolves the issue #1321 

## Proposed changes

I re-ordered the equation that calculated Dropout to preserve the input dtype. The underlying problem in Issue #1321 mentioned https://github.com/ml-explore/mlx/issues/1321#issuecomment-2285606325

## Checklist

### DIFF
--- a/python/mlx/nn/layers/dropout.py
+++ b/python/mlx/nn/layers/dropout.py
@@ -32,7 +32,7 @@ class Dropout(Module):
 
         mask = mx.random.bernoulli(self._p_1, x.shape)
 
-        return (1 / self._p_1) * mask * x
+        return mask * x / self._p_1
 
 
 class Dropout2d(Module):
@@ -85,7 +85,7 @@ class Dropout2d(Module):
         mask_shape[-2] = mask_shape[-3] = 1
 
         mask = mx.random.bernoulli(p=self._p_1, shape=mask_shape)
-        return (1 / self._p_1) * mask * x
+        return mask * x / self._p_1
 
 
 class Dropout3d(Module):
@@ -134,4 +134,4 @@ class Dropout3d(Module):
         mask_shape[-2] = mask_shape[-3] = mask_shape[-4] = 1
 
         mask = mx.random.bernoulli(p=self._p_1, shape=mask_shape)
-        return (1 / self._p_1) * mask * x
+        return mask * x / self._p_1
--- a/python/mlx/nn/layers/dropout.py
+++ b/python/mlx/nn/layers/dropout.py
@@ -32,7 +32,7 @@ class Dropout(Module):
 
         mask = mx.random.bernoulli(self._p_1, x.shape)
 
-        return mask * x / self._p_1
+        return (mask * x) / self._p_1
 
 
 class Dropout2d(Module):
@@ -85,7 +85,7 @@ class Dropout2d(Module):
         mask_shape[-2] = mask_shape[-3] = 1
 
         mask = mx.random.bernoulli(p=self._p_1, shape=mask_shape)
-        return mask * x / self._p_1
+        return (mask * x) / self._p_1
 
 
 class Dropout3d(Module):
@@ -134,4 +134,4 @@ class Dropout3d(Module):
         mask_shape[-2] = mask_shape[-3] = mask_shape[-4] = 1
 
         mask = mx.random.bernoulli(p=self._p_1, shape=mask_shape)
-        return mask * x / self._p_1
+        return (mask * x) / self._p_1
--- a/python/mlx/nn/layers/dropout.py
+++ b/python/mlx/nn/layers/dropout.py
@@ -32,7 +32,7 @@ class Dropout(Module):
 
         mask = mx.random.bernoulli(self._p_1, x.shape)
 
-        return (mask * x) / self._p_1
+        return (mask * x) * (1 / self._p_1)
 
 
 class Dropout2d(Module):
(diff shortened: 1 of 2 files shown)


awni:
Thanks for the fix! Two comments:

- Could you add a test to https://github.com/ml-explore/mlx/blob/main/python/tests/test_nn.py#L962
- I would use parenthesis to be extra explicit here. The python order makes sense but it might not be as obvious to people reading the code that the order shouldn't be changed.

bhargavyagnik:
Hi @awni , I found a serious bug in the complete testing directory.

https://github.com/ml-explore/mlx/blob/1086dc4db070d4f7d211f2bb760f560b98f969f7/python/tests/test_nn.py#L1002

While I had already tested before submission, I realized something is wrong and thus read the mlx_test.py code.
Here you are not overriding the assertTrue in any other form.

Thus many of the test cases written in python/tests with `self.assertTrue`  like 

https://github.com/ml-explore/mlx/blob/1086dc4db070d4f7d211f2bb760f560b98f969f7/python/tests/test_nn.py#L1002

https://github.com/ml-explore/mlx/blob/1086dc4db070d4f7d211f2bb760f560b98f969f7/python/tests/test_nn.py#L960

assertTrue(x,y) would treat x as expression and y as message.

I think the correct function to use is assertEqual or we can use assertTrue( x==y) .

If you agree, I can open another PR regarding these changes.

awni:
> I think the correct function to use is assertEqual or we can use assertTrue( x==y) .

Those are indeed bugs. The right function is `assertEqual`. Would be great if you could send a PR to fix it.

bhargavyagnik:
Yeah, I'll add them in this PR itself. There were less than I anticipated.

bhargavyagnik:
I added the changes, let me know if you have any suggestions. Thanks for the review !!

python/mlx/nn/layers/dropout.py
```diff
@@ -32,7 +32,7 @@ def __call__(self, x):
 
         mask = mx.random.bernoulli(self._p_1, x.shape)
 
-        return (1 / self._p_1) * mask * x
+        return (mask * x) / self._p_1
```
awni, line 35: Sorry I missed this. You should still do the scaling with a multiplication by `(1 / self._p1)` which will be a single divide followed by a bunch of multiplies as opposed to a bunch of divides. Divide is more expensive than multiply.

python/tests/test_nn.py
```diff
@@ -939,73 +939,73 @@ def test_rope(self):
...
-        self.assertTrue(y.dtype, mx.float32)
+        self.assertEqual(y.shape, x.shape)
+        self.assertEqual(y.dtype, mx.float32)
 
         x = mx.ones((2, 4, 4, 4, 4), dtype=mx.bfloat16)
         y = nn.Dropout3d(0.5)(x)
-        self.assertTrue(y.shape, x.shape)
-        self.assertTrue(y.dtype, mx.bfloat16)
+        self.assertEqual(y.shape, x.shape)
+        self.assertEqual(y.dtype, mx.bfloat16)
```
awni, line 1003: Thank you for fixing those!!

python/mlx/nn/layers/dropout.py
bhargavyagnik: Oh, okay I'll make that change

awni (approved):
Looks great. Will merge when the tests clear.

python/mlx/nn/layers/dropout.py
bhargavyagnik: Hi @awni 
Sorry, I'm still learning so please excuse me if this is out of topic and a very small impact question but ..
```python
import mlx.core as mx
import timeit

x = mx.random.uniform(shape=(100000, 1000))
y = mx.random.uniform(shape=(100000, 1000))

x = x * 124.215**6
y = y * 135.353**6
def direct_division():
    return mx.eval(x/y)

def multiply_reciprocal():
    return mx.eval(x * (1/y))

print("Direct division:", timeit.timeit(direct_division, number=1000))
print("Multiply reciprocal:", timeit.timeit(multiply_reciprocal, number=1000))

```

The output is  :
Direct division: 6.718534375000672
Multiply reciprocal: 10.876147124999989

I read some articles over https://stackoverflow.com/questions/1117674/why-is-multiplying-cheaper-than-dividing?rq=3, https://stackoverflow.com/questions/15745819/why-is-division-more-expensive-than-multiplication

I understood that division is more expensive than multiplication, but when I observe the performance of `direct_division`, it is faster than `multiply_reciprocal`..
I also experimented by multiplying x and y with 1.7976931348623157e+308 but still seems direct_division is faster. What could be the underlying reason?

awni: No problem. In your example `y` is an array so you are dividing a bunch of stuff in both cases. So the second one is strictly more work than the first (it includes the first in fact).

In the Dropout example `p` is a scalar. So a benchmark would be:

```
import mlx.core as mx
import timeit

x = mx.random.uniform(shape=(100000, 1000))
y = 5.0

def direct_division():
    return mx.eval(x/y)

def multiply_reciprocal():
    return mx.eval(x * (1/y))

print("Direct division:", timeit.timeit(direct_division, number=1000))
print("Multiply reciprocal:", timeit.timeit(multiply_reciprocal, number=1000))
```

Also you probably still won't see a time difference because both operations are memory bandwidth bound. But if you measure power use (which isn't so easy to do reliably) you would might see lower power consumption for the second. It could be marginal.. but in this case why not since it's trivial to use the slightly more efficient version.

bhargavyagnik: Yeah, i see. I was just exploring so thought to test it out. Thank you so much for the quick reply!

Issue #1238, opened 2024-06-27
[BUG] mx.linalg.inv crash


petertsoi:
**Describe the bug**
When trying to invert a small 3x3 (camera intrinsics matrix), mlx crashes.

**To Reproduce**
```
import mlx.core as mx
intrinsics = mx.array([[1162.38, 0, 618.25], [0, 1156.83, 410.75], [0, 0, 1]])
mx.linalg.inv(intrinsics)

libc++abi: terminating due to uncaught exception of type std::runtime_error: [Inverse::eval_gpu] Metal inversion NYI.
zsh: abort      python
```

**Expected behavior**
Works properly in numpy

**Desktop (please complete the following information):**
 - macOS 14.4.1
 - mlx:  0.15.1
 - python: 3.11.9

petertsoi:
Also not working in mlx-swift, which is where I'm using it from but reproduced in python so filed it here. The CPU backend appears to work though.

awni:
Yes, this isn't a bug, the GPU back-end is not yet implemented. It's most likely going to take some time before we have GPU support for matrix inversion. I changed this to be a feature req rather than a bug, and we can leave the issue open.

awni:
My recommendation is to use the CPU for now. You can do something like:

```python
out = mx.llinalg.inv(x, stream=mx.cpu)
```

Just for that operation.

angeloskath:
By the way if all you want to do is 3x3 matrix inversion it is way faster to write it explicitly and compile it with mlx. The inversion would be as simple as the following:

```python
import mlx.core as mx


@partial(mx.compile, shapeless=True)
def _inverse_3x3(a11, a12, a13, a21, a22, a23, a31, a32, a33):
    det = (
        a11 * a22 * a33
        + a12 * a23 * a31
        + a13 * a21 * a32
        - a11 * a23 * a32
        - a12 * a21 * a33
        - a13 * a22 * a31
    )
    c11 = (a22 * a33 - a23 * a32) / det
    c12 = (a13 * a32 - a12 * a33) / det
    c13 = (a12 * a23 - a13 * a22) / det
    c21 = (a23 * a31 - a21 * a33) / det
    c22 = (a11 * a33 - a13 * a31) / det
    c23 = (a13 * a21 - a11 * a23) / det
    c31 = (a21 * a32 - a22 * a31) / det
    c32 = (a12 * a31 - a11 * a32) / det
    c33 = (a11 * a22 - a12 * a21) / det
[...]

PR #1325, opened 2024-08-13
Custom Metal Kernels from Python

barronalex:
This PR let's you define custom Metal kernels through the C++/Python API.

The user provides the body of the Metal kernel and then we use the shape/dtype of the arrays it's called with together with provided `output_shapes`/`output_dtypes` to determine what the function signature should be.

Metal attributes such as `[[thread_position_in_grid]]` are automatically added to the function signature if present in the body.

For example:
```python
source = """
    uint elem = thread_position_in_grid.x;
    T tmp = inp[elem];
    out[elem] = metal::exp(tmp);
"""
A = mx.random.normal(shape=(4, 16)).astype(mx.float16)
    name="myexp",
    source=source
)
out = kernel(
    inputs={"inp": A},
    template={"T": mx.float32},
    grid=(A.size, 1, 1),
    threadgroup=(256, 1, 1),
    output_shapes={"out": A.shape},
    output_dtypes={"out": A.dtype},
    verbose=True,  # prints the full generated kernel
)
```
Generates:
```cpp
template <typename T>
[[kernel]] void custom_kernel_myexp(
  const device float16_t* inp [[buffer(0)]],
  const constant int* inp_shape [[buffer(1)]],
  const constant size_t* inp_strides [[buffer(2)]],
  const constant int& inp_ndim [[buffer(3)]],
  device float16_t* out [[buffer(4)]],
  uint3 thread_position_in_grid [[thread_position_in_grid]]) {

    uint elem = thread_position_in_grid.x;
    T tmp = inp[elem];
    out[elem] = metal::exp(tmp);

}

template [[host_name("custom_kernel_myexp_float_")]] [[kernel]] decltype(custom_kernel_myexp<float>) custom_kernel_myexp<float>;
```

@angeloskath's nice 3x3 inversion example from #1238 becomes:
```python
def invert_3x3(A):
    source = """
        int elem = thread_position_in_grid.x;
        int index = elem * 9;
        T a11 = a[index];
        T a12 = a[index + 1];
        T a13 = a[index + 2];
        T a21 = a[index + 3];
        T a22 = a[index + 4];
        T a23 = a[index + 5];
        T a31 = a[index + 6];
        T a32 = a[index + 7];
        T a33 = a[index + 8];
        T det = (
            a11 * a22 * a33
            + a12 * a23 * a31
            + a13 * a21 * a32
            - a11 * a23 * a32
            - a12 * a21 * a33
            - a13 * a22 * a31
        );
        out[index] = (a22 * a33 - a23 * a32) / det;
        out[index + 1] = (a13 * a32 - a12 * a33) / det;
        out[index + 2] = (a12 * a23 - a13 * a22) / det;
        out[index + 3] = (a23 * a31 - a21 * a33) / det;
        out[index + 4] = (a11 * a33 - a13 * a31) / det;
        out[index + 5] = (a13 * a21 - a11 * a23) / det;
        out[index + 6] = (a21 * a32 - a22 * a31) / det;
        out[index + 7] = (a12 * a31 - a11 * a32) / det;
        out[index + 8] = (a11 * a22 - a12 * a21) / det;
    """
        name="invert3x3",
        source=source,
   )
   out = kernel(
        inputs={"a": A},
        template={"T": A.dtype},
        output_shapes={"out": A.shape},
        output_dtypes={"out": A.dtype},
        grid=(A.size // 9, 1, 1),
        threadgroup=(256, 1, 1),
        verbose=True,
    )
    return out["out"]

A = mx.random.normal(shape=(16, 3, 3))
A = mx.matmul(A, mx.swapaxes(A, -1, -2))
A_inv = mx.linalg.inv(A, stream=mx.cpu)
A_inv_metal = invert_3x3(A)
mx.allclose(A_inv, A_inv_metal, atol=1e-4)
```

Fusing the `mx.split`/`mx.concatenate` get's you a nice speed up, particularly for large batch sizes:
inv

### DIFF
--- a/docs/src/dev/custom_metal_kernels.rst
+++ b/docs/src/dev/custom_metal_kernels.rst
@@ -47,7 +47,7 @@ The full function signature will be generated using:
     so we add ``device float16_t* out``.
 * Template parameters passed using ``template``
     In the above, ``template={"T": mx.float32}`` adds a template of ``template <typename T>`` to the function
-    and instantiates the template with ``myexp_(hash)<float>``.
+    and instantiates the template with ``myexp_float<float>``.
     Template parameters can be ``mx.core.Dtype``, ``int`` or ``bool``.
 * Metal attributes used in ``source`` such as ``[[thread_position_in_grid]]``
     These will be added as function arguments.
@@ -58,12 +58,9 @@ Putting this all together, the generated function signature for ``myexp`` is as 
 .. code-block:: cpp
 
   template <typename T>
-  [[kernel]] void myexp_7406405795239204910(
+  [[kernel]] void myexp_float(
     const device float16_t* inp [[buffer(0)]],
-    const constant int* inp_shape [[buffer(1)]],
-    const constant size_t* inp_strides [[buffer(2)]],
-    const constant int& inp_ndim [[buffer(3)]],
-    device float16_t* out [[buffer(4)]],
+    device float16_t* out [[buffer(1)]],
     uint3 thread_position_in_grid [[thread_position_in_grid]]) {
 
           uint elem = thread_position_in_grid.x;
@@ -72,7 +69,7 @@ Putting this all together, the generated function signature for ``myexp`` is as 
 
   }
 
-  template [[host_name("myexp_7406405795239204910")]] [[kernel]] decltype(myexp_7406405795239204910<float>) myexp_7406405795239204910<float>;
+  template [[host_name("myexp_float")]] [[kernel]] decltype(myexp_float<float>) myexp_float<float>;
 
 You can print the generated code for a ``mx.fast.metal_kernel`` by passing ``verbose=True``.
 
@@ -85,7 +82,7 @@ Generally this makes writing the kernel easier, since we don't have to worry abo
 when indexing.
 
 If we want to avoid this copy, ``metal_kernel`` automatically passes ``a_shape``, ``a_strides`` and ``a_ndim`` for each
-input array ``a``.
+input array ``a`` if any are present in ``source``.
 We can then use MLX's built in indexing utils to fetch the right elements for each thread.
 
 Let's convert ``myexp`` above to support arbitrarily strided arrays without relying on a copy from ``ensure_row_contiguous``:
--- a/docs/src/dev/custom_metal_kernels.rst
+++ b/docs/src/dev/custom_metal_kernels.rst
@@ -17,9 +17,11 @@ Let's write a custom kernel that computes ``exp`` elementwise:
           out[elem] = metal::exp(tmp);
       """
 
-      outputs = mx.fast.metal_kernel(
+      kernel = mx.fast.metal_kernel(
           name="myexp",
           source=source,
+      )
+      outputs = kernel(
           inputs={"inp": a},
           template={"T": mx.float32},
           grid=(a.size, 1, 1),
@@ -71,7 +73,7 @@ Putting this all together, the generated function signature for ``myexp`` is as 
 
   template [[host_name("custom_kernel_myexp_float")]] [[kernel]] decltype(custom_kernel_myexp_float<float>) custom_kernel_myexp_float<float>;
 
-You can print the generated code for a ``mx.fast.metal_kernel`` by passing ``verbose=True``.
+You can print the generated code for a ``mx.fast.metal_kernel`` by passing ``verbose=True`` when you call it.
 
 Using Shape/Strides
 -------------------
@@ -99,9 +101,11 @@ Let's convert ``myexp`` above to support arbitrarily strided arrays without rely
           out[elem] = metal::exp(tmp);
       """
 
-      outputs = mx.fast.metal_kernel(
+      kernel = mx.fast.metal_kernel(
           name="myexp_strided",
-          source=source,
+          source=source
+      )
+      outputs = kernel(
           inputs={"inp": a},
           template={"T": mx.float32},
           grid=(a.size, 1, 1),
--- a/docs/src/dev/custom_metal_kernels.rst
+++ b/docs/src/dev/custom_metal_kernels.rst
@@ -53,7 +53,7 @@ The full function signature will be generated using:
     Template parameters can be ``mx.core.Dtype``, ``int`` or ``bool``.
 * Metal attributes used in ``source`` such as ``[[thread_position_in_grid]]``
     These will be added as function arguments.
-    All the attributes defined in Table 5.8 of the Metal Shading Language Specification are supported.
+    All the attributes defined in Table 5.8 of the `Metal Shading Language Specification <https://developer.apple.com/metal/Metal-Shading-Language-Specification.pdf>`_ are supported.
 
 Putting this all together, the generated function signature for ``myexp`` is as follows:
 
--- a/docs/src/dev/custom_metal_kernels.rst
+++ b/docs/src/dev/custom_metal_kernels.rst
@@ -1,5 +1,3 @@
-.. _custom_metal_kernels:
-
 Custom Metal Kernels
 ====================
 
@@ -19,16 +17,16 @@ Let's write a custom kernel that computes ``exp`` elementwise:
           out[elem] = metal::exp(tmp);
       """
 
-      kernel = mx.fast.MetalKernel(
+      outputs = mx.fast.metal_kernel(
           name="myexp",
           source=source,
+          inputs={"inp": a},
+          template={"T": mx.float32},
           grid=(a.size, 1, 1),
           threadgroup=(256, 1, 1),
           output_shapes={"out": a.shape},
           output_dtypes={"out": a.dtype},
       )
-      kernel.template(T=mx.float32)
-      outputs = kernel(inp=a)
       return outputs["out"]
 
   a = mx.random.normal(shape=(4, 16)).astype(mx.float16)
@@ -45,8 +43,8 @@ We are only required to pass the body of the Metal kernel in ``source``.
 * The keys and values of ``output_shapes`` and ``output_dtypes``.
     In the above, ``out`` is an ``mx.float16`` ``mx.array``
     so we add ``device float16_t* out [[buffer(3)]]``.
-* Template parameters passed using ``MetalKernel.template()``.
-    In the above, ``kernel.template(T=mx.float32)`` adds a template of ``template <typename T>`` to the function
+* Template parameters passed using ``template``.
+    In the above, ``template={"T": mx.float32}`` adds a template of ``template <typename T>`` to the function
     and instantiates the template with ``custom_kernel_myexp<float>``.
     Template parameters can be ``mx.core.Dtype``, ``int`` or ``bool``.
 * Metal attributes used in ``source`` such as ``[[thread_position_in_grid]]`` or ``[[simdgroup_index_in_threadgroup]]``.
@@ -104,17 +102,17 @@ Let's convert ``myexp`` above to support arbitrarily strided arrays without rely
           out[elem] = metal::exp(tmp);
       """
 
-      kernel = mx.fast.MetalKernel(
+      outputs = mx.fast.metal_kernel(
           name="myexp_strided",
           source=source,
+          inputs={"inp": a},
+          template={"T": mx.float32},
           grid=(a.size, 1, 1),
           threadgroup=(256, 1, 1),
           output_shapes={"out": a.shape},
           output_dtypes={"out": a.dtype},
           ensure_row_contiguous=False,
       )
-      kernel.template(T=mx.float32)
-      outputs = kernel(inp=a)
       return outputs["out"]
 
   a = mx.random.normal(shape=(4, 16)).astype(mx.float16)
--- a/docs/src/dev/custom_metal_kernels.rst
+++ b/docs/src/dev/custom_metal_kernels.rst
@@ -47,7 +47,7 @@ The full function signature will be generated using:
     so we add ``device float16_t* out``.
 * Template parameters passed using ``template``
     In the above, ``template={"T": mx.float32}`` adds a template of ``template <typename T>`` to the function
-    and instantiates the template with ``myexp_float<float>``.
+    and instantiates the template with ``custom_kernel_myexp_float<float>``.
     Template parameters can be ``mx.core.Dtype``, ``int`` or ``bool``.
 * Metal attributes used in ``source`` such as ``[[thread_position_in_grid]]``
     These will be added as function arguments.
@@ -58,7 +58,7 @@ Putting this all together, the generated function signature for ``myexp`` is as 
 .. code-block:: cpp
 
   template <typename T>
-  [[kernel]] void myexp_float(
+  [[kernel]] void custom_kernel_myexp_float(
     const device float16_t* inp [[buffer(0)]],
     device float16_t* out [[buffer(1)]],
     uint3 thread_position_in_grid [[thread_position_in_grid]]) {
@@ -69,7 +69,7 @@ Putting this all together, the generated function signature for ``myexp`` is as 
 
   }
 
-  template [[host_name("myexp_float")]] [[kernel]] decltype(myexp_float<float>) myexp_float<float>;
+  template [[host_name("custom_kernel_myexp_float")]] [[kernel]] decltype(custom_kernel_myexp_float<float>) custom_kernel_myexp_float<float>;
 
 You can print the generated code for a ``mx.fast.metal_kernel`` by passing ``verbose=True``.
 
--- a/docs/src/dev/custom_metal_kernels.rst
+++ b/docs/src/dev/custom_metal_kernels.rst
@@ -8,7 +8,7 @@ Simple Example
 
 Let's write a custom kernel that computes ``exp`` elementwise:
 
-.. code_block:: python
+.. code-block:: python
 
   def exp_elementwise(a: mx.array):
       source = """
@@ -33,27 +33,29 @@ Let's write a custom kernel that computes ``exp`` elementwise:
   b = exp_elementwise(a)
   assert mx.allclose(b, mx.exp(a))
 
-We are only required to pass the body of the Metal kernel in ``source``.
-``mx.fast.MetalKernel`` will generate the full function signature using:
+.. note::
+    We are only required to pass the body of the Metal kernel in ``source``.
 
-* The keys and shapes/dtypes of the arrays passed to ``MetalKernel.__call__()``.
-    In the above, ``a`` is an ``mx.float16`` ``mx.array`` and we pass it with the key ``inp``
-    so we will add ``const device float16_t* inp [[buffer(0)]]`` to the signature.
-    It will also add `inp_shape`, `inp_strides` and `inp_ndim` for convenience.
-* The keys and values of ``output_shapes`` and ``output_dtypes``.
-    In the above, ``out`` is an ``mx.float16`` ``mx.array``
-    so we add ``device float16_t* out [[buffer(3)]]``.
-* Template parameters passed using ``template``.
+The full function signature will be generated using:
+
+* The keys and shapes/dtypes of ``inputs``
+    In the above, ``a`` is an ``mx.array`` of type ``mx.float16`` and we pass it with the key ``inp``
+    so we will add ``const device float16_t* inp`` to the signature.
+    ``inp_shape``, ``inp_strides`` and ``inp_ndim`` are also added for convenience.
+* The keys and values of ``output_shapes`` and ``output_dtypes``
+    In the above, ``out`` is an ``mx.array`` of type ``mx.float16``
+    so we add ``device float16_t* out``.
+* Template parameters passed using ``template``
     In the above, ``template={"T": mx.float32}`` adds a template of ``template <typename T>`` to the function
-    and instantiates the template with ``custom_kernel_myexp<float>``.
+    and instantiates the template with ``myexp_(hash)<float>``.
     Template parameters can be ``mx.core.Dtype``, ``int`` or ``bool``.
-* Metal attributes used in ``source`` such as ``[[thread_position_in_grid]]`` or ``[[simdgroup_index_in_threadgroup]]``.
+* Metal attributes used in ``source`` such as ``[[thread_position_in_grid]]``
     These will be added as function arguments.
     All the attributes defined in Table 5.8 of the Metal Shading Language Specification are supported.
 
 Putting this all together, the generated function signature for ``myexp`` is as follows:
 
-.. code_block:: cpp
+.. code-block:: cpp
 
   template <typename T>
   [[kernel]] void myexp_7406405795239204910(
@@ -72,25 +74,23 @@ Putting this all together, the generated function signature for ``myexp`` is as 
 
   template [[host_name("myexp_7406405795239204910")]] [[kernel]] decltype(myexp_7406405795239204910<float>) myexp_7406405795239204910<float>;
 
-You can print the generated code for a ``mx.fast.MetalKernel`` by passing ``verbose=True`` when you create it.
-
-A hash of the source code and template arguments is added to function name above to prevent cache invalidation.
+You can print the generated code for a ``mx.fast.metal_kernel`` by passing ``verbose=True``.
 
 Using Shape/Strides
 -------------------
 
-``mx.fast.MetalKernel`` supports an initialization argument ``ensure_row_contiguous`` which is ``True`` by default.
+``mx.fast.metal_kernel`` supports an argument ``ensure_row_contiguous`` which is ``True`` by default.
 This will copy the ``mx.array`` inputs if needed before the kernel is launched to ensure that the memory layout is row contiguous.
 Generally this makes writing the kernel easier, since we don't have to worry about gaps or the ordering of the dims
 when indexing.
 
-If we want to avoid this copy, ``MetalKernel`` automatically passes ``a_shape``, ``a_strides`` and ``a_ndim`` for each
-input array ``a`` (as shown in the generated function signature above).
-With these we can use MLX's built in indexing utils to fetch the right elements for each thread.
+If we want to avoid this copy, ``metal_kernel`` automatically passes ``a_shape``, ``a_strides`` and ``a_ndim`` for each
+input array ``a``.
+We can then use MLX's built in indexing utils to fetch the right elements for each thread.
 
 Let's convert ``myexp`` above to support arbitrarily strided arrays without relying on a copy from ``ensure_row_contiguous``:
 
-.. code_block:: python
+.. code-block:: python
 
   def exp_elementwise(a: mx.array):
       source = """
@@ -116,7 +116,7 @@ Let's convert ``myexp`` above to support arbitrarily strided arrays without rely
       return outputs["out"]
 
   a = mx.random.normal(shape=(4, 16)).astype(mx.float16)
-  # Make a non-contiguous
+  # make non-contiguous
   a = a[::2]
   b = exp_elementwise(a)
   assert mx.allclose(b, mx.exp(a))
--- a/docs/src/usage/custom_metal_kernels.rst
+++ b/docs/src/usage/custom_metal_kernels.rst
@@ -1,0 +1,124 @@
+.. _custom_metal_kernels:
+
+Custom Metal Kernels
+====================
+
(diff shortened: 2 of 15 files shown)


mzy2240:
Wow this is fantastic! Just wondering if this PR got approved and merged, how hard would it be to support [Accelerate](https://developer.apple.com/documentation/accelerate?language=objc) in a similar fashion? I can see a HUGE potential of this PR!

python/src/array.cpp
```diff
@@ -247,6 +247,15 @@ void init_array(nb::module_& m) {
           Returns:
             tuple(int): A tuple containing the sizes of each dimension.
         )pbdoc")
+      .def_prop_ro(
```
angeloskath, line 250: That's probably not what you had in mind. Since everything in mlx is lazy there will be no strides available before the array is computed so this will be wrong unfortunately.

We 've had the discussion about including strides many times I think we should keep them unavailable from the public API. They are only useful inside an op implementation.

I would instead allow access to `<name of array argument>_[shape|strides|ndim]` any of which would be automatically included in the arguments if present in the kernel source.

mlx/backend/metal/custom_kernel.cpp
```diff
@@ -0,0 +1,65 @@
...
+    std::vector<array>& outputs) {
+  auto& s = stream();
+
+  for (auto& out : outputs) {
+    out.set_data(allocator::malloc_or_wait(out.nbytes()));
+  }
+
+  std::vector<array> copies;
+
+  auto check_input = [&copies, &s](const array& x) {
```
angeloskath, line 19:
```suggestion
  auto check_input = [&copies, &s](const array& x) -> const array& {
```

mlx/backend/metal/custom_kernel.cpp
```diff
@@ -0,0 +1,65 @@
...
+    bool no_copy = x.flags().row_contiguous;
+    if (no_copy) {
```
angeloskath, line 21:
```suggestion
    if (!ensure_row_contiguous_ || no_copy) {
```

mlx/backend/metal/custom_kernel.cpp
```diff
@@ -0,0 +1,65 @@
...
+  compute_encoder->setComputePipelineState(kernel);
+  int index = 0;
+  for (int i = 0; i < inputs.size(); i++) {
+    array in = inputs[i];
+    if (ensure_row_contiguous_) {
+      in = check_input(in);
+    }
+    compute_encoder.set_input_array(in, index);
+    index++;
+  }
```
angeloskath, line 47:
```cpp
int index = 0;
for (const array & in : inputs) {
  compute_encoder.set_input_array(check_input(in), index++);
}
```

mlx/backend/metal/custom_kernel.cpp
```diff
@@ -0,0 +1,65 @@
...
+
+  const auto [tx, ty, tz] = threadgroup_;
+  MTL::Size group_dims = MTL::Size(tx, ty, tz);
+  const auto [gx, gy, gz] = grid_;
+  MTL::Size grid_dims = MTL::Size(gx, gy, gz);
+  compute_encoder->dispatchThreads(grid_dims, group_dims);
+
+  d.get_command_buffer(s.index)->addCompletedHandler(
```
angeloskath, line 59:
```cpp
if (!copies.empty()) {
  ...
}
```

mlx/fast.cpp
```diff
@@ -1,8 +1,14 @@
 // Copyright © 2023-2024 Apple Inc.
 
+#include <algorithm>
 #include <cassert>
+#include <iostream>
```
angeloskath, line 5: Forgotten? I tend to do that a lot :-)

mlx/fast.cpp
```diff
@@ -859,4 +865,214 @@ array affine_dequantize(
...
+
+  int index = 0;
+  constexpr int max_constant_array_size = 8;
+  // Add inputs
+  for (auto& [name, arr] : inputs) {
+    auto dtype = get_type_string(arr.dtype());
+    std::string location = "constant";
+    std::string ref = "*";
+    // Single ints/float/bools are passed by ref
+    if (arr.ndim() == 0) {
```
angeloskath, line 954: Unfortunately that is not correct. They should be passed as constants iff they are available at dispatch time. Otherwise they have to be passed as device pointers or device references.

mlx/fast.h
```diff
@@ -61,4 +63,54 @@ array affine_dequantize(
...
+        output_dtypes_(output_dtypes),
+        grid_(grid),
+        threadgroup_(threadgroup),
+        ensure_row_contiguous_(ensure_row_contiguous),
+        verbose_(verbose) {
+    validate_output_shapes();
+  }
+
+  std::map<std::string, array> run(
+      std::map<std::string, array>& inputs,
```
angeloskath, line 90: Although I see why this would be so much nicer and I am not sure we should change it I have to point out that it deviates from the rest of the C++ API which takes vectors of arrays. Also I would change that to `unordered_map`.

angeloskath (commented):
This is extremely cool and in my todo list for so long! Thank you for taking it up!

I left a bunch of comments. I think the general approach is great! Having the user not write the kernel signature is an awesome decision that solves a bunch of problems imho. There is a few things to decide and an even fewer things to fix and this is going to be an amazing addition!

mlx/fast.cpp
angeloskath: Nah it is fine. I am not sure we would leave print in there. I would rather have the generated kernel stored or something if verbose, not sure... It is definitely good for now.

Also I just wrote the comment before looking at the print.

If I had to guess of a principled way of printing the kernel then I would make `verbose` to be of type `std::optional<std::ostream>` and if it has a value I would print there. This way one can save it in a string or print it from python and so on and so forth.

awni:
This is really neat!

I didn't look at the internals yet.. but one thing I'm wondering about is why the grid (and a few other parameters) gets specified when you construct the kernel as opposed to when you launch it? Is that a constraint due to some other reason or is there a way to make it more flexible?

barronalex:
Really everything is happening in `kernel.__call__` since we can't construct the full function without knowing the names and types of the inputs. I've gone back and forth but I have `grid` etc being passed to `kernel.__init__` mainly because the syntax looked a little bit prettier.

Another option would be forgoing the object altogether and just passing everything in one function call:

```
mx.fast.metal_kernel(
    name="myexp",
    source=source,
    grid=(4, 1, 1),
    threadgroup=(2, 1, 1),
    inputs={"a": a},
    template_args={"T": a.dtype},
    output_shapes={"out": out.shape},
    output_dtypes={"out": out.dtype},
)
```

Definitely open to suggestions

awni:
I guess maybe a better question is what do you do if you want to use the same kernel with outputs with different shapes?

Do you make a new `MetalKernel` in that case? Under the hood will it recompile if only the grid changes?

If that's the expected workflow I think the single function version is nicer.

angeloskath:
Well when I was thinking about how to do that I was thinking that we would simply pass a function from inputs to grid dimensions.

We could even define this function to be from input shapes to grid dimensions but except if we need the strides (which we shouldn't) I think using the inputs is more general.

barronalex:
I'm using a hash of the source and the template arguments to get the `host_name` so it won't recompile if you change the `output_shapes`, `grid` and/or `threadgroup` but have the same `source` and `template`.

I did consider having `grid` accept a function and then you'd need `output_shapes`/`output_dtypes` to accept a function as well. It seemed simpler to let the user write the logic directly as in the `invert_3x3` example above. There shouldn't be much overhead from regenerating the kernel and it's happening at graph construction time anyway, but maybe it's worth measuring to make sure.

mlx/fast.h
```diff
@@ -61,4 +62,52 @@ array affine_dequantize(
     int bits = 4,
     StreamOrDevice s = {});
 
+typedef std::variant<int, bool, Dtype> TemplateArg;
+
+class MetalKernel {
```
awni, line 67: Indeed it might be better to expose this as a function in the C++ API as it will simplify binding code. O/w we have to bind this class in MLX C to make it work (or special case this file).

mlx/fast.h
```diff
@@ -2,6 +2,7 @@
 
 #pragma once
 
+#include <map>
```
awni, line 5: I think you're using `unordered_map` everywhere so this is not needed.

docs/src/usage/custom_metal_kernels.rst
```diff
@@ -0,0 +1,124 @@
+.. _custom_metal_kernels:
```
awni, line 1: I suggest we put this in the [developer guides](https://github.com/ml-explore/mlx/tree/main/docs/src/dev) as it's not really intended for the typical user. Also I think that makes it easier to change the API if we need the flexibility while we are understanding the usage patterns.

mlx/fast_primitives.h
```diff
@@ -1,5 +1,6 @@
 // Copyright © 2024 Apple Inc.
 
+#include <map>
```
awni, line 3: nit: also not needed?

mlx/fast.h
```diff
@@ -61,4 +62,52 @@ array affine_dequantize(
...
+        grid_(grid),
+        threadgroup_(threadgroup),
+        ensure_row_contiguous_(ensure_row_contiguous),
+        verbose_(verbose) {
+    validate_output_shapes();
+  }
+
+  std::unordered_map<std::string, array> run(
+      std::unordered_map<std::string, array>& inputs,
+      StreamOrDevice s = {});
```
awni, line 91: Ideally we can match the C++ and Python APIs. So depending on how this changes it could either be an overload of operator() or you return a lambda from the `metal_kernel` function.

mlx/fast.cpp
```diff
@@ -859,4 +864,229 @@ array affine_dequantize(
...
+  } else {
+    for (auto kv : output_shapes_) {
+      if (output_dtypes_.find(kv.first) == output_dtypes_.end()) {
+        validated = false;
+      }
+    }
+  }
+  if (!validated) {
+    throw std::invalid_argument(
+        "`output_shapes` and `output_dtypes` must have the same keys.");
```
awni, line 884: Could you add the `[...]` as a prefix to the error message for consistency with other error messages. Not sure what should go inside yet something like `[metal_kernel] `output_shapes` ...`

mlx/fast.cpp
```diff
@@ -859,4 +864,229 @@ array affine_dequantize(
...
+      {"thread_position_in_grid", "uint3"},
+      {"thread_position_in_threadgroup", "uint3"},
+      {"threadgroup_position_in_grid", "uint3"},
+      {"threadgroups_per_grid", "uint3"},
+      {"threads_per_grid", "uint3"},
+      {"threads_per_simdgroup", "uint"},
+      {"thread_per_threadgroup", "uint3"},
+  };
+  std::vector<std::pair<std::string, std::string>> attrs;
+  for (auto [attr, dtype] : metal_attributes) {
```
awni, line 940: nit `const auto&`

mlx/fast.cpp
```diff
@@ -859,4 +864,229 @@ array affine_dequantize(
...
+      kernel_source << "  const constant size_t* " << name
+                    << "_strides [[buffer(" << index << ")]]," << std::endl;
+      index++;
+      kernel_source << "  const constant int& " << name << "_ndim [[buffer("
+                    << index << ")]]," << std::endl;
+      index++;
+    }
+  }
+  // Add outputs
+  for (auto [name, dtype] : output_dtypes_) {
```
awni, line 972: same `const auto&`

mlx/fast.cpp
```diff
@@ -859,4 +864,229 @@ array affine_dequantize(
...
+                  << " [[buffer(" << index << ")]]";
+    if (index < inputs.size() + output_shapes_.size() - 1 || attrs.size() > 0) {
+      kernel_source << "," << std::endl;
+    } else {
+      kernel_source << ") {" << std::endl;
+    }
+    index++;
+  }
+  // Add metal attributes e.g. `threadgroup_index_in_grid`
+  for (auto [attr, dtype] : attrs) {
```
awni, line 983: same `const auto&`

mlx/fast.cpp
awni: > If I had to guess of a principled way of printing the kernel then I would make verbose to be of type std::optional<std::ostream> and if it has a value I would print there. This way one can save it in a string or print it from python and so on and so forth.

I like that suggestion more than logging to `std::cout`. Another option that might work is to expose a "kernel_string" attribute /method which gives you the generated kernel in a `std::string`.

mlx/fast.cpp
```diff
@@ -859,4 +864,229 @@ array affine_dequantize(
...
+  template_def << ">";
+  return template_def.str();
+}
+
+std::unordered_map<std::string, array> MetalKernel::run(
+    std::unordered_map<std::string, array>& inputs,
+    StreamOrDevice s_) {
+  auto s = to_stream(s_);
+  if (s.device != Device::gpu) {
+    throw std::invalid_argument("MetalKernel only works on GPU.");
```
awni, line 1021: Nit: same comment re including the function / Class::method name in `[...]` as a prefix.

docs/src/usage/custom_metal_kernels.rst
barronalex: Definitely agree. I've moved it.

mlx/fast.h
barronalex: I ended up moving everything to a single function. I don't think the overhead of constructing the signature is going to matter much in practice and as you suggested, it unifies the Python/C++ APIs and makes the binding code a bit simpler.

mlx/fast_primitives.h
barronalex: Instead of removing I ended up switching the `unordered_map`s back to `map`s since it's useful to have ordering when constructing the signature and it seems unlikely to make a big difference performance wise.

barronalex:
Thanks for the review @awni!
I think it's in a better state now but let me know if you're OK with the single function approach.

mlx/fast_primitives.h
awni: Sorry I should have thought of this. But it would be better to use an `unordered_map` if it's not too tricky, because we have bindings for it in [MLX C already](https://github.com/ml-explore/mlx-c/blob/main/mlx/c/private/map.h). Otherwise we'll have to bind an ordered and an unordered map there.

python/src/fast.cpp
```diff
@@ -183,4 +188,102 @@ void init_fast(nb::module_& parent_module) {
...
+      Run a custom Metal kernel.
+
+      Args:
+        name (str): Name for the kernel. Must be unique in this process.
+        source (str): Source code. This is the body of a function in Metal,
+            the function signature will be generated for you. The function parameters
+            are determined by the names and shapes of ``inputs`` and
+            ``output_shapes``/``output_dtypes``.
+        inputs (dict[str, array]): Inputs. These will be added to the function signature and passed to the Metal kernel.
+            The keys will the names of the arguments to the kernel.
```
awni, line 271:
```suggestion
            The keys will be the names of the arguments to the kernel.
```

python/src/fast.cpp
```diff
@@ -183,4 +188,102 @@ void init_fast(nb::module_& parent_module) {
...
+      "source"_a,
+      "inputs"_a,
+      "output_shapes"_a,
+      "output_dtypes"_a,
+      "grid"_a,
+      "threadgroup"_a,
+      "template"_a = nb::none(),
+      "ensure_row_contiguous"_a = true,
+      "verbose"_a = false,
+      nb::kw_only(),
```
awni, line 257: Wdyt about making more of these arguments `kw_only`? Maybe even all of them?

python/src/fast.cpp
```diff
@@ -183,4 +188,102 @@ void init_fast(nb::module_& parent_module) {
...
+      "stream"_a = nb::none(),
+      nb::sig(
+          "def metal_kernel(name: str, source: str, inputs: Mapping[str, Union[scalar, list, tuple, numpy.ndarray, array]], output_shapes: Mapping[str, Sequence[int]], output_dtypes: Mapping[str, Dtype], grid: tuple[int, int, int], threadgroup: tuple[int, int, int], template: Mapping[str, Union[bool, int, Dtype]] | None = None, ensure_row_contiguous: bool = True, verbose: bool = False, *, stream: Stream | Device | None = None) -> dict[str, array]"),
```
awni, line 260: Nit: You used mapping in the type here but `dict` below. I'm not sure `Mapping` is accurate here.. I don't think objects which implement the interface will auto convert through the binding. So maybe use `dict` here as well?

docs/src/dev/custom_metal_kernels.rst
```diff
@@ -0,0 +1,119 @@
...
+* The keys and values of ``output_shapes`` and ``output_dtypes``
+    In the above, ``out`` is an ``mx.array`` of type ``mx.float16``
+    so we add ``device float16_t* out``.
+* Template parameters passed using ``template``
+    In the above, ``template={"T": mx.float32}`` adds a template of ``template <typename T>`` to the function
+    and instantiates the template with ``custom_kernel_myexp_float<float>``.
+    Template parameters can be ``mx.core.Dtype``, ``int`` or ``bool``.
+* Metal attributes used in ``source`` such as ``[[thread_position_in_grid]]``
+    These will be added as function arguments.
+    All the attributes defined in Table 5.8 of the Metal Shading Language Specification are supported.
```
awni, line 54: Could be nice to include a link here: https://developer.apple.com/metal/Metal-Shading-Language-Specification.pdf

awni:
> I'm using a hash of the source and the template arguments to get the host_name so it won't recompile if you change the output_shapes, grid and/or threadgroup but have the same source and template.

> I did consider having grid accept a function and then you'd need output_shapes/output_dtypes to accept a function as well. It seemed simpler to let the user write the logic directly as in the invert_3x3 example above. There shouldn't be much overhead from regenerating the kernel and it's happening at graph construction time anyway, but maybe it's worth measuring to make sure.

Coming back to this for a minute. I'm still not sure if we should separate the kernel generation from the kernel calling into two steps or not. Even if most of the work doesn't happen until you call it.. it feels a bit more intuitive from a usage standpoint based on the way one typically builds and runs a kernel. For example taking your exp custom kernel, it might look like this instead:

```python
source = """
    uint elem = thread_position_in_grid.x;
    T tmp = inp[elem];
    out[elem] = metal::exp(tmp);
"""

# Build
    name="myexp",
    source=source,
) 

# Call
output = kernel(
    inputs={"inp": a},
    template={"T": mx.float32},
    grid=(a.size, 1, 1),
    threadgroup=(256, 1, 1),
    output_shapes={"out": a.shape},
    output_dtypes={"out": a.dtype},
)  
```

Would it be potentially more scalable if the source is quite long / we can do more at "compile" time?

Does it feel more complicated from a usage standpoint?

FWIW I'm on board with what you have now.. but just want to discuss this a bit since nows the easiest time to make API changes..

barronalex:
I've been going back and forth as well.

I think I agree that the above is more readable and it let's us cache the generated kernel if the inputs/outputs are the same which means the overhead can be kept to a minimum if we're calling in a tight loop.

I'll change it back.

mlx/backend/metal/custom_kernel.cpp
```diff
@@ -0,0 +1,82 @@
+#include "mlx/backend/metal/copy.h"
```
awni, line 1: Nit: Can you add a license comment to the top of this file

python/src/fast.cpp
```diff
@@ -183,4 +188,102 @@ void init_fast(nb::module_& parent_module) {
         Returns:
           array: The quantized version of ``w``
       )pbdoc");
+
+  m.def(
+      "metal_kernel",
+      [](std::string name,
+         std::string source,
+         std::map<std::string, nb::handle> inputs_,
```
awni, line 196: I would use `ScalarOrArray` here (like the other ops) and use `to_array` to do the conversion. That will support everything except list which I think is a bit more consistent with out other operations.

python/src/fast.cpp
```diff
@@ -183,4 +188,102 @@ void init_fast(nb::module_& parent_module) {
...
+                  "Invalid template argument. Must be `mlx.core.Dtype`, `int` or `bool`.");
+            }
+          }
+        }
+        std::map<std::string, array> inputs;
+        for (auto [name, in] : inputs_) {
+          if (name != "stream") {
+            auto value = nb::cast<ArrayInitType>(in);
+            auto arr = create_array(value, std::nullopt);
+            inputs.insert({name, arr});
```
awni, line 231: If you switch to `ScalarOrArray` that will simplify to
```suggestion
            inputs.insert({name, to_array(arr)});
```

python/src/fast.cpp
```diff
@@ -183,4 +188,102 @@ void init_fast(nb::module_& parent_module) {
...
+         std::map<std::string, Dtype> output_dtypes,
+         std::tuple<int, int, int> grid,
+         std::tuple<int, int, int> threadgroup,
+         std::optional<std::map<std::string, nb::handle>> template_args_,
+         bool ensure_row_contiguous,
+         bool verbose,
+         StreamOrDevice s) -> std::map<std::string, array> {
+        std::map<std::string, fast::TemplateArg> template_args;
+        if (template_args_) {
+          for (auto [name, value] : template_args_.value()) {
```
awni, line 207: nit `const auto&`

python/src/fast.cpp
```diff
@@ -183,4 +188,102 @@ void init_fast(nb::module_& parent_module) {
...
+            // Handle bool, int and dtype template args
+            if (PyBool_Check(value.ptr())) {
```
awni, line 209: Nit:  `nb::isinstance<bool>` (we use that pattern elsewhere)

python/src/fast.cpp
```diff
@@ -183,4 +188,102 @@ void init_fast(nb::module_& parent_module) {
...
+              bool bool_val = nb::cast<bool>(value);
+              template_args.insert({name, bool_val});
+            } else if (PyLong_Check(value.ptr())) {
+              int int_val = nb::cast<int>(value);
+              template_args.insert({name, int_val});
+            } else if (
+                nb::type_check(value.type()) &&
+                nb::type_info(value.type()) == typeid(Dtype)) {
```
awni, line 217: Nit: `nb::isinstance<Dtype>`

python/src/fast.cpp
```diff
@@ -183,4 +188,102 @@ void init_fast(nb::module_& parent_module) {
...
+         bool verbose,
+         StreamOrDevice s) -> std::map<std::string, array> {
+        std::map<std::string, fast::TemplateArg> template_args;
+        if (template_args_) {
+          for (auto [name, value] : template_args_.value()) {
+            // Handle bool, int and dtype template args
+            if (PyBool_Check(value.ptr())) {
+              bool bool_val = nb::cast<bool>(value);
+              template_args.insert({name, bool_val});
+            } else if (PyLong_Check(value.ptr())) {
```
awni, line 212: Nit: `nb::isinstance<int>`

python/src/fast.cpp
awni, line 260: Also if we switch to `ScalarOrArray` maybe we simplify the signature to `Union[scalar, array]` just to make it a bit shorter (it's quite long now).

Also could you use `Union` for the `stream` as well? The `|` is only in 3.10 + so just to be extra careful about Python compatibility (since we support >= 3.8).

python/src/fast.cpp
barronalex: Thanks! I spent while looking through the nanobind docs trying to figure out how to do this. This way is much better.

python/src/fast.cpp
barronalex: Sounds good to me.

mlx/fast.cpp
awni: What do you want to do here @barronalex ? If we intend to change the API I would either leave the argument out for now (to make it easier to add) or change it.

mlx/fast.h
```diff
@@ -63,4 +64,32 @@ array affine_dequantize(
...
+ public:
+  MetalKernel(
+      const std::string& name,
+      const std::string& source,
+      bool ensure_row_contiguous)
+      : name_(name),
+        source_(source),
+        ensure_row_contiguous_(ensure_row_contiguous) {}
+
+  std::map<std::string, array> run(
```
awni, line 79: Shouldn't this be `operator()` to match the Python API?

python/src/fast.cpp
```diff
@@ -1,9 +1,15 @@
...
 
 #include <nanobind/nanobind.h>
+#include <nanobind/stl/map.h>
 #include <nanobind/stl/optional.h>
+#include <nanobind/stl/string.h>
 #include <nanobind/stl/tuple.h>
 #include <nanobind/stl/variant.h>
+#include <nanobind/stl/vector.h>
+
+#include "python/src/convert.h"
```
awni, line 11: You don't need this anymore I think.

python/src/fast.cpp
```diff
@@ -186,4 +192,136 @@ void init_fast(nb::module_& parent_module) {
...
+                  template_args.insert({name, bool_val});
+                } else if (nb::isinstance<int>(value)) {
+                  int int_val = nb::cast<int>(value);
+                  template_args.insert({name, int_val});
+                } else if (nb::isinstance<Dtype>(value)) {
+                  Dtype dtype = nb::cast<Dtype>(value);
+                  template_args.insert({name, dtype});
+                } else {
+                  throw std::invalid_argument(
+                      "Invalid template argument. Must be `mlx.core.Dtype`, `int` or `bool`.");
```
awni, line 281: nit: `[metal_kernel]` as prefix here.

python/src/fast.cpp
```diff
@@ -186,4 +192,136 @@ void init_fast(nb::module_& parent_module) {
...
+          nb::kw_only(),
+          "stream"_a = nb::none(),
+          nb::sig(
+              "def __call__(self, inputs: Mapping[str, Union[scalar, array]], output_shapes: Mapping[str, Sequence[int]], output_dtypes: Mapping[str, Dtype], grid: tuple[int, int, int], threadgroup: tuple[int, int, int], template: Optional[Mapping[str, Union[bool, int, Dtype]]] = None, verbose: bool = false, *, stream: Union[None, Stream, Device] = None)"),
+          R"pbdoc(
+            Run the kernel.
+
+            Args:
+              inputs (dict[str, array]): Inputs. These will be added to the function signature and passed to the Metal kernel.
+                  The keys will the names of the arguments to the kernel.
```
awni, line 311: nit typo:
```suggestion
                  The keys will be the names of the arguments to the kernel.
```

python/src/fast.cpp
```diff
@@ -186,4 +192,136 @@ void init_fast(nb::module_& parent_module) {
...
+          "output_shapes"_a,
+          "output_dtypes"_a,
+          "grid"_a,
+          "threadgroup"_a,
+          "template"_a = nb::none(),
+          "verbose"_a = false,
+          nb::kw_only(),
+          "stream"_a = nb::none(),
+          nb::sig(
+              "def __call__(self, inputs: Mapping[str, Union[scalar, array]], output_shapes: Mapping[str, Sequence[int]], output_dtypes: Mapping[str, Dtype], grid: tuple[int, int, int], threadgroup: tuple[int, int, int], template: Optional[Mapping[str, Union[bool, int, Dtype]]] = None, verbose: bool = false, *, stream: Union[None, Stream, Device] = None)"),
```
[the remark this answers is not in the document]
barronalex: I think `Mapping` is ok here. It's what nanobind automatically generates for `std::map` and is recommended by the [python docs](https://docs.python.org/3/library/typing.html#typing.Dict).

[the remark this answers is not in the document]
awni: Thanks for checkin!

awni (approved):
LGTM!

Note we may need to rearrange the internals a bit in `fast.h` to make it easy to bind to MLX C but that is mostly an implementation detail. Let's discuss with @andresy when he is back about how best to bind this.

python/src/fast.cpp
barronalex: Missed this one -- I agree let's make them all kw_only.

awni: Yea.. that will also make the API more flexible since we won't need to worry about back-compatibility with positions in the argument list.

PR #1326, opened 2024-08-13
GPU mx.sign for complex64

barronalex:
As above.


barronalex:
OK a bit more subtlety here than I first thought.

NumPy has actually changed how it computed `np.sign` for complex numbers between version 1 and 2 (see this [issue](https://github.com/numpy/numpy/pull/25441)).
Our Linux tests run NumPy 2 but the Mac tests run NumPy 1 so it was passing for Mac but not Linux.

I've updated the PR to mirror NumPy 2 and do `mx.sign(x) = x / abs(x)`

awni:
Oh thanks for digging into that! I was really curious what was going on there.

awni (approved):
Looks great!!

Issue #1324, opened 2024-08-13
[BUG] Convergence issue in MLX when compared to PyTorch


mzy2240:
**Describe the bug**

The MLX model failed to converge after the second iteration while the PyTorch counterpart works fine.
![losses](https://github.com/user-attachments/assets/17da66fa-7459-4c51-aebb-06700e5486dc)
The plot is coming from the minimal reproducible example. In the actual use case, due to much greater amount of training data, the difference is much more obvious: the pytorch model losses reduce to 0.1 within 10 iterations, while the MLX model stuck at around 1.0 after the first iteration.

**To Reproduce**

[debug 2.ipynb.zip](https://github.com/user-attachments/files/16603797/debug.2.ipynb.zip)
[dataset.zip](https://github.com/user-attachments/files/16603799/dataset.zip)
To reproduce, I have attached the zipped notebook file. You only need to unzip and open it, no need to unzip the dataset.

**Expected behavior**
The training losses should keep the trend of decrease after the second iterations.

**Desktop (please complete the following information):**
 - OS Version: [e.g. MacOS 14.1.2] 13.6.6
 - Version [e.g. 0.7.0] 0.16.3

**Additional context**
Related to [discussion](https://github.com/ml-explore/mlx/discussions/1322)

awni:
There's quite a few differences here between the two versions with some bugs as well. Let me delineate everything:

1. Everything in MLX is batch first. So your forward function is buggy right now. You should change it to the following (basically get rid of all the transposing):

```python
    def __call__(self, src, tgt):
        src = self.encode(src)
        out = self.decode(tgt, src)
        return out
    
    def encode(self, src):
        src_start = self.input_projection(src)
        batch_size, in_sequence_len = src_start.shape[:2]
        pos_encoder = mx.repeat(mx.arange(0, in_sequence_len).reshape((1,-1)), batch_size, axis=0)
        pos_encoder = self.input_pos_embedding(pos_encoder)
        src = src_start + pos_encoder
        src = self.encoder(src, mask=None)
        return src
    
    def decode(self, tgt, memory):
        tgt_start = self.output_projection(tgt)
        batch_size, out_sequence_len = tgt_start.shape[:2]
        pos_decoder = mx.repeat(mx.arange(0, out_sequence_len).reshape((1,-1)), batch_size, axis=0)
        pos_decoder = self.output_pos_embedding(pos_decoder)
        tgt = tgt_start + pos_decoder
[...]

PR #1327, opened 2024-08-13
Fix transformer encoder with `norm_first=False`

awni:
As title.

Relevant to #1324


barronalex (approved):
Nice catch! Kind of impressive this flew under the radar for so long.

PR #1330, opened 2024-08-14
Read arrays from files faster

awni:
This runs about twice as fast on my laptop:

```python
import mlx.core as mx
from huggingface_hub import snapshot_download
import glob

model_path = snapshot_download(
    repo_id="mlx-community/Meta-Llama-3-8B-Instruct-4bit",
    allow_patterns=["*.safetensors"],
)
weight_files = glob.glob(str(model_path + "/model*.safetensors"))
weights = {}
for wf in weight_files:
    weights.update(mx.load(wf))
mx.eval(weights)
```

Running it with `time`:
Pre: 1.4s
Post: 0.7s

### DIFF
--- a/mlx/io/load.h
+++ b/mlx/io/load.h
@@ -22,6 +22,7 @@ class Reader {
       std::ios_base::seekdir way = std::ios_base::beg) = 0;
   virtual void read(char* data, size_t n) = 0;
   virtual std::string label() const = 0;
+  virtual ~Reader() = default;
 };
 
 class Writer {
@@ -34,6 +35,7 @@ class Writer {
       std::ios_base::seekdir way = std::ios_base::beg) = 0;
   virtual void write(const char* data, size_t n) = 0;
   virtual std::string label() const = 0;
+  virtual ~Writer() = default;
 };
 
 class FileReader : public Reader {
@@ -41,7 +43,7 @@ class FileReader : public Reader {
   explicit FileReader(std::string file_path)
       : fd_(open(file_path.c_str(), O_RDONLY)), label_(std::move(file_path)) {}
 
-  ~FileReader() {
+  ~FileReader() override {
     close(fd_);
   }
 
@@ -93,6 +95,10 @@ class FileWriter : public Writer {
   explicit FileWriter(std::string file_path)
       : fd_(open(file_path.c_str(), O_CREAT | O_WRONLY | O_TRUNC, 0644)),
         label_(std::move(file_path)) {}
+
+  ~FileWriter() override {
+    close(fd_);
+  }
 
   bool is_open() const override {
     return fd_ >= 0;
--- a/mlx/io/load.h
+++ b/mlx/io/load.h
@@ -5,9 +5,8 @@
 #include <fcntl.h>
 #include <sys/stat.h>
 #include <unistd.h>
-#include <fstream>
-#include <istream>
 #include <memory>
+#include <sstream>
 
 namespace mlx::core {
 
@@ -68,7 +67,16 @@ class FileReader : public Reader {
   }
 
   void read(char* data, size_t n) override {
-    ::read(fd_, data, n);
+    while (n != 0) {
+      auto m = ::read(fd_, data, n);
+      if (m <= 0) {
+        std::ostringstream msg;
+        msg << "[read] Unable to read " << n << " bytes from file.";
+        throw std::runtime_error(msg.str());
+      }
+      data += m;
+      n -= m;
+    }
   }
 
   std::string label() const override {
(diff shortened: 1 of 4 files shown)


angeloskath:
This is a bit unexpected. Do we have any reasoning as to why? Are we simply avoiding a copy from the kernel into the `ifstream` buffer?

mlx/io/load.h
```diff
@@ -36,39 +39,44 @@ class Writer {
...
+    if (way == std::ios_base::beg) {
+      lseek(fd_, off, 0);
+    } else {
+      lseek(fd_, off, SEEK_CUR);
+    }
   }
 
   void read(char* data, size_t n) override {
-    is_.read(data, n);
+    ::read(fd_, data, n);
```
angeloskath, line 71: Shouldn't this be sth like

```cpp
while (n > 0) {
  size_t m = ::read(fd_, data, n);
  data += m;
  n -= m;
}
```

awni:
I don’t know why it’s so much faster. I just noticed ifstream was getting about half the peak b/w. So I benchmarked it against read and mmap. Read was the fastest, even faster than mmap but not by as much. The internet doesn’t have much detailed info on this for macOS and it is likely platform dependent…

mlx/io/load.h
awni: I guess it should be. I wonder if we should throw or maintain an error bit if it fails..

angeloskath: Good point! We probably need to check and throw which would be the `ifstream` behavior.

angeloskath:
Sounds good then I 'll give it a closer look. Indeed mmap shouldn't be faster than a single read (no random access involved). I just assumed that all these would be properly optimized so the overhead for a single contiguous read of say 10MB should not matter at all 🤷‍♂️

awni:
I updated write as well to use an file descriptor and `write(...)`. For this benchmark:

```python
import time
import mlx.core as mx
from huggingface_hub import snapshot_download
import glob

model_path = snapshot_download(
    repo_id="mlx-community/Meta-Llama-3-8B-Instruct-4bit",
    allow_patterns=["*.safetensors"],
)
tic = time.time()
weight_files = glob.glob(str(model_path + "/model*.safetensors"))
weights = {}
for wf in weight_files:
    weights.update(mx.load(wf))
mx.eval(weights)
toc = time.time()
ms = 1e3 * (toc - tic)
print(f"Read time {ms:.3f}")

tic = time.time()
mx.save_safetensors("model.safetensors", weights)
toc = time.time()
ms = 1e3 * (toc - tic)
print(f"Write time {ms:.3f}")
````

There are the pre and post numbers on M1 max

Pre:
```
Read time 1916.287
Write time 3363.711
```

Post:
```
Read time 918.398
Write time 1818.641
```

File: mlx/io/load.h [hunk 2]
```diff
@@ -36,77 +38,102 @@ class Writer {
...
-        label_(std::move(file_path)) {}
+      : fd_(open(file_path.c_str(), O_RDONLY)), label_(std::move(file_path)) {}
+
+  ~FileReader() {
+    close(fd_);
+  }
 
   bool is_open() const override {
-    return is_.is_open();
+    return fd_ > 0;
```
angeloskath, line 49: I think it is fine to keep this as is instead of checking with `fcntl(fd_, F_GETFD)`. We are managing the whole lifecycle of the file descriptor so we will fail on `read` instead.

Perhaps it had some point when a stream could be passed in the constructor so we wouldn't know if it was open/closed/good etc.

mlx/io/load.h
```diff
@@ -36,77 +38,102 @@ class Writer {
...
 class FileReader : public Reader {
  public:
-  explicit FileReader(std::ifstream is)
-      : is_(std::move(is)), label_("stream") {}
   explicit FileReader(std::string file_path)
-      : is_(std::ifstream(file_path, std::ios::binary)),
-        label_(std::move(file_path)) {}
+      : fd_(open(file_path.c_str(), O_RDONLY)), label_(std::move(file_path)) {}
+
+  ~FileReader() {
```
angeloskath, line 44: I think this should be virtual defined in the base class and also defined for the writer as well.

angeloskath (approved):
I guess except for the virtual destructors this is good to go .

Great find with awesome speedups 🚀

mlx/io/load.h
awni: Good catch, fixed.

awni:
This is really neat!

I didn't look at the internals yet.. but one thing I'm wondering about is why the grid (and a few other parameters) gets specified when you construct the kernel as opposed to when you launch it? Is that a constraint due to some other reason or is there a way to make it more flexible?

awni:
Oops, thanks for catching that. Too many tabs open 😅 !

PR #1333, opened 2024-08-16
move has no effect on const variables

RSilicon:
## Proposed changes
Remove std::move for const variables

## Checklist

### DIFF
--- a/mlx/backend/accelerate/softmax.cpp
+++ b/mlx/backend/accelerate/softmax.cpp
@@ -331,7 +331,7 @@ void Softmax::eval_cpu(const std::vector<array>& inputs, array& out) {
       return x_copy;
     }
   };
-  array in = check_input(std::move(inputs[0]));
+  array in = check_input(inputs[0]);
   out.set_data(
       allocator::malloc_or_wait(in.data_size() * in.itemsize()),
       in.data_size(),
--- a/mlx/compile.cpp
+++ b/mlx/compile.cpp
@@ -824,7 +824,7 @@ std::function<std::vector<array>(const std::vector<array>&)> compile(
       // Mark the entry as not empty since we are about to fill it
       entry.empty = false;
       // Set the constants
-      entry.constants = std::move(constants);
+      entry.constants = constants;
       // Trace to build the graph
       std::tie(entry.inputs, entry.outputs) = compile_trace(fun, inputs);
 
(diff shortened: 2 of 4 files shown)


mlx/compile.cpp
```diff
@@ -824,7 +824,7 @@ std::function<std::vector<array>(const std::vector<array>&)> compile(
       // Mark the entry as not empty since we are about to fill it
       entry.empty = false;
       // Set the constants
-      entry.constants = std::move(constants);
+      entry.constants = constants;
```
angeloskath, line 827: The constants are actually passed by value here or did I miss something.

mlx/backend/accelerate/softmax.cpp
```diff
@@ -331,7 +331,7 @@ void Softmax::eval_cpu(const std::vector<array>& inputs, array& out) {
       return x_copy;
     }
   };
-  array in = check_input(std::move(inputs[0]));
+  array in = check_input(inputs[0]);
```
angeloskath, line 334: Nice, this was left because initially it was written like this

```
array in_pre = inputs[0]; // copy first
auto in = check_input(std::move(in_pre));
```

mlx/primitives.cpp
```diff
@@ -977,7 +977,7 @@ std::vector<array> Convolution::vjp(
           }
         }
 
-        grad = slice(grad, std::move(starts), std::move(stops), stream());
```
awni, line 980: Those ones aren't const are they?

mlx/compile.cpp
```diff
@@ -824,7 +824,7 @@ std::function<std::vector<array>(const std::vector<array>&)> compile(
       // Mark the entry as not empty since we are about to fill it
       entry.empty = false;
       // Set the constants
-      entry.constants = std::move(constants);
```
awni, line 827: That one is not a constant is it?

mlx/primitives.cpp
```diff
@@ -977,7 +977,7 @@ std::vector<array> Convolution::vjp(
...
+        grad = slice(grad, starts, stops, stream());
```
angeloskath, line 980: Slice takes its arguments by value. What we should actually do is check if there are places that could pass by move yet pass by copy.

angeloskath (changes requested):
The first 2 were left

awni:
@RSilicon are you coming back to this?

RSilicon:
Yes I am.

PR #1334, opened 2024-08-18
Eigenvalues and eigenvectors

kashif:
## Proposed changes

## Checklist


kashif:
thanks @awni my question is:

* the output with the lapack can in theory return 2 things... is that an issue with the unitary primitive? the output depending on the params can be values, vectors, or I believe both...
* how do i initialize an empty array in cpp with a given shape?

python/src/linalg.cpp
```diff
@@ -353,4 +353,41 @@ void init_linalg(nb::module_& parent_module) {
...
+      "eigvalsh",
+      &eigvalsh,
+      "a"_a,
+      "upper"_a = true,
+      nb::kw_only(),
+      "stream"_a = nb::none(),
+      nb::sig(
+          "def eigvalsh(a: array, upper: bool = True, *, stream: Union[None, Stream, Device] = None) -> array"),
+      R"pbdoc(
+        Compute the eigenvalues of a complex Hermitian or real symmetric matrix.
```
barronalex, line 366: We don't do a brilliant job of this in the other `linalg` functions, but we should probably check if complex inputs work and raise a not implemented error if not.

mlx/primitives.h
```diff
@@ -2158,4 +2158,23 @@ class Cholesky : public UnaryPrimitive {
   bool upper_;
 };
 
+class Eigvalsh : public UnaryPrimitive {
+ public:
+  explicit Eigvalsh(Stream stream, bool upper, bool compute_vectors)
```
barronalex, line 2163: Given it can compute eigenvectors, this can probably be called `Eigh` rather than `Eigvalsh`?

mlx/backend/common/eigvalsh.cpp
```diff
@@ -0,0 +1,160 @@
...
+#endif
+
+namespace mlx::core {
+
+namespace {
+
+// Delegate to the eigenvalue decomposition taking into account differences in
+// LAPACK implementations (basically how to pass the 'jobz' and 'uplo' strings
+// to fortran).
+int ssyevd_wrapper(char jobz, char uplo, float* matrix, float* w, int N) {
```
barronalex, line 21: I don't think it needs to be in this PR, but it would be nice to add support for arbitrary matrices (not just hermitian/symmetric ones).

Hopefully that could just be a flag on the `Eig` primitive that then uses a different `lapack` incantation?

mlx/linalg.cpp
```diff
@@ -339,4 +339,40 @@ array cholesky_inv(
...
+  if (a.shape(-1) != a.shape(-2)) {
+    throw std::invalid_argument(
+        "[linalg::eigvalsh] Eigenvalues are only defined for square matrices.");
+  }
+
+  std::vector<int> out_shape(a.shape().begin(), a.shape().end() - 1);
+  out_shape.back() = a.shape(-1);
+
+  auto out = array::make_arrays(
+      {out_shape, compute_vectors ? a.shape() : std::vector<int>()},
```
barronalex, line 371: I don't think you need to pass an empty second shape here if `compute_vectors=false`.
Instead, you could pass just one output if you only need eigenvalues and check `outputs.size()` in `Eigvalsh::eval`.

barronalex (commented):
This is nicely done and a great addition -- thanks!
Left a few comments/suggestions.

mlx/linalg.h
```diff
@@ -74,4 +74,8 @@ array pinv(const array& a, StreamOrDevice s = {});
 
 array cholesky_inv(const array& a, bool upper = false, StreamOrDevice s = {});
 
+array eigvalsh(const array& a, bool upper = false, StreamOrDevice s = {});
+
+array eigh(const array& a, bool upper = false, StreamOrDevice s = {});
```
awni, line 79: Shouldn't that return both the eigenvalues and eigenvectors like numpy?

mlx/primitives.h
```diff
@@ -2158,4 +2158,36 @@ class Cholesky : public UnaryPrimitive {
   bool upper_;
 };
 
+class Eigvalsh : public UnaryPrimitive {
```
awni, line 2161: Sorry if @barronalex already made this suggestion: but I think it makes sense to merge the `Eighvalsh` and `Eigh` into a single primitive. And have the ops use the same primitive but just return only the eigenvalues in the case of `eighvalsh`. 

It looks like the work is done anyway.. and the underlying implementations are basically identical.

kashif:
@awni  is the current implementation more like what you mean?

mlx/backend/accelerate/primitives.cpp
```diff
@@ -81,6 +81,7 @@ DEFAULT_MULTI(SVD)
 DEFAULT(Transpose)
 DEFAULT(Inverse)
 DEFAULT(Cholesky)
+DEFAULT_MULTI(EighPrimitive)
```
awni, line 84: Please rename this to `Eigh` for consistency with other primitive names.

awni:
Yes I think it's better this way. I'm debating if we should bother with the `compute_eigenvectors` flag as opposed to just having the primitive always populate both outputs and then just extracting the output you need in the function in `ops.cpp`.

kashif:
@awni  still getting seg-faults... would you have a few mins to help debug?

adhulipa:
> @awni still getting seg-faults... would you have a few mins to help debug?

@kashif curious what is the piece of code causing seg-faults for you?

kashif:
thanks @awni  that looks much cleaner!

awni (approved):
Thanks for the contribution! Will merge once the tests clear.

kashif:
i thank you! i learned a lot and hope to contribute more bindings!

kashif:
failing because now in the` lapack.h` the `#include <cblas.h>` is missing... should i add it?

PR #1336, opened 2024-08-19
Fix contiguity check

angeloskath:
As title.

The col/row checks were swapped. All our slices were reporting non-contiguous unfortunately since this refactoring was introduced some time ago.

Possible speedups in various places.


jagrit06 (approved):
Thanks for adding the additional tests!

PR #1337, opened 2024-08-19
RoPE with frequencies as optional input

awni:
This enables alternative ropes like Yarn, RoPE with fine-tuned parameters, NTK (used in llama 3.1), Su RopE (long-context phi3) to use the `mx.fast.rope` op.


awni:
Relevant PR in MLX examples https://github.com/ml-explore/mlx-examples/pull/945

awni:
### Benchmarks

Timings on M2 Ultra, all models are 4-bit.

Model       | Slow RoPE | Fast RoPE
----- | ----- | ------
Llama 3.1 8B | 86.6 | 105.2
Phi-3 128k | 102.9 | 173.0
DeepSeekV2 Chat Lite | 100.3 | 111.5

mlx/backend/metal/rope.cpp
```diff
@@ -1,4 +1,5 @@
 // Copyright © 2023-2024 Apple Inc.
+#include <iostream> // TODO
```
angeloskath, line 2: Forgotten.

angeloskath (approved):
Looks great! 🚀

python/src/fast.cpp
```diff
@@ -82,23 +82,25 @@ void init_fast(nb::module_& parent_module) {
...
 
         Args:
             a (array): Input array.
             dims (int): The feature dimensions to be rotated. If the input feature
-                is larger than dims then the rest is left unchanged.
+              is larger than dims then the rest is left unchanged.
             traditional (bool): If set to ``True`` choose the traditional
-                implementation which rotates consecutive dimensions.
+              implementation which rotates consecutive dimensions.
             base (float): The base used to compute angular frequency for
```
awni, line 98: I'm thinking about making `base` an `Optional[float]` and making it an error to set both `base` and `freqs`. Wdyt?

mlx/backend/metal/rope.cpp
awni: Haven't done that in a while 😅

python/src/fast.cpp
angeloskath: Yeah I like it! The less ambiguity the better!

PR #1339, opened 2024-08-20
2d gather specialization

### LINKED ISSUE #1338: [BUG] Integer overflow when gathering into very large outputs
```python
import mlx.core as mx
import numpy as np

shape = (1000, 5000000)
a_mx = mx.random.uniform(shape=shape)
a_np = np.array(a_mx)
idx = mx.arange(0, 4400000)
print(np.allclose(a_np[:, np.array(idx)], a_mx[:, idx]))
```

Should print `True`.

There's actually a potential perf optimization here because right now we do the gather with two indices (we make an `arange` for the `:` slice). So then we get (implicitly) huge index arrays for the gather.

awni:
Use a 2D grid for indices with >= 2 dimensions.

Closes #1338 

Also a bit faster for the following:

```python
shape = (1000, 500000)
a = mx.random.uniform(shape=shape)
idx = mx.arange(0, 440000)
mx.eval(a[:, idx])
```

Timing pre/post gives 14.69 / 12.58 ms.

NB: This isn't a general fix for all large arrays with gather. There are still some edge cases that we may want to at least check and throw if not fix. The naive general fix would require a `size_t elem_to_loc` which will be slow so I'm somewhat reluctant there..

### DIFF
--- a/mlx/backend/metal/indexing.cpp
+++ b/mlx/backend/metal/indexing.cpp
@@ -95,11 +95,21 @@ void Gather::eval_gpu(const std::vector<array>& inputs, array& out) {
     slice_size *= s;
   }
 
-  // Launch 2D grid of threads: indices x slice
-  size_t dim0 = out.size() / slice_size;
-  size_t dim1 = slice_size;
-  auto group_dims = get_block_dims(dim0, dim1, 1);
-  MTL::Size grid_dims = MTL::Size(dim0, dim1, 1);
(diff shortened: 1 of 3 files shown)


barronalex (approved):
Nice!
Is it worth adding a test for the failure as well? I guess these are massive arrays so it would have to be in nightly?

awni:
Indeed, I already added a test to our nightly integration tests for this.

PR #1340, opened 2024-08-20
Fix rope

awni:
Fix a bug in rope using the wrong stride for the case where the input is row contiguous and donatable.


angeloskath (approved):
🙏

PR #1343, opened 2024-08-21
fix io

awni:
As title.


mlx/io/load.h
```diff
@@ -141,6 +141,8 @@ class FileWriter : public Writer {
  private:
   int fd_;
   std::string label_;
+  size_t max_bytes_;
```
angeloskath, line 144: Forgotten...

angeloskath (approved):
Thanks for the fix. A member var is forgotten otherwise looks great!

PR #1344, opened 2024-08-21
Bump nanobind + fix extensions

### LINKED ISSUE #1342: [BUG] Metal extensions still use old nanobind version
**Describe the bug**

In the example extension it still uses the old nanobind version (pre 2.0 upgrade). See: `mlx/examples/extensions/pyproject.toml`. Using nanobind==2.0 in the pyproject.toml leads to segmentation fault. While this does not lead to immediate errors for the example extension, this prevents me from upgrading my own mlx extensions from mlx 0.16.1 to 0.16.2 due to incompatibilities and unexpected errors such as `libc++abi: terminating due to uncaught exception of type std::runtime_error: library not found. Failed to load device library from...`

awni:
Bump Nanobind to 2.1.0 and make it consistent between core and extension example.

Closes #1342

### DIFF
--- a/examples/extensions/pyproject.toml
+++ b/examples/extensions/pyproject.toml
@@ -2,7 +2,7 @@
 requires = [
   "setuptools>=42",
   "cmake>=3.24",
-  "mlx>=0.9.0",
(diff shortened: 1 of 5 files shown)


angeloskath (approved):
🙏

awni:
That was fast 🏃

PR #1348, opened 2024-08-23
Fix multiple attributes in MetalKernel

barronalex:
Fix bug when using multiple Metal attributes in a `mx.fast.metal_kernel`.


awni (approved):
Looks good!

PR #1349, opened 2024-08-23
Further reduction tuning

angeloskath:
Still small regression (~1%) on the transformer benchmark and improvements across the board on the micro benchmarks.

### DIFF
--- a/mlx/backend/metal/reduce.h
+++ b/mlx/backend/metal/reduce.h
@@ -16,7 +16,8 @@ void all_reduce_dispatch(
     const std::string& op_name,
     CommandEncoder& compute_encoder,
     metal::Device& d,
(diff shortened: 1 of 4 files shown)


mlx/backend/metal/reduce.h
```diff
@@ -16,7 +16,8 @@ void all_reduce_dispatch(
     const std::string& op_name,
     CommandEncoder& compute_encoder,
     metal::Device& d,
-    const Stream& s);
+    const Stream& s,
+    std::vector<array>& copies);
```
awni, line 20: That's funny that the compiler/linker didn't complain.. I guess it's not used outside `reduce.cpp`

awni (approved):
LGTM!! 🚀

PR #1351, opened 2024-08-23
Reductions update

angeloskath:
This adds a few new column reduction kernels. Mainly adds 2 pass kernels to enable parallelization for long columns without the use of atomics. It also simplifies the small column reduction kernel.

In CNN training it gets us ~5% away from the performance before the removal of atomics and in transformers it gets us about 1% improvement.

Namely, for transformer training

```
0.833 it/s main
0.841 it/s w atomics
0.847 it/s this PR
```

for CNN training

```
1216 samples/s main
6150 samples/s old-reductions
5815 samples/s this PR
```

Besides recovering full perf there are two remaining todos:
1. There is repetitive code in the kernels which could probably be simplified
2. Some indexing for really large arrays needs to be re-checked. At least everything is in size_t now so the only thing to do is properly fix the grid dimensions and corresponding offset calculations in the kernels.

However, I do think this should be merged if anything just for the CNN training perf recovery.


mlx/backend/metal/kernels/reduction/reduce_col.h
```diff
@@ -20,170 +15,128 @@ template <
...
-    }
-
-    short stride = reduction_stride;
-    short size = reduction_size;
-    short blocks = stride / N_READS;
-    short extra = stride - blocks * N_READS;
+  U totals[n_reads];
+  for (int i = 0; i < n_reads; i++) {
+    totals[i] = Op::init;
+  }
```
jagrit06, line 28:
```suggestion
  U totals[n_reads] = {Op::init};
```

It shouldn't really matter, but aleast on my M2 Ultra system with the test `python benchmarks/python/comparative/bench_mlx.py sum_and_add --axis 0 --size 16x4096 --size 16x4096` it gave a small improvement.

angeloskath: Yeah, perhaps we could resort to an `if constexpr (Op::init == 0)` because otherwise this is not the same as before (at least in C++, maybe metal has something I missed).

Basically,

```cpp
int x[3] = {0}; // initialize everything to 0
int y[3] = {1}; // initialize the first to 1 and the rest to 0
```

jagrit06: Good point, I missed that, yeah we can stick with what it was before, no need for a `if constexpr (Op::init == 0)`

mlx/backend/metal/reduce.h
```diff
@@ -37,6 +37,7 @@ void strided_reduce_general_dispatch(
     const std::vector<int>& axes,
     CommandEncoder& compute_encoder,
     metal::Device& d,
-    const Stream& s);
+    const Stream& s,
+    std::vector<array>& copies);
```
awni, line 41: Just as a side note, this idiom is not so critical anymore (though I certainly don't mind if we keep it this way). I think (?) we started writing functions with temporaries pass-by-reference vectors to avoid completion handler bloat. But if you do a `d.add_temporaries` inside the `strided_reduce...` it will put the temporary buffer in a vector and we only add one completion handler at the very end after the command buffer is committed.

awni (approved):
Looks great!!

mlx/backend/metal/reduce.h
angeloskath: Oh you 're right! Force of habit. I will change all the reductions to use that idiom because it makes the internal API simpler. (see normalization implementations in `fast` for instance).

mlx/backend/metal/reduce.cpp
```diff
@@ -469,51 +484,126 @@ void strided_reduce_small(
...
-        MTL::Size(threadgroup_size, out_grid_dims.width, out_grid_dims.height);
-    group_dims = MTL::Size(threadgroup_size, 1, 1);
+  const int n_reads = 4;
+  size_t reduction_stride_blocks =
+      (args.reduction_stride + n_reads - 1) / n_reads;
+  size_t total = args.reduction_size * args.non_col_reductions;
+  size_t threadgroup_x = std::min(reduction_stride_blocks, 32ul);
+  size_t threadgroup_y = std::min(
+      8ul,
+      std::min(kernel->maxTotalThreadsPerThreadgroup() / threadgroup_x, total));
```
zcbenz, line 507: Since `threadgroup_x` can only be 32 at most, I think there is no way for `maxTotalThreadsPerThreadgroup() / threadgroup_x` to be smaller than 8 and this `std::min` is not needed?

awni: - I don't know what the minimum value of `total` is... so it depends on that
- Also it's unlikely but in general there is no guarantee that `maxTotalThreadsPerThreadgroup > 32 * 8`. If you use a lot of shared resources in the kernel then that number can be smaller. I don't know how this kernel will fare in those cases. We often put little asserts on that to make sure `maxTotalThreadsPerThreadgroup` is in the range we expect.

zcbenz: That makes sense, thanks!

PR #1352, opened 2024-08-23
Add `grid_sample` example to `metal_kernel` docs

barronalex:
Add a much more involved [grid_sample](https://pytorch.org/docs/stable/generated/torch.nn.functional.grid_sample.html) example to the `metal_kernel` docs.
It produces the same outputs as the PyTorch version with default arguments and is much faster than the MLX ops version.

I also added two small features to `metal_kernel` to get the example to work:

- `zero_outputs`: set the output arrays to zero before the kernel runs
- `atomic_outputs`: use Metal atomics for the output array arguments

Both are disabled by default.

My only concern is that the example might be a bit too long to be clear, but it definitely exercises all of the custom kernel features.

I think `grid_sample` is fairly commonly used in vision models, so I'll put this in a full C++ primitive as well in a separate PR.

### DIFF
--- a/docs/src/dev/custom_metal_kernels.rst
+++ b/docs/src/dev/custom_metal_kernels.rst
@@ -275,8 +275,8 @@ its custom vjp transform so MLX can differentiate it.
 The backwards pass requires atomically updating ``x_grad``/``grid_grad`` and so
 requires a few extra ``mx.fast.metal_kernel`` features:
 
-* ``zero_outputs=True``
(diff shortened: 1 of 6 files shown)


angeloskath (approved):
I love it!

One thing that comes to mind is instead of `zero_outputs` we could have an initialization with a scalar but perhaps it's too much and zero is better. Wdyt?

barronalex:
That's a nice idea! I'll put it in.

PR #1353, opened 2024-08-24
Chore: update pre-commit hooks

NripeshN:
Back at it again 😅


awni (approved):
Thanks! Would you mind bumping in MLX Examples and MLX Data as well?

PR #1356, opened 2024-08-24
Put MPI ops in default stream with async for faster comms

awni:
This is an RFC / WIP to see if we should move in this direction. 

As you can see there is a slight gain in perf when you mix GPU work with communication.  It may be the case that that gap will widen with some pending upstream improvements in metal synchronization. Will check that.

Benchmark | Pre | Post
----- | ---- | -----
all_sum_with_sine  | 3.69 sec | 3.42 sec
2 process single machine 4-bit Mistral | 39 toks/sec | 41 toks/sec |

### DIFF
--- a/benchmarks/python/distributed_bench.py
+++ b/benchmarks/python/distributed_bench.py
@@ -1,0 +1,66 @@
+# Copyright © 2024 Apple Inc.
+
+"""
+Run with:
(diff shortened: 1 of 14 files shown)


awni:
Ok @angeloskath this is good for review.

angeloskath (approved):
This is great! Thanks!

PR #1358, opened 2024-08-25
Add optional headers to ``mx.fast.metal_kernel``

### LINKED ISSUE #1357: [Feature Request] Metal kernels external includes.
Ability to compile custom kernels has unlocked ability to write apps that can use Metal with very little setup (just install MLX) and in Python. While it works for simple things, ability to allow includes of external .metal files would unlock another level and allow write more complicated apps using MLX framework. For example, I'm currently thinking of writing a raytracer using this feature.

barronalex:
Resolves #1357

One question here:
I was thinking about allowing a `Path` or `Sequence[Path]` as well as a string in the headers argument. Do you think that's worth having? Feels a bit more natural in the python API than C++.

### DIFF
--- a/python/tests/test_fast.py
+++ b/python/tests/test_fast.py
@@ -651,7 +651,7 @@ class TestFast(mlx_tests.MLXTestCase):
         mx.random.seed(7)
         a = mx.random.normal(shape=(2, 2))
         kernel = mx.fast.metal_kernel(
-            name="basic",
+            name="helper",
             header="""
             template <typename T>
             T do_exp(T x) {
--- a/python/tests/test_fast.py
+++ b/python/tests/test_fast.py
@@ -551,7 +551,7 @@ class TestFast(mlx_tests.MLXTestCase):
     @unittest.skipIf(not mx.metal.is_available(), "Metal is not available")
     def test_custom_kernel_basic(self):
         mx.random.seed(7)
-        a = mx.random.normal(shape=(3, 6))
+        a = mx.random.normal(shape=(2, 2))
         kernel = mx.fast.metal_kernel(
             name="basic",
             source="""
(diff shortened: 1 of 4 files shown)


awni:
> I was thinking about allowing a Path or Sequence[Path] as well as a string in the headers argument.

I think what you have now is sufficiently flexible. I would probably keep it this way as it's easy enough to concatenate multiple files and pass them as a single string. 

Just a question / comment on naming:

- header (singular) seems slightly better than headers (plural) as it's just a single header?
- include or includes might make sense if it's a file or set of files to include (?)

barronalex:
Good point, I like `includes` a bit better so have changed it to that.

angeloskath:
I might be nitpicking but I think (and I agree) that the point was `includes` or `include` makes sense if the argument was a file. `header` makes more sense to me too since it is a header/heading for the source code.

python/tests/test_fast.py
```diff
@@ -646,6 +646,34 @@ def test_custom_kernel_strides(self):
...
+        )
+        out = kernel(
+            inputs={"a": a},
+            grid=(4, 1, 1),
+            threadgroup=(2, 1, 1),
+            output_shapes={"out1": (2, 2)},
+            output_dtypes={"out1": mx.float32},
+            stream=mx.gpu,
+        )
+        print(out["out1"])
```
angeloskath, line 674: Forgotten print.

python/tests/test_fast.py
```diff
@@ -646,6 +646,34 @@ def test_custom_kernel_strides(self):
...
+            }
+            """,
+            source="""
+                uint elem = thread_position_in_grid.x;
+                out1[elem] = do_exp(a[elem]);
+            """,
+        )
+        out = kernel(
+            inputs={"a": a},
+            grid=(4, 1, 1),
```
angeloskath, line 668: This launches 4 threads for 18 values and the check below `a[:2, :2]` should fail. It doesn't because there is no `self.assertTrue` on `mx.allclose`. Same goes for some tests further up in this file.

angeloskath (changes requested):
This looks great, I would also choose `header` instead of `includes`.

Also the tests have some small typos in them that went unnoticed, after editing them we can merge.

python/tests/test_fast.py
barronalex: Thanks for catching this!

angeloskath:
@barronalex  For the 2nd test perhaps `metal::precise::exp(x)` would fix the error? We use precise for the unary exp.

angeloskath (approved):
Looks great. Let's merge when the tests pass.

PR #1359, opened 2024-08-25
Fix ternary for large arrays

awni:
- Use the vector specialization of ternary when all the inputs have the same contiguity.
- Fix the 2D dispatch
- Add input donation


angeloskath (approved):
Looks great!

PR #1360, opened 2024-08-25
int() and float() for mx.array

barronalex:
Make size 1 `mx.array`s work with `int()` and `float()`.

### DIFF
--- a/python/src/convert.cpp
+++ b/python/src/convert.cpp
@@ -160,6 +160,10 @@ nb::ndarray<> mlx_to_dlpack(const array& a) {
 }
 
 nb::object to_scalar(array& a) {
+  if (a.size() != 1) {
+    throw std::invalid_argument(
+        "[convert] Only length-1 arrays can be converted to Python scalars.");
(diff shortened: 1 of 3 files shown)


python/src/convert.cpp
```diff
@@ -160,6 +160,10 @@ nb::ndarray<> mlx_to_dlpack(const array& a) {
 }
 
 nb::object to_scalar(array& a) {
+  if (a.size() != 1) {
+    throw std::invalid_argument(
+        "[convert] Only length-1 arrays can be converted to Python scalars.");
+  }
```
awni, line 166: I think the call to `.item` would already raise an error in that case so this is redundant. But maybe your intnetion was a nicer error message? (which I think it is).

awni (approved):
Looks great, thanks!

python/src/convert.cpp
barronalex: I didn't realize, but the error is a bit nicer so I guess it stays :)

PR #1361, opened 2024-08-26
Fix extension metal library finding

awni:
I broke this in https://github.com/ml-explore/mlx/pull/1310.. didn't catch the library path finding was dependent on the way the function was linked 😅 .


barronalex (approved):
Looks good!

PR #1366, opened 2024-08-27
Adds send/recv ops in distributed

angeloskath:
As title.

### DIFF
--- a/mlx/distributed/no_distributed.cpp
+++ b/mlx/distributed/no_distributed.cpp
@@ -34,8 +34,8 @@ Stream communication_stream() {
 
 void all_sum(Group group, const array& input, array& output) {}
 void all_gather(Group group, const array& input, array& output) {}
-void send(Group group, const array& input, int dst);
-void recv(Group group, array& out, int src);
+void send(Group group, const array& input, int dst) {}
+void recv(Group group, array& out, int src) {}
 
 } // namespace detail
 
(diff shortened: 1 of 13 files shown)


python/src/distributed.cpp
```diff
@@ -121,4 +122,88 @@ void init_distributed(nb::module_& parent_module) {
...
+      "x"_a,
+      "dst"_a,
+      nb::kw_only(),
+      "group"_a = nb::none(),
+      "stream"_a = nb::none(),
+      nb::sig(
+          "def send(x: array, dst: int, *, group: Optional[Group] = None, stream: Union[None, Stream, Device] = None) -> array"),
+      R"pbdoc(
+        Send an array from the current process to the process that has rank
+        ``dst`` in the group..
```
awni, line 138: Nit extra `.`

awni (approved):
Looks great!

mlx/distributed/no_distributed.cpp
```diff
@@ -34,6 +34,8 @@ Stream communication_stream() {
 
 void all_sum(Group group, const array& input, array& output) {}
 void all_gather(Group group, const array& input, array& output) {}
+void send(Group group, const array& input, int dst);
+void recv(Group group, array& out, int src);
```
awni, line 38: I think those need `{}` to fix the linking errors.

awni:
Oh also we should add them to the doc https://github.com/ml-explore/mlx/blob/main/docs/src/python/distributed.rst

PR #1368, opened 2024-08-27
Make eval just wait if all arrays are scheduled

angeloskath:
Makes sure that `eval` simply waits for arrays to be ready if they are already scheduled instead of scheduling a synchronizer task.

I wasn't able to make a simple test that showcases the issue but as @davidkoski found out during generation if we use `mx.eval(y)` instead of `y.item()` we get a slowdown which this PR fixes.


awni (approved):
Looks great, thanks for the fix!

PR #1369, opened 2024-08-28
Even Faster I/O

awni:
A couple benchmarks on M2 Ultra:

Bench | Pre | Post |
---- | ---- | ---- |
Loading 4-bit Gemma 27B | 2.15 s | 1.47 s |
Genrating with Flux | 8.93 s | 7.14 s |

Benchmarks on M1 Max:

Bench | Pre | Post |
---- | ---- | ---- |
Mistral 7B 16-bit | 2.73 | 2.05
Llama 8B 4bit | 0.64 | 0.42 |
Flux | 29.0 | 22.4 

Gemma benchmark:

```python
model, tokenizer = load("mlx-community/gemma-2-27b-it-4bit", lazy=True)
tic = time.time()
mx.eval(model)
toc = time.time()
print(toc - tic)
```

Flux benchmark:

```
diffusionkit-cli --model-version FLUX.1-schnell --prompt "A cat samurai" --steps 4
```

### DIFF
--- a/mlx/backend/common/load.cpp
+++ b/mlx/backend/common/load.cpp
@@ -33,8 +33,7 @@ void Load::eval(const std::vector<array>& inputs, array& out) {
   assert(inputs.size() == 0);
   out.set_data(allocator::malloc_or_wait(out.nbytes()));
 
-  reader_->seek(offset_);
-  reader_->read(out.data<char>(), out.nbytes());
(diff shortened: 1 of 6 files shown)


angeloskath (approved):
Looks great. I changed `readfn` to account for the fact that `pread` like `read` may read less bytes than what was requested, let me know if it is too verbose, otherwise merge at will.

PR #1372, opened 2024-08-28
Async load

awni:
This uses the same trick as the comms to put Load ops in their own stream.

A simple benchmark shows that this can enable pipelining Loading with GPU work:

Bench | Pre | Post |
---- | --- | --- |
op_and_load | 12.26343 msec | 6.74928 msec
op | 6.50007 msec | 6.48391 msec
load | 4.63016 msec | 4.60564 msec

The code for the benchmark:

```python
import mlx.core as mx
import time

def time_fn(fn, *args, **kwargs):
    print(f"Timing {fn.__name__} ...", end=" ")

    # warmup
    for _ in range(5):
        mx.eval(fn(*args, **kwargs))

    num_iters = 100
    tic = time.perf_counter()
    for _ in range(num_iters):
        x = mx.eval(fn(*args, **kwargs))
    toc = time.perf_counter()

    msec = 1e3 * (toc - tic) / num_iters
    print(f"{msec:.5f} msec")

def op_and_load(x, y):
    z = x @ y
    a = mx.load("model.safetensors")["a"]
    return z + a

def op(x, y):
    return x @ y

def load():
    return mx.load("model.safetensors")


n = 4096
#a = mx.random.uniform(shape=(n, n))
#mx.save_safetensors("model.safetensors", {"a": a})
x = mx.random.uniform(shape=(n, n))
y = mx.random.uniform(shape=(n, n))

time_fn(op_and_load, x, y)
time_fn(op, x, y)
time_fn(load)
```

### DIFF
--- a/mlx/backend/common/load.cpp
+++ b/mlx/backend/common/load.cpp
@@ -5,10 +5,8 @@
 #include <utility>
 
 #include "mlx/allocator.h"
-#include "mlx/io/load.h"
+#include "mlx/backend/common/load.h"
 #include "mlx/primitives.h"
-
-namespace mlx::core {
 
 namespace {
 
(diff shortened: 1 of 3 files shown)


angeloskath (approved):
🚀

PR #1373, opened 2024-08-28
Fix JIT reductions

barronalex:
Looks like we broke the JIT when we updated the reductions recently.

I added the python tests with JIT enabled to CI so this is more obvious next time. Hopefully it doesn't slow down the dev cycle too much.


angeloskath (approved):
Thank you for fixing this!

It looks perfect to me, let's merge after the tests pass.

PR #1374, opened 2024-08-29
Even Even Faster IO

awni:
A couple changes:

- Add static io in the thread pool. The GPU load now submits to this thread pool. In order to keep the event signaling in the correct order, the tasks from this thread pool are waited on in the io stream before signaling the GPU.
- Make the parallel file reader static to protect against the now potential situation with lots of files simultaneously being read creating an explosion of threads. 
- The maximum number of active threads (from thread pools) is now 8 - 4 for each of the inner and out thread pools.

M2 Ultra:

Bench | 0.17.1 | Main | Post
--- | ---- | --- | ---
4-bit Gemma 27B | 2.20 | 1.47 | 0.75 |
Flux Generation | 8.61 |   7.08 | 6.53

M1 Max:

Bench | 0.17.1 | Main | Post
--- | ---- | --- | ---
4-bit Llama 8B | 0.64 | 0.46 | 0.26

### DIFF
--- a/mlx/backend/metal/primitives.cpp
+++ b/mlx/backend/metal/primitives.cpp
@@ -202,15 +202,18 @@ void Load::eval_gpu(const std::vector<array>& inputs, array& out) {
   static Stream io_stream = new_stream(Device::cpu);
   out.set_data(allocator::malloc_or_wait(out.nbytes()));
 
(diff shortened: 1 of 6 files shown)


angeloskath (approved):
🚀

PR #1376, opened 2024-08-29
Patch bump

angeloskath:
As title. There is quite a few stuff probably doesn't warrant a full release though

- Fast IO
- GPU waits on comms
- pinv
- send/recv
- JIT reduction fix
- MetalKernel headers

and probably more.


awni (approved):
🚀

PR #1379, opened 2024-08-30
Fix overflow in quantize/dequantize

barronalex:
Resolves https://github.com/ml-explore/mlx-examples/issues/967

Doesn't seem to make any speed difference when quantizing an 8B model:

Before: 1.62 sec
After: 1.60 sec


awni (approved):
Thanks for tracking that down!!

PR #1381, opened 2024-08-30
Fix compile with byte sized constants

awni:
Printing the `uint8` and `int8` directly to the string stream produces badly formatted constants which don't compile, e.g.

```
  auto tmp_C = static_cast<int8_t>(<U+0003>);
```

This fixes that.


angeloskath (approved):
Thanks for the fix!

Issue #1382, opened 2024-08-30
[Bug] `argsort` crashed on Xcode 16.1 beta


kemchenj:
Reproducible code:

```swift
let array = MLXArray.zeros([66512])
let sortedArray = argSort(array)
sortedArray.eval()
```

Environment:

- Xcode: Version 16.1 beta (16B5001e)
- macOS: 15.1 Beta (24B5035e) 

Here is the backtrace:

```
* thread #5, stop reason = signal SIGABRT
    frame #0: 0x00000001931ea600 libsystem_kernel.dylib`__pthread_kill + 8
    frame #1: 0x0000000103f0bfa8 libsystem_pthread.dylib`pthread_kill + 288
    frame #2: 0x000000019312f908 libsystem_c.dylib`abort + 128
  * frame #3: 0x00000001000c1500 Example1`std::__1::vector<unsigned long, std::__1::allocator<unsigned long>>::operator[][abi:de180100](this=0x0000600001dac018 size=1, __n=1) const at vector:1400:3
    frame #4: 0x000000010044b854 Example1`std::__1::tuple<std::__1::vector<int, std::__1::allocator<int>>, std::__1::vector<std::__1::vector<unsigned long, std::__1::allocator<unsigned long>>, std::__1::allocator<std::__1::vector<unsigned long, std::__1::allocator<unsigned long>>>>> mlx::core::collapse_contiguous_dims<unsigned long>(shape=size=2, strides=size=2) at utils.h:60:13
    frame #5: 0x0000000100bdac38 Example1`void mlx::core::copy_gpu_inplace<unsigned long>(in=0x00000001700296c0, out=0x00006000011d4160, data_shape=size=2, strides_in_pre=size=2, strides_out_pre=size=1, inp_offset=0, out_offset=0, ctype=Vector, s=0x0000600001dc2120) at copy.cpp:62:27
    frame #6: 0x0000000100bdaac8 Example1`mlx::core::copy_gpu_inplace(in=0x00000001700296c0, out=0x00006000011d4160, ctype=Vector, s=0x0000600001dc2120) at copy.cpp:157:10
    frame #7: 0x0000000100cce7b8 Example1`mlx::core::(anonymous namespace)::multi_block_sort(s=0x0000600001dc2120, d=0x00000001017406e0, in=0x00006000011d4180, out=0x00006000011d4160, axis=0, bn=512, tn=8, n_blocks=17, argsort=true) at sort.cpp:239:5
[...]

PR #1383, opened 2024-08-31
Fix copy in the sort primitive

angeloskath:
Sort assumed that copy didn't care about shapes which unfortunately it did and caused out of bounds access in `collapse_contiguous_dims`. This basically fixes https://github.com/ml-explore/mlx/issues/1382 .

This PR does the following:
1. Add assertions in `copy` so at least in debug mode it should crash if misused
2. Refactor the copy of the result of multi-block sort to be hopefully simpler.
 
The refactoring idea is the following, sort has the result in a contiguous array that has the following shape `x.moveaxis(reduction_axis, -1).reshape(-1, x.shape[reduction_axis])`. We simply create the strides such that the sorted result is transposed back into the shape of `out` and do the copy by manually passing the correct shape and strides.

### DIFF
--- a/mlx/backend/metal/copy.cpp
+++ b/mlx/backend/metal/copy.cpp
@@ -156,6 +156,7 @@ void copy_gpu_inplace(
     array& out,
     CopyType ctype,
     const Stream& s) {
+  assert(in.shape() == out.shape());
   return copy_gpu_inplace(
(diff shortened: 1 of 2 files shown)


awni (approved):
Thanks for the quick fix!!

PR #1384, opened 2024-08-31
Xcode 160

awni:
Ok I think this is finally ready 🤞 

Also the Circle config update is really nice as it halves the overall CI time.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -92,7 +92,7 @@ jobs:
           name: Install Python package
           command: |
             source env/bin/activate
-            CMAKE_BUILD_PARALLEL_LEVEL=`sysctl -n hw.ncpu` pip install -e . -v
+            DEBUG=1 CMAKE_BUILD_PARALLEL_LEVEL=`sysctl -n hw.ncpu` pip install -e . -v
       - run:
           name: Generate package stubs
           command: |
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -38,8 +38,12 @@ jobs:
       - run:
           name: Install Python package
           command: |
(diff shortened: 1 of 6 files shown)


.circleci/config.yml
```diff
@@ -314,7 +314,7 @@ workflows:
           matrix:
             parameters:
               python_version: ["3.8", "3.9", "3.10", "3.11", "3.12"]
-              xcode_version: ["15.0.0", "15.2.0"]
+              xcode_version: ["15.0.0", "15.2.0", "16.0.0"]
```
awni, line 317: I'm not actually sure if this will work for the release. It needs to be tested. It might make the same wheel (14.0) as using xcode 15.2.0

awni: Actually it puts 15_0 in the wheel so should be good:
`mlx-0.17.2.dev20240831+161b2e8f-cp312-cp312-macosx_15_0_arm64.whl`

awni:
For some reason xcode 16.0 keeps stalling on the build ...

NripeshN:
> For some reason xcode 16.0 keeps stalling on the build ...

Not completely sure if this is the case, but `BNNSLayerParametersBroadcastMatMul`, `BNNSFilterCreateLayerBroadcastMatMul`, `BNNSFilterApplyTwoInput`, `BNNSFilterDestroy` is deprecated as of MacOS 15.0
The file in question is `mlx/backend/accelerate/matmul.cpp`

In my local machine I am using Xcode 16.1 and the deprecation warnings appear to be at the same step as to when the CI pipeline just stops. It does continue to compile locally later though.

angeloskath (approved):
🚀

PR #1385, opened 2024-08-31
Support for 1D and 2D grid in mx.fast.metal_kernel

jeethu:
## Proposed changes

This change allows using metal kernels with 1D and 2D grids, in addition to 3D grids. Previously, `grid` and `threadgroup` argument had to be 3 element tuples of integers. With this change, they can be one of `int` (1D), `tuple[int, int]` (2D) and `tuple[int, int, int]` (3D).

## Checklist

### DIFF
--- a/mlx/fast.cpp
+++ b/mlx/fast.cpp
@@ -1011,7 +1011,7 @@ void write_signature(
       {"threadgroups_per_grid", grid_dtype},
       {"threads_per_grid", grid_dtype},
       {"threads_per_simdgroup", "uint"},
-      {"thread_per_threadgroup", grid_dtype},
+      {"threads_per_threadgroup", grid_dtype},
   };
   std::vector<std::pair<std::string, std::string>> attrs;
   for (const auto& [attr, dtype] : metal_attributes) {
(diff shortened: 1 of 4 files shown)


angeloskath:
Although nice, I don't necessarily see the reason as one can simply pass the tuple as follows `(N, 1, 1)` or of course `(N, M, 1)`. Subsequently in the kernel you can access just the `.x` member or the `.xy` swizzle.

Is there something that is not possible with the above?

angeloskath:
@jeethu is there something I missed? Do you think we need the 1D and 2D grid capability?

jeethu:
> @jeethu is there something I missed? Do you think we need the 1D and 2D grid capability?

I'd written this to port a few old of my old 1D and 2D metal kernels for use with MLX without changing the kernel source. Your suggestion does sound reasonable. Happy to close this PR, if you think it introduces too much of a maintenance overhead.

Also, there's a [minor typo](https://github.com/ml-explore/mlx/blob/main/mlx/fast.cpp#L975) fix (`s/thread_per_threadgroup/threads_per_threadgroup/`) in this PR that should probably go in another change.

PR #1387, opened 2024-09-01
add std as method

### LINKED ISSUE #1386: mx.std() doesn't exist?   [BUG]
**Describe the bug**
mx.std() doesn't exist?

**To Reproduce**
```python
import mlx.core as mx
mx.array([1,2,3]).std()
```
AttributeError: 'array' object has no attribute 'std'

**Expected behavior**
https://ml-explore.github.io/mlx/build/html/python/_autosummary/mlx.core.std.html

 - OS Version: MacOS 14.1.2
 - Version 0.17.2

**Additional context**
```python
mx.array([1,2,3]).var().sqrt() # works for now
```

awni:
- Add std as method to array, close #1386

### DIFF
--- a/docs/src/python/array.rst
+++ b/docs/src/python/array.rst
@@ -53,8 +53,9 @@ Array
     array.sqrt
     array.square
     array.squeeze
-    array.swapaxes
+    array.std
     array.sum
+    array.swapaxes
     array.transpose
     array.T
     array.var
--- a/docs/src/python/ops.rst
+++ b/docs/src/python/ops.rst
@@ -78,6 +78,7 @@ Operations
    hadamard_transform
    identity
    inner
+   isfinite
    isclose
    isinf
    isnan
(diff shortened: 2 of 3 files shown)


bhargavyagnik:
We could also update tests for all ops in the test_array.py
https://github.com/ml-explore/mlx/blob/58dca7d8464d6ce25f0fd94eb3ece4cc0c119632/python/tests/test_array.py#L1426

I did a small check, 
`
not_in_ops = [
    "abs",
    "conj",
    "cummax",
    "cummin",
    "cumprod",
    "cumsum",
    "diag",
    "diagonal",
    "flatten",
    "log10",
    "log2",
    ("moveaxis",1,2),
    "round",
    "split",
    "squeeze",
    ("swapaxes",1,2),
    "std"
]`
these many are missing. If you want, I can add in this PR or create a different PR for it.

angeloskath (approved):
👍

PR #1389, opened 2024-09-02
Fix reduce edge case

angeloskath:
There is currently a bug when reducing an axis that is a transposed singleton axis and the axis just before that.

Basically the following:

```python
x = mx.random.normal((5, 5, 1, 5))

# fixed in this PR
x.transpose(0, 2, 1, 3).sum((0, 1))

# all the following are fine
x.transpose(0, 2, 1, 3).sum(0)
x.transpose(0, 2, 1, 3).sum((0, 2))
x.transpose(0, 2, 1, 3).sum((0, 3))
x.transpose(0, 2, 1, 3).sum((1, 2))
x.transpose(0, 2, 1, 3).sum((1, 3))
...
```


awni (approved):
🙏  thanks for the fix!

PR #1390, opened 2024-09-02
Add ConvTransposed layers

EduardoPach:
## Proposed changes

This PR adds `ConvTranspose1d`, `ConvTranspose2d`, and `ConvTranspose3d` to `mlx.nn`, following interest expressed in https://github.com/ml-explore/mlx/discussions/1001. Additionally, equivalence tests were included to compare the outputs with PyTorch's implementations, as the [example implementation](https://github.com/ml-explore/mlx-examples/blob/main/segment_anything/segment_anything/mask_decoder.py#L202-L252) referenced in the discussion did not match PyTorch outputs upon comparison.

## Checklist

### DIFF
--- a/docs/src/python/nn/layers.rst
+++ b/docs/src/python/nn/layers.rst
@@ -16,6 +16,9 @@ Layers
    Conv1d
    Conv2d
    Conv3d
+   ConvTransposed1d
+   ConvTransposed2d
+   ConvTransposed3d
    Dropout
    Dropout2d
    Dropout3d
--- a/python/mlx/nn/layers/__init__.py
+++ b/python/mlx/nn/layers/__init__.py
@@ -54,7 +54,14 @@ from mlx.nn.layers.activations import (
 )
 from mlx.nn.layers.base import Module
 from mlx.nn.layers.containers import Sequential
-from mlx.nn.layers.convolution import Conv1d, Conv2d, Conv3d
+from mlx.nn.layers.convolution import (
+    Conv1d,
+    Conv2d,
+    Conv3d,
+    ConvTranspose1d,
+    ConvTranspose2d,
+    ConvTranspose3d,
+)
 from mlx.nn.layers.dropout import Dropout, Dropout2d, Dropout3d
 from mlx.nn.layers.embedding import Embedding
 from mlx.nn.layers.linear import Bilinear, Identity, Linear
--- a/python/mlx/nn/layers/convolution.py
+++ b/python/mlx/nn/layers/convolution.py
@@ -200,3 +200,168 @@ class Conv3d(Module):
         if "bias" in self:
             y = y + self.bias
         return y
+
+
+class ConvTranspose1d(Module):
+    def __init__(
+        self,
+        in_channels: int,
+        out_channels: int,
+        kernel_size: int,
+        stride: int = 1,
+        padding: int = 0,
+        dilation: int = 1,
+        bias: bool = True,
+    ):
+        super().__init__()
+        # Add +1 to padding to match the behavior of PyTorch
+        padding += 1
+
+        scale = math.sqrt(1 / (in_channels * kernel_size))
+        self.weight = mx.random.uniform(
+            low=-scale,
+            high=scale,
+            shape=(out_channels, kernel_size, in_channels),
+        )
+        if bias:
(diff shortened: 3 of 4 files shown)


EduardoPach:
Oh, I've just seen that https://github.com/ml-explore/mlx/pull/1245 aims to do the same thing

EduardoPach:
> Oh, I've just seen that #1245 aims to do the same thing

@awni tagging you here just to get a feedback

mlaves:
@awni Sorry that I let my PR #1245 become stale. I try to finish it up asap.
I also like the simplistic approach of this PR. However, I think that without changing `conv_general`, this approach will yield incorrect gradients. That's why I introduced a new primitive in my PR.

EduardoPach:
> @awni Sorry that I let my PR #1245 become stale. I try to finish it up asap.
> 
> I also like the simplistic approach of this PR. However, I think that without changing `conv_general`, this approach will yield incorrect gradients. That's why I introduced a new primitive in my PR.

Makes sense, tbh I haven't checked the gradients I was only using this version for inference. I do believe your PR look in a much better shape 😊 looking forward to it

mlaves:
@EduardoPach Here's a quick test to see that this PR will compute incorrect gradients:

```python
import mlx.core as mx
from mlx import nn
import torch
import numpy as np

def set_same_weights(mlx_layer, pt_layer):
    weight_shape = mlx_layer.weight.shape
    np_weight = np.ones(weight_shape).astype(np.float32)

    mlx_layer.weight = mx.array(np_weight)
    pt_layer.weight = torch.nn.Parameter(
        torch.from_numpy(np_weight).permute(-1, *range(x.ndim - 1))
    )

img_size = 8
in_channels = 6
out_channels = 3
kernel_size = 2
stride = 2
padding = 0
dilation = 1

x = np.random.randn(1, img_size, img_size, in_channels).astype(np.float32)
mlx_input = mx.array(x)
pt_input = torch.from_numpy(x).permute(0, 3, 1, 2)

mlx_convt2d = nn.ConvTranspose2d(
    in_channels=in_channels,
    out_channels=out_channels,
    kernel_size=kernel_size,
    stride=stride,
    padding=padding,
    dilation=dilation,
    bias=False,
)

pt_convt2d = torch.nn.ConvTranspose2d(
    in_channels=in_channels,
    out_channels=out_channels,
    kernel_size=kernel_size,
    stride=stride,
    padding=padding,
    dilation=dilation,
    bias=False,
)

set_same_weights(mlx_convt2d, pt_convt2d)

mlx_output = np.array(mlx_convt2d(mlx_input))
pt_output = pt_convt2d(pt_input)
pt_output_np = pt_output.detach().permute(0, 2, 3, 1).numpy()

print(mlx_output.shape, pt_output.shape)
print(np.allclose(np.array(mlx_output), pt_output_np, atol=1e-5))

def loss_fn(model, X, y):
    return mx.mean(nn.losses.mse_loss(model(X), y))

loss_and_grad_fn = nn.value_and_grad(mlx_convt2d, loss_fn)

t = torch.rand(mlx_output.shape)
loss, grad = loss_and_grad_fn(mlx_convt2d, mlx_input, mx.array(t.numpy()))

loss = torch.nn.functional.mse_loss(pt_output, t.permute(0,-1,1,2))
loss.backward()

np.allclose(grad["weight"], pt_convt2d.weight.grad.permute(1,2,3,0))  # will be False
```

awni:
> I think that without changing conv_general, this approach will yield incorrect gradients.

Good point @mlaves. IIRC we discussed not adding a new primitive but the possibility of passing the primitive the right argument. If that is possible it would be a lot less code which is nice (since most of it is duplicate). 

In that case maybe we keep #1245  and close this?

EduardoPach:
> > I think that without changing conv_general, this approach will yield incorrect gradients.
> 
> 
> 
> Good point @mlaves. IIRC we discussed not adding a new primitive but the possibility of passing the primitive the right argument. If that is possible it would be a lot less code which is nice (since most of it is duplicate). 
> 
> 
> 
> In that case maybe we keep #1245  and close this? 

Yup, I think that's the best option

awni:
@mlaves coming back to your example.. it doesn't quite make sense to me that the conv transpose operation outputs match but the gradients don't. Isn't that a bug in the gradient?

PR #1391, opened 2024-09-03
Test to prevent bugs like #1386

bhargavyagnik:
## Proposed changes

Added some test cases to prevent further issues similar to https://github.com/ml-explore/mlx/issues/1386
All test cases pass right now which is good.

## Checklist


awni (approved):
Thanks for the tests!

PR #1394, opened 2024-09-04
Fix slice data size

awni:
- Fix the data size computation in slice
- This introduced a scenario in downstream ops where they could use too much data so perform a copy when the input is not `contiguous` in `softmax`, `layer_norm` and `rms_norm`
- Add some documentation on array metadata
- More tests

Also confirmed the following benchmark uses the expected amount of RAM:

```
import mlx.core as mx

result = [mx.zeros((2, 8192, 128))[:, -1:]]
for _ in range(100):
    result.append(mx.softmax(result[-1], axis=-1))
mx.eval(result)
print(mx.metal.get_peak_memory())
```

### DIFF
--- a/mlx/backend/common/slicing.cpp
+++ b/mlx/backend/common/slicing.cpp
@@ -6,18 +6,16 @@ namespace mlx::core {
 
 std::tuple<bool, int64_t, std::vector<int64_t>> prepare_slice(
     const array& in,
-    std::vector<int>& start_indices,
-    std::vector<int>& strides) {
+    const std::vector<int>& start_indices,
+    const std::vector<int>& strides) {
   int64_t data_offset = 0;
   bool copy_needed = false;
   std::vector<int64_t> inp_strides(in.ndim(), 0);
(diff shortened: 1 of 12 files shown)


mlx/backend/common/slicing.cpp
```diff
@@ -6,45 +6,33 @@ namespace mlx::core {
...
 
 void shared_buffer_slice(
     const array& in,
     const std::vector<size_t>& out_strides,
     size_t data_offset,
+    size_t data_size,
     array& out) {
   // Compute row/col contiguity
-  auto [data_size, is_row_contiguous, is_col_contiguous] =
+  auto [no_bsx_size, is_row_contiguous, is_col_contiguous] =
```
angeloskath, line 29: My only comment would be: do we want to change the `check_contiguity` function as well to name this param differently? Just to avoid confusion wrt what `data_size` refers to.

angeloskath (approved):
Awesome. Left a nitpick address if you feel like it.

PR #1395, opened 2024-09-04
Fix jit reduce

awni:
- The JIT reduce wasn't using the template parameters properly, this fixes that.
- I think there were also some library naming issues which would cause us to rebuild libraries even when we already had built them for some kernels. So this diff also gives each kernel it's own library.
- Removed the defaults for the integer template parameters that should be set so that the compiler will catch this in the future.

Hopefully this will fix the issues we were seeing like so:

```
 -[MTLDebugComputeCommandEncoder _validateThreadsPerThreadgroup:]:1028: failed assertion `(threadsPerThreadgroup.width(1) * threadsPerThreadgroup.height(1) * threadsPerThreadgroup.depth(1))(1) must be <= 0. (kernel threadgroup size limit)'
```

Though to be honest I'm not sure why it would...


angeloskath (approved):
Thanks for fixing this!

PR #1397, opened 2024-09-07
Replace the use of `result_of_t` with `invoke_result_t`

xnorai:
## Proposed changes

Sometime after v0.16.1 code was introduced that uses `result_of_t`. That has been deprecated in C++17 and removed in C++20. This code is in the header, so people using C++20 will not be able to compile with MLX as a dep. The recommended replacement is `invoke_result_t`. It was introduced in C++17. The change is trivial, and it does not change any logic. I have tested this by installing from source, and then building with MLX as a dependency. The compile error is gone.

Note: this change implies C++17 standard will be used to build.

## Checklist

### DIFF
--- a/mlx/io/threadpool.h
+++ b/mlx/io/threadpool.h
@@ -37,7 +37,7 @@ class ThreadPool {
   ThreadPool(size_t);
   template <class F, class... Args>
   auto enqueue(F&& f, Args&&... args)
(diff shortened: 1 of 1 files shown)


angeloskath (approved):
Looks great. I 'll merge when the tests clear.

PR #1401, opened 2024-09-09
Throw for certain cases of non captured inputs in compile

### LINKED ISSUE #1399: [BUG]  Segfault running compiled function without capture
The following code segfaults. It's not a correct use of `mx.compile` since not everything is well captured. But it should not segfault.

```python
import mlx.core as mx
import mlx.nn as nn

class Cache:

    def __init__(self):
        self.c = None

    def update_and_fetch(self, c):
        if self.c is None:
            self.c = mx.zeros((8, 128))

        self.c[0 : 1] = c
        return self.c[0 : 1]

def fn(x, cache):
    y = cache.update_and_fetch(x)
    return x * y

def run():
    x = mx.random.uniform(shape=(1, 128))
    cache = Cache()
    def sample(x):
        return fn(x, cache)
    sample(x)
    y = mx.compile(sample)(x)
    mx.eval(y)


def main():
    run()

if __name__ == "__main__":
    main()
```

awni:
Closes #1399 

Also fixed the typo mentioned in #1385 and an indentation issue in the docs for `metal_kernel`.

### DIFF
--- a/mlx/compile.cpp
+++ b/mlx/compile.cpp
@@ -306,21 +306,27 @@ std::pair<std::vector<array>, std::vector<array>> compile_trace(
 // Traverses the graph to build a tape and a map of array ids to their parents
 std::pair<std::vector<array>, ParentsMap> compile_dfs(
     const std::vector<array>& inputs,
-    const std::vector<array>& outputs) {
+    const std::vector<array>& outputs,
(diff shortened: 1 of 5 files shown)


angeloskath (approved):
🙏

PR #1402, opened 2024-09-09
Fix copying scalars by adding fill_gpu

### LINKED ISSUE #1398: [BUG] pad with constant value fails in cpp
**Describe the bug**
pad with constant value always fails by `Assertion failed: (in.shape() == out.shape()), function copy_gpu_inplace, file copy.cpp, line 159.`

**To Reproduce**
```
 auto x = array({1.0f, 2.0f, 3.0f, 4.0f}, {2, 2});
 std::vector<int> axes = {0,1};
 std::vector<int> lower_padded = {1,1};
 std::vector<int>  higer_padded = {1,1};
 auto value = array(0);
 std::string mode = "constant";
 auto padded_x = pad(x, axes, lower_padded, higer_padded, value, mode);
 std::cout << padded_x << std::endl;
```
Include code snippet
```python
python works well
```

**Expected behavior**

 - OS Version: [e.g. MacOS 15.0.0]
 - Version [branch main]

**Additional context**
it may led by #1383 added `assert(in.shape() == out.shape());`

awni:
Closes #1398

### DIFF
--- a/mlx/backend/metal/conv.cpp
+++ b/mlx/backend/metal/conv.cpp
@@ -552,7 +552,7 @@ void winograd_conv_2D_gpu(
 
   // Fill with zeros
   array zero_arr = array(0, in.dtype());
-  copy_gpu(zero_arr, in_padded, CopyType::Scalar, s);
+  fill_gpu(zero_arr, in_padded, s);
   copies_w.push_back(zero_arr);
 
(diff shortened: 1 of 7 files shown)


angeloskath (approved):
Nice! That 'll save us from those slippery bugs...

PR #1403, opened 2024-09-11
Fix module attribute set, reset, set

awni:
Pretty bad bug when doing something like:

```python
model.attribute = mx.array([1])
model.attribute = None
model.attribute = mx.array([2])
assert model.attribute == None # <- would be true
````


angeloskath (approved):
Nice. Thanks!

PR #1404, opened 2024-09-11
Buffer size

angeloskath:
Adds the ability to query the allocator for the size of a buffer which enables us to add donation heuristics to limit memory abuse by donating a huge buffer while only a small portion of it is used.

The metal allocator already tracks the size of the buffer so this PR simply exposes that. It also adds the same functionality to the common allocator by allocating an extra `size_t` per buffer which holds the size of the buffer.

Finally in the binary ops I added a heuristic that doesn't allow donation if the buffer is more than 16KB bigger than needed. In the allocator we have set this to one VM page which is typically ~8KB so this heuristic mimics that one only a bit more permissive.

### DIFF
--- a/mlx/backend/common/binary.h
+++ b/mlx/backend/common/binary.h
@@ -43,13 +43,18 @@ void set_binary_op_output_data(
     array& out,
     BinaryOpType bopt,
     bool donate_with_move = false) {
+  constexpr size_t donation_extra = 16384;
+  bool b_donatable = b.is_donatable() && b.itemsize() == out.itemsize() &&
+      b.buffer_size() <= out.nbytes() + donation_extra;
+  bool a_donatable = a.is_donatable() && a.itemsize() == out.itemsize() &&
+      a.buffer_size() <= out.nbytes() + donation_extra;
   switch (bopt) {
     case BinaryOpType::ScalarScalar:
       out.set_data(
           allocator::malloc_or_wait(out.itemsize()), 1, a.strides(), a.flags());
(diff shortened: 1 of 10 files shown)


mlx/backend/common/binary.h
```diff
@@ -43,13 +43,18 @@ void set_binary_op_output_data(
     array& out,
     BinaryOpType bopt,
     bool donate_with_move = false) {
+  constexpr size_t donation_extra = 16384;
+  bool b_donatable = b.is_donatable() && b.itemsize() == out.itemsize() &&
+      b.buffer_size() <= out.nbytes() + donation_extra;
+  bool a_donatable = a.is_donatable() && a.itemsize() == out.itemsize() &&
+      a.buffer_size() <= out.nbytes() + donation_extra;
```
awni, line 50: It might be worth encapsulating this logic somewhere so that we can re-use it in unary / binary_two/ ternary / everywhere else that we have potential donation. (I think it is also possible in those cases to donate overly large buffers?)

awni (approved):
Looks great! I left a comment, not sure if it's worth addressing now or for later.

Also, I don't think this changes anything but fwiw the VM page size (`pagesize`) is also usually 16k.

mlx/backend/common/binary.h
angeloskath: Is something like that what you had in mind?

awni: Exactly 👍

PR #1405, opened 2024-09-12
More fixes for arrays with large sizes

awni:
This PR ballooned 🎈  a bit:

- Fix compile for larger size (when contiguous)
- Add size cap for contiguous collapse to not merge dimensions when the resulting shape is too big
- Always collapse singleton dimensions in contiguous collapse
- Remove loop unrolled specializations in copy and binary kernels. These were buying us virtually nothing except code and binary bloat (68MB -> 63MB with on JIT)
- Fix elem to loc for a couple cases to not down cast `size_t` inputs

### Benchmarks

TLDR no difference for real model benchmarks:

Bench | Pre | Post
---- | ---- | ---
Transformer batch 2 | It/sec 5.903 | It/sec 5.907 
Transformer batch 4 | 3.125 | It/sec 3.119
LeNet |  6.874 s| 6.880 s
MLP Size: 64 | 0.667 s | 0.645 s
MLP Size: 640 | 0.625 | 0.609 s
MLP Size: 6400 | 7.921 s | 7.910 s
Mistral 7B 16-bit prompt  | 1406.151 toks/sec |  1412.264 toks/sec
Mistral 7B 16-bit generation | 39.520 toks/sec | 39.537 toks/sec
Mistral 7B 4-bit prompt | 1286.011 toks/sec | 1293.704 toks/sec
Mistral 7B 4-bit generation |93.147 toks/sec | 93.726 toks/sec

This microbenchmark is faster because it gets routed to a 3D kernel now since singleton dims are collapsed:

```python
D = 512
a = mx.random.uniform(shape=(D, 1, D, 1, D))
b = a.transpose((4, 3, 2, 1, 0))

def fn(a):
    return [a + b for _ in range(5)]
```
```text
Pre:  14.25713 msec
Post: 10.88140 msec
```

### Tests

Added three tests to our long running CI which pass now:

```python
def compiled_function(x):
    def silu(x):
        return x * mx.sigmoid(x)

    out = mx.compile(silu)(x)
    expected = silu(x)
    assert mx.allclose(out, expected, atol=1e-3, rtol=1e-4)


def copy_general():
    shape = (4, 4096, 4096, 64)
    a = mx.random.uniform(shape=shape)
    a = a.view(mx.uint8)

    out = a[::-1]

    for s in [0, 4]:
        assert mx.array_equal(out[s, 0], a[4 - s - 1, 0])
        assert mx.array_equal(out[s, -1], a[4 - s - 1, -1])


def transposed_binary():
    shape = (2, 2, 4096, 4096, 64)
    a = mx.random.uniform(shape=shape)
    a = a.view(mx.uint8)
    b = a.swapaxes(0, 1)

    out = a + b

    slices = [
        (0, 0, 0, 0, slice(None)),
        (-1, -1, -1, -1, slice(None)),
    ]
    for s in slices:
        out_slice = a[*s] + b[*s]
        assert mx.array_equal(out[*s], out_slice)
```


awni:
> So this would still fail when we can't collapse to <4 dims and the total size is >UINT_MAX. Although rare, it wouldn't be too hard to fix that.

Yea there are still some cases in general that are broken. I can add more fixes here. For example we shouldn't collapse when it leads to overflow. Also most of the general kernels that use elem_to_loc are broken for inputs > `UINT_MAX` since they compute the index in 32 bits..

angeloskath:
Yeah 100%. Collapse contiguous dims should not collapse when it leads to overflow. It is kinda trivial to fix that actually (at least it was in the reductions). Either way I am not suggesting that you do that for this PR just saying.

awni:
@angeloskath probably good to review this again when you have a moment.

mlx/backend/common/utils.h
```diff
@@ -44,58 +44,26 @@ std::vector<stride_t> make_contiguous_strides(const std::vector<int>& shape) {
...
-
-  return std::make_tuple(out_shape, out_strides);
-}
+    const std::vector<std::vector<int64_t>>& strides,
+    int64_t size_cap = std::numeric_limits<int64_t>::max());
+std::tuple<std::vector<int>, std::vector<std::vector<size_t>>>
+collapse_contiguous_dims(
+    const std::vector<int>& shape,
+    const std::vector<std::vector<size_t>>& strides,
+    size_t size_cap = std::numeric_limits<size_t>::max());
```
angeloskath, line 56: This is a nitpick but since the collapsed dimensions will have to fit in the returned shape I think this can be set to `std::numeric_limits<int>::max()` (same below).

mlx/backend/metal/compiled.cpp
```diff
@@ -288,6 +306,15 @@ void Compiled::eval_gpu(
         collapse_contiguous_dims(output_shape, initial_strides);
```
angeloskath, line 306: Maybe `size_cap` to `INT32_MAX`?

angeloskath (approved):
This is awesome! Thanks for doing all the changes.

I left practically one comment regarding the default value for `size_cap`. Additionally, a general comment would be why remove the 4 and 5 dim specializations? Was it simply that they weren't faster? I can imagine it could be useful for `copy`...

mlx/backend/common/utils.h
awni: That's a good point. I guess it's more wishful thinking at this point. I'll update it to to that for now.

mlx/backend/common/utils.cpp
```diff
@@ -0,0 +1,90 @@
+// Copyright © 2023-2024 Apple Inc.
+
+#include "mlx/utils.h" // TODO
+#include <iostream> // TODO
```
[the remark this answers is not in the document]
awni: 😬

awni:
> Additionally, a general comment would be why remove the 4 and 5 dim specializations? Was it simply that they weren't faster?

Right, the reason I pulled those is because they are barely faster. Between 1-3% in microbenchmarks and no difference in any of the real benchmarks as you can see. What it gives you is loop unrolling over the general `elem_to_loc` which isn't buying us much.

- That didn't seem enough to justify the code / binary bloat 
- I would actually like to do a follow on with speeding up copy/binary for general dims with some specializations by doing more work per thread to amortize the cost of computing the indices. Doing this for all the different ND specializations would be too much probably.
- If you prefer I can wait to remove these kernels in a follow on.

angeloskath:
No I assumed it was something along these lines. I am glad to see them go together with a bunch of specializations of `elem_to_loc`!

Perhaps using the `looped_elem_to_loc` would help provide more work per thread with minimal additional code.

Merge this whenever you 're ready.

PR #1412, opened 2024-09-14
Ensure io/comm streams are active before eval

awni:
Previously seeing occasional segfaults due to new stream creation in the scheduler not being thread safe. This resolves that (not seeing a segfault in 20 runs now). I didn't come up with a good way to test this reliably.. so not new tests added.


angeloskath (approved):
Great job finding that!

PR #1417, opened 2024-09-16
remove fmt dependencies from MLX install

jjuang-apple:
This PR fixes the below problem when using MLX install
```
CMake Error at /Users/jjuang/test_install/share/cmake/MLX/MLXTargets.cmake:61 (set_target_properties):
  The link interface of target "mlx" contains:

    fmt::fmt-header-only

  but the target was not found.  Possible reasons include:

    * There is a typo in the target name.
    * A find_package call is missing for an IMPORTED target.
    * An ALIAS target is missing.
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -203,7 +203,7 @@ FetchContent_Declare(fmt
   EXCLUDE_FROM_ALL
 )
 FetchContent_MakeAvailable(fmt)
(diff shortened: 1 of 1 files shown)


awni (approved):
Thanks for the fix!

As discussed offline, we should revisit the headers we export.. rather than just dumping everything. Ideally it would just be the headers in `mlx.h` but we still need some internal headers to support extensions.

PR #1419, opened 2024-09-16
Faster RNN layers

awni:
Use transpose matrices to get faster inference with RNNs (same as with linear layers).

Compiled | Pre | Post
---- | ---- | ----
Yes | 15.1 ms |  10.2 ms
No | 21.3 ms | 17.3 ms

Benchmark:

```python
import time
import mlx.nn as nn
import mlx.core as mx

lstm = nn.LSTM(512, 512)
mx.eval(lstm)

x = mx.random.uniform(shape=(1, 300, 512))

@mx.compile
def fun(x):
    return lstm(x)

for _ in range(5):
    mx.eval(fun(x))

tic = time.time()
for _ in range(100):
    mx.eval(fun(x))
toc = time.time()
print(10 * (toc - tic))
```

Note this is a breaking change.. but I don't think our RNN layers are used much if at all. Probably better to change it now..

### DIFF
--- a/python/mlx/nn/layers/recurrent.py
+++ b/python/mlx/nn/layers/recurrent.py
@@ -81,7 +81,7 @@ class RNN(Module):
         all_hidden = []
         for idx in range(x.shape[-2]):
             if hidden is not None:
-                hidden = x[..., idx, :] + hidden @ self.Whh.T
+                hidden = mx.addmm(x[..., idx, :], hidden, self.Whh.T)
             else:
                 hidden = x[..., idx, :]
             hidden = self.nonlinearity(hidden)
@@ -269,7 +269,7 @@ class LSTM(Module):
         for idx in range(x.shape[-2]):
             ifgo = x[..., idx, :]
             if hidden is not None:
-                ifgo = ifgo + hidden @ self.Wh.T
+                ifgo = mx.addmm(ifgo, hidden, self.Wh.T)
             i, f, g, o = mx.split(ifgo, 4, axis=-1)
 
             i = mx.sigmoid(i)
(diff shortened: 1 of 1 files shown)


angeloskath (approved):
👍

awni:
Using admm in the inner loop gets another millisecond 🚀

barronalex:
This is really nice!

Just for fun, a custom kernel to get rid of the `mx.split` gets another ~15% faster on my M1 Max:

```
def lstm_custom(ifgo, cell):
        name="lstm",
        input_names=["ifgo", "cell", "hidden_size"],
        output_names=["hidden_state", "cell_state"],
        header="""
        template <typename T>
        T sigmoid(T x) {
            auto y = 1 / (1 + metal::exp(-metal::abs(x)));
            return (x < 0) ? 1 - y : y;
        }
        """,
        source="""
            uint elem = thread_position_in_grid.x;
            int index = elem / hidden_size * hidden_size * 4 + elem % hidden_size;

            T i = sigmoid(ifgo[index]);
            T f = sigmoid(ifgo[index + hidden_size]);
            T g = metal::precise::tanh(ifgo[index + hidden_size * 2]);
            T o = sigmoid(ifgo[index + hidden_size * 3]);

            cell_state[index] = f * cell[index] + i * g;
            hidden_state[index] = o * metal::precise::tanh(cell_state[index]);
        """,
    )
    out_shape = list(ifgo.shape)
    out_shape[-1] //= 4
    return kernel(
        inputs=[ifgo, cell, out_shape[-1]],
        output_shapes=[out_shape, out_shape],
        output_dtypes=[ifgo.dtype, ifgo.dtype],
        grid=(ifgo.size // 4, 1, 1),
        threadgroup=(256, 1, 1),
        template=[("T", ifgo.dtype)]
    )
```

barronalex:
I also toyed with parallelizing across layers as this [post](https://developer.nvidia.com/blog/optimizing-recurrent-neural-networks-cudnn-5/) suggests but for just `num_layers=2` it wasn't really any faster.

awni:
Very cool! That gets it down to 7.67ms in my benchmark!

barronalex (approved):
🚀

PR #1420, opened 2024-09-17
Build Docs: Custom Metal Kernels

cdreetz:
## Proposed changes

The site showed an old version of the docs compared to what the rst files showed. Just rebuilt and look like its fixed locally. No code updates, just pushing the build.

## Checklist


angeloskath:
@cdreetz we appreciate that and we 'll rebuild and upload asap but I don't think there is any way we can actually review 800k lines diff. Would it be ok if we closed this and rebased on on our own and pushed in an hour or so?

awni:
👍  I'll push the updated docs rn. Sorry for the delay.

awni:
FYI this is done.

cdreetz:
@angeloskath yes sorry i didnt realize it included all that other stuff.

@awni thank you awni!

PR #1421, opened 2024-09-17
Optimization for general ND copies

awni:
- Add specializations for more work per thread to amortize cost of indexing
- Use 2-array index computation for gg copy to save a bunch of divides (helps a lot)

Benchmarking general and gg copy for 5D arrays:

bench | pre | post
---- | ---- | ----
general | 118.26 ms  |  79.3 ms
gg | 230.2 ms |  107.0 ms

Benchmark below:

```python
D = 64

a = mx.zeros((D, D, D, D, D))
a = mx.transpose(a, (1, 0, 4, 2, 3))

def fn_g(a):
    return [a[..., ::-1] for _ in range(5)]

def fn_gg(a):
    return [mx.pad(a, ((0, 0), (0, 0), (0, 0), (0, 0), (4, 0))) for _ in range(5)]
```

### DIFF
--- a/mlx/backend/metal/copy.cpp
+++ b/mlx/backend/metal/copy.cpp
@@ -73,9 +73,11 @@ void copy_gpu_inplace(
         }
       };
   auto [shape, strides_in_, strides_out_] = maybe_collapse();
+  int ndim = shape.size();
 
   bool use_2d = out.data_size() > UINT32_MAX;
   auto& d = metal::device(s.device);
+  int work_per_thread = 1;
   std::string kernel_name;
   {
     std::ostringstream kname;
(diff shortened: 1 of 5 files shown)


angeloskath (approved):
Very nice! And for the cherry on top here is the results for a nearest upsampling benchmark which will be quite common.

| bench    | pre  | post |
| -------- | ------- |-------|
| upsample 2x  | 1.82 ms    | 0.87 ms |
| upsample 4x  |  6.21 ms   | 2.22 ms |

and the benchmark code

```python
D = 64

b = mx.zeros((D, D, D, D))

def upsample(b, s):
    b = b[:, :, None, :, None, :]
    b = mx.broadcast_to(b, (D, D, s, D, s, D))
    b = b.reshape(D, s*D, s*D, D)
    return b

time_fn(upsample, b, 2, msg="upsample 2x")
time_fn(upsample, b, 4, msg="upsample 4x")
```

PR #1434, opened 2024-09-24
Faster cpu ops

awni:
- Use a contiguous iterator for faster copy, unary, binary, ternary 
- Some fixes for large arrays (use size_t in most places now)
- Remove some dim specializations to reduce bloat and binary size
- Add vector specialization for ternary
- Apply dimension collapsing for all the above ops

### Benchmarks

Everything is way faster up to arbitrary dimensions. A few benchmarks:

Bench | Pre | Post |
--- | --- | ---
Copy 7D | 5.811 | 5.867
Copy 8D | 11.734 | 11.017
Binary |273.774  | 5.9
Unary | 16.071 | 0.850

Note the copy is about the same because it had a 7D specialization. Now it only has up to 3D specialized.

```python
mx.set_default_device(mx.cpu)

def copy(x):
    return x[..., ::-1]

def binary(x, y):
    return x + y

def unary(x):
    return mx.negative(x)

D = 10
x = mx.zeros((D, D, D, D, D, D, D))
x = mx.transpose(x, (1, 0, 4, 2, 3, 6, 5))
timeit(copy, x)

y = mx.zeros((D, D, D, D, D, D, D))
timeit(binary, x, y)

z = mx.random.uniform(shape=(12,)*6)[..., ::2]
mx.eval(unary(z))
timeit(unary, z)
```

### DIFF
--- a/mlx/backend/common/copy.cpp
+++ b/mlx/backend/common/copy.cpp
@@ -26,344 +26,6 @@ void copy_vector(const array& src, array& dst) {
   std::copy(src_ptr, src_ptr + src.data_size(), dst_ptr);
 }
 
-template <typename SrcT, typename DstT, typename stride_t>
-void copy_general_dim1(
-    const array& src,
-    array& dst,
-    const std::vector<int>& data_shape,
-    const std::vector<stride_t>& i_strides,
-    int64_t i_offset) {
-  const SrcT* src_ptr = src.data<SrcT>();
-  DstT* dst_ptr = dst.data<DstT>();
-  stride_t src_idx = i_offset;
-  stride_t dst_idx = 0;
-  for (int i = 0; i < data_shape[0]; ++i) {
-    dst_ptr[dst_idx++] = static_cast<DstT>(src_ptr[src_idx]);
-    src_idx += i_strides[0];
-  }
-}
-
-template <typename SrcT, typename DstT>
-inline void copy_general_dim1(const array& src, array& dst) {
-  return copy_general_dim1<SrcT, DstT, size_t>(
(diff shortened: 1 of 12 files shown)


angeloskath (approved):
It took some reading cause this moves some code that was added to fix an oob access in the copy. I am pretty sure it just moved further down in copy so no regression there let me know if I missed something.

Otherwise it looks great. Love to see all those red lines in the diff.

awni:
> It took some reading cause this moves some code that was added to fix an oob access in the copy.

Which code are you referring to? I can double check that I didn't mess that up.

mlx/backend/common/primitives.cpp
```diff
@@ -406,16 +406,7 @@ void Reshape::eval(const std::vector<array>& inputs, array& out) {
...
-    copy_inplace<size_t>(
-        in,
-        out,
-        in.shape(),
-        in.strides(),
-        out_strides,
-        0,
-        0,
-        CopyType::General);
+    copy_inplace(in, out, CopyType::General);
```
angeloskath, line 409: This is the diff I was referring to. Git commit 03cf033f82 .

mlx/backend/common/copy.cpp
```diff
@@ -26,463 +26,115 @@ void copy_vector(const array& src, array& dst) {
...
-        new_strides[1],
-        src_offset,
-        dst_offset);
-  }
+  copy_general_general<SrcT, DstT, StrideT>(
+      src,
+      dst,
+      data_shape,
+      i_strides,
+      make_contiguous_strides<StrideT>(data_shape),
```
angeloskath, line 123: But it should be fine due to that.

PR #1436, opened 2024-09-25
Dtype is copy assignable

awni:
It's convenient to make it copy assignable. I don't think we lose anything (?), using methods in C++ is also a bit more consistent with everything else in the API.

### DIFF
--- a/mlx/utils.cpp
+++ b/mlx/utils.cpp
@@ -58,11 +58,11 @@ inline void PrintFormatter::print(std::ostream& os, complex64_t val) {
 PrintFormatter global_formatter;
 
 Dtype result_type(const std::vector<array>& arrays) {
-  std::vector<Dtype> dtypes(1, bool_);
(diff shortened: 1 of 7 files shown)


mlx/utils.cpp
```diff
@@ -58,11 +58,11 @@ inline void PrintFormatter::print(std::ostream& os, complex64_t val) {
 PrintFormatter global_formatter;
 
 Dtype result_type(const std::vector<array>& arrays) {
-  std::vector<Dtype> dtypes(1, bool_);
+  Dtype t = bool_;
   for (auto& arr : arrays) {
-    dtypes.push_back(promote_types(dtypes.back(), arr.dtype()));
+    t = promote_types(t, arr.dtype());
```
angeloskath, line 63: ❤️

angeloskath (approved):
Looks great!

PR #1437, opened 2024-09-25
Some overhead reductions in mx.fast.metal_kernel

awni:
Benchmark:

Pre: 1.12 ms
Post: 0.28 ms

I think we could do more work at "compile" time to get it down further.. but this is a good step / low hanging fruit.


barronalex (approved):
Looks great!

Issue #1438, opened 2024-09-26
[BUG] mlx 0.17.3 crashes when converting certain models with mlx-lm


M-I:
**Describe the bug**

Trying to convert  deepseek-ai/DeepSeek-V2.5 and Qwen/Qwen2.5-72B-Instruct ends with failure, when deepseek-ai/DeepSeek-Coder-V2-Lite-Instruct succeeds.  Reverting to 0.17.1 with pip install mlx==0.17.1 fixed it.

**To Reproduce**

Include code snippet
```python
from mlx_lm import load, generate, convert
convert("deepseek-ai/DeepSeek-V2.5", q_bits=2, quantie=true)
```
or  `mlx_lm.convert --hf-path deepseek-ai/DeepSeek-V2.5 -q --q-bits 2`

With mlx 0.17.3, the output is:
```
[INFO] Loading 
Fetching 62 files: 100%|______________________________________________________________________________| 62/62 [00:00<00:00, 14207.10it/s
[INFO] Quantizing
libc++abi: terminating due to uncaught exception of type std::runtime_error: [METAL] Command buffer execution failed: Caused GPU Timeout Error (00000002:kIOGPUCommandBufferCallbackErrorTimeout)
 zsh: abort      HF_HOME=/Volumes/barr_1TB_1/huggingface python3
```
Then back in zsh it ouputs:
```
m-i@MBP  :$> /Users/m-i/miniconda3/envs/mlx/lib/python3.11/multiprocessing/resource_tracker.py:254: UserWarning: resource_tracker: There appear to be 1
leaked semaphore objects to clean up at shutdown
  warnings.warn('resource_tracker: There appear to be %d '
```

**Expected behavior**

**Desktop (please complete the following information):**
 - OS Version: MacOS 15.0
 - Version 0.17.3

awni:
What hardware are you running this on?

M-I:
MBP M3 Max 128GB

M-I:
@awni Since you uploaded https://huggingface.co/mlx-community/DeepSeek-V2.5-MLX-AQ4_1_64 on the 7/09/2024, I suppose it was with mlx 0.17.2 at most? I can try 0.17.2 if you want, but since 0.17.1 works for me, maybe it happened between 0.17.2 and 0.17.3?

awni:
Hi @M-I I'm having no trouble quantizing the Qwen 2 72B model this on my M1 Max with 32GB:

```
[...]

PR #1442, opened 2024-09-27
avoid io timeout for large arrays

awni:
When reading very large arrays the GPU shared event has the potential for timing out.

So e.g. the following works now:

```python
#x = mx.zeros((4096, 4096, 1024))
#mx.save_safetensors("model", {"x":x})

outs = mx.load("model.safetensors")
mx.eval(outs)
```

No perceivable difference in IO benchmarks that I tested.

May close / help with #1438.


angeloskath (approved):
👍

PR #1444, opened 2024-09-27
Add support for grouped 1D convolutions to the nn API

lucasnewman:
## Proposed changes

This enables the groups parameter for nn.Conv1D and fixes the weights creation to handle the groups.

I realize the backwards pass isn't supported with groups > 1, but this API is already exposed through the functional API so I don't think it's necessarily a worse user experience to have the forward pass available in the nn API as well.

## Checklist

### DIFF
--- a/python/mlx/nn/layers/convolution.py
+++ b/python/mlx/nn/layers/convolution.py
@@ -25,6 +25,8 @@ class Conv1d(Module):
         padding (int, optional): How many positions to 0-pad the input with.
             Default: ``0``.
         dilation (int, optional): The dilation of the convolution.
+        groups (int, optional): The number of groups for the convolution.
+            Default: ``1``.
         bias (bool, optional): If ``True`` add a learnable bias to the output.
             Default: ``True``
     """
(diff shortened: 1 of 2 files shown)


python/mlx/nn/layers/convolution.py
```diff
@@ -45,25 +46,29 @@ def __init__(
         self.weight = mx.random.uniform(
             low=-scale,
             high=scale,
-            shape=(out_channels, kernel_size, in_channels),
+            shape=(out_channels, kernel_size, int(in_channels / groups)),
```
awni, line 49: I think it would be good to add a check above if `in_channels` is not divisible by groups and raise a `ValueError`.

awni:
Looks good thanks! Could you address the comment then we can test/merge?

python/mlx/nn/layers/convolution.py
lucasnewman: Sure thing, I added it.

angeloskath: It is an absolute nitpick but perhaps also use `in_channels // groups` instead of `int(in_channels / groups)` here?

lucasnewman: Yep, no problem, I updated it.

awni (approved):
Thanks!!

PR #1446, opened 2024-09-30
Fix building on Linux

zcbenz:
## Proposed changes

Set `MLX_BUILD_METAL` only for arm64 mac, and set `MLX_BUILD_ARM` for arm on all platforms.

## Checklist


awni (approved):
Thanks for fixing that!

PR #1449, opened 2024-10-01
Conv grad with groups + bugfix

awni:
- Enable convolution gradient for groups > 1
- Fix bug with flipped grouped convolutions 
- Add some tests
- Refactor out duplicate matmul code


angeloskath (approved):
🚀

Issue #30, opened 2023-12-06
[feature request] Plans for linear solvers in MLX?


sck-at-ucy:
It would be awesome if linear algebra operations could be supported directly in MLX, for example the equivalent of PyTorch linalg.solve, which PyTorch currently supports only on CPU but not on mps .

Thank you.

Maalvi14:
`mx.linalg.solve` added on #1451 . Issue can be closed.

PR #1451, opened 2024-10-03
CPU LU factorization and linear solvers

abeleinin:
## Proposed changes

Adds `mx.linalg.solve` backend following the [numpy implementation](https://numpy.org/doc/2.0/reference/generated/numpy.linalg.solve.html). 

For the LAPACK routines, I chose to use a combination of `sgetrf` and `sgetrs`, which is equivalent to calling `sgesv` (https://stackoverflow.com/a/36065619) because it removes a transpose of `a` in the C++.

## Checklist


awni:
This looks good to me. I'm debating with myself if we should try to do this in high-level ops rather than add a new primitive for it.

On the one-hand it is definitely going to be way less efficient. On the other hand it will run on the GPU and have transform support out of the box.

I'm also ok landing this and rearranging the back-end at a later time if ti makes sense.

abeleinin:
Were you thinking of an approach using `linalg.lu_factor` and `linalg.solve_triangular` instead? Then `linalg.solve` could just use those ops. I think @barronalex mentioned possibly working on `linalg.solve_triangular` in a previous review, so maybe those linalg ops should be a higher priority than just `linalg.solve` support.

awni:
> Then linalg.solve could just use those ops. I think @barronalex mentioned possibly working on linalg.solve_triangular

I think exactly that would be better and more flexible than what we have now. I was even wondering if we could do `linalg.solve_triangular` using existing MLX ops rather than wrap the lapack routine.

abeleinin:
Ok, working on adding a `LUF` primitive and subsequent `linalg.lu_factor`, `linalg.solve_triangular`, and `linalg.solve` functions.

abeleinin:
Added `LUF` primitive which adds support for `linalg.lu` and `linalg.lu_factor` functions. 

Also added two linear solvers, `linalg.solve_triangular` and `linalg.solve` using higher-level ops.

This pr closes #30

sck-at-ucy:
This will be really useful, thank you.

acsweet:
I was wondering how close this is to being complete?
It would be great for the Keras backend.

awni:
We should be able to land it soon.

awni (approved):
Thanks for the addition!

PR #1452, opened 2024-10-03
Fix normalization check_input

angeloskath:
This PR does 2 things:

1. Fixes a segfault that came from the normalization gradient due to getting a reference to `copies.back()` which was invalidated when `copies.push_back(...)` was called for a 2nd time.
2. Normalizes the callback code by checking if we have copies to clear (unless we always do) and moves the copies vector into the lambda instead of copying it.


awni (approved):
Nice find + fix, thanks!

PR #1455, opened 2024-10-05
Add the roll op

### LINKED ISSUE #1454: [FEATURE] MLX equivalent of torch.roll
https://pytorch.org/docs/stable/generated/torch.roll.html

angeloskath:
Adds the roll operation using `split` and `concatenate`. Closes #1454 .

### DIFF
--- a/python/src/ops.cpp
+++ b/python/src/ops.cpp
@@ -4795,4 +4795,54 @@ void init_ops(nb::module_& m) {
       Returns:
         array: The output array.
     )pbdoc");
+  m.def(
+      "roll",
+      [](const array& a,
+         const IntOrVec& shift,
+         const IntOrVec& axis,
+         StreamOrDevice s) {
+        return std::visit(
+            [&](auto sh, auto ax) -> array {
+              using T = decltype(ax);
+              using V = decltype(sh);
+
+              if constexpr (std::is_same_v<V, std::monostate>) {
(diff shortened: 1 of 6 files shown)


python/src/ops.cpp
```diff
@@ -4795,4 +4795,54 @@ void init_ops(nb::module_& m) {
...
+
+        If the axis is not provided the array is flattened, rolled and then the
+        shape is restored.
+
+        Args:
+          a (array): Input array
+          shift (int or tuple(int)): The number of places by which elements
+            are shifted. If positive the array is rolled to the right, if
+            negative it is rolled to the left. If an int is provided but the
+            axis is a tuple then the same value is used for all axes.
```
awni, line 4844: The numpy docs are either confusing or wrong:

> If a tuple, then axis must be a tuple of the same size, and each of the given axes is shifted by the corresponding number.

But the actual behavior suggests you can provide a tuple of shifts and a single axis and it sums them up, like what you implemented...

awni (approved):
Very nice! Thanks for the addition.

python/src/ops.cpp
angeloskath: Yeah weird right? I initially was throwing if `len(shift) > len(axis)` but then I thought it costs us nothing to be numpy compatible... so I added it.

PR #1460, opened 2024-10-06
Fix complex power on Metal

### LINKED ISSUE #1458: [BUG] Is this a bug when raising complex numbers to a power?
**Describe the bug**
Raising to a power complex64 does not work as expected in case of base having a negative real part.

**To Reproduce**

Include code snippet
```
import mlx.core as mx
import numpy as np
a = [-1.2+0.1j]
a_m = mx.array(a)
a_n = np.array(a)
print(a_m**2.2)
print(a_n**2.2)```

**Expected behavior**
MLX gives a wrong result calculating (-1.2+0.1j)**2.2

 - OS Version: MacOS 14.6.1 ]
 - Version: 0.18.0

awni:
Closes #1458

### DIFF
--- a/mlx/backend/metal/kernels/binary_ops.h
+++ b/mlx/backend/metal/kernels/binary_ops.h
@@ -217,7 +217,7 @@ struct Power {
 
   template <>
   complex64_t operator()(complex64_t x, complex64_t y) {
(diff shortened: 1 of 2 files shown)


angeloskath (approved):
Thanks!

PR #1461, opened 2024-10-06
Bump nanobind to 2.2

### LINKED ISSUE #1380: [BUG] `Invalid type  ndarray` for complex, non-contiguous `np.array` input to `mx.array`
**Describe the bug**
Cannot initialize `mx.array` from a complex, non-contiguous `np.array`, and the error message is confusing.

**To Reproduce**

```python
import numpy as np
import mlx.core as mx


contiguous_array = np.random.rand(10, 10).astype(np.complex64)

# Make the array non-contiguous by transposing it
non_contiguous_array = contiguous_array.T

# Ensure the array is non-contiguous
assert not non_contiguous_array.flags["C_CONTIGUOUS"]

# Attempt to move the non-contiguous array to the GPU
gpu_array = mx.array(non_contiguous_array)
# Perform a simple operation just to check
result = mx.sum(gpu_array)
```

**Expected behavior**
Expected to run without errors.

Actual error:
```
  File ".../mlx_complex.py", line 14, in <module>
    gpu_array = mx.array(non_contiguous_array)
ValueError: Invalid type  ndarray received in array initialization.
```

Note: workarounds are:
* forcing the array to be contiguous with `np.ascontiguousarray`, which works, but took me a while to figure out that this is a requirement
* use real-valued array (okay in this toy example, but I am interested in using this for complex data)

 - OS Version: Sonoma 14.6.1 (23G93)
 - Version: 0.16.3
```
❯ pip list | grep mlx
mlx                                               0.16.3
❯ pip list | grep numpy  
numpy                                             1.26.4
```

**Additional context**

awni:
- Bump nanobind
- Closes #1380

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -31,7 +31,7 @@ jobs:
           name: Install dependencies
           command: |
             pip install --upgrade cmake
-            pip install nanobind==2.1.0
+            pip install nanobind==2.2.0
             pip install numpy
             sudo apt-get update
             sudo apt-get install libblas-dev liblapack-dev liblapacke-dev
@@ -83,7 +83,7 @@ jobs:
             source env/bin/activate
             pip install --upgrade pip
             pip install --upgrade cmake
-            pip install nanobind==2.1.0
+            pip install nanobind==2.2.0
             pip install numpy
             pip install torch
             pip install tensorflow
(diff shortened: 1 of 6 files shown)


awni:
The extension version should technically be pinned at 18.1 / the next MLX release but I will keep it at 18.0 so our tests clear for the time being and bump it at the next version bump.

angeloskath (approved):

PR #1462, opened 2024-10-06
Fix deep recursion with siblings

awni:
This should fix recursing during destruction with siblings creating a very deep recursion.

Close https://github.com/ml-explore/mlx-swift/issues/139

Added some tests to the long running CI:

```python
def deep_graph():
    # Deep graph destroyed without eval
    x = mx.array([1.0, 2.0])
    for _ in range(100_000):
        x = mx.sin(x)
    del x

    # Deep graph with siblings destroyed without eval
    x = mx.array([1, 2])
    for _ in range(100_000):
        x = mx.concatenate(mx.split(x, 2))
    del x

    # Deep graph with eval
    x = mx.array([1.0, 2.0])
    for _ in range(100_000):
        x = mx.sin(x)
    mx.eval(x)
```

### DIFF
--- a/mlx/array.cpp
+++ b/mlx/array.cpp
@@ -245,7 +245,7 @@ array::ArrayDesc::~ArrayDesc() {
   std::vector<std::shared_ptr<ArrayDesc>> for_deletion;
 
   for (array& a : inputs) {
-    if (a.array_desc_.use_count() == 1) {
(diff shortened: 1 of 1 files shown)


angeloskath (approved):
Thanks for the fix!

The tests seem fast enough for the normal CI as well don't they? Either way it's fine,

awni:
> The tests seem fast enough for the normal CI as well don't they?

I think you're right. I'll add them here instead.

PR #1468, opened 2024-10-07
Fix array is_available race cases

awni:
There was previously a race condition with the use of `is_available` on an array. The `is_available` check did not guarantee the array was ready, only that it had `eval_*` called. This attempts to fix that.

- Add a `Status::evaluated` for arrays which have had `eval_gpu` called but are not ready. The purpose of that status over say `Status::scheduled` is to know if it's safe to detach an array on the next eval if it was previously eval'ed as a tracer.
- `Status::available` now a sufficient but not necessary condition for the array to be ready.
- Add `is_signaled` to events for a quick wait-free way to check if the even is ready.
- Use `is_signaled` along with the array status to implement `is_available`
- Add `array::wait` as a simple helper to wait on an array and set its status correctly when done waiting.

Added a test which would previously fail almost every time I ran it.


CMakeLists.txt
```diff
@@ -47,13 +47,13 @@ if(${CMAKE_SYSTEM_NAME} MATCHES "Darwin")
...
     else()
+      set(MLX_BUILD_METAL OFF)
       message(WARNING "Building for x86_64 arch is not officially supported.")
     endif()
-  elseif(${CMAKE_SYSTEM_PROCESSOR} MATCHES "arm64")
-    set(MLX_BUILD_METAL ON)
   endif()
 
 else()
+  set(MLX_BUILD_METAL OFF)
```
awni, line 56: A previous change broke the ability to specify `MLX_BUILD_METAL` since it always set it to `ON`. This fixes that.

angeloskath (approved):
Looks good!

PR #1471, opened 2024-10-08
Add `mx.random.permutation`

awni:
As title.

### DIFF
--- a/mlx/random.cpp
+++ b/mlx/random.cpp
@@ -463,15 +463,14 @@ array permutation(
     int axis /* = 0 */,
     const std::optional<array>& key /* = std::nullopt */,
     StreamOrDevice s /* = {} */) {
-  auto indices = argsort(bits({x.shape(axis)}, key, s), s);
-  return take(x, indices, axis, s);
+  return take(x, permutation(x.shape(axis), key, s), axis, s);
 }
 
 array permutation(
     int x,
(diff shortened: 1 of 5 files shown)


mlx/random.cpp
```diff
@@ -458,4 +458,20 @@ array laplace(
...
+  auto indices = argsort(bits({x.shape(axis)}, key, s), s);
+  return take(x, indices, axis, s);
+}
+
+array permutation(
+    int x,
+    const std::optional<array>& key /* = std::nullopt */,
+    StreamOrDevice s /* = {} */) {
+  return permutation(arange(x, s), 0, key, s);
+}
```
angeloskath, line 475: Total nitpick but I would swap the two as follows:

```cpp
array permutation(
    const array& x,
    int axis /* = 0 */,
    const std::optional<array>& key /* = std::nullopt */,
  auto indices = permutation(x.shape(axis), key, s);
  return take(x, indices, axis, s);
}

array permutation(
    int x,
    const std::optional<array>& key /* = std::nullopt */,
  return argsort(bits({x}, key, s), s);
}
```

angeloskath (approved):
Left a small nitpick, otherwise looks great!

mlx/random.cpp
awni: Oh that's so much better.

PR #1477, opened 2024-10-10
patch bump 0.18.1

awni:
🚀


examples/extensions/requirements.txt
```diff
@@ -1,4 +1,4 @@
 setuptools>=42
 cmake>=3.24
-mlx>=0.18.0
+mlx>=0.18.1
```
awni, line 3: This will cause the tests to fail. But after the release they should pass so I will rerun them then.

angeloskath (approved):

PR #1478, opened 2024-10-11
Make the GPU device more thread safe

awni:
Multiples GPU streams was completely broken due to race conditions on several shared data structures in the device (kernels, libraries, etc).

This refactors and adds some shared mutexes to fix it. In terms of the Metal device public methods, the main change is to remove stuff and provide a single mechanism to get a library `get_library(name, builder_function)` so we can control access to the underlying data structure without the possibility for  a race.

- [] run some benchmarks to make sure no impact

Note, MLX is still pretty far from being thread safe. Some things we would need to fix:

- Stream creation is not thread safe or async eval safe. We assume stream creation happens on the main thread and if a new stream is created during an async eval that is a race.
- CPU compilation is not thread safe and also currently not stream safe. The latter should be fixed as it is a bug.
- RNG state is not thread safe but is stream and async eval safe since it's not accessed during eval.
- Compiled function cache is not thread safe but also not accessed during eval

### DIFF
--- a/mlx/backend/metal/device.cpp
+++ b/mlx/backend/metal/device.cpp
@@ -440,12 +440,13 @@ MTL::Library* Device::get_library_(const std::string& name) {
 MTL::Library* Device::get_library(
     const std::string& name,
     const std::function<std::string(void)>& builder) {
-  std::shared_lock rlock(library_mtx_);
-  if (auto it = library_map_.find(name); it != library_map_.end()) {
-    return it->second;
+  {
+    std::shared_lock rlock(library_mtx_);
+    if (auto it = library_map_.find(name); it != library_map_.end()) {
+      return it->second;
+    }
   }
 
-  rlock.unlock();
   std::unique_lock wlock(library_mtx_);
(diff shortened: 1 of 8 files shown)


mlx/backend/metal/device.cpp
```diff
@@ -465,68 +431,31 @@ MTL::ComputePipelineState* Device::get_kernel_(
...
+  std::shared_lock rlock(library_mtx_);
+  if (auto it = library_map_.find(name); it != library_map_.end()) {
+    return it->second;
   }
 
-  auto mtl_lib = get_library_(desc);
-
-  if (cache) {
-    library_map_.insert({name, mtl_lib});
+  rlock.unlock();
```
barronalex, line 448: Personal preference: I feel like the locking/unlocking code is a bit easier to read with scopes rather than calling `unlock` directly.
e.g.
```suggestion
 {
   std::shared_lock rlock(library_mtx_);
   ...
 }
 {
   std::unique_lock wlock(library_mtx_);
   ...
 }
```

barronalex (approved):
Awesome! Nice improvement and it's a negative diff.

PR #1482, opened 2024-10-13
A few updates for CPU

awni:
- Fix streams in affine quantize
- Bulk eval in safetensors to avoid unnecessary serialization of work
- Stream safety in CPU compilation
- Fix CPU compile preamble  for some linux machines
- Few other minor nits


angeloskath (approved):
Thanks!

PR #1486, opened 2024-10-14
Faster metal compiled kernels + some fixes

awni:
- More work per thread for ndim > 3 compiled kernels to amortize cost of indexing
- Fix compile for non-contiguous large array sizes (> uint32_max). Add test to long running CI
- Speed up indexing computations for dynamic dim case (ndim > 8)
- Always use work per thread > 1 for non contiguous elementwise ops (unary, binary, copy, etc). Not sure why I had the dim=1 specialization in there. It's always faster when the last dimension > 1 and collapsing should remove the case when the last dimension is 1. (Added benefit it reduces binary size by 1MB)
- Fix bug in compile for ndim > 8 (dynamic dims) + add test

### Benchmarks

On M1 Max

Bench | Pre (ms) | Post (ms)
---- | ---- | -----
Strided compile dim=6 | 22.961 | 9.387
Strided compile dim=9 | 476.715 | 213.611
Binary last dim = 2 | 3.324 | 2.007
Binary last dim = 4 | 6.387 | 2.683

Code for benchmarks:

```python
D = 24
a = mx.random.uniform(shape=(D,)*6)
a = a.transpose(1, 0, 3, 2, 5, 4)
b = mx.random.uniform(shape=(D,)*6)
mx.eval(a, b)


def fn(a, b):
    return mx.negative(a + b)

timeit(mx.compile(fn), a, b)

D = 8
a = mx.random.uniform(shape=(D,)*9)
a = a.T
b = mx.random.uniform(shape=(D,)*9)
mx.eval(a, b)

def fn(a, b):
    return mx.negative(a + b)

timeit(mx.compile(fn), a, b)

D = 28
a = mx.random.uniform(shape=(D,)*4 + (2, D))
a = a.transpose(1, 0, 3, 2, 5, 4)
b = mx.random.uniform(shape=(D,)*5 + (2,))
mx.eval(a, b)

def fn(a, b):
    return a + b

timeit(fn, a, b)

D = 28
a = mx.random.uniform(shape=(D,)*4 + (4, D))
a = a.transpose(1, 0, 3, 2, 5, 4)
b = mx.random.uniform(shape=(D,)*5 + (4,))
mx.eval(a, b)

def fn(a, b):
    return a + b

timeit(fn, a, b)
```

### DIFF
--- a/mlx/backend/metal/compiled.cpp
+++ b/mlx/backend/metal/compiled.cpp
@@ -13,6 +13,8 @@
 
 namespace mlx::core {
 
+constexpr int WORK_PER_THREAD = 4;
+
 inline void build_kernel(
     std::ostream& os,
     const std::string& kernel_name,
@@ -23,7 +25,8 @@ inline void build_kernel(
     bool contiguous,
     int ndim,
     bool dynamic_dims,
-    bool use_big_index = false) {
+    bool use_big_index = false,
+    int work_per_thread = 1) {
   // All outputs should have the exact same shape and will be row contiguous
   auto output_shape = outputs[0].shape();
   auto output_strides = outputs[0].strides();
@@ -38,8 +41,8 @@ inline void build_kernel(
   int cnt = 0;
 
   // Start the kernel
-  os << "[[host_name(\"" << kernel_name << "\")]]" << std::endl
-     << "[[kernel]] void " << kernel_name << "(" << std::endl;
+  os << "[[host_name(\"" << kernel_name << "\")]]\n"
+     << "[[kernel]] void " << kernel_name << "(\n";
 
   // Add the input arguments
   for (auto& x : inputs) {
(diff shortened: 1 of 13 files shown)


mlx/backend/metal/compiled.cpp
```diff
@@ -38,8 +41,8 @@ inline void build_kernel(
   int cnt = 0;
 
   // Start the kernel
-  os << "[[host_name(\"" << kernel_name << "\")]]" << std::endl
-     << "[[kernel]] void " << kernel_name << "(" << std::endl;
+  os << "[[host_name(\"" << kernel_name << "\")]]\n"
+     << "[[kernel]] void " << kernel_name << "(\n";
```
angeloskath, line 45: Do the explicit newlines make it easier to follow/smaller?

angeloskath (approved):
This looks really good and has awesome perf gains!

mlx/backend/metal/compiled.cpp
awni: For ostringstream `std::endl` gets resolved to `\n`. My only rationale here is that I think `os << "a" << "\n"` is a lot slower (in the worst case) than `os << "a\n"`.

I was thinking to change all of it to string concatenation per our previous conversation.. but I didn't want to blow up this diff. It probably has a small to negligible impact on startup cost but honestly I haven't measured it yet so 🤷‍♂️ .

PR #1492, opened 2024-10-15
Fix submodule stubs

awni:
The submodules stubs don't import the packages they need so the types in the signatures previously would show up as `Any`.

This fixes that by adding the imports.

Some comments for future fixes here:
- Nanobind has some facility to auto-detect and include imports but it would require writing the signatures with fully qualified names (e.g. `typing.Union` instead of just `Union`) which gets quite verbose and annoying. 

- The alternative would be to let nanobind auto-generate the signatures.. this is not a bad option. The only downside is some of the variants can be extremely long / verbose and I don't see a way to have aliases for those.

### DIFF
--- a/python/src/linalg.cpp
+++ b/python/src/linalg.cpp
@@ -187,7 +187,7 @@ void init_linalg(nb::module_& parent_module) {
       nb::kw_only(),
       "stream"_a = nb::none(),
       nb::sig(
-          "def qr(a: array, *, stream: Union[None, Stream, Device] = None) -> tuple(array, array)"),
+          "def qr(a: array, *, stream: Union[None, Stream, Device] = None) -> Tuple[array, array]"),
       R"pbdoc(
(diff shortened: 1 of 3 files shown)


awni:
How it looks in VS Code now

Screenshot 2024-10-15 at 3 32 23 PM 2

python/src/linalg.cpp
```diff
@@ -187,7 +187,7 @@ void init_linalg(nb::module_& parent_module) {
       nb::kw_only(),
       "stream"_a = nb::none(),
       nb::sig(
-          "def qr(a: array, *, stream: Union[None, Stream, Device] = None) -> tuple(array, array)"),
```
barronalex, line 190: Slightly sad this doesn't work

barronalex (approved):
Thanks for fixing this! Nice to have the previews working.

PR #1496, opened 2024-10-17
remove hazard tracking

awni:
Draft PR to turn off hazard tracking. Removing hazard tracking opens a bunch of doos including using heaps efficiently, residency sets.

This is a draft because I'm not 100% sure it's right and I'd also like to do some benchmarking to see if it affects perf at all.

Basically we use MTL::Fence to put command buffers in order. We could of course try to do something more clever.. but so far I don't see any performance impact doing it the simple way suggesting we aren't getting much out of command buffers running in parallel.


awni:
Unfortunately training is more impacted by this change 😞 . The transformer training benchmark goes down to 5.6 its/sec from 5.9 or so. So we'll probably need something more sophisticated to manage dependencies between command buffers.

Issue #1459, opened 2024-10-05
[Feature] Sparse_ init


cvnad1:
Request to add Sparse_ initialization similar tohttps://pytorch.org/docs/stable/nn.init.html#torch.nn.init.sparse_. I have gone through the official documentation of MLX but couldn't find the above initialization method. I think it would be a good addition to MLX.

cvnad1:
@awni May I know if the above has an update?

awni:
We could put that initialization in our init package. I'll mark this as a feature request. Just curious, what do you intend to use it for?

Saanidhyavats:
@awni We were experimenting with different initialization techniques and figured out that sparse init is not implemented. I think it  can be a good enhancement feature.

cvnad1:
The PR is merged and hence closing the issue.

PR #1498, opened 2024-10-18
[Feature] Added Sparse Initialization

cvnad1:
## Proposed changes

Adding Feature Sparse Initialization as proposed in the issue #1459 

## Checklist

### DIFF
--- a/python/mlx/nn/init.py
+++ b/python/mlx/nn/init.py
@@ -349,8 +349,12 @@ def he_uniform(
 
     return initializer
 
+
 def sparse(
-    mean: float = 0.0, std: float = 1.0, sparsity: float = None, dtype: mx.Dtype = mx.float32
+    mean: float = 0.0,
+    std: float = 1.0,
+    sparsity: float = None,
+    dtype: mx.Dtype = mx.float32,
 ) -> Callable[[mx.array], mx.array]:
     r"""An initializer that returns a sparse matrix.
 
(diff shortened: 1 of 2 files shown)


Saanidhyavats:
@awni, @cvnad1 and I have added sparse init method with documentation and tests. Let us know if any change needs to be done

python/mlx/nn/init.py
```diff
@@ -348,3 +348,56 @@ def initializer(
...
+        array([[-1.91187, -0.117483],
+       [0, 0]], dtype=float32)
+    """
+
+    def initializer(a: mx.array) -> mx.array:
+
+        if sparsity == None:
+            raise ValueError(f"Sparsity argument cannot be {sparsity}")
+
+        if a.ndim != 2:
```
angeloskath, line 388: What is preventing us from initializing all arrays here? I think this check should be removed.

angeloskath (changes requested):
Thanks for the addition!

I think this implementation is unnecessarily complicated. The same can be achieved via a mask. Something as simple as

```python
v = mx.random.normal(shape=a.shape, scale=std, loc=mean, dtype=dtype)
m = mx.random.bernoulli(1-sparsity, shape=a.shape)
return m * v
```

python/mlx/nn/init.py
cvnad1: @angeloskath, I believe that `mx.random.Bernoulli` might generate matrices where entire columns consist solely of 1s or 0s. Our approach, however, emphasizes a more even distribution of sparsity across dimensions, specifically the columns in this case. Sparse initialization is primarily used for initializing 2D weights, and ensuring an even distribution of sparsity becomes increasingly complex as the number of dimensions grows. Therefore, we opted to stick with the current implementation.

angeloskath: Still the loop is a significantly inefficient way of implementing this. We will have thousands of columns and hundreds of arrays. A more efficient implementation exactly equivalent to your original one is the following:

```python
order = mx.argsort(mx.random.uniform(shape=(a.shape)), axis=0)
a = mx.random.normal(shape=a.shape, scale=std, loc=mean, dtype=dtype)
a[order[:num_zeros], mx.arange(cols)[None]] = 0
```

Another detail is that you probably want to initialize each row to be partially 0 instead of each column. In all linear layers we do `x @ w.T` so having a completely 0 row is probably the thing you are trying to avoid.

cvnad1: @angeloskath Refactored the code with the suggested changes. Could you please review it?

angeloskath (approved):

zcbenz:
@cvnad1 Sorry for pinging you on an old PR but can you take a look at #3725 to verify whether its claim is correct?

PR #1506, opened 2024-10-21
Working 64-bit scans

### LINKED ISSUE #673: [Feature] Support cumsum (and other scans) on 64-bit types
**Describe the bug**
When applying cumsum to an int64 array, there's a kernel error.

**To Reproduce**

```python
import mlx.core as mx

a = mx.array([0, 1, 2], dtype=mx.int64)
mx.cumsum(a)

>>> libc++abi: terminating due to uncaught exception of type std::runtime_error: [metal::Device] Unable to load kernel contiguous_scan_inclusive_sum_int64_int64

[1]    48047 abort      python
```

 - macOS 14.2
 - branch: main

angeloskath:
This fixes #673 . We can merge it as is to close that issue but I probably need to also revamp the kernels a bit as they seem quite inefficient and also don't support huge arrays.


angeloskath:
I think this is ready for review. The kernels may need another pass but for now they should be ok to merge.

```
Size      | Axis | Transpose | Before | After
----------+------+-----------+--------+-------
1024x1024 |    0 |    False  |  0.039 | 0.034
1024x1024 |    1 |    False  |  0.031 | 0.033
1024x1024 |    0 |    True   |  0.131 | 0.031
1024x1024 |    1 |    True   |  0.049 | 0.031
8192x8192 |    0 |    False  |  0.657 | 0.638
8192x8192 |    1 |    False  |  0.730 | 0.732
8192x8192 |    0 |    True   |  1.867 | 0.732
8192x8192 |    1 |    True   |  1.476 | 0.637
```

awni:
This looks really great! (I made a small change to the jit path to use the `get_template_definition` Alex added. It simplifies maintenance moving forward since we don't need to have the separate `scan.h` file.)

awni (approved):
🚀

PR #1509, opened 2024-10-21
Remove Hazard tracking with Fences

awni:
Removes hazard tracking and replaces it with Fences where needed - basically in between command buffers.

The basic idea is:

- Every command buffer has a unique fence
- We record a map of ouput arrays -> fence
- For the next command buffer, we look up its inputs in the map to see if we need to wait on a fence
- When command buffers finish they remove their outputs from the output->fence map in their completion handler

The only thing to be aware of for dev is that the `completionHandler` stuff for storing temporaries is replaces with `d.add_temporaries(std::move(copies), s.index)`. The change to the way we store temporaries is not critical actually.. I had it in from an earlier iteration to avoid a resource leak. Since temps never cross command buffer boundaries they are removed from the inputs/outputs that we synchronize on. I think it's kind of nice which is why I kept it as is and a bit more efficient.. but I am happy to split it / remove it / etc.

### Benchmarks on M2 Ultra

Benchmark | Pre | Post 
--- | --- | --- 
Mistral 7B generation | 113.6 toks/sec | 113.6 toks/sec
Transformer batch=2 | 5.91 its/sec | 5.90 its/sec
Transformer batch=4 | 3.13 its/sec | 3.13 its/sec
MNIST | 0.60 sec | 0.60 sec
LeNet | 7.25 sec | 7.25 sec

### DIFF
--- a/docs/src/usage/compile.rst
+++ b/docs/src/usage/compile.rst
@@ -33,12 +33,12 @@ Let's start with a simple example:
   # Compile the function
   compiled_fun = mx.compile(fun)
 
-  # Prints: array(2.36788, dtype=float32) 
+  # Prints: array(2.36788, dtype=float32)
   print(compiled_fun(x, y))
 
 The output of both the regular function and the compiled function is the same
 up to numerical precision.
-   
+
 The first time you call a compiled function, MLX will build the compute
 graph, optimize it, and generate and compile code. This can be relatively
 slow. However, MLX will cache compiled functions, so calling a compiled
@@ -96,7 +96,7 @@ element-wise operations:
 
 .. code-block:: python
 
-  def gelu(x):  
+  def gelu(x):
       return x * (1 + mx.erf(x / math.sqrt(2))) / 2
 
 If you use this function with small arrays, it will be overhead bound. If you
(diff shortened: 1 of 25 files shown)


awni:
@angeloskath I did some cleanup + added some commentary for future refence..

I'm a bit torn about putting stuff into `DeviceStream` as a methods.. on first pass it mostly adds verbosity without saving that much IMO. I think making this nicer requires a bigger refactor which I'm not yet sure how to do.

From a object interaction standpoint, when dealing with streams (queue, encoders, buffers, etc) the `Device` is basically a thin wrapper around `DeviceStream`. The `CommandEncoder` is basically only touched by the `DeviceStream`. But the interaction between `DeviceStream` and the `CommandEncoder` is coupled via the inputs/outputs. 

Some ideas for future improvement (nothing I'm too crazy about yet):
- One option to decouple that is to make `set_input_array` a method on device which takes an `encoder` as a second argument but records the inputs and outputs in the `DeviceStream`. It's kind of an annoying change..
- Another option is to just be more explicit about the fact that the `CommandEncoder` is part of the `DeviceStream`. Maybe nest it or merge it.
- Could allow access to `DeviceStream` directly rather than accessing it through the `Device` itself.

I'm leaning towards trying to get this over the finish line. Feel free to toy around with it if you have any ideas.



 In theory we could expose `DeviceStream`.. but it might be good to sleep on that.

angeloskath (approved):
Looks great. I appreciate the extra comments :-)

PR #1510, opened 2024-10-21
Wired

awni:
- Builds on #1509 to expose residency sets when available (>= OS 15).
- Exposes `mx.metal.set_wired_limit` to enable / disable wiring buffers
- The allocator will wire buffers up to the wired limit.

See MLX LM companion PR for benchmarks: https://github.com/ml-explore/mlx-examples/pull/1069


mlx/backend/metal/resident.cpp
```diff
@@ -0,0 +1,96 @@
...
+#include "mlx/backend/metal/resident.h"
+#include "mlx/backend/metal/metal_impl.h"
+
+namespace mlx::core::metal {
+
+// TODO maybe worth including tvos / visionos
+#define supported __builtin_available(macOS 15, iOS 18, *)
+
+ResidencySet::ResidencySet(MTL::Device* d) {
+  if (supported) {
```
awni, line 12: Couple things here:
- This is a runtime check because otherwise we can't use this feature in MLX Swift binaries without breaking macOS<15 support.
- It has to be `if (__builtin_available(...))` so that's why it's with the preprocessor instead of say a `bool` or something like that.

angeloskath: Does it also have to be `if (supported)`? I am asking because

```cpp
if (!supported) {
  return;
}
```

could be a tiny bit simpler.

angeloskath (approved):
Looks perfect!

mlx/backend/metal/resident.cpp
awni: It's crazy but it actually does. There might be a `__not_builtin_available` or something

awni: I can't find anything like that sadly.

PR #1514, opened 2024-10-22
Remove use of vector<const T>

dgovil:
## Proposed changes

LLVM/Clang upstream have stopped supporting `std::vector<const T>` as documented here https://reviews.llvm.org/D120996

This was not accepted by other compilers anyway, since vectors can't truly have const members. 
This PR makes MLX compatible with the change in the above link.

## Checklist


angeloskath (approved):
Thanks for the fix! I will merge after the tests clear.

PR #1519, opened 2024-10-23
C++20 compatibility for fmt

xnorai:
## Proposed changes

Before this change, if you try to compile MLX in C++20 mode, you will get errors along the lines of:

```
mlx/backend/metal/jit_kernels.cpp:459:26: error: call to consteval function 'fmt::basic_format_string<char, fmt::detail::statically_named_arg<const std::string &, char, 5, detail_exported::fixed_string<char, 5>{"name"}>, fmt::detail::statically_named_arg<std::string, char, 6, detail_exported::fixed_string<char, 6>{"atype"}>, fmt::detail::statically_named_arg<std::string, char, 6, detail_exported::fixed_string<char, 6>{"otype"}>>::basic_format_string<std::string_view, 0>' is not a constant expression
```

The solution is proposed here: https://github.com/fmtlib/fmt/issues/2438.

With these changes, code builds in C++20 mode.

## Checklist

### DIFF
--- a/mlx/backend/metal/jit_kernels.cpp
+++ b/mlx/backend/metal/jit_kernels.cpp
@@ -445,8 +445,9 @@ MTL::ComputePipelineState* get_steel_gemm_splitk_accum_kernel(
     kernel_source << metal::utils() << metal::gemm()
                   << metal::steel_gemm_splitk()
                   << fmt::format(
-                         axbpy ? steel_gemm_splitk_accum_axbpy_kernels
-                               : steel_gemm_splitk_accum_kernels,
+                         fmt::runtime(
+                             axbpy ? steel_gemm_splitk_accum_axbpy_kernels
+                                   : steel_gemm_splitk_accum_kernels),
                          "name"_a = lib_name,
                          "atype"_a = get_type_string(in.dtype()),
                          "otype"_a = get_type_string(out.dtype()));
@@ -481,7 +482,7 @@ MTL::ComputePipelineState* get_steel_gemm_masked_kernel(
     kernel_source << metal::utils() << metal::gemm()
                   << metal::steel_gemm_masked()
                   << fmt::format(
-                         steel_gemm_masked_kernels,
+                         fmt::runtime(steel_gemm_masked_kernels),
                          "name"_a = lib_name,
                          "itype"_a = get_type_string(out.dtype()),
                          "outmasktype"_a = out_mask_type,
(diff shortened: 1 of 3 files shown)


awni:
I tried compiling with C++20 and the current code didn't complain. Just curious, which compiler are you using?

mlx/backend/metal/kernels.h
```diff
@@ -212,7 +212,7 @@ get_template_definition(std::string name, std::string func, Args... args) {
   std::string base_string = R"(
 template [[host_name("{0}")]] [[kernel]] decltype({1}) {1};
   )";
-  return fmt::format(base_string, name, s.str());
+  return fmt::format(fmt::runtime(base_string), name, s.str());
```
awni, line 215: I think this one is easy to fix without disabling the run time check by puting the format string as the argument In `fmt::format`?

mlx/backend/metal/jit_kernels.cpp
```diff
@@ -481,7 +482,7 @@ MTL::ComputePipelineState* get_steel_gemm_masked_kernel(
     kernel_source << metal::utils() << metal::gemm()
                   << metal::steel_gemm_masked()
                   << fmt::format(
-                         steel_gemm_masked_kernels,
+                         fmt::runtime(steel_gemm_masked_kernels),
```
awni, line 485: This should already be a constexpr 🤔  https://github.com/ml-explore/mlx/blob/main/mlx/backend/metal/jit/steel_gemm.h#L26

xnorai: Looks like you're right. I was patching against an earlier revision. Removed.

mlx/backend/metal/kernels.h
xnorai: Done.

xnorai:
@awni my understanding is CMake will use Clang, and so will Bazel. The version is:

```
clang --version
Apple clang version 16.0.0 (clang-1600.0.26.3)
Target: arm64-apple-darwin24.0.0
Thread model: posix
InstalledDir: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin
```

To test C++20 build, clone main and then do the following:
- Replace `CMAKE_CXX_STANDARD` with `20`
- Run `CMAKE_BUILD_PARALLEL_LEVEL=8 pip install -e ".[dev]"` (in a venv)

Unpatched build will produce an extensive error spew, and it will fail to build.

xnorai (commented):
Addressed feedback. PTAL

xnorai:
Maybe we should set this to C++20 in the next PR so it doesn't drift?

awni:
> Maybe we should set this to C++20 in the next PR so it doesn't drift?

Yes I'm supportive of that. We still run CI with xcode 15.0 so there are [some C++20 features](https://developer.apple.com/xcode/cpp/#c++20) we can't use yet. But I think it's fine to set the version to 20 in general.

awni (approved):
Thanks for the fixes!

PR #1521, opened 2024-10-24
Add mx.array.__format__

### LINKED ISSUE #1520: [FEATURE] Adding (string) __format__ to values extracted from array
**Describe the feature**
Be able to use f-strings formatting on elements of arrays.

**To Reproduce**
```python
import mlx.core as mx
a = mx.arange(1)
print(f"{a[0]:.3f}")
# TypeError: unsupported format string passed to array.__format__
b = mx.random.normal()
print(f"{b:.3f}")
# TypeError: unsupported format string passed to array.__format__
```

**Expected behavior**
Displaying the float to the nth digit like `print(f"{float(a[0]):.{n}f}")` would do, without converting to a float.

**Desktop:**
 - MacBook Pro M3 Max (14 cores, 96 GB)
 - MacOS 14.6.1
 - MLX 0.19.0

**Additional context**
I think it would help make the transition from NumPy or PyTorch slightly smoother.

barronalex:
Resolves #1520

### DIFF
--- a/python/src/array.cpp
+++ b/python/src/array.cpp
@@ -848,7 +848,7 @@ void init_array(nb::module_& m) {
       .def(
           "__format__",
           [](array& a, nb::object format_spec) {
-            if (a.size() > 1) {
+            if (a.ndim() == 0) {
               throw nb::type_error(
                   "unsupported format string passed to mx.array.__format__");
             }
--- a/python/src/array.cpp
+++ b/python/src/array.cpp
@@ -846,6 +846,16 @@ void init_array(nb::module_& m) {
       .def("__int__", [](array& a) { return nb::int_(to_scalar(a)); })
       .def("__float__", [](array& a) { return nb::float_(to_scalar(a)); })
       .def(
+          "__format__",
+          [](array& a, nb::object format_spec) {
+            if (a.size() > 1) {
+              throw nb::type_error(
+                  "unsupported format string passed to mx.array.__format__");
+            }
+            auto obj = to_scalar(a);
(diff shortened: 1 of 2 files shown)


python/tests/test_array.py
```diff
@@ -1875,6 +1875,17 @@ def test_to_scalar(self):
         with self.assertRaises(ValueError):
             int(a)
 
+    def test_format(self):
+        a = mx.arange(3)
+        self.assertEqual(f"{a[0]:.2f}", "0.00")
+
+        b = 0.35487
+
+        self.assertEqual(f"{b:.1f}", "0.4")
```
awni, line 1884: I don't think this is testing anything in MLX?

python/src/array.cpp
```diff
@@ -845,6 +845,16 @@ void init_array(nb::module_& m) {
           nb::rv_policy::none)
       .def("__int__", [](array& a) { return nb::int_(to_scalar(a)); })
       .def("__float__", [](array& a) { return nb::float_(to_scalar(a)); })
+      .def(
+          "__format__",
+          [](array& a, nb::object format_spec) {
+            if (a.size() > 1) {
```
awni, line 851: I notice numpy only supports scalar (`a.ndim() == 0`). Should we do the same?

awni (approved):
Thanks for the fast fix!! Left a couple minor comments. Take a look then feel free to merge when ready.

python/tests/test_array.py
barronalex: That's silly -- good catch!

PR #1523, opened 2024-10-25
Fix synchronization bug

awni:
No noticeable change (within variance) in LLM generation or transformer training on M2 Ultra.

Bench | Pre | Post 
----- | ----- |-----
Mistral 7B 4-bit | 112.94 tokens-per-sec | 113.119 tokens-per-sec |
Transformer LM Train | It/sec 5.903 | It/sec 5.908 |

### DIFF
--- a/mlx/backend/metal/device.cpp
+++ b/mlx/backend/metal/device.cpp
@@ -130,10 +130,11 @@ CommandEncoder::~CommandEncoder() {
   enc_->release();
 }
 
-void CommandEncoder::set_array(
+void CommandEncoder::set_input_array(
     const array& a,
     int idx,
     int64_t offset /* = 0 */) {
+  all_inputs_.insert(a.buffer().ptr());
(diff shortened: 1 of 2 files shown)


mlx/backend/metal/device.cpp
```diff
@@ -130,10 +130,11 @@ CommandEncoder::~CommandEncoder() {
   enc_->release();
 }
 
-void CommandEncoder::set_array(
+void CommandEncoder::set_input_array(
     const array& a,
     int idx,
     int64_t offset /* = 0 */) {
+  all_inputs_.insert(a.buffer().ptr());
```
barronalex, line 137: So now all `all_inputs_` also contains all outputs?

barronalex (approved):
Nice!

mlx/backend/metal/device.cpp
awni: Exactly

PR #1525, opened 2024-10-25
BFS graph evaluation order

awni:
This is a draft implementation of changing the graph evaluation order from DFS to BFS. So far results are promising, more benchmarks forth-coming.

There might be a faster/simpler way to do the traversal, this also needs some scrutiny.

### Benchmarks:

On M2 Ultra

Benchmark | DFS  | BFS 
----- | ---- | ----
Simple matmul + unary | 1.725 ms | 1.023 ms
Mistral 7B 4-bit, 512 tokens | 113.0  toks/sec | 120.5 toks/sec
Transformer train 153 M |   5.9 its/sec | 6.22 its/sec
MNIST | 0.608 seconds | 0.565 seconds
LeNet | 7.241 seconds | 5.212 seconds
ResNet 22 Training CIFAR | 1021.85 samples/second | 1175.79 samples/second

Memory Use

Benchmark | DFS | BFS
---- | --- | -----
Mistral 7B 4-bit, 512 tokens | 3.900 GB	| 3.900 GB
Transformer train 153 M	| 5.534 GB | 5.480 GB
ResNet 22 Training CIFAR |1.643 GB |	2.264 GB

Simple benchmark code:

```python
a = mx.random.uniform(shape=(256, 256))
b = mx.random.uniform(shape=(256, 256))

def fun():
    outs = []
    for _ in range(100):
        x = a @ b
        x = mx.abs(x)
        outs.append(x)
    return outs
```

### DIFF
--- a/mlx/transforms.cpp
+++ b/mlx/transforms.cpp
@@ -40,7 +40,7 @@ int detail::InTracing::tracing_counter{0};
 int detail::RetainGraph::tracing_counter{0};
 
 array eval_impl(std::vector<array> outputs, bool async) {
-  std::queue<array> tape;
+  std::vector<array> tape;
 
   // stream events to use for synchronization
   std::unordered_map<uint32_t, Event> events;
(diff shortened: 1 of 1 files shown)


angeloskath (approved):
This is so well done!!!

The implementation is so much simpler (and probably a lot faster) than my old priority queue implementation... it's embarrassing 😊

Not sure why it is draft still. Let's go 🚀🚀🚀

awni:
> You might want to run some lora training as well. I had a pretty chunky slowdown (~15%) when combined with checkpointing and QLoRA on FLUX.

Confirmed no regression with LoRA / QLoRA in MLX LM. Per offline discussion we'll merge this and continue investigating slowdown in QLoRA FLUX with gradient checkpointing.

PR #1535, opened 2024-10-28
Faster bits and bernoulli

awni:
- Better grid dim for much faster bits
- Use bits directly for bernoulli

On M1 Max

Benchmark | Pre (ms) | Post (ms)
--- | ---- | ----
Uniform | 21.283 | 4.320 
Bernoulli | 21.791 |  1.911

Statistical test for Bernoulli

P | Pre `abs(p-phat)` | Post `abs(p-phat)
--- | ----- | ----
0.010| 0.0000270 | 0.0000191
0.500| 0.0000430 | 0.0000903
0.990| 0.0000566 | 0.0000239

Code for benchmarks + test:

```python
shape = (32_000_000,)

timeit(lambda : mx.random.uniform(shape=shape))

timeit(lambda : mx.random.bernoulli(0.1, shape))

n = 10_000_000

for p in [0.01, 0.5, 0.99]:
    b = mx.random.bernoulli(p, shape=(n,))
    phat = b.sum() / n
    diff = abs(p - phat)
    print(f"P={p:.3f}, diff = {diff:.7f}")
```

### DIFF
--- a/mlx/backend/metal/kernels/random.metal
+++ b/mlx/backend/metal/kernels/random.metal
@@ -34,8 +34,8 @@ rbits threefry2x32_hash(const thread uint2& key, uint2 count) {
 [[kernel]] void rbitsc(
     device const uint32_t* keys,
     device char* out,
-    device const bool& odd,
-    device const uint& bytes_per_key,
+    constant const bool& odd,
+    constant const uint& bytes_per_key,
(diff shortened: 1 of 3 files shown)


thegodone:
Yes I see that bernoulli is bad top on torch but only on cpu, I made a issue last week.

angeloskath (approved):

PR #1537, opened 2024-10-28
Skip using Residency sets in VMs

carlocab:
## Proposed changes

Attempting to use residency sets in a VM throws[^1]

    libc++abi: terminating due to uncaught exception of type std::runtime_error: [metal::Device] Unable to construct residency set.

Not quite sure if this is the best fix, but it does make the error go
away.

Note that it was previously possible to run simple programs that used
mlx in a VM prior to 0eb56d5be02a826b13519a1418e345150f5e9526. See
related discussion at Homebrew/homebrew-core#195627.

[^1]: https://github.com/Homebrew/homebrew-core/actions/runs/11525831492/job/32105148462#step:3:56

## Checklist

I didn't run `pre-commit`, but it did run `clang-format` on the single file I changed. It reordered some existing includes, which didn't seem correct to me so I dropped that change. I can commit it again if desired.

### DIFF
--- a/mlx/backend/metal/resident.cpp
+++ b/mlx/backend/metal/resident.cpp
@@ -1,5 +1,8 @@
 // Copyright © 2024 Apple Inc.
 
+#include <sys/sysctl.h>
+#include <cstddef>
+
 #include "mlx/backend/metal/resident.h"
 #include "mlx/backend/metal/metal_impl.h"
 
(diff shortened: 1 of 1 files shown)


awni:
Thanks for sending a fix for this. I will do a quick investigation as prior to now we did not have to special case metal features for inside/outside VM.

mlx/backend/metal/resident.cpp
```diff
@@ -8,8 +11,20 @@ namespace mlx::core::metal {
...
 #define supported __builtin_available(macOS 15, iOS 18, *)
 
+// Trying to create a residency set in a VM leads to errors.
+static bool in_vm() {
+  static int in_vm = -1;
+  if (in_vm != -1)
+    return in_vm;
+
+  std::size_t len = sizeof(in_vm);
+  sysctlbyname("kern.hv_vmm_present", &in_vm, &len, NULL, 0);
```
awni, line 21: Small nit: if you put this in a lambda then you don't have to do the check for `-1`. Something like:

```suggestion
  auto check_vm() {
    int inv_vm;
    std::size_t len = sizeof(in_vm);
    sysctlbyname("kern.hv_vmm_present", &in_vm, &len, NULL, 0);
    return in_vm
  };
  static int in_vm = check_vm();
  return in_vm;
```

carlocab: Thanks for the suggestion! I write basically no C++, so using a lambda didn't occur to me. Let me test this out locally and update this PR.

carlocab:
Thanks, @awni. Updated this slightly, but it seems to work fine.

awni:
Per offline discussion, checking for Metal3 support is sufficient for this to run in a VM. I simplified the checking code a bit based on that.

awni (approved):
Thanks for flagging and fixing!

PR #1539, opened 2024-10-28
Fix format for array without spec

### LINKED ISSUE #1538: [BUG] formatted printing of mlx arrays broken
**Describe the bug**
Printing of arrays with format strings is broken, output `TypeError: unsupported format string passed to mx.array.__format__
`. 

**To Reproduce**

The following code worked in 0.19, broken in 0.19.1, presumably due to #1521

```python
import mlx.core as mx

a = mx.arange(1)
print(a) #works
print(f"{a}") #broken
```

**Expected behavior**
Expected output: 
```
array([0], dtype=int32)
array([0], dtype=int32)
```

awni:
Close #1538

### DIFF
--- a/python/src/array.cpp
+++ b/python/src/array.cpp
@@ -848,12 +848,19 @@ void init_array(nb::module_& m) {
       .def(
           "__format__",
           [](array& a, nb::object format_spec) {
-            if (a.ndim() > 0) {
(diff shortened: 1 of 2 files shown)


barronalex (approved):
Thanks for the quick fix!

PR #1543, opened 2024-10-30
Fix thread group for large arrays

awni:
FYI this is an important cliff to be aware of. The following example grid/group dimension is a performance bug:

```python
grid=(10_000_000, 1, 1),
threadgroup=(1, 1024, 1),
```

Bench | Pre | Post
---- | ---- | ----
Unary | 1978.538 (ms) | 43.691 (ms)
Binary | 1979.676 (ms) | 50.079 (ms)
Copy | 1923.015 (ms) | 44.385 (ms)
Compiled |1977.468 (ms) | 43.802 (ms)

Benchmark code:

```python
def big_unary():
    x = mx.zeros((3, 2**31 - 1), mx.int8)
    timeit(mx.abs, x, num_iter=5)

def big_binary():
    x = mx.zeros((3, 2**31 - 1), mx.int8)
    timeit(mx.add, x, x, num_iter=5)

def big_copy():
    x = mx.zeros((3, 2**31 - 1), mx.int8)
    timeit(lambda x: x.astype(mx.uint8), x, num_iter=5)

def big_compiled():
    x = mx.zeros((3, 2**31 - 1), mx.int8)

    @mx.compile
    def fun(x):
        return mx.abs(-x)

    timeit(fun, x, num_iter=5)
```

### DIFF
--- a/mlx/backend/metal/binary.cpp
+++ b/mlx/backend/metal/binary.cpp
@@ -110,6 +110,7 @@ void binary_op_gpu_inplace(
     compute_encoder.set_output_array(outputs[1], arg_idx++);
   }
 
+  auto thread_group_size = kernel->maxTotalThreadsPerThreadgroup();
   if (bopt == BinaryOpType::General) {
     // Launch up to 3D grid of threads
     size_t dim0 = ndim > 0 ? shape[ndim - 1] : 1;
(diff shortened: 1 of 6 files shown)


angeloskath (approved):
Out of curiosity why not change `get_2d_grid_dims` to always return the larger number first?

PR #1551, opened 2024-11-01
Add dispatchThreads to custom kernel doc

### LINKED ISSUE #1547: [DOCS] Incomplete mlx.core.fast.metal_kernel
**Describe the bug**

While writing a custom kernel using mlx.core.fast's metal_kernel() function, I found myself looking through mlx's source code multiple times. 
It wasn't clear that the grid argument to provide the created callable is for the `dispatchThreads` and not `dispatchThreadgroups` underlying call.

**Expected behavior**

The mlx.core.fast.metal_kernel's docs page should include info like the used attribute detection & other which are mentionned in the custom_metal_kernels guide OR a link to the custom_metal_kernels guide.
It should also be mentionned that the grid attribute is the grid size in Threads and not Threadgroups.

barronalex:
Resolves #1547 

Ideally I'd like to add `metal_kernel.__call__` to the autosummary generated by Sphinx, but when I do that I run into errors when it tries to summarize the `nb::cpp_function`.

This PR adds some more info directly to the doc as a temporary solution.

### DIFF
--- a/docs/src/dev/custom_metal_kernels.rst
+++ b/docs/src/dev/custom_metal_kernels.rst
@@ -76,7 +76,7 @@ Putting this all together, the generated function signature for ``myexp`` is as 
 
   template [[host_name("custom_kernel_myexp_float")]] [[kernel]] decltype(custom_kernel_myexp_float<float>) custom_kernel_myexp_float<float>;
 
-Note: ``grid`` and ``threadgroup`` are parameters to the Metal `dispatchThreads function <https://developer.apple.com/documentation/metal/mtlcomputecommandencoder/2866532-dispatchthreads>`_.
+Note: ``grid`` and ``threadgroup`` are parameters to the Metal `dispatchThreads <https://developer.apple.com/documentation/metal/mtlcomputecommandencoder/2866532-dispatchthreads>`_ function.
 This means we will launch ``mx.prod(grid)`` threads, subdivided into ``threadgroup`` size threadgroups.
 
 Passing ``verbose=True`` to ``mx.fast.metal_kernel.__call__`` will print the generated code for debugging purposes.
--- a/docs/src/dev/custom_metal_kernels.rst
+++ b/docs/src/dev/custom_metal_kernels.rst
@@ -1,3 +1,5 @@
+.. _custom_metal_kernels:
+
 Custom Metal Kernels
 ====================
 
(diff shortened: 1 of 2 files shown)


angeloskath (approved):
This looks great. I would add a link to this page in the metal kernel docstring. It is not the most generic thing but I think it would be a great improvement that doesn't have to deal with the `nb:cpp_function` shenanigans.

docs/src/dev/custom_metal_kernels.rst
```diff
@@ -76,6 +76,9 @@ Putting this all together, the generated function signature for ``myexp`` is as
 
   template [[host_name("custom_kernel_myexp_float")]] [[kernel]] decltype(custom_kernel_myexp_float<float>) custom_kernel_myexp_float<float>;
 
+Note: ``grid`` and ``threadgroup`` are parameters to the Metal `dispatchThreads <https://developer.apple.com/documentation/metal/mtlcomputecommandencoder/2866532-dispatchthreads>`_ function.
+This means we will launch ``mx.prod(grid)`` threads, subdivided into ``threadgroup`` size threadgroups.
```
awni, line 80: This could be a good spot to put a little warning here to make sure the thread group dimension is always less than or equal to the corresponding grid dimension.

Something like: 

> For optimal performance, each thread group dimension should be less than or equal to the corresponding grid dimension.

python/src/fast.cpp
```diff
@@ -278,7 +278,9 @@ void init_fast(nb::module_& parent_module) {
               output_shapes (List[Sequence[int]]): The list of shapes for each output in ``output_names``.
               output_dtypes (List[Dtype]): The list of data types for each output in ``output_names``.
               grid (tuple[int, int, int]): 3-tuple specifying the grid to launch the kernel with.
+                This will be passed to MTLComputeCommandEncoder::dispatch_threads.
               threadgroup (tuple[int, int, int]): 3-tuple specifying the threadgroup size to use.
+                This will be passed to MTLComputeCommandEncoder::dispatch_threads.
```
awni, line 283: Nit:

```suggestion
                This will be passed to ``MTLComputeCommandEncoder::dispatch_threads``.
              threadgroup (tuple[int, int, int]): 3-tuple specifying the threadgroup size to use.
                This will be passed to ``MTLComputeCommandEncoder::dispatch_threads``.
```

awni (approved):
Thanks!

Leikoe:
Exactly what I needed, thank you 😄

PR #1553, opened 2024-11-02
Add python 3.13 to circle

awni:
I ran the tests locally in a 3.13 env and they all passed so I think it's ok to do a 3.13 distribution.

Closes https://github.com/ml-explore/mlx-examples/issues/1083


angeloskath (approved):
🙏

PR #1558, opened 2024-11-03
Sdpa fix

angeloskath:
Fixes the incorrect assumption that `k` and `v` will have the same underlying layout.

Thanks to @tlikhomanenko for uncovering the bug!


tlikhomanenko:
Thanks for the quick fix!

awni (approved):
Thanks!

PR #1561, opened 2024-11-04
Use fewer barriers

awni:
Everything after a barrier waits on everything before so we don't need to synchronize on anything that happened before a barrier. This change implements that.

Some benchmarks on M2 Ultra:

Bench | Pre | Post
----- | ---- | ----
4-bit Mistral 7B generating 512 toks | 122.0 | 123.8
4-bit Llama 1B generating 512 toks | 420.2  | 431.3
Transformer training | 6.237 | 6.265 |

On LeNet and MNIST no change observed.

With this change we can hit > 130 toks/sec for 4-bit Mistral 7B by increasing ops per buffer:

```
MLX_MAX_OPS_PER_BUFFER=80 mlx_lm.generate --model mlx-community/Mistral-7B-Instruct-v0.3-4bit --prompt "Write a story about Einstein"  --temp 0.0 --max-tokens 512
```

Pre: `125.939 tokens-per-sec`
Post: `130.900 tokens-per-sec`

### DIFF
--- a/mlx/backend/metal/device.cpp
+++ b/mlx/backend/metal/device.cpp
@@ -164,7 +164,7 @@ void CommandEncoder::maybeInsertBarrier() {
   if (needs_barrier_) {
     enc_->memoryBarrier(MTL::BarrierScopeBuffers);
     needs_barrier_ = false;
-    prev_outputs_ = next_outputs_;
+    prev_outputs_ = std::move(next_outputs_);
   } else {
     prev_outputs_.insert(next_outputs_.begin(), next_outputs_.end());
   }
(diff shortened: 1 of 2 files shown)


mlx/backend/metal/device.cpp
```diff
@@ -161,19 +156,32 @@ void CommandEncoder::set_output_array(
...
-    outputs_.insert(buf);
+    next_outputs_.insert(buf);
+  }
+}
+
+void CommandEncoder::maybeInsertBarrier() {
+  if (needs_barrier_) {
+    enc_->memoryBarrier(MTL::BarrierScopeBuffers);
+    needs_barrier_ = false;
+    prev_outputs_ = next_outputs_;
```
angeloskath, line 167: Maybe `std::move`? You could do an `extract()` and move insert loop below but that may be unnecessarily complicated given that we are copying around some integers.

angeloskath (approved):
Looks great! All that is necessary and absolutely nothing more 😊

PR #1563, opened 2024-11-05
Fix gather vmap

### LINKED ISSUE #1517: [BUG] Indexing inside double vmap raises 'ValueError: [transpose] Recived 3 axes for array with 4 dimensions.'
**Describe the bug**
MLX raises a cryptic error (`ValueError: [transpose] Recived 3 axes for array with 4 dimensions.`) when a function is vmapped multiple times in a specific way and we index one argument by the other.

See example below as the steps to reproduce are quite specific.

**To Reproduce**
```python
import mlx.core as mx

def f(args):
    (x, y) = args
    return x[y]

f = mx.vmap(f, [0, None])
f = mx.vmap(f, [0, 0])
f([mx.ones((2, 3, 4)), mx.zeros(2, dtype=mx.int32)])
# ValueError: [transpose] Recived 3 axes for array with 4 dimensions.
```
3 is the expected number of dimensions returned from f while 4 is the original number of dimensions. Perhaps this gives a clue for where the error originates from? By increasing the number of dimensions in the arrays we get different numbers in the error message:
```python
import mlx.core as mx

def f2(args):
    (x, y) = args
    return x[y]

f2 = mx.vmap(f2, [0, None])
f2 = mx.vmap(f2, [0, 0])
f2([mx.ones((2, 3, 4, 5, 6)), mx.zeros((2, 3, 4), dtype=mx.int32)])
# ValueError: [transpose] Recived 7 axes for array with 8 dimensions.
```

**Expected behavior**
Expected not an error and expected `f([mx.ones((2, 3, 4)), mx.zeros(2, dtype=mx.int32)]).shape` to output `(2, 3)`.

 - OS Version: [e.g. MacOS 15.0.1]
 - Version [e.g. 0.19.0]

**Additional context**
Discovered while trying to implement a custom interpolation method.

awni:
Our gather vmap was a bit buggy. This fixes it and adds some more tests. Closes #1517 1517

### DIFF
--- a/mlx/primitives.cpp
+++ b/mlx/primitives.cpp
@@ -1,5 +1,4 @@
 // Copyright © 2023-2024 Apple Inc.
-
 #include <algorithm>
 #include <cassert>
 #include <cmath>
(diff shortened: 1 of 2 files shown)


angeloskath (approved):
Looks great. There are no tests for `!src_vmapped` but I tested it a bit and everything seems fine.

The logic for double checking is
- Move all vmapped index axes to the first index or the src
- Expand if an index is not vmapped
- Add an extra index dimension or for efficiency update the slices if we didn't have vmapped indices

The one confusing part was the conditional squeeze in the end but I think if src_vmapped and !indices_vmapped there wouldn't be an extra dimension since we updated the slice size instead of adding a dimension. If !src_vmapped and indices_vmapped then we do want the extra dim. So all is great!

awni:
> There are no tests for !src_vmapped but I tested it a bit and everything seems fine.

There are actually [some in C++](https://github.com/ml-explore/mlx/blob/main/tests/vmap_tests.cpp#L365-L396).

> The logic for double checking is

Exactly right.

> The one confusing part was the conditional squeeze in the end but I think if src_vmapped and !indices_vmapped there wouldn't be an extra dimension since we updated the slice size instead of adding a dimension. If !src_vmapped and indices_vmapped then we do want the extra dim.

Exactly right 😄

PR #1568, opened 2024-11-07
Bfs width limit

awni:
Add a width limit to the BFS. Made it configurable with an env var `MLX_BFS_MAX_WIDTH` to facilitate benchmarking and keep it flexible.

The default is temporarily 10 but that needs to be validated on benchmarks:

A simple benchmark like the following which is worst case for BFS will use way less RAM:

```python
arrs = [mx.zeros((4096, 128), mx.int64) for _ in range(1000)]
arrs = [x.astype(mx.int8) for x in arrs]
mx.eval(arrs)
print(mx.metal.get_peak_memory() / 1e6)
```

Before the width limit:  4,251 MB
width 1: 689 MB
width 10: 729 MB
width 50: 792 MB

### Benchmarks M3 Max

LLM inference Llama 3.2 1B 4-bit, 512 tokens

Width | toks/sec | RAM
--- | --- | ---
1 | 357.2 | 0.724 GB
5 | 373.2 | 0.723 GB
10 | 374.0 | 0.724 GB
inf | 375.0 | 0.724 GB

Transformer Training

Width | it/sec | RAM
--- | --- | ---
1 | 3.32 | 5.204
10 | 3.47 | 5.222
50 | 3.48 | 5.48
inf | 3.48 | 5.48

CIFAR ResNet

Width | im/sec | RAM
--- | --- | ---
1 | 3152 | 1.468 GB
10 | 3087 | 1.515 GB
20 | 3168 | 1.671 GB
50 | 3887 | 1.895 GB 
inf | 4023 | 2.264  GB

Quantizing Llama 8B to 4-bit
Width | Time (s) | RAM (GB)
--- | --- | ---
1 | 2.92 | 9.37 
10 | 2.83 | 9.59
20 | 2.89 | 10.20
50 |  2.90 | 10.22
inf | 3.14 | 18.17

### DIFF
--- a/mlx/backend/metal/metal.cpp
+++ b/mlx/backend/metal/metal.cpp
@@ -1,31 +1,17 @@
 // Copyright © 2023-2024 Apple Inc.
-#include <cstdlib>
 #include <memory>
 
 #include "mlx/backend/metal/device.h"
 #include "mlx/backend/metal/utils.h"
 #include "mlx/primitives.h"
 #include "mlx/scheduler.h"
+#include "mlx/utils.h"
 
 namespace mlx::core::metal {
 
 bool is_available() {
   return true;
 }
-
-int max_ops_per_buffer() {
-  auto get_val = []() {
-    if (const char* buff_str = std::getenv("MLX_MAX_OPS_PER_BUFFER")) {
-      return atoi(buff_str);
-    } else {
-      return 10;
-    }
-  };
-  static int max_ops_per_buffer_ = get_val();
-  return max_ops_per_buffer_;
-}
-
-#define MAX_OPS_PER_BUFFER max_ops_per_buffer()
 
 inline void check_error(MTL::CommandBuffer* cbuf) {
   if (cbuf->status() == MTL::CommandBufferStatusError) {
@@ -77,7 +63,8 @@ std::function<void()> make_task(array arr, bool signal) {
       out.set_status(array::Status::evaluated);
     }
 
-    if (signal || d.get_command_buffer_ops(s.index) >= MAX_OPS_PER_BUFFER) {
+    if (signal ||
+        d.get_command_buffer_ops(s.index) >= env::max_ops_per_buffer()) {
       d.end_encoding(s.index);
       if (signal) {
         command_buffer->encodeSignalEvent(
--- a/mlx/transforms.cpp
+++ b/mlx/transforms.cpp
@@ -26,7 +26,7 @@ int bfs_max_width() {
       return 10;
     }
   };
-  static int bfs_max_width_ get_val();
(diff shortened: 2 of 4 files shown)


awni:
I think we should land this. It basically gets 99% of the advantage of BFS but protects from the extreme edge cases. Based on experimentation, I think the default should be somewhere between 10-50. There will always be a trade-off between memory use and parallelism.. but that range seems pretty stable / gets most of the best of both DFS / BFS.

angeloskath (approved):
I agree let's merge it.

The BFS/DFS hybrid is very smart code that takes 3-4 reads to build some confidence in one's understanding. I mean it both as a compliment and criticism :-) .

I am impressed but I know I will have to re-read it carefully if I ever come back to this code. I do not have an improvement to propose and it is simple enough once one tracks all indices in their head carefully so we should merge it.

Again very nice code though. Using the ordered STL containers was slowing things down (the many months back that I touched this part of the code) while careful looping seems to have no negative effect in performance.

It might be nice to re-implement this in our `graph_utils` to retain the property of printing/exporing a graph in the order of evaluation.

awni:
> It might be nice to re-implement this in our graph_utils to retain the property of printing/exporing a graph in the order of evaluation.

It's a good idea. I want to think on it a bit though as naively it's a lot of code duplication (which is ok when it's simple but I'd prefer to avoid pasting this complex bit of code in other places and the `graph_utils` is so nice and simple right now..). For exporting to dot it won't matter much since dot will determine the order in the image. But for printing it would be nice to have the same order.

angeloskath:
Yeah I agree 100%. We shouldn't copy/paste. It is a good opportunity to benchmark and explore rewriting this code in a location where performance is much less critical. As far as dot goes, I kinda like that one can trace the evaluation order based on the names of the nodes which is true in dot images as well. I remember I used it a lot when reasoning about checkpointing.

awni:
I'm going with 20 as the default.. it seems to be a sweet spot for most workloads. The ResNet is a bit of an outlier because of the suboptimal gradient computation for the weights. I think fixing that with a single Conv primitive will make a smaller BFS width limit more favorable for it as well.

awni:
> As far as dot goes, I kinda like that one can trace the evaluation order based on the names of the nodes which is true in dot images as well. I remember I used it a lot when reasoning about checkpointing.

Very good point. Will aim to fix that in a follow up.

angeloskath:
> Will aim to fix that in a follow up.

Yeah no worries, I didn't mean to dump it on you :-) Just thinking out-loud.

> I'm going with 20 as the default.

Exactly what you wrote. Given also the 10 ops per buffer it is unlikely that we 'll get significant parallelization speedups with more than 20 primitives anyway.

PR #1569, opened 2024-11-07
Add groups in nn.Conv2d

awni:
- Add groups for `nn.Conv2d`
- Some nits to isclose to reduce graph size which is a little out of hand..


angeloskath (approved):
Nice! Manual run of `simplify` :-P

PR #1572, opened 2024-11-07
Fully wrap the command encoder

awni:
- In order to protect from direct access to the command encoder, this fully wraps it inside the `metal::CommandEncoder`.
- I used snake case for method names for consistency with the rest of the code (and updated other methods to be snake case..like `dispatch_threads`) but if you prefer we keep it camelCase lmk.


angeloskath (approved):
This is the lord's work 😊

Thanks!

angeloskath:
I actually like the snake case. It also makes it clear that this is wrapped and not the original object.

PR #1574, opened 2024-11-08
Fix per-example mask + docs in sdpa

awni:
- This worked in the regular case but was broken in the gqa case since we reshape to 5d.
- FIx + test
- More in the docs on this.


angeloskath (approved):
🙏

PR #1575, opened 2024-11-08
Throw for invalid *fft{2,n} cases and remove test

awni:
Numpy 2.0 deprecated passing a shape with axes unspecified to fft2, fftn2, etc. Just following the behavior there and it gets rid of warnings in the tests.

It also makes the C++ and Python APIs align more closely which is nice.


barronalex (approved):
Nice -- thanks!

PR #1577, opened 2024-11-08
Fix OOB access in qmv

barronalex:
This was breaking quantized generation with `SmolLM2-135M-Instruct` specifically with group size 32.

No clear difference in performance before and after (tested with group size 64 which still worked on `main`).

Before:
```
Prompt: 38 tokens, 629.657 tokens-per-sec
Generation: 574 tokens, 261.939 tokens-per-sec
Peak memory: 0.164 GB
```
After:
```
Prompt: 38 tokens, 630.536 tokens-per-sec
Generation: 574 tokens, 262.050 tokens-per-sec
Peak memory: 0.164 GB
```

### DIFF
--- a/mlx/backend/metal/kernels/quantized.h
+++ b/mlx/backend/metal/kernels/quantized.h
@@ -628,10 +628,12 @@ METAL_FUNC void qmv_impl(
       const device T* sl = scales + row * in_vec_size_g;
       const device T* bl = biases + row * in_vec_size_g;
 
-      U s = sl[0];
-      U b = bl[0];
-      result[row] += qdot_safe<U, values_per_thread, bits>(
-          wl, x_thread, s, b, sum, remaining);
+      if (remaining > 0) {
+        U s = sl[0];
+        U b = bl[0];
+        result[row] += qdot_safe<U, values_per_thread, bits>(
+            wl, x_thread, s, b, sum, remaining);
+      }
     }
 
     for (int row = 0; row < results_per_simdgroup; row++) {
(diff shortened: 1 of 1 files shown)


mlx/backend/metal/kernels/quantized.h
```diff
@@ -628,10 +628,12 @@ METAL_FUNC void qmv_impl(
...
-      U s = sl[0];
-      U b = bl[0];
-      result[row] += qdot_safe<U, values_per_thread, bits>(
-          wl, x_thread, s, b, sum, remaining);
+      if (remaining > 0) {
+        U s = sl[0];
+        U b = bl[0];
+        result[row] += qdot_safe<U, values_per_thread, bits>(
+            wl, x_thread, s, b, sum, remaining);
+      }
```
barronalex, line 636: Even though `sum` and `accum` are zero when `remaining=0`, `scale * accum + sum * bias can still produce bad outputs if the uninitialized memory at `s` or `b` is nan or inf

angeloskath: That's awesome! Thanks for finding and fixing this! I can't be sure without measuring but a possibly better fix is to skip everything if remaining is 0. And it would also be needed at lines 563.

angeloskath (approved):
Thanks for the fix! Tracking these down is really hard!

I left a comment I think it would make it a bit clearer and probably result in the same exact perf as your fix.

Also not certain looking at the benchmark but that is with 32 group size on SmolLM? Cause other sizes may not route to this kernel and could route to the `qmv_fast_impl`...

barronalex:
No worries! Totally agree with the comment -- I updated it.

The benchmark is with group size 64 since that worked with `main`. The qmv size is (192, 576) so it should never go to `qmv_fast` (also I don't think group size affects the routing at the moment).

PR #1578, opened 2024-11-09
Cpu fast quantize

awni:
Maybe worth merging this. We have the fast op already so it's a small addition. And the benchmarks look pretty decent.

Quantizing Llama 3.1 8B on CPU on M3 Max:

Pre: 33.34 seconds
Post: 16.542 seconds


angeloskath (approved):
👍🚀

PR #1584, opened 2024-11-12
Dispatch bf16 at run time when using the JIT

awni:
This dispatches based on metal version at runtime when using the JIT.

The idea is to simplify distributing a single binary while supporting older OS yet getting newer features when available.

In general I think these are good guidelines to follow that give the most flexibility in terms of building applications with MLX:

1. As much as possible minimize where we conditionally compile based on the metal version 
2. When using the JIT we can easily choose the right branch at run-time with negligible cost. As much as possible we should put things which dispatch in Metal in the JIT (when `MLX_METAL_JIT=True`) so that we at least have this option.
3. When not using the JIT you have two options: either deploy a single binary for the minimal target without using newer features OR ship multiple binaries (like we do in PyPi).

CC @davidkoski

### DIFF
--- a/mlx/backend/metal/kernels/utils.h
+++ b/mlx/backend/metal/kernels/utils.h
@@ -4,7 +4,9 @@
 
 #include <metal_math>
 
-// bf16.h is conditionally included based on the metal version
+// The correct bf16.h is included based on the metal version
+// by giving the correct path to -I during compilation
+// e.g. mlx/backend/metal/kernels/metal_3_0/ for Metal 3.0
 #include "bf16.h"
 
 #include "mlx/backend/metal/kernels/bf16_math.h"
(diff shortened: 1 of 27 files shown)


mlx/backend/metal/kernels/utils.h
```diff
@@ -3,7 +3,11 @@
 #pragma once
 
 #include <metal_math>
-#include "mlx/backend/metal/kernels/bf16.h"
+
+// bf16.h is conditionally included based on the metal version
+#include "bf16.h"
```
awni, line 8: Instead of using the preprocessor include the correct bf16 file in CMake here.

angeloskath: I know what you mean but only because I read the PR. Perhaps "The appropriate bf16.h will be included based on the metal version" or "A different bf16.h will be included based on the metal version"?

angeloskath (approved):
This is great! I left one tiny comment, I am not even sure it is an improvement. It's just that it is the only place that we do not have absolute includes and it can be confusing.

PR #1587, opened 2024-11-14
Fix donation in sdpa

awni:
Previously the query would never be donated in the non-copy case because we incremented the ref count before the donation check.


barronalex (approved):
Nice find!!

PR #1589, opened 2024-11-16
Faster indexing math in a few kernels

awni:
- Use `uint` instead of `size_t` in unary, binary, ternary, copy, and jit compile kernels when possible.

- Generally changing code that I touch from using `ostringstream` to `string` concatenation / formatting which prior benchmarking shows can be a lot faster. I also think it can be more readable.

### Microbenchmarks on M1 Max:

Bench | Pre (ms) | Post (ms)
----- | ---- | ----
Unary general 5D | 4.721 | 2.950 
Binary general 3D | 6.164 |  4.761
Binary general 5D | 6.147 |  5.947
Ternary general 3D | 6.209 |  5.236 
Compiled batch norm | 4.784 | 3.171 
Gather 3D |3.913 | 3.105
Scatter 5D | 43.536 | 10.827

### Microbenchmarks on M3 Max:

Bench | Pre (ms) | Post (ms)
----- | ---- | ----
Unary general 5D | 3.412 | 3.093
Binary general 3D | 3.063 | 3.032
Binary general 5D | 5.989 |  5.828
Ternary general 3D | 3.351 |  3.339
Compiled batch norm | 3.66 | 2.62

### CIFAR ResNet Training benchmark on M3 Max

Pre: 28.2k ims/sec
Post 34.2k ims/sec

### No JIT binary sizes:

Pre: uncompressed 69M, compressed 19M
Post: uncompressed 77M, compressed 19M

### DIFF
--- a/mlx/backend/metal/utils.h
+++ b/mlx/backend/metal/utils.h
@@ -61,4 +61,15 @@ inline void debug_set_primitive_buffer_label(
 
 std::string get_primitive_string(Primitive* primitive);
 
+template <typename T>
+void concatenate(std::string& acc, T first) {
+  acc += first;
+}
+
+template <typename T, typename... Args>
+void concatenate(std::string& acc, T first, Args... args) {
+  acc += first;
+  concatenate(acc, args...);
+}
+
 } // namespace mlx::core
--- a/docs/src/usage/function_transforms.rst
+++ b/docs/src/usage/function_transforms.rst
@@ -184,8 +184,8 @@ Let's time these two different versions:
   print(timeit.timeit(lambda: mx.eval(naive_add(xs, ys)), number=100))
   print(timeit.timeit(lambda: mx.eval(vmap_add(xs, ys)), number=100))
 
-On an M1 Max the naive version takes in total ``0.390`` seconds whereas the
-vectorized version takes only ``0.025`` seconds, more than ten times faster.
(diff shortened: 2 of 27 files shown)


mlx/backend/metal/utils.h
```diff
@@ -61,4 +61,15 @@ inline void debug_set_primitive_buffer_label(
...
 
+template <typename T>
+void concatenate(std::string& acc, T first) {
+  acc += first;
+}
+
+template <typename T, typename... Args>
+void concatenate(std::string& acc, T first, Args... args) {
+  acc += first;
+  concatenate(acc, args...);
```
angeloskath, line 72: ❤️

angeloskath (approved):
This looks really great!

It is interesting that work per thread of 2 is faster than 1 but presumably slower than 4 for uint indexing.

awni:
> It is interesting that work per thread of 2 is faster than 1 but presumably slower than 4 for uint indexing.

It is interesting indeed. I tested all the ones I changed for at least one case mostly on an M1 Max.. but probably there is some variability based on the actual op and machine. 

In general more threads seems to be better unless we hit against indexing being expensive then we have to amortize the cost.

mlx/backend/metal/utils.h
awni: Thanks for that code. Sorry I didn't attribute you ;)

PR #1591, opened 2024-11-16
More buffer donation with no-ops

awni:
No-ops could previously break buffer donation chains because we would hold extra references to their underlying data buffers. This fixes that.

On small Transformer training we get a small decrease in memory use

Pre: It/sec 2.754, Peak memory 5.386 (GB)
Post: It/sec 2.763, Peak memory 5.347 (GB)

I went through a few benchmarks and didn't notice much difference in speed.


angeloskath (approved):
Nice!

PR #1595, opened 2024-11-17
Use osx deployment target to pick Metal version

### LINKED ISSUE #1170: [Enhancement] be able to override MLX_METAL_VERSION when running cmake
**Describe the bug**

`mlx-swift` needs to have the various preambles (e.g. jit directory with metal shaders) built and checked in to the repository.  Currently cmake picks up the SDK that is present on the machine doing this work but the actual jit happens using the runtime OS (or perhaps the minimum OS target).  Anyway, I need to override MLX_METAL_VERSION to METAL_3_0 for generating these files (longer discussion offline about what that means and where we might go with it).

Ideally I would be able to override the MLX_METAL_VERSION e.g. with `cmake .. -DMLX_METAL_VERSION=METAL_3_0`

**To Reproduce**

N/A

**Expected behavior**
Ideally I would be able to override the MLX_METAL_VERSION e.g. with `cmake .. -DMLX_METAL_VERSION=METAL_3_0`

N/A

awni:
Should close #1170. You can build with:

```
cmake .. -DCMAKE_OSX_DEPLOYMENT_TARGET=13.5
```

and it will use the correct version of Metal for that OS.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -89,14 +89,14 @@ elseif(MLX_BUILD_METAL)
   # Throw an error if xcrun not found
   execute_process(
     COMMAND zsh "-c" "/usr/bin/xcrun -sdk macosx --show-sdk-version"
-    OUTPUT_VARIABLE MACOS_VERSION COMMAND_ERROR_IS_FATAL ANY)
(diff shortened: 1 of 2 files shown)


angeloskath (approved):
👍

PR #1596, opened 2024-11-18
Allocate raw JSON metadata buffer on the heap, and limit its size

xnorai:
## Proposed changes

We've run into an issue where on iOS, where the stack size is smaller, stack allocation of a buffer for a large Safetensors header blows up the stack. Also, there seems to be no upper limit for how large this buffer can be. This PR fixes both of the issues by:

- Allocating the buffer on the heap
- Imposing the upper bound of 1GiB on its size.

## Checklist

### DIFF
--- a/mlx/io/safetensors.cpp
+++ b/mlx/io/safetensors.cpp
@@ -110,7 +110,8 @@ SafetensorsLoad load_safetensors(
   }
 
   uint64_t jsonHeaderLength = 0;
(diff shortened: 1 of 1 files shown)


thegodone:
should it be automatically dynamic to a given device memory resource instead of 1 Gb ?

awni (approved):
LGTM! Thanks for the fix. Will merge when tests clear.

PR #1597, opened 2024-11-18
2-Pass Sdpa Inference Kernel

angeloskath:
This PR aims to improve long context generation performance by increasing parallelization for large numbers of keys/values. There are mild benefits for smaller machines and very significant benefits for Ultra machines.

The main benefit for small machines stems from accessing the keys and values in a more cache friendly way when there is GQA and for the Ultra machines it stems from launching more thread groups which allows using more of the chip.

#### Speedup for M2 Max

The following speedup is in total tokens per second and not attention speedup. Note the phi model which does not improve does not have GQA. The 1 pass SDPA on the M2 Max achieves ~350 to 380 GB/s read for sequence length ~2048 so there isn't really much room for speedup.

m2-max-speedup

#### Speedup for M2 Ultra

Again the speedup is in total tokens per second and not attention specific. The M2 Ultra is sped up for all cases, no GQA required. The 2048 sequence length without GQA peaks at >800GB/s which also means there is probably little room for improvement (there could be for longer sequences).

m2-ultra-speedup

### DIFF
--- a/mlx/backend/metal/kernels/sdpa_vector.h
+++ b/mlx/backend/metal/kernels/sdpa_vector.h
@@ -130,7 +130,7 @@ template <typename T, int D>
     uint3 tid [[threadgroup_position_in_grid]],
     uint simd_gid [[simdgroup_index_in_threadgroup]],
     uint simd_lid [[thread_index_in_simdgroup]]) {
-  constexpr int BN = 32;
+  constexpr int BN = 8;
   constexpr int BD = 32;
   constexpr int elem_per_thread = D / BD;
   constexpr int stride = BN * D;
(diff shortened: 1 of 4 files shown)


mlx/backend/metal/kernels/sdpa_vector.h
```diff
@@ -114,3 +113,182 @@ template <typename T, int D>
...
+    // Update the output accumulator
+    for (int i = 0; i < elem_per_thread; i++) {
+      o[i] = o[i] * factor + exp_score * values[i];
+    }
+
+    // Move the pointers to the next kv
+    keys += blocks * stride;
+    values += blocks * stride;
+  }
+  threadgroup_barrier(mem_flags::mem_threadgroup);
```
awni, line 204: That barrier may be uneccessary?

awni (approved):
Amazing speedup!! LGTM!

mlx/backend/metal/kernels/sdpa_vector.h
angeloskath: This is actually a great catch. It should be inside the loop. Same for the 1 pass. The reasoning is it makes sure the thread group all reads the block at the same time so one simdgroup cannot just run ahead. I had seen it provides a small improvement but then in one of the edits it probably got restored back outside the loop.

angeloskath: I take it back... I tested again and I get mixed results which is probably why I reverted it in some previous commit. I will remove it since indeed conceptually it needn't be there.

PR #1601, opened 2024-11-19
Fix cmake version

awni:
Fixes #1595 for when `CMAKE_OSX_DEPLOYMENT_TARGET` is empty.


barronalex (approved):
Nice! Thanks for the quick fix!

PR #1612, opened 2024-11-21
Contiguous op + primitive

awni:
Adds a contiguous operation and primitive to facilitate getting contiguous data out of MLX. Useful for e.g. serialization, networking, and possibly bindings to other languages.

Helps with e.g. https://github.com/ml-explore/mlx/discussions/1608


awni:
@angeloskath I notice for distributed ops we always do the contiguity check inside the raw op on the CPU, like [here](https://github.com/ml-explore/mlx/blob/main/mlx/distributed/mpi/mpi.cpp#L333).

It's probably not a big deal but it's also pretty easy to change it to do the copy on the requested stream:

- Since we have `eval_cpu` and `eval_gpu` split out for the MPI ops we could easily do that inside those functions instead and then require that the raw ops get contiguous arrays.
- We could put a call to `contiguous` in the op implementation itself. 
 
I kind of prefer the first option since it doesn't make the graph larger... but I guess both are pretty reasonable. Thoughts?

barronalex (approved):
Looks great!

angeloskath:
The reasoning was that a distributed backend may allow non contiguous transfers so we don't need to copy. In practice the array is almost always contiguous since it is the concatenation of a lot of weights or the output of a matmul etc.

angeloskath (approved):

PR #1613, opened 2024-11-21
3 and 6 bit quantization

barronalex:
Add 3 and 6 bit quantization options. These use the same absmax quantization technique as the existing quants.

In all the kernels, we read 3 uint8 values for the 3/6 bit `qmm` in place of 1 uint32 for the powers of two. For 3 bit, that corresponds to 8 values, for 6 bit, 4 values. Otherwise the kernels stay pretty much the same.

## Performance

- There's a nice speed up from 8->6 bit despite the more awkward bit packing.
- On M2 Ultra 3bit is faster than 4 bit, but it's the other way around on M1 Max.
- The 2/4/8 bit performance is unchanged.

Screenshot 2024-11-21 at 14 49 53

Performance below is from some zero shot reasoning benchmarks in [lm_eval](https://github.com/EleutherAI/lm-evaluation-harness). 

- Seems like the 6 bit could be a nice tradeoff over 8 bit since there's minimal accuracy loss and it's faster.
- There's a bigger drop off from 4->3 bits. More to come on addressing this in follow up PRs.

Screenshot 2024-11-21 at 14 54 05


python/src/fast.cpp
```diff
@@ -161,49 +161,6 @@ void init_fast(nb::module_& parent_module) {
             array: The output array.
       )pbdoc");
 
-  m.def(
```
barronalex, line 164: I removed the quantize with scales and bias option for now because I don't think it was being used and it was complicating the code slightly. Happy to add it back in later if need be.

mlx/backend/common/quantized.cpp
```diff
@@ -22,13 +45,12 @@ void _qmm(
...
     int K) {
   constexpr int bitmask = (1 << bits) - 1;
-  constexpr int pack_factor = 32 / bits;
+  constexpr int pack_factor = bits == 3 ? 8 : bits == 6 ? 4 : 8 / bits;
+  constexpr int bytes_per_pack = (bits == 3 || bits == 6) ? 3 : 1;
   constexpr int packs_in_group = group_size / pack_factor;
-  const int Ng = N / group_size;
-  const int Nw = N / pack_factor;
 
   for (int m = 0; m < M; m++) {
```
barronalex, line 52: I ran a microbench and seems like this doesn't affect the cpu `qmm` throughput:

```
python benchmarks/python/comparative/bench_mlx.py quant_matmul_t_32_4 --size 1x1x1024 --size 1x2048x128 --size 1x2048x32 --size 1x2048x32 --dtype float16 --dtype uint32 --dtype float16 --dtype float16 --cpu
```

Before: 6.2653ms
After: 6.2657ms

barronalex:
Quick check that I didn't slow down model quantization for 2/4/8:

Before:
Quant time: 3.0901970863342285
CPU Quant time: 20.717879056930542

After:
GPU Quant time: 3.2021162509918213
CPU Quant time: 19.711817026138306

awni:
Wow, these results are really nice!!

mlx/backend/metal/kernels/quantized.h
```diff
@@ -442,25 +628,34 @@ METAL_FUNC void qmv_fast_impl(
...
+  constexpr int pack_factor = bits == 3 ? 8 : bits == 6 ? 4 : 32 / bits;
+  constexpr int bytes_per_pack = power_of_2_bits ? 1 : 3;
   constexpr int values_per_thread = pack_factor * packs_per_thread;
   constexpr int block_size = values_per_thread * SIMD_SIZE;
   constexpr int scale_step_per_thread = group_size / values_per_thread;
 
+  // When bits is a power of two, read 1 uint32_t at a time
+  // When bits is 3 or 6, read 3 uint8_ts at a time
+  using W_T =
+      typename ConditionalType<power_of_2_bits, uint32_t, uint8_t>::type;
```
angeloskath, line 644: Although I love the conditional type, I think we could just always use uint8 can't we?

Just change the bytes per pack to 4. Regardless the loading functions read in whatever type is convenient.

angeloskath (approved):
It looks fantastic.

I left a comment about the conditional type and bytes per pack constexpr. I think it applies in all cases in the kernels. I think some code can be simplified by using always uint8 but I don't know if it will have speed implications (most likely not since it is only for adjusting the pointer location).

mlx/backend/metal/kernels/quantized.h
barronalex: I tried this and it was a bit slower for some reason, hence the conditional type stuff. Maybe I can try tuning that way a bit more, it would be nicer.

angeloskath: Huh interesting. We can merge it as is and investigate later. This basically should result in exactly the same code for 2, 4 and 8 bits.

barronalex: Sounds good to me. For prompt processing it didn't make a difference so it's all uint8 in the `qmm`s.

PR #1616, opened 2024-11-22
Cleaner `qmv`/`qvm`

barronalex:
Turns out there's no speed difference so adding @angeloskath's nice suggestion from #1613.


angeloskath (approved):
👍

PR #1617, opened 2024-11-22
version bump 0.21

awni:
🚀


examples/extensions/requirements.txt
```diff
@@ -1,4 +1,4 @@
 setuptools>=42
 cmake>=3.24
-mlx>=0.18.1
+mlx>=0.21.0
```
awni, line 3: This should be bumped because of changes to the cmake package + device.h headers used in the extension.

barronalex (approved):
🚀

PR #1629, opened 2024-11-26
Fix some leaks and races

awni:
- Fixes 3 memory leaks because of not having objc pools in place :(
- Fixes race condition setting wired and cache limits (needed a mutex)

Closes https://github.com/ml-explore/mlx-examples/issues/1124

We should consider running our test suite with `OBJC_DEBUG_MISSING_POOLS=YES` and checking for leaks in the future as it is quite easy to introduce leaks between the C++ and objc interface. It doesn't quite run clean:
- one of the PyTorch tests triggers it as well (e.g. no MLX involved) 
- we leak some stuff at program termination..


angeloskath (approved):
🙏

PR #1634, opened 2024-11-29
feat(nn): add CausalConv1d

khirotaka:
## Proposed changes
This pull request introduces `CausalConv1d`, a 1-dimensional causal convolution layer designed for time series tasks such as [WaveNet](https://arxiv.org/abs/1609.03499) and [Temporal Convolutional Network](https://arxiv.org/abs/1803.01271).

The `CausalConv1d` layer ensures that each output at time \( t \) is computed using inputs up to and including time \( t \), without depending on future inputs. This is crucial for maintaining causality in time series predictions.

## Checklist

### DIFF
--- a/docs/src/python/nn/layers.rst
+++ b/docs/src/python/nn/layers.rst
@@ -16,6 +16,7 @@ Layers
    BatchNorm
    CELU
    Conv1d
+   CausalConv1d
    Conv2d
    Conv3d
    ConvTranspose1d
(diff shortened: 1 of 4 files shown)


awni:
This is really nicely implemented. Since it's a fairly simple modification of Conv1D that's used only in a few specific cases, let's hold off on adding it to `mlx.nn`.

We can revisit this if it becomes a common need.

PR #1637, opened 2024-12-02
Fix transformer decoder post norm LN

### LINKED ISSUE #1636: [Bug?] Wrong layer norm used in TransformerDecoderLayer
Shouldn't it be `self.ln2(x+y)` instead of `ln1` on line 241? The PyTorch version uses `self.norm2` (https://github.com/pytorch/pytorch/blob/b2fe1b94093e902e3d59d940fc1f9ed3ac12e826/torch/nn/modules/transformer.py#L1085)

https://github.com/ml-explore/mlx/blob/974bb54ab2f8e8450f856ac215df160374080b33/python/mlx/nn/layers/transformer.py#L235-L241

awni:
Closes #1636


barronalex (approved):
👍

PR #1638, opened 2024-12-02
Mixed quantization parameters in nn.quantize

barronalex:
Allow `class_predicate` to return a dict parameters for `to_quantized` as well as a bool.

This lets you specify a mixed quantization, e.g. embed/lm_head in 6 bits, everything else in 4 bits.

### DIFF
--- a/python/mlx/nn/layers/quantized.py
+++ b/python/mlx/nn/layers/quantized.py
@@ -1,7 +1,7 @@
 # Copyright © 2023-2024 Apple Inc.
 
 import math
-from typing import Callable, Optional
+from typing import Callable, Optional, Union
 
 import mlx.core as mx
 from mlx.nn.layers.base import Module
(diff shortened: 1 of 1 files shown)


python/mlx/nn/layers/quantized.py
```diff
@@ -197,7 +206,7 @@ def _extra_repr(self):
         out_dims, in_dims = self.weight.shape
         in_dims *= 32 // self.bits
         return (
-            f"input_dims={in_dims}, output_dims={out_dims}, bias={'bias' in self},"
+            f"input_dims={in_dims}, output_dims={out_dims}, bias={'bias' in self}, "
```
angeloskath, line 209: ❤️

angeloskath (approved):
Looks great! I like that it is backwards compatible!

PR #72, opened 2023-12-08
AdamW implementation, without correction

jbarrow:
Implementation of AdamW, but without correction of the first and second moments. Following the convention of the Adam implementation in the repo.

If you want the paper convention of correction, that implementation is here:

https://github.com/jbarrow/mlxllama/blob/cb2a808eefafc340db91e7d8f97ca334f6afff2a/mlxllama/optim.py

PR #1640, opened 2024-12-03
Support bias correction in Adam and AdamW optimizers

mt-caret:
## Proposed changes

The original implementation of AdamW doesn't include bias correction (https://github.com/ml-explore/mlx/pull/72). I found this causes problems when using it to learn a trivial task such as memorization using a [GPT2-like architecture](https://github.com/karpathy/build-nanogpt) whereas the equivalent pytorch implementation doesn't exhibit this behavior; the changes in this PR resolve the issue.

I've manually confirmed that this matches Pytorch behavior up to some small floating point differences, and also replicated a simpler version of it in mlx tests which breaks for main but passes in my branch.

## Checklist

### DIFF
--- a/python/mlx/optimizers/optimizers.py
+++ b/python/mlx/optimizers/optimizers.py
@@ -398,7 +398,7 @@ class Adam(Optimizer):
     r"""The Adam optimizer [1].
 
     Our Adam implementation follows the original paper and omits the bias
-    correction in the first and second moment estimates. In detail,
+    correction in the first and second moment estimates by default. In detail,
 
     [1]: Kingma, D.P. and Ba, J., 2015. Adam: A method for stochastic
     optimization. ICLR 2015.
@@ -423,12 +423,14 @@ class Adam(Optimizer):
         learning_rate: Union[float, Callable[[mx.array], mx.array]],
         betas: List[float] = [0.9, 0.999],
         eps: float = 1e-8,
+        bias_correction: bool = False,
     ):
         super().__init__()
 
         self._maybe_schedule("learning_rate", learning_rate)
         self.betas = betas
         self.eps = eps
+        self.bias_correction = bias_correction
 
     def init_single(self, parameter: mx.array, state: dict):
         """Initialize optimizer state"""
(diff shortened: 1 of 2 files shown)


guillaume-osmo:
yes thanks for adding this option this is indeed impactful

python/mlx/optimizers/optimizers.py
```diff
@@ -398,7 +398,7 @@ class Adam(Optimizer):
     r"""The Adam optimizer [1].
 
     Our Adam implementation follows the original paper and omits the bias
-    correction in the first and second moment estimates. In detail,
+    correction in the first and second moment estimates by default. In detail,
```
angeloskath, line 401: I would simply remove that comment and document the `bias_correction` argument below in the args.

python/mlx/optimizers/optimizers.py
```diff
@@ -449,15 +459,17 @@ def apply_single(self, gradient: mx.array, parameter: mx.array, state: dict):
...
 
 
 class AdamW(Adam):
     r"""The AdamW optimizer [1].
 
     Following the above convention, in contrast with [1], we do not use bias
-    correction in the first and second moments for AdamW. We update the weights
-    with a weight_decay (:math:`\lambda`) value:
+    correction in the first and second moments for AdamW by default. We update
+    the weights with a weight_decay (:math:`\lambda`) value:
```
angeloskath, line 472: Same as for Adam.

angeloskath (changes requested):
Thanks for adding this. Several people have requested this so I think it is time to add it.

I left some comments regarding the implementation. I think it will be both cleaner and have the benefit of no extra cost in the case where we don't use `mx.compile`. Let me know what you think.

python/mlx/optimizers/optimizers.py
mt-caret: I've adjusted the comment accordingly. Does this look reasonable to you?

python/mlx/optimizers/optimizers.py
```diff
@@ -449,15 +459,17 @@ def apply_single(self, gradient: mx.array, parameter: mx.array, state: dict):
         state["m"] = m
         state["v"] = v
 
-        return parameter - lr * m / (mx.sqrt(v) + eps)
+        return parameter - step_size * m / (
+            mx.sqrt(v) / bias_correction2_sqrt + eps
+        ).astype(gradient.dtype)
```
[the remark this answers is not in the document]
mt-caret: That makes sense; done.

angeloskath (approved):
Looks good, thanks!

angeloskath:
@mt-caret Could you skip the test if there is no PyTorch available? Then I can merge thanks!

mt-caret:
I should've looked more carefully at the other tests which use torch, my bad. Fixed!

mt-caret:
...and also applied the pre-commit thing 😅

PR #1644, opened 2024-12-03
Compile changes if stream changes

### LINKED ISSUE #1624: [BUG] Compiled functions always use the first stream they were called with
```python
@mx.compile
def fun(x):
    return mx.abs(mx.negative(x))

x = mx.array([1,2,3])

with mx.stream(mx.default_stream(mx.cpu)):
    mx.eval(fun(x))

with mx.stream(mx.default_stream(mx.gpu)):
    mx.eval(fun(x))
```

The `fun` will always run on the `cpu` because it was first called with that stream as the default.

This is not desirable and an easy source of performance issues (e.g. https://github.com/ml-explore/mlx-examples/pull/1122).

Here is what I think is the expected behavior:

```python
# Runs on the current default stream of the current default device when the function is called
@mx.compile
def fun(x):
    return mx.abs(mx.negative(x))

# negative runs on the default stream of the cpu
# abs runs on the default stream of the gpu
@mx.compile
def fun(x):
    return mx.abs(mx.negative(x, stream=mx.cpu), stream=mx.gpu)

# negative runs on stream1
# abs runs on stream2
@mx.compile
def fun(x):
    return mx.abs(mx.negative(x, stream=stream1), stream=stream2)

x = mx.array([1,2,3])

with mx.stream(mx.default_stream(mx.gpu)):
    mx.eval(fun(x))

with mx.stream(mx.default_stream(mx.cpu)):
    mx.eval(fun(x))
```

awni:
If the default stream or device change then we recompile the function.  

I looked into the alternative of updating the stream dynamically but it was quite a bit more complicated, especially considering the fact that we often use different primitives for different devices.

Another option is to update the compiled primitive only if the default device changes and dynamically update the stream. This is still complex since we need to track if the stream was set as a default or set explicitly. Presumably if you set it explicitly then you don't want it to change under your feet.

Closes #1624

### DIFF
--- a/mlx/compile.cpp
+++ b/mlx/compile.cpp
@@ -211,6 +211,8 @@ std::uintptr_t get_function_address(const std::function<T(U...)>& fun) {
 class CompilerCache {
  public:
   struct CacheEntry {
+    CacheEntry(Stream stream) : stream(stream) {};
+    Stream stream;
     std::vector<array> inputs;
     std::vector<array> outputs;
     std::vector<array> tape;
@@ -227,6 +229,7 @@ class CompilerCache {
       const std::vector<uint64_t>& constants) {
     // Find the cache entries for |fun_id|.
     std::vector<CacheEntry>& entries = cache_[fun_id];
+
     // Compare if 2 arrays have same shape and dtype.
     auto has_same_shape_and_dtype = [shapeless](
                                         const std::vector<array>& in1,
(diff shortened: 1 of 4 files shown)


angeloskath (approved):
Looks great! Tiny change and I think it gets us >90% of the benefit.

awni:
Just FYI there is a flaky test on linux. It's unrelated to this PR. I'll look into it.

PR #1645, opened 2024-12-03
Shape and Strides 1 / N

awni:
Adds a typedef for shape and strides to facilitate switching them if needed.

- `Strides` is currently an alias for `std::vector<size_t>` with the intention of switching it to `std::vector<int64_t>`
- `Shape` is currently an alias for `std::vector<int>` with possibility of eventually switching uint or maybe a larger precisoin
- Might be interesting to try a small vector optimization for these types which is another reason to prefer the typedef over just switching the value type.

If you all are cool with the naming / approach then I can start making the change to use the typedefs and try and switch to a `int64_t` stride.

CC @davidkoski

### DIFF
--- a/mlx/array.cpp
+++ b/mlx/array.cpp
@@ -31,7 +31,7 @@ array::array(const std::complex<float>& val, Dtype dtype /* = complex64 */)
 }
 
 array::array(
-    std::vector<int> shape,
+    Shape shape,
     Dtype dtype,
     std::shared_ptr<Primitive> primitive,
     std::vector<array> inputs)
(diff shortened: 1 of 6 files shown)


davidkoski:
I wonder if the type should be `ptrdiff_t` (difference between two pointers in the same allocation) since that is what a stride describes?  I am sure `int64_t` works as well

awni:
> I wonder if the type should be `ptrdiff_t`

Yea.. that could work too. I kind of prefer if it has a fixed size (and is not platform dependent) for the sake of serialization. Though I don't think MLX is used on many/any 32-bit machines so for all intents and purposes they are equivalent.

angeloskath (approved):
Looks good!

PR #1650, opened 2024-12-05
allow compiling lambdas in C++

awni:
Allows compiling lambda functions in C++.

There is an edge case difference between Python and C++ to note. We have slightly less ability to cache captured lambdas in C++ than in  Python because they aren't reference counted pointers in C++. In Python:

```python
f = lambda ...
cf = mx.compile(f) 
# calling cf doesn't recompile (unless you change input shape/typ)

# Assuming cf is still in scope
df = mx.compile(f) # also doesn't recompile
```

The second one works because `f` is basically a shared pointer to an underlying object and we use the pointer value as the function id and hold a reference to it inside the compiled `cf`. If `cf` goes out of scope then we will in fact recompile `df`.

In C++

```c++
auto f = [=](...) { ... };  # captures something
auto cf = compile(f);
# calling cf doesn't recompile (unless you change input shape/typ)
auto df = compile(f); # recompiles
```

`f` is implicitly cast to a `std::function` and so we don't have the address of the original lambda. But even if we did it's a bad idea to use if because `f` is a temporary object so the address can belong to a new lambda when `f` is destroyed and we can't detect when the underlying `f` is destroyed.

I don't really think this is an issue because compiling once and calling many times is the sensible usage. However, if it becomes an issue we can provide a `compile(shared_ptr<std::function>)` overload and then safely not recompile if you did:

```c++
compile(std::make_share(f));
```

### DIFF
--- a/mlx/compile.h
+++ b/mlx/compile.h
@@ -17,6 +17,17 @@ std::function<std::vector<array>(const std::vector<array>&)> compile(
     std::vector<array>(fun)(const std::vector<array>&),
     bool shapeless = false);
 
+// Convert capture-less lambdas to function pointers.
+template <
+    typename F,
+    typename = std::enable_if_t<
+        std::is_convertible_v<F, decltype(+std::declval<F>())>>>
+std::function<std::vector<array>(const std::vector<array>&)> compile(
+    F&& f,
+    bool shapeless = false) {
+  return compile(+f, shapeless);
+}
+
 /** Globally disable compilation.
  * Setting the environment variable ``MLX_DISABLE_COMPILE`` can also
  * be used to disable compilation.
--- a/mlx/compile.h
+++ b/mlx/compile.h
@@ -10,7 +10,11 @@ enum class CompileMode { disabled, no_simplify, no_fuse, enabled };
 
 /** Compile takes a function and returns a compiled function. */
 std::function<std::vector<array>(const std::vector<array>&)> compile(
-    const std::function<std::vector<array>(const std::vector<array>&)>& fun,
+    std::function<std::vector<array>(const std::vector<array>&)> fun,
+    bool shapeless = false);
+
+std::function<std::vector<array>(const std::vector<array>&)> compile(
+    std::vector<array>(fun)(const std::vector<array>&),
     bool shapeless = false);
 
 /** Globally disable compilation.
--- a/tests/compile_tests.cpp
+++ b/tests/compile_tests.cpp
@@ -739,6 +739,29 @@ TEST_CASE("test compile lambda") {
   auto out = compile(+fun)({array(-1)});
   CHECK_EQ(out[0].item<int>(), 1);
 
+  decltype(compile(nullptr)) c_local_fun;
+  {
+    auto local_fun = [](const std::vector<array>& inputs) {
+      return std::vector<array>{abs(inputs[0])};
+    };
+    c_local_fun = compile(+local_fun);
+  }
+
+  // This is ok even though local_fun is out of scope
+  out = c_local_fun({array(-1)});
+  CHECK_EQ(out[0].item<int>(), 1);
+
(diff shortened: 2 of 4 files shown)


tests/compile_tests.cpp
```diff
@@ -730,3 +730,53 @@ TEST_CASE("test compile change streams") {
...
+
+  auto out = compile(+fun)({array(-1)});
+  CHECK_EQ(out[0].item<int>(), 1);
+
+  decltype(compile(nullptr)) c_local_fun;
+  {
+    auto local_fun = [](const std::vector<array>& inputs) {
+      return std::vector<array>{abs(inputs[0])};
+    };
+    c_local_fun = compile(+local_fun);
```
angeloskath, line 747: I understand what is going on but I kinda hate that this is allowed. It will be a head scratcher for many people down the road. Not sure it is changeable or even documentable cause it is in the calling code and not in MLX.

The weirdness comes from the fact that a capture-less lambda is convertible to a function pointer (let alone the + operator weirdness) and persists further than the pointer pointing to it.

angeloskath (approved):
This is such a great change! 🙏

tests/compile_tests.cpp
awni: Right so this is a really interesting case (sorry I meant to mention it in the PR comment).

If a lambda is capture-less it is convertible to a function pointer and so we don't need to manage its lifetime (we can treat it just like a plain old function). What I did there was an suboptimal attempt to enable that optimization.

What happens if you don't add the `+` is you get a compiler error which for sure will be confusing. We could document that.. or we could just remove the specialization for now and treat capture-less lambdas the same way we treat captured ones. (I was exploring a third option which used compile time checking if the lambda has a capture and using a specialization if not, but I didn't get it to work yet and it gets pretty messy in the header which I don't love).

angeloskath: Yeah exactly. I think treating it as a normal lambda would be ok. It would incur just the cost of allocating a pointer on the heap (that simply points to a function pointer).

It isn't too bad honestly. I just don't think a lot of people will get it the first time and the fact that you have to use the `+` idiom to not get a compiler error is kinda bad.

angeloskath: Oh I guess another benefit is that you can "recompile" the same capture-less lambda and it won't recompile cause it will be just a pointer while if we treat it as a lambda it will be recompiled.

awni: Right that is the main benefit I was thinking of. But I don't like the `+` thing either so let's pull remove it and we can try to do a better fix later.

angeloskath: If the only problem is the `+` then having a pointer type to the function sth like `using ArrayFunPtr = std::vector<array>(*)(const std::vector<array>&);` and then used as

```cpp
auto local_fun = [](const std::vector<array>& inputs) {
  return std::vector<array>{abs(inputs[0])};
};
c_local_fun = compile((ArrayFunPtr)local_fun);
```

is probably more understandable. (didn't compile it but it should work or with `static_cast<ArrayFunPtr>(...)`)

mlx/compile.h
```diff
@@ -10,9 +10,24 @@ enum class CompileMode { disabled, no_simplify, no_fuse, enabled };
...
+// Convert capture-less lambdas to function pointers.
+template <
+    typename F,
+    typename = std::enable_if_t<
+        std::is_convertible_v<F, decltype(+std::declval<F>())>>>
+std::function<std::vector<array>(const std::vector<array>&)> compile(
+    F&& f,
+    bool shapeless = false) {
+  return compile(+f, shapeless);
+}
```
awni, line 29: @angeloskath this works. It gets the benefit of compiling the capture-less lambda without needing any explicit casting of `+`. Not too bad of a template monstrosity..

PR #1651, opened 2024-12-06
feat: Add orthogonal initializer and corresponding tests

### LINKED ISSUE #1586: Add orthogonal initialization
Hi, 

I would like to use [orthogonal initialization](https://pytorch.org/docs/stable/nn.init.html#torch.nn.init.orthogonal_) as I would like to use it for a project of mine. Is it planned to add it in the near future?

NripeshN:
closes #1586


awni (approved):
Thanks!

PR #1652, opened 2024-12-06
Chore: Add default language in pre-commit and bump hooks

NripeshN:
I had the following issue with black when `default_language_version` was not mentioned:

```
black....................................................................Failed
- hook id: black
- exit code: 1

Python 3.12.5 has a memory safety issue that can cause Black's AST safety checks to fail. Please upgrade to Python 3.12.6 or downgrade to Python 3.12.4
```

Added default language as 3.10 and bumped up the hooks to solve this issue

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,13 +1,17 @@
+default_language_version:
+  python: python3.10
+
 repos:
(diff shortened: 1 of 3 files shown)


awni (approved):
Looks good, thanks!

awni:
Do you mind updating MLX Examples with the same Python settings as well?

PR #1653, opened 2024-12-06
More primitives for compiling with shapeless

awni:
Added several more primitives to allow compiling with `shapeless=True` (and hopefully exporting with `shapeless=True`).

- `Arange`
- `Gather`
- `Transpose`
- `Concatenate`
- `Matmul`
- `QuantizedMatmul`
- `fast::AffineQuantize`
- `fast::LayerNorm`
- `fast::RMSNorm`
- `fast::RoPE`
- `fast::ScaledDotProductAttention`

### DIFF
--- a/mlx/compile.cpp
+++ b/mlx/compile.cpp
@@ -1,5 +1,4 @@
 // Copyright © 2023-2024 Apple Inc.
-
 #include <cstdlib>
 #include <map>
 #include <unordered_map>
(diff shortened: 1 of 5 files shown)


angeloskath (approved):
Looks good. There isn't really a reason to not do a shapeless compile for all primitives is there? Except having to write output shape functions.

awni:
> There isn't really a reason to not do a shapeless compile for all primitives is there? 

I think for almost all primitives we should allow it.

A couple primitives are somewhat dangerous to allow for shapeless compile. Mostly `Reshape` and `Broadcast` where the output shape is hardcoded and there is not currently a way to infer it from the input shape. It might still be nice to allow them to be shapelessly compiled because you may know that the shapes for that part of the graph are fixed.. something to think about.

PR #1654, opened 2024-12-06
Fix a flaky test and the pre-commit hooks.

awni:
Fix a flaky test and the pre-commit hooks.

Having the default language specified breaks things if you don't have the version available so it doesn't seem like a good idea to specify it.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,6 +1,3 @@
-default_language_version:
-  python: python3.10
-
 repos:
 -   repo: https://github.com/pre-commit/mirrors-clang-format
     rev: v19.1.4
--- a/python/tests/test_array.py
+++ b/python/tests/test_array.py
@@ -1,5 +1,6 @@
 # Copyright © 2023-2024 Apple Inc.
 
+import gc
 import operator
 import pickle
 import resource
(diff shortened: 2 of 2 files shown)


.pre-commit-config.yaml
NripeshN, line 1: Hi @awni,

As mentioned on my previous PR(https://github.com/ml-explore/mlx/pull/1652#issue-2721913665), this would break on few machines. It is best to setup a default python version to maintain consistency. Happy to create a new PR to fix this, will wait on your confirmation to add it on the other repos too.


```
(venv) nripeshn@Nripeshs-MacBook-Pro ~/D/P/m/e/extensions (pr/1685)> pre-commit run --all-files
clang-format.............................................................Passed
black....................................................................Failed
- hook id: black
- exit code: 1

Python 3.12.5 has a memory safety issue that can cause Black's AST safety checks to fail. Please upgrade to Python 3.12.6 or downgrade to Python 3.12.4

isort....................................................................Passed
cmake-format.............................................................Passed
```

awni: Yea.. the problem is setting a default version requires that the default version be present. So it basically fixes the python version you can use for pre-commit hooks.

Maybe it's simpler if you upgrade or downgrade your Python instead?

PR #1655, opened 2024-12-06
Fix performance bug in qvm

barronalex:
I assumed this [refactor](https://github.com/ml-explore/mlx/pull/1616) wouldn't matter for the `qvm` since it doesn't make a difference for the `qmv` but turns out it really slows things down at long context:

Before:
```
Prompt: 9 tokens, 4.092 tokens-per-sec
Generation: 100 tokens, 15.318 tokens-per-sec
Peak memory: 5.410 GB
```
After:
```
Prompt: 9 tokens, 29.152 tokens-per-sec
Generation: 100 tokens, 29.119 tokens-per-sec
Peak memory: 5.410 GB
```


awni:
That's pretty wild. Does this mean 6-bit will be very slow for long-context KV cache compared to 8-bit?

awni (approved):
LGTM, thanks for debugging that!

Issue #1513, opened 2024-10-22
[Feature] Windows support


zcbenz:
I tried building MLX on Windows, of course it didn't build but I think it is not very hard to make it work there.

A few problems I met so far:

* CMake can not find OpenBLAS automatically on Windows, I had to hardcode the path:
  `-DBLAS_LIBRARIES=openblas64/lib/libopenblas.a -DLAPACK_LIBRARIES=openblas64/lib/libopenblas.a -DBLAS_INCLUDE_DIRS=openblas64/include -DLAPACK_INCLUDE_DIRS=openblas64/include`
* MSVC has no `_Complex` but `lapack_complex_float` is defined to it.
* gguf-tools does not support Windows but we can turn off it with `-DMLX_BUILD_GGUF=OFF`.
* `io/load.h` uses unix system calls for IO, I can make it work on Windows by replacing them with `stdio.h` equivalents.
* The `make_compiled_preamble.sh` needs a Windows version.
* `mx.compile` is going to need many works to find and call MSVC.

And there will likely be much more problems since compilations are mostly blocked by `io/load.h`.

I don't really like Windows nor enjoy working on it, but the majority of world is using it and I figured out whatever I'm going to work on I would need Windows support to reach wider audience. What do you think?

zcbenz:
I found someone already did some work on this: https://github.com/lighttransport/mlx/commits/lte/

awni:
Thanks for all the PRs to get this working. I wonder what you think about a good long-term strategy for supporting Windows. 

We don't have plans to test or optimize regularly for Windows machines. So that might make maintenance a little tricky -> compilation could inadvertently break, performance might be slow, etc.

Luckily so far many of the PRs to get windows support have been quite simple (and in some cases improve the quality of the code) so it's not been a difficult decision to merge them. But I'm also wondering about what the right strategy is for PRs which introduce more complexity?e

I don't have answers to the above questions.. but maybe a good thing to discuss a bit more while you are in the process of experimenting on Windows.

zcbenz:
I'll add Windows build to node-mlx, so at least the C++ interface will be tested routinely, which is how the x64 mac build is being tested.

It is fine if an unsupported build breaks in master and relies on contributors for fixes, most large open source projects I worked with work this way. The key is to leave good comments for the untested code path so maintainers know whether to leave them be or just remove them.

[...]

PR #1658, opened 2024-12-07
Download OpenBLAS binaries when building with MSVC

zcbenz:
## Proposed changes

There is currently no distribution of OpenBLAS on Windows that can be found by `find_package`, this change makes CMake download and link with the official release on GitHub.

## Checklist


zcbenz:
With some testing I found out that the official binaries were built with MinGW and are not really usable for compilation with native ABI (MSVC), I'll seek another solution.

PR #1659, opened 2024-12-07
Work around MSVC error C2580

zcbenz:

It seems to be a bug of MSVC that it can not distinguish between:

```c++
  array& operator=(array&& other) && = delete;
  array& operator=(array&& other) & = default;
```

and gives error:

```
C:\cygwin\home\zcbenz\codes\mlx\mlx/array.h(70,10): error C2580: 'mlx::core::array
 &mlx::core::array::operator =(mlx::core::array &&) &': multiple versions of a def
aulted special member functions are not allowed [C:\cygwin\home\zcbenz\codes\mlx\b
uild\mlx.vcxproj]
```

### DIFF
--- a/mlx/array.h
+++ b/mlx/array.h
@@ -67,7 +67,9 @@ class array {
   array& operator=(array&& other) && = delete;
 
   /** Default copy and move constructors otherwise. */
+#if !defined(_MSC_VER)
   array& operator=(array&& other) & = default;
+#endif
   array(const array& other) = default;
   array(array&& other) = default;
 


awni:
It looks like it was fixed alreaady? https://developercommunity.visualstudio.com/t/error-c2580-multiple-versions-of-a-defaulted-speci/468431

Is it reasonable to use a newer version of MSVC? The fix looks like from 4+ years ago..

zcbenz:
Updated my Visual Studio and it is gone 👍 

The bug was 4 years ago but the fix only got released a few months ago, we are quite lucky this time.

PR #1660, opened 2024-12-07
Use correct complex type for MSVC

zcbenz:

OpenBLAS's default define for complex type does not compile on MSVC, check the "Change in complex types for Visual Studio 2017 and up" in its document:
https://github.com/OpenMathLib/OpenBLAS/blob/develop/docs/install.md


awni (approved):
Thanks! Will merge when tests clear.

PR #1662, opened 2024-12-07
Use auto* for pointers

zcbenz:

MSVC can not deduce `auto` to be a pointer and would complain `delete p` not working.


awni (approved):
I think that might be incorrect in the spec, but it deduced it correctly somehow with clang.

PR #1664, opened 2024-12-07
Do not pass integers to isnan

zcbenz:

Passing integers to isnan is a compilation error in MSVC, see also https://stackoverflow.com/questions/61646166/how-to-resolve-fpclassify-ambiguous-call-to-overloaded-function.


awni (approved):
Thanks! Will merge after tests clear!

PR #1665, opened 2024-12-07
Using math defines requires switch in MSVC

zcbenz:

### DIFF
--- a/mlx/ops.cpp
+++ b/mlx/ops.cpp
@@ -1,4 +1,10 @@
 // Copyright © 2023-2024 Apple Inc.
+
+#ifdef _MSC_VER
+// Required for ussing M_PI in MSVC.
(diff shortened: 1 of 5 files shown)


mlx/ops.cpp
```diff
@@ -1,4 +1,10 @@
 // Copyright © 2023-2024 Apple Inc.
+
+#ifdef _MSC_VER
+// Required for using M_PI in MSVC.
+#define _USE_MATH_DEFINES
+#endif
+
```
awni, line 7: I think we can keep it simple and just leave these `#defines _USE_MATH_DEFINES` ungaurded. They work w/ clang and I believe g++ as well.

zcbenz: Updated.

awni (approved):
Thanks!

PR #1666, opened 2024-12-07
Provide a pread implementation for MSVC

zcbenz:

There is no `pread` on Windows, but we can easily emulate it with win32 APIs.


awni (approved):
Thanks!

PR #1668, opened 2024-12-07
Metal-cpp version bump

ameszy:
Apple has released the stable version of Metal-cpp for macOS 15 and iOS 18. CMakeLists.txt is updated to build with it instead of the beta one.


awni (approved):
Thanks!

awni (changes requested):
Please run the format hooks.

ameszy:
I admit I omitted the format hooks thinking it was just a small change so probably I won't need those. I'm sorry for that and for the inconvenience caused. The new commit now includes the style fix done by those scripts.

awni (approved):
Thanks!

PR #1671, opened 2024-12-07
Use int64 stride everywhere

awni:
Compiles and tests clear with the switched type. Sorry for the big diff. It's 99% boiler plate changes with a few simplifications (removing unneeded template parameters, using cleaner instantiations, etc) here and there.

Should probably run a couple benchmarks before merging this just to make sure nothing is affected..seems unlikely but you never know.

I will leaving enabling negative strides in `as_strided` and potentially a copy-less negative strided slice for a follow up. I think most of our kernels should "just work" with negative strides.. but that could end up being more involved.


angeloskath (approved):
Looks great! I love that you took the time to also refactor repeated instantiations of `check_transpose`, very nice!

I didn't do a `grep size_t` and `grep std::vector<size_t>` etc I assume you did. Let me know if you want me to go over the greps as well.

awni:
> I didn't do a grep size_t and grep std::vector<size_t> etc I assume you did.

I didn't actually. I was hoping the compiler would catch everything for me 😅 . I'll do a quick grep just to make sure

PR #1672, opened 2024-12-08
Convert filesystem path to std::string explicitly

zcbenz:

The `std::filesystem::path` can not be silently converted to `std::string` on Windows because the native path uses a different char type.


awni (approved):
Thanks!

awni (approved):
Thanks!

PR #1673, opened 2024-12-08
Use int64_t instead of ssize_t

zcbenz:

`ssize_t` is a POSIX type and not available in MSVC, I think `int64_t` should do the same.


awni (approved):
Thanks!

jeethu:
`ptrdiff_t` is available in MSVC and everywhere else. IMO, that's a better replacement for `ssize_t` than hardcoding it to `int64_t`.

PR #1674, opened 2024-12-08
Build OpenBLAS from source code for MSVC

zcbenz:

When building with MSVC, there is no central package manager that allows installing and searching third party packages, so we can not find BLAS with `find_package`.

This PR adds a `MLX_BUILD_BLAS_FROM_SOURCE` flag to download and build OpenBLAS from source code which is turned on for MSVC by default. Other platforms can also set it on to test the same build.

The `dlfcn-win32` dependency is also changed to build from source for the same reason.


CMakeLists.txt
```diff
@@ -113,16 +114,52 @@ elseif(MLX_BUILD_METAL)
   target_link_libraries(mlx PUBLIC ${METAL_LIB} ${FOUNDATION_LIB} ${QUARTZ_LIB})
 endif()
 
+if(WIN32)
+  if(MSVC)
```
awni, line 118: I'm just curious why we check both of these. Is it common to use a compiler other than MSVC on windows?

CMakeLists.txt
```diff
@@ -113,16 +114,52 @@ elseif(MLX_BUILD_METAL)
...
+    # Download and build OpenBLAS from source code.
+    FetchContent_Declare(
+      openblas
+      GIT_REPOSITORY https://github.com/OpenMathLib/OpenBLAS.git
+      GIT_TAG v0.3.28
+      EXCLUDE_FROM_ALL)
+    set(BUILD_STATIC_LIBS ON) # link statically
+    set(NOFORTRAN ON) # msvc has no fortran compiler
+    FetchContent_MakeAvailable(openblas)
+    target_link_libraries(mlx PRIVATE openblas)
```
awni, line 158: You linked openblas privately here (which seems correct). Should we update it below as well to link privately for consistency?

CMakeLists.txt
zcbenz: Some people use MinGW (https://github.com/JuliaPackaging/Yggdrasil/pull/9761), which provides layers for some POSIX APIs.

CMakeLists.txt
zcbenz: Updated.

awni (approved):
Ty!

CMakeLists.txt
```diff
@@ -152,14 +189,7 @@ if(MLX_BUILD_CPU)
     message(STATUS "Blas lib " ${BLAS_LIBRARIES})
     message(STATUS "Blas include " ${BLAS_INCLUDE_DIRS})
     target_include_directories(mlx PRIVATE ${BLAS_INCLUDE_DIRS})
-    target_link_libraries(mlx PUBLIC ${BLAS_LIBRARIES})
-
-    if(WIN32)
-      find_package(dlfcn-win32 REQUIRED)
```
stemann, line 158: Sorry to see this gone!

A `USE_SYSTEM_DLFCN_WIN32` option could have kept it.

PR #1675, opened 2024-12-08
Make compiled preamble work in MSVC

zcbenz:

This PR adds a `make_compiled_preamble.ps1` script for MSVC build which basically duplicates what the shell script does. The `mx.compile` API still needs some massive changes to work, but this change will be able to unblock the compilation of MLX first.


awni (approved):
LGTM, thanks!

PR #1677, opened 2024-12-09
Fix compile hasher for string constants.

### LINKED ISSUE #1676: [BUG] Unexpected recompilation in a training loop when using certain APIs
**Describe the bug**

I was debugging some recompilation issues in a training loop that eventually caused memory to be exhausted and training to fail. 
I'm not totally sure this is a bug, but it seemed unexpected:

In 1) below, when we decode a string from bytes prior to the call to the compiled training step (as you might from a dataloader), it seems to capture some additional state during compilation that changes by the moment, so every call to the function will recompile.

If you use a string directly like 2), the function is compiled as expected and reused.

```python
from __future__ import annotations
from functools import partial
import time

import mlx.core as mx
import mlx.nn as nn
from mlx.optimizers import Adam

class MLP(nn.Module):
    def __init__(self):
        super().__init__()
        self.proj_in = nn.Linear(32, 128)
        self.mlp = [nn.Linear(128, 128) for _ in range(3)]
        self.proj_out = nn.Linear(128, 1)

    def __call__(self, x, text):
        x = self.proj_in(x)
        for layer in self.mlp:
            x = nn.relu(layer(x))
        return self.proj_out(x)


def train(model, optimizer):
    def loss_fn(m, x, text, y):
        y_pred = m(x, text)
        return nn.losses.mse_loss(y_pred, y)

    state = [model.state, optimizer.state]

    def train_step(x, text, y):
        loss_and_grad_fn = nn.value_and_grad(model, loss_fn)
        loss, grads = loss_and_grad_fn(model, x, text, y)

        return loss

    model.train()

    text_placeholder = "foo".encode("utf-8")

    for step in range(50):
        x = mx.random.normal((100, 32))
        y = mx.random.normal((100, 1))

        # 1) captures constant state and recompiles
        text = [bytes(text_placeholder).decode("utf-8")]

        # 2) no recompilation
        # text = ["foo"]

        loss = train_step(x, text, y)
        mx.eval(state)
        
        print(f"Step {step}, loss: {loss}")
        time.sleep(0.5)


train(MLP(), Adam(learning_rate=1e-3))
```

I used this patch to verify when recompilation was happening (it would be super nice to have an environment variable for this!)

```diff
diff --git a/mlx/compile.cpp b/mlx/compile.cpp
index 9bfdb6ee9..2a7a7acc8 100644
--- a/mlx/compile.cpp
+++ b/mlx/compile.cpp
@@ -1,5 +1,6 @@
 // Copyright © 2023-2024 Apple Inc.
 #include <cstdlib>
+#include <iostream>
 #include <map>
 #include <unordered_map>
 #include <unordered_set>
@@ -269,6 +270,24 @@ class CompilerCache {
       if (has_same_shape_and_dtype(inputs, entry.inputs) &&
           constants == entry.constants) {
         return entry;
+      } else if (constants != entry.constants) {
+        std::cout << " -- constants not equal" << std::endl;
+        std::cout << "constants = [";
+        for (size_t i = 0; i < constants.size(); ++i) {
+            std::cout << constants[i];
+            if (i < constants.size() - 1) {
+                std::cout << ", ";
+            }
+        }
+        std::cout << "]" << std::endl;
+        std::cout << "entry constants = [";
+        for (size_t i = 0; i < entry.constants.size(); ++i) {
+            std::cout << entry.constants[i];
+            if (i < entry.constants.size() - 1) {
+                std::cout << ", ";
+            }
+        }
+        std::cout << "]" << std::endl;
       }
     }
     // Otherwise append a new cache entry
@@ -850,6 +869,8 @@ std::function<std::vector<array>(const std::vector<array>&)> compile(
 
     // No matching cache entry existed, so compile
     if (entry.empty) {
+      std::cout << "Recompiling..." << std::endl;
+      
       // Mark the entry as not empty since we are about to fill it
       entry.empty = false;
       // Set the constants
```

I can't tell exactly what the constants being captured are (RNG state?), but it looks like this:

```
Recompiling...
Step 0, loss: 0.9941226840019226
 -- constants not equal
constants = [18446744073709551533, 18446744073709551557, 18446744073709551533, 4376530480, 18446744073709551557, 18446744073709551521]
entry constants = [18446744073709551533, 18446744073709551557, 18446744073709551533, 4376535088, 18446744073709551557, 18446744073709551521]
Recompiling...
Step 1, loss: 1.1245471239089966
 -- constants not equal
constants = [18446744073709551533, 18446744073709551557, 18446744073709551533, 4376536304, 18446744073709551557, 18446744073709551521]
entry constants = [18446744073709551533, 18446744073709551557, 18446744073709551533, 4376535088, 18446744073709551557, 18446744073709551521]
 -- constants not equal
constants = [18446744073709551533, 18446744073709551557, 18446744073709551533, 4376536304, 18446744073709551557, 18446744073709551521]
entry constants = [18446744073709551533, 18446744073709551557, 18446744073709551533, 4376530480, 18446744073709551557, 18446744073709551521]
Recompiling...
[... repeats with the cache growingly indefinitely ...]
```

Does this seem expected? Am I doing something dumb?

 - OS Version: macOS 15.1.1
 - Version: 0.21.1

awni:
closes #1676

### DIFF
--- a/python/src/transforms.cpp
+++ b/python/src/transforms.cpp
@@ -444,8 +444,7 @@ struct PyCompiledFun {
         auto r = obj.attr("__hash__")();
         constants.push_back(nb::cast<int64_t>(r));
       } else if (nb::isinstance<nb::int_>(obj)) {
-        auto r = nb::cast<int64_t>(obj);
-        constants.push_back(static_cast<uint64_t>(r));
+        constants.push_back(nb::cast<int64_t>(obj));
       } else if (nb::isinstance<nb::float_>(obj)) {
         auto r = nb::cast<double>(obj);
         constants.push_back(*reinterpret_cast<uint64_t*>(&r));
--- a/python/src/transforms.cpp
+++ b/python/src/transforms.cpp
@@ -433,19 +433,19 @@ struct PyCompiledFun {
         auto d = nb::cast<nb::dict>(obj);
         constants.push_back(dict_identifier);
         for (auto item : d) {
-          auto r = item.first.attr("__hash__");
-          constants.push_back(*reinterpret_cast<uint64_t*>(&r));
+          auto r = item.first.attr("__hash__")();
+          constants.push_back(nb::cast<int64_t>(r));
           recurse(item.second);
         }
       } else if (nb::isinstance<array>(obj)) {
         inputs.push_back(nb::cast<array>(obj));
         constants.push_back(array_identifier);
       } else if (nb::isinstance<nb::str>(obj)) {
-        auto r = obj.attr("__hash__");
-        constants.push_back(*reinterpret_cast<uint64_t*>(&r));
+        auto r = obj.attr("__hash__")();
+        constants.push_back(nb::cast<int64_t>(r));
       } else if (nb::isinstance<nb::int_>(obj)) {
         auto r = nb::cast<int64_t>(obj);
-        constants.push_back(*reinterpret_cast<uint64_t*>(&r));
+        constants.push_back(static_cast<uint64_t>(r));
       } else if (nb::isinstance<nb::float_>(obj)) {
         auto r = nb::cast<double>(obj);
         constants.push_back(*reinterpret_cast<uint64_t*>(&r));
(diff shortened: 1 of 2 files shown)


angeloskath (approved):
Thanks!

PR #1678, opened 2024-12-09
Use Py_ssize_t in python bindings

zcbenz:

Same with https://github.com/ml-explore/mlx/pull/1673 that the use of `ssize_t` has to be replaced. In this PR I use `Py_ssize_t ` instead since they are going to be assigned to `Py_buffer` and other types do not compile.

This PR also includes a tiny fix that too small to be a separate PR.


awni (approved):
Ty!!

PR #1679, opened 2024-12-09
Work around function std name collision in MSVC

zcbenz:

There is a bug in MSVC that, if the `std` function is exposed to current scope with `using namespace mlx::core`, all following `std::xxx` invocations that try to access the `namespace std` will fail with very unhelpful errors:

```
C:\cygwin\home\zcbenz\codes\mlx\python\src\array.cpp(28,18): error C2874: using-de
claration causes a multiple declaration of 'std' [C:\cygwin\home\zcbenz\codes\mlx\
build\temp.win-amd64-cpython-311\Release\mlx.core\python\src\core.vcxproj]
      C:\Program Files\Microsoft Visual Studio\2022\Community\VC\Tools\MSVC\14.42.
  34433\include\numeric(22,1):
      see declaration of 'std'
C:\Users\zcbenz\AppData\Roaming\Python\Python311\site-packages\nanobind\include\na
nobind\nb_cast.h(160,27): error C2872: 'std': ambiguous symbol [C:\cygwin\home\zcb
enz\codes\mlx\build\temp.win-amd64-cpython-311\Release\mlx.core\python\src\core.vc
xproj]
  (compiling source file '../../../../../../python/src/convert.cpp')
      C:\Program Files\Microsoft Visual Studio\2022\Community\VC\Tools\MSVC\14.42.
  34433\include\numeric(22,1):
      could be 'std'
      C:\Users\zcbenz\AppData\Roaming\Python\Python311\site-packages\nanobind\incl
  ude\nanobind\nb_cast.h(160,27):
      or       'std'
```

(MSVC thinks `std::` is ambiguous.)

There are 2 solutions, one taken by this PR is to rename the `std` function to `std_dev` and use it internally, while adding a `std` alias only available in public `mlx.h` header. It makes minimum changes to existing code with the cost of adding a new header.

The other solution would be replacing the `using namespace mlx::core` with `namespace mx = mlx::core` and use `mx::xxx()` instead of `xxx()` in the python bindings code. I prefer this one but it would result in a relatively large PR.

### DIFF
--- a/mlx/mlx.h
+++ b/mlx/mlx.h
@@ -13,7 +13,7 @@
 #include "mlx/fft.h"
 #include "mlx/io.h"
 #include "mlx/linalg.h"
-#include "mlx/ops.h"
+#include "mlx/ops_public.h"
 #include "mlx/random.h"
 #include "mlx/stream.h"
 #include "mlx/transforms.h"
--- a/mlx/ops.cpp
+++ b/mlx/ops.cpp
@@ -1704,17 +1704,17 @@ array var(
   return var(a, std::vector<int>{axis}, keepdims, ddof, to_stream(s));
 }
 
-array std(
+array std_dev(
     const array& a,
     bool keepdims,
     int ddof /* = 0*/,
     StreamOrDevice s /* = {}*/) {
   std::vector<int> axes(a.ndim());
   std::iota(axes.begin(), axes.end(), 0);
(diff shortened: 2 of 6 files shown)


awni:
I kind of prefer the second solution as well ... it feels like the right way to fix name collisions and how the C++ namespace should be used in user code. It is a big and tedious diff though 😓

zcbenz:
Will go the 2nd solution, feels like the right way to go.

PR #1681, opened 2024-12-09
Fix library output directory for MSVC

zcbenz:

For MSVC you have to specify library output dir for every config otherwise the file would be put into `out/Debug`, `out/Release`, etc. instead of `out` directly.


python/src/CMakeLists.txt
```diff
@@ -29,8 +29,21 @@ if(NOT MLX_PYTHON_BINDINGS_OUTPUT_DIRECTORY)
...
+               LIBRARY_OUTPUT_DIRECTORY_DEBUG
+               ${MLX_PYTHON_BINDINGS_OUTPUT_DIRECTORY}
+               LIBRARY_OUTPUT_DIRECTORY_RELWITHDEBINFO
+               ${MLX_PYTHON_BINDINGS_OUTPUT_DIRECTORY}
+               LIBRARY_OUTPUT_DIRECTORY_MINSIZEREL
+               ${MLX_PYTHON_BINDINGS_OUTPUT_DIRECTORY})
+else()
+  set_target_properties(core PROPERTIES LIBRARY_OUTPUT_DIRECTORY
+                                        ${MLX_PYTHON_BINDINGS_OUTPUT_DIRECTORY})
+endif()
```
awni, line 46: Why not just use the first condition for every platform and discard the check for MSVC?

zcbenz: Updated. It seems that some build systems other than MSVC also require this to work.

awni (approved):
Thank you 👍

PR #1682, opened 2024-12-09
No reshapes in quantized embedding

awni:
Don't do reshapes in quantized embedding layer as they are not needed. No perceptible change in inference speed .. but simpler more compilable graph.

Also fixed some casting bugs where nanobind would cast scalar arrays to integers by mistake.


angeloskath (approved):
👍

PR #1683, opened 2024-12-10
Add squeeze primitive

awni:
This adds an explicit `squeeze` primitive. The purpose is to enable shapeless compilation of anything that uses squeeze.

It turns out we use squeeze in a lot of places (e.g. after a gather / take, after any kind of reduction, etc). In those cases we end up doing explicit reshapes which break shapeless compilation. With a squeeze primitive we can shapelessly compile these without any problems.

In a similar vein it might make sense to add an `ExpandDims` primitive. The primitive's are pretty simple.. so this seems worth-while..


mlx/primitives.cpp
```diff
@@ -3,6 +3,8 @@
 // Required for using M_2_SQRTPI in MSVC.
 #define _USE_MATH_DEFINES
 
+#include <iostream> // TODO
```
angeloskath, line 6: forgotten :-)

angeloskath (approved):
Nice!

PR #1684, opened 2024-12-10
Fix shared library not exporting symbols on Windows

zcbenz:

On Windows a shared library does not have its symbols exported by default, and usually devs would add a dllexport prefix for each symbol to export in source code.

In this PR I use a feature of CMake to automate it, and due to the feature not exporting global variables, I also change the `global_formatter` variable to be a getter function instead.


awni (approved):
Looks good, will merge when tests clear.

PR #1685, opened 2024-12-10
Remove "using namespace mlx::core" in benchmarks/examples

zcbenz:

In order to make the code compilable with MSVC we must avoid introducing function `std` into current scope (see https://github.com/ml-explore/mlx/pull/1679 for more information).

This PR converts the code in benchmarks/examples to use `mx::` namespace alias, as an alternative solution to #1679. The downside is lots of lines get overwritten and it would be harder to check the code history in future.

I'll convert the code in python/src if you are OK with this. 

(The refactoring was mostly done with the help of AI editors, I check the code for a few times but errors can happen.)

### DIFF
--- a/examples/extensions/axpby/axpby.cpp
+++ b/examples/extensions/axpby/axpby.cpp
@@ -60,7 +60,7 @@ mx::array axpby(
   return mx::array(
       /* const std::vector<int>& shape = */ out_shape,
       /* mx::Dtype dtype = */ out_dtype,
-      /* std::unique_ptr<Primitive> primitive = */
+      /* std::unique_ptr<mx::Primitive> primitive = */
       std::make_shared<Axpby>(to_stream(s), alpha, beta),
       /* const std::vector<mx::array>& inputs = */ broadcasted_inputs);
 }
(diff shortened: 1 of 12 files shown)


awni (approved):
Looks good, thanks! Will merge when the tests clear.

examples/extensions/axpby/axpby.cpp
```diff
@@ -32,9 +32,9 @@ namespace mlx::core {
  *  Follow numpy style broadcasting between x and y
  *  Inputs are upcasted to floats if needed
  **/
-array axpby(
-    const array& x, // Input array x
-    const array& y, // Input array y
+mx::array axpby(
```
awni, line 35: I'm not sure why this is inside the `mlx::core` namespace. I think it makes more sense to keep it in a top-level / user name space and use the `namespace mx = ...`

awni: Do you mind also updating `axbpy.h`?

zcbenz: I have fixed the building and updated the extension to use its own namespace.

zcbenz:
Sorry I forgot to test the code on a M chip mac, can you run tests again?

PR #1687, opened 2024-12-10
`ExpandDims` primitive

awni:
Building on https://github.com/ml-explore/mlx/pull/1683. Adds an `ExpandDims` primitive for dimension expansion.

With this we can shapelessly compile most matmuls, reductions, stack, gathers, etc.

E.g. this works now:

```python
a = mx.array([0.0, 1.0, 2.0])
b = mx.array([0.0, 1.0, 2.0])

fun = mx.compile(lambda a, b: a @ b, shapeless=True)
```

And this:

```python
fun = mx.compile(lambda x: x.sum(0), shapeless=True)
```


angeloskath (approved):
Nice! Perhaps we should just merge this instead of `squeeze` and `expand` separately 😀

PR #1688, opened 2024-12-10
Allow no copy negative strides in as_strided and slice

awni:
Allows arrays to have negative strides. Mostly the size caps have to change to make sure things fit in `int` instead of `uint` when we are doing strided (general) ops.


angeloskath (approved):
Very straightforward, nice!

PR #1689, opened 2024-12-11
Remove "using namespace mlx::core" in python/src

zcbenz:

In order to make the code compilable with MSVC we must avoid introducing function `std` into current scope (see https://github.com/ml-explore/mlx/pull/1679 for more information).

Similar to https://github.com/ml-explore/mlx/pull/1685, this PR removes `using namespace mlx::core` in `python/src` code.


awni (approved):
Thanks for the big change!

PR #1690, opened 2024-12-11
Install mlx.dll into the same dir with python bindings on Windows

zcbenz:

On Windows the statically linked DLLs must be put in the same directory with the python bindings (or in the system DLL locations), and there is no compiler option to change that. This PR changes the install destination of `mlx.dll` to `CMAKE_INSTALL_PREFIX` instance of `CMAKE_INSTALL_BINDIR`.

The dlfcn-win32 dependency is also changed to be linked statically.


CMakeLists.txt
```diff
@@ -125,10 +125,11 @@ if(WIN32)
     dlfcn-win32
     GIT_REPOSITORY https://github.com/dlfcn-win32/dlfcn-win32.git
     GIT_TAG v1.4.1
+    SOURCE_SUBDIR use-an-unexist-path-to-skip-cmake-of-dlfcn-win32
     EXCLUDE_FROM_ALL)
   FetchContent_MakeAvailable(dlfcn-win32)
   target_include_directories(mlx PRIVATE "${dlfcn-win32_SOURCE_DIR}/src")
-  target_link_libraries(mlx PRIVATE dl)
+  target_sources(mlx PRIVATE ${dlfcn-win32_SOURCE_DIR}/src/dlfcn.c)
```
awni, line 132: it might be simpler to `set(BUILD_SHARED_LIBS OFF)` ?

zcbenz: Updated the code to set `BUILD_SHARED_LIBS`, also bumped the required cmake version for using `block`.

awni (approved):
Thanks! I'm ok with the cmake bump.

awni:
We should update our documentation though I think it specifies cmake 3.24. I can send a patch here or you can add it if you have time.

zcbenz:
Updated the docs and requirements.

awni:
Thank you!

zcbenz:
Sorry there was a cmake link error, can you run tests again?

stemann:
This broke building/installing shared libs on (MinGW) Windows for non-Python applications: https://buildkite.com/julialang/yggdrasil/builds/15595#0193cee0-171b-4396-ace6-6e62a0345b11

E.g., `libmlx.dll` ends up in a non-standard location (directly in CMAKE_INSTALL_PREFIX it seems).

awni:
It should be statically linked 🤔

zcbenz:
Static linking seems to be the simplest solution, I'll send a PR to do static linking in `setup.py` and revert the change on install location.

PR #1692, opened 2024-12-11
Flatten and unflatten

awni:
- Primitives for flatten / unflatten
- op + binding `unflatten`

Let's us do some more dynamic work in the graph. For example instead of:

```python
mx.reshape(q, (B, L, num_heads, -1))
```

we do:

```python
mx.unflatten(q, -1, (num_heads, -1))
```

and it can be dynamic w.r.t to sequence length.


angeloskath (approved):
Looks great. Just to be clear, this still is in reality only partially shapeless since the unflatten still defines that part of a shape as a constant. I don't think we have other ops in the shapeless list that have the same property. 🤔

awni:
> Just to be clear, this still is in reality only partially shapeless since the unflatten still defines that part of a shape as a constant.

Yes. 

> I don't think we have other ops in the shapeless list that have the same property

Actually I snuck one by you before.  `gather` has that property as well since it takes `slice_size` which is based on the shape of the input. 

Maybe it makes sense to have some error handling in the shape inference routines.. I think it's going to be quite difficult to have shapeless be completely dynamic but it's still pretty nice to have shapeless as an option..even with some runtime error allowances.

awni:
After this I can shapelessly compile the generation part of LLama inference with a KV cache using concatenate.

The prompt processing doesn't use the fused SDPA and there is a non-fused broadcast as wells a broadcast and a reshape for the mask. So broadcast would be the next thing to tackle (or we use add mask support + the head dim for the fused SDPA).

PR #1694, opened 2024-12-12
Make sure the inputs are deleted in reversed order

zcbenz:

The old code relied on a behavior that elements inserted to `unordered_map` are iterated in reversed order, which is however not the case with MSVC.

I don't know if it is intentional that the inputs must be destroyed in reversed order, but without doing so would result in crashes when running autograd tests.

### DIFF
--- a/mlx/array.cpp
+++ b/mlx/array.cpp
@@ -1,6 +1,6 @@
 // Copyright © 2023-2024 Apple Inc.
 #include <functional>
-#include <unordered_map>
+#include <set>
 
 #include "mlx/array.h"
 #include "mlx/ops.h"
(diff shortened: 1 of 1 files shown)


awni:
> I don't know if it is intentional that the inputs must be destroyed in reversed order, but without doing so would result in crashes when running autograd tests.

What kind of crash? It's not clear to me why those need to be done in reverse order .. maybe it's correct, not sure.

zcbenz:
If you checkout the branch and change the code:

```diff
     // The inputs are destroyed in reversed order.
-    for (auto a = input_arrs.rbegin(); a != input_arrs.rend(); ++a) {
+    for (auto a = input_arrs.begin(); a != input_arrs.end(); ++a) {
```

run the cpp test you can reproduce the crash (or assertion more accurately).

My observation is, some arrays not available for deletion got added to `for_deletion`.

awni:
Closing in favor of #1714

PR #1697, opened 2024-12-13
Make mx.compile work on Windows

zcbenz:

This PR adds code to invoke MSVC on Windows. Most of the new code are about getting the paths information of Visual Studio.

### DIFF
--- a/mlx/backend/common/compiled_cpu.cpp
+++ b/mlx/backend/common/compiled_cpu.cpp
@@ -284,6 +284,11 @@ inline void build_kernel(
 
   NodeNamer namer;
 
+#ifdef _MSC_VER
+  // Export the symbol
+  os << "__declspec(dllexport) ";
+#endif
+
   // Start the kernel
   os << "void " << kernel_name << "(void** args) {" << std::endl;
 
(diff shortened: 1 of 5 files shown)


mlx/backend/common/compiled_cpu.cpp
```diff
@@ -102,11 +193,30 @@ void* compile(
...
+        source_file_name.str(),
+        shared_lib_name.str(),
+        libpaths);
+#else
+    std::string build_command = fmt::format(
+        "g++ -std=c++17 -O3 -Wall -fPIC -shared '{0}' -o '{1}'",
+        source_file_path,
+        shared_lib_path);
+#endif
+    auto return_code = system(build_command.c_str());
```
awni, line 219: I think it would be good to encapsulated the compiler into it's own class to contain the if-defs and make this more maintainable / extensible.

Something like:

```c++
static std::string JitCompiler::build_command(
  const std::string& source_file_path, const std::string& shared_lib_name)

// And we could include any special source code headers needed in the class as well. 
static std::string JitCompiler::get_header()
```

zcbenz: I think it is good timing to move the code to a separate file, I have updated the PR.

awni (approved):
Looks great, thanks for the refactor / improvements!

zcbenz:
I pushed a change to fix the lint error, somehow pre-commit did not run automatically on my Windows env.

PR #1698, opened 2024-12-13
Read/write files in binary mode

zcbenz:

On Windows we need to explicitly open files in binary mode to write the data as-is.


awni (approved):
Thanks!

PR #1699, opened 2024-12-14
Make sure gguf_ctx is closed when error happens

zcbenz:

This PR makes sure the `gguf_ctx` is closed when exception is thrown in `save_gguf` and `load_gguf`. Without doing would leave the files opened and result in test failures on Windows, as it does not allow writing to already opened files.


awni (approved):
Nice! Though I thought you weren't building gguf on windows?

zcbenz:
I'm making gguf-tools work on Windows.

PR #1700, opened 2024-12-14
Use psutil to get memory info on Windows

zcbenz:

The `resource` package does not work on Windows, and according to https://github.com/giampaolo/psutil/issues/1096, the equivalent of `ru_maxrss` is `peak_wset` on Windows.


awni (approved):
LGTM!

PR #1701, opened 2024-12-14
Shorter kernel name for Windows

zcbenz:

Windows has a famous [maximum path length limitation](https://learn.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation) issue, while the OS has solved it the [Visual Studio is still haunted by it](https://developercommunity.visualstudio.com/t/compiler-cant-find-source-file-in-path-/10221576).

### DIFF
--- a/mlx/backend/common/compiled_cpu.cpp
+++ b/mlx/backend/common/compiled_cpu.cpp
@@ -80,7 +80,8 @@ void* compile(
     std::ostringstream file_name;
     file_name
         << std::string_view(kernel_name).substr(0, max_file_name_length - 16);
-    auto file_id = std::hash<std::string>{}(kernel_name);
+    auto file_id =
+        std::hash<std::string>{}(kernel_name.substr(max_file_name_length - 16));
     file_name << "_" << std::hex << std::setw(16) << file_id << std::dec;
     kernel_file_name = file_name.str();
   } else {
--- a/mlx/backend/common/compiled_cpu.cpp
+++ b/mlx/backend/common/compiled_cpu.cpp
@@ -68,10 +68,14 @@ void* compile(
   std::string source_code = source_builder();
   std::string kernel_file_name;
 
-  // Deal with long kernel names. Maximum length for files on macOS is 255
-  // characters. Clip file name with a little extra room and append a 16
-  // character hash.
+  // Deal with long kernel names. Maximum length for filename on macOS is 255
+  // characters, and on Windows the maximum length for whole path is 260. Clip
+  // file name with a little extra room and append a 16 character hash.
+#ifdef _WIN32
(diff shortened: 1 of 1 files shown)


mlx/backend/common/compiled_cpu.cpp
```diff
@@ -68,10 +68,14 @@ void* compile(
   std::string source_code = source_builder();
   std::string kernel_file_name;
 
-  // Deal with long kernel names. Maximum length for files on macOS is 255
-  // characters. Clip file name with a little extra room and append a 16
-  // character hash.
+  // Deal with long kernel names. Maximum length for filename on macOS is 255
+  // characters, and on Windows the maximum length for whole path is 260. Clip
+  // file name with a little extra room and append a 16 character hash.
+#ifdef _WIN32
```
awni, line 74: This totally fine to merge.. but with very short file-names it's not inconceivable to get collisions on file-names (which would be bad). Especially [given this](https://stackoverflow.com/questions/7968674/unexpected-collision-with-stdhash). Some ideas to help with that:

I might change the hash to only hash the part of the file name that is clipped to reduce probabilities of collisions. If the prefix is different there won't be a collision anyway. And if it's the same, the suffix must be different and we want to use the suffix for the hash.

We might also want to do some checks on the source.. but that is more involved and we can leave that for future work if it becomes an issue.

zcbenz: It will take quite a while before anyone actually starts using the Windows port seriously, I prefer leaving the problem for future.

awni: You can at least make this change or let me edit the PR:

```
    auto file_id =
        std::hash<std::string>{}(kernel_name.substr(max_file_name_length - 16));
```

It's simple enough and much safer.

zcbenz: Please edit the PR for me 🙏

awni: I tried but I don't think I have permission :\. Usually it's the default to allow .. but it didn't work in your case. Check out the guide [here](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork).

zcbenz: It is likely because my fork is in an org account, I'll update this PR later.

awni: Thank you 🙏

zcbenz: Update!

awni (approved):
Thanks!

PR #1702, opened 2024-12-14
Eval before cleanup so model file is unlocked

zcbenz:

Without evaluating the model weights the `Load` primitive will hold the weights file which mades it undeletable on Windows.


awni (approved):
LGTM

PR #1705, opened 2024-12-15
More shape type

awni:
- Use the `Shape` type everywhere that is needed to make the code compile if it's changed from e.g. `vector<int>` to `vector<int64_t>`
- Added a `ShapeElem` type to specify the element type of the `Shape`. It's pretty convenient to have during the transition and maybe even beyond that. Could use `decltype` to infer it but it's a bit verbose.


angeloskath (approved):
👍

PR #1708, opened 2024-12-16
Fix unflatten vjp

awni:
By some miracle this didn't fail a test when I added it..

Closes https://github.com/ml-explore/mlx-examples/issues/1168


barronalex (approved):
Nice!

PR #1714, opened 2024-12-17
Fix deletion of non-evaled arrays with siblings

awni:
We should only delete siblings if all siblings are deletable. Replacing #1694


awni:
@zcbenz would be helpful if you could verify this runs for you on windows now.

zcbenz:
I confirm this fixes the same crashes with #1694, thanks!

angeloskath (approved):
I will not say it so that I don't jinx it...

👍

PR #1715, opened 2024-12-18
Define NOMINMAX before windows.h

zcbenz:

Fix build failure with MSVC after #1706.


awni (approved):
Thanks!

PR #1716, opened 2024-12-18
Link python extension with mlx statically on Windows

zcbenz:

https://github.com/ml-explore/mlx/pull/1690 changed the install location of mlx.dll on Windows, which breaks MinGW build (see https://github.com/ml-explore/mlx/pull/1690#issuecomment-2545147230).

This PR reverts the change on the install location of mlx.dll, and makes the python extension link with MLX statically instead.

Note that the change in #1690 that forces `dlfcn-win32` to be linked statically is not reverted, otherwise users have to also ship a `dl.dll` along with `mlx.dll` when they build MLX as shared library.


setup.py
```diff
@@ -77,11 +77,16 @@ def build_extension(self, ext: CMakeExtension) -> None:
         # Pass version to C++
         cmake_args += [f"-DMLX_VERSION={self.distribution.get_version()}"]  # type: ignore[attr-defined]
 
-        if sys.platform.startswith("darwin"):
+        if platform.system() == "Darwin":
```
awni, line 80: Looks good overall.. but I'm just curious what if anything motivated the switch to `platform.system` from `sys.platform`?

zcbenz: MLX's codebase is using both `platform.system` and `sys.platform` but more code is using the former, so I chose the former for the new code and made this file use the same style.

awni (approved):

setup.py
```diff
@@ -77,11 +77,16 @@ def build_extension(self, ext: CMakeExtension) -> None:
...
             # Cross-compile support for macOS - respect ARCHFLAGS if set
             archs = re.findall(r"-arch (\S+)", os.environ.get("ARCHFLAGS", ""))
             if archs:
                 cmake_args += ["-DCMAKE_OSX_ARCHITECTURES={}".format(";".join(archs))]
+        if platform.system() != "Windows":
+            # On Windows DLLs must be put in the same dir with the extension
+            # while cmake puts mlx.dll into the "bin" sub-dir. Link with mlx
+            # statically to work around it.
+            cmake_args += ["-DBUILD_SHARED_LIBS=ON"]
```
stemann, line 89: I had a hard time reading this, as it said "statically" and appended `-DBUILD_SHARED_LIBS=ON` (because I missed the `!=` in `!= "Windows"`.

Wouldn't it be more straightforward to have `-DBUILD_SHARED_LIBS=ON` in the default `cmake_args`, and then append an extra `-DBUILD_SHARED_LIBS=OFF` when `platform.system() == "Windows"`? (IIUC, the last cmake arg should take precedence, right?)

zcbenz: Or more explicitly so readers don't have to check which arg takes precedence.

```python
if platform.system() = "Windows":
    cmake_args += ["-DBUILD_SHARED_LIBS=OFF"]
else:
    cmake_args += ["-DBUILD_SHARED_LIBS=ON"]
```

I'm not opposed to making the code more readable, I'm leaving the decision to @awni .

awni: I like the last option you suggest @zcbenz it is a lot easier to follow.

zcbenz: Updated.

PR #1721, opened 2024-12-18
Check mask dtype in SDPA

barronalex:
Make sure the mask dtype matches the output dtype in sdpa.

Discussed in https://github.com/ml-explore/mlx-examples/pull/1173


awni (approved):
Looks great! If for some reason it ends up being too string we can revisit.

PR #1722, opened 2024-12-19
Dynamic broadcasting for shapeless compile/export

awni:
- Add a `BroadcastAxes` primitive with an optional `ignore_axes` parameter. `ignore_axes` specifies the axes to ignore from the inputs during broadcasting. Can use this for a shapeless broadcast when we do batched matmuls and in several other places.

- Add `broadcast_arrays` to python. Same API as the NumPy function and can be used instead of `broadcast_to` if dynamic broadcasting is needed.

- Add a flag to `InTracing` to determine if we are in a "dynamic" trace. This way we don't blow up the graph with a bunch of unnecessary ops unless they are needed. Also guarded `NumberOfElements` with this for consistency.

- In order to make the shapes dynamic with broadcast, if we are in a dynamic / shapeless trace then instead of using a unary broadcast we use a n-ary broadcast as follows:

1.  The first output is a broadcast which takes the first input and all of the other inputs as inputs
2. The remaining outputs get themself as input as well as the first already broadcast output

This works the same for both the regular `Broadcast` and `BroadcastAxes`

Some more notes on the implementation:
- Originally I attempted doing the above with a multi-output primitive but this ended up being cleaner especially considering the interaction with compile which doesn't fuse multi-output primitives.
- The reason for having the new primitive (rather than adding `ignore_axes` as an optional input on the existing `Broadcast` is to make it easy to distinguish them (e.g. in compile) since we treat them quite differently.


mlx/ops.cpp
```diff
@@ -1320,37 +1321,98 @@ array broadcast_to(
...
   return outputs;
 }
 
+std::pair<array, array>
+broadcast_arrays(const array& a, const array& b, StreamOrDevice s) {
+  auto out = broadcast_arrays({a, b}, s);
+  return {out[0], out[1]};
+}
+
+std::pair<array, array> broadcast_arrays(
```
angeloskath, line 1396: I don't think anybody is actually using these two. Probably wrote them because they were used before but now the vector in/out is more convenient anyway so I think they can be deleted. They are internal to ops.cpp anyway.

mlx/ops.cpp
```diff
@@ -1320,37 +1321,98 @@ array broadcast_to(
...
-std::vector<array>
-broadcast_arrays(const array& a, const array& b, StreamOrDevice s /* = {} */) {
-  auto shape = broadcast_shapes(a.shape(), b.shape());
-  return {broadcast_to(a, shape, s), broadcast_to(b, shape, s)};
+/** Broadcast the input arrays against one another while ignoring the
+ * axes specified in `ignore_axes`. The `ignore_axes` should be:
+ * - negative values indicating axes from the end
+ * - sorted in increasing order
+ */
+std::vector<array> broadcast_arrays(
```
angeloskath, line 1329: Maybe make a note that this should be internal to ops.cpp ? Just in case we forget and expose it in ops.h .

Basically my fear is that we will somehow pass positive axes to ignore which will most likely segfault. Easy to fix but kind of a bad failure mode.

mlx/ops.cpp
```diff
@@ -1320,37 +1321,98 @@ array broadcast_to(
...
+    std::vector<int> ignore_axes,
+    StreamOrDevice s) {
+  if (inputs.size() <= 1) {
+    return inputs;
+  }
+  std::vector<array> outputs;
+  auto shape = BroadcastAxes::output_shape(inputs, ignore_axes);
+  for (auto& in : inputs) {
+    for (auto ax : ignore_axes) {
+      shape[shape.size() + ax] = in.shape(ax);
```
angeloskath, line 1340: Another maybe better solution would be to check that `shape.size() + ax < shape.size()` or sth like that.

mlx/ops.cpp
```diff
@@ -1320,37 +1321,98 @@ array broadcast_to(
...
+    }
+    if (in.shape() == shape) {
+      outputs.push_back(in);
+    } else if (detail::in_dynamic_tracing()) {
+      outputs.push_back(array(
+          shape,
+          in.dtype(),
+          std::make_shared<BroadcastAxes>(to_stream(s), ignore_axes),
+          outputs.empty() ? inputs : std::vector<array>{in, outputs.front()}));
```
angeloskath, line 1349: This is just to save work in compile right? Otherwise moving each array into first position would be the natural way of doing this right? For instance what if I only want to use the 2nd output. This way the first array will be broadcasted as well because we depend on it to compute the 2nd output. Maybe something like the following?

```cpp
std::swap(inputs[0], inputs[I]);
outputs.push_back(array(
          shape,
          inputs[0].dtype(),
          std::make_shared<BroadcastAxes>(to_stream(s), ignore_axes),
          inputs));
std::swap(inputs[0], inputs[i]);
```

angeloskath (approved):
Looks great! I left some minor comments nothing particularly important.

mlx/ops.cpp
awni: They are both used in a small number of places in this file. They are really just a convenience so one doesn't have to write the following every time you broadcast two arrays and need to extract the components.

```
auto out = broadcast_arrays({x, y});
x = out[0];
y = out[1];
```

mlx/ops.cpp
awni: Yes, will do. I should probably also add a check that they are all negative and sorted.

mlx/ops.cpp
awni: Yes that's exactly right, the first input is always broadcast if any of the outputs are evaluated. 

I think your suggestion is pretty reasonable.. I did indeed do it this way to avoid needing to redo the full broadcasting of all arrays during shape inference for each input. For a large number of inputs it might matter, but in practice I think they will be pretty similar since

1. We only use this right now for broadcast two arrays 
2. We almost always use both outputs since this is used for e.g. matmul, qmm, etc

So I'm happy to change it.

awni: There is an issue with this solution that I didn't notice before. Function transformations like VJP / JVP request the grad for the extra placeholder inputs to the broadcast primitive. 

I could do stop gradients on all the extra inputs before passing them to the primitive. It's not horrible, but yet another complication :\.

Maybe I should bite the bullet and make it a multi-output primitive. But there the challenge is to get compile to fuse and fix code generation for multi-output broadcasts.

awni: I ended up going with stop grad, it's not too complex. I also added https://github.com/ml-explore/mlx/pull/1759 which will optimize away those extra primitives from the tape when we compile so it should not be noticeable in most cases (we only use this primitive in shapeless compiles / export)

PR #1723, opened 2024-12-19
More lenient mask type check in SDPA

barronalex:
Apologies for the double PR, I did some more testing with mlx_lm and some of the models break with the stricter check.


awni (approved):
Perfect thanks for double checking that!

PR #1724, opened 2024-12-19
Allow offset to be an mx.array for `mx.fast.rope`

awni:
Fairly simple change... rope offset can be a scalar array or an int.

### DIFF
--- a/mlx/fast.cpp
+++ b/mlx/fast.cpp
@@ -338,34 +338,50 @@ array rope(
     bool traditional,
     float base,
     float scale,
-    int offset,
     bool forward,
     StreamOrDevice s) {
   auto& x = inputs[0];
+  auto& offset = inputs[1];
   if (x.ndim() < 3) {
     std::ostringstream msg;
     msg << "[rope] Input must have at least 3 dimensions but got input with "
         << x.ndim() << " dimensions.";
     throw std::invalid_argument(msg.str());
   }
-  if (inputs.size() == 2 &&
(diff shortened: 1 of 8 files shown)


python/src/ops.cpp
```diff
@@ -1367,9 +1371,9 @@ void init_ops(nb::module_& m) {
             dtype,
             s);
       },
-      "start"_a,
-      "stop"_a,
-      "step"_a = nb::none(),
+      "start"_a.noconvert(),
+      "stop"_a.noconvert(),
+      "step"_a.noconvert() = nb::none(),
```
awni, line 1376: I added this because nanobind will otherwise implicitly convert `mx.arange(mx.array(1))` to work since scalar arrays are convertible to Python integers. It's not technically a bug.. but I think it's better to avoid implicit graph evaluations like that.

mlx/fast.cpp
```diff
@@ -467,11 +483,23 @@ array rope(
...
+    const array& x,
+    int dims,
+    bool traditional,
+    std::optional<float> base,
+    float scale,
+    int offset,
+    const std::optional<array>& freqs /* = std::nullopt */,
+    StreamOrDevice s /* = {} */) {
+  return rope(
+      x, dims, traditional, base, scale, array(offset, uint32), freqs, s);
```
angeloskath, line 500: I think this should be `int32` here shouldn't it?

angeloskath (approved):
Nice! Left a single nitpick comment.

PR #1726, opened 2024-12-19
Add `mx.finfo` and use it when making causal mask

### LINKED ISSUE #1725: [BUG] `nn.MultiHeadAttention.create_additive_causal_mask` should not give NaN for low precision types
Possible fixes:
- remove type param
- upcast and downcast
- add finfo and use the actual min for the type

```
>>> import mlx.nn as nn
>>> import mlx.core as mx
>>> nn.MultiHeadAttention.create_additive_causal_mask(4, mx.float16) 
array([[nan, -inf, -inf, -inf],
       [nan, nan, -inf, -inf],
       [nan, nan, nan, -inf],
       [nan, nan, nan, nan]], dtype=float16)
```

awni:
Closes #1725

### DIFF
--- a/docs/src/python/data_types.rst
+++ b/docs/src/python/data_types.rst
@@ -66,3 +66,4 @@ documentation for more information. Use :func:`issubdtype` to determine if one
    Dtype
    DtypeCategory
    issubdtype
+   finfo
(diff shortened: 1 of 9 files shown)


barronalex (approved):
Thanks for fixing this! Looks great!

PR #1727, opened 2024-12-19
Shapeless slice update and broadcast when possible

awni:
- Allow slice update to be shapeless since we know it's output shape
- Allow broadcast to be shapeless when possible and throw an error if it's not possible: (condition is `broadcast_shapes(in.shape(), shape) == shape`)
- Remove redundant shapeless condition + checks. The shapeless validation was a redundant run-time check. The compiled function will already throw if a primitive does not have a shape inference implemented (e.g. `output_shapes`) so there was really no need for that extra check.
- `a[0] = x` uses slice update instead of scatter
- Simplify + speedup `slice::vjp` by using `slice_update` instead of scatter

### DIFF
--- a/mlx/primitives.cpp
+++ b/mlx/primitives.cpp
@@ -740,6 +740,13 @@ bool Broadcast::is_equivalent(const Primitive& other) const {
   return shape_ == b_other.shape_;
 }
 
+std::vector<Shape> Broadcast::output_shapes(const std::vector<array>& inputs) {
+  if (broadcast_shapes(inputs[0].shape(), shape_) != shape_) {
+    throw std::invalid_argument("[Broadcast] Unable to infer broadcast shape");
+  }
+  return {shape_};
+}
+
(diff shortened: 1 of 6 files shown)


mlx/primitives.cpp
```diff
@@ -3585,63 +3592,9 @@ std::vector<array> Slice::vjp(
...
-
-  // Concatenate all the indices and axes
-  inds.insert(inds.end(), single_inds.begin(), single_inds.end());
-  ind_axes.insert(
-      ind_axes.end(), single_ind_axes.begin(), single_ind_axes.end());
-
-  return {scatter_add(
-      zeros_like(primals[0], stream()), inds, cotan, ind_axes, stream())};
+  auto out = zeros_like(primals[0], stream());
+  return {slice_update(
```
angeloskath, line 3596: ❤️

angeloskath (approved):
Looks great! Love all the removals.

PR #1731, opened 2024-12-23
Fix Distributed Communication documentation (correct usage of `mx.distributed.init().size()` in `all_reduce_grads`)

danilopeixoto:
## Proposed changes

This fix updates the Distributed Communication documentation to reflect the correct usage of `mx.distributed.init().size()` in the `all_reduce_grads` function. The previous implementation incorrectly assumed `mx.distributed.init()` directly provided the number of processes.

## Checklist

### DIFF
--- a/docs/src/usage/distributed.rst
+++ b/docs/src/usage/distributed.rst
@@ -131,7 +131,8 @@ have to :func:`mlx.utils.tree_map` the gradients with following function.
 .. code:: python
 
     def all_avg(x):
-        return mx.distributed.all_sum(x) / mx.distributed.init().size()
+        N = mx.distributed.init().size()
+        return mx.distributed.all_sum(x) / N
 
 Putting everything together our training loop step looks as follows with
 everything else remaining the same.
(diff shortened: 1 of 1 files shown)


docs/src/usage/distributed.rst
```diff
@@ -131,7 +131,8 @@ have to :func:`mlx.utils.tree_map` the gradients with following function.
 .. code:: python
 
     def all_avg(x):
-        return mx.distributed.all_sum(x) / mx.distributed.init().size()
+        N = mx.distributed.init().size()
```
danilopeixoto, line 134: Explicitly execute the initialization first, even if `mx.distributed.all_sum(x)` is lazily evaluated.

docs/src/usage/distributed.rst
```diff
@@ -141,12 +142,13 @@ everything else remaining the same.
     from mlx.utils import tree_map
 
     def all_reduce_grads(grads):
-        N = mx.distributed.init()
```
angeloskath, line 144: That is indeed incorrect! Thanks for the fix.

docs/src/usage/distributed.rst
angeloskath: I don't think this is necessary.

angeloskath (changes requested):
Thanks for the fix! Could you change the `all_avg` function back and then I can merge this?

I think it is time to have a more general update to the distributed docs since the better option is to use `nn.utils.average_gradient` anyway.

docs/src/usage/distributed.rst
danilopeixoto: The [Undo aesthetic code changes](https://github.com/danilopeixoto/mlx/commit/280e38e5cca20ee04f3299b1b21cc623cef9e052) commit reversed that. Thanks!

angeloskath (approved):

PR #1732, opened 2024-12-24
Explicit parentheses around some logical operators

louen:
## Proposed changes

This minor change is here to satisfy my warning flags ([`-Wlogical-op-parentheses`](https://clang.llvm.org/docs/DiagnosticsReference.html#wlogical-op-parentheses)) which recommend making the priority between `&&` and `||` explicit with parentheses. The change is just cosmetic and does not alter the behavior, but it makes it a bit more explicit which operator gets evaluated in which order. 

## Checklist

### DIFF
--- a/mlx/array.cpp
+++ b/mlx/array.cpp
@@ -119,7 +119,7 @@ void array::eval() {
 }
 
 bool array::is_tracer() const {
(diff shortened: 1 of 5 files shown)


awni (approved):
Thanks!

Issue #1729, opened 2024-12-21
[Feature] Adan Optimizer


cvnad1:
@angeloskath 

I would like to add the latest improvement to [Adam Optimizer](https://optax.readthedocs.io/en/latest/api/optimizers.html#optax.adan) which is The ADAptive Nesterov momentum algorithm(Adan) Optimizer which is reference implementation from [paper](https://arxiv.org/abs/2208.06677).

I believe this will be a great addition to the MLX framework given its superior performance than Adam and given the fact it is already available in Optax.

cvnad1:
@awni closing the issue as I have raised a new issue in mlx-optimizers as suggested.

PR #1733, opened 2024-12-26
Added Adan Optimizer

cvnad1:
## Proposed changes

Added Adan Optimizer and the tests according to the issue #1729. Kindly go through the changes and share your thoughts.

## Checklist


cvnad1:
@angeloskath 

I went ahead and added the Adan Optimizer and also modified tests. Kindly review the code and give your valuable feedback.

stockeh:
If Adan is added to MLX, then what's the criterion for adding optimizers? 

[PyTorch](https://pytorch.org/docs/stable/optim.html#algorithms) is more conservative in their core lib relative to [Optax](https://optax.readthedocs.io/en/latest/api/optimizers.html). The [mlx-optimizers](https://github.com/stockeh/mlx-optimizers) package could serve as a location for new optimizers as it already has non-traditional implementations (e.g., ADOPT, LAMB, Muon, QHAdam, etc.).

awni:
Actually, I don't think we will merge this into MLX core.  In fact @cvnad1 I will close this. @stockeh may be interested in having a PR at to mlx-optimizers.

> If Adan is added to MLX, then what's the criterion for adding optimizers?

We don't have a strict criterion. But for the most part we aren't planning to add new optimizers unless there is a really compelling reason to add something.

PR #1735, opened 2024-12-27
Fix concatenate/slice_update vjp + reduce binary size

awni:
- Make sure inputs to concatenate and slice update have the same type otherwise the vjp has incorrect type. This will insert an extra copy in some cases.. but it's also pretty rare to, for example, concatenate arrays of different types. 
- `GeneralGeneral` copy on the GPU assumes the input and the output have the same type to reduce binary size. We don't ever use it unless this is the case.

Binary size pre: 83 MB
Binary size post: 78 MB


angeloskath (approved):
👍

PR #1736, opened 2024-12-27
MLX in C++ example

awni:
Adds:
- Very minimal example project using MLX in C++ with Cmake
- Documentation page on the same in the developer docs
- Convenience commands to get the `--cmake-dir` and `--version` using `mlx` as a python package. E.g. `python -m mlx --cmake-dir` gets the `cmake` directory for `find_package`. Simplifies using MLX in CMake as a dependency.

### DIFF
--- a/docs/src/dev/mlx_in_cpp.rst
+++ b/docs/src/dev/mlx_in_cpp.rst
@@ -1,0 +1,121 @@
+Using MLX in C++
+================
+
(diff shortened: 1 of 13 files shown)


angeloskath (approved):
I love this :-)

So far I have been building a separate mlx library for C++ dev. Didn't think of making this.

PR #1737, opened 2024-12-29
Fix export.cpp compilation with MSVC

zcbenz:
Compilation failed with `export.cpp(293,1): error C1061: compiler limit: blocks nested too deeply`, luckily it can be fixed by making code simpler.


awni (approved):
Thanks!!

PR #1741, opened 2024-12-30
Dynamic slicing

awni:
Adds a `slice` and `slice_update` which take a dynamic (`array`) starting location. The APIs are:

```c++
/** Slice an array with dynamic starting indices. */
array slice(
    const array& a,
    const array& start,
    std::vector<int> axes,
    Shape slice_size,
    StreamOrDevice s = {});

/** Update a slice from the source array with dynamic starting indices. */
array slice_update(
    const array& src,
    const array& update,
    const array& start,
    std::vector<int> axes,
    StreamOrDevice s = {});
``` 

For the `slice` the output shape is specified in `slice_size`.

I also bound these ops to python. For `slice` it's not so simple to use `getitem` since we don't know the shape so if we want it in python it probably needs to be as an op.

For `slice_update` we could use `setitem`. But it's a bit awkward since for `setitem` we would need to specify the slice to update like `a:b` and `b` is just ignored (it wouldn't make sense to check its value since it's an array).

These ops are implemented with two new primitives `DynamicSlice` and `DynamicSliceUpdate`. They are pretty similar to `Slice` and `SliceUpdate` but just different enough that it seemed cleaner to make new primitives rather than fill the `Slice` and `SliceUpdate` functions with conditions.


tests/ops_tests.cpp
```diff
@@ -353,6 +353,51 @@ TEST_CASE("test slice update") {
   CHECK(array_equal(slice(out, {4}, {8}, {1}), y).item<bool>());
 }
 
+TEST_CASE("test dynamic slice") {
+  auto src = reshape(arange(6), {2, 3});
+  ;
```
angeloskath, line 358: nit

tests/vmap_tests.cpp
```diff
@@ -496,3 +496,34 @@ TEST_CASE("test vmap SVD") {
...
+  }
+
+  {
+    auto fun = [](std::vector<array> inputs) {
+      return std::vector<array>{
+          slice_update(inputs[0], inputs[1], array({1}), {0})};
+    };
+    auto x = zeros({2, 2});
+    auto upd = ones({2, 1});
+    ;
```
angeloskath, line 521: same

angeloskath (approved):
Looks good.

I was wondering whether it makes sense to compute the offset in the kernel vs a separate tiny kernel launch 🤔

I like these as functions instead of `__getitem__` and `__setitem__`. I think it is possible but weird to have them as python slices. We may need to concat scalars or sth like that so functions is better for sure.

awni:
> I was wondering whether it makes sense to compute the offset in the kernel vs a separate tiny kernel launch 🤔

It's a good question.. I don't actually know the answer to that.. but I suspect it will depend a lot on how many dimensions you slice. The offset is ndim `int64` multiplications which as we know are relatively expensive. It seemed a bit wasteful to have every thread compute the same value so the alternative I was thinking about is to put it in shared memory for the thread group. But it was more tedious to implement so I didn't try it.

I think a good plan it to actually time the slice update dynamic vs not (e.g. with and without the offset computation) and see if it merits further optimization.

awni:
These two benchmarks are indistinguishable:

```
a = mx.zeros((2048, 2048))
upd = mx.random.uniform(shape=(1024, 1024))
starts = mx.array([1024, 1024])
axes = (0, 1)

def timeit(fn):
    for _ in range(5):
        mx.eval(fn())

    tic = time.time()
    for _ in range(100):
        mx.eval(fn())
    toc = time.time()
    ms = 10 * (toc - tic)
    print(f"{ms:.3f} (ms)")

def slice_update():
    x = a
    for _ in range(5):
        x[1024:, 1024:] = upd
    return x

def dynamic_slice_update():
    x = a
    for _ in range(5):
        x = mx.slice_update(x, upd, starts, axes)
    return x

timeit(slice_update)
timeit(dynamic_slice_update)
```

If I make the size really small (like 16x16) to get just the overhead then the dynamic slice update is about 4% slower on an M1 max.

There is also the the added challenge of dealing with dimension collapse. Given that we collapse the array shapes ahead of the copy we'd have to also encode the uncollapsed strides for both the input and output  and the uncollapsed ndim. It gets a bit ugly.

Overall I don't think it's worth optimizing yet. But let's revisit if needed.

PR #1742, opened 2025-01-01
Shapeless compile in docs and partially shapeless reshape

awni:
- Add a section on shapeless compilation in the compile usage guide
- Allow reshape to be shapeless when the inferred shape is valid


angeloskath (approved):
🙏

PR #1743, opened 2025-01-02
Transfer exceptions thrown in stream threads to main thread

zcbenz:
This PR catches exceptions when evaluating arrays in stream threads, and re-throw them in the main thread.

With this change errors happened in array evaluation can be caught in language bindings as exceptions, rather than hard crashes. Which is especially valuable for Windows, where the uncaught C++ exceptions are not printed in console and even a compilation failure requires attaching a debugger to find out the error message.

A few uncertain things with this PR:

* Only `std::runtime_error` is caught, catching other types of exceptions requires recording the type information of exception which I think does not worth the efforts.
* The tests are using `matmul` to throw errors, is there a reliable way to throw errors on all platforms?
* A new `failure` status is added to array, I'm not certain if there are side effects.

### DIFF
--- a/mlx/array.cpp
+++ b/mlx/array.cpp
@@ -113,6 +113,8 @@ void array::eval() {
   // Ensure the array is ready to be read
   if (status() == Status::unscheduled) {
     mlx::core::eval({*this});
+  } else if (status() == Status::failure) {
+    throw *exception();
   } else {
     wait();
   }
--- a/mlx/array.h
+++ b/mlx/array.h
@@ -358,6 +358,9 @@ class array {
     // status of `x` in `auto x = a + b; eval(x);`
     scheduled,
 
+    // Exception was thrown when evaluating the array.
+    failure,
+
     // The array's `eval_*` function has been run, but the computation is not
     // necessarily complete. The array will have memory allocated and if it is
     // not a tracer then it will be detached from the graph.
@@ -400,6 +403,14 @@ class array {
   }
   // Check if the array is a tracer array
   bool is_tracer() const;
+
+  // Set and get the exception thrown when evaluating the array.
+  void set_exception(const std::runtime_error& e) {
+    array_desc_->exception = std::make_unique<std::runtime_error>(e);
+  }
+  std::runtime_error* exception() const {
+    return array_desc_->exception.get();
+  }
 
   void set_data(allocator::Buffer buffer, Deleter d = allocator::free);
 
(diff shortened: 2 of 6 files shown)


awni:
So far our stance on this has been to throw as few possible exceptions from the stream threads as possible and try to push as much error handling to the main thread. I'm wondering if you could give more information on where this comes up for you / why you added this? 

Was it that there is no bf16 matmul on windows / linux and it crashed? Is it important to catch the exception in that case? I assume you will rewrite the program to not use the bf16 matmul rather than have an exception handler.

I'm a bit cautious about adding mostly because it's not clear at all that when you throw an exception in a stream thread, the program state is valid any longer. And it could be bad to try and continue running from there.

zcbenz:
My major complaint is `mx.compile`, compilation can fail on CI or user's machines and it is hard crash without an error message (on Windows). I think I can work around it by adding a `mx.test_compilation` API which compiles the preamble and throws in the main thread on failure.

Another thing helpful would be to explicitly catch and print the exception in stream threads, and then abort (with some work it is even possible to print the stack trace), which gives a consistent behavior for fatal exceptions on all platforms.

awni:
> My major complaint is mx.compile, compilation can fail on CI or user's machines and it is hard crash without an error message (on Windows).

I see, I didn't realize you don't get any kind of error message if it crashes on the stream thread. I can see how that makes debugging very difficult.

For a CPU-only build a pretty easy option is to have a flag maybe just for development to indicate if the task should run on the main thread or on the stream thread. Basically change [this line](https://github.com/ml-explore/mlx/blob/main/mlx/transforms.cpp#L237) to something like:

```c++
if (MLX_NO_THREADING) {
  task();
} else {
  scheduler::enqueue(stream, std::move(task));
}
```

It's not so scalable but could be handy for dev.

> Another thing helpful would be to explicitly catch and print the exception in stream threads, and then abort (with some work it is even possible to print the stack trace), which gives a consistent behavior for fatal exceptions on all platforms.

This would be really nice indeed. Printing useful error messages in the stream thread and aborting is preferable to propagating them to the main thread. I think it should be simpler as well. Stack trace would also be very nice.

zcbenz:
> For a CPU-only build a pretty easy option is to have a flag maybe just for development to indicate if the task should run on the main thread or on the stream thread.

It matters in production too, some user machines may have an old compiler not supporting C++17, or even no compilers at all. Apps (for example LM Studio) need a way to check if `mx.compile` works on the machine so they can give a meaningful error to users.

PR #1749, opened 2025-01-04
Optionally specify names for arrays when exporting

angeloskath:
This allows passing arrays as kwargs to `export_to_dot` in order to assign names to nodes in the graph. Example usage:

```python
model = ...
x = ...
loss, grads = nn.value_and_grad(...)(x)
weight_names = dict([
  (k.replace(".", "/"), v)
  for k, v in tree_flatten(model.parameters())
])
grad_names = dict([
  ("grad/" + k.replace(".", "/"), v)
  for k, v in tree_flatten(grads)
])
mx.export_to_dot("fwd_bwd.dot", loss, grads, loss=loss, **weight_names, **grad_names)
```

### DIFF
--- a/mlx/graph_utils.cpp
+++ b/mlx/graph_utils.cpp
@@ -117,7 +117,8 @@ void export_to_dot(
       [&](const array& x) {
         if (!x.has_primitive()) {
           input_set.insert(x.id());
-          os << "{ rank=source; " << namer.get_name(x) << "; }" << std::endl;
+          os << "{ rank=source; \"" << namer.get_name(x) << "\"; }"
+             << std::endl;
           return;
         }
 
@@ -131,7 +132,8 @@ void export_to_dot(
           os << "; }" << std::endl;
           // Arrows to primitive's inputs
           for (auto& a : x.inputs()) {
-            os << namer.get_name(a) << " -> " << x.primitive_id() << std::endl;
+            os << '"' << namer.get_name(a) << "\" -> " << x.primitive_id()
+               << std::endl;
           }
         }
 
@@ -141,10 +143,11 @@ void export_to_dot(
           if (output_set.find(a.id()) != output_set.end()) {
             os << "rank=sink; ";
           }
-          os << namer.get_name(a);
-          os << "; }" << std::endl;
+          os << '"' << namer.get_name(a);
+          os << "\"; }" << std::endl;
           if (x.has_primitive()) {
-            os << x.primitive_id() << " -> " << namer.get_name(a) << std::endl;
+            os << x.primitive_id() << " -> \"" << namer.get_name(a) << '"'
+               << std::endl;
           }
         }
       },
--- a/mlx/graph_utils.cpp
+++ b/mlx/graph_utils.cpp
@@ -105,18 +105,34 @@ void print_graph(
 void export_to_dot(
     std::ostream& os,
     NodeNamer namer,
-    const std::vector<array>& outputs) {
+    const std::vector<array>& nodes) {
+  // Perform one DFS to mark arrays as intermediate if they are used as inputs
+  // to other arrays.
+  std::unordered_set<std::uintptr_t> intermediate_set;
+  depth_first_traversal(
+      [&](const array& x) {
+        // No primitive so it is an input
+        if (!x.has_primitive()) {
+          return;
+        }
+
+        for (auto& a : x.inputs()) {
+          intermediate_set.insert(a.id());
+        }
+      },
+      nodes);
+
+  // Now we got everything we need to make the graph. Arrays can be one of 3
+  // things:
+  //  1. Inputs, when they have no primitive ie are evaluated
+  //  2. Intermediates, when they are the intermediate set
+  //  3. Outputs, if they are not inputs and not intermediates
+
   os << "digraph {" << std::endl;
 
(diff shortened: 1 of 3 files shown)


angeloskath:
Here is a full example of how the above would look.

```python
import mlx.core as mx
import mlx.nn as nn
from mlx.utils import tree_flatten


class SimpleMLP(nn.Module):
    def __init__(self):
        super().__init__()

        self.fc1 = nn.Linear(2, 32)
        self.fc2 = nn.Linear(32, 1)

    def __call__(self, x):
        x = nn.relu(self.fc1(x))
        x = nn.sigmoid(self.fc2(x))
        return x


def loss_fn(model, x, y):
    return (model(x) - y).square().mean()


model = SimpleMLP()
x = mx.random.normal((128, 2))
y = mx.random.uniform(shape=(128, 1))
mx.eval(model.parameters(), x, y)

loss_and_grad = nn.value_and_grad(model, loss_fn)
l, g = loss_and_grad(model, x, y)

weight_names = dict([
  (k.replace(".", "/"), v)
  for k, v in tree_flatten(model.parameters())
])
grad_names = dict([
  ("grad/" + k.replace(".", "/"), v)
  for k, v in tree_flatten(g)
])
mx.export_to_dot("graph.dot", l, g, loss=l, **weight_names, **grad_names)
```

and then running `dot -T png graph.dot >graph.png` results in

graph

awni:
That's really nice.. so much more interpretable!

I'm wondering if it makes sense to just specify the outputs one time rather than twice though:

```python
mx.export_to_dot("graph.dot", loss=l, **weight_names, **grad_names)
```

angeloskath:
Hmm I like your suggestion but I just wrote that and it changes the graph since we have configured outputs to be at the bottom and inputs at the top.

I can probably mark the outputs as the arrays that have no children but that is not necessarily the same. For instance you can have an output that is a function of another output and in that case it would be marked as an intermediate (which maybe it is).

Wdyt?

awni:
> we have configured outputs to be at the bottom and inputs at the top

Good point. I wonder if it would look ok if we removed the "rank=sink" and let dot place the outputs as it would. I think it will place them in a reasonable viewing location and with naming it will be a lot easier to see what is going on.

I also think it would work if you only mark outputs as arrays without parents.

I think I prefer on of the above to specifying the output twice since I could see it being confusing that this doesn't do anything: `mx.export_to_dot("graph.dot", output=output)` . That said I think you and I are the only people that use this function right now 🤣

angeloskath:
Agreed, I will try both and keep the one that looks better.

angeloskath:
Added a second DFS to automatically find the output set. Now we get exactly the same graph as above by specifying

```python
mx.export_to_dot("graph.dot", loss=l, **weights, **grads)
```

I think I like it a bit more than completely removing `source` and `sink`. I am adding this image here let me know wyt.

 
graph3

mlx/graph_utils.h
```diff
@@ -12,20 +12,50 @@ struct NodeNamer {
...
+      os, NodeNamer{}, std::vector<array>{std::forward<Arrays>(outputs)...});
+}
 
 template <typename... Arrays, typename = enable_for_arrays_t<Arrays...>>
-void export_to_dot(std::ostream& os, Arrays&&... outputs) {
-  export_to_dot(os, std::vector<array>{std::forward<Arrays>(outputs)...});
+inline void
+export_to_dot(std::ostream& os, NodeNamer namer, Arrays&&... outputs) {
+  export_to_dot(
+      os, namer, std::vector<array>{std::forward<Arrays>(outputs)...});
```
awni, line 58: nit:
```suggestion
      os, std::move(namer), std::vector<array>{std::forward<Arrays>(outputs)...});
```

mlx/graph_utils.h
```diff
@@ -12,20 +12,50 @@ struct NodeNamer {
...
+  print_graph(
+      os, NodeNamer{}, std::vector<array>{std::forward<Arrays>(outputs)...});
+}
 
 template <typename... Arrays, typename = enable_for_arrays_t<Arrays...>>
-void print_graph(std::ostream& os, Arrays&&... outputs) {
-  print_graph(os, std::vector<array>{std::forward<Arrays>(outputs)...});
+inline void
+print_graph(std::ostream& os, NodeNamer namer, Arrays&&... outputs) {
+  print_graph(os, namer, std::vector<array>{std::forward<Arrays>(outputs)...});
```
awni, line 36: Nit:
```suggestion
  print_graph(os, std::move(namer), std::vector<array>{std::forward<Arrays>(outputs)...});
```

mlx/graph_utils.cpp
```diff
@@ -96,20 +102,39 @@ void print_graph(std::ostream& os, const std::vector<array>& outputs) {
...
-  std::unordered_set<std::uintptr_t> output_set;
-  for (auto& o : outputs) {
-    output_set.insert(o.id());
-  }
-  std::unordered_set<std::uintptr_t> input_set;
-  NodeNamer namer;
   depth_first_traversal(
       [&](const array& x) {
         if (!x.has_primitive()) {
-          input_set.insert(x.id());
```
awni, line 111: I guess this was never used..

awni (approved):
Looks great, thanks!!

PR #1750, opened 2025-01-04
mpi send use input as output

awni:
1. Change send to return input as output to make it easier to eval the right graph
2. Fix bug in recv forgetting to end encoding

As for 1, I found it much simpler to implement pipeline parallelism if the `send` has that behavior. Otherwise one has to keep track of the output of send as an additional output to be evaluated which is quite cumbersome (it has to be wired through all the way to the `eval` call which can be far away from the call to `send`).

For example with this API you can just replace the `model.__call__` with the following simple change and everything else stays the same:

```python
       if cache is None:
            cache = [None] * len(self.layers)

        if rank < world_size - 1:
            h = mx.distributed.recv_like(h, (rank + 1))

        for layer, c in zip(self.layers, cache):
            h = layer(h, mask, c)

        # Send to the next process
        if rank != 0:
            h = mx.distributed.send(h, (rank - 1) % world_size)
```

I also think we don't loose anyhting here as no additional computation is being done and one can always treat the output of `send` as a seperate variable than the input to get the old behavior, basically:

```python
y = mx.distributed.send(x, dst)
```

### DIFF
--- a/mlx/backend/metal/distributed.cpp
+++ b/mlx/backend/metal/distributed.cpp
@@ -90,13 +90,14 @@ void Send::eval_gpu(
 
   auto& in = inputs[0];
   auto& out = outputs[0];
+  move_or_copy(in, out);
 
   // Schedule an async send on the comm stream
   auto task = [in = in, out = out, group = group(), dst = dst_]() mutable {
     if (in.event().valid()) {
       in.event().wait();
     }
-    distributed::detail::send(group, in, dst);
+    distributed::detail::send(group, out, dst);
     out.event().signal();
   };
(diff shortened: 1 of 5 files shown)


mlx/backend/metal/distributed.cpp
```diff
@@ -133,6 +135,7 @@ void Recv::eval_gpu(
   // Encode a wait event as there is no input for the recv to encode a signal.
   auto& s = stream();
   auto& d = metal::device(s.device);
+  d.end_encoding(s.index);
```
angeloskath, line 138: 🙏

angeloskath (approved):
Thanks looks great! Indeed strictly better than returning a dummy array.

PR #1752, opened 2025-01-06
Refactor distributed backend

angeloskath:
This PR allows for multiple distributed backends available at the same. For now it doesn't add or remove functionality but simply moves the distributed implementation in the `Group` instead of free standing functions in the `distributed` namespace.

The plan is for `distributed::is_available()` to check all available backends and maybe add a list available backends function. Additionally, `distributed::init()` shall take some extra parameters to allow initializing different backends.

I should run some long running training tests before we merge this but it should be at least good to review.


mlx/distributed/distributed_impl.h
```diff
@@ -6,6 +6,21 @@
 
 namespace mlx::core::distributed::detail {
 
+/**
+ * Abstract base class of a distributed group implementation.
+ */
+class GroupImpl {
+ public:
+  virtual int rank() = 0;
+  virtual int size() = 0;
```
awni, line 15: Nit: I would make these const here and in `group.h` to avoid having issues accessing them with a const group. E.g.

```
virtual int rank() const = 0
```

mlx/distributed/distributed.cpp
```diff
@@ -0,0 +1,100 @@
...
+Group init(bool strict /* = false */) {
+  static std::shared_ptr<detail::GroupImpl> default_group = nullptr;
+
+  if (default_group == nullptr) {
+    default_group = mpi::init(strict);
+  }
+
+  if (default_group == nullptr) {
+    default_group = std::make_shared<detail::EmptyGroup>();
+  }
```
awni, line 92: This is a very minor nit.. but you can put the static initialization in a little lambda to avoid needing to do these checks:

```suggestion
  auot init_group = [strict]() {
    auto default_group = mpi::init(strict);
    if (default_group == nullptr) {
      default_group = std::make_shared<detail::EmptyGroup>();
    }
    return default_group;
  };
  static std::shared_ptr<detail::GroupImpl> default_group = init_group();
```
```suggestion
  static std::shared_ptr<detail::GroupImpl> default_group = nullptr;

  if (default_group == nullptr) {
    default_group = mpi::init(strict);
  }

  if (default_group == nullptr) {
    default_group = std::make_shared<detail::EmptyGroup>();
  }
```
```suggestion
  static std::shared_ptr<detail::GroupImpl> default_group = nullptr;

  if (default_group == nullptr) {
    default_group = mpi::init(strict);
  }

  if (default_group == nullptr) {
    default_group = std::make_shared<detail::EmptyGroup>();
  }
```

mlx/distributed/mpi/mpi.cpp
```diff
@@ -233,149 +236,115 @@ MPIWrapper& mpi() {
...
+};
 
 bool is_available() {
   return mpi().is_available();
 }
 
-Group init(bool strict /* = false */) {
-  static std::shared_ptr<MPIGroupImpl> global_group = nullptr;
+std::shared_ptr<GroupImpl> init(bool strict /* = false */) {
+  static std::shared_ptr<MPIGroup> global_group = nullptr;
```
awni, line 334: Could also do the static initialization with a lambda here

awni: Although I wonder.. should this one need to be static? We only intend to call this once right?

awni (approved):
Looks great! Thanks for the refactor!

angeloskath:
Thanks for the comments, they 're awesome :-)

mlx/distributed/mpi/mpi.cpp
angeloskath: Yeah I am removing the static. But we should really only call this once if it succeeds. If it fails we can call it again.

mlx/distributed/distributed_impl.h
angeloskath: Hmm actually these can't be const. I made the group members const as they don't really affect the group. Even split should really return a different group without affecting the current group.

PR #1753, opened 2025-01-06
Allow boolean mask in sdpa

awni:
Allow a boolean (in addition to additive) mask in sdpa.

Nice for a couple reasons:
- Simplifies mask creation, we don't need to worry about the type, can just make a boolean mask
- Uses less memory for long context.

For the following:

```python
# Test with boolean causal mask
N = 8192
indices = mx.arange(N)
mask = get_mask(N)
x = mx.random.normal(shape=(1, 2, N, 32))
y = mx.fast.scaled_dot_product_attention(
    x, x, x, scale=1.0, mask=mask,
)
mx.eval(y)
print(mx.metal.get_peak_memory() / 1e6)
```

A boolean mask uses 610 MB
An additive mask uses 878 MB

### DIFF
--- a/mlx/backend/common/ternary.h
+++ b/mlx/backend/common/ternary.h
@@ -67,7 +67,12 @@ void set_ternary_op_output_data(
       }
       break;
     case TernaryOpType::General:
-      out.set_data(allocator::malloc_or_wait(out.nbytes()));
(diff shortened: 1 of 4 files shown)


angeloskath (approved):
Sweet!

PR #1754, opened 2025-01-07
Print exceptions in eval_cpu/eval_gpu and abort

zcbenz:
Catch exceptions when calling `eval_cpu`/`eval_gpu`, print them, and then abort.

Some platforms like Windows do not print uncaught exceptions in console, and this change gives them a chance to print out the error before termination.

### DIFF
--- a/mlx/backend/metal/metal.cpp
+++ b/mlx/backend/metal/metal.cpp
@@ -47,7 +47,11 @@ std::function<void()> make_task(array arr, bool signal) {
       }
 
       debug_set_primitive_buffer_label(command_buffer, arr.primitive());
-      arr.primitive().eval_gpu(arr.inputs(), outputs);
+      try {
+        arr.primitive().eval_gpu(arr.inputs(), outputs);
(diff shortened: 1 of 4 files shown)


awni:
This is very nice, thank you!

For now, I would just print the exception as is without changing the primitive throw error messages. 
- You changed a few but not all of them (e.g. in `accelerate/` and `metal/`) so the remaining ones will not be correct.
- In some cases the print is not just the primitive class name but includes some extra stuff so it will be a little odd looking.

Those `[name::method]` prefixes are mostly because we don't get nice stack traces. If/when we can print nicer stack traces it might be worth just removing them entirely. I see c++ 23 [has some support for this](https://en.cppreference.com/w/cpp/header/stacktrace). Maybe we can conditionally use it if available. I haven't seen a good lightweight alternative for c++ < 23 but I didn't look into it much.

zcbenz:
Got it. Have updated the PR.

awni (approved):
Thanks!!

PR #1756, opened 2025-01-08
Include command stdout in error message

zcbenz:
When compilation fails, users can see the reason in error message:

`Terminating due to uncaught exception: [Compile::eval_cpu] Failed to compile function Cf4BroadcastBDf4AddAC_VS_f4f4_11160318154034397263_contiguous: Failed to execute command with return code 127: "g++ -std=c++17 -O3 -Wall -fPIC -shared '/var/folders/4v/b5_8jfds0x71cym60rd_km1m0000gn/T/Cf4BroadcastBDf4AddAC_VS_f4f4_11160318154034397263_contiguous.cpp' -o '/var/folders/4v/b5_8jfds0x71cym60rd_km1m0000gn/T/libCf4BroadcastBDf4AddAC_VS_f4f4_11160318154034397263_contiguous.so' 2>&1", the output is: sh: g++: No such file or directory`

### DIFF
--- a/mlx/backend/common/compiled_cpu.cpp
+++ b/mlx/backend/common/compiled_cpu.cpp
@@ -7,6 +7,8 @@
 #include <mutex>
 #include <shared_mutex>
 
+#include <fmt/format.h>
+
 #include "mlx/backend/common/compiled.h"
 #include "mlx/backend/common/compiled_preamble.h"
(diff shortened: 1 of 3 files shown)


awni (approved):
LGTM, thanks !!

PR #1757, opened 2025-01-08
Add boolean mask support in vector SDPA

awni:
Add optional boolean mask in vector SDPA metal kernel:

```
Timing sdpa ... 3.01268 msec
Timing attention ... 7.13981 msec
Timing attention_mask ... 7.50264 msec
Timing sdpa_mask ... 2.24858 msec
```

Using the mask + sdpa is a lot faster than the naive primitive with a mask.

### DIFF
--- a/mlx/backend/metal/kernels/sdpa_vector.h
+++ b/mlx/backend/metal/kernels/sdpa_vector.h
@@ -4,6 +4,8 @@
 
 using namespace metal;
 
+constant bool has_mask [[function_constant(20)]];
+
 template <typename T, int D>
 [[kernel]] void sdpa_vector(
     const device T* queries [[buffer(0)]],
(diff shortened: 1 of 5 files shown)


mlx/backend/metal/kernels/sdpa_vector.h
```diff
@@ -54,34 +62,39 @@ template <typename T, int D>
...
 
-    // Compute the i-th score
-    U score = 0;
-    for (int i = 0; i < elem_per_thread; i++) {
-      score += q[i] * k[i];
-    }
-    score = simd_sum(score);
+      // Compute the i-th score
+      U score = 0;
+      for (int j = 0; j < elem_per_thread; j++) {
```
barronalex, line 73: +1 for removing the shadowing on i here

barronalex (approved):
That’s an awesome speed up!!
Super clean with the function constant too 🚀

PR #1758, opened 2025-01-09
Fix batched qmv bug

barronalex:
Resolves https://github.com/ml-explore/mlx-examples/issues/1182

This fixes a (B, M, K) x (B, N, K) `qmv`/`qvm` when B > 1 and M > 1.

This only occurs for small B < 6 since otherwise we use `qmm_t`/`qmm_n` which explains why it took a little while to pop up.

Previously we weren't adjusting the output stride by M, so we wouldn't write the entire output.

### DIFF
--- a/mlx/backend/metal/kernels/quantized.h
+++ b/mlx/backend/metal/kernels/quantized.h
@@ -1216,7 +1216,7 @@ METAL_FUNC void adjust_matrix_offsets(
     const device T*& scales,
     const device T*& biases,
     device T*& y,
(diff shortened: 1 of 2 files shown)


awni:
Thanks for the fix!! I'll let @angeloskath take a look and approve as I'm not so familiar with that kernel.

angeloskath (approved):
Great find! Thanks for the fix!

Feel free to merge after moving the fix to qmvs/qvms.

PR #1759, opened 2025-01-09
Simplify removes no-ops from the tape

awni:
It's a minor optimization but:

- reduces the tape size when no-ops are not fused
- exposes more possibility for merging arrays in simplify
- increases fusability given the depth limit


angeloskath (approved):
Nice!!!

angeloskath:
We could potentially remove `is_noop` from `is_fusable` now but I don't think it hurts to leave it there.

PR #1760, opened 2025-01-09
Use sdpa and exportable functions in transformer multi head attention

awni:
Use flatten/ unflatten to make the function exportable + use SDPA cause it will be faster / lower memory footprint in some cases.


angeloskath (approved):
👍

PR #1762, opened 2025-01-10
0ul is not size_t on MSVC

zcbenz:
Fix compilation error with MSVC due to args passed `std::max` have different types.


awni (approved):
Sorry about that 😅

PR #1764, opened 2025-01-10
Limit grad recursion depth by not recursing through non-grad inputs

awni:
Previously we enclose non-gradient inputs in a lambda which then get recursed through when we build the VJP graph.

This change to allow an internal VJP to take a `argnums` list so that the VJP can have access to all the inputs including the ones for which no gradient is requested.

- Also fixes a bug in python grad where we return the wrong result when given multiple argnums of the same value.

An example case which would previously recurse to 10k+ and now just recurses a couple of times:

```python
fun = lambda x, y: x * y
x = mx.array(2.0)
for _ in range(10000):
    x = mx.abs(x)
y = mx.array(3.0)
dfdx = mx.grad(fun)(x, y)
```

### DIFF
--- a/mlx/transforms.cpp
+++ b/mlx/transforms.cpp
@@ -278,7 +278,8 @@ void eval(std::vector<array> outputs) {
 std::pair<std::vector<array>, std::vector<array>> vjp(
     const std::function<std::vector<array>(const std::vector<array>&)>& fun,
     const std::vector<array>& primals,
-    const std::vector<array>& cotans) {
+    const std::vector<array>& cotans,
+    const std::vector<int>& argnums) {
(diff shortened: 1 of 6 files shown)


awni:
@davidkoski this limits the stack-overflow issue with VJP that we were discussing offline. It does require a change in how VJP is called from Python -> C++. I'm not sure how it's done in Swift (if you enclosed the non-grad inputs or not). But it may also require a change there to pass in the non-grad inputs to `value_and_grad` rather than enclosing them.

angeloskath (approved):
A bit of mind-bender but it makes perfect sense afterwards. Nice!

PR #1765, opened 2025-01-11
Fix and cleanup event signal/wait for metal

awni:
Events should be held in completion handler so we know they are not destroyed too early. Previously this would crash with internal error:

```python
a = mx.ones((5, 5))
for _ in range(10):
    a = mx.distributed.all_sum(a)
mx.eval(a)

tic = time.perf_counter()
for _ in range(10):
    a = mx.distributed.all_sum(a)
mx.eval(a)
```


angeloskath (approved):
👍

PR #1768, opened 2025-01-14
Fix synchronization bug for GPU stream async CPU work

awni:
Fixes synchronization bugs for asynchronous CPU work that happens in the GPU stream.

The basic premise is that the output events were getting signaled before all the work in the GPU stream actually completed.

For example:

```python
a = mx.ones(5,) # a event value 1
x = mx.load() # x has event value 2
z = mx.all_sum(a) # wait on event value 1
```

The `load` would signal the event value to `2` without ensuring that `a` was actually finished causing the `all_sum` to start prematurely.

The fix is to simply make a new event for primitives like `Load` and the distributed ops that need to synchronize with the GPU.

### DIFF
--- a/mlx/backend/metal/distributed.cpp
+++ b/mlx/backend/metal/distributed.cpp
@@ -12,11 +12,11 @@
 
 namespace mlx::core::distributed {
 
-void signal_and_wait(const array& in, const array& out) {
-  if (in.event().valid()) {
-    encode_signal(in.event());
+void signal_and_wait(const Event& e_signal, const Event& e_wait) {
+  if (e_signal.valid()) {
+    encode_signal(e_signal);
   }
-  encode_wait(out.event());
+  encode_wait(e_wait);
 }
 
 void AllReduce::eval_gpu(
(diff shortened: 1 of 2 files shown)


mlx/backend/metal/distributed.cpp
```diff
@@ -92,15 +98,14 @@ void Send::eval_gpu(
...
   };
   scheduler::enqueue(detail::communication_stream(), std::move(task));
 
-  // Encode a signal event for the input but not a wait since we don't need to
-  // wait on the output.
+  // Encode a signal event for the input
   if (in.event().valid()) {
     encode_signal(in.event());
   }
+  // TODO do we need an event wait as well?
```
angeloskath, line 108: That is pretty tricky. Many comm backends have a synchronous/blocking and non-blocking version but I feel that this may be a bit much for now. So I think going without an event is better as it will allow us to better pipeline multiple sends or computation after send in general.

angeloskath (approved):
Perfect! That would have gone unnoticed for long...

PR #1772, opened 2025-01-17
Matmul JVPs

### LINKED ISSUE #1771: "Not implemented" error with mlx.core.jvp
I'm running into the following error -- "ValueError: [Primitive::jvp] Not implemented for AddMM" -- when using JVPs. This error does not seem to occur with VJPs. Here is simple example:

```
import mlx.core as mx
import mlx.nn as nn

class Net(nn.Module):

    def __init__(self, in_dim: int):
        super().__init__()
        self.layer = nn.Linear(in_dim, in_dim)

    def __call__(self, x: mx.array) -> mx.array:
        return self.layer(x) 

d = 2
model = Net(d)
x = mx.random.normal((d,))
v = mx.random.normal((d,))
```

Now, if I perform a VJP, things work without error:

```
mx.vjp(model, [x], [v])
```

Alternatively, if I perform a JVP I get the above error:

```
mx.jvp(model, [x], [v])
```

```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Cell In[12], [line 1](vscode-notebook-cell:?execution_count=12&line=1)
----> [1](vscode-notebook-cell:?execution_count=12&line=1) mx.jvp(model, [x], [v])

ValueError: [Primitive::jvp] Not implemented for AddMM.
```

Assuming this is not an issue with my implementation, is there a fix for this type of error?

Thanks!

Edit: I've posted this in the wrong place, can this be moved from ml-explore/mlx-examples to ml-explore/mlx?

awni:
Added JVP for:
- Matmul
- AddMM
- QuantizedMatmul

I also shrunk some of the quantized tests since they were taking several minutes to complete on the CPU and we have the long running tests for those now.

Closes #1771

### DIFF
--- a/mlx/primitives.cpp
+++ b/mlx/primitives.cpp
@@ -246,6 +246,36 @@ std::vector<array> AddMM::vjp(
   return vjps;
 }
 
+std::vector<array> AddMM::jvp(
+    const std::vector<array>& primals,
+    const std::vector<array>& tangents,
+    const std::vector<int>& argnums) {
+  std::vector<array> jvp;
+  for (int i = 0; i < argnums.size(); ++i) {
+    auto arg = argnums[i];
+    if (arg == 0) {
+      if (jvp.empty()) {
+        jvp.push_back(matmul(tangents[i], primals[1], stream()));
+      } else {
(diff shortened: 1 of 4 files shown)


barronalex (approved):
Looks great! I played around with the tests and everything still passed :)

Good call on making the quantized tests faster.

PR #1773, opened 2025-01-17
Faster synchronization `Fence` primitive

awni:
- Adds a `Fence` synchronization primitive for polling-based fast CPU / GPU synchronization
- Conditionally builds the above for Metal 3.2 and up.
- Adds an environment variable `MLX_METAL_FAST_SYNCH` to toggle using the fast primitive or not (default `0`).
- Updates distributed Metal ops to use the `Fence` so we have the option to make them fast.

- Distributed LLM tested on M1 Max, M2 Ultra, M3 Max and working.
- Benchmarks suggests 10x reduction in synchronization latency for all-reduce (15-20 microseconds depending on the machine, compared to 150-200 with shared events).

Microbenchmarks on M3 Max:

```
mpirun -np 2 python benchmarks/python/synchronize_bench.py
```

All Reduce: time per iteration 0.149781 (ms)
All gather: time per iteration 0.136384 (ms)

```
MLX_METAL_FAST_SYNCH=1 mpirun -np 2 python benchmarks/python/synchronize_bench.py
```
All Reduce: time per iteration 0.016508 (ms)
All gather: time per iteration 0.015927 (ms)


mlx/backend/metal/device.cpp
```diff
@@ -338,6 +354,7 @@ CommandEncoder& Device::get_command_encoder(int index) {
   if (stream.encoder == nullptr) {
     stream.encoder = std::make_unique<CommandEncoder>(stream.buffer);
     stream.fence = std::make_shared<Fence>(device_->newFence());
+    stream.encoder->wait_for_fence(stream.event_fence);
```
angeloskath, line 357: I am kinda confused by this wait here. In the case where we don't use the `Fence` at all when would the `event_fence` be updated?

awni: Ok, I pushed a change to get rid of this that should (and seems) to work, so you can disregard all the previous stuff with `event_fence`.

Basically it requires modifying the call to `wait_gpu` to take an array that we want to be sure is ready before any future kernels that depend on it run. It reuses our existing synchronization machinery (barriers + fences) and is nice in that it only encoders which actually depend on the output will wait for it.

I had to add a way to `register_output_array` since it's not actually part of a kernel.. but I think it's cleaner/ more efficient / and doesn't require this random `stream_event` which was very icky.

angeloskath: > So if you never update it the wait it is essentially a no-op.

What a horrible hidden API. I assumed wait without update is a deadlock hence the confusion before.

> I had to add a way to `register_output_array`.

Yeah, so much better! Also only waiting on things that matter rather than everything on the whole stream. Plus avoid waiting on two fences which could have been the case before.

angeloskath (approved):
Absolutely beautiful!

mlx/backend/metal/kernels/fence.metal
```diff
@@ -0,0 +1,52 @@
...
+#define __METAL_MEMORY_SCOPE_SYSTEM__ 3
+namespace metal {
+constexpr constant metal::thread_scope thread_scope_system =
+    static_cast<thread_scope>(__METAL_MEMORY_SCOPE_SYSTEM__);
+}
+#endif
+
+#include <metal_atomic>
+
+[[kernel]] void input_coherent(
```
zcbenz, line 15: What does the `input_coherent` do? Does it mean to ensure the data gets fully transferred from cpu to gpu?

awni: It's to ensure the data is safe to be read by the CPU.

zcbenz: Is this a metal specific thing?

I have a weird flaky test in cuda backend that an array loaded from cpu sometimes ends up with wrong value in gpu:
https://github.com/ml-explore/mlx/blob/9daa6b003f548569e9186502d5acb9b74b91bcbe/tests/load_tests.cpp#L36

I wonder if I need to do something similar here, though it is completely different hardware.

awni: It's very metal specific. In fact.. most of the time this kerne is not used. It's only enabled for an experimental polling based synchronization when you set `MLX_FAST_METAL_SYNCH=1`. 

If you are using standard cuda synchronization primitives you should need anything like that.

zcbenz: Cuda does not have a standard synchronization primitive for signaling a gpu wait from cpu, I was using `cuda::atomic` but its `wait/notify` APIs had some problems with CUDA 11 (https://github.com/ml-explore/mlx/issues/2137) so I'm currently using busy wait for fence which is almost identical to metal backend's fast fence.

awni: > so I'm currently using busy wait for fence which is almost identical to metal backend's fast fence.

Oh right. Well in that case.. we are in pretty uncharted territory. Do you know of examples that have used a similar concept in CUDA? That would be a good place to look and see if they have to do anything fancy to ensure the input safe to read. 

In some Metal we need this because even if the fence value is written by the GPU kernel (which certainly implies and the input is done being computed) it can still be in cache and not necessarily "coherent" with the rest of the device (and system in unified memory).

zcbenz: No I couldn't find any open source code synchronizes gpu with cpu work, and there is actually no framework using cuda's unified memory APIs. I also searched nvidia forum and there was probably 2 posts related to this topic in past 10 years.

PR #1776, opened 2025-01-18
Recompile when shapeless is different

awni:
Since the graph can change if `shapeless` changes from `False` to `True` we need to compile different functions.


angeloskath (approved):
👍

PR #1782, opened 2025-01-21
Use `get_template_definition` instead of `fmt::format`

angeloskath:
This is basically a cherry-pick from the `cpp20` branch. It just simplifies building some kernels via jit.

Regarding C++ 20, I don't think it is worth moving to C++ 20 yet, it is too much hassle for gcc at the moment (especially if we want to get rid of `fmt` which requires gcc >= 13). For Macs OSX minimum deployment target works fine but also comes with requirement to build on Xcode 16 so I think we should revisit in a few months or if we have another need for it.


awni:
Sounds good to me! Thanks for digging into that and for this clean up.

awni (approved):

PR #1784, opened 2025-01-21
Ring distributed backend

angeloskath:
Initial version of a custom ring reduce backend. It can be used with Ethernet as well but the main focus is on thunderbolt because:
1. MPI ring reduce is not a very good fit and doesn't saturate the TB interface.
2. it makes it much easier to add more cables for even more bandwidth

The comparison is as follows for a 4-way all reduce using 4 M2 Ultras:
all_reduce

The bandwidth is computed as `total bytes * 3/4 * 2 / time` the `3/4` is due to the ring reduce actually sending less than 2*B bytes on the wire.

In terms of time per all reduce
all_reduce_time

which means that for 2GB all reduce we go from 1.3 seconds over Ethernet and 0.45s with MPI to 0.27s with our reduce.

Things to note:
1. `mx.distributed.init()` takes a `backend` argument and initializes the corresponding backend. If no argument is provided it tries to initialize in order `ring > mpi`. The reason is that ring will fail if `MLX_HOSTFILE` and `MLX_RANK` is not provided while mpi may succeed and "hide" the rest of the backends.
2. `mx.distributed.init()` called for a 2nd time without an argument returns the previously initialized backend. This allows the rest of the code to remain unchanged and agnostic of the communication group.
3. Small edits to the Threadpool to make it resizable.
4. The ring backend can be improved/simplified so take this as a 1st version mostly defining the API.

Finally, I am in the process of preparing another PR with changes in the docs and possibly a launcher for this backend so people don't have to write their own shell scripts to launch with this backend.


mlx/io/threadpool.h
```diff
@@ -38,9 +38,13 @@ class ThreadPool {
   template <class F, class... Args>
   auto enqueue(F&& f, Args&&... args)
       -> std::future<typename std::invoke_result_t<F, Args...>>;
+  void resize(size_t);
```
awni, line 41: Maybe it makes sense to move this file up a level into `mlx/`

mlx/distributed/distributed.cpp
```diff
@@ -80,20 +83,47 @@ Group Group::split(int color, int key /* = -1 */) const {
...
-  };
-  static std::shared_ptr<detail::GroupImpl> default_group = init_group();
+Group init(bool strict /* = false */, Backend bk /* = Backend::Any */) {
+  static std::unordered_map<Backend, std::shared_ptr<detail::GroupImpl>>
+      backends;
+
+  // Already initialized so return the group.
+  if (auto g = backends.find(bk); g != backends.end()) {
+    return Group(g->second);
+  }
```
awni, line 93: Does it ever make sense to have two back-ends simultaneously active? If not .. maybe this should throw in that case?

awni (commented):
Really nice, I didn't read the actual ring.cpp yet but I think the API changes look great.

One high-level question, is there anyway to test this back-end in CI?

angeloskath:
Yeah 100%. I should have added that actually. Will do.

It doesn't actually need thunderbolt connections any socket will do. I regularly test for correctness using the loop back interface which will be the CI test as well.

awni (approved):
LGTM!! This is good to go right?

angeloskath:
Yeah it is good to go. I am writing docs for it in a different PR and there will be more PRs improving the performance and adding features (these should be limited in `ring.cpp`). I want to apply your comments and then we can merge.

mlx/distributed/distributed.cpp
angeloskath: I think it does make sense to have more than one backend active at a given time. For instance we could have MPI over Ethernet and a thunderbolt ring. It would be slightly weird code but we can expect MPI to always have more features than the ring backend.

PR #1788, opened 2025-01-23
Einsum ellipsis

### LINKED ISSUE #1717: [Feature] Ellipsis for broadcasting with einsum
I noticed that using mlx.core.einsum with ellipsis produces an error. In numpy, I've found that ellipses with np.einsum can be quite useful. Are there plans of including this feature in the future? 

**To reproduce**
MLX code:
```
import mlx.core as mx

x = mx.random.normal(shape=(3, 4, 5, 6))
y = mx.random.normal(shape=(3, 4, 5, 6))

z = mx.einsum('...ij,...kj->...ik', x, y)
```
and the resulting error: 
```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Cell In[4], [line 7]
      [4] x = mx.random.normal(shape=(3, 4, 5, 6))
      [5] y = mx.random.normal(shape=(3, 4, 5, 6))
----> [7] z = mx.einsum('...ij,...kj->...ik', x, y)

ValueError: [einsum] Subscripts must be letters, but got '.'.
```

In contrast, the following code works without issue in numpy: 
```
import numpy as np

x = np.random.normal(size=(3, 4, 5, 6))
y = np.random.normal(size=(3, 4, 5, 6))

z = np.einsum('...ij,...kj->...ik', x, y)
```

Thanks!

angeloskath:
Closes #1717.

I added the tests I could think of but maybe I am missing an important case? Also this adds lines 653 to 665 which will run every time even if we don't have an ellipsis. I could instead move them into 701 and run them the first time to encounter an ellipsis. This way the overhead of parsing the ellipsis should be even less. Let me know if you prefer that.

### DIFF
--- a/mlx/einsum.cpp
+++ b/mlx/einsum.cpp
@@ -669,7 +669,7 @@ std::pair<std::vector<PathNode>, PathInfo> einsum_path_helper(
     bool have_ellipsis = false;
     int cnt_before = 0, cnt_after = 0;
     for (int i = 0; i < subscript.size(); i++) {
-      if (!isalpha(subscripts[i])) {
+      if (!isalpha(subscript[i])) {
         if (i + 2 >= subscript.size() || subscript[i] != '.' ||
             subscript[i + 1] != '.' || subscript[i + 2] != '.') {
           std::ostringstream msg;
--- a/mlx/einsum.cpp
+++ b/mlx/einsum.cpp
@@ -71,6 +71,7 @@ std::pair<std::vector<std::string>, std::string> parse(std::string subscripts) {
   } else {
     // Implicit mode:
     // - repeats are summed
+    // - ellipses are placed in the beginning of the output
     // - remaining output axes are ordered alphabetically
     lhs = subscripts;
     std::unordered_map<char, int> temp;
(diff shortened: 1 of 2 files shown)


barronalex (commented):
Looks great!

I converted some of [NumPy's tests](https://github.com/numpy/numpy/blob/2e0d1f133a115374bf9b964c41972c2d02410db8/numpy/_core/tests/test_einsum.py#L786C1-L826C78) to double check and a couple are failing:

```
        A = mx.random.normal(shape=(4, 3))
        B = mx.random.normal(shape=(3, 2))
        ref = mx.einsum('ik,kj->ij', A, B)
        self.assertTrue(mx.allclose(mx.einsum('...k,kj', A, B), ref))  # used to raise error

        a = mx.random.normal(shape=(2, 3, 4, 5))
        v = mx.random.normal(shape=(4,))
        ref = mx.einsum('ijkl,k->ijl', a, v)
        self.assertTrue(mx.allclose(mx.einsum('...kl,k', a, v), ref))
        self.assertTrue(mx.allclose(mx.einsum('...kl,k...', a, v), ref))
 ```
 
 Possibly something to do with only one operand having an ellipsis?

angeloskath:
Awesome! Thanks for both comments. I did a small search for tests but obviously I failed 😕

I will copy several from numpy's tests and fix the error. The string abomination was aiming for minimum allocations but I do think that a) premature optimization the root of all evil etc b) the two iterator version of replace should do exactly 1 allocation as well.

angeloskath:
So the bug had to do with the fact that I didn't handle the implicit contraction case at all (who uses that btw?).

It should be fine now but I will do another pass to add more of numpy's tests and then we can merge.

barronalex (approved):
Awesome!

Issue #1794, opened 2025-01-25
[BUG] LAPACK `info` is not negative


NAThompson:
**Describe the bug**

When a non-positive semidefinite matrix is passed into the Cholesky decomposition, the `info` is 2, not a negative number.

**To Reproduce**

Apply the following patchfile:

```diff
diff --git a/mlx/backend/common/cholesky.cpp b/mlx/backend/common/cholesky.cpp
index 62807e6..3f4beeb 100644
--- a/mlx/backend/common/cholesky.cpp
+++ b/mlx/backend/common/cholesky.cpp
@@ -5,6 +5,7 @@
 #include "mlx/backend/common/lapack.h"
 #include "mlx/linalg.h"
 #include "mlx/primitives.h"
+#include <iostream>

 namespace mlx::core {

@@ -51,6 +52,8 @@ void cholesky_impl(const array& a, array& factor, bool upper) {
       throw std::runtime_error(msg.str());
     }

+    std::cout << "Cholesky info = " << info << std::endl;
+
     // Zero out the upper/lower triangle while advancing the pointer to the
     // next matrix at the same time.
     for (int row = 0; row < N; row++) {
```

And run this code:

```python
#!/usr/bin/env python3

import mlx
import mlx.core

A = mlx.core.array([[1.0, 2.0], [2.0, 1.0]])
print(f"Input A = {A}")
L = mlx.core.linalg.cholesky(A, stream=mlx.core.cpu)
print(f"L={L}")

A_recovered = L@L.T
print(f"L@L^T = {A_recovered}")
```

Info = 2 will be printed.

**Expected behavior**

The comment `mlx/backend/common/cholesky.cpp`  in `cholesky_impl` (confusingly) claim it does not and should not throw, but then it explicitly throws on `info < 0`:

```cpp
    // TODO: We do nothing when the matrix is not positive semi-definite
    // because throwing an error would result in a crash. If we figure out how
    // to catch errors from the implementation we should throw.
    if (info < 0) {
      std::stringstream msg;
      msg << "[cholesky] Cholesky decomposition failed with error code "
          << info;
      throw std::runtime_error(msg.str());
    }
```

However, this condition is never satisfied, as error states are communicated with positive integers. Hence the desired fix is not quite obvious-should the error check condition be fixed and the throw be maintained?

angeloskath:
[...]

PR #1795, opened 2025-01-25
Translate LAPACK exception codes to an actionable error message.

NAThompson:
## Checklist

### DIFF
--- a/mlx/backend/common/cholesky.cpp
+++ b/mlx/backend/common/cholesky.cpp
@@ -41,13 +41,17 @@ void cholesky_impl(const array& a, array& factor, bool upper) {
         /* lda = */ &N,
         /* info = */ &info);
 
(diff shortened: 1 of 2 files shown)


angeloskath:
As I mentioned in #1794 (and as the comment above the check mentions) I think we should not throw when a matrix is not positive semi-definite since this can't be caught and will result in a crash.

awni:
Agreed, we shouldn't throw for non-PSD matrices. Maybe there are a couple other options we can think about (though not necessarily for this PR):

- Have a hard assert mode, when you run you code and a detectable error happens (like this or in indexing reading/writing out of bounds) there is a hard crash. Could toggle that with a env flag
- Could fill the output with NaN in the case of non-PSD matrices rather than a partially computed Q/R.

awni:
@NAThompson do you want to come back to this or should we close it for now?

PR #1797, opened 2025-01-26
Fix rope fallback to not upcast

awni:
As title.

### DIFF
--- a/mlx/fast.cpp
+++ b/mlx/fast.cpp
@@ -392,7 +392,7 @@ array rope(
           s);
     };
 
-    auto inv_freqs = inputs.size() == 3 ? astype(reciprocal(inputs[2], s), t)
+    auto inv_freqs = inputs.size() == 3 ? astype(reciprocal(inputs[2], s), t, s)
                                         : default_inv_freqs();
     auto theta =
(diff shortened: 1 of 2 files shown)


mlx/fast.cpp
```diff
@@ -392,8 +392,8 @@ array rope(
           s);
     };
 
-    auto inv_freqs =
-        inputs.size() == 3 ? reciprocal(inputs[2], s) : default_inv_freqs();
+    auto inv_freqs = inputs.size() == 3 ? astype(reciprocal(inputs[2], s), t)
```
angeloskath, line 395: So presumably this is for precision that we don't do `reciprocal(astype(inputs[2], t, s), s)` ?

Also small issue here is that `astype` should also get the stream.

```suggestion
    auto inv_freqs = inputs.size() == 3 ? astype(reciprocal(inputs[2], s), t, s)
```

angeloskath (approved):
Nice!

mlx/fast.cpp
awni: Yes it seemed safer but it's not really based on anything concrete. Good catch with the stream!

PR #1798, opened 2025-01-27
CPU benchmarks for Cholesky decomposition.

NAThompson:
## Checklist


awni:
Thanks for the PR. I don't think we are particularly in need of a standalone benchmark for this.. if you are planning to do some work on speeding up Cholesky feel free to submit a benchmark with it at that time.

awni:
Cool, per offline conversation, I'll keep this closed but we can re-open/merge when necessary

PR #1811, opened 2025-01-30
Allow different value dimensions in sdpa_vector

angeloskath:
As title. Specifically it adds a kernel for 192 query/key dimensions and 128 value dimensions.

On the M2 Ultra the speed ups are as follows (over non fused sdpa):

**With GQA**
speedup_gqa

**Without GQA**
speedup


awni:
Wow, the GQA speedup is massive!!

angeloskath:
> Wow, the GQA speedup is massive!!

There isn't an improvement over what we had before (as you can see in the code).

PR #1812, opened 2025-01-30
Remove JSON Dependency

barronalex:
A basic JSON parser that allows us to remove the [nlohmann/json](https://github.com/nlohmann/json) dependency.

Should be pretty much complete with the exception of unicode support.

### DIFF
--- a/mlx/io/load.h
+++ b/mlx/io/load.h
@@ -274,7 +274,6 @@ struct json {
   }
 
   json& operator[](const char* key) {
-    std::cout << "KEY " << key << std::endl;
     auto& m = std::get<json_object>(this->data);
     if (m.find(key) == m.end()) {
       m.insert({key, nullptr});
@@ -343,9 +342,9 @@ struct json {
       data;
 };
 
-json parse_json(std::istream& s, bool allow_extra = false);
+json parse_json(std::istream& s);
 
-json parse_json(const std::string& s, bool allow_extra = false);
+json parse_json(const std::string& s);
 
 } // namespace io
 } // namespace mlx::core
--- a/mlx/io/load.h
+++ b/mlx/io/load.h
@@ -4,6 +4,7 @@
 
 #include <memory>
 #include <sstream>
+#include <variant>
 
 #include <fcntl.h>
 #ifdef _MSC_VER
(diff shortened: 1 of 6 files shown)


mlx/io/load.h
```diff
@@ -170,5 +171,180 @@ class FileWriter : public Writer {
...
+using json_object = std::unordered_map<std::string, json>;
+
+struct json {
+  json() : data(json_object()) {}
+
+  template <typename T>
+  json(T x) : data(x) {}
+
+  // TODO: templatize this
+  json(std::vector<size_t> x) {
```
awni, line 186: Why not templatize these vector functions? Should be fairly straight-forward to remove the duplicates below?

```
template <typename T>
json(std::vector<T> x) {
...
```

awni: Oh I see, you have `long` for both `int` and `long` and `double` for bout `double` and `float` 🤔

mlx/io/load.h
```diff
@@ -4,6 +4,7 @@
 
 #include <memory>
 #include <sstream>
+#include <variant>
```
awni, line 7: Not critical (maybe overkill).. but I think it would make sense to reorganize:
- move the json declarations into mlx/io/json.h (and include this for json only)
- move the json definitions into mlx/io/json.cpp

mlx/io/safetensors.cpp
```diff
@@ -120,34 +117,37 @@ SafetensorsLoad load_safetensors(
...
+        metadata_map.insert({meta_key, meta_value});
       }
       continue;
     }
-    const std::string& dtype = item.value().at("dtype");
-    const Shape& shape = item.value().at("shape");
-    const std::vector<size_t>& data_offsets = item.value().at("data_offsets");
+    const std::string& dtype = item_value["dtype"];
+    // TODO: revert this to Shape
+    const std::vector<int>& shape = item_value["shape"];
```
awni, line 141: That's interesting, `Shape` is just an alias for `vector<int>`. It doesn't compile if you use it there?

awni (approved):
Looks great, thanks for the update! Left a couple minor comments.. but otherwise g2g!

mlx/io/load.h
barronalex: Sounds great!

mlx/io/load.h
barronalex: Totally agree -- for reasons I don't understand the compiler does not like it when I templatize here. I'll play around with it a little.

mlx/io/safetensors.cpp
barronalex: This was a red herring in the end. Shape works fine.

mlx/io/json.cpp
```diff
@@ -0,0 +1,281 @@
...
+  return num;
+}
+
+json parse_json_number(std::istream& s) {
+  auto num = read_digits(s);
+  if (num.find_first_of(".eE") != std::string::npos) {
+    return json(std::stod(num));
+  } else {
+    return json(std::stol(num));
+  }
```
angeloskath, line 28:
```suggestion
  size_t pos;
  json ret;
  if (num.find_first_of(".eE") != std::string::npos) {
    ret = json(std::stod(num, &pos));
  } else {
    ret = json(std::stol(num, &pos));
  }
  if (pos != num.size()) {
    throw std::invalid_argument(...);
  }
  return std::move(ret);
```

The above is a bit of a problem because `read_digits` reads any character in `[0-9-eE.]` which means it could be not a number eg `1.2.3.4` and in this case we should fail. `std::stod` and friends return the number of characters processed so it is a fairly easy change to check for correctness.

angeloskath (commented):
This is a very nice addition. Very well done.

I left a few comments on things I missed the first time around. They mostly have to do with edge case error checking for invalid JSON so not absolutely critical but it 'd be nice to fix before merging.

awni:
```
cmake .. -DMLX_BUILD_CPU=ON -DMLX_BUILD_SAFETENSORS=OFF
```

is currently broken... but rather than fix it, I think we can just land this.

awni:
@barronalex we can close this right?

PR #1817, opened 2025-02-01
Refactor common into cpu specific and truly common

awni:
I think this is a nice change. Easier to understand what is what and what goes where, simplifies includes between different backends and simplifies the cpu / no_cpu build.

The new structure is like so:

```
backend/
  common/
  cpu/
  no_cpu/
  metal/
  no_metal/
```

- `eval_cpu` always goes in `cpu`, `eval_gpu` always goes in metal
- Primitives with a common implementation can have a private method `eval` or use a free function which is implemented in common
- Other utilities used by both back-ends also go in common (e.g. binary/ternary/copy types, reduction planning, compile helpers, etc).


angeloskath (approved):
This looks great!

I did go through it but admittedly only skimmed most of it since it is mostly moving stuff around. If there is something specific you want me to look closely let me know.

awni:
No it's pretty much just moving stuff around. I don't think I wrote any new code. The most invasive change was splitting a couple of files up (binary.h, reduce.h) so that the common part lived in common and the CPU part lived in cpu/.

PR #1819, opened 2025-02-01
Added mlx::core::version() returning std::string(MLX_VERSION)

stemann:
## Proposed changes

Added function for getting the MLX version built.

## Checklist

### DIFF
--- a/mlx/version.h
+++ b/mlx/version.h
@@ -1,0 +1,24 @@
+// Copyright © 2023-2024 Apple Inc.
+
+#pragma once
+
+#include <string>
+
+#define STRINGIFY(x) #x
+#define TOSTRING(x) STRINGIFY(x)
+
+namespace mlx::core {
+
+std::string version() {
+  return std::string(TOSTRING(MLX_VERSION));
+}
+
+int version_major() { return MLX_VERSION_MAJOR; }
+
+int version_minor() { return MLX_VERSION_MINOR; }
+
+int version_patch() { return MLX_VERSION_PATCH; }
+
+int version_numeric() { return MLX_VERSION_NUMERIC; }
+
+} // namespace mlx::core
--- a/mlx/version.h
+++ b/mlx/version.h
@@ -9,8 +9,8 @@
 
 namespace mlx::core {
 
-std::string version() {
-  return std::string(TOSTRING(MLX_VERSION));
+constexpr const char* version() {
+  return TOSTRING(MLX_VERSION);
 }
 
 constexpr int version_major() {
--- a/mlx/version.h
+++ b/mlx/version.h
@@ -13,12 +13,20 @@ std::string version() {
   return std::string(TOSTRING(MLX_VERSION));
 }
 
-int version_major() { return MLX_VERSION_MAJOR; }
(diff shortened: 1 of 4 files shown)


awni:
Just curious, what do you use this for? 

I'm not opposed to making the version available in the C++ API. I'm just wondering if we should make it a `constexpr` or a preprocessor macro so that it can be used at compile time as well. Either way it might be good to change the format to be a single number like `MAJORMINORPATCH` rather than `MAJOR.MINOR.PATCH` so it can easily be used in conditionals.

stemann:
It’s mostly for convenience, e.g. for a user of MLX.jl (Julia),

* where the user is using MLX.jl version X
* dependent on MLX C version Y,
* where MLX.jl has an indirect dependency via MLX C version Y on MLX version Z,

that the user can more easily check the version (Z - and Y would also be nice) - or write version dependent code (if that should be necessary) - without importing the Julia package manager (Pkg.jl) as a dependency.

stemann:
Regarding the format, I was actually considering to do it like LibTorch, with separate identifiers for major, minor, and patch - in addition to the string version (`"MAJOR.MINOR.PATCH"`) - and the numeric `MAJORMINORPATCH`.

It one puts the version in the CMake project, CMake should take care of splitting it up into major, minor, and patch.

awni:
I think that's a good idea! Those should be macros and can be used at compile time. I like their naming scheme also we could adapt to MLX. Do you want to do that in this diff?

stemann:
I have added the compile-time definitions to `CMakeLists.txt`.

mlx/version.h
```diff
@@ -0,0 +1,24 @@
...
+  return std::string(TOSTRING(MLX_VERSION));
+}
+
+int version_major() { return MLX_VERSION_MAJOR; }
+
+int version_minor() { return MLX_VERSION_MINOR; }
+
+int version_patch() { return MLX_VERSION_PATCH; }
+
+int version_numeric() { return MLX_VERSION_NUMERIC; }
```
angeloskath, line 22: Nit: I would make all of these constexpr.

angeloskath (commented):
Looks great to me. I would make the functions in `version.h` constexpr but other than that it's great.

mlx/version.h
```diff
@@ -0,0 +1,24 @@
+// Copyright © 2023-2024 Apple Inc.
```
awni, line 1: I think it's worth including this in `mlx.h` by default. It's low footprint and probably expected to be there if you do `"#include <mlx.h>`.

awni:
Looks really nice, thanks for adding that! Would you mind addressing the two comments and then we can merge it.

mlx/version.h
stemann: Fixed

mlx/version.h
stemann: Perhaps fixed - let me know if `version()` should also be constexpr... (that would require it to return something else than std::string)

stemann: Also changed `version()` to be constexpr (returning `const char*`)

awni (approved):
🚀

PR #1830, opened 2025-02-04
Avoid duplicate malloc with custom kernel init

awni:
As title.


angeloskath (approved):
Good catch!

barronalex (approved):
Nice catch!

PR #1833, opened 2025-02-05
Add use_optimal_threadgroups optional flag to run custom kernel with optimal occupancy

spichardo:
## Proposed changes

This flag can be used to use dispatch_threadgroups instead of dispatch_threads (the default) to execute the custom kernel. If True, the optimal grid and threadgroup will be calculated for optimal occupancy. However, the programming of the kernel must assume a flattened operation equivalent to a new grid specification with [grid[0]*grid[1]*grid[2],1,1]. Default: ``False``

I also added a benchmark to show the improvement in `benchmarks/python/custom_kernel_dispatchThreadgroups.py` with the flag on and off. In a M3 Max, we can see an improvement about 13% with a moderately complex kernel after being called 12,000 times.

## Checklist

### DIFF
--- a/benchmarks/python/custom_kernel_dispatchThreadgroups.py
+++ b/benchmarks/python/custom_kernel_dispatchThreadgroups.py
@@ -1,0 +1,672 @@
+import time
+
+import matplotlib.pyplot as plt
+import mlx.core as mx
+import numpy as np
+
+"""
+Test to compare speed using use_optimal_threadgroups = True with default (False) in custom kernel execution
+
+We use a relatively heavy kernel called 12,000 times. The kernel solves the Bio Heat Thermal Equaion (BHTE)
+"""
+
+BHTE_header = """
+#include <metal_stdlib>
+using namespace metal;
+#define pi M_PI_F
+#define Tref 43.0
+
+typedef float FloatingType;
+"""
+
+BHTE_body = """
+#ifdef _METAL
+    kernel  void BHTEFDTDKernel(device float *d_output [[ buffer(0) ]], 
+                                device float *d_output2 [[ buffer(1) ]],
+                                device const float *d_input [[ buffer(2) ]], 
+                                device const float *d_input2 [[ buffer(3) ]],
+                                device const float *d_bhArr [[ buffer(4) ]],
+                                device const float *d_perfArr [[ buffer(5) ]], 
+                                device const unsigned int *d_labels [[ buffer(6) ]],
+                                device const float *d_Qarr [[ buffer(7) ]],
+                                device const unsigned int *d_pointsMonitoring [[ buffer(8) ]],
+                                device float *d_MonitorSlice [[ buffer(9) ]],
+                                device float *d_Temppoints [[ buffer(10) ]],
+                                constant float * floatParams [[ buffer(11) ]],
+                                constant unsigned int * intparams [[ buffer(12) ]],
+                                uint gid[[thread_position_in_grid]])	
+    {
+#endif
+    #ifdef _MLX
+    uint gid = thread_position_in_grid.x;
+    #endif
+
+    #define CoreTemp floatParams[0]
+    #define dt floatParams[1]
+    #define sonication intparams[0]
+    #define outerDimx intparams[1]
+    #define outerDimy intparams[2]
+    #define outerDimz intparams[3]
+    #define TotalStepsMonitoring intparams[4]
+    #define nFactorMonitoring intparams[5]
+    #define n_Step intparams[6]
+    #define SelJ intparams[7]
+    #define StartIndexQ intparams[8]
+    #define TotalSteps intparams[9]
+    
+    // x,y,z indices for grid
+    const int gtidx =  gid/(outerDimy*outerDimz);
+    const int gtidy =  (gid - gtidx*outerDimy*outerDimz)/outerDimz;
+    const int gtidz =  gid - gtidx*outerDimy*outerDimz - gtidy*outerDimz;
+
+    unsigned int DzDy = outerDimz*outerDimy;
+    
+    unsigned int coord = gtidx * DzDy + gtidy * outerDimz + gtidz;
+    
+    float R1,R2,dtp;
+    if(gtidx > 0 && gtidx < outerDimx-1 && gtidy > 0 && gtidy < outerDimy-1 && gtidz > 0 && gtidz < outerDimz-1)
+    {
+
+        const unsigned int label = d_labels[coord];
+
+        d_output[coord] = d_input[coord] + d_bhArr[label] * 
+                          (d_input[coord + 1] + d_input[coord - 1] + d_input[coord + outerDimz] + d_input[coord - outerDimz] +
+                           d_input[coord + DzDy] + d_input[coord - DzDy] - 6.0 * d_input[coord]) +
+                           d_perfArr[label] * (CoreTemp - d_input[coord]);
+        if (sonication)
+        {
+            d_output[coord] += d_Qarr[coord+StartIndexQ];
+        }
+        
+        R2 = (d_output[coord] >= Tref)?0.5:0.25; 
+        R1 = (d_input[coord] >= Tref)?0.5:0.25;
+
+        if(fabs(d_output[coord]-d_input[coord])<0.0001)
+        {
+            d_output2[coord] = d_input2[coord] + dt * pow((float)R1,(float)(Tref-d_input[coord]));
+        }
+        else
+        {
+            if(R1 == R2)
+            {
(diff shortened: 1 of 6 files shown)


angeloskath:
Hi, thanks for the PR! I am not sure it adds something though in the sense that you can simply call the custom kernel with these group and grid dims directly. Why do we need to calculate them? Simply to get the maximum number of threads per thread group?

spichardo:
@angeloskath: It won't work that easily, for starting, we need to access `threadExecutionWidth()` and `maxTotalThreadsPerThreadgroup()` of the pipeline object to do those calculations. As far I can see in the code, those low-level functions of the Metal pipeline object are not exposed yet through MLX to the Python layer. This recipe is based on this note in the official documentation https://developer.apple.com/documentation/metal/calculating-threadgroup-and-grid-sizes.  

Also, we discovered long time ago that we needed to use `dispatch_threadgroups` instead of  `dispatch_threads`  to see the better impact of the recipe. 

I hope this explains why adding a new optional functionality to the call of custom kernel in MLX may be required to use this recipe.

Other approaches could be to expose `threadExecutionWidth()` and `maxTotalThreadsPerThreadgroup()` as part of the properties if the MLX custom kernel, and be able to specify to use `dispatch_threadgroups` instead of `dispatch_threads` when calling the kernel. Either way, some extra functionality is needed.

angeloskath:
Sure but it is more complicated than that anyway since the maximum occupancy isn't necessarily the optimal grid anyway.

Do you have a use case where passing the grid manually doesn't work? Thread execution width is 32 for all intents and purposes. Then the maximum number of threads is going to depend on the kernel (most times it will be 1024) but have you actually encountered a use case where you couldn't set it to 1024 on one device but you could on another and this wasn't something that you could select on the fly in the calling code?

spichardo:
I added a benchmark test where I select a threadgroup size of (256,1,1) and compare that to the recipe, performance was improved by 13% in an execution of 12,000 calls to the kernel.  It is at [benchmarks/python/custom_kernel_dispatchThreadgroups.py](https://github.com/ml-explore/mlx/pull/1833/files#diff-660694a626b03a6587af9ebf4ba32ddd60065b9885c533b934311ac077fa5eca)

Certainly this improvement is more obvious with heavy kernels, in scenarios where you solve time-domain systems where you need to call thousands of time the kernel. But that is exactly our application, we run thousands of these simulations.

We have studied this long and far with very heavy kernels for years; first with Swift-based libraries, then with py-metal-compute. And the conclusion was clear, trying to tune manually was simply too tedious, and the recipe ended being a great compromise to ensure better performance without asking an end-user to worry on this. In many cases, I remember we only managed to match the performance of the recipe when doing manual adjustments in a few , in many other , we didn't. And the test was done using from small matrix sizes to large ones. So we were happy to apply this formula across all our kernels.

It is also important to put in context that our interest on this recipe is because we have our end users do not need to deal with fine tuning the optimal threadgroup size, which can have important impact, and we have matrix sizes ranging small to very large, so it is a challenging situation. The recipe using threadExecutionWidth and maxTotalThreadsPerThreadgroup helped us to keep performance similar NVidia GPUs for several of the most intense kernels.

If this is too disruptive, it is ok, we can just have our own fork for our purposes where we will have this feature enable. I just thought the recipe was simple to implement, and since it is optional, it should not break anything.

I'm sensitive our project is not in the mainstream of the intended projects for MLX, but it can still help significantly other scientific projects that are not ML-oriented. And this type of fine-tuning can help a lot.

angeloskath:
So firstly, I think that your use case is perfect for MLX! Doing physics with custom kernels and being able to take advantage of fast matmuls and general ML, autograd and so on and so forth is a perfect combination.

Going back to the specifics, I just don't understand why you can't use `(1024, 1, 1)` as your group dimensions. Do you have a use case where the maximum thread group size varies among devices and is it hard to manually find it? The reason I am saying this is that your code will almost always set the group dimensions to `(1024, 1, 1)`.

spichardo:
Running with [1024,1,1] produces a wall-time close to the recipe time, indeed, just a hair slower , but it gets worse as you increase the domain size. 

Some quick results with different grid sizes (each tests runs the kernel 12,000 times) in a M3 Max:

[114,114,193] :   Constant group size = 2.6 s, recipe = 2.4s
[281,281,481]:   Constant group size = 36.1 s, recipe = 35.7s
[422,422,721]:  Constant group size = 130.5s, recipe = 123s

I tested changing the order which one was done first, it didn't change the result.

While I agree using the constant value of 1024 gets close, it still remains above the recipe.

This test was with moderate kernel, we are preparing with our biggest kernel (much more complicated than this one), so we can see how this evolves when doing that test. We can test with our smaller kernels, there may be that indeed some cases there may be no difference.

angeloskath:
Interesting, so does this mean that the difference comes from `dispatch_threads` vs `dispatch_threadgroups`? Everything else should be the same in your case. The grid size math just result in thread groups of size `(1024, 1, 1)`.

If that is the case can you change line 90 to also dispatch thread groups to confirm? The runtimes should be identical in that case.

spichardo:
Hi, not sure if that makes sense because `dispatch_threadgroups `is supposed to work differently from `dispatch_threads`, but I did the change as suggested in line 90, and now, using the constant group size skyrocketed the computational time, just in the first small matrix test I got 

[114,114,193] : Constant group size = **635 s**, recipe = 2.5s

I didn't even test for the larger matrix sizes

`dispatch_threadgroups` is supposed to receive a `threadgroupsPerGrid` (first parameter) that is normalized by the dimensions of threadsPerThreadgroup (second parameter). In `dispatch_threads`, the first parameter is `threadsPerGrid`, which is often just the size of the domain where to run the kernel (i.e. [114\*114\*193,1,1]). If I use that as input in `dispatch_threadgroups`, I end with far too many calls to the kernel with indexes that were outside the dimensions of the valid region of the domain.

So I made a second test where threadgroupsPerGrid was calculated with the formula (from [here](https://developer.apple.com/documentation/metal/calculating-threadgroup-and-grid-sizes))
[(totadomainsize+1023)/1024,1,1]

[114,114,193] : Constant group size = 2.7 s, recipe = 2.7s
[281,281,481]: Constant group size = 36.3, recipe = 36.7s
[422,422,721]: Constant group size = 123s, recipe = 123s

Now we got pretty much the same result, so there is definitively a small advantage  in using `dispatch_threadgroups` over `dispatch_threads`

angeloskath:
Thanks for persisting by the way and also for being nice and running all those benchmarks :-)

spichardo:
My pleasure, we are really invested in squeezing the more we can from the GPU for our [application](https://github.com/ProteusMRIgHIFU/BabelBrain).

I printed the values of threadExecutionWidth() and maxTotalThreadsPerThreadgroup(), and they came back with 32 and 1024, respectively, which was a source of the discussion.  

And just for my education, is it always expected that these functions would return 32 and 1024? I never analyzed with other kernels to see whether these values would change depending on the kernel.  That is why we used the recipe recommended in the official documentation, as we didn't want to assume anything. CUDA also has a similar approach to calculate thread and group sizes. You mentioned that "*most times it will be 1024*", but can we be sure? That is why we thought adding this new `use_optimal_threadgroups` flag could help to optimize GPU utilization without having to worry to test which `threadgroupsPerGrid` and `threadsPerThreadgroup` values to use.

angeloskath:
The 32 is pretty much guaranteed. We use it as a constant in many places in MLX. The 1024 depends on the kernel but for most cases it is 1024.

The problem is that this doesn't optimize GPU utilization or occupancy. It just uses the maximum number of threads per thread group. The GPU will schedule several thread groups at the same time anyway. Unfortunately this number needs to be chosen on a per kernel basis if performance is absolutely critical. A great example for instance is our qmv or matmul kernels. Even though 1024 threads per thread group are possible we launch 128 because it utilizes the gpu better.

Going back to the proposed change, I ran on my laptop with `dispatch_threads` and `dispatch_threadgroups` and I don't see any difference. I will run on a bigger machine as well to run larger grids and if there is no difference we 'll circle back to see why there is a difference on your case but I will likely close the PR.

angeloskath:
Ran it on the M2 Ultra as well. I get the exact same numbers with `dispatch_threads` and `dispatch_threadgroups` so I will close this PR.

If you do encounter problems with choosing the thread group size manually, feel free to open an issue or a discussion for them separately.

Btw I get 17.4 with both when using the default grid. I do however have consistently 18 (with both) for the first run probably due to caching or other issues which could be skewing your results above. I would suggest when testing this to run one benchmark and recompile in between or launch it again with different arguments instead of running both in the same program one after the other. The obvious choice is to run it several times and get sth like the median but when a benchmark takes 2 minutes ro run that becomes slightly impractical.

spichardo:
Thanks, this was a great discussion!

PR #1834, opened 2025-02-06
Loading empty list is ok when strict = false

awni:
As title.


barronalex (approved):
Thanks for fixing this! I ran into it earlier today.

PR #1835, opened 2025-02-06
Fix metal sort for certain cases

awni:
- contiguous sort on shape [...., z, 1] at axis=-2  was broken
- fix bug dereferencing potentially invalid ptr
- For now throw if the stride is too big (rather than silently fail)


angeloskath (approved):
🙏

PR #1843, opened 2025-02-07
Fp64 on the CPU

### LINKED ISSUE #799: Support for mlx.float64
**Describe the bug**
Would it be possible to support float64 types. For some numerical simulations, having float64 is important for the accuracy of the simulation. The goal is to use mlx for automatic differentiation in these types of scenarios.

awni:
- Adds most common ops for double support on the CPU
- Throws if using flota64 arrays on the GPU

I didn't implement everything for type double even on the CPU. Specifically:
- Most lapack ops (qrf, eigh, lapack, etc)
- FFTs
- Quantized ops

I think we can add the rest on an as needed basis and maybe for some it won't make sense to ever add them (e.g. quantized).

Partially closes #799

### DIFF
--- a/docs/src/python/data_types.rst
+++ b/docs/src/python/data_types.rst
@@ -51,9 +51,18 @@ The default floating point type is ``float32`` and the default integer type is
    * - ``float32``
      - 4 
      - 32-bit float
+   * - ``float64``
+     - 4 
+     - 64-bit double
    * - ``complex64``
      - 8 
      - 64-bit complex float
+
+
+.. note::
+
+    Arrays with type ``float64`` only work with CPU operations. Using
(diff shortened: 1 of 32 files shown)


angeloskath (approved):
Very nice! Extremely small diff for a whole new type being added.

awni:
CPU benchmarks on MNIST (chosen to maximally stress overhead).

TLDR any differences seem to be in the noise.

Benchmark | Pre | Post
---- | ---- | ----
10 layers, compiled | 0.493 | 0.489
10 layers, not compiled | 0.662 | 0.657
100 layers, compiled |  2.787 |  2.795
100 layers, not compiled | 4.126 | 4.136

sck-at-ucy:
Thank you 🙏🏻☺️

PR #1844, opened 2025-02-08
Fix compilation error on Windows

zcbenz:
The env name can be full uppercased on some systems.

The `std::negate` needs `functional` header.


awni (approved):
Thanks!

PR #1845, opened 2025-02-08
Split vmap incorrect

angeloskath:
Well the vmap of split was completely incorrect. It was returning a single output axis for all the outputs.

Thanks to @tlikhomanenko for finding the bug.


awni (approved):
Good catch, thanks!

Issue #1846, opened 2025-02-08
[BUG] Source build with python 3.13.0 fails; 3.9.17 works


faisalmemon:
https://ml-explore.github.io/mlx/build/html/install.html
says the pre-requisites for python are:
"Using a native Python >= 3.9"
but actually when performing the editable install
`CMAKE_BUILD_PARALLEL_LEVEL=8 pip install -e ".[dev]"`
you get an install failure at the end if your python is newer than 3.9.x, such as 3.13.0.

```
      [100%] Linking CXX shared module /tmp/tmpjioegrhn.build-lib/mlx/core.cpython-39-darwin.so
      [100%] Built target core
      Install the project...
      -- Install configuration: "Release"
      Traceback (most recent call last):
        File "/private/tmp/pip-build-env-b3tvd6sc/overlay/lib/python3.13/site-packages/setuptools/command/editable_wheel.py", line 139, in run
          self._create_wheel_file(bdist_wheel)
          ~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^
        File "/private/tmp/pip-build-env-b3tvd6sc/overlay/lib/python3.13/site-packages/setuptools/command/editable_wheel.py", line 340, in _create_wheel_file
          files, mapping = self._run_build_commands(dist_name, unpacked, lib, tmp)
                           ~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        File "/private/tmp/pip-build-env-b3tvd6sc/overlay/lib/python3.13/site-packages/setuptools/command/editable_wheel.py", line 263, in _run_build_commands
          self._run_build_subcommands()
          ~~~~~~~~~~~~~~~~~~~~~~~~~~~^^
        File "/private/tmp/pip-build-env-b3tvd6sc/overlay/lib/python3.13/site-packages/setuptools/command/editable_wheel.py", line 290, in _run_build_subcommands
          self.run_command(name)
          ~~~~~~~~~~~~~~~~^^^^^^
        File "/private/tmp/pip-build-env-b3tvd6sc/overlay/lib/python3.13/site-packages/setuptools/_distutils/cmd.py", line 339, in run_command
          self.distribution.run_command(command)
          ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^
[...]

Issue #200, opened 2023-12-17
Common Installation Problems


awni:
The docs have more information on [troubleshooting installation](https://ml-explore.github.io/mlx/build/html/install.html).

Two of the more common issues are described below (as well as in the docs).

# Installing from PyPI

A common problem is `pip install mlx` does not find a matching distribution:

```
ERROR: Could not find a version that satisfies the requirement mlx (from versions: none)
ERROR: No matching distribution found for mlx
```

If this happens you are probably using a non-native Python. The output of:

```
python -c "import platform; print(platform.processor())"
```

Should be `arm` not `i386`. You can fix this by [using an environment with a native Python using Conda](https://stackoverflow.com/questions/65415996/how-to-specify-the-architecture-or-platform-for-a-new-conda-environment-apple). Another option is to simply reinstall Conda and setup a new environment.

# Installing from Source

You see the following error when you try to build:

```
error: unable to find utility "metal", not a developer tool or in PATH
```

To fix this, first make sure you have Xcode installed:

```
xcode-select --install
```

Then set the active developer directory:

```
sudo xcode-select --switch /Applications/Xcode.app/Contents/Developer
```

ariaattar:
<img width="526" alt="Screen Shot 2023-12-21 at 3 50 06 PM" src="https://github.com/ml-explore/mlx/assets/72599441/aefdbde0-7856-44e9-902a-c7aeff0ba042">

Having the same issue, platform is outputting `arm` is there something else I should check?

awni:
operating system?

matthewfernst:
Hi there, I'm getting the following error during `make install` is there something silly I'm missing? I was following the steps mentioned [here](https://ml-explore.github.io/mlx/build/html/install.html). It works if I do `sudo make install`, but when I try running a `isnan` that I'm implementing I'm getting this error `[metal::Device] Unable to load kernel ssnaneqint32`. I'm guessing something is wrong with the install? Thank you!

```bash
CMake Error at mlx/backend/metal/kernels/cmake_install.cmake:41 (file):
  file INSTALL cannot copy file
  "/Users/matthewernst/Developer/AIML/mlx/build/mlx/backend/metal/kernels//mlx.metallib"
  to "/usr/local/lib/mlx.metallib": Permission denied.
Call Stack (most recent call first):
  mlx/backend/metal/cmake_install.cmake:42 (include)
  mlx/cmake_install.cmake:57 (include)
  cmake_install.cmake:42 (include)
```

[...]

PR #1848, opened 2025-02-09
Add advice to use miniconda if editable package fails. (#1846)

faisalmemon:
## Proposed changes

This change adds advice for those affected by #1846 
The installation/build process works for different python versions if minconda was used to consistently version the environment.

## Checklist

The revised documentation looks like this:
revised-documentation


awni:
Thanks for the addition. I don't think we need to add this just yet for the one-off issue you had. Let's see if it's a common problem and then we can revisit.

awni:
You could post the steps that worked for you here for now: https://github.com/ml-explore/mlx/issues/200

PR #1856, opened 2025-02-11
Cycle leak break

### LINKED ISSUE #1832: mx.distributed.finalize() ?
I am running into the same issue with several codes the use mx.distributed. The code runs fine but then the different ranks quit ungracefully because there is nothing equivalent  `mx.distributed.finalize() ` to terminate mpi gracefully. Is there a way to have the ranks exit more gracefully? Am I missing something?

awni:
Enable Python garbage collector to break reference cycles with MLX function transformations and other custom MLX objects which hold references Python objects.

There are two mechanisms. For functions which were built with `nb::cpp_function` we now use `mlx_func` as a light-weight wrapper with garbage collection. Python object dependencies have to be registered on construction like so:

```
mlx_fun(wrapped_fun, dep1, dep2, ...)
```

For classes that hold `nb::object`, `nb::callable` etc. we install Python cycle breaking functions in the type slots.

Closes #1832 

Thanks a ton to @wjakob for the help implementing this (see discussion https://github.com/wjakob/nanobind/discussions/919)

### DIFF
--- a/python/src/mlx_func.cpp
+++ b/python/src/mlx_func.cpp
@@ -1,0 +1,108 @@
+// Copyright © 2025 Apple Inc.
+
+#include "python/src/mlx_func.h"
+
+// A garbage collected function which wraps nb::cpp_function
+// See https://github.com/wjakob/nanobind/discussions/919
+
+struct gc_func {
+  PyObject_HEAD
+      // Vector call implementation that forwards calls to nanobind
+      PyObject* (*vectorcall)(PyObject*, PyObject* const*, size_t, PyObject*);
+  // The function itself
+  PyObject* func;
+  // A non-owning reference to dependencies owned by 'func'
+  std::vector<PyObject*> deps;
+};
+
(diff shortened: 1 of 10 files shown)


python/src/mlx_func.cpp
```diff
@@ -0,0 +1,108 @@
...
+  for (auto d : w->deps) {
+    Py_VISIT(d);
+  }
+  Py_VISIT(Py_TYPE(self));
+  return 0;
+};
+
+int gc_func_tp_clear(PyObject* self) {
+  gc_func* w = (gc_func*)self;
+  Py_CLEAR(w->func);
```
angeloskath, line 30: No need to clear the dependencies cause they are non-owning?

angeloskath (approved):
It looks great! Going through the issue and the docs referenced therein also very illuminating.

python/src/mlx_func.cpp
awni: Exactly

PR #1857, opened 2025-02-12
Fixes output donation for IO ops on the GPU

angeloskath:
The CPU task would hold a reference to the output buffer so it couldn't be donated to the next op. This now holds a non-owning reference so it can be donated.

As an example when casting with convert from fp16 to bf16 a 7B model, a fully donatable operation, before we would use 6.8GB peak memory and now 5.3GB. Of course there should be no changing in quantizing since it isn't a donatable operation unfortunately.

One question to be answered is whether we like the method name `unsafe_weak_copy`, perhaps `non_owning_copy` or sth along those lines would be better.

### DIFF
--- a/mlx/array.cpp
+++ b/mlx/array.cpp
@@ -231,17 +231,6 @@ array::~array() {
   }
 }
 
-array array::unsafe_weak_copy() const {
-  return array(
-      array_desc_->data->buffer,
-      shape(),
-      dtype(),
-      strides(),
-      data_size(),
-      flags(),
-      [](auto b) {});
-}
-
 void array::ArrayDesc::init() {
   strides.resize(shape.size());
   size = 1;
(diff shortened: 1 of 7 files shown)


mlx/array.cpp
```diff
@@ -219,6 +231,17 @@ array::~array() {
   }
 }
 
+array array::unsafe_weak_copy() const {
+  return array(
+      array_desc_->data->buffer,
```
awni, line 236: nit: `buffer()`

In theory we could put this function somewhere else (e.g. somewhere private) if we don't want to expose it in the array API since it uses all public methods.

awni (approved):
Very clean. Looks great. Thanks for adding the test!

mlx/array.cpp
angeloskath: Yeah that's a good point. I moved to `mlx/backend/metal/utils.h` cause it seemed appropriate. Let me know what you think and then we can merge.

awni: Looks great. Thanks!

PR #1858, opened 2025-02-12
More buffer donation in some cases

awni:
Previously if an input didn't get donated it was held by the command buffer and no longer eligible for donation.

So sequences like:

```python
a = x + y
b = mx.abs(x)
```

Assuming `x` didn't get donated to `a` then it is impossible for it to be donated to `b`.

This changes that by removing buffers that are safe to not hold in the command buffer.

Benchmark on Transformer training:
Command: `python main.py --gpu --batch_size 8`
Pre:   `Iter 20: Train loss 8.181, It/sec 0.853, Peak memory 17.431 (GB)`
Post: `Iter 20: Train loss 8.260, It/sec 0.885, Peak memory 16.330 (GB)`

### DIFF
--- a/mlx/backend/metal/metal.cpp
+++ b/mlx/backend/metal/metal.cpp
@@ -1,6 +1,5 @@
 // Copyright © 2023-2024 Apple Inc.
 #include <memory>
-
 #include "mlx/backend/metal/device.h"
 #include "mlx/backend/metal/event.h"
 #include "mlx/backend/metal/utils.h"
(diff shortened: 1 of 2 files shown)


angeloskath (approved):
That is pretty cool.

My understanding is, if use count is >1 then someone else is holding it as an input or sth similar. So the last primitive (last as defined by the tape eval order) to use an array is responsible for keeping its buffer around.

awni:
That’s exactly right

PR #1859, opened 2025-02-13
Support loading F8_E4M3 from safetensors

barronalex:
Loads F8_E4M3 tensors in bf16 from safetensors.

The load only works on the GPU at the moment because it uses a custom kernel.


angeloskath (approved):
Looks great!

PR #1860, opened 2025-02-13
Fix creating an array with an int64 scalar

barronalex:
Fixes a bug where mx.array(2**32) results in a `bad_cast` error.
Now large/small ints are automatically upcasted to int64.


awni (approved):
LGTM, thanks for the fix!

PR #1862, opened 2025-02-13
Bitwise Inverse

barronalex:
Add bitwise inverse (complement)


mlx/backend/cpu/unary.cpp
```diff
@@ -85,6 +85,12 @@ void ArcTanh::eval_cpu(const std::vector<array>& inputs, array& out) {
   unary_fp(in, out, detail::ArcTanh());
 }
 
+void BitwiseInverse::eval_cpu(const std::vector<array>& inputs, array& out) {
+  assert(inputs.size() == 1);
+  const auto& in = inputs[0];
+  unary_int(in, out, detail::BitwiseInverse());
```
awni, line 91: This doesn't support `bool`, should we add that to `unary_int`?

awni: Nvm I see you supported it with logical_not, nice.

python/src/ops.cpp
```diff
@@ -4833,6 +4833,28 @@ void init_ops(nb::module_& m) {
...
+      "bitwise_inverse",
+      [](const ScalarOrArray& a_, mx::StreamOrDevice s) {
+        auto a = to_array(a_);
+        return mx::bitwise_inverse(a, s);
+      },
+      nb::arg(),
+      nb::kw_only(),
+      "stream"_a = nb::none(),
+      nb::sig(
+          "def bitwise_inverse(a: Union[scalar, array], b: Union[scalar, array], stream: Union[None, Stream, Device] = None) -> array"),
```
awni, line 4846:
```suggestion
          "def bitwise_inverse(a: Union[scalar, array], stream: Union[None, Stream, Device] = None) -> array"),
```

barronalex:
Nice catch on inverse vs invert!

mlx/backend/cpu/simd/base_simd.h
```diff
@@ -95,6 +95,16 @@ DEFAULT_UNARY(sqrt, std::sqrt)
 DEFAULT_UNARY(tan, std::tan)
 DEFAULT_UNARY(tanh, std::tanh)
 
+template <typename T>
+Simd<T, 1> bitwise_invert(Simd<T, 1> in) {
+  return ~in.value;
+}
```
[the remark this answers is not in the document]
awni: I notice it's only defined for the scalar ops so I assume it's unused o/w we should define it for the vector ops as well.

awni:
I left one more comment and also one last thing:

Can you add this to the list of fusable unary primitives in compile: https://github.com/ml-explore/mlx/blob/main/mlx/compile.cpp#L19 ?

mlx/backend/cpu/simd/base_simd.h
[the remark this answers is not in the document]
barronalex: Good catch, I removed it.

PR #1863, opened 2025-02-13
Revert "More buffer donation in some cases (#1858)"

awni:
This reverts commit d274ae77f2083cef3af3ebd01465c26f9c958cb1.

Unfortunately I think this was a bogus optimization. It's not stream safe and I don't think there is an easy way to make it so.

Two streams that use the same input can arrive to the `use_count` test simultaneously and both get a `use_count` of 2 and both proceed to let go of the buffer which destroys it before it's done being needed.

### DIFF
--- a/mlx/backend/metal/metal.cpp
+++ b/mlx/backend/metal/metal.cpp
@@ -1,5 +1,6 @@
 // Copyright © 2023-2024 Apple Inc.
 #include <memory>
+
 #include "mlx/backend/metal/device.h"
 #include "mlx/backend/metal/event.h"
 #include "mlx/backend/metal/utils.h"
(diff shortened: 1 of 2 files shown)


angeloskath (approved):
Ouch :-( 

Do you think it makes sense to protect with a lock and re-apply?

awni:
I think there is also a race condition involved with `async_eval`. It's difficult to pin down... but this diff also resolves the bug with generation when I increase the max ops per buffer and a lock doesn't fix that.

I think what's happening is that here we rely on the user holding a reference to all the buffers that don't have parents in the graph after everything is detached until the eval is finished. But with async eval you can delete something before it's done being evaluated. Kind of like this case:

```python
def fn():
    x = mx.zeros((5, 5))
    x = mx.abs(x)
    return x

x = fn()
mx.async_eval(y)
del x
```

awni:
I'm going to merge this to resolve the bugs and maybe we revive it if we can come up with a clean solution. Lmk if you have any ideas about that.

PR #1869, opened 2025-02-15
redesign for faster cpu/gpu synch

awni:
Notes on important changes:

- All `eval_cpu` are now logically split in two parts and look a lot like how `eval_gpu` looks.
  - Setup which does allocate the output, sets flags, strides etc
  - Dispatch the kernel. This just puts a task in the streams thread for now. In the future we may make it multi threaded.
  - Like the GPU the CPU task now registers inputs and outputs with the command encoder  `add_input_array`, `add_output_array`
  - The CPU task is also required to register temporaries `add_temporary`
- All task setup for any stream (CPU/GPU/etc) is now done on the main thread. Concretely `eval_cpu` and `eval_gpu` are now called on the main thread. It didn't make much sense from a perf standpoint to have sub-threads for this and it was overly complicated from a synchronization standpoint (as one would need to synchronize for both setup and completion). Furthermore this really simplifies things like thread safety of the eval moving forward. I think it's overall a nice change. In-fact there is now some room to pipeline CPU setup with CPU work so potential for speedups there.
- Use `fence` for synchronization within an eval across streams. Use `event` for synchronization outside the eval loop or across multiple eval loops (e.g. when using `async_eval`).
- The way donation is done has changed. There is no longer a requirement to move the data shared pointer out of the array when donating (deleted the `array::move_shared_buffer` API). This was a bit tedious.. so instead eval driver handles it by not retaining input buffers that got donated to their respective outputs.

CPU-only benchmarks:

Bench | Pre | Post |
----- | ----- | -----
MNIST layers=10, compiled | 0.662 | 0.659
MNIST layers=10, not compiled | 0.785 | 0.740


mlx/distributed/ring/ring.cpp
```diff
@@ -593,114 +583,127 @@ class RingGroup : public GroupImpl {
...
+  }
+
+  void recv(array& out, int src, Stream stream) override {
+    auto& encoder = cpu::get_command_encoder(stream);
+    encoder.set_output_array(out);
+    encoder.dispatch(
+        [out_ptr = out.data<char>(), nbytes = out.nbytes(), src, this]() {
+          int right = (rank_ + 1) % size_;
+          int left = (rank_ + size_ - 1) % size_;
+          if (src == right) {
```
angeloskath, line 626: Not sure if this will be auto-fixed when you rebase on main so I am leaving a comment to remember to check that the check here is swapped to `left` then `right` opposite to `send` above.

mlx/backend/cpu/eval.cpp
```diff
@@ -0,0 +1,44 @@
...
+  for (auto& in : arr.inputs()) {
+    buffers.insert(in.data_shared_ptr());
+  }
+  for (auto& s : arr.siblings()) {
+    buffers.insert(s.data_shared_ptr());
+  }
+  // Remove the output if it was donated to by an input
+  if (auto it = buffers.find(arr.data_shared_ptr()); it != buffers.end()) {
+    buffers.erase(it);
+  }
```
angeloskath, line 38: Shouldn't this also be done for the siblings? Otherwise we guarantee they won't be donated later on. Sth like the following in the loop ought to do it.

```cpp
if (buffers.find(s.data_shared_ptr()) == buffers.end()) {
  buffers.insert(s.data_shared_ptr());
}
```

mlx/backend/cpu/inverse.cpp
```diff
@@ -103,34 +103,57 @@ void tri_inv(array& inv, int N, int i, bool upper) {
...
-  copy(a, inv, a.flags().row_contiguous ? CopyType::Vector : CopyType::General);
+  if (a.flags().row_contiguous && a.is_donatable()) {
+    inv.copy_shared_buffer(a);
+  } else {
+    copy(
+        a,
+        inv,
+        a.flags().row_contiguous ? CopyType::Vector : CopyType::General,
+        stream);
+  }
```
angeloskath, line 126: I think we can simply replace the if with the original copy. It will either donate or copy.

mlx/backend/cpu/masked_mm.cpp
```diff
@@ -104,31 +105,39 @@ void BlockMaskedMM::eval_cpu(const std::vector<array>& inputs, array& out) {
     return;
   }
 
+  auto& encoder = cpu::get_command_encoder(stream());
   if (K == 0) {
-    std::memset(static_cast<void*>(out.data<float>()), 0, out.nbytes());
+    encoder.set_output_array(out);
+    encoder.dispatch([out_ptr = out.data<void>(), nbytes = out.nbytes()]() {
+      std::memset(out_ptr, 0, nbytes);
```
angeloskath, line 112: Absolute nitpick but I think this might be slightly better with `std::fill_n(out_ptr, size, 0);` mostly because it is typed but ultimately it would be the same.

mlx/backend/cpu/encoder.h
```diff
@@ -0,0 +1,50 @@
...
+  }
+
+  // Hold onto temporaries until any already scheduled tasks which use it as
+  // an input are complete.
+  void add_temporaries(std::vector<array> arrays) {
+    if (arrays.empty()) {
+      return;
+    }
+    scheduler::enqueue(stream_, [arrays = std::move(arrays)]() {});
+  }
```
angeloskath, line 36: Not a problem necessarily but `add_temporaries` on the CPU "CommandEncoder" has to be after `dispatch` while on the GPU it can be before. One solution would be for the `CommandEncoder` to collect temporaries and then `cpu::eval` to dispatch the task that keeps them in memory possibly together with the rest of the buffers.

This would require `cpu::eval` to know about the `CommandEncoder` but it is already the case in the metal side so I don't think this is a problem and it will allow us to use `add_temporary` where it makes intuitive sense and reduce the number of tasks in the queue.

angeloskath (commented):
So far it looks great. I read about half way through. I didn't read the important bits yet, neither the changes in `event.h` nor `transforms.cpp` and `fence.h`. The API changes look really great imho and unify cpu and gpu primitive implementations.

mlx/backend/cpu/inverse.cpp
awni: Good point!

mlx/backend/cpu/encoder.h
awni: I think it's a very good suggestion as it's actually really not good that `add_temporary` has to be called at just the right spot and otherwise it could fail silently.

mlx/backend/cpu/conv.cpp
```diff
@@ -98,219 +113,237 @@ void slow_conv_2D(
...
+                    wt_stride_W = wt.strides()[2],
+                    wt_stride_C = wt.strides()[3],
+
+                    out_stride_N = out.strides()[0],
+                    out_stride_H = out.strides()[1],
+                    out_stride_W = out.strides()[2],
+                    out_stride_O = out.strides()[3],
+
+                    padding,
+                    wt_strides,
```
jagrit06, line 159: It looks like we end up making 2 copies of the `wt_strides` here 
These are the slower CPU conv kernels, so its not an urgent change though

jagrit06 (commented):
Looks great! I still need to go through the fence and event parts in detail, will do that soon!

mlx/backend/cpu/conv.cpp
awni: It's not two copies.. the naming here is a bit unfortunate:

- `wt_strides` is the strides for applying the kernel.
- `wt.strides()` and `wt_stride_*` are the strides of the underlying array 

This is as it was.. it is due for a bit of a clean up though.

awni:
Fast synchronization works well with the redesign. On an M1 max, synch latencies are 10x less which is the same as when doing custom synchronization within the primitive.

```
mpirun -np 2 python benchmarks/python/synchronize_bench.py 
```
All Reduce: time per iteration 0.232273 (ms)
All gather: time per iteration 0.234606 (ms)

```
MLX_METAL_FAST_SYNCH=1 mpirun -np 2 python benchmarks/python/synchronize_bench.py
```

All Reduce: time per iteration 0.019380 (ms)
All gather: time per iteration 0.021714 (ms)

awni:
For the GPU benchmarks are not noticeably different:

Generation speed is unchanged:
```
mlx_lm.generate --model mlx-community/Mistral-7B-Instruct-v0.3-4bit --prompt "Write a story about Einstein" -m 512
```

Pre: Generation: 512 tokens, 129.661 tokens-per-sec
Post: Generation: 512 tokens, 129.765 tokens-per-sec


Transformer training speed / memory use unchanged:
Pre: Iter 30: Train loss 7.969, It/sec 6.254, Peak memory 5.525 (GB)
Post: Iter 30: Train loss 7.981, It/sec 6.253, Peak memory 5.525 (GB)

angeloskath (approved):
It looks good. Love the event for syncing across evals. The Event value is not needed anymore but I think it is nice this way as it can be used potentially for more things.

PR #1870, opened 2025-02-15
Compiler warnings are errors

awni:
Warnings are errors. We have pretty relaxed warning settings in MLX so this shouldn't be too strict, but if it is annoying we can also enable it only for our CI or something like that.


angeloskath (approved):
Nice!

xhochy:
It's annoying ;) I commented on the merged commit: https://github.com/ml-explore/mlx/commit/5274c3c43f932ba2bddf0235efd7437daecd7b09#commitcomment-153633843

It breaks releases on conda-forge.

PR #1874, opened 2025-02-17
Add bincount method to array class

### LINKED ISSUE #1786: Missing something like `numpy.bincount`
While converting a KNN implementation from NumPy to MLX Swift  I stumbled upon the usage of a `bincount` to count the number of occurrences of each value in array. I couldn't find a MLX equivalent so I did the following:

```swift
let prediction =  kNearestLabels
        .asArray(Int.self) // MLXArray -> [Int]
        .reduce(into: [:]) { $0[$1, default: 0] += 1 }  // count labels
        .max(by: { $0.1 < $1.1 })!  // max '(label, count)' pair
        .key // label (prediction)
```

The NumPy code did this in just one line: 

```swift
np.bincount(k_nearest_labels).argmax()
```

Is there a more "MLX way" of doing this? If not, maybe MLX should have this `bincount` method, seems quite handy.

anupamme:
Fixes #1786

Add `bincount` method to MLX to count occurrences of each value in an array.

* **mlx/array.cpp**: Implement `bincount` method using `std::unordered_map` to store counts and return an array with the counts of each value in the input array.
* **mlx/array.h**: Declare `bincount` method in the `array` class.
* **python/src/array.cpp**: Add `bincount` method to the Python API for the `array` class and implement it using the `bincount` method from `mlx/array.cpp`.
* **python/tests/test_array.py**: Add test cases for the `bincount` method in the `array` class to verify its correctness with various input arrays.

---

For more details, open the [Copilot Workspace session](https://copilot-workspace.githubnext.com/ml-explore/mlx/pull/1874?shareId=cfbe7bbd-6b78-4b29-943f-4bf7ed4a5012).

### DIFF
--- a/mlx/array.cpp
+++ b/mlx/array.cpp
@@ -1,4 +1,3 @@
-// Copyright © 2023-2024 Apple Inc.
 #include <functional>
 #include <unordered_map>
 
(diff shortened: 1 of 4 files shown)


angeloskath:
Hi, the linked issue has a much more concise and correct implementation as follows

```python
def bincount(input, max_val: int):
    return mx.zeros(max_val, dtype=mx.int32).at[input].add(1)
```

this also runs in the GPU and doesn't require evaluating `input` before computing the bin count.

PR #1876, opened 2025-02-18
Enforce triangular matrix form in `tri_inv`

### LINKED ISSUE #1871: solve_triangular result in the case of an incorrectly passed value for upper argument
Hi,

I was trying to add solve_triangular in Keras MLX backend support using `mx.linalg.solve_triangular` and it failed the test where the upper field value is wrongly passed.
All other frameworks give the same output while MLX gives a different one so I was wondering what would the reason be.

Here is the code I used:

```
import numpy as np
import mlx.core as mx
import tensorflow as tf
import jax.scipy as jsp
import torch

a = np.array(
    [
        [1., 2.],
        [0., 5.]
    ]
)
lower = True 
upper = not lower

b = np.array([2., 10.])

x_jax = jsp.linalg.solve_triangular(a, b, lower=lower)
print(f"JAX result:\n{x_jax}")

x_tf = tf.squeeze(
    tf.linalg.triangular_solve(
        a,
        tf.expand_dims(b, axis=-1),
        lower=lower
    ),
    axis=-1
)
print(f"TF result:\n{x_tf}")

x_torch = torch.linalg.solve_triangular(
    torch.tensor(a),
    torch.unsqueeze(torch.tensor(b), axis=-1),
    upper=upper
).squeeze(axis=-1)
print(f"Torch result:\n{x_torch}")

x_mlx = mx.linalg.solve_triangular(
    mx.array(a),
    mx.array(b),
    upper=upper,
    stream=mx.cpu
)

print(f"MLX result:\n{x_mlx}")
```

The output is the following:
```
JAX result:
[2. 2.]
TF result:
[2. 2.]
Torch result:
tensor([2., 2.])
MLX result:
array([22, 2], dtype=float32)
```

abeleinin:
Fixes #1871.

### DIFF
--- a/mlx/linalg.cpp
+++ b/mlx/linalg.cpp
@@ -284,7 +284,7 @@ array tri_inv(
     const array& a,
     bool upper /* = false */,
     StreamOrDevice s /* = {} */) {
-  return inv_impl(a, /*tri=*/true, upper, s);
+  return inv_impl(upper ? triu(a) : tril(a), /*tri=*/true, upper, s);
 }
 
 array cholesky(
--- a/mlx/linalg.cpp
+++ b/mlx/linalg.cpp
@@ -284,7 +284,7 @@ array tri_inv(
     const array& a,
     bool upper /* = false */,
     StreamOrDevice s /* = {} */) {
-  return inv_impl(upper ? triu(a) : tril(a), /*tri=*/true, upper, s);
+  return inv_impl(a, /*tri=*/true, upper, s);
 }
 
 array cholesky(
--- a/mlx/backend/cpu/inverse.cpp
+++ b/mlx/backend/cpu/inverse.cpp
@@ -81,7 +81,22 @@ void general_inv(array& inv, int N, int i) {
 void tri_inv(array& inv, int N, int i, bool upper) {
   const char uplo = upper ? 'L' : 'U';
   const char diag = 'N';
-  int info = strtri_wrapper(uplo, diag, inv.data<float>() + N * N * i, N);
+  float* data = inv.data<float>() + N * N * i;
+  int info = strtri_wrapper(uplo, diag, data, N);
+
+  // zero out the other triangle
+  if (upper) {
+    for (int i = 0; i < N; i++) {
+      std::fill(data, data + i, 0.0f);
+      data += N;
+    }
+  } else {
+    for (int i = 0; i < N; i++) {
+      std::fill(data + i + 1, data + N, 0.0f);
+      data += N;
+    }
+  }
+
   if (info != 0) {
     std::stringstream ss;
     ss << "inverse_impl: triangular inversion failed with error code " << info;
--- a/python/tests/test_linalg.py
+++ b/python/tests/test_linalg.py
@@ -175,6 +175,13 @@ class TestLinalg(mlx_tests.MLXTestCase):
                     mx.allclose(M @ M_inv, mx.eye(M.shape[0]), rtol=0, atol=1e-5)
                 )
 
+        # Ensure that tri_inv will 0-out the supposedly 0 triangle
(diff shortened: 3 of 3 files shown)


mlx/linalg.cpp
```diff
@@ -284,7 +284,7 @@ array tri_inv(
     const array& a,
     bool upper /* = false */,
     StreamOrDevice s /* = {} */) {
-  return inv_impl(a, /*tri=*/true, upper, s);
+  return inv_impl(upper ? triu(a) : tril(a), /*tri=*/true, upper, s);
```
awni, line 287: This is a nice simple fix.. but at the same time I think it would be more efficient if we handled the arguments in L275 correctly. 

I'm just a bit confused because we use [strtri](https://www.netlib.org/lapack/explore-html-3.6.1/de/d76/strtri_8f_a0e63d634f0c29d919e96a78c58a38bd7.html) for the triangular case. And it says the non-used triangle should not be referenced. So is there a bug in the dispatch or what's going wrong there?

abeleinin: We could handle it on L275, but I think it's an issue with the `Inverse` CPU backend. Specifically, the identity $(A^{-1})^T = (A^T)^{-1}$ we use to avoid transposing the input to col-contiguous form (see comment on [L93-L96 inverse.cpp](https://github.com/ml-explore/mlx/blob/4c1dfa58b7c076bae187e249fa49c6a0337a724c/mlx/backend/cpu/inverse.cpp#L93)).

Currently, for `tri_inv` we're relying on the `uplo` flag to transpose the input ([L82 inverse.cpp](https://github.com/ml-explore/mlx/blob/4c1dfa58b7c076bae187e249fa49c6a0337a724c/mlx/backend/cpu/inverse.cpp#L82)) instead of determining the triangular form, which is what `strtri` expects. The solution I'm thinking of is to copy the input to col-contiguous and pass it with the correct `uplo` flag. I can update the PR if you think this approach seems more reasonable.

angeloskath:
I move the `triu` and `tril` into the `strtri_wrapper` it because it is significantly faster than the general operation. I am adding some numbers for comparison (on my laptop)

Before
```
Timing invert (2, 32, 32) ... 0.19954 msec
Timing invert (128, 32, 32) ... 10.13377 msec
Timing invert (2, 128, 128) ... 1.87379 msec
Timing invert (128, 128, 128) ... 114.81336 msec
Timing invert (2, 1024, 1024) ... 114.22994 msec
Timing invert (8, 1024, 1024) ... 469.61235 msec
Timing invert (2, 4096, 4096) ... 5269.60609 msec
```

Using triu and tril
```
Timing invert (2, 32, 32) ... 6.24470 msec
Timing invert (128, 32, 32) ... 18.27467 msec
Timing invert (2, 128, 128) ... 8.72328 msec
Timing invert (128, 128, 128) ... 155.31334 msec
Timing invert (2, 1024, 1024) ... 155.92783 msec
Timing invert (8, 1024, 1024) ... 560.89717 msec
Timing invert (2, 4096, 4096) ... 5606.98880 msec
```

Current
```
Timing invert (2, 32, 32) ... 0.21191 msec
Timing invert (128, 32, 32) ... 10.60309 msec
Timing invert (2, 128, 128) ... 1.82560 msec
Timing invert (128, 128, 128) ... 118.94545 msec
Timing invert (2, 1024, 1024) ... 124.48269 msec
Timing invert (8, 1024, 1024) ... 545.40635 msec
Timing invert (2, 4096, 4096) ... 5385.82807 msec
```

I think it is clearly faster with a similar diff size. I am still not sure I like the cost though because some times it can be >10%.

awni (approved):
LGTM!

abeleinin:
@angeloskath Cool speed up!

PR #1879, opened 2025-02-18
GPU Hadamard for large N

barronalex:
Add support for large Hadamard transforms on the GPU.

For $N=2^{24}$ the GPU version is about 50x faster than the CPU:

```
Timing hadamard_transform ... 2.32494 msec
Timing hadamard_transform ... 123.39948 msec
```

### DIFF
--- a/mlx/ops.cpp
+++ b/mlx/ops.cpp
@@ -473,8 +473,19 @@ array hadamard_transform(
     std::optional<float> scale_ /* = std::nullopt */,
     StreamOrDevice s /* = {} */) {
   // Default to an orthonormal Hadamard matrix scaled by 1/sqrt(N)
-  float scale = scale_.has_value() ? *scale_ : 1.0f / std::sqrt(a.shape(-1));
+  int n = a.ndim() > 0 ? a.shape(-1) : 1;
+  float scale = scale_.has_value() ? *scale_ : 1.0f / std::sqrt(n);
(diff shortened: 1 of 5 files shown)


angeloskath:
Looks great! As per offline discussion let's move this into the primitive instead.

angeloskath:
This should be fine to review and merge now.

The kernel is a bit faster than copying and calling the contiguous one and it also has the benefit of being completely in-place which the transpose copy can't do.

hadamard

The 16K elements in-place transform was not launching correctly (max threads per thread group 832) so I reduced the limit to 8K. @barronalex if you remember encountering the same issue before let me know how you did fix it 🤔.

mlx/ops.cpp
```diff
@@ -473,8 +473,10 @@ array hadamard_transform(
     std::optional<float> scale_ /* = std::nullopt */,
     StreamOrDevice s /* = {} */) {
   // Default to an orthonormal Hadamard matrix scaled by 1/sqrt(N)
-  float scale = scale_.has_value() ? *scale_ : 1.0f / std::sqrt(a.shape(-1));
+  int n = a.shape(-1);
+  float scale = scale_.has_value() ? *scale_ : 1.0f / std::sqrt(n);
   auto dtype = issubdtype(a.dtype(), floating) ? a.dtype() : float32;
```
awni, line 478: Do you mind adding a check here that `a.ndim() > 1`? (No one shall make MLX segfault!)

awni (approved):
Awesome!!

Issue #1873, opened 2025-02-17
[BUG] Running Conv2d multiple times returns different results in 0.23.0


kaeru-shigure:
**Describe the bug**

Running Conv2d multiple times returns different results in 0.23.0

**To Reproduce**

# Warning

**When repeatedly processing the issue in version 0.23.0, the GPU usage of the Window Server increased and couldn't recover without a reboot. It might be due to another cause, but please be cautious.**

```python
import mlx.core as mx
import mlx.nn as nn
from importlib.metadata import version

print(f"Running on MLX version {version('mlx')}")

TEST_STEPS = 200
ok_count = 0
ng_seed = -1
for i in range(TEST_STEPS):
    mx.random.seed(i)
    x = mx.random.normal((1, 1, 168, 320,))
    layer = nn.Conv2d(320, 320, kernel_size=3, stride=1, padding=1)
    ok = mx.allclose(layer(x), layer(x))
    print("." if ok else "!", end="")
    if i%50 == 49: print("")
    if ok:
        ok_count += 1
    else:
        ng_seed = i
        break # Freezes if continues, DO NOT DELETE IT.

if ng_seed != -1:
    print(f"\nNG seed: {ng_seed}")
else:
    print("All OK")
print(f"PASSED {ok_count}/{TEST_STEPS}")
```
Always no problem in 0.22.1:
```
Running on MLX version 0.22.1
..................................................
..................................................
..................................................
..................................................
All OK
PASSED 200/200
```
Randomly fails in 0.23.0:
```
Running on MLX version 0.23.0
.!
NG seed: 1
PASSED 1/200
```
```
Running on MLX version 0.23.0
..!
NG seed: 2
PASSED 2/200
```

**Expected behavior**
Same result how many times I run it.

**Desktop (please complete the following information):**
 - OS Version: MacOS Sequoia 15.3
 - Version: 0.23.0, py3.12.4

awni:
@jagrit06 maybe this is related to #1803? Do you mind checking if there is a bug here?

filipstrand:
I can confirm this change of behaviour too in mflux from yesterday: https://github.com/filipstrand/mflux/issues/128. 

Tried to install latest mlx from source from a couple of different commits and it seems like this one caused the issue https://github.com/ml-explore/mlx/commit/2dc307f2e6ee9364f591878cdd5fccd1eef8fd58

angeloskath:
[...]

PR #1803, opened 2025-01-27
Winograd Update for Small batches 

jagrit06:
## Proposed changes

- Build in padding to Winograd kernels 
- Add new fused Winograd kernel 
- Enable weight flipping in Winograd kernels 

## Checklist

Put an `x` in the boxes that apply.

- [x] I have read the [CONTRIBUTING](https://github.com/ml-explore/mlx/blob/main/CONTRIBUTING.md) document
- [x] I have run `pre-commit run --all-files` to format my code / installed pre-commit prior to committing changes
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have updated the necessary documentation (if needed)

PR #1882, opened 2025-02-18
Fix convs by reverting #1803

angeloskath:
Simply reverts #1803 and adds a test that fails on main. Also bumps the version for a patch release.

The bug appears to be an OOB write given the test in #1873 .


barronalex (approved):
👍

PR #1883, opened 2025-02-19
Ensure linspace always contains start and stop

### LINKED ISSUE #1872: [BUG] linspace should always include the stop value
**Describe the bug**
`linspace` truncates values in some cases

**To Reproduce**

(lazily casting to numpy to print to see precision)
```python
import mlx.core as mx
import jax.numpy as jnp
import numpy as np

range = (0.21436651, 0.96568555)
num = 6
a = mx.linspace(range[0], range[1], num, dtype=mx.float32)
print(np.array(a))
a = jnp.linspace(range[0], range[1], num, dtype=jnp.float32)
print(np.array(a))
# [0.21436651 0.3646303  0.5148941  0.6651579  0.8154217  0.9656855 ]
# [0.21436651 0.3646303  0.5148941  0.6651579  0.81542176 0.96568555]
```

**Expected behavior**
I'd expect `linspace` to produce the same results as jax, without truncation for the same dtype

 - OS Version: 15.2
 - Version: 0.23.0

**Additional context**
I was using this to create bin edges for a histogram, and it was excluding the right most value because of the truncation when calculating counts. I can work around this by setting the last value if it's not an easy fix.

angeloskath:
Fixes #1872 . It makes `linspace` a bit slower (at least if we are not using `mx.compile`), however, I think it is reasonable to expect that the `start` and `stop` are always exact (or the nearest representable value).

### DIFF
--- a/mlx/ops.cpp
+++ b/mlx/ops.cpp
@@ -230,16 +230,16 @@ array linspace(
     throw std::invalid_argument(msg.str());
   }
   if (num == 1) {
-    return astype(array({start}), dtype, to_stream(s));
+    return astype(array({start}), dtype, s);
   }
-  array sequence = arange(0, num, float32, to_stream(s));
-  float step = (stop - start) / (num - 1);
+  array t = divide(arange(0, num, float32, s), array(num - 1, float32), s);
+  array t_bar = subtract(array(1, float32), t, s);
   return astype(
-      add(multiply(sequence, array(step), to_stream(s)),
-          array(start),
-          to_stream(s)),
+      add(multiply(t_bar, array(start, float32), s),
+          multiply(t, array(stop, float32), s),
+          s),
       dtype,
-      to_stream(s));
+      s);
 }
 
 array astype(array a, Dtype dtype, StreamOrDevice s /* = {} */) {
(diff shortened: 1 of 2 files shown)


mlx/ops.cpp
```diff
@@ -230,16 +230,16 @@ array linspace(
...
   return astype(
-      add(multiply(sequence, array(step), to_stream(s)),
-          array(start),
-          to_stream(s)),
+      add(multiply(t_bar, array(start, float32), s),
+          multiply(t, array(stop, float32), s),
+          s),
       dtype,
-      to_stream(s));
+      s);
```
awni, line 242: You could do it in a couple fewer array ops by simplifying:

```
start * (1 - t) + stop * t
```

To:

```
(stop - start) * t - start
```

angeloskath: So the reason I went this way is cause unfortunately the simpler `(stop - start) * t + start` doesn't give us the numerical stability we want. The issue comes from expecting `stop - start + start` to equal `stop` but that is not a property of floats.

I haven't properly verified it but I am nearly certain that `x * 0 == 0` and `x * 1 == x` is a property of floats (with the exceptions of `inf` and `nan` of course. So as a result the slower version I wrote does guarantee that the final value will be `stop` as we also have the guarantee that `x + 0 = x` in floats.

PR #1884, opened 2025-02-19
Raise an exception in the rope op if input is integer

angeloskath:
We could implicitly cast but I think that makes a bit more sense. It is very likely that someone accidentally is passing ints in.


awni:
It looks like we may need to revise the tolerance for RopE fp16 🤔

awni (approved):
Looks great, thanks! I think the test is just a tolerance issue so if we fix that should be good to merge.

PR #1885, opened 2025-02-19
Ring update

angeloskath:
Updates the ring distributed backend as follows:

- We add all sockets to an object that provides asynchronous send/recv.
    - The current implementation is using a thread per socket and polls manually to send/recv the data.
    - I had a kqueue implementation that was slower but it can be something to revisit.
- Rewrite all_sum using the above which is both simpler and more performant
- Implement send/recv to neighbors

All gather coming next.


awni (approved):
Looks great!!

angeloskath:
Adding the benchmark against MPI ring reduce for anybody that is interested. This is on a single TB ring and MPI is configured to perform ring reduce over the same interfaces.

ring_reduce

PR #1887, opened 2025-02-19
Limit compile buffers

### LINKED ISSUE #1880: Metal Compiled primitive: Process finished with exit code 134 (interrupted by signal 6:SIGABRT)
I have been experimenting with including a Transformer layer in Physics-Informed NN (PINN) that uses autograd (mx.grad) for the Partial Differential Equation operator. If I try to use the standard

```
@partial(mx.compile, inputs=state, outputs=state)  # optional: can remove if you prefer
def train_step_2_5D():
    val, grads = nn.value_and_grad(mlp_model, loss_fn_2_5D)()
    return val, grads
```

the code crashes with the error below. 

----------------

Terminating due to uncaught exception: [compile] Too many inputs/outputs fused in the Metal Compiled primitive which exhausted the available argument buffers for the kernel. Please file an issue with the function that results in this error. The name of the kernel is 'AEf4BroadcastBAFf4AddAAEAGf4MultiplyDEAHf4MultiplyCAGAIf4AddAFAHAJf4AddAIFAKf4MultiplyIJALf4AddHAKAMf4MultiplyGALANf4BroadcastAMAOf4AddAJANAPf4MultiplyGKAQf4BroadcastAPARf4MultiplyDAQASf4MultiplyCARATf4AddAOASAUf4AddATLAVf4MultiplyMNAWf4MultiplyAVOAXf4MultiplyGAWAYf4BroadcastAXAZf4MultiplyDAYBAf4MultiplyCAZBBf4AddAUBABCf4AddBBPBDf4NegativeAWBEf4MultiplyQBDBFf4MultiplyIBEBGf4AddBFRBHf4MultiplyGBGBIf4BroadcastBHBJf4AddBCBIBKf4MultiplySTBLf4AddBKUBMf4MultiplyVWBNf4AddBMXBOf4MultiplyYZBPf4DivideBOAABQf4NegativeBPBRf4AddABACBSf4AddBRADBTf4MultiplyMBSBUf4MultiplyBTOBVf4AddBQBUBWf4NegativeBVBXf4MultiplyQBWBYf4MultiplyIBXBZf4AddBNBYCAf4MultiplyGBZCBf4BroadcastCACCf4AddBLCBCDf4MultiplyGBVCEf4BroadcastCDCFf4MultiplyDCECGf4MultiplyCCFCHf4AddCCCGCIf4AddBJCH_VVVVVVVVVVVVVVVVVVVVVVVVVVVVVV_f4f4f4f4f4f4f4f4f4f4f4f4f4f4f4f4f4f4f4f4f4f4f4f4f4f4f4f4f4f4_11160318154034397263_strided_1'

Process finished with exit code 134 (interrupted by signal 6:SIGABRT)

-----------------

Turning the compilation off allows the code to run fine. If I remove the Transformer layer from the PINN and leave the MLP layers the code can be compiled and is pretty fast. The Transformer layer is actually pretty minimal 

```
        input_dim=4,
        hidden_dim=16,
        output_dim=4,  # 4 outputs now
        n_tokens=2,
        num_layers=1,
        num_heads=1,
        mlp_dims=32,
        post_mlp_dims=32,
        activation=nn.silu,
        checkpoint=False
```

but I guess it might be the case that even with such a small Transformer when combined with the autograd for the PDE operator it becomes too large for the compilation? Just guessing 🤓. 

I wonder if there is a way to work around this so I can compile the code, because judging from the version without the Transformer layer that be compiled, it will produce a significant speed gain.

awni:
Limits the number of input/output buffers for compiled primitive to avoid crashing in Metal. 

Closes #1880 

Note this is an approximation in that we still break the primitive in the case that the computation fans out to many possible inputs and then fans back in to just a few. It should be a rare case ... but there could be more optimal approaches to deal with that. It seems overly complex for now though.

Another option is to use Metal argument buffers which could hold unlimited buffers. The problem with this approach is that there are some limitations with Metal 2 / VMs. We [removed argument encoders](https://github.com/ml-explore/mlx/pull/683) for that reason. It's possible they can be made to work, not certain.

### DIFF
--- a/mlx/compile.cpp
+++ b/mlx/compile.cpp
@@ -15,6 +15,7 @@
 namespace mlx::core {
 
 constexpr int max_compile_depth = 11;
+constexpr int max_compile_arrays = 24;
 
 bool is_unary(const Primitive& p) {
   return (
@@ -570,6 +571,7 @@ void compile_fuse(
 
     std::function<void(const array&, int, const Stream&, const Shape&)> recurse;
     std::unordered_set<uintptr_t> cache;
+    std::unordered_set<uintptr_t> input_set;
     recurse = [&](const array& a,
                   int depth,
                   const Stream& s,
@@ -587,6 +589,8 @@ void compile_fuse(
       if (depth >= max_compile_depth || !a.has_primitive() ||
           a.primitive().stream() != s || !is_fusable(a.primitive()) ||
           (output_map.find(a.id()) != output_map.end() && a.shape() != shape)) {
+        // Possible input
+        input_set.insert(a.id());
         return;
       }
 
@@ -607,9 +611,20 @@ void compile_fuse(
       // Arrays with a mix of parents outside the compilable section
       // are not fusable
       if (!all_parents_in) {
+        // Possible input
+        input_set.insert(a.id());
         return;
       }
 
+      if (output_map.find(a.id()) != output_map.end()) {
+        input_set.insert(a.id());
+      } else {
+        // Not an input anymore since fusing it
+        input_set.erase(a.id());
+      }
+      if (input_set.size() >= max_compile_arrays) {
+        return;
+      }
       cache.insert({a.id()});
 
       for (auto& in : a.inputs()) {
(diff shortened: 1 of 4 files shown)


angeloskath (approved):
🙏

PR #1889, opened 2025-02-20
Allow non-square lu

### LINKED ISSUE #1886: Support non-square matrices LU decomposition
Could it be possible to support non square matrices for LU decomposition in order to add it as part of Keras MLX backend support?
Currently:
`ValueError: [linalg::lu] Only defined for square matrices.`

awni:
Allow non-square LU factorization.

Closes #1886


barronalex (approved):
Looks great!

PR #1892, opened 2025-02-20
fix version and expose command queue getter

awni:
Fixes the versions in the header so that downstream code can actually use them without setting the defines (which kind of defeats the purpose). Also they broke compiling with MLX as a library without setting the defines.

- Swap where we defining C++ version to version.h and reading it in CMake
- Use a compiled `version()` rather than inline in the header

Exposes command queue for setting properties (note this is an internal MLX API). It can be used like so:
```C++
  auto stream = default_stream(mx::Device::gpu);
  auto& d = mx::metal::device(mx::Device::gpu);
  MTL::CommandQueue* q = d.get_queue(stream);

  // Set some stuff on q here.

  return 0;
}
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,8 +1,14 @@
 cmake_minimum_required(VERSION 3.25)
 
 if(NOT MLX_VERSION)
-  set(MLX_VERSION 0.23.1)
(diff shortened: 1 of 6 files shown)


angeloskath (approved):
Nice thanks!

Out of curiosity when do we need access to the queue directly?

PR #1894, opened 2025-02-22
Adds nuclear norm support

### LINKED ISSUE #1791: [Feature] Singular value norms in `linalg::matrix_norm`
Could we get support for singular value norms implemented in `linalg::matrix_norm`? (via `linalg::norm`)

Thank you!

abeleinin:
## Proposed changes

resolves #1791 by adding support for nuclear norm to `mx.linalg.norm`.

approach adds `compute_uv` parameter to `mx.linalg.svd` following the [np.linalg.svd](https://numpy.org/doc/2.2/reference/generated/numpy.linalg.svd.html) docs. this removes unnecessary calculation of `U` and `Vt` components improving nuclear norm efficiency.

### DIFF
--- a/mlx/linalg.cpp
+++ b/mlx/linalg.cpp
@@ -112,8 +112,16 @@ inline array matrix_norm(
   if (ord == "f" || ord == "fro") {
     return l2_norm(a, axis, keepdims, s);
   } else if (ord == "nuc") {
-    throw std::runtime_error(
-        "[linalg::norm] Nuclear norm not yet implemented.");
+    auto a_moveaxis = moveaxis(a, axis[1], -1, s);
+    a_moveaxis = moveaxis(
+        a_moveaxis, (axis[0] > axis[1] ? axis[0] - 1 : axis[0]), -2, s);
+    auto nuclear_norm = sum(svd(a_moveaxis, false, s).at(0), -1, false, s);
+    if (keepdims) {
+      std::vector<int> sorted_axes(axis);
+      std::sort(sorted_axes.begin(), sorted_axes.end());
+      nuclear_norm = expand_dims(nuclear_norm, sorted_axes, s);
+    }
+    return nuclear_norm;
   } else {
     std::ostringstream msg;
     msg << "[linalg::norm] Invalid ord value '" << ord << "' for matrix norm.";
@@ -210,7 +218,8 @@ std::pair<array, array> qr(const array& a, StreamOrDevice s /* = {} */) {
   return std::make_pair(out[0], out[1]);
 }
 
-std::vector<array> svd(const array& a, StreamOrDevice s /* = {} */) {
+std::vector<array>
+svd(const array& a, bool compute_uv /* = true */, StreamOrDevice s /* = {} */) {
   check_cpu_stream(s, "[linalg::svd]");
   if (a.dtype() != float32) {
     std::ostringstream msg;
@@ -230,13 +239,21 @@ std::vector<array> svd(const array& a, StreamOrDevice s /* = {} */) {
   const auto n = a.shape(-1);
   const auto rank = a.ndim();
 
-  auto u_shape = a.shape();
-  u_shape[rank - 2] = m;
-  u_shape[rank - 1] = m;
-
   auto s_shape = a.shape();
   s_shape.pop_back();
   s_shape[rank - 2] = std::min(m, n);
+
+  if (!compute_uv) {
+    return array::make_arrays(
+        {s_shape},
+        {a.dtype()},
+        std::make_shared<SVD>(to_stream(s), compute_uv),
+        {a});
+  }
+
+  auto u_shape = a.shape();
+  u_shape[rank - 2] = m;
+  u_shape[rank - 1] = m;
 
   auto vt_shape = a.shape();
   vt_shape[rank - 2] = n;
@@ -245,7 +262,7 @@ std::vector<array> svd(const array& a, StreamOrDevice s /* = {} */) {
   return array::make_arrays(
       {u_shape, s_shape, vt_shape},
       {a.dtype(), a.dtype(), a.dtype()},
-      std::make_shared<SVD>(to_stream(s)),
+      std::make_shared<SVD>(to_stream(s), compute_uv),
       {a});
 }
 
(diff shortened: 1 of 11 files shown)


mlx/linalg.cpp
```diff
@@ -112,8 +112,16 @@ inline array matrix_norm(
...
-    throw std::runtime_error(
-        "[linalg::norm] Nuclear norm not yet implemented.");
+    auto a_moveaxis = moveaxis(a, axis[1], -1, s);
+    a_moveaxis = moveaxis(
+        a_moveaxis, (axis[0] > axis[1] ? axis[0] - 1 : axis[0]), -2, s);
+    auto nuclear_norm = sum(svd(a_moveaxis, false, s).at(0), -1, false, s);
+    if (keepdims) {
+      std::vector<int> sorted_axes(axis);
+      std::sort(sorted_axes.begin(), sorted_axes.end());
+      nuclear_norm = expand_dims(nuclear_norm, sorted_axes, s);
```
angeloskath, line 122: Similar to the above this can be

```cpp
std::vector<int> sorted_axes = (row < col) ? {row, col} : {col, row};
```

tests/linalg_tests.cpp
```diff
@@ -224,6 +227,14 @@ TEST_CASE("[mlx.core.linalg.norm] string ord") {
...
             array({14.28285686, 39.7617907}))
             .item<bool>());
+  CHECK(allclose(
+            norm(x, "nuc", std::vector<int>{0, 1}, false, Device::cpu),
+            array({25.045408, 26.893724, 28.831797}))
+            .item<bool>());
+  CHECK(allclose(
+            norm(x, "nuc", std::vector<int>{1, 2}, false, Device::cpu),
+            array({15.491934, 40.211937}))
+            .item<bool>());
```
angeloskath, line 237: Maybe add a test with negative axes?

mlx/linalg.cpp
```diff
@@ -112,8 +112,16 @@ inline array matrix_norm(
   if (ord == "f" || ord == "fro") {
     return l2_norm(a, axis, keepdims, s);
   } else if (ord == "nuc") {
-    throw std::runtime_error(
-        "[linalg::norm] Nuclear norm not yet implemented.");
+    auto a_moveaxis = moveaxis(a, axis[1], -1, s);
+    a_moveaxis = moveaxis(
+        a_moveaxis, (axis[0] > axis[1] ? axis[0] - 1 : axis[0]), -2, s);
```
angeloskath, line 117: I think the following is a little more readable and works with negative axes as well.

```cpp
auto row = (axis[0] < 0) ? axis[0] + a.ndim() : axis[0];
auto col = (axis[1] < 0) ? axis[1] + a.ndim() : axis[1];
auto a_matrix = (row > col)
    ? moveaxis(moveaxis(a, row, -1, s), col, -1, s)
    : moveaxis(moveaxis(a, col, -1, s), row, -2, s);
```

angeloskath: I don't think we should care but obviously we merge the two `moveaxis` into one `transpose` if it matters in the future.

mlx/linalg.cpp
```diff
@@ -230,22 +239,30 @@ std::vector<array> svd(const array& a, StreamOrDevice s /* = {} */) {
...
   auto s_shape = a.shape();
   s_shape.pop_back();
   s_shape[rank - 2] = std::min(m, n);
 
+  if (!compute_uv) {
+    return array::make_arrays(
+        {s_shape},
+        {a.dtype()},
+        std::make_shared<SVD>(to_stream(s), compute_uv),
+        {a});
```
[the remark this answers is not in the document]
abeleinin: Sorry, maybe this is supposed to be inferred by your comment, but did you want the return type of `svd` to be `std::variant<array, std::vector<array>>`?

abeleinin:
Thanks for the comments!

I added spectral norm support and more test coverage for vmap. The only comment I had a question about was the return type of `svd` since it's currently a `std::vector<array>` in the cpps. Maybe `std::variant` makes more sense, just let me know and I'll update the pr.

python/src/linalg.cpp
```diff
@@ -226,10 +231,12 @@ void init_linalg(nb::module_& parent_module) {
             a (array): Input array.
             stream (Stream, optional): Stream or device. Defaults to ``None``
               in which case the default stream of the default device is used.
+            compute_uv (bool, optional): If ``True``, return the ``U``, ``S``, and ``Vt`` components.
+              If ``False``, return only the ``S`` array. Default: ``True``.
```
awni, line 235: Nit: move this above `stream`.

mlx/linalg.cpp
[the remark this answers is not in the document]
angeloskath: Oops missed the braces.

No I wouldn't change the return type. It's a nitpicky optimization cause `make_arrays` tries to set up the siblings so it does some small extra things that are unnecessary. I am completely fine if you left it with `make_arrays`.

abeleinin:
Sorry for the delay on this review. There's some inconsistency in the float32 precision of the `svd` op between numpy, torch, and mlx. Here's some example code that illustrates this:
```py
import torch
import numpy as np
import mlx.core as mx

a = np.arange(18, dtype=np.float32).reshape((2,3,3))

print("numpy:", np.linalg.svd(a, compute_uv=False))
print("torch:", torch.svd(torch.tensor(a), compute_uv=False)[1])
print("mlx  :", mx.linalg.svd(mx.array(a), compute_uv=False, stream=mx.cpu))

ord = -2
axis = (1,2)
print("numpy: ", np.linalg.norm(a, ord=ord, axis=axis))
print("torch: ", torch.linalg.norm(torch.tensor(a), ord=ord, axis=axis))
print("mlx  : ", mx.linalg.norm(mx.array(a), ord=ord, axis=axis, stream=mx.cpu))
```

```
numpy: [[1.4226707e+01 1.2652260e+00 4.9790280e-16]
 [3.9759212e+01 4.5272526e-01 7.0096280e-16]]
torch: tensor([[1.4227e+01, 1.2652e+00, 4.3978e-08],
        [3.9759e+01, 4.5272e-01, 2.9630e-07]])
mlx  : array([[14.2267, 1.26523, 3.78331e-08],
       [39.7592, 0.452724, 2.65557e-07]], dtype=float32)
```

```
numpy:  [4.979028e-16 7.009628e-16]
torch:  tensor([4.3978e-08, 2.9630e-07])
mlx  :  array([3.78331e-08, 2.65557e-07], dtype=float32)
```

I replaced the expected values with what we see from numpy and tweaked the `rtol` value for that unit test.

angeloskath (approved):

PR #1897, opened 2025-02-23
Fix cpu compile

awni:
In some cases CPU-only compile was broken on macOS (e.g. when building C++ from source without python).

Tracked this down to different behavior from `/usr/bin/cc` and `/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/ ` (which is quite odd since they should be the same).

But explicitly specifying the flags is better anyway since it shrinks the size of the compiled preamble.


angeloskath (approved):
👍

PR #1898, opened 2025-02-24
Fix leaking objc

awni:
Minor fix so that `OBJC_DEBUG_MISSING_POOLS=YES` runs clean most of the time.


mlx/backend/metal/device.cpp
```diff
@@ -249,6 +249,7 @@ Device::~Device() {
   for (auto& l : library_map_) {
     l.second->release();
   }
+  stream_map_.clear();
```
awni, line 252: This needs to be cleared while the pool is in place.

angeloskath (approved):
🙏

PR #1902, opened 2025-02-25
Add a prepare tb ring script

angeloskath:
Adds a script to prepare a thunderbolt ring.

```shell
mlx.prepare_tb_ring --hosts host1,host2,host3,host4 --auto-setup --hostfile-output hostfile.json
```
The above should ssh to the hosts. Grab the connectivity per host and figure out any rings with a greedy algorithm to favor bigger rings. It will then disable the bridge interface and configure unique subnetworks in pairs. Finally it will save a hostfile to be used with `mlx.launch`.

```shell
mlx.prepare_tb_ring --hosts host1,host2,host3,host4 --dot
```
Does everything except start configuring the hosts. Prints a graphviz graph to visualize the ring topology.

```shell
mlx.prepare_tb_ring --hosts host1,host2,host3,host4
```
Instead of attempting to configure everything via ssh it prints a set of commands per node so we can just copy, paste and run on each node. Easier to deal with `sudo` etc.

A few questions/comments:
- Should it be a different script? I am not using much more than the `Host` dataclass and the corresponding `parse_hostlist` and `parse_hostfile`.
- The configuration is done using `ifconfig` which resets on reboot. This is on purpose to avoid making a mess accidentally. Worst case scenario someone reboots and everything should be back to normal.
- We are not making a lot of checks for the rings which may result in broken configuration. For instance if we have rings of different size or of different orientations, the configuration will be correct but the resulting hostfile won't be usable.
- The one thing I did do for normalizing the rings is to make sure that smaller ranks connect to larger ranks so `a -> b -> c -> a` will always form this circle regardless of interface order (ie regardless of which ports we connected).

### DIFF
--- a/python/mlx/distributed_run.py
+++ b/python/mlx/distributed_run.py
@@ -133,6 +133,11 @@ def log_warning(*args, **kwargs):
     print("\033[33m[WARN]", *args, "\033[0m", **kwargs)
 
 
+def log_error(*args, **kwargs):
+    kwargs["file"] = sys.stderr
+    print("\033[31m[ERROR]", *args, "\033[0m", **kwargs)
+
+
 def parse_hostfile(parser, hostfile):
     """Parse the json hostfile that contains both the hostnames to ssh into and
     the ips to communicate over when using the ring backend.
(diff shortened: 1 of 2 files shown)


awni (approved):
I think it's fine to keep it in the same script. It might be nice to have a more flexible naming strategy for these scripts.. though maybe it's not necessary yet:

- `mlx.prepare --backend "ring"` (or maybe `distributed.prepare` / `distributed_prepare`

Fits nicely with `mlx.launch --backend "ring"`. 

If other backends have need of a configuration / preparation step it could be worth-while.

angeloskath:
Yeah I agree the `prepare_tb_ring` is kinda lame as a name. Otoh it isn't a prepare ring because the ring backend works over Ethernet and even when the connectivity is not a ring. It is really a prepare thunderbolt ring.

I think it makes sense to name the entry point something like `mlx.distributed_config` or indeed `mlx.distributed_prepare` and then route to the tb ring configuration either by options or subcommands.

Wdyt?

awni:
I like both, maybe slight preference for mlx.distributed_config. 

How would the options differ if we wanted to use thunderbolt for MPI? I guess the script could take both the protocol and the backend as options?

angeloskath:
Yeah I think the following should be reasonable options (I will add routing but not implement them atm).

```
# Does what we do now
mlx.distributed_config --backend ring --over thunderbolt --hosts a,b,c

# Collects the IPs from the ethernet interface and generates a hostfile for mlx.launch
mlx.distributed_config --backend ring --over ethernet --hosts a,b,c

# Sets up the ring for MPI and generates rankfile needed by MPI to do ring reduce
mlx.distributed_config --backend mpi --over thunderbolt --hosts a,b,c

# Not sure this has something to do tbh
mlx.distributed_config --backend mpi --over ethernet --hosts a,b,c
```

PR #1906, opened 2025-02-26
Fix ring of 2 and allow scalars in API

angeloskath:
As the title suggests this makes 3 changes:

1. Fix the send/recv for a ring of size 2. This slipped the tests that check for a ring of 8 perhaps we should write tests for various ring sizes.
2. Change the `mx.distributed` ops to accept python values. This is a convenience when writing eval and/or training loops where we can gather or sum scalars across nodes with something like `mx.distributed.all_sum(loss) / mx.distributed.all_sum(count)` where the above are python scalars.
3. Enable TCP_NODELAY for the sockets that may improve latency for small all reductions and send/recv.

### DIFF
--- a/mlx/distributed/ring/ring.cpp
+++ b/mlx/distributed/ring/ring.cpp
@@ -3,6 +3,7 @@
 #include <arpa/inet.h>
 #include <fcntl.h>
 #include <netdb.h>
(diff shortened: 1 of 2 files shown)


awni (approved):
LGTM!

PR #1913, opened 2025-02-28
Fix slice data size

awni:
Fix data size computation for slice.


angeloskath (approved):
👍

Is there some specific case that we should add to the tests?

awni:
Sorry I got lazy and didn't add one. Forthcoming

PR #1915, opened 2025-02-28
RMS norm without scaling

angeloskath:
Allow rms norm without scaling and add function constants to reduce writes and memory usage if weights are not provided.

This started from https://github.com/ml-explore/mlx-examples/pull/1308 but I think it is generally a good idea.

Layernorm benchmark without weights.
```
Before: 20.87674 msec
After:  18.60866 sec
```

### DIFF
--- a/python/src/fast.cpp
+++ b/python/src/fast.cpp
@@ -25,12 +25,12 @@ void init_fast(nb::module_& parent_module) {
       "rms_norm",
       &mx::fast::rms_norm,
       "x"_a,
-      "weight"_a,
+      "weight"_a.none(),
       "eps"_a,
       nb::kw_only(),
       "stream"_a = nb::none(),
       nb::sig(
-          "def rms_norm(x: array, weight: array, eps: float, *, stream: Union[None, Stream, Device] = None) -> array"),
+          "def rms_norm(x: array, weight: Optional[array], eps: float, *, stream: Union[None, Stream, Device] = None) -> array"),
       R"pbdoc(
         Root Mean Square normalization (RMS norm).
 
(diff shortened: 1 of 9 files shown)


angeloskath:
Oh also we can wait to merge this after #1869 because this will be easier to rebase anyway.

python/src/fast.cpp
```diff
@@ -25,22 +25,22 @@ void init_fast(nb::module_& parent_module) {
...
 
         The normalization is with respect to the last axis of the input ``x``.
 
         Args:
             x (array): Input array.
-            weight (array): A multiplicative weight to scale the result by.
+            weight (array, optional): A multiplicative weight to scale the result by.
               The ``weight`` should be one-dimensional with the same size
-              as the last axis of ``x``.
+              as the last axis of ``x``. If set to ``None`` the no scaling happens.
```
awni, line 43:
```suggestion
              as the last axis of ``x``. If set to ``None`` then no scaling happens.
```

awni:
Very nice! And no need to wait for #1869.. I don't think the rebase will be too complicated.

One thing I'm wondering, on the off chance you happen to know, is what is the scope of a function constant? It seems like what you did works fine.. which suggests they are limited to the `.air` / compilation unit.

awni (approved):

angeloskath:
> One thing I'm wondering, on the off chance you happen to know, is what is the scope of a function constant? It seems like what you did works fine.. which suggests they are limited to the .air / compilation unit.

That is a very good point that I hadn't bothered to consider. According to the spec it seems that they are program level constants and we have 65k of them 🤔.

We hash by the function name though so in our case we will recreate the function setting the constant to the correct value each time. So presumably it doesn't matter and whether Metal will recompile depends on whether we asked the library with this set of function constants.

Given the above, my conclusion is that we should be fine and the 65k limit is only relevant to how many we can have set at any given point. But we need to be careful for kernels that use other constants and may be used as libraries.

PR #1914, opened 2025-02-28
Adding support for the Muon Optimizer

Goekdeniz-Guelmez:
## Proposed changes

First contribution to the MLX repo. Add the Muon optimizer to MLX's optimizer suite. Muon (MomentUm Orthogonalized by Newton-schulz) is a novel optimizer that combines momentum-based SGD with orthogonalization of parameter updates via Newton-Schulz iterations. This optimizer has shown promising results for training neural networks, particularly for convolutional and transformer architectures.
The implementation follows the approach described in [https://kellerjordan.github.io/posts/muon/ ](https://kellerjordan.github.io/posts/muon/), adapting it to MLX's framework. The optimizer performs standard SGD-momentum updates, followed by an orthogonalization step that replaces each 2D parameter's update with the nearest orthogonal matrix using an efficient Newton-Schulz iteration.
Key features of this implementation:
- Support for standard optimizer features (learning rate, momentum, weight decay, Nesterov)
- Efficient Newton-Schulz orthogonalization that works with bfloat16
- Special handling for parameters of different dimensions
- Appropriate scaling for non-square matrices

## Checklist

- [x] I have read the [CONTRIBUTING](https://github.com/ml-explore/mlx/blob/main/CONTRIBUTING.md) document
- [x] I have run `pre-commit run --all-files` to format my code / installed pre-commit prior to committing changes
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have updated the necessary documentation (if needed)

PR #1916, opened 2025-03-03
Add a multi optimizer

angeloskath:
It mainly adds a convenient way of using more than one optimizer for a single model based on the discussion in #1914.

```python
optimizer = MultiOptimizer(
    [MuonOptimizer(learning_rate=0.02), AdamW(learning_rate=0.001)],
    [lambda name, weight: weight.ndim >= 2]
)
# business as usual
```

As @stockeh demonstrated very nicely, in the issue linked above, it isn't something that is hard to do but it feels a pretty fundamental capability that we may want to make more convenient.

I am not sold on the `MultiOptimizer` approach I wrote (it was the simplest one) so feel free to suggest better strategies if you want.

### DIFF
--- a/python/mlx/optimizers/optimizers.py
+++ b/python/mlx/optimizers/optimizers.py
@@ -158,8 +158,10 @@ class MultiOptimizer(Optimizer):
     """Wraps a list of optimizers with corresponding weight predicates/filters
     to make it easy to use different optimizers for different weights.
 
-    The predicates take the full "path" of the weight and return True if it
-    should be considered for this optimizer.
+    The predicates take the full "path" of the weight and the weight itself and
+    return True if it should be considered for this optimizer. The last
+    optimizer in the list is a fallback optimizer and no predicate should be
+    given for it.
 
     Args:
         optimizers (list[Optimizer]): A list of optimizers to delegate to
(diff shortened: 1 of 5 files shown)


python/mlx/optimizers/optimizers.py
```diff
@@ -154,6 +154,75 @@ def _maybe_schedule(
...
+        filters (list[Callable[[str, array], bool]): A list of predicates that
+            should be one less than the provided optimizers.
+    """
+
+    def __init__(self, optimizers, filters: list = []):
+        super().__init__()
+        self._state = {}
+
+        if len(filters) != len(optimizers) - 1:
+            raise ValueError(f"Given {len(filters)} but {len(optimizers)-1} needed.")
```
awni, line 175:
```suggestion
            raise ValueError(f"Given {len(filters)} filters but {len(optimizers)-1} needed.")
```

awni (approved):
Looks great!

python/mlx/optimizers/optimizers.py
```diff
@@ -154,6 +154,75 @@ def _maybe_schedule(
         self.state[name] = param
 
 
+class MultiOptimizer(Optimizer):
+    """Wraps a list of optimizers with corresponding weight predicates/filters
+    to make it easy to use different optimizers for different weights.
+
+    The predicates take the full "path" of the weight and return True if it
+    should be considered for this optimizer.
```
awni, line 162: I would add a line here like:

> The last optimizer in the list is the fallback optimizer and no predicate should be given for it.

PR #1920, opened 2025-03-03
Allow debugging in distributed mode

angeloskath:
I hope this script isn't becoming too complicated but this PR broadcasts the stdin to all the processes. This allows debugging distributed jobs. Basically we can simply `import pdb; pdb.set_trace();` and then all processes will stop and wait for input that will be broadcasted to all.

I find it pretty convenient for testing.

The main thing is to switch from a bash monitoring script to a python monitoring script such that we can call `exec(...)` instead of reading the script from `stdin`.

### DIFF
--- a/python/mlx/distributed_run.py
+++ b/python/mlx/distributed_run.py
@@ -14,6 +14,8 @@ import time
 from collections import Counter
 from dataclasses import dataclass
 from pathlib import Path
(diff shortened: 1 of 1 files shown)


awni (approved):
LGTM thanks for adding this. This will be super useful!

PR #1921, opened 2025-03-04
Implement Weight Normalization, addressing issue #1888

### LINKED ISSUE #1888: Feature Request: Add Weight Normalization Support (weight_norm)
MLX currently lacks built-in support for weight normalization, which is a crucial feature for various deep learning architectures, particularly in audio processing and generative models. Weight normalization is a reparameterization technique that decouples the magnitude and direction of weight vectors, often leading to better conditioning and faster convergence.
Current Situation:

- No built-in equivalent to PyTorch's torch.nn.utils.weight_norm
- Users need to implement custom solutions, which may not be optimal or consistent

Proposed Solution:
I've developed a reference implementation that could serve as a starting point:

```python
import mlx.core as mx
import numpy as np
from typing import Optional, List, Union, Tuple

def compute_norm(x: mx.array, 
                p: int, 
                dim: Optional[Union[int, List[int]]] = None, 
                keepdim: bool = False) -> mx.array:
    """
    Compute the p-norm of a tensor along specified dimensions.
    
    Args:
        x: Input array
        p: Order of the norm (1 or 2)
        dim: Dimension(s) along which to compute the norm
        keepdim: Whether to keep the reduced dimensions
    
    Returns:
        MLX array containing the computed norm
    """
    if p not in [1, 2]:
        raise ValueError("Only p-norms with p of 1 or 2 are supported")
    
    # Handle dimension input
    if dim is None:
        dim = tuple(range(x.ndim))
    elif isinstance(dim, int):
        dim = (dim,)
    
    if p == 1:
        # L1 norm
        return mx.sum(mx.abs(x), axis=dim, keepdims=keepdim)
    else:
        # L2 norm
        return mx.sqrt(mx.sum(x * x, axis=dim, keepdims=keepdim))

def weight_norm(weight_v: mx.array, 
                weight_g: mx.array, 
                dim: Optional[int] = None) -> mx.array:
    """
    Applies weight normalization to the input tensor.
    
    Weight normalization reparameterizes weight vectors in a neural network 
    as a magnitude scalar times a direction vector: w = g * v/||v||
    
    Args:
        weight_v: Weight direction tensor (v)
        weight_g: Weight magnitude tensor (g)
        dim: Dimension along which to normalize. If None, normalize over all dims
            except dim=-1
    
    Returns:
        Normalized weight tensor
    """
    rank = len(weight_v.shape)
    
    if dim is not None:
        # Adjust negative dim
        if dim < -1:
            dim += rank
            
        # Create list of axes to normalize over
        axes = list(range(rank))
        if dim != -1:
            axes.remove(dim)
    else:
        # Default behavior: normalize over all dimensions
        axes = list(range(rank))
    
    # Compute L2 norm of v along specified axes
    norm_v = compute_norm(weight_v, p=2, dim=axes, keepdim=True)
    
    # Normalize and scale by g: w = g * (v / ||v||)
    normalized_weight = weight_v / (norm_v + 1e-7)  # Add epsilon for numerical stability
    return normalized_weight * weight_g

# Example usage:
def test_weight_norm():
    # Create sample tensors
    v = mx.random.normal((64, 3, 3))  # Direction tensor
    g = mx.random.normal((64, 1, 1))  # Magnitude tensor
    
    # Apply weight normalization
    w = weight_norm(v, g, dim=0)
    
    # Verify shape
    assert w.shape == v.shape
    
    # Verify norm along specified dimension
    norm_w = compute_norm(w, p=2, dim=[1, 2], keepdim=True)
    mx.eval(norm_w)  # Force computation
    
    return w, norm_w

if __name__ == "__main__":
    normalized_weight, weight_norm = test_weight_norm()
```

cavit99:
## Proposed changes

This PR implements weight normalization for MLX, addressing issue #1888. Weight normalization is a reparameterization technique that decouples the magnitude of a weight tensor from its direction, making optimization more efficient by improving the conditioning of the optimization problem. Is particularly important for audio processing, among other applications.

## Key Features

- Core C++ implementation of `mx.weight_norm` with optimized paths for different dimensions
- Python module `weight_norm.py` with user-friendly API and layer wrappers
- Proper handling of MLX's channel ordering differences from PyTorch
- Workaround for the `linalg::norm` 2-axes limitation
- Convenience classes for common layer types (Linear, Conv1d, Conv2d)
- Comprehensive test suite validating mathematical properties and cross-framework compatibility

### Core C++ Implementation

The core `weight_norm` operation is implemented with three different paths based on the number of axes to normalize over:

1. **Direct path** for 1-2 axes using optimized `linalg::norm` kernels
2. **Reshape-based approach** for >2 axes, which:
   - Identifies dimensions to keep vs. normalize
   - Handles special cases: normalizing all dims, keeping one dim, keeping multiple dims
   - Reshapes appropriately to leverage the optimized 2D norm kernel
   - Reshapes results back for broadcasting

### Python Layer

The Python implementation:
- Provides a `weight_norm` function that wraps MLX modules
- Handles dimension ordering differences for different layer types
- Computes initial g parameter as the norm of the original weight
- Overrides the module's forward pass to apply weight normalization on-the-fly
- Includes convenience classes (WeightNormLinear, WeightNormConv1d, WeightNormConv2d)

## Testing and Verification

Testing follows a comprehensive two-pronged approach:

### 1. Mathematical Property Tests
- Verify that the normalized weights have the correct norm (equals g)
- Confirm that the direction of normalized weights matches v
- Validate that changing g correctly scales the weight norms
- Test edge cases like normalizing over all dimensions

### 2. Cross-Framework Verification
- Compare against PyTorch's weight normalization
- Test both independent implementations and direct weight transfer
- Document expected differences between frameworks and how to achieve exact equivalence

### 3. Performance Benchmarks
download
Benchmarked on Apple M3 Max shows MLX outperforms PyTorch MPS:
- Linear layers (1 axis): 4.90x-5.26x speedup
- Conv1d layers (2 axes): 1.46x-2.05x speedup  
- Conv2d layers (3 axes): 1.50x-1.76x speedup

### Core API
```python
import mlx.core as mx

# Create parameters
v = mx.random.normal((64, 3, 3))  # Direction tensor
g = mx.random.normal((64, 1, 1))  # Magnitude tensor

# Apply weight normalization
w = mx.weight_norm(v, g, axes=[1, 2])
```

### Module API
```python
import mlx.nn as nn
from mlx.nn.layers.weight_norm import weight_norm

# Apply to existing layer
linear = nn.Linear(10, 20)
linear_wn = weight_norm(linear)

# Use convenience class
conv1d_wn = nn.WeightNormConv1d(16, 32, kernel_size=3)
```
Resolves #1888.

## Checklist

- [X ] I have run `pre-commit run --all-files` to format my code / installed pre-commit prior to committing changes
- [ X] I have updated the necessary documentation (if needed)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -85,3 +85,5 @@ build/
 
 # Jetbrains
 .cache
+build*/
+*.dylib
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -29,7 +29,7 @@ set(CMAKE_INSTALL_MESSAGE NEVER)
 # ----------------------------- Configuration -----------------------------
 option(MLX_BUILD_TESTS "Build tests for mlx" ON)
 option(MLX_BUILD_EXAMPLES "Build examples for mlx" ON)
-option(MLX_BUILD_BENCHMARKS "Build benchmarks for mlx" OFF)
+option(MLX_BUILD_BENCHMARKS "Build benchmarks for mlx" ON)
 option(MLX_BUILD_PYTHON_BINDINGS "Build python bindings for mlx" OFF)
 option(MLX_BUILD_METAL "Build metal backend" ON)
 option(MLX_BUILD_CPU "Build cpu backend" ON)
--- a/mlx/ops.cpp
+++ b/mlx/ops.cpp
@@ -11,6 +11,7 @@
 #include <sstream>
 
 #include "mlx/fast.h"
+#include "mlx/linalg.h"
 #include "mlx/ops.h"
 #include "mlx/primitives.h"
 #include "mlx/transforms.h"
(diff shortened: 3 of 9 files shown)


Blaizzy:
Thanks a lot @cavit99, this is great work!

One tiny nit:
- Could we change the weight naming to weight_g and weight_v? Makes easier to map from torch and remember.

Blaizzy:
As far as I can tell from intial testing.

This PR does address my issues:

Screenshot 2025-03-04 at 7 11 59 PM

cavit99:
> Thanks a lot @cavit99, this is great work!
> 
> One tiny nit:
> 
> * Could we change the weight naming to `weight_g` and `weight_v`? Makes easier to map from torch and remember.

agreed from my side, so I pushed that change to the PR, thank you

Blaizzy:
Perfect! 🤩

Now we wait for @awni :)

awni:
I'm not certain about including this as `WeightNorm` as I thought `WeightNorm` is not used so much anymore.. thoughts?

Either way we should not make free functions in C++ and Python for this. It should just be a layer  in `mlx.nn`.

awni:
@cavit99 are you planning coming back to this PR?

Blaizzy:
@cavit99 I'm porting SparkTTS and this would be a great addition :)

Turns out 90% of the audio models we ported in MLX-Audio use WeightNorm layer either in the model or the codec.

cavit99:
sorry I will do this by EOD tomorrow

awni:
I'm going to close this as inactive. We're open to revisiting the addition of weight norm in the future.

PR #1922, opened 2025-03-04
SDPA support for small batch (over sequence) queries

awni:
Benchmarks on M4 max with the following config:

```
L = 16384
Lq = 4
H = 32
H_k = H // 4
D = 128
V = 128
```

Timing sdpa ... 4.70581 msec
Timing attention ... 16.99894 msec

Also updated RoPE to route to a faster path for shapes like `[1, H, L, D]`.

The intention here is mostly to speed up spec dec. Will share some benchmarks here: https://github.com/ml-explore/mlx-examples/pull/1319

### DIFF
--- a/mlx/backend/metal/kernels/sdpa_vector.h
+++ b/mlx/backend/metal/kernels/sdpa_vector.h
@@ -18,9 +18,11 @@ template <typename T, int D, int V = D>
     const constant size_t& v_stride,
     const constant float& scale,
     const device bool* mask [[function_constant(has_mask)]],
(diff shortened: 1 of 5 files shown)


angeloskath (approved):
Awesome!

The choice for `query_transposed` vs passing strides and so on was for performance reasons?

awni:
> The choice for query_transposed vs passing strides and so on was for performance reasons?

Good question.. I didn't try passing the query strides in so I couldn't say if there is much perf difference, probably minor.. the diff seemed simpler with the function constant. On the other hand it could be more general to allow arbitrary strides in the sequence and head dimension.

PR #1925, opened 2025-03-05
Affine quant always in fp32

barronalex:
Change the affine quant to always use fp32 for the scale/bias calculation.

This gets rid of the looping mentioned in this [thread](https://www.reddit.com/r/ollama/comments/1j1e5k8/for_mac_users_ollama_is_getting_mlx_support/) for Llama 3.2 3B.

Test prompt:
```
mlx_lm.generate --temp 0 --max-tokens 256 --model "mlx-community/Llama-3.2-3B-Instruct-8bit" --prompt "List the ISO 3-letter codes of all countries in JSON." --max-tokens 1024
```

### DIFF
--- a/mlx/backend/cpu/quantized.cpp
+++ b/mlx/backend/cpu/quantized.cpp
@@ -543,8 +543,8 @@ void quantize(
   T* scales = scales_.data<T>();
   T* biases = biases_.data<T>();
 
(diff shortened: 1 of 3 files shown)


awni (approved):
Very nice!! Thanks for updating the CPU + fallback as well, that's great.

PR #1926, opened 2025-03-05
Stop fence term

angeloskath:
Make sure that we don't leave any kernels waiting for a value that won't be updated by updating in the destructor.

Also fix `update_gpu` to only ever increase the fence value.


angeloskath:
I am closing this (at least for now). Shared pointers won't necessarily be destroyed when SIGTERM is signaled and there isn't much point beyond that.

PR #1928, opened 2025-03-05
Only fail when 10 consecutive socket errors occur

angeloskath:
Allow some send/recv errors to occur before failing. Also don't fail in `distributed_config` if we have found at least 1 ring.


awni (approved):
LGTM!

Issue #1785, opened 2025-01-21
[FEATURE] `mx.repeat` with variable number of repetitions


kyrollosyanny:
**Describe the bug**
In numpy, ```np.repeat``` can accept both int and array type for the repetition. The main use case for passing an array is if each element in the original array needs to repeated a different number of times. Currently, this fails in MLX.  

**To Reproduce**
In numpy:

```
a=np.array([1,2,3,4])

print(np.repeat(a,a))

[1 2 2 3 3 3 4 4 4 4]
``` 

in MlX, this does not work since repetition needs to be an integer. 

It would be great if we can add this behavior to ```mx.repeat``` unless there is another efficient way to do this. 

Thanks

ThomasLin617:
Hi! I’m new here and would like to take this issue.

Plan: add vector-length support for `mx.repeat` via a small pure-Python fallback with a basic pytest.  
Let me know if that’s okay. Thanks!

PR #1935, opened 2025-03-06
`mx.repeat` with variable number of repetitions, issue #1785

kisljak:
## Proposed changes

Fixes https://github.com/ml-explore/mlx/issues/1785

Implementation of `mx.repeat` for variable number of repetitions. I also updated checking of axis in a method `tile`, which was only using duplicit code already implemented elsewhere.

Finally I added tests to check the new functionality

## Checklist


angeloskath:
MLX doesn't support ops with a data dependent output size yet unfortunately. This op assumes that the repeats is a vector available when called so it would not solve the linked issue which requires an array as the repeats argument.

I will close it but feel to post on the issue for more details.

PR #1938, opened 2025-03-06
always use json

awni:
`ring.cpp` depends on JSON as well so we simply always include it

Fixes the build when you disable safetensors


angeloskath (approved):
🙏

PR #1939, opened 2025-03-07
Fix unreachable warning

chunyang-wen:
## Proposed changes

The editor reports code is not reachable. We are using a Literal type and it will introduce static type check. It makes the lint confusing I think.

iShot_2025-03-07_20 40 27

- This change also removes the duplicate string collection for `Reduction`.

```python
>>> z = f"Invalid reduction. Must be one of {get_args(Reduction)}."
>>> z
"Invalid reduction. Must be one of ('none', 'mean', 'sum')."
```
- Fail fast might be better than we compare several times and finally it fails.
## Checklist


awni (approved):
Looks good, thanks!

chunyang-wen:
@awni Please help merge the PR. I have no right to merge. Thanks!

Issue #1918, opened 2025-03-03
[BUG] Code snippet runs 4x slower since mlx 0.19.0


gabrieldemarmiesse:
**Describe the bug**

When developping our model, we noticed that changing the version of mlx had a huge impact on performance, notably that mlx 0.18.1 was faster than all mlx versions released afterwards.

**To Reproduce**

Here is a minimal reproducible example:
```python
# /// script
# requires-python = "==3.12.9"
# dependencies = [
#     "mlx==0.23.1",
# ]
# ///
import time

import mlx.core as mx
import mlx.nn as nn


def something(kv):
    xs = mx.zeros(shape=(1, 32, 1, 128), dtype=mx.bfloat16)
    for _ in range(128):
        xs = xs + mx.fast.scaled_dot_product_attention(xs, kv, kv, scale=0.1)
    return xs


def main():
    WARMUP = 5
    TOTAL_STEPS = 40
    mx.random.seed(299792458)
    
    kv = mx.zeros(shape=(1, 1024, 4096), dtype=mx.bfloat16)
    kv = kv.reshape(1, -1, 2, 32, 128)
    kv = kv[:, :, 0].transpose(0, 2, 1, 3)
    mx.eval(kv)

    sum_times = 0
    for i in range(TOTAL_STEPS):
        t1 = time.time()
        mx.eval(something(kv))
        t2 = time.time()
        if i >= WARMUP:
            sum_times += t2 - t1

    print(f"average time per step: {(sum_times / (TOTAL_STEPS - WARMUP)) * 1000:1f} ms")

main()
```

```bash
gabrieldemarmiesse@Kyutais-Laptop V2LLMs % uv add --script bench.py mlx==0.17.1 && uv run bench.py
average time per step: 10.089261 ms
gabrieldemarmiesse@Kyutais-Laptop V2LLMs % uv add --script bench.py mlx==0.18.1 && uv run bench.py
average time per step: 9.984575 ms
gabrieldemarmiesse@Kyutais-Laptop V2LLMs % uv add --script bench.py mlx==0.19.0 && uv run bench.py
average time per step: 44.592115 ms
gabrieldemarmiesse@Kyutais-Laptop V2LLMs % uv add --script bench.py mlx==0.19.3 && uv run bench.py
average time per step: 44.662435 ms
gabrieldemarmiesse@Kyutais-Laptop V2LLMs % uv add --script bench.py mlx==0.20.0 && uv run bench.py
average time per step: 44.754716 ms
[...]

PR #1950, opened 2025-03-10
Support transposed head/seq for kv

awni:
Closes: #1918 

Benchmark on M2 Ultra:

```
L = 16384
H = 32
H_k = H // 4
D = 128
V = 128
dtype = mx.float16
```

Pre:
Timing sdpa ... 3.31429 msec

Post:
Timing sdpa ... 1.66841 msec

Benchmark in  #1918 

Pre: average time per step: 5.838094 ms
Post: average time per step: 2.821009 ms


angeloskath (approved):
🙏

PR #1952, opened 2025-03-10
Reduce binary size

awni:
Reduce the binary size by moving the kernel dispatches to an outer level and adding back `unsafe_weak_copy`:

Binary size built with:

```
cmake .. \                                
  -DCMAKE_BUILD_TYPE=MinSizeRel \
  -DBUILD_SHARED_LIBS=ON \
  -DMLX_BUILD_CPU=ON \
  -DMLX_BUILD_SAFETENSORS=ON \
  -DMLX_BUILD_GGUF=OFF \
  -DMLX_METAL_JIT=ON
```

Pre: 35MB
Post: 12MB (potentially shrink more by updating a few more of the Primitives with the new pattern)


angeloskath (approved):
LGTM!

PR #1953, opened 2025-03-10
Fix copy for large arrays with JIT

### LINKED ISSUE #1951: [BUG] JIT looks for copyfloat32float32 but generates copyfloatfloat
**Describe the bug**

JIT looks for copyfloat32float32 but generates copyfloatfloat

**To Reproduce**

I am not sure the repro steps -- I see this via mlx-swift running a large video through Qwen2.5VL. It looks like the operation above this is a reshape.

This looks like it is caused by:

https://github.com/ml-explore/mlx/blob/3c3e558c603b4aba87ae1369e776b44a8cdf5a48/mlx/backend/metal/utils.cpp#L42

```
std::string type_to_name(const Dtype& t) {
  std::string tname;
  switch (t) {
...
    case float32:
      tname = "float32";
      break;
    case float64:
```

vs

https://github.com/ml-explore/mlx/blob/3c3e558c603b4aba87ae1369e776b44a8cdf5a48/mlx/backend/common/compiled.cpp#L49

```
std::string get_type_string(Dtype d) {
  switch (d) {
    case float32:
      return "float";
```

I think :-)

Seen in mlx v0.21.0 via mlx-swift.

awni:
Closes #1951

### DIFF
--- a/mlx/backend/metal/jit_kernels.cpp
+++ b/mlx/backend/metal/jit_kernels.cpp
@@ -177,7 +177,12 @@ MTL::ComputePipelineState* get_copy_kernel(
     kernel_source +=
         get_template_definition("s_" + lib_name, "copy_s", in_type, out_type);
     kernel_source +=
+        get_template_definition("s2_" + lib_name, "copy_s2", in_type, out_type);
+    kernel_source +=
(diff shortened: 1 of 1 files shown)


angeloskath (approved):
🙏

PR #1962, opened 2025-03-14
Use same accumulation precision in gemv as gemm

### LINKED ISSUE #1958: [BUG] Using Matrix Multiplication on Batched Arrays Leads to Different Results
**Describe the bug**
When doing matrix multiplication, increasing batch size leads to different results than a batch size of 1 when using non fp32 data types.

**To Reproduce**

Include code snippet
```python
a = mx.random.normal([1, 1, 256])
_b = mx.random.normal([1, 1, 256])
b = mx.concatenate([a, _b], axis=1)
c = mx.random.normal([256, 64])

print(a.astype(mx.bfloat16) @ c.astype(mx.bfloat16))
print(b.astype(mx.bfloat16) @ c.astype(mx.bfloat16))
```

Can swap the data types to be different. It seems fp32 is the only one that works.

**Expected behavior**
The first indexes of the results should be the same (as it is when using fp32), but are wildly different using other data types. Result from running the above snippet:

```
array([[[2.01562, -21.375, 7.6875, ..., 1.21875, -8.0625, 11.125]]], dtype=bfloat16)
array([[[1.89844, -21.25, 7.53125, ..., 1.26562, -8.0625, 11.125],
        [-2.3125, 17.375, -19, ..., 11.8125, 5.96875, 18]]], dtype=bfloat16)
```

The correct answer from running in fp32 is:

```
array([[[1.89315, -21.2555, 7.5196, ..., 1.25635, -8.06082, 11.1537]]], dtype=float32)
array([[[1.89315, -21.2555, 7.51959, ..., 1.25635, -8.06082, 11.1537],
        [-2.30559, 17.4162, -19.0379, ..., 11.7927, 6.01239, 18.0915]]], dtype=float32)
```

 - OS Version: MacOS 15.0
 - Version: 0.23.1

awni:
This changes the gemv to accumulate in a higher precision which matches the behavior of gemm.

I think this is the right thing to do in general and will close #1958 

The benchmarks are confusing and need some investigation though.

For BF16 it's noticeably slower but for fp16 the speed is unchanged 🤔 . In both cases the math should be easily hidden by loading time so the bf16 result doesn't make sense.

### DIFF
--- a/mlx/backend/metal/kernels/gemv.metal
+++ b/mlx/backend/metal/kernels/gemv.metal
@@ -23,7 +23,8 @@ template <
     const int SN, /* Simdgroup cols (in threads) */
     const int TM, /* Thread rows (in elements) */
     const int TN, /* Thread cols (in elements) */
-    const bool kDoAxpby> /* Do out = alpha * out + beta * bias */
+    const bool kDoAxpby, /* Do out = alpha * out + beta * bias */
+    typename AccT = float>
 struct GEMVKernel {
   MLX_MTL_CONST int threadsM = BM * SM;
   MLX_MTL_CONST int threadsN = BN * SN;
@@ -97,7 +98,7 @@ struct GEMVKernel {
       const constant float& alpha [[buffer(7)]],
       const constant float& beta [[buffer(8)]],
       const constant int& bias_stride [[buffer(14)]],
-      threadgroup T* tgp_memory [[threadgroup(0)]],
+      threadgroup AccT* tgp_memory [[threadgroup(0)]],
       uint3 tid [[threadgroup_position_in_grid]],
       uint3 lid [[thread_position_in_threadgroup]],
       uint simd_gid [[simdgroup_index_in_threadgroup]],
@@ -106,7 +107,7 @@ struct GEMVKernel {
     (void)lid;
 
     // Thread local accumulation results
-    thread T result[TM] = {0};
+    thread AccT result[TM] = {0};
     thread T inter[TN];
     thread T v_coeff[TN];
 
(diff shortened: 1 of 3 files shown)


awni:
Adding reviewers in case you spot something that I missed.. I'll share some benchmarks shortly.

barronalex:
This seems very reasonable to me but probably worth @jagrit06 taking a look before we merge.
BF16 is quite a few more ops than fp16 if it's emulated right?

awni:
I found a way to do this that doesn't result in much slow down for the important case `x @ W.T` and only slight slowdown for bf16 for the less important case (`x @ W`).

Type | Transposed | Time Pre | Time Post |
---- | ---- | ----- | -----
float16 | True |  2.772 (ms)| 2.778 (ms)
float16 | False | 8.019 (ms)| 8.025 (ms)
bfloat16 | True | 2.859 (ms)| 2.829 (ms)
bfloat16 | False |9.545 (ms)| 10.093 (ms)
float32 | True |  5.022 (ms)| 5.029 (ms)
float32 | False | 12.393 (ms)| 12.364 (ms)

angeloskath:
I hesitated writing this yesterday (cause I only skimmed through the changes) but it makes sense cause we only do the casting once instead of every time we do a multiplication. On the other hand we need twice the registers so it could have a big impact on smaller machines. Let me run a quick benchmark on my M2 air.

angeloskath:
I am running a square benchmark `1x4096 @ 4096x4096` but there is no difference on my laptop so I think it is good to go.

Before
```
Timing float32	x @ w	 ... 7.65121 msec
Timing float32	x @ w.T	 ... 7.88494 msec
Timing float16	x @ w	 ... 3.93161 msec
Timing float16	x @ w.T	 ... 3.98962 msec
Timing bfloat16	x @ w	 ... 3.97440 msec
Timing bfloat16	x @ w.T	 ... 3.97418 msec
```

After
```
Timing float32	x @ w	 ... 7.58956 msec
Timing float32	x @ w.T	 ... 7.77694 msec
Timing float16	x @ w	 ... 3.98002 msec
Timing float16	x @ w.T	 ... 3.93565 msec
Timing bfloat16	x @ w	 ... 3.95123 msec
Timing bfloat16	x @ w.T	 ... 4.01029 msec
```

PR #1966, opened 2025-03-16
Do not define MLX_VERSION globally

zcbenz:
A global MLX_VERSION define would trigger a recompilation of the whole project every time when bumping the version.

Also fix MLX_VERSION being empty when not building from `setup.py`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -9,6 +9,7 @@ if(NOT MLX_VERSION)
   string(REGEX MATCH "#define MLX_VERSION_PATCH ([0-9]+)" _ "${_mlx_h_version}")
   set(_patch ${CMAKE_MATCH_1})
   set(MLX_PROJECT_VERSION "${_major}.${_minor}.${_patch}")
+  set(MLX_VERSION ${MLX_PROJECT_VERSION})
 else()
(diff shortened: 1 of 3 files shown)


awni:
> A global MLX_VERSION define would trigger a recompilation of the whole project every time when bumping the version.

Do you mean downstream projects that use MLX as a library?

awni:
Wouldn't anyone that included `mlx/mlx.h` (which includes `mlx/version.h` still need to rebuild here? I'm not quite following what this is helping with.

zcbenz:
It is about the recompilation of MLX itself and its dependencies, like openblas when building from source.

When calling `setup.py` it sets MLX_VERSION to current day or current hash, which forces a recompilation of the whole project whenever there is a new commit since last call.

The thing annoys me most is the recompilation of openblas though, so a simpler alternative fix would be restricting MLX_VERSION to mlx itself to avoid recompilation of dependencies only.

awni:
> When calling setup.py it sets MLX_VERSION to current day or current hash, which forces a recompilation of the whole project whenever there is a new commit since last call.

😮 that's bad.. now that you mention it I noticed we started rebuilding from scratch every day 🤯  and I never thought it was because of that.

PR #1969, opened 2025-03-18
Fix multistream GPU deadlock

awni:
There is some chance for deadlock:
- GPU stream 1: does some work
- GPU stream 2: does work which depends on the first
- GPU stream 2: reaches its command queue buffer limit and waits on the main thread
- GPU stream 1: is not committed so it never signals to the second
- GPU stream 2: main thread is stuck because command buffers are stuck waiting on GPU stream 1

The fix here takes over the waiting from the Metal API and moves it into eval. I tried several options for this but most of the cleaner solutions resulted in some perf regression so here we are.

### DIFF
--- a/mlx/backend/cpu/encoder.h
+++ b/mlx/backend/cpu/encoder.h
@@ -39,13 +39,24 @@ struct CommandEncoder {
 
   template <class F, class... Args>
   void dispatch(F&& f, Args&&... args) {
+    num_ops_ = (num_ops_ + 1) % 10;
     auto task = std::bind(std::forward<F>(f), std::forward<Args>(args)...);
-    scheduler::enqueue(stream_, std::move(task));
+    if (num_ops_ == 0) {
+      scheduler::notify_new_task(stream_);
+      auto task_wrap = [s = stream_, task = std::move(task)]() mutable {
+        task();
+        scheduler::notify_task_completion(s);
+      };
+      scheduler::enqueue(stream_, std::move(task_wrap));
+    } else {
+      scheduler::enqueue(stream_, std::move(task));
+    }
   }
 
  private:
   Stream stream_;
   std::vector<array> temporaries_;
+  int num_ops_{0};
 };
 
 CommandEncoder& get_command_encoder(Stream stream);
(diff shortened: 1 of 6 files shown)


awni:
I ran a few benchmarks and didn't observe any perf regression. I tested on M2 Ultra and M3 Max a few sized LLMs (1B-32B).

angeloskath:
I may be sleep deprived but is this changing anything in the metal side of things? I feel I either need to read and think way more carefully or there is a part of the diff missing ...

awni:
Yea sorry it's a bit subtle. 

The deadlock happens when the main thread reaches the command buffer limit for stream 2 and waits. But previous command buffers in stream 2 are waiting on stream 1. The command buffer in stream 1 is not yet committed.

The fix is to increase the command buffer limit on the queue so it is effectively unused. That way the main thread never waits inside `eval_gpu` to get a command buffer. Instead the wait happens in the main eval loop where we already are committing command buffers in all streams.

The reason I changed the CPU task counting is because it was a simple way to reuse the `MAX_ACTIVE_TASKS` for both the CPU and GPU.

angeloskath:
Ok yeah I see. I like handling the high level logic in eval, that's nice.

For instance now a CPU task is 10 tasks which is somewhat symmetric to the GPU but it is encoded in a different place and I guess the reason was to get 10*10 =100 as it were before...

For the GPU, we have max ops per buffer, max size per buffer, max buffers per queue and max tasks. I am not saying we don't need all of them but it is hard to say for instance why 20 fixes the problem and 12 doesn't. We can have at most 10 buffers in flight after which `eval` should commit all streams and wait for one to finish so we would never get >10 uncommitted buffers right?

Sorry for the stupid questions, I am just trying to reduce the magic numbers especially if they have complex interactions across files and devices.

awni:
> We can have at most 10 buffers in flight after which eval should commit all streams and wait for one to finish so we would never get >10 uncommitted buffers right?

Exactly. So the `20` is just a placeholder for # > 10. I can check the default value and if it's large enough we can just remove it.

angeloskath (approved):
👍 Let's merge it then.

I would put the magic number 10 on the cpu side in a constexpr variable on top instead of having it as a magic number. If anything we can draw some parallel to the metal backend if we happen to look back at it in a few months or late at night 😅

PR #1970, opened 2025-03-19
interruptable eval

awni:
CC @davidkoski  wdyt?

### DIFF
--- a/mlx/transforms.cpp
+++ b/mlx/transforms.cpp
@@ -35,6 +35,11 @@ class Synchronizer : public Primitive {
   DEFINE_PRINT(Synchronize);
 };
 
+std::atomic<bool>& interrupt_flag() {
+  static std::atomic<bool> interrupt_{false};
+  return interrupt_;
+}
+
 // Initialize the static tracing members from transforms_impl.h
 //
 // These are used to implement the in_tracing() function the returns true if we
@@ -249,6 +254,12 @@ array eval_impl(std::vector<array> outputs, bool async) {
     if (!arr.is_tracer()) {
       arr.detach();
     }
+
+    if (interrupt_flag()) {
+      interrupt_flag() = false;
+      synchronizer.attach_event(Event{stream});
+      break;
+    }
   }
 
   // Signal the event in its stream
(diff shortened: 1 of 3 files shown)


mlx/transforms.cpp
```diff
@@ -263,6 +274,10 @@ array eval_impl(std::vector<array> outputs, bool async) {
   return synchronizer;
 }
 
+void interrupt_eval() {
+  interrupt_flag() = true;
+}
```
davidkoski, line 279: I think you need a way to clear this if you raced vs the eval.  E.g. you didn't hit line 259 above it will still be in `true` state on the next call to `eval()`

mlx/transforms.cpp
```diff
@@ -249,6 +254,12 @@ array eval_impl(std::vector<array> outputs, bool async) {
     if (!arr.is_tracer()) {
       arr.detach();
     }
+
+    if (interrupt_flag()) {
+      interrupt_flag() = false;
+      synchronizer.attach_event(Event{stream});
```
davidkoski, line 260: Does this leave the current work enqueued and cause the eval to return?  Or does it drain it (or is that not a thing)?

awni: Work that has already been enqueued will remain so. That includes CPU and GPU work. The GPU command buffers are all committed when this returns.

From https://developer.apple.com/documentation/metal/preparing-your-metal-app-to-run-in-the-background?language=objc we may have to wait for the command buffers to be scheduled.

mlx/transforms.cpp
awni: Yes good point. There is some subtlety here. I changed it to only set the interrupt flag if an eval is actually ongoing. If it's not ongoing then the flag is not set.

So if you call interrupt before you call eval on the other thread then the eval will keep going. I think this is kind of expected since there is nothing to interrupt and you don't want to interrupt future evals.  But it does introduce the possibility of a race. I added a return value to `interrupt_eval` so the caller can check if the eval they were expecting to interrupt was actually interrupted.

davidkoski: Seems reasonable

aPaleBlueDot:
👀  I'd greatly appreciate this feature, as it'll solve the app crashes.

zcbenz:
I'm closing this PR since the work will unlikely be continued.

Issue #1971, opened 2025-03-19
[BUG] nullptr reference in allocator::Buffer::raw_ptr


zcbenz:
Apply the patch below and run C++ tests, and you can see there are lots of nullptr references.

```patch
diff --git a/mlx/backend/metal/allocator.cpp b/mlx/backend/metal/allocator.cpp
index f2c95be2..4306d13d 100644
--- a/mlx/backend/metal/allocator.cpp
+++ b/mlx/backend/metal/allocator.cpp
@@ -20,6 +20,8 @@ Allocator& allocator() {
 }
 
 void* Buffer::raw_ptr() {
+  if (!ptr_)
+    fprintf(stderr, "raw_ptr is null\n");
   return static_cast<MTL::Buffer*>(ptr_)->contents();
 }
 
```

awni:
This comes up with empty arrays which are valid (and should have null `data_ptr`). Dereferencing the nullptr seems risky though :). Somehow getting the contents of `MTL::Buffer*` when it is nullptr is giving us the right result.. but I'm not sure if that's guaranteed or not.

Probably safer to change it to:

```c++
if (!ptr_) {
  return nullptr;
}
...
```

PR #1972, opened 2025-03-19
Guard nullptr dereference

awni:
Close https://github.com/ml-explore/mlx/issues/1971


zcbenz:
The cpu/allocator also needs the guard.

zcbenz:
Yes. It could return (nullptr + 1) and the caller can not determine whether the returned ptr is nullptr by comparing to nullptr.

angeloskath (approved):
Great, feel free to merge after guarding the CPU allocator as well.

PR #1974, opened 2025-03-20
Update smooth_l1_loss in losses.py

jiyzhang:
According the definition of smooth_l1_loss, the line 

`diff = predictions - targets`

Should be updated to 

`diff = mx.abs(predictions - targets)`

After the modification, the result is consistent with PyTorch smooth_l1_loss

## Proposed changes

## Checklist


angeloskath (approved):
Thanks for catching and fixing this.

PR #1975, opened 2025-03-20
Cross platform libmpi loading

### LINKED ISSUE #1964: [BUG] Failed to read `libmpi.so` on Linux container
**Describe the bug**

I am trying to run distributed MLX using MPI within Linux container.
However, the MLX fails due to this hardcoded value to read `libmpi.dylib` file: https://github.com/ml-explore/mlx/blob/main/mlx/distributed/mpi/mpi.cpp#L50-L53

For the Linux the actual file name is: `libmpi.so`. For example:
```
$ realpath libmpi.so

/usr/lib/aarch64-linux-gnu/openmpi/lib/libmpi.so
```

cc @awni @tenzen-y

**To Reproduce**

Include code snippet
```python
import mlx.core as mx

dist = mx.distributed.init(strict=True, backend="mpi")
x = mx.distributed.all_sum(mx.ones(10))
print(f"RANK: {dist.rank()}")
print(f"WORLD_SIZE: {dist.size()}")
print(x)
```

Execute with `mpirun`:

```
$ mpirun -n 2 python train.py

  File "/home/mpiuser/train.py", line 3, in <module>
    dist = mx.distributed.init(strict=True, backend="mpi")
RuntimeError: Cannot initialize MPI
  File "/home/mpiuser/train.py", line 3, in <module>
    dist = mx.distributed.init(strict=True, backend="mpi")
RuntimeError: Cannot initialize MPI
```

**Expected behavior**

MLX distributed should work on Linux container correctly.

 - OS Version: [e.g. MacOS 14.1.2] Debian Trixie release.
Base image: https://github.com/kubeflow/mpi-operator/blob/master/build/base/Dockerfile

 - Version [e.g. 0.7.0]

angeloskath:
Well this started in order to close #1964. Unfortunately gcc seems to be requiring all the functions declared in `<mpi.h>` to be available or linked to by the `*.so`. So basically if we `#include <mpi.h>` we need to link with `libmpi.so`. So to avoid that I simply declared what types we need since we are loading all functions dynamically anyway.

Let me know wyt. I am ok with not merging this.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -112,6 +112,7 @@ jobs:
           command: |
             python3 -m unittest discover python/tests -v
             mpirun --bind-to none -host localhost:8 -np 8 python python/tests/mpi_test_distributed.py
+            mlx.launch --verbose -n 8 python/tests/ring_test_distributed.py
       - run:
           name: Build CPP only
           command: |
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -89,6 +89,7 @@ jobs:
             pip install numpy
             sudo apt-get update
             sudo apt-get install libblas-dev liblapack-dev liblapacke-dev
+            sudo apt-get install openmpi-bin openmpi-common libopenmpi-dev
       - run:
           name: Install Python package
           command: |
@@ -110,6 +111,7 @@ jobs:
           name: Run Python tests
           command: |
             python3 -m unittest discover python/tests -v
+            mlx.launch --verbose -n 8 --backend mpi --mpi-arg='--bind-to-none' python/tests/mpi_test_distributed.py
       - run:
           name: Build CPP only
           command: |
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -111,7 +111,7 @@ jobs:
           name: Run Python tests
           command: |
             python3 -m unittest discover python/tests -v
-            mlx.launch --verbose -n 8 --backend mpi --mpi-arg='--bind-to-none' python/tests/mpi_test_distributed.py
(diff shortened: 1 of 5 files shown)


angeloskath:
I mean we can. I just hate to have to define the types myself which means we may have to be a bit more careful when a new major MPI version is released. For now it is just `MPI_Status` so it isn't too big a deal.

To give you a concrete example MPICH uses an integer `MPI_Comm` and probably a different `MPI_Status` as well. I know we only support openmpi for now but you get the point.

awni:
Do you think it's possible to only include `<mpi.h>` and the stuff we need for it if the shared library is available?

So if I'm buildling from source on a linux machine and `libmpi.so` isn't there then we don't build the MPI bits at all?

angeloskath:
That's what we do. The problem is that on linux if we do include `<mpi.h>` then we also have to link to `libmpi.so`. This has the issue that then the mlx library can only be loaded on platforms that have `libmpi.so` while now we look for it at runtime and if it isn't found we bail.

I don't know why this isn't a problem on the Mac tbh. Basically clang only exports the names that are used in the code and not everything that it found in `<mpi.h>` . This seems a much better strategy in general imho.

awni (commented):
Let's land it. If the types become an issue we'll figure something else out.

angeloskath:
Yeah I don't know what I didn't do that from the get go. It should be good now, I am checking the MPI library version at runtime instead of at build time and I removed the check at build time as well. No need to have MPI at all when building.

Feel free to approve when you want and then we can merge.

awni (approved):
LGTM!

stemann:
You may want to take a peek at https://github.com/eschnett/MPItrampoline - A C library (built for the Julia ecosystem) for handling the differences among MPI implementations.

PR #1976, opened 2025-03-20
Fix malloc or wait deadlock

awni:
A similar possibility for deadlock here as #1969  because of malloc_or_wait in multiple streams.

The fix here removes `malloc_or_wait` entirely and instead to waits in the main eval loop if the memory limit is exceeded.

Generally speaking I like this change because it decouples `malloc` from the scheduler.

One thing that would be good to do is to enable this for the CPU only build as well which would require adding the memory limit and memory counting to the common allocator and making the memory API more back-end agnostic. If we like this approach I can do a follow on diff with that right after this.


angeloskath (approved):
👍

PR #1980, opened 2025-03-20
fix mask in sdpa

awni:
Broadcast mask before op.


jagrit06 (approved):
LGTM! Thanks for fixing it up1

PR #1982, opened 2025-03-21
Move memory APIs into top level mlx.core

awni:
This is part 1 of a two part diff.  This part just moves the memory APIs from `mx.metal` to `mx`. The second part will make the memory APIs do something useful for the CPU-only allocator.


python/tests/test_memory.py
```diff
@@ -0,0 +1,59 @@
...
+
+import unittest
+
+import mlx.core as mx
+import mlx_tests
+
+
+class TestMemory(mlx_tests.MLXTestCase):
+    def test_memory_info(self):
+        old_limit = mx.metal.set_cache_limit(0)
```
angeloskath, line 11: nit: Forgotten.

angeloskath (commented):
Looks good. You want to skip the memory test on Linux until part 2 or sth else?

awni:
Yes I'll skip that test until part 2. We weren't running it before on linux either

angeloskath (approved):
👍

PR #1985, opened 2025-03-21
Add a ring all gather

angeloskath:
Simple enough and brings the ring and mpi backends to feature parity except group splitting of course.

It achieves the same bandwidth speeds as the `all_sum` which is expected and significantly better than open MPI. 48 Gbps for a single ring and ~71 for a double ring.

### DIFF
--- a/mlx/distributed/ring/ring.cpp
+++ b/mlx/distributed/ring/ring.cpp
@@ -603,8 +603,8 @@ class RingGroup : public GroupImpl {
     return size_;
   }
 
-  void all_sum(const array& input_, array& output, Stream stream) override {
-    SWITCH_TYPE(output, all_sum<T>(input_, output, stream));
+  void all_sum(const array& input, array& output, Stream stream) override {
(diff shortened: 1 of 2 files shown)


mlx/distributed/ring/ring.cpp
```diff
@@ -603,16 +603,47 @@ class RingGroup : public GroupImpl {
...
   }
 
   std::shared_ptr<GroupImpl> split(int color, int key = -1) override {
     throw std::runtime_error("[ring] Group split not supported.");
   }
 
   void all_gather(const array& input, array& output, Stream stream) override {
-    throw std::runtime_error("[ring] All gather not supported.");
+    auto& encoder = cpu::get_command_encoder(stream);
+    encoder.set_input_array(input);
```
awni, line 616: Technically I think we want to set the output here as well. It's not currently used so doesn't matter.. but may be good to minimize future issues.

awni (approved):
Looks great, thanks!

PR #1987, opened 2025-03-21
Remove unused import

chunyang-wen:
## Proposed changes

Remove unused import.

## Checklist

### DIFF
--- a/python/mlx/optimizers/optimizers.py
+++ b/python/mlx/optimizers/optimizers.py
@@ -1,6 +1,5 @@
 # Copyright © 2023-2024 Apple Inc.
 
-import math
 from typing import Callable, List, Optional, Tuple, Union
 
 import mlx.core as mx


chunyang-wen:
`__all__` controls what will be imported when you use `from a import *`.

awni:
> __all__ controls what will be imported when you use from a import *.

Indeed I get what `__all__` does. What I don't get is why you added a bunch of stuff to it. Can you explain that part?

As far as I can tell this doesn't change the behavior of e.g. `from mlx.nn import *` since by default all of the names that don't start with `_` are included.

chunyang-wen:
Oh, sorry for not making myself clear. I'd like to remove the all warning that the variables are imported but not used.

awni:
What about this instead (from Google search):
```
# pyright: reportUnusedVariable=false
```

chunyang-wen:
Ok. I will just delete the unused import. Leave other stuff unchanged.

awni (approved):
Thanks!

PR #1988, opened 2025-03-21
Add stats and limit to common allocator and enable tests

awni:
- Add memory limit to common allocator
- Add supported memory stats (peak size, active size) to common allocator.


angeloskath (approved):
Looks great!

PR #1989, opened 2025-03-21
Disable mpi on version mismatch

angeloskath:
Sorry, forgot to actually disable MPI (instead of only complaining) on version mismatch in the previous PR.


awni (approved):
LGTM

PR #1991, opened 2025-03-22
fix docs

awni:
Fix docs for recent memory API change.


docs/Doxyfile
```diff
@@ -13,7 +13,7 @@ EXCLUDE_PATTERNS       = */private/*
 CREATE_SUBDIRS         = NO
 FULL_PATH_NAMES        = YES
 RECURSIVE              = YES
-GENERATE_HTML          = YES
```
barronalex, line 16: Is this a big change?

barronalex (approved):
Nice!

docs/Doxyfile
awni: No we never used that so I just disabled it. Sphinx uses the xml output

Issue #1994, opened 2025-03-23
[BUG] Creating a f64 scalar array returns a f32 scalar array


minghuaw:
**Describe the bug**
Creating a scalar array of type `float64` would actually return a scalar array of type `float32` (using the MLX-C) because of the following lines.

https://github.com/ml-explore/mlx/blob/f018e248cd75dbb65668f418d6afb67842ea28b7/mlx/dtype.cpp#L150-L153

**To Reproduce**
Found this while updating the rust binding, so I don't have a c code for this right now. I can prep one if needed but the cause is already shown above

**Expected behavior**
A scalar `Array` of type `float64` is expected.

**Desktop (please complete the following information):**
 - OS Version: MacOS 15.2
 - Version 0.23.1

**Additional context**
n/a

awni:
I don't think that is the cause. When you explicitly pass the dtype to the array constructor that function should not be called. It would be useful to see the code that reproduces this.

minghuaw:
The c binding provides a function 'mlx_array_new_float64' which internally creates a new scalar array with 'mlx::core::array(val)' where 'val' is a 'double'. This eventually calls this constructor 

'''cpp
template <typename T>
  explicit array(T val, Dtype dtype = TypeToDtype<T>());
'''

To me it makes sense that you'd want to default floating points to f32 instead of f64 given the limited support of the latter on GPU. I guess the problem is that the c binding probably shouldn't name itself 'mlx_array_new_float64' which implies creating a float64 scalar array from a double

minghuaw:
I totally agree with the current approach of picking up float32 dtype given the limited support of float64 on GPU. This is more about whether the c binding should offer a function that implies something in its name while it actually doesn't. I'll close this and open a new issue in the mlx-c repo

PR #1995, opened 2025-03-23
fix: Fix `TypeToDtype<double>` to return `float64`

minghuaw:
## Proposed changes

Closes: #1994 

This PR addresses the problem where a scalar array of type `float32` is returned when attempting create a scalar array of type `float64`

## Checklist

### DIFF
--- a/mlx/dtype.cpp
+++ b/mlx/dtype.cpp
@@ -149,7 +149,7 @@ TypeToDtype<float>::operator Dtype() {
 
 template <>
 TypeToDtype<double>::operator Dtype() {
-  return float32;
+  return float64;
 }
 
 template <>
--- a/tests/array_tests.cpp
+++ b/tests/array_tests.cpp
@@ -178,6 +178,11 @@ TEST_CASE("test array types") {
     basic_dtype_test(float, float32);
   }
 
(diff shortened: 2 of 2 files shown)


awni:
I don't think we want to make this change.

```
array(1.0);
```

should pick up type `float32`. If you want a `float64` array you have to explicitly ask for it.

awni:
Is there a different behavior you noticed that should be fixed?

minghuaw:
> I don't think we want to make this change.
> 
> 
> 
> ```
> 
> array(1.0);
> 
> ```
> 
> 
> 
> should pick up type `float32`. If you want a `float64` array you have to explicitly ask for it.

I absolutely agree with you current approach that the user has to explicitly set the dtype to float64 in order to create one given the limited support of `float64` on GPU. I guess the problem should be worded differently. The problem is then more about whether the c binding should offer `mlx_array_new_float64` at all. Because its name implies creating a scalar array of type `float64` when it actually doesn't. 

Let me close this and the related issue and open it in the mlx-c repo

awni:
Fwiw I do think `mlx_array_new_float64` should return a float64 array

PR #2006, opened 2025-03-25
Wire cache

awni:
Keep cache memory wired if wired limit is set.

This helps a decent amount when training on longer sequences / larger batch sizes.

Lora fine tuning benchmark on a Llama 1B with batch size 4 and sequence length ~2048:

Pre: Tokens/sec 1819.193
Post (wire limit with wired cache): Tokens/sec 2082.386

Didn't notice any change in LLM inference (which is the only other place we use the wired limit as far as I know).

### DIFF
--- a/mlx/backend/metal/allocator.cpp
+++ b/mlx/backend/metal/allocator.cpp
@@ -33,8 +33,11 @@ namespace metal {
 
 namespace {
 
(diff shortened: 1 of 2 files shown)


angeloskath (approved):
Looks great!

PR #2007, opened 2025-03-26
Sdpa specialization for head dim 256

awni:
The Gemma3 models use a head dim of 256 so it seems worth adding a specialization for it.

There is a reasonable speedup.. given the trivial addition / cost it seems worth while. Benchmarks on M4 Max generating 512 tokens.

Model | Prompt | Pre tok/sec | Post tok/sec
------ | ----- | ---- |  -----
1B 4bit | short (14 toks) | 280.46 | 308.223
1B 4bit | long (17k toks) | 264.660 | 284.337
4B 4bit | short (14 toks) | 140.80 | 142.64
4B 4bit | long (17k toks) | 112.419 | 120.819


angeloskath (approved):
🚀

PR #2009, opened 2025-03-27
iinfo and scalar overflow detection

awni:
- Add `iinfo`
- throw in cases like the following rather than silently oveflowing

```
y = mx.array(2000000000, dtype=mx.int32)
x = 3000000000
y + x
```


angeloskath (approved):
👍

PR #2013, opened 2025-03-27
feat: Triangular LR Scheduler

dalongbao:
## Proposed changes
Implemented Triangular LR Scheduler as used in [one of Keller Jordan's CIFAR10 speedruns](https://gist.github.com/KellerJordan/7a99ef4f300a6941c12b544a0030b2a9)

## Checklist


angeloskath:
Thank you for the PR! However, I am closing this as you can achieve the exact same schedule by joining two linear schedules so there is no need for a specific scheduler for this.

PR #2016, opened 2025-03-28
Use minimum deployment target instead of xcode in Circle CI

awni:
The big change here is how we build and test for different OS. Instead of coupling the OS we build for with the xcode version we use in Circle, we always use the latest xcode but specify the OS to build for using `MACOSX_MINIMUM_TARGET`.

There are a couple benefits:
- Good to build our releases with the latest xcode: binaries smaller and have better optimizations
- Useful to decouple xcode from OS. That way we can build a release for newer or older OS without needing that OS to be available on xcode (e.g. we can now build a release for macOS 15 and use more features that require macos 15 / Metal 3.2).

Minor additions:
- Change the resource class to an M2
- Fix cmake warning compiler setting to actually do it's job

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -24,8 +24,8 @@ jobs:
         type: boolean
         default: false
     macos:
-      xcode: "15.2.0"
-    resource_class: macos.m1.medium.gen1
+      xcode: "16.2.0"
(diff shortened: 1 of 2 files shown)


angeloskath (approved):
Nice!

PR #2018, opened 2025-03-28
causal vector sdpa

awni:
Adds the causal option to the fast vector sdpa. Since we use `"causal"` in MLX LM now we no longer route to the fast SDPA when the query length is more than 1 (e.g. for spec decoding).

Also got rid of the memory efficient threshold. Since we don't really need this.. and in general it's good not to dispatch differently based on the shape of the input (for e.g. shapeless compilation).


angeloskath (approved):
Looks great!

PR #2021, opened 2025-03-29
Add missing funcs to docs

smpanaro:
## Proposed changes

Follow up to https://github.com/ml-explore/mlx/pull/2020, was curious if there were others (didn't realize it was an easy fix to add them).

There are a couple I wasn't sure if are intended to be in the docs:
- checkpoint
- concat + permute_dims (aliases)

## Checklist

### DIFF
--- a/docs/src/python/ops.rst
+++ b/docs/src/python/ops.rst
@@ -36,10 +36,12 @@ Operations
    bitwise_or
    bitwise_xor
    block_masked_mm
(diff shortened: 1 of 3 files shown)


angeloskath (approved):
Thanks, this is great!

Async eval started slightly experimental but a) it is now quite battle tested and b) it is already documented on the Swift side so I don't think we should refrain from documenting it.

Issue #2023, opened 2025-03-30
[BUG] Unexpected return value for `sign` of some 4-element `uint64` array


stemann:
`sign` for a some (randomly generated) 4-element `uint64` array does not return the expected return value (an array of `0x0000000000000001`), but instead returns an array of `0xffffffffffffffff`.

The following test fails:

```c++
    array x_uint64({uint64_t(0xa11cc311cb6acd70), uint64_t(0x7a375ac3ebb533f3), uint64_t(0x734969adf9d7190c), uint64_t(0xb400515a4f673424)});
    array expected({uint64_t(0x0000000000000001), uint64_t(0x0000000000000001), uint64_t(0x0000000000000001), uint64_t(0x0000000000000001)});
    CHECK(array_equal(sign(x_uint64), expected).item<bool>());
```

**To Reproduce**

```c++
// fails
    array x_uint64({uint64_t(0xa11cc311cb6acd70), uint64_t(0x7a375ac3ebb533f3), uint64_t(0x734969adf9d7190c), uint64_t(0xb400515a4f673424)});
    array expected({uint64_t(0x0000000000000001), uint64_t(0x0000000000000001), uint64_t(0x0000000000000001), uint64_t(0x0000000000000001)});
    CHECK(array_equal(sign(x_uint64), expected).item<bool>());

// succeeds, wrong return value
    x_uint64 = array({uint64_t(0xa11cc311cb6acd70), uint64_t(0x7a375ac3ebb533f3), uint64_t(0x734969adf9d7190c), uint64_t(0xb400515a4f673424)});
    expected = array({uint64_t(0xffffffffffffffff), uint64_t(0xffffffffffffffff), uint64_t(0xffffffffffffffff), uint64_t(0xffffffffffffffff)});
    CHECK(array_equal(sign(x_uint64), expected).item<bool>());

// succeeds, the first three elements
    x_uint64 = array({uint64_t(0xa11cc311cb6acd70), uint64_t(0x7a375ac3ebb533f3), uint64_t(0x734969adf9d7190c)});
    expected = array({uint64_t(0x0000000000000001), uint64_t(0x0000000000000001), uint64_t(0x0000000000000001)});
    CHECK(array_equal(sign(x_uint64), expected).item<bool>());

// succeeds, the first two elements
    x_uint64 = array({uint64_t(0xa11cc311cb6acd70), uint64_t(0x7a375ac3ebb533f3)});
[...]

PR #2024, opened 2025-03-30
Added uint64 tests for sign

stemann:
## Proposed changes

Tests for #2023 

## Checklist


angeloskath:
Added a fix, the issue was only on CPU wrt simd sign for unsigned numbers. To me it seems that the previous code should be working but it was converting to all 1s (bitwise). It sometimes seemed to worked when the simd-width was 1 of course.

angeloskath (approved):

PR #2027, opened 2025-03-30
fix residency set clearing

awni:
Forgot to remove cleared buffers from the residency set in one more place.


angeloskath (approved):
👍

Sorry for missing it in the PR review

PR #2028, opened 2025-03-30
Custom logsumexp

awni:
Add this to speedup and reduce memory use when computing `logsumexp` during e.g. lora fine tuning. The large vocab (200k+ in some cases) makes this optimization important. Also since the `logsumexp` is in high precision this let's us avoid up casting the logits and instead only upcasting the per-token loss prior to the reduction.

QLoRA fine tuning Gemma 3 1B:

Pre / Post

Iter 10: Train loss 3.595, Learning Rate 1.000e-05, It/sec 0.227, Tokens/sec 812.842, Trained Tokens 35845, Peak mem 66.335 GB

Iter 10: Train loss 3.594, Learning Rate 1.000e-05, It/sec 0.345, Tokens/sec 1235.568, Trained Tokens 35845, Peak mem 49.164 GB

And a microbenchmark:

```python
shape = (4096, 120_000)

x = mx.random.uniform(shape=shape)

def fun(x):
    return [mx.logsumexp(x, axis=-1, keepdims=True) for _ in range(100)]
```

Pre: 1588.345
Post: 257.241


angeloskath (approved):
Looks great! The memory savings look even greater :-) 

Left one nitpick...

mlx/backend/metal/kernels/logsumexp.metal
```diff
@@ -0,0 +1,35 @@
...
+#include <metal_common>
+#include <metal_simdgroup>
+
+using namespace metal;
+
+// clang-format off
+#include "mlx/backend/metal/kernels/utils.h"
+#include "mlx/backend/metal/kernels/logsumexp.h"
+
+#define instantiate_logsumexp(name, itype)                          \
```
angeloskath, line 12: Sorry I think I forgot to comment it before (it's a nitpick anyway)...

Maybe use `instantiate_kernel` here?

python/tests/test_ops.py
```diff
@@ -690,15 +690,34 @@ def test_broadcast(self):
...
+        self.assertTrue(mx.allclose(mx.logsumexp(x), logsumexp(x)))
+
+        # Broadcast
+        x = mx.broadcast_to(mx.random.uniform(shape=(2, 1, 8)), (2, 2, 8))
+        self.assertTrue(mx.allclose(mx.logsumexp(x), logsumexp(x)))
+
+        # Large
+        x = mx.random.uniform(shape=(1025,))
+        x = mx.broadcast_to(mx.random.uniform(shape=(2, 1, 8)), (2, 2, 8))
+        self.assertTrue(mx.allclose(mx.logsumexp(x), logsumexp(x)))
```
zcbenz, line 720: This assertion tests the same thing with previous one.

awni: Yea looks like a mistake

PR #2029, opened 2025-03-31
Swap -inf for finite_minimum value

angeloskath:
Fixes issue https://github.com/ml-explore/mlx-lm/issues/58 by avoiding -inf altogether and using `finite_min`.

I don't foresee much of a problem except in case of underflow where an attention score results in infinity and then our supposed infinity becomes a proper value. Given that the type is `float` in all cases that is even more unlikely. Let me know if I missed something.

I will add Awni's test as well.


awni (approved):
Very nice! Thanks for the quick fix!

PR #2031, opened 2025-04-01
Tune quant dispatch

awni:
Tune quant dispatch conditions to qmv vs qmm.

The optimal is both machine and matrix size dependent unfortunately so this is only an approximation. It's conservative in that we continue to route to the qmm unless it's pretty clear that there is a speedup to be had.

In the plots below:
- x-axis is batch size
- y-axis is time in milliseconds
- dashed line is qmm
- solid line is qmv

Here's a benchmark comparing the two at size = 4096 on a few machines:

Screenshot 2025-04-01 at 8 08 00 AM

Another benchmark for just M2 Ultra for 3 different sized matrices:

Screenshot 2025-04-01 at 8 09 12 AM

Same benchmark on M3 Max:

Screenshot 2025-04-01 at 8 09 35 AM

### DIFF
--- a/mlx/backend/metal/kernels/quantized.h
+++ b/mlx/backend/metal/kernels/quantized.h
@@ -586,13 +586,13 @@ METAL_FUNC void qmv_quad_impl(
   // Adjust positions
   const int in_vec_size_w = in_vec_size / pack_factor;
   const int in_vec_size_g = in_vec_size / group_size;
-  const int out_row = tid.x * quads_per_simd * results_per_quadgroup + quad_gid;
+  const int out_row = tid.y * quads_per_simd * results_per_quadgroup + quad_gid;
 
   w += out_row * in_vec_size_w + quad_lid * packs_per_thread;
   scales += out_row * in_vec_size_g + quad_lid / scale_step_per_thread;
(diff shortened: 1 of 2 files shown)


awni:
For Mistral 7B batched we get a nice speedup especially at small batch sizes:

Batch size | Pre ms | Post ms
----- | ---- | -----
6 | 31.971 | 19.400
8 | 31.817 | 25.431
10 |31.416| 31.274


Benchmark:

```python
import time
import mlx.core as mx
from mlx_lm import load

model, _ = load("mlx-community/mistral-7B-Instruct-v0.3-4bit")

for batch_size in [6, 8, 10, 12, 14, 20]:
    prompt = mx.array([[0] * batch_size])

    for _ in range(10):
        mx.async_eval(model(prompt))

    tic = time.time()
    for _ in range(10):
        mx.async_eval(model(prompt))
    toc = time.time()
    ms = 1000 * (toc - tic) / 10
    print(f"{ms:.3f}")
```

awni:
It's true I didn't tune for those at all. I just checked `qmv` and on m4 max the crossing point is 14 (for size == 4352), so close to `qmv_fast`. I'm gues quad will be quite different. If anything I would expect the crossing point to be further out since the matrices are probably quite small with that one.

PR #2032, opened 2025-04-01
Residency set fix forreal

awni:
Arg I had a typo in the previous fix.

Closes https://github.com/ml-explore/mlx-lm/issues/64


angeloskath (approved):
Damn again it shows I should be more careful when reviewing...

PR #2040, opened 2025-04-03
Gather mm new kernel and small refactoring

angeloskath:
This is probably good enough to start reviewing. I need to fix the JIT build and also add some benchmark numbers. For now I am only routing matrix vector products but I think small matrices could also be routed there.

So the change list is

- Move gather out of `steel_gemm_fused` so that both are easier to edit
- Refactor GatherMM::eval_gpu to just route so that it is a bit easier to follow and add specializations etc
- Add `store_slice` to `BlockMMA` and the tiles internally so we can store only a slice of the output (a variation of `store_safe` basically)

There is more to be done in refactoring steel and simplifying routing and computing the arguments for matmul, I haven't done much there at all. I am sure Jagrit agrees and has a lot of stuff cooking.

### DIFF
--- a/mlx/backend/metal/jit_kernels.cpp
+++ b/mlx/backend/metal/jit_kernels.cpp
@@ -584,6 +584,40 @@ MTL::ComputePipelineState* get_steel_gemm_masked_kernel(
   return d.get_kernel(kernel_name, lib);
 }
 
+MTL::ComputePipelineState* get_steel_gemm_gather_kernel(
+    metal::Device& d,
+    const std::string& kernel_name,
+    const std::string& hash_name,
+    const metal::MTLFCList& func_consts,
+    const array& out,
+    bool transpose_a,
+    bool transpose_b,
+    int bm,
+    int bn,
+    int bk,
+    int wm,
+    int wn) {
+  const auto& lib_name = kernel_name;
+  auto lib = d.get_library(lib_name, [&]() {
+    std::ostringstream kernel_source;
+    kernel_source << metal::utils() << metal::gemm()
+                  << metal::steel_gemm_gather()
+                  << get_template_definition(
+                         lib_name,
+                         "gemm",
+                         get_type_string(out.dtype()),
+                         bm,
+                         bn,
+                         bk,
+                         wm,
+                         wn,
+                         transpose_a,
+                         transpose_b);
+    return kernel_source.str();
+  });
+  return d.get_kernel(kernel_name, lib, hash_name, func_consts);
+}
+
 MTL::ComputePipelineState* get_gemv_masked_kernel(
     metal::Device& d,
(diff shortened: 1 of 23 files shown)


mlx/backend/metal/jit_kernels.cpp
```diff
@@ -479,6 +479,40 @@ MTL::ComputePipelineState* get_steel_gemm_fused_kernel(
...
+                         "gemm",
+                         get_type_string(out.dtype()),
+                         bm,
+                         bn,
+                         bk,
+                         wm,
+                         wn,
+                         transpose_a,
+                         transpose_b);
+    return kernel_source.str();
```
awni, line 511: minor nit: I would use concatenate / `+=` here as well.. I'm trying to move all of them to that same pattern over time since it's faster.

awni:
This looks really great so far + a lot of nice refactoring!

angeloskath:
I am running the benchmark in the repo with different parameters and reporting below

```
N = batch size * seq len
D = model dims
M = hidden dims
E = number of experts
I = active experts
```

On an M2 Ultra

N | D | M | E | I | Before unsorted (ms) | Before sorted (ms) | Now sorted (ms)
--- | --- | --- | --- | --- | --- | --- | --- 
1024 | 1024 | 1024 | 32 | 4 | 12.7 | 10.2 | 1.9
1024 | 1024 | 1024 | 8 | 4 | 10.3 | 10.2 | 1.7
1024 | 4096 | 1024 | 256 | 4 | 171.8 | 40.6 | 26.6
1024 | 4096 | 1024 | 256 | 8 | 345 | 79 | 35.3
1024 | 4096 | 1024 | 8 | 4 | 46.9 | 39.4 | 5.5
512 | 1024 | 1024 | 256 | 4 | 19 | 5.8 | 4.8
512 | 1024 | 1024 | 256 | 1 | 5.3 | 2.8 | 3.4 

On my laptop

N | D | M | E | I | Before unsorted (ms) | Before sorted (ms) | Now sorted (ms)
--- | --- | --- | --- | --- | --- | --- | --- 
1024 | 1024 | 1024 | 32 | 4 | 268 | 68.1 | 10.8
1024 | 1024 | 1024 | 8 | 4 | 149 | 68.1 | 9.6
512 | 1024 | 1024 | 256 | 4 | 190 | 35 | 37
512 | 1024 | 1024 | 256 | 1 | 48 | 23 | 26

Basically if there are enough samples to fill the experts we see very significant speedups otherwise we can even have some slowdown. The slowdown is in cases where we are accessing many experts in each group of 16 inputs. Since in the benchmark the experts are also uniformly distributed, it is a kind of worst case scenario.

jagrit06 (approved):
Looks good to me! There's a lot of changes, but each step is fairly straightforward

PR #2041, opened 2025-04-03
Min / max reductions

nastya236:
## Proposed changes

Added min and max reductions for mpi and ring backends.
## Checklist


awni:
Looks very good! Thanks!

nastya236:
Thank you @angeloskath for the suggestion!
Indeed passing functor instead of function pointer helps a bit. Here are some numbers:

| Shape            | Functor (1e-6 s) | Function pointer (1e-6 s)     |
|------------------|------------------|------------------------|
| `(7,)`           | 3.85243940       | 3.58418038             |
| `(10,)`          | 2.93540827       | 2.93708057             |
| `(1024,)`        | 2.82622088       | 3.01080145             |
| `(1024, 1024)`   | 2.87798059       | 3.14998790             |
| `(16384, 16384)` | 5.02713025       | 5.93123841             |

There is a small degradation for small shapes but I guess it is not as important.

angeloskath:
Hm not sure what the measurement is. 5μs to add `16k x 16k` and 4μs for 7 seems a bit odd.

nastya236:
| Shape            | Functor (s)       | Function pointer (s)      |
|------------------|------------------|------------------|
| `(7,)`           | 0.00023820        | 0.00023999       |
| `(10,)`          | 0.00017402        | 0.00017373       |
| `(1024,)`        | 0.00018299        | 0.00017948       |
| `(1024, 1024)`   | 0.00053568        | 0.00053819       |
| `(16384, 16384)` | 0.03507892        | 0.03514215       |

Oh of course — sorry, that was a silly mistake :) Now it makes much more sense, I think. From the numbers, the difference looks quite minimal, but let me know if you'd prefer I keep this version or revert back :)

angeloskath:
So I calculated the assembly and at least clang is always inlining the function pointer call resulting in exactly identical assembly to the functor. I think it is fine to leave it with the functor, it is also a bit more flexible so I am merging this as is.

Thanks again @nastya236 for the super clean implementation.

angeloskath (approved):

PR #2043, opened 2025-04-03
patch bump

awni:
🚀


andresy:
Should there be a minor release update, given the API of `scaled_dot_product_attention()` changed?

jagrit06 (approved):

angeloskath (approved):
I don't mind a minor vs a patch. @andresy I know the rules say it should be a minor but we kinda go by how impactful the change has been I think :-)

awni:
Yes we’re not so strict with semver for low footprint APIs. I’d rather do this now as a patch for the bug fixes and a minor next week post new MOE kernels

PR #2045, opened 2025-04-04
Fix ci release

angeloskath:
Unreasonably large diff but Circle doesn't have a way to exclude subsets from the matrix.


awni (approved):
Thanks for fixing that!

The delineation of the exluded conditions is a bit unfortunate.. but I guess we'll be able to remove it after we stop updating 13.5 soon enough.

angeloskath:
> The delineation of the exluded conditions is a bit unfortunate

Yeah you 'd think they would allow partial matches right?

PR #2046, opened 2025-04-05
add a half simd gemm fallback

awni:
Adds a SIMD gemm fallback option to support half types on other platforms.


angeloskath (approved):
Beautifully simple.

PR #2047, opened 2025-04-06
wait_for_one should wait for task to finish

zcbenz:
Make `wait_for_one` wait for one task to finish, see https://github.com/ml-explore/mlx/discussions/1941 for discussions on this.

Potentially it could make some work slower as the scheduler now actually waits until active tasks becomes lesser.


angeloskath (approved):
Actually after #1869 this change is pretty safe since the only place that uses `wait_for_one` is the `eval_impl()` and it can only mean that a task finished since only `eval_impl()` is submitting stuff to the scheduler.

Issue #2015, opened 2025-03-28
Casting negative float to uint is undefined behavior


zcbenz:
This test checks the result of converting `-3.0f` to `uint32_t` being same on CPU and GPU:

https://github.com/ml-explore/mlx/blob/5580b47291caed127e54a427a265e7776add2997/tests/creations_tests.cpp#L138-L149

Which however on my machine does not pass, because in CPU the result is 4294967293 while in CUDA it is 0.

After trying to debug CUDA, I realized this is an undefined behavior: https://embeddeduse.com/2013/08/25/casting-a-negative-float-to-an-unsigned-int/#cast-became-undefined .

I can just disable the test in CUDA, but I wonder if this test should be removed at all.

awni:
Yes probably a good idea to remove the test for undefined behavior.

PR #2048, opened 2025-04-06
Remove test of converting negative float to uint

zcbenz:
Close https://github.com/ml-explore/mlx/issues/2015.

It is an [undefined behavior](https://embeddeduse.com/2013/08/25/casting-a-negative-float-to-an-unsigned-int/#cast-became-undefined) that GPU and CPU may give different results.


awni (approved):
Thanks!

PR #2051, opened 2025-04-07
Only request residency once

awni:
Removes redundant and potentially wasteful calls to `requestResidency`. Only requests residency once at the time of construction.


angeloskath (approved):
👍

PR #2053, opened 2025-04-08
Fix causal mask sdpa vec

awni:
As title 😬  🐛


angeloskath (approved):
🐛

angeloskath:
I 've had one of those recently and it is hard because you do get a kernel back and it might even work or partially work 🙀

PR #2055, opened 2025-04-08
Do not load the default lib if another is requested

angeloskath:
Before:
Calling `register_library("foo")` succeeds and registers the default metal library under the key `foo`.

Now:
It will fail to load the library and will report an error with the attempted collocated path.


awni (approved):
Awesome thanks!!

PR #2058, opened 2025-04-10
Fix MSVC build due to use of M_LN2

zcbenz:
Similar to https://github.com/ml-explore/mlx/pull/1665.

### DIFF
--- a/mlx/backend/cpu/unary.cpp
+++ b/mlx/backend/cpu/unary.cpp
@@ -1,5 +1,8 @@
 // Copyright © 2024 Apple Inc.
 
+// Required for using M_LN2 in MSVC.
+#define _USE_MATH_DEFINES
+
 #include <cassert>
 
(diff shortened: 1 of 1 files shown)


mlx/backend/cpu/unary.cpp
```diff
@@ -1,5 +1,8 @@
 // Copyright © 2024 Apple Inc.
 
+// Required for using M_LN2 in MSVC.
+#define _USE_MATH_DEFINES
```
awni, line 4: Why put it here instead of in the actual `simd/base_simd.h` file?

zcbenz: `cmath` was included before `base_simd.h`, so putting it here does not work, we have to put it somewhere before any header includes `cmath`.

awni (approved):
Thanks!

PR #2059, opened 2025-04-10
Remove static initializers

hdeng-apple:
## Proposed changes

Trying to eliminate most of static initializers to reduce overhead during app launch. These were found by "-Werror=global-constructors"

## Checklist

### DIFF
--- a/mlx/3rdparty/pocketfft.h
+++ b/mlx/3rdparty/pocketfft.h
@@ -537,7 +537,11 @@ inline size_t &num_threads()
   static thread_local size_t num_threads_=1;
   return num_threads_;
   }
-static const size_t max_threads = std::max(1u, std::thread::hardware_concurrency());
+inline const size_t &max_threads()
+  {
+  static thread_local const size_t max_threads_ = std::max(1u, std::thread::hardware_concurrency());
+  return max_threads_;
+  }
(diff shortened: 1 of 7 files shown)


mlx/3rdparty/pocketfft.h
```diff
@@ -537,7 +537,11 @@ inline size_t &num_threads()
   static thread_local size_t num_threads_=1;
```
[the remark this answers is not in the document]
hdeng-apple: Sure, I have reverted.

mlx/io/load.h
```diff
@@ -99,9 +99,11 @@ class ParallelFileReader : public Reader {
     return "file " + label_;
   }
 
+ private:
+  static ThreadPool& thread_pool();
+
```
awni, line 104: Is the purpose here to ensure the thread pool is active when this class is constructed? 

Also, small nit: let's just put this in the same `private:` section below.

hdeng-apple: I put it back to `private:` below.
The previous `thread_pool_` creates a shared instance when app is launched. Changing to local static variable changes to lazily initialized at first time it was called.

awni (approved):
Thanks!!

PR #2060, opened 2025-04-10
Fix the test and add custom min/max reductions for uncommon MPI types

angeloskath:
Fixes the MPI test and also adds `all_max`, `all_min` for float16, bfloat16 and complex64.


awni (approved):
Looks great! Not sure why that linux test is failing 🤔

PR #2061, opened 2025-04-10
Search mlx.metallib in macOS framework "Resources" dir

hdeng-apple:
## Proposed changes

`get_colocated_mtllib_path` returns the metallib in the directory of executable, but macOS frameworks put it in `Resources` dir.

## Checklist

### DIFF
--- a/mlx/backend/metal/device.cpp
+++ b/mlx/backend/metal/device.cpp
@@ -1,6 +1,7 @@
 // Copyright © 2023-2024 Apple Inc.
 
 #include <cstdlib>
+#include <filesystem>
 #include <sstream>
 
 #include <sys/sysctl.h>
@@ -14,6 +15,8 @@
 #include "mlx/backend/metal/metal_impl.h"
 #include "mlx/backend/metal/utils.h"
 #include "mlx/utils.h"
+
+namespace fs = std::filesystem;
 
 namespace mlx::core::metal {
 
@@ -79,12 +82,18 @@ MTL::Library* try_load_bundle(
 // Firstly, search for the metallib in the same path as this binary
 std::pair<MTL::Library*, NS::Error*> load_colocated_library(
     MTL::Device* device,
-    const std::string& lib_name) {
-  std::string lib_path = get_colocated_mtllib_path(lib_name);
-  if (lib_path.size() != 0) {
-    return load_library_from_path(device, lib_path.c_str());
+    const std::string& relative_path) {
+  std::string binary_dir = get_binary_directory();
+  if (binary_dir.size() == 0) {
+    return {nullptr, nullptr};
   }
-  return {nullptr, nullptr};
+
+  auto path = fs::path(binary_dir) / relative_path;
+  if (!path.has_extension()) {
+    path.replace_extension(".metallib");
+  }
+
+  return load_library_from_path(device, path.c_str());
 }
 
 std::pair<MTL::Library*, NS::Error*> load_swiftpm_library(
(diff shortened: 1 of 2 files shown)


angeloskath:
Hi could you share a bit more about the cases that this fixes? Is there an issue reported?

hdeng-apple:
> Hi could you share a bit more about the cases that this fixes? Is there an issue reported?

If I understand correctly, `Device::Device()` calls `get_colocated_mtllib_path()` which finds mlx.metallib in executable's directory, However when users need to integrate MLX into an macOS app or framework, they may want to follow the convention
 - In macOS [frameworks](https://developer.apple.com/library/archive/documentation/CoreFoundation/Conceptual/CFBundles/BundleTypes/BundleTypes.html#//apple_ref/doc/uid/10000123i-CH101-SW28), binary in `Versions/A/ `, resources in `Versions/A/Ressource`.
 - In macOS [app](https://developer.apple.com/library/archive/documentation/CoreFoundation/Conceptual/CFBundles/BundleTypes/BundleTypes.html#//apple_ref/doc/uid/10000123i-CH101-SW19), binary in `Contents/MacOS/`, resources in `Contents/Resources/`.

Is there a better way in MLX to get the path? Such as defining custom relative path or use [resourcePath](https://developer.apple.com/documentation/foundation/bundle/resourcepath?language=objc) ?

angeloskath:
So when there is a Swift PM bundle we try to load from there. You can check main as this code has also been recently refactored slightly. https://github.com/ml-explore/mlx/blob/e9e268336b464839034f3f9260ad841380865338/mlx/backend/metal/device.cpp#L90

The TL;DR is that we go through all resource bundles and try to find the library there. We are looking for `default.metallib` instead of `mlx.metallib` because this is how it is built in MLX Swift currently (iirc).

Basically, did you encounter any issue with loading the metallib in an app? Also is this for loading the default metallib compiled with MLX or one from an extension? In the latter case you can also just pass the path to the metallib or the containing folder when you `register_library()`.

hdeng-apple:
I'm using MLX C++. If I link the lib and copy the mlx.metallib into my framework or app, the app crashes with error message
`terminating due to uncaught exception of type std::runtime_error: Failed to load the default metallib. library not found metal library file`

angeloskath:
I see, I think the best way forward is the following. First let's rebase this over main because the library loading behavior changed recently.

Then `get_collocated_mtllib_path` should be renamed to `get_binary_path` and `load_default_library` should use that to build candidate paths to search the default lib. Let me know if you would rather I edit directly in your PR.

hdeng-apple:
Thanks for the suggestion. I think I'd prefer to let you edit the PR directly to better organize the code related to paths.

hdeng-apple:
Or shall we let non-SwiftPM users get paths using NSBundle (load with bundle ID or Obj-C class), which make the path more portable and follows the convention of Apple platforms. The current solution in my PR does not help apps because I think it is more dangerous to test `../Resources`.

angeloskath:
I pushed what I was talking about. Let me know if this covers your use case. Now the default behavior is as follows:

1. Look for `mlx.metallib` relative to the binary
2. Look for `Resources/mlx.metallib` relative to the binary
3. Look for `default.metallib` in any bundle available from SwiftPM
4. Look for whatever path was defined during compilation

Regarding NSBundle, that sounds great. We are still talking about the default lib though right? Cause any other custom lib should be loaded by providing a full path to the `load_library` function which can be done however it is most convenient to the user.

hdeng-apple:
Thanks! I have verified it works in my case.

> We are still talking about the default lib though right?

Yes. I'm talking about the default mlx.metallib built in MLX.

angeloskath (approved):

PR #2062, opened 2025-04-11
fix fft bugs

### LINKED ISSUE #2034: [BUG] irfft does not produce correct results
The inverse STFT seems to not produce correct results, ie. should the following forward and inverse regenerate the input (white noise).  But I get a result with dropouts when using mlx.core.irfft
For comparison I used the scipy irfft

```python
import mlx.core as mx 
import numpy as np  
import scipy
import matplotlib .pyplot as plt

def show_wav(title, data, sample_rate):
    time = np.linspace(0, len(data) / sample_rate, num=len(data))
    plt.figure(figsize=(10, 4))
    plt.title(title)
    plt.plot(time, data)
    plt.show()

# Config
sample_rate = 44100
audio_ch = 1
n_fft = 6144
sample_count = 10 * sample_rate

# Create white noise buffer
white_noise = np.random.normal(0, 1, [audio_ch, sample_count]).astype(np.float32)

# Forward FFT
fft_result = []
i = 0
while i < sample_count:
    ei = min(i + n_fft, sample_count)
    segment = mx.array(white_noise[..., i:ei])
    r = mx.fft.rfft(segment, n=n_fft)
    fft_result.append(r)
    i += n_fft
print("FFT segments:", len(fft_result))

# Inverse FFT
inv_result = None
for segment in fft_result:
    r = mx.fft.irfft(segment, n=n_fft)
    # r = mx.array(scipy.fft.irfft(np.array(segment), n=n_fft))
    if inv_result is None:
        inv_result = r
    else:
        inv_result = mx.concatenate([inv_result, r], axis=1) 

show_wav("Inverted result", inv_result[0], sample_rate=sample_rate)
```

awni:
Fixes a few bugs in Metal FFT mostly related to not properly holding temporaries.

Closes #2034


barronalex (approved):
Thanks for fixing this!!

PR #2068, opened 2025-04-11
Add float mask to sdpa vector

angeloskath:
Besides the usefulness of allowing for additive masks in the vector kernel, it also fixes a bug because we are currently routing to the fast op for say 1 query and a float mask but in the primitive we are routing to the vector kernel which doesn't support it.

| Model | Token context | Before | After |
| --- | --- | --- | --- |
| Llama 1B | 80 | 464.2 | 464.2
| Llama 8B | 80 | 124.3 | 124.0
| Llama 70B | 80 | 16.3 | 16.2
| QwQ 32B | 60 | 33.2 | 33.1
| Llama 1B | 1400 | 392 | 394.9
| Llama 8B | 1400 | 115 | 115.1
| Llama 70B | 1400 | 15.4 | 15.4
| QwQ 32B | 1400 | 31.1 | 31.2

I think it is safe to say that at least the no mask kernel is not affected by this change speed wise.

### DIFF
--- a/mlx/backend/metal/kernels/scaled_dot_product_attention.metal
+++ b/mlx/backend/metal/kernels/scaled_dot_product_attention.metal
@@ -1,11 +1,11 @@
 #include <metal_stdlib>
 
-#include "mlx/backend/metal/kernels/sdpa_vector.h"
+// clang-format off
 #include "mlx/backend/metal/kernels/utils.h"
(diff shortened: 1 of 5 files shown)


awni (approved):
Looks great!! Thanks for adding/fixing that. I'm sure someone would have run into that bug at some point in the future.

tlikhomanenko:
Thanks for the fix!

PR #2069, opened 2025-04-12
LogCumSumExp

djphoenix:
## Proposed changes

Add `mx.logcumsumexp` op, that is crucial to e.g. implement log-scale minLSTM/minGRU efficiently.


awni:
This looks really nicely done!! Would you mind adding tests for a few edge cases:

- array of all -inf `mx.array([-float("inf")]* 8)`
- array of all -inf but one `mx.array([-float("inf"), 0, -float("inf")])`
- array of some inf and some -inf `mx.array([-float("inf"), float("inf"), -float("inf")])`

djphoenix:
@awni thanks for brief review. Seems like that these edge cases are erronous in current implementation... Well, there are some checks missed in LAE. I will add test cases and covering checks.

djphoenix:
@awni I just reused LogAddExp from binary_ops (both for CPU and GPU). Added mentioned test cases, now these are pass.

angeloskath (approved):
It looks great!

I changed the `LogAddExp` and `lae` to `Logaddexp` (at least where it made sense) and also checked the vjp with finite differences.

Thanks for the addition!

Issue #2001, opened 2025-03-25
MLX default weight init seems incorrect


ag487:
Seems the default mlx weight initialization is incorrect. Specifically github mlx/python/mlx/nn/layers/linear.py and other similar references.

Generally the starting point for weight init is a normal distribution with mean = zero and variance = 1 / input_dim  or 2 / input_dim (with relu). The issue is discussed in papers like Xavier Glorot "Understanding the difficulty of training deep feedforward neural networks" or Kaiming He "Delving Deep into Rectifiers.."

Like many, mlx uses a uniform distribution to approximate a normal distribution. In this case, mlx uses a uniform distribution with bounds sqrt(1 / input_dim). Mathematically the variance of a uniform distribution with bounds a b is (b - a)^2 / 12. To maintain a zero mean, we set a = -b. To achieve variance = 1 / input_dim = b^2 / 3. Requires bound b = sqrt(3 / input_dim).

As a comparison, look at github pytorch/torch/nn/modules/linear.py the func reset_parameters() uses init.kaiming_uniform_(). Then within pytorch/torch/nn/init.py kaiming_uniform_() multiplies the bound by sqrt(3).

MLX should make a similar adjustment. General training accuracy should improve as a result.

angeloskath:
Indeed our initialization is `sqrt(3)` or someone could even say `sqrt(6)` off from the standard kaiming uniform. To be honest I think at some point it was using normal instead of uniform which made a bit more sense but still.

Personally I don't think this matters as much. Since these analyses both the normalization layers and optimizers have actually changed significantly that I don't think a factor of `~1.5` is going to have a big impact. If people disagree we could fix the default init fairly easily.

AvikantSrivastava:
Hey @awni @angeloskath @ag487 I have raised the PR https://github.com/ml-explore/mlx/pull/2071/files
That adds a parameter for `kaiming_init` 
```
layer = nn.Linear(input_dims=4, output_dims=8, kaiming_init=True)
```

can you folks take a look?

angeloskath:
So the default behavior is actually exactly PyTorch's see the last comment in the PR. I should have checked whether PT is using Kaiming uniform which they aren't.

Closing the issue but as I mentioned in the PR feel free to open an issue specific to any initialization related problems you may encounter.

PR #2071, opened 2025-04-12
Adjust nn layer init with kaiming_init

AvikantSrivastava:
## Proposed changes

Issue: https://github.com/ml-explore/mlx/issues/2001

This PR adds support  Kaiming uniform initialization in the Linear and Bilinear layers of MLX. This brings MLX in line with initialization practices used by PyTorch and others, especially beneficial for models using ReLU activations.

## Checklist

### DIFF
--- a/python/mlx/nn/layers/linear.py
+++ b/python/mlx/nn/layers/linear.py
@@ -55,7 +55,6 @@ class Linear(Module):
         super().__init__()
         init_fn = init.he_uniform()
         self.weight = init_fn(mx.zeros((output_dims, input_dims)))
-
         if bias:
             self.bias = init_fn(mx.zeros((output_dims,)))
 
@@ -108,7 +107,6 @@ class Bilinear(Module):
         super().__init__()
         init_fn = init.he_uniform()
         self.weight = init_fn(mx.zeros((output_dims, input_dims)))
-
         if bias:
             self.bias = init_fn(mx.zeros((output_dims,)))
 
--- a/python/mlx/nn/layers/linear.py
+++ b/python/mlx/nn/layers/linear.py
@@ -45,9 +45,17 @@ class Linear(Module):
           not use a bias. Default is ``True``.
     """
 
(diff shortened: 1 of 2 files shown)


awni:
I don't we should add the kwarg. Just change the initialization. If someone want a different initialization they can use an [initializer](https://ml-explore.github.io/mlx/build/html/python/nn/init.html)

AvikantSrivastava:
> I don't we should add the kwarg. Just change the initialization. If someone want a different initialization they can use an [initializer](https://ml-explore.github.io/mlx/build/html/python/nn/init.html)

@awni Re requesting for a review, I have made the changes that you suggested

AvikantSrivastava:
@awni Can you take a look? 
:)

angeloskath:
Hi @AvikantSrivastava, I think we should close this PR. I am sorry if you wasted time on this.

This started with issue #2001 as you mention. At the time I kinda mentioned that it is not important but didn't even check if PT indeed does use Kaiming uniform initialization. In fact, PT uses exactly the same initialization as we do. The simple check is

```python
import torch
import mlx.nn as nn

D = 100
print(torch.nn.Linear(D, D).weight.std(), nn.Linear(D, D).weight.std())
```

but looking at PT's code they actually use `a=5**0.5` and they even have a comment that it is the same as simple uniform. https://github.com/pytorch/pytorch/blob/1eba9b3aa3c43f86f4a2c807ac8e12c4a7767340/torch/nn/modules/linear.py#L114

If you want to further check this you can walk through the code where `a` is used to compute a `gain` for Kaiming uniform  as `gain = (2 / (1 + a**2))**0.5 = 1/3**0.5` which is precisely the difference we were talking about in the issue.

Given the above I am closing the PR. If you do encounter initialization related issues, please open an issue but you can always just initialize the model by applying any custom init function.

AvikantSrivastava:
This makes a lot of sense @angeloskath, Thanks for the clarification :)

PR #2073, opened 2025-04-13
Introduce macros for dispatching dynamic dtypes as static types

zcbenz:
This PR ports some macros originated from PyTorch that were used for dispatching runtime dtypes to compile time C++ types. The code was modified from [the modified version in ExecuTorch](https://github.com/pytorch/executorch/blob/a664d7b3616e5307cd1a770895b805773b0e29c9/runtime/core/exec_aten/util/scalar_type_util.h), which has some nice improvements that enable nested dispatching.

For example, code that does:

```c++
  switch (dtype) {
    case bool_:
      op<bool>();
      break;
    case uint8:
      op<uint8_t>();
      break;
    ...
  }
```

can be simplified with:

```c++
MLX_SWITCH_ALL_TYPES(dtype, CTYPE, [&]() {
  op<CTYPE>();
});
```

Code that used to need several functions for dispatching:

```c++
template <typename A>
void dispatch_b(dtype b) {
  switch (b) {
    case bool_:
      op<A, bool>();
      break;
    ...
  }
}

void dispatch_ab(dtype a, dtype b) {
  switch (a) {
    case bool_:
      dispatch_b<bool>(b);
      break;
    ...
  }
}
```

can be replaced with

```c++
MLX_SWITCH_ALL_TYPES(a, CTYPE_A, [&]() {
  MLX_SWITCH_ALL_TYPES(b, CTYPE_B, [&]() {
    op<CTYPE_A, CTYPE_B>();
  });
});
```

It also has a version for checking unhandled dtype automatically:

```c++
MLX_SWITCH_INT_TYPES_CHECKED(dtype, "iinfo", CTYPE, [&]() {
  set_iinfo_limits<CTYPE>(min, max);
});
```

which throws "Unhandled dtype xxx for iinfo".

The code in `mlx/utils.cpp` are updated to use the macros.

### DIFF
--- a/mlx/dtype_utils.h
+++ b/mlx/dtype_utils.h
@@ -1,0 +1,207 @@
+// Copyright © 2025 Apple Inc.
+// Copyright © Meta Platforms, Inc. and affiliates.
+//
+// This source code is licensed under the BSD-style license found in
+// https://github.com/pytorch/executorch/blob/main/LICENSE
+//
+// Forked from
+// https://github.com/pytorch/executorch/blob/main/runtime/core/exec_aten/util/scalar_type_util.h
+
+#pragma once
+
+#include "mlx/dtype.h"
+
+#include <fmt/format.h>
+
+namespace mlx::core {
+
+// Return string representation of dtype.
+const char* dtype_to_string(Dtype arg);
+
+// Macros that iterate across different subsets of Dtypes.
+//
+// For all of these macros, the final `_` parameter is the name of another macro
+// that takes two parameters: the name of a C type, and the name of the
+// corresponding Dtype enumerator.
+//
+// Note that these macros should use fully-qualified namespaces (starting with
+// `::`) to ensure that they can be called safely in any arbitrary namespace.
+#define MLX_FORALL_INT_TYPES(_) \
+  _(uint8_t, uint8)             \
+  _(uint16_t, uint16)           \
+  _(uint32_t, uint32)           \
+  _(uint64_t, uint64)           \
+  _(int8_t, int8)               \
+  _(int16_t, int16)             \
+  _(int32_t, int32)             \
+  _(int64_t, int64)
+
+#define MLX_FORALL_FLOAT_TYPES(_) \
+  _(float16_t, float16)           \
+  _(float, float32)               \
+  _(double, float64)              \
+  _(bfloat16_t, bfloat16)
+
+// Calls the provided macro on every Dtype, providing the C type and the
+// Dtype name to each call.
+//
+// @param _ A macro that takes two parameters: the name of a C type, and the
+//          name of the corresponding Dtype enumerator.
+#define MLX_FORALL_DTYPES(_) \
+  MLX_FORALL_INT_TYPES(_)    \
+  MLX_FORALL_FLOAT_TYPES(_)  \
+  _(bool, bool_)             \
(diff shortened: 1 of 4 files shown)


mlx/dtype_utils.h
```diff
@@ -0,0 +1,214 @@
...
+//
+// MLX_SWITCH_ALL_TYPES(input.dtype(), CTYPE, [&]() {
+//   output.data<CTYPE>[0] = input.data<CTYPE>[0];
+// });
+//
+// Note that these can be nested as well:
+//
+// MLX_SWITCH_ALL_TYPES(input.dtype(), CTYPE_IN, [&]() {
+//   MLX_SWITCH_ALL_TYPES(output.dtype(), CTYPE_OUT, [&]() {
+//     output.dat<CTYPE_OUT>[0] = input.data<CTYPE_IN>[0];
```
awni, line 178:
```suggestion
//     output.data<CTYPE_OUT>[0] = input.data<CTYPE_IN>[0];
```

mlx/utils.cpp
```diff
@@ -277,50 +248,7 @@ std::ostream& operator<<(std::ostream& os, const Dtype::Kind& k) {
...
-      print_array<float>(os, a);
-      break;
-    case float64:
-      print_array<double>(os, a);
-      break;
-    case complex64:
-      print_array<complex64_t>(os, a);
-      break;
-  }
+  MLX_SWITCH_ALL_TYPES(a.dtype(), CTYPE, [&]() { print_array<CTYPE>(os, a); });
```
awni, line 251: It might be better to take the function directly rather than a lambda. I think (at least with Apple Clang) there is some non-trivial binary size overhead to making lambdas and it's also a bit simpler to do something like:

```suggestion
  MLX_SWITCH_ALL_TYPES(a.dtype(), CTYPE, print_array<CTYPE>(os, a));
```

If for some reason you really want to use a lambda you could do:

```
  MLX_SWITCH_ALL_TYPES(a.dtype(), CTYPE, [&]() { print_array<CTYPE>(os, a); }());
```

zcbenz: The suggested does look much better, I have updated the macros to take statement instead of lambda.

mlx/dtype_utils.h
```diff
@@ -0,0 +1,215 @@
...
+#define MLX_INTERNAL_SWITCH_CHECKED(TYPE, NAME, ...)                    \
+  [&] {                                                                 \
+    switch (TYPE) {                                                     \
+      __VA_ARGS__                                                       \
+      default:                                                          \
+        throw std::invalid_argument(fmt::format(                        \
+            "Unhandled dtype %s for %s", dtype_to_string(TYPE), NAME)); \
+    }                                                                   \
+  }()
+
```
awni, line 109: Not clear to me why this needs to be wrapped in a lambda? Am I missing something or can we just remove that?

zcbenz: The original design was to enable something like `xxx = MLX_SWITCH_TYPES(...)`, I have updated the PR to remove it.

awni: I see, thanks for the explanation.

awni (approved):
This looks really nice. We should adopt it in more places, especially the CPU back-end where we have lot's of verbose switch cases.

awni:
Will merge after the tests clear!

Issue #1004, opened 2024-04-18
[Feature] Support Short-Time Fourier Transforms


Rifur13:
Short-Time Fourier Transforms (STFT) and its inverse are used extensively in signal processing.

Function definition will match scipy:
https://docs.scipy.org/doc/scipy/reference/generated/scipy.signal.stft.html
https://docs.scipy.org/doc/scipy/reference/generated/scipy.signal.istft.html

Rifur13:
You can assign this to me.

awni:
Cool.. if we include this, I wonder where it would go? I'm not opposed to including it in core or a sub-package but I think there is a decision to be made there.

I see jax implemented it in a `scipy.signal` subpackage https://jax.readthedocs.io/en/latest/_autosummary/jax.scipy.signal.stft.html, and PyTorch implemented it at the top level https://pytorch.org/docs/stable/generated/torch.stft.html

Also our Whisper implementation already has an STFT in MLX you might find useful: https://github.com/ml-explore/mlx-examples/blob/main/whisper/whisper/audio.py#L104-L127

Rifur13:
I think adding them in core is best - they’re fundamental algorithms and used frequently.  A separate `signal` sub-package makes sense if you anticipate adding more functions from `scipy.signal` in he near future. (https://jax.readthedocs.io/en/latest/jax.scipy.html#module-jax.scipy.signal)

Thanks for pointing me to the Whisper implementation, I haven't seen in yet

tqtifnypmb:
One tiny issue about [`stft` of mlx-example/whisper](https://github.com/ml-explore/mlx-examples/blob/main/whisper/whisper/audio.py#L104-L127), the `noverlap` parameter does not actually represent the overlap of frames; instead, it is semantically equivalent to `hop_length` in `torch.stft`. 

I found it a little bit misleading while implementing `istft`. @awni

awni:
Is it not the same as `noverlap` in the Scipy implementation? https://docs.scipy.org/doc/scipy/reference/generated/scipy.signal.stft.html 

I think that's a good reference to follow.

Rifur13:
Still working on this CL, shouldn't be much longer.

``` t = (x.size - nperseg + noverlap) // noverlap``` does look wrong though, maybe that's where the confusion is. The denominator should be the number of steps.

tqtifnypmb:
[...]

PR #2074, opened 2025-04-13
Added Short time fourier transform and ISTFT implementations

ParamThakkar123:
## Proposed changes

This PR introduces the implementations of Short Time and Inverse Short Time Fourier Transforms in mlx as addressed in issue #1004 

## Checklist

### DIFF
--- a/python/mlx/stft.py
+++ b/python/mlx/stft.py
@@ -1,0 +1,115 @@
+import mlx.core as mx
+import numpy as np
+
+def stft(
+    x: mx.array,
+    n_fft: int = 2048,
+    hop_length: int = None,
+    win_length: int = None,
+    window: mx.array = None,
+    center: bool = True,
+    pad_mode: str = "reflect",
+    normalized: bool = False,
+    onesided: bool = True,
+    return_complex: bool = True,
+) -> mx.array:
+
+    if hop_length is None:
+        hop_length = n_fft // 4
+        
+    if win_length is None:
+        win_length = n_fft
+        
+    if window is None:
+        window = mx.ones(win_length)
+
+    if win_length < n_fft:
+        pad_left = (n_fft - win_length) // 2
+        pad_right = n_fft - win_length - pad_left
+        window = mx.pad(window, [(pad_left, pad_right)])
+
+    if center:
+        pad_width = n_fft // 2
+        x = mx.pad(x, [(pad_width, pad_width)], mode=pad_mode)
+
+    n_frames = 1 + (x.shape[0] - n_fft) // hop_length
+
+    frames = mx.stack([
+        x[i * hop_length : i * hop_length + n_fft] * window
+        for i in range(n_frames)
+    ])
+
+    stft = mx.fft.fft(frames, n=n_fft, axis=-1)
+    
+    if normalized:
+        stft = stft / mx.sqrt(n_fft)
+        
+    if onesided:
+        stft = stft[..., :n_fft//2 + 1]
+        
+    if not return_complex:
+        stft = mx.stack([stft.real, stft.imag], axis=-1)
+        
+    return stft
+
+def istft(
+    stft_matrix: mx.array,
+    hop_length: int = None,
+    win_length: int = None,
+    window: mx.array = None,
+    center: bool = True,
+    length: int = None,
+    normalized: bool = False,
+    onesided: bool = True,
+) -> mx.array:
+    if hop_length is None:
+        hop_length = stft_matrix.shape[-2] // 4
+        
+    if win_length is None:
+        win_length = stft_matrix.shape[-2]
+        
+    if window is None:
+        window = mx.ones(win_length)
+        
+    if win_length < stft_matrix.shape[-2]:
+        pad_left = (stft_matrix.shape[-2] - win_length) // 2
+        pad_right = stft_matrix.shape[-2] - win_length - pad_left
+        window = mx.pad(window, [(pad_left, pad_right)])
+        
+    if stft_matrix.shape[-1] == 2:
+        stft_matrix = stft_matrix[..., 0] + 1j * stft_matrix[..., 1]
+        
+    if onesided:
+        n_fft = 2 * (stft_matrix.shape[-1] - 1)
+        full_stft = mx.zeros((*stft_matrix.shape[:-1], n_fft), dtype=stft_matrix.dtype)
+        full_stft[..., :stft_matrix.shape[-1]] = stft_matrix
+        full_stft[..., stft_matrix.shape[-1]:] = mx.conj(stft_matrix[..., -2:0:-1])
+        stft_matrix = full_stft
+        
+    frames = mx.fft.ifft(stft_matrix, n=stft_matrix.shape[-1], axis=-1)
+    
+    if normalized:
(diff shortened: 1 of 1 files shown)


ParamThakkar123:
I referred librosa's documentation on stft to write my implementation: 
https://librosa.org/doc/main/generated/librosa.stft.html

python/mlx/stft.py
```diff
@@ -0,0 +1,116 @@
+import mlx.core as mx
+import numpy as np
```
angeloskath, line 2: Numpy is not used anywhere.

angeloskath (commented):
This looks well done from a first pass.

The main question to answer is whether this should be in the main mlx vs on a per library/use case implementation. I also think this shouldn't be in the top level of the package. We should come up with a better location if we are to merge this (I don't have a very good suggestion atm).

Regarding the PR content, I would definitely add docstrings and tests for the two functions.

ParamThakkar123:
@angeloskath would adding to the code to the utils.py be a better option ?

ParamThakkar123:
> One option is to go the route of cloning scipy in MXL kind of like Jax (e.g. put this in a new package `mlx.scipy.signal`). It's a big package so I'm not really sure that's a good idea.
> 
> Another option is to put this in `mlx.core.fft`, but that would require a C++ implementation instead of a Python one (which may actually be handy to be honest because then we get it in all the APIs with simple bindings).

Okay, I would implement a C++ version of this and add it to mlx.core.fft

ParamThakkar123:
@awni should I create a separate namespace for this ??

ParamThakkar123:
in fft.cpp

ParamThakkar123:
@awni @angeloskath I have added a C++ implementation of stft in mlx.core.fft i.e. fft.cpp file. Can you please review it ??

awni:
The next step is to remove the Python one and make a binding in python/src/fft.cp

ParamThakkar123:
@awni I have added bindings and also removed the python implementation

mlx/fft.cpp
```diff
@@ -190,3 +192,133 @@ array irfftn(const array& a, StreamOrDevice s /* = {} */) {
 }
 
 } // namespace mlx::core::fft
+
+namespace mlx::core::stft {
```
awni, line 196: We don't need a new namespace for this. Just put in the `fft` namespace.

mlx/fft.cpp
```diff
@@ -190,3 +192,133 @@ array irfftn(const array& a, StreamOrDevice s /* = {} */) {
...
+
+  array win = (window.size() == 0) ? ones({win_length}, float32, s) : window;
+
+  if (win_length < n_fft) {
+    int pad_left = (n_fft - win_length) / 2;
+    int pad_right = n_fft - win_length - pad_left;
+
+    array left_pad = zeros({pad_left}, float32, s);
+    array right_pad = zeros({pad_right}, float32, s);
+    win = concatenate({left_pad, win, right_pad}, 0, s);
```
awni, line 222: Use `mx::pad`

mlx/fft.cpp
```diff
@@ -190,3 +192,133 @@ array irfftn(const array& a, StreamOrDevice s /* = {} */) {
...
+  }
+
+  array padded_x = x;
+  if (center) {
+    int pad_width = n_fft / 2;
+
+    array left_pad = zeros({pad_width}, x.dtype(), s);
+    array right_pad = zeros({pad_width}, x.dtype(), s);
+    padded_x = concatenate({left_pad, padded_x, right_pad}, 0, s);
```
awni, line 231: Use `mx::pad`

mlx/fft.cpp
```diff
@@ -190,3 +192,133 @@ array irfftn(const array& a, StreamOrDevice s /* = {} */) {
...
+  }
+
+  int n_frames = 1 + (padded_x.shape(0) - n_fft) / hop_length;
+
+  std::vector<array> frames;
+  for (int i = 0; i < n_frames; ++i) {
+    array frame =
+        slice(padded_x, {i * hop_length}, {i * hop_length + n_fft}, s);
+    frames.push_back(multiply(frame, win, s));
+  }
```
awni, line 241: Use `as_strided` and a regular multiplication.

mlx/fft.cpp
```diff
@@ -190,3 +192,133 @@ array irfftn(const array& a, StreamOrDevice s /* = {} */) {
...
+
+  array stacked_frames = stack(frames, 0, s);
```
awni, line 243: Remove.

mlx/fft.cpp
```diff
@@ -190,3 +192,133 @@ array irfftn(const array& a, StreamOrDevice s /* = {} */) {
...
+  }
+
+  if (onesided) {
+    stft_result = slice(stft_result, {}, {n_fft / 2 + 1}, s);
+  }
+
+  return stft_result;
+}
+
+array istft(
```
awni, line 259: Same comments here as for `stft`

ParamThakkar123:
@awni can you please review this PR ?

ParamThakkar123:
I have added all the fixes and tests

mlx/fft.cpp
```diff
@@ -1,6 +1,9 @@
 // Copyright © 2023 Apple Inc.
+
+#include <cmath>
 #include <numeric>
 #include <set>
+#include <sstream>
```
awni, line 6: Why add that? Doesn't look used.

python/src/fft.cpp
```diff
@@ -459,4 +459,100 @@ void init_fft(nb::module_& parent_module) {
...
+            length,
+            normalized,
+            s);
+      },
+      "stft_matrix"_a,
+      "hop_length"_a = -1,
+      "win_length"_a = -1,
+      "window"_a = mx::array(),
+      "center"_a = true,
+      "length"_a = -1,
```
awni, line 540: -1 is not the right value for optional. If it is optional and we want a default, use `none`.

mlx/fft.cpp
```diff
@@ -188,4 +191,122 @@ array irfftn(const array& a, StreamOrDevice s /* = {} */) {
   return fft_impl(a, true, true, s);
 }
 
-} // namespace mlx::core::fft
+array stft(
```
awni, line 194: Don't these need to be defined in the header? Does the code currently compile?

I highly recommend building and running the tests locally. It will shorten the number of times we have to go back and forth on review.

ParamThakkar123:
@awni I fixed the headers. Compiled and tested it and it works fine on my end.

ParamThakkar123:
@awni @angeloskath

ParamThakkar123:
@awni can you please review this PR ?

ParamThakkar123:
@awni @angeloskath ???

tests/fft_tests.cpp
```diff
@@ -309,7 +309,81 @@ TEST_CASE("test fft grads") {
   CHECK_EQ(vjp_out.shape(), Shape{5, 5});
 }
 
-TEST_CASE("test fftshift and ifftshift") {
+<<<<<<< HEAD
```
yrom, line 312: conflicts were not resolved correctly.

PR #2075, opened 2025-04-13
CUDA backend: backbone

zcbenz:

This PR adds backbone facilities for CUDA backend, and implements the `Arange` primitive.

Notes:

* ~I only tested building with CUDA 11.7, for machines with CUDA 12 there are likely building errors.~
* The code uses `mxcuda` as namespace instead of `cuda`, because when using `cuda::xxx` inside the `mlx::core` namespace, for many cases C++ would only search `xxx` in the global `::cuda` namespace imported by cuda headers and then complain symbol not found.
* There are 2 types of `Event` implementations: one is fast but only synchronizes GPU tasks, one is slow but can synchronize between CPU/GPU. It is dynamically chosen based on how `Event` is used.
* A mlx `Stream` currently only maps to a single cuda stream, but the interface is designed to support multiple streams in future.


mlx/backend/cuda/kernels/fp16_math.cuh
```diff
@@ -0,0 +1,103 @@
...
+      typename = cuda::std::enable_if_t<is_arithmetic_except<T, HALF>>> \
+  __forceinline__ __device__ HALF operator OP(HALF x, T y) {            \
+    return FLOAT2HALF(HALF2FLOAT(x) OP static_cast<float>(y));          \
+  }                                                                     \
+  template <                                                            \
+      typename T,                                                       \
+      typename = cuda::std::enable_if_t<is_arithmetic_except<T, HALF>>> \
+  __forceinline__ __device__ HALF operator OP(T x, HALF y) {            \
+    return FLOAT2HALF(static_cast<float>(x) OP HALF2FLOAT(y));          \
+  }
```
jagrit06, line 63: Should the output type here be `__half` ? If the operation is `float * half` the result should be `float` - or is this particularly meant just for the Arange function for now ?

zcbenz: Good question, I think it should return float for float * half. It was defined because some ops are doing arithmetics between half types and numeric literals, I will try making the overloads do proper promotions.

zcbenz: I agree it makes sense to get rid of the overloads that introduce type promotions like `float * half`. How do you feel about the safer overloads like `int * half`? In the case of Arange it is actually the `i * step` part that needs the overload, we can do `start + T(i) * step`, but I think it does not harm having `i * step` implicitly converting to half.

zcbenz:
Updated the PR to fix building for CUDA 12 and remove the overloads for `half x float` (overloads for `half x int` are kept).

zcbenz:
Updated the PR with 2 changes:

* Avoid freeing buffers in cpu streams to avoid dead locks (https://github.com/ml-explore/mlx/issues/2099).
* Remove the usages of `cudaLaunchHostFunc` API and use our own worker thread to free buffers.

mlx/CMakeLists.txt
```diff
@@ -48,6 +48,8 @@ add_subdirectory(${CMAKE_CURRENT_SOURCE_DIR}/io)
 
 if(MLX_BUILD_METAL)
   add_subdirectory(${CMAKE_CURRENT_SOURCE_DIR}/backend/metal)
+elseif(MLX_BUILD_CUDA)
+  add_subdirectory(${CMAKE_CURRENT_SOURCE_DIR}/backend/cuda)
 else()
   add_subdirectory(${CMAKE_CURRENT_SOURCE_DIR}/backend/no_metal)
```
awni, line 54: I feel like we should change the name of that directory from `no_metal` to `no_gpu`. It would be more fitting / clearer in the context of the CUDA backend.

mlx/backend/cuda/metal.cpp
```diff
@@ -0,0 +1,62 @@
...
+
+#include "mlx/backend/metal/metal.h"
+#include "mlx/backend/cuda/device.h"
+#include "mlx/backend/metal/metal_impl.h"
+#include "mlx/primitives.h"
+#include "mlx/scheduler.h"
+
+#include <nvtx3/nvtx3.hpp>
+
+namespace mlx::core::metal {
```
awni, line 11: I think it's worth generalizing this part of the code. 

For example, checking `metal::is_available` doesn't really make sense for the CUDA back-end.

Here's some options:

- Add a `mx::is_available(Device)` which checks on the given device. 
- For the GPU device we have a shared header that lives in `backend/gpu` and in the namespace `gpu`. 
- The implementations then go in `metal`, `cuda`,  and `no_gpu` respectively.
- We could potentially mark `metal::is_available` as deprecated as well.. 

In general a good aim is to work towards having device-general code. So in user-land it would be really nice to get rid of all those calls to `metal.is_available()` and instead check for the device.  And if we need back-end specific availability checks have them as internal API.

awni: Another option is to put these in the `mx_cuda` namespace and duplicate some of the headers. I think this is suboptimal though because then we need to duplicate stuff in calling code.

mlx/backend/metal/CMakeLists.txt
```diff
@@ -121,5 +121,7 @@ endif()
 
 add_subdirectory(${CMAKE_CURRENT_SOURCE_DIR}/kernels)
 
-target_compile_definitions(mlx
-                           PRIVATE METAL_PATH="${MLX_METAL_PATH}/mlx.metallib")
+target_compile_definitions(
+  mlx
+  PUBLIC MLX_USE_METAL
```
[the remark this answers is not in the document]
zcbenz: At some point I needed it to disable some metal-only tests, I will remove it.

stemann:
Is the `mx` in the `mxcuda` name (namespace name) a bit of a Pythonic legacy (`mx`/`np`)? Seems a bit strange when the `mlx` namespace includes the`l` in `mlx`.

awni:
> Is the mx in the mxcuda name (namespace name) a bit of a Pythonic legacy (mx/np)? Seems a bit strange when the mlx namespace includes the in mlx.

I guess we conventionally alias `mlx::core` with `mx`.  That namespace is not part of the public API so I think it's ok as is. If you have better suggestions for naming it though let us know! (Note the obvious `mlx::core::cuda` doesn't work due to name clashing with the `cuda` namespace`)

awni:
@zcbenz let me know when this is rebased on main and I will review again.

stemann:
> > Is the mx in the mxcuda name (namespace name) a bit of a Pythonic legacy (mx/np)? Seems a bit strange when the mlx namespace includes the in mlx.
> 
> I guess we conventionally alias `mlx::core` with `mx`. That namespace is not part of the public API so I think it's ok as is. If you have better suggestions for naming it though let us know! (Note the obvious `mlx::core::cuda` doesn't work due to name clashing with the `cuda` namespace`)

`mx` is only used in the Python examples/API, right?

I would just suggest to name it `mlx_cuda` or `mlxcuda`.

awni:
> mx is only used in the Python examples/API, right?

By convention whenever we use the `mlx::core` namespace in C++ we also alias it to `mx`. See e.g. https://github.com/ml-explore/mlx/blob/e496c5a4b4d06e07ad204bdf3226df4f1bc3e259/examples/cpp/linear_regression.cpp#L13

awni:
I see your point @stemann ... maybe worth thinking on some other options for the `mx_cuda` namespace:

- `mlx::core::mlxcuda`
- `mlx::core::cu` (cu for cuda)
- `mlx::core::nv` or `mlx::core::nvda` (for nvidia)

That said I think this is a more important question if we plan to expose the namespace in the public API (which we currently don't).

zcbenz:
Rebased the PR.

mlx/backend/cuda/primitives.cu
```diff
@@ -0,0 +1,185 @@
...
+  nvtx3::scoped_range r("Arange::eval_gpu");
+  assert(inputs.size() == 0);
+  out.set_data(allocator::malloc(out.nbytes()));
+  if (out.size() == 0) {
+    return;
+  }
+  auto& s = stream();
+  auto& encoder = mxcuda::get_command_encoder(s);
+  encoder.set_output_array(out);
+  encoder.launch_kernel([&, this](cudaStream_t stream) {
```
awni, line 30: It looks like the `launch_kernel` calls this task synchronously right? So the lambda will be called before we leave the outer `eval_gpu`? Just checking cause we capture `out` by reference here.

awni: I'm also wondering if this level of indirection is necessary.. but I don't fully understand what the `launch_kernel` is doing.

It looks like it gets the stream, dispatches the kernel and then checks if there are any existing errors?

Is the purpose here also that we want to register with the device that a kernel has been launched so it's important to control that?

zcbenz: The lambda is always immediately called inside `launch_kernel`.

The `launch_kernel` does following things:
1. Get the cuda stream, currently there is only one cuda stream per mlx stream, but in future it could schedule one from a pool based on parameters passed to `launch_kernel`.
2. Call `cudaSetDevice` to set the device that kernel runs on. This is mandatory to support multi-device.
3. Call `fun()` which actually launches the kernel.
4. Check if the launch has succeeded. 
5. Set a flag indicating a kernel has launched, which is used later to determine whether we should schedule the completion callbacks to run after kernel's completion.

So we have some works to do both before and after the kernel launch, which led to the design.

A way to avoid using lambda is doing something like this:

```c++
{
  auto scope = device.prepare_kernel_launch();
  kernel<<<1, 1, 0, scope.stream>>>();
}
```

Though it could add more complexity.

awni:
I vote `mlx::core::cu`.  Kind of matches things like `cuBLAS`, `cuDNN`.. etc. `mlx::core::mlx_cuda` seems a little to redundant.

zcbenz:
Updated PR to rebase on main (#2145) and rename `mxcuda` to `cu`.

awni:
I think we should go ahead and merge this. 
- It's very well decoupled from the rest of MLX and so I'm not too worried about interference / complexity there.
- Let's not hesitate to make big changes if needed on the design. But overall it looks quite nice!

awni (approved):
Awesome awesome!! We're getting there! 

I will merge it once the tests clear.

PR #2079, opened 2025-04-17
Fix Python Garbage Collection edge case

awni:
See https://github.com/wjakob/nanobind/pull/1015 for context.


angeloskath (approved):
Looks good, thanks!

Issue #2077, opened 2025-04-15
[BUG] shift operations on int8 return int16


acsweet:
**Describe the bug**
`right_shift` and `left_shift` on `int8` arrays returns `int16` result.

**To Reproduce**

```python
import mlx.core as mx

dtypes = [mx.int8, mx.int16, mx.int32, mx.int64, mx.uint8, mx.uint16, mx.uint32, mx.uint64]
for dtype in dtypes:
    x = mx.ones((1,), dtype=dtype)
    y = mx.ones((1,), dtype=dtype)
    right_shift = mx.right_shift(x, y)
    if x.dtype != right_shift.dtype:
        print('result dtype mismatch for right_shift:', x.dtype, y.dtype, right_shift.dtype)
    left_shift = mx.left_shift(x, y)
    if x.dtype != left_shift.dtype:
        print('result dtype mismatch for left_shift:', x.dtype, y.dtype, left_shift.dtype)

# result dtype mismatch for right_shift: mlx.core.int8 mlx.core.int8 mlx.core.int16
# result dtype mismatch for left_shift: mlx.core.int8 mlx.core.int8 mlx.core.int16
```

**Expected behavior**
Useing right_shift or left_shift on an `int8` results in an `int16` result, for all other dtypes it returns a result the same dtype as the inputs. I would expect the same behavior here.

**Desktop (please complete the following information):**
 - OS Version: 15.2
 - Version: 0.24.2

**Additional context**
Easy to work around for now, but I noticed this while working on bit shift ops for the Keras mlx backend implementation (https://github.com/keras-team/keras/issues/19571).

ParamThakkar123:
I would like to take on this issue

ParamThakkar123:
@awni On it

PR #2080, opened 2025-04-17
Fixed shift operations issue

ParamThakkar123:
## Proposed changes

Please include a description of the problem or feature this PR is addressing. If there is a corresponding issue, include the issue #2077 .

## Checklist

### DIFF
--- a/tests/ops_tests.cpp
+++ b/tests/ops_tests.cpp
@@ -3874,3 +3874,30 @@ TEST_CASE("test contiguous") {
   CHECK(x.flags().col_contiguous);
   CHECK_EQ(x.strides(), decltype(x.strides()){1, 2});
 }
+
+TEST_CASE("test bitwise shift operations") {
+  std::vector<Dtype> dtypes = {
+      int8, int16, int32, int64, uint8, uint16, uint32, uint64};
+
+  for (const auto % dtype : dtypes) {
+    array x = full({4}, 1, dtype);
+    array y = full({4}, 2, dtype);
+
+    auto left_shift_result = left_shift(x, y);
+    CHECK_EQ(left_shift_result.dtype(), dtype);
+    CHECK(array_equal(left_shift_result, array({4, 4, 4, 4}, dtype))
+              .item<bool>());
+
+    auto right_shift_result = right_shift(full({4}, 4, dtype), y);
+    CHECK_EQ(right_shift_result.dtype(), dtype);
+    CHECK(array_equal(right_shift_result, full({4}, 1, dtype)).item<bool>());
+  }
+
+  array x = array({127, -128}, int8); // Maximum and minimum int8 values
+  array y = array({1, 1}, int8); // Shift by 1
+  auto left_shift_result = left_shift(x, y);
+  auto right_shift_result = right_shift(x, y);
+
+  CHECK(array_equal(left_shift_result, array({-2, 0}, int8)).item<bool>());
+  CHECK(array_equal(right_shift_result, array({63, -64}, int8)).item<bool>());
+}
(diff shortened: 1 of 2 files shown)


awni:
Can you please add a test for this? Also I don't think the fix is correct. We want the behavior to be:

- If `result_type(a, b)` is bool -> output type is uint8
- Else output type is `result_type(a, b)`.

ParamThakkar123:
@awni added the test cases and fixes

tests/ops_tests.cpp
```diff
@@ -3874,3 +3874,30 @@ TEST_CASE("test contiguous") {
   CHECK(x.flags().col_contiguous);
   CHECK_EQ(x.strides(), decltype(x.strides()){1, 2});
 }
+
+TEST_CASE("test bitwise shift operations") {
+  std::vector<Dtype> dtypes = {
+      int8, int16, int32, int64, uint8, uint16, uint32, uint64};
+
+  for (const auto % dtype : dtypes) {
```
awni, line 3882: Does that compile?

ParamThakkar123: No it won't actually 😅 . Made a typo after testing 😅

tests/ops_tests.cpp
```diff
@@ -3874,3 +3874,30 @@ TEST_CASE("test contiguous") {
...
+    CHECK(array_equal(right_shift_result, full({4}, 1, dtype)).item<bool>());
+  }
+
+  array x = array({127, -128}, int8);
+  array y = array({1, 1}, int8);
+  auto left_shift_result = left_shift(x, y);
+  auto right_shift_result = right_shift(x, y);
+
+  CHECK(array_equal(left_shift_result, array({-2, 0}, int8)).item<bool>());
+  CHECK(array_equal(right_shift_result, array({63, -64}, int8)).item<bool>());
```
awni, line 3902: Why do you have a separate test case for `int8` here?

ParamThakkar123: I thought that might be necessary for this particular issue.

ParamThakkar123: But would still need your thoughts on it's necessity @awni

ParamThakkar123: > I thought that might be necessary for this particular issue.

in order to check if they still return `int8`

awni: Isn't that checked in the previous loop? You are missing the check for `bool`.

ParamThakkar123: yeah, it's checked but I am also trying to check for overflows if they happen in case of int8 because it has smaller range overall

ParamThakkar123: Should I add the check for bools in the loop itself @awni ??

awni: No I would add that separate since the output type should be `uint8`

tests/ops_tests.cpp
```diff
@@ -3874,3 +3874,41 @@ TEST_CASE("test contiguous") {
...
+
+  array x_bool = full({4}, true, bool_);
+  array y_bool = full({4}, 1, int32);
```
awni, line 3905: This must be a typo.. I don't see how these tests could pass if `y_bool` has type `int32`.

ParamThakkar123:
yes @awni actually it was a typo 😅 . Fixed that

awni (approved):
Looks good! Thanks for fixing that.

ParamThakkar123:
Thanks @awni !!

ParamThakkar123:
Moving on to fixing the next issues!

PR #2087, opened 2025-04-19
Use unordered map for kwargs in export/import

awni:
Use unordered kwargs for import/export to simplify bindings to MLX C.

### DIFF
--- a/python/src/export.cpp
+++ b/python/src/export.cpp
@@ -16,7 +16,7 @@ namespace mx = mlx::core;
 namespace nb = nanobind;
 using namespace nb::literals;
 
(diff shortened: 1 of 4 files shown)


python/src/export.cpp
```diff
@@ -31,7 +30,7 @@ validate_and_extract_inputs(
     }
   };
   std::vector<mx::array> args_;
-  std::map<std::string, mx::array> kwargs_;
+  mx::Kwargs kwargs_;
```
angeloskath, line 33: Nit: you can also do `mx::Args` instead of vector above.

angeloskath (approved):
👍

PR #2090, opened 2025-04-19
SDPA fixes

### LINKED ISSUE #1604: [Enhancement Request] SDFA does not support head dimension of size 192 (capped at 128)
**Describe the bug**
SDFA currently does not support head dimension outside of 64, 96, and 128. 

**To Reproduce**
Fused attention falls back to regular operation when head dimension is not in (64, 96, 128)
https://github.com/ml-explore/mlx/blob/main/mlx/fast.cpp#L644-L645

**Request**
SDFA to support non-common head dimensions (still multiple of 32)

 - OS Version: MacOS 15.2
 - MLX Version 0.20.0

**Additional context**
Awni's suggestion:  «generalize it so that any even head dim or maybe multiple of 32 is supported»

ParamThakkar123:
## Proposed changes

Please include a description of the problem or feature this PR is addressing. If there is a corresponding issue, include the issue.
Fixes #1604 

## Checklist


awni:
@ParamThakkar123 this PR is just introducing a bug as we don’t have kernels for those sizes.

If you want to work in this issue seriously you should write tests and run some benchmarks. We’re not going to merge a change like this without either.

ParamThakkar123:
Sure @awni 👍

Issue #2089, opened 2025-04-19
Expose `broadcast_shapes`


williambdean:
mlx has np.broadcast_shapes like logic in utils file but doesn't appear to expose it to outward facing API

Reference: 

https://github.com/ml-explore/mlx/blob/00794c42bc734bd817b494fca2b0ccecd9097a58/mlx/utils.cpp?plain=1#L94-L95

ita9naiwa:
Hi, is it okay I take this?

williambdean:
Thanks @ita9naiwa !

PR #2091, opened 2025-04-19
Add broadcast_shapes in python API

ita9naiwa:
## Proposed changes

Add a `broadcast_shapes` function to the Python API. (related #2089)

Since this operation is simple (working only on a list of shape tuples), a pure Python implementation is preferable to reduce C++ binding boilerplate and keep the API surface minimal.

If necessary, I’m open to switching to a C++-exposed version.

## Checklist


awni:
> If necessary, I’m open to switching to a C++-exposed version.

Yes please switch to that. This should be a function in `mlx.core` and there's no need to duplicate the broadcasting logic in Python.

ita9naiwa:
Okay.
I'll make following changes soon.
- place `broadcast_shapes` in `mlx.core`
- use C++ api
thanks for your comment!

ita9naiwa:
fixed! 
@awni could you please review this?

python/src/ops.cpp
```diff
@@ -5189,4 +5189,65 @@ void init_ops(nb::module_& m) {
...
+          if (nb::isinstance<nb::tuple>(shapes[i])) {
+            nb::tuple t = nb::cast<nb::tuple>(shapes[i]);
+            for (size_t j = 0; j < t.size(); ++j) {
+              shape.push_back(nb::cast<int>(t[j]));
+            }
+          } else if (nb::isinstance<nb::list>(shapes[i])) {
+            nb::list l = nb::cast<nb::list>(shapes[i]);
+            for (size_t j = 0; j < l.size(); ++j) {
+              shape.push_back(nb::cast<int>(l[j]));
+            }
```
awni, line 5213: These loops are not necessary. You should be able to get a `mx::Shape` from the `list` or `tuple` in a single line.

python/tests/test_broadcast.py
```diff
@@ -0,0 +1,44 @@
+# Copyright © 2025 Apple Inc.
+
+import mlx.core
+import mlx_tests
+
+
+class TestBroadcast(mlx_tests.MLXTestCase):
+    def test_broadcast_shapes(self):
```
awni, line 8: I don't think we need a separate file for this test since I don't see us adding anything here. Just put it in the ops test for now.

python/src/ops.cpp
ita9naiwa: Thanks!

python/tests/test_broadcast.py
ita9naiwa: Right. I moved `broadcast_shapes` test into `test_ops.py`

python/src/ops.cpp
ita9naiwa: fixed!

ita9naiwa:
fixed! Thanks for quick reivew!

python/src/ops.cpp
```diff
@@ -5189,4 +5189,54 @@ void init_ops(nb::module_& m) {
...
+              "broadcast_shapes expects a sequence of shapes");
+
+        std::vector<mx::Shape> shape_vec;
+        shape_vec.reserve(shapes.size());
+
+        for (size_t i = 0; i < shapes.size(); ++i) {
+          if (!nb::isinstance<mx::Shape>(shapes[i]) &&
+              !nb::isinstance<nb::tuple>(shapes[i]))
+            throw std::invalid_argument(
+                "broadcast_shapes expects a sequence of shapes (tuple or list of ints)");
```
awni, line 5206:
```suggestion
                "[broadcast_shapes] Expects a sequence of shapes (tuple or list of ints).");
```

python/src/ops.cpp
```diff
@@ -5189,4 +5189,54 @@ void init_ops(nb::module_& m) {
...
+
+          shape_vec.push_back(nb::cast<mx::Shape>(shapes[i]));
```
awni, line 5208: What if we just broadcast here instead of making a temporary vector of shapes?

```suggestion
          auto shape = nb::cast<mx::Shape>(shapes[i]);
          result = i == 0 ? shape : mx::broadcast_shapes(result, shape);
```

python/src/ops.cpp
```diff
@@ -5189,4 +5189,54 @@ void init_ops(nb::module_& m) {
...
+        }
+
+        mx::Shape result = shape_vec[0];
+        for (size_t i = 1; i < shape_vec.size(); ++i)
+          result = mx::broadcast_shapes(result, shape_vec[i]);
```
awni, line 5213: Then remove this and put the initializer further up.

awni (approved):
Very nice, thanks!

ita9naiwa:
@awni Sorry I made a last minute change... can you check this??

I think my previous commit wasn't safe (although test passing)

prev one:
```c++
        mx::Shape result;
        for (size_t i = 0; i < shapes.size(); ++i) {
          if (!nb::isinstance<mx::Shape>(shapes[i]) &&
              !nb::isinstance<nb::tuple>(shapes[i]))
            throw std::invalid_argument(
                "[broadcast_shapes] Expects a sequence of shapes (tuple or list of ints).");
          result = mx::broadcast_shapes(result, nb::cast<mx::Shape>(shapes[i]));
        }
```

current:
```c++
mx::Shape result = nb::cast<mx::Shape>(shapes[0]);
        for (size_t i = 1; i < shapes.size(); ++i) {
          if (!nb::isinstance<mx::Shape>(shapes[i]) &&
              !nb::isinstance<nb::tuple>(shapes[i]))
            throw std::invalid_argument(
                "[broadcast_shapes] Expects a sequence of shapes (tuple or list of ints).");
          result = mx::broadcast_shapes(result, nb::cast<mx::Shape>(shapes[i]));
        }
```

awni:
The tests are running. Once they clear we can merge this.

PR #2092, opened 2025-04-19
Added output_padding parameters in conv_transpose

### LINKED ISSUE #1423: [Feature] Add output_padding option to ConvTranspose2d
Passing the same parameters to `nn.ConvTranspose2d` does not revert `nn.Conv2d` in terms of dimension shape conservation, and an `output_padding` option can solve it.

Reference from [PyTorch's docs](https://pytorch.org/docs/stable/generated/torch.nn.ConvTranspose2d.html):

> The padding argument effectively adds dilation * (kernel_size - 1) - padding amount of zero padding to both sizes of the input. This is set so that when a [Conv2d](https://pytorch.org/docs/stable/generated/torch.nn.Conv2d.html#torch.nn.Conv2d) and a [ConvTranspose2d](https://pytorch.org/docs/stable/generated/torch.nn.ConvTranspose2d.html#torch.nn.ConvTranspose2d) are initialized with same parameters, they are inverses of each other in regard to the input and output shapes. However, when stride > 1, [Conv2d](https://pytorch.org/docs/stable/generated/torch.nn.Conv2d.html#torch.nn.Conv2d) maps multiple input shapes to the same output shape. output_padding is provided to resolve this ambiguity by effectively increasing the calculated output shape on one side. Note that output_padding is only used to find output shape, but does not actually add zero-padding to output.

ParamThakkar123:
## Proposed changes
Fixes #1423 

## Checklist

### DIFF
--- a/mlx/ops.cpp
+++ b/mlx/ops.cpp
@@ -3769,6 +3769,7 @@ array conv_transpose_general(
     std::vector<int> stride,
     std::vector<int> padding,
     std::vector<int> dilation,
+    std::vector<int> output_padding,
     int groups,
     StreamOrDevice s) {
   std::vector<int> padding_lo(padding.size());
@@ -3782,7 +3783,8 @@ array conv_transpose_general(
 
     int in_size = 1 + (conv_output_shape - 1);
     int out_size = 1 + stride[i] * (input.shape(1 + i) - 1);
-    padding_hi[i] = in_size - out_size + padding[i];
+    padding_hi[i] = in_size - out_size + padding[i] +
+        output_padding[i]; // Adjust with output_padding
   }
 
   return conv_general(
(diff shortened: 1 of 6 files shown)


ParamThakkar123:
@awni can you please review this PR ?

angeloskath (commented):
This looks great to me! Do you mind adding it to the `nn.layers` and also add tests in `test_conv_transpose.py` as well, we already have a PyTorch dependent test there so it seems fitting.

ParamThakkar123:
Sure @angeloskath . I will make these changes

ParamThakkar123:
@angeloskath I added the output padding parameters to conv_transpose layers in `nn.layers` and also added the tests for them

angeloskath (approved):
I fixed the tests and docstrings, let's merge after the tests pass on CI.

PR #2094, opened 2025-04-19
Complex scan

djphoenix:
## Proposed changes

Add complex variants for:

- log1p (cpu/gpu)
- logaddexp (cpu/gpu)
- logcumsumexp (cpu/gpu)

Also re-enabled (already implemented) complex scan ops for CPU (I am doubt why these were disabled).

TODO: add tests for enabled ops (tested in separate file during development, including gradients).


djphoenix:
I have added tests for log1p and cumsum/cumprod.

Unfortunately, numpy does not support complex logaddexp (and, as result, logcumsumexp as logaddexp.accumulate), so we can't refer to it. @awni @angeloskath  How we can workaround this?

awni:
> Unfortunately, numpy does not support complex logaddexp

You can write it by hand in numpy using `add`, `exp`, `max`, `log` etc and compare against that. I would just check edge cases (like multiple inf, -inf etc) manually and if you aren't sure what the result should be look at PyTorch or Jax as a reference.

djphoenix:
Sure, I already have such implementations in separate file (to compare with). Just in doubt where to place them to use in tests.

awni:
If it's a short snippet of code I would just put them somewhere at the top of the ops test. If it's short it's probably fine to duplicate them. Alternatively you could put it in https://github.com/ml-explore/mlx/blob/main/python/tests/mlx_tests.py

djphoenix:
OK, now I think it seems done (if I have not missed edge cases again, huh).

mlx/backend/metal/kernels/binary_ops.h
```diff
@@ -130,6 +130,25 @@ struct LogAddExp {
         ? maxval
         : (maxval + log1p(metal::exp(minval - maxval)));
   };
+
+  template <>
```
awni, line 134: Can be a plain overload, specialization not needed.
```suggestion
```

awni (approved):
This looks really nice! Thanks for the addition!

PR #2100, opened 2025-04-21
Return const ref in array::data_shared_ptr

zcbenz:
There are 2 cases for calling this method:

1. `buffers.insert(in.data_shared_ptr())`
2. `auto it = buffers.find(arr.data_shared_ptr())`

For case 1 there is no change, for case 2 we can avoid creating a temporary shared_ptr.


awni (approved):
LGTM! Thanks!

PR #2109, opened 2025-04-22
irfft throws instead of segfaults on scalars

### LINKED ISSUE #2105: [BUG] `mx.fft.irfftn` causes core dumped when the input's dimension is 0
**Describe the bug**
I notice that when the input array of `mx.fft.rfftn` or `mx.fft.ifftn` is 0-dimension, they will report the ValueError, showing that `[fftn] Requires array with at least one dimension.`

While `mx.fft.irfftn` will directly causes the core dumped.

**To Reproduce**

Include code snippet
```python
import mlx.core as mx
import numpy as np
arg_1 = np.random.randn()
print(arg_1)
mx_signal = mx.array(arg_1, dtype=mx.float16)
print(mx_signal)
print(mx_signal.shape)
fft_result = mx.fft.irfftn(mx_signal)
print(fft_result)
```

Image

**Expected behavior**
print the ValueError

 - MacOS 15.1.1
 - Version 0.24.1

awni:
As title, closes #2105

### DIFF
--- a/mlx/fft.cpp
+++ b/mlx/fft.cpp
@@ -1,5 +1,4 @@
 // Copyright © 2023 Apple Inc.
-
 #include <numeric>
 #include <set>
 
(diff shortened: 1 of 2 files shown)


angeloskath (approved):
🙏

PR #2121, opened 2025-04-25
Fix the error message in `mx.right_shift` and `mx.left_shift`

Redempt1onzzZZ:
- add check on the input type of `mx.right_shift` and `mx.left_shift`
- these two APIs can throw a more precise message now
image


awni:
Ok I looked at the issue, and I understand your fix now.

awni (approved):
Thanks!

awni:
Closes #2121 2120

Issue #2088, opened 2025-04-19
Use different letter than H in the input docstring


williambdean:
https://github.com/ml-explore/mlx/blob/00794c42bc734bd817b494fca2b0ccecd9097a58/python/src/ops.cpp?plain=1#L3421

awni:
Good point, they should be different.

XueSongTap:
Can I work on this ?

saikiranpennam:
Hello. I would like to work on this issue. I think it will be a greate start for me to know the library more by starting with documentation. I propose `kH` as the new notation and if that doesn’t work for more clearer documentation, happy to take your input on this.

PR #2123, opened 2025-04-25
Clarify dimension notation in conv1d, conv2d, and conv3d docstrings

charan-003:
This PR addresses [issue #2088](https://github.com/ml-explore/mlx/issues/2088) by updating the docstrings for conv1d, conv2d, and conv3d in ops.cpp to use distinct letters for input and kernel dimensions. The previous documentation used the same letter (e.g., H) for both the input spatial dimension and the kernel size, which could be confusing.

Summary of changes:

Updated conv1d docstring to use L for input sequence length and K for kernel size.
Updated conv2d docstring to use H and W for input height and width, and KH, KW for kernel height and width.
Updated conv3d docstring to use D, H, W for input depth, height, and width, and KD, KH, KW for kernel depth, height, and width.
Ensured that the new notation is consistent and unambiguous across all three functions.

### DIFF
--- a/python/src/ops.cpp
+++ b/python/src/ops.cpp
@@ -3455,8 +3455,8 @@ void init_ops(nb::module_& m) {
         1D convolution over an input with several channels
 
         Args:
-            input (array): Input array of shape ``(N, H, C_in)``.
-            weight (array): Weight array of shape ``(C_out, H, C_in)``.
+            input (array): Input array of shape ``(N, L, C_in)``.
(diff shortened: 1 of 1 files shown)


charan-003:
> This is great, thanks. Would you mind making the same change for the transposed convs? Then we can merge it.

@awni Looks good, thanks! Can you confirm if `python/mlx/nn/layers/convolution_transpose.py` is the right file for the transposed convs change?

awni (approved):
Thank !

PR #2129, opened 2025-04-26
Enable vjp for quantized scale and bias

awni:
I'm playing around a bit with fine-tuning the scale and bias parameters to improve the quant quality as a potentially simpler / more general alternative to AWQ.

I figure this addition is useful to have either way.. but if you prefer I can hold off on adding it.


angeloskath (approved):
Love it!

PR #2131, opened 2025-04-28
return library if it is successfully loaded

skyzh:
## Proposed changes

The logic doesn't look right: if the library gets loaded, it will proceed to load it from other directories.

I had to manually copy the metallib to the site-packages/mlx/lib in my own project as it didn't get picked up: https://github.com/skyzh/tiny-llm/blob/c64e9ef52f16c052b4f245f3b08a70d09677f7cc/build_ext.sh#L5

## Checklist


awni:
@angeloskath this looks right to me but I'll let you confirm / approve since you most recently touched that part of the code.

angeloskath (approved):
Thanks for catching that!

Issue #2134, opened 2025-04-28
[Feature Request] Add fftshift functionality


kyrollosyanny:
**Describe the bug**
MLX currently does not have an ```fftshift``` function that can be used before `mlx.fft.fft2` to take the Fourier transform of functions with frequencies at the proper location. It would be useful to add this. Thanks

**Expected behavior**
The final usage should be something like this: ```FT=mx.fft.fftshift(mx.fft.fft2(mx.fft.ifftshift(x)))```

PR #2135, opened 2025-04-29
add fftshift and ifftshift fft helpers

wisefool769:
## Proposed changes

Adds helper function `fftshift`, which moves the zero-frequency component to the center of the array. Also includes its inverse `ifftshift`.  These are found in similar API's like PyTorch, numpy, and MatLab. 

Closes https://github.com/ml-explore/mlx/issues/2134

## Checklist

### DIFF
--- a/mlx/fft.cpp
+++ b/mlx/fft.cpp
@@ -184,8 +184,79 @@ array irfftn(
     StreamOrDevice s /* = {} */) {
   return fft_impl(a, axes, true, true, s);
 }
+
 array irfftn(const array& a, StreamOrDevice s /* = {} */) {
   return fft_impl(a, true, true, s);
+}
+
+array fftshift(
+    const array& a,
+    const std::vector<int>& axes,
+    StreamOrDevice s /* = {} */) {
+  if (axes.empty()) {
+    return a;
+  }
+
+  Shape shifts;
+  for (int ax : axes) {
+    // Convert negative axes to positive
+    int axis = ax < 0 ? ax + a.ndim() : ax;
+    if (axis < 0 || axis >= a.ndim()) {
+      std::ostringstream msg;
+      msg << "[fftshift] Invalid axis " << ax << " for array with " << a.ndim()
+          << " dimensions.";
+      throw std::invalid_argument(msg.str());
+    }
+    // Match PyTorch's implementation
+    shifts.push_back(a.shape(axis) / 2);
+  }
+
+  return roll(a, shifts, axes, s);
+}
+
+array ifftshift(
+    const array& a,
+    const std::vector<int>& axes,
+    StreamOrDevice s /* = {} */) {
+  if (axes.empty()) {
+    return a;
+  }
+
+  Shape shifts;
+  for (int ax : axes) {
+    // Convert negative axes to positive
+    int axis = ax < 0 ? ax + a.ndim() : ax;
+    if (axis < 0 || axis >= a.ndim()) {
+      std::ostringstream msg;
+      msg << "[ifftshift] Invalid axis " << ax << " for array with " << a.ndim()
+          << " dimensions.";
+      throw std::invalid_argument(msg.str());
+    }
+    // Match PyTorch's implementation
+    int size = a.shape(axis);
+    shifts.push_back((size + 1) / 2);
+  }
+
+  return roll(a, shifts, axes, s);
+}
+
+// Default versions that operate on all axes
+array fftshift(const array& a, StreamOrDevice s /* = {} */) {
+  if (a.ndim() < 1) {
+    return a;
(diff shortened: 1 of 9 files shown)


mlx/fft.cpp
```diff
@@ -184,8 +184,79 @@ array irfftn(
...
+  for (int ax : axes) {
+    // Convert negative axes to positive
+    int axis = ax < 0 ? ax + a.ndim() : ax;
+    if (axis < 0 || axis >= a.ndim()) {
+      std::ostringstream msg;
+      msg << "[fftshift] Invalid axis " << ax << " for array with " << a.ndim()
+          << " dimensions.";
+      throw std::invalid_argument(msg.str());
+    }
+    // Match PyTorch's implementation
```
awni, line 210: Does this match NumPy's implementation? If so let's say that instead.

mlx/fft.cpp
```diff
@@ -184,8 +184,79 @@ array irfftn(
...
+  for (int ax : axes) {
+    // Convert negative axes to positive
+    int axis = ax < 0 ? ax + a.ndim() : ax;
+    if (axis < 0 || axis >= a.ndim()) {
+      std::ostringstream msg;
+      msg << "[ifftshift] Invalid axis " << ax << " for array with " << a.ndim()
+          << " dimensions.";
+      throw std::invalid_argument(msg.str());
+    }
+    // Match PyTorch's implementation
```
awni, line 235: Same comment.

python/src/fft.cpp
```diff
@@ -459,4 +459,57 @@ void init_fft(nb::module_& parent_module) {
...
+      "a"_a,
+      "axes"_a = nb::none(),
+      "stream"_a = nb::none(),
+      R"pbdoc(
+        Shift the zero-frequency component to the center of the spectrum.
+
+        Args:
+            a (array): The input array.
+            axes (list(int), optional): Axes over which to perform the shift.
+               If None, shift all axes. Each axis can be negative and will be
```
awni, line 482:
```suggestion
               If ``None``, shift all axes. Each axis can be negative and will be
```

python/src/fft.cpp
```diff
@@ -459,4 +459,57 @@ void init_fft(nb::module_& parent_module) {
...
+               converted to a positive axis using the same rules as NumPy.
```
awni, line 483: I think we can leave out the negative part as almost all APIs in MLX allow negative axes.

```suggestion
               If ``None``, shift all axes.
```

python/src/fft.cpp
```diff
@@ -459,4 +459,57 @@ void init_fft(nb::module_& parent_module) {
...
+        } else {
+          return mx::fft::ifftshift(a, s);
+        }
+      },
+      "a"_a,
+      "axes"_a = nb::none(),
+      "stream"_a = nb::none(),
+      R"pbdoc(
+        The inverse of fftshift. While identical to fftshift for even-length axes,
+        the behavior differs for odd-length axes.
```
awni, line 504:
```suggestion
        The inverse of :func:`fftshift`. While identical to :func:`fftshift` for even-length axes,
        the behavior differs for odd-length axes.
```

python/src/fft.cpp
```diff
@@ -459,4 +459,57 @@ void init_fft(nb::module_& parent_module) {
...
+
+        Args:
+            a (array): The input array.
+            axes (list(int), optional): Axes over which to perform the inverse shift.
+               If None, shift all axes. Each axis can be negative and will be
+               converted to a positive axis using the same rules as NumPy.
```
awni, line 510:
```suggestion
               If ``None``, shift all axes.
```

awni (commented):
This looks quite nice. Left a few minor comments, please address, then we can merge!

awni:
@wisefool769 one of the linux tests failed. Are you able to see the error message? Let us know if you need help debugging.

awni:
> test_fftshift_errors (test_fft.TestFFT) ... /bin/bash: line 1:  3966 Floating point exceptionpython3 -m unittest discover python/tests -v

awni:
Oh nice find! Yes we should definitely fix it. Are you up for sending a PR or adding a fix here?

awni:
> Is it a goal of MLX to smooth over platform-specific differences like this, to achieve reproducibility between Linux and Mac?

I think for genuinely undefined behavior it's fine for the behavior to be platform specific. 

But an np.roll on an empty array should probably not be undefined.

So I guess we can take it on a case-by-case basis. If you find other examples of similar issues feel free to send a PR or file an issue and we can discuss more.

awni (approved):
LGTM!

awni:
Thanks for the contribution!

PR #2138, opened 2025-04-29
Generalize gpu backend

awni:
This is intended to make #2075 fit better with MLX.

So concretely:

- Adds `is_available(Device)` to the top-level. Kept `metal::is_available` for now, but we should prefer to use the new version for general GPU availability checks to make user code more agnostic to Metal / CUDA. The implementation is back-end  specific.
- Moves any metal namespace functions that were used in non-backend specific code to the `gpu` namespace. These are defined in `gpu/eval.h`. And change all calls of these (mostly in scheduler / eval)
- Moved `no_metal` to `no_gpu`

Metal-only specific functions that are in the user interface are now in `metal.h`. These have a `no_metal.cpp` implementation for when the Metal backend is not available. We don't have too many of these (GPU tracing, `device_info`, and the legacy `is_available`). It might make sense in the future to move some of them to the general GPU interface.. depending on what is common between Metal and CUDA.

CC @zcbenz please take a look here. My hope is that rebasing #2075 on top of this will make things a bit simpler and more general.


mlx/backend/metal/device.h
```diff
@@ -266,4 +266,6 @@ class Device {
 
 Device& device(mlx::core::Device);
 
+std::unique_ptr<void, std::function<void(void*)>> new_scoped_memory_pool();
```
awni, line 269: This function is internal only to the metal backend now.

mlx/backend/gpu/available.h
```diff
@@ -0,0 +1,9 @@
+// Copyright © 2025 Apple Inc.
+
+#pragma once
+
+namespace mlx::core::gpu {
+
+bool is_available();
```
zcbenz, line 7: Should it take a device parameter to support multi-device?

zcbenz (approved):

mlx/backend/gpu/available.h
awni: Good question. I think of this one as more of a general "backend availability" API than a "device" availability. But I haven't really thought through what our API for multi-GPU-device on single node looks like. 

I would say let's cross that bridge when we get there... we can always add an overload of this one / update it since it's a pretty internal API.

angeloskath (approved):
Looks great to me!

zcbenz:
Should there be a python `mx.is_available(device)` API?

yiakwy-xpu-ml-framework-team:
Hi @awni is there anyone working on adding ANE backend ?

PR #2139, opened 2025-04-30
Remove metal-only tests

zcbenz:
Move the general gpu tests from `metal_tests.cpp` to `gpu_tests.cpp` so they can be used in cuda backend.


tests/metal_tests.cpp
```diff
@@ -1,12 +1,9 @@
 // Copyright © 2023-2024 Apple Inc.
 
-#include <array>
 #include "doctest/doctest.h"
```
awni, line 3: I think we should simply remove this file.. the test is pretty trivial

awni (approved):
I would just remove the `metal_tests.cpp` file. Would you mind making that change and then we can merge?

zcbenz:
Updated the PR to remove metal-only tests.

PR #2141, opened 2025-04-30
Fix building with uv

zcbenz:
Add missing files needed for building from source distribution, which is used when using uv.

I also added `uv.lock` to gitignore - as long as mlx does not use uv for releasing, it should be ignored for people building with uv.


awni (approved):
LGTM, thanks!

PR #2142, opened 2025-04-30
Allow quant layer to be unfrozen

awni:
Allow parameters in quant linear layer to be unfrozen.

The over-ridden `unfreeze` wasn't really doing much for us anyway since it doesn't get called by parent models. So it only stops quant layers from being unfrozen if you call `unfreeze` directly.

I'm not sure if we want to force the `weight` to stay frozen if you call `unfreeze()` directly without any keys. We could add it back for both the linear layer (and add it for the embedding layer).


angeloskath (approved):
👍

PR #2145, opened 2025-04-30
Move common gpu primitives to backend/gpu

zcbenz:
Many primitives are implemented with the copy op, this PR moves them into `backend/gpu/` so metal and cuda backends share the same code.

Most changes in this PR are just moving code, except a few things:

1. The `concatenate_gpu` function has a few lines of metal-specific code (`start_concurrent()`), which I choose to put under a `#ifdef`.
2. In cuda backend I need to put [nvtx](https://nvidia.github.io/NVTX/) calls in each `eval_gpu` to make them show in profiler, so I added a `MLX_PROFILER_RANGE` macro for this purpose.


mlx/backend/gpu/slicing.cpp
```diff
@@ -36,9 +40,11 @@ void concatenate_gpu(
   flags.row_contiguous = false;
   flags.col_contiguous = false;
   flags.contiguous = false;
+#if defined(MLX_USE_METAL)
   auto& d = metal::device(s.device);
   auto& compute_encoder = d.get_command_encoder(s.index);
   auto concurrent_ctx = compute_encoder.start_concurrent();
+#endif
```
awni, line 47: What do you think about refactoring that like so:

1. Make a `concatenate_gpu_parallel` (or `metal::concatenate_gpu`, ..) in the metal backend
2. Two setup the concurrent context there
3. Call `concatenate_gpu` inside `concatenate_gpu_parallel` after the concurrent context is setup.
4. Call `concatenate_gpu_parallel` everywhere we call `concatenate_gpu` (which is only a couple places).

I think the extra indirection is preferable to the ifdef.. but if you think it's too tedious / not worth it, I'm also fine to revisit later.

awni (approved):
Looks nice!  That will dedup a lot of code between back-ends.

I left one suggestion regarding the ifdef for the concurrent context. Please take a look.

Otherwise good to go!

mlx/backend/gpu/slicing.cpp
zcbenz: I prefer revisiting this later, I don't actually fully understand why `start_concurrent()` is needed here so I'm not confident to make right changes.

angeloskath: Let me try to explain the concurrent context with a simple example.

Let's say we have `f(x, y)` and `g(x, y)` where `x` is the input and `y` the output. If these functions affect the same positions of `y` for instance `f <- add` and `g <- multiply` then they need to synchronize between them which will be done automatically by `set_output_array`.

So you can call the kernels `f` and `g` as follows and it will work

```cpp
enc.set_compute_pipeline_state(f_kernel)
enc.set_input_array(x)
enc.set_output_array(y)
enc.dispatch_threads(...)

enc.set_compute_pipeline_state(g_kernel)
enc.set_input_array(x)
enc.set_output_array(y)
enc.dispatch_threads(...)  // this inserts a barrier because y
                           // is the output of the previous call
```

However, in the case where `f` and `g` write to different parts of `y` (like the case of concatenate) then we can get better utilization of the GPU when we don't synchronize.

```cpp
{
  auto concurent = enc.start_concurrent()

  enc.set_compute_pipeline_state(f_kernel)
  enc.set_input_array(x)
  enc.set_output_array(y)
  enc.dispatch_threads(...)

  enc.set_compute_pipeline_state(g_kernel)
  enc.set_input_array(x)
  enc.set_output_array(y)
  enc.dispatch_threads(...)  // no barrier because of the concurrency context
}

enc.set_compute_pipeline_state(g_kernel)
enc.set_input_array(x)
enc.set_output_array(y)
enc.dispatch_threads(...)  // barrier again because it got destroyed
```

angeloskath: Btw there is no equivalent concept in CUDA (apart from running with different streams which probably shouldn't be handled here).

I think that means that `concatenate_gpu` is fundamentally not a common op so it should probably just have different implementations. Maybe it's not worth it for 3 lines 🤷‍♂️

angeloskath (approved):
This looks great to me. I think it is ok to have `concatenate_gpu` separate for metal, we don't need to have it in the common part.

Generally I prefer to err on the side of verbosity rather than complexity (even though in this case the ifdef can barely be called complexity).

angeloskath:
Hmm not sure what happened with double commenting, sorry for the noise.

mlx/backend/gpu/slicing.cpp
zcbenz: Thanks for the excellent explanation!

I agree this could be likely handled differently in cuda. I have updated the PR to move `concatenate_gpu` implementation out of `backend/gpu`.

angeloskath:
@awni are we waiting on this for some reason? It should be good to go.

PR #2147, opened 2025-05-01
Fix the launcher when ran locally

### LINKED ISSUE #2144: [BUG] Using mlx.launch with ring backend throws at end of program
Simple repro:

```python
for _ in range(100):
    print("hello world")
```

```
mlx.launch -n 2 test.py
```

Gives me something like:

```python
hello worlException in thread Thread-2 (node_thread):
  File "/Users/awnihannun/miniconda3/lib/python3.12/threading.py", line 1073, in _bootstrap_inner
    self.run()
  File "/Users/awnihannun/miniconda3/lib/python3.12/threading.py", line 1010, in run
    self._target(*self._args, **self._kwargs)
  File "/Users/awnihannun/mlx/python/mlx/distributed_run.py", line 292, in node_thread
    outfile.write(msg)
BlockingIOError: [Errno 35] write could not complete without blocking
```

angeloskath:
Fixes #2144 .

Interestingly, the reason for this is that setting STDIN to non blocking makes STDOUT non blocking as well when they both are the TTY. It makes a bit of sense in hindsight but I would never have guessed it if you had asked me.

The solution is simple enough, bite the bullet and do all IO in a non-blocking way.

### DIFF
--- a/python/mlx/distributed_run.py
+++ b/python/mlx/distributed_run.py
@@ -270,9 +270,11 @@ def launch_ring(parser, hosts, args, command):
 
         # Repeat the stdout and stderr to the local machine
         to_read = [p.stdout.fileno(), p.stderr.fileno()]
-        to_write = [p.stdin.fileno()]
+        to_write = [p.stdin.fileno(), sys.stdout.fileno(), sys.stderr.fileno()]
         pidfile = ""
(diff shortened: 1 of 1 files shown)


awni (approved):
Beautiful! Thanks for the quick fix!

PR #2148, opened 2025-05-01
Fix shapeless export to throw on dim mismatch

### LINKED ISSUE #2124: [BUG] Export should throw if dimensions change
```python
import mlx.core as mx

def fun(*args):
    return sum(args)

with mx.exporter("fun.mlxfn", fun, shapeless=True) as exporter:
    exporter(mx.array(1))
    exporter(mx.array(1), mx.array(2))
    exporter(mx.array(1), mx.array(2), mx.array(3))

f2 = mx.import_function("fun.mlxfn")

a = mx.array(10)
b = mx.array([5, 10, 20])

print(fun(a, b))
print(f2(a, b))
```

CC @davidkoski

awni:
For some reason the dim check didn't make it into the export. Now the behavior matches shapeless compile.

Closes #2124

### DIFF
--- a/mlx/export.cpp
+++ b/mlx/export.cpp
@@ -470,6 +470,9 @@ bool FunctionTable::match(
     if (x.dtype() != y.dtype()) {
       return false;
     }
+    if (x.ndim() != y.ndim()) {
+      return false;
+    }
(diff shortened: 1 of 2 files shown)


angeloskath (approved):
🙏

PR #2150, opened 2025-05-02
Fix compile merging

awni:
This fixes the merging of arrays to leave the parents map in a valid state because we rely on that later in compile fusion.

We now get proper fusion of graphs with no-ops.


angeloskath (approved):
👍

PR #2152, opened 2025-05-05
Fix batched vector sdpa

awni:
Mostly the vector sdpa was broken when the batch dim != 1 for any of the inputs. This fixes those cases + adds a few tests.


angeloskath (approved):
Thanks!

PR #2153, opened 2025-05-05
fix input coherent kernel launch

awni:
As title.

Thanks @angeloskath for the repro:

```python
import mlx.core as mx

x = mx.ones((1024 * 1024 + 1))
mx.eval(x)

x = mx.add(x, 2, stream=mx.gpu)
x = mx.subtract(x, 2, stream=mx.cpu)
x = mx.multiply(x, 2, stream=mx.gpu)
mx.eval(x)

print((x == 2).all())
```


angeloskath (approved):
❤️

PR #2158, opened 2025-05-07
CUDA backend: unary ops

zcbenz:

This PR implements unary ops for CUDA backend.

* A `cucomplex_math.cuh` file is added to implement arithmetic operators for cuComplex.
* In `fp16_math.cuh` there are arithmetic functions added for supporting both floats and halfs. For functions that CUDA does not provide a builtin version, halfs are converted to floats.
* A custom iterator class `general_iterator` is added to allow passing strided input to `thrust::transform`. (thrust calls it "fancy iterator")
* A `const_param` utility is added to convert `std::vector` to fixed-sized `cuda::std::array`, which is used to pass shapes and strides as CUDA does not allow passing dynamical-sized arguments to kernels. As result a hard-coded `MAX_NDIM` is introduced, which I set to 8, [in PyTorch they set it to 25](https://github.com/pytorch/pytorch/blob/f320c7b76682720aab420b1ecd1eaa8ad9171be8/aten/src/ATen/cuda/detail/OffsetCalculator.cuh#L15-L31).
* For deciding which types are supported in ops, I use a constexpr `supports_unary_op` function, which can prevent compilation of unsupported combinations of types and ops.


mlx/backend/cuda/kernels/utils.cuh
```diff
@@ -0,0 +1,48 @@
...
+///////////////////////////////////////////////////////////////////////////////
+// CUDA kernel utils
+///////////////////////////////////////////////////////////////////////////////
+
+// To pass shape/strides to kernels via constant memory, their size must be
+// known at compile time.
+#define MAX_NDIM 8
+
+using Shape = cuda::std::array<int32_t, MAX_NDIM>;
+using Strides = cuda::std::array<int64_t, MAX_NDIM>;
```
angeloskath, line 20: Should these be named `DeviceShape` and `DeviceStrides` to avoid confusion? Not sure asking for opinions...

mlx/backend/cuda/kernels/iterators/general_iterator.cuh
```diff
@@ -0,0 +1,121 @@
...
+#include <thrust/iterator/iterator_adaptor.h>
+#include <cuda/std/utility>
+
+namespace mlx::core::cu {
+
+// Iterating non-contiguous array.
+template <typename Iterator, typename IdxT = int64_t>
+class general_iterator
+    : public thrust::
+          iterator_adaptor<general_iterator<Iterator, IdxT>, Iterator> {
```
angeloskath, line 16: This is pretty cool. I am leaving this just as a comment also because I am not sure how `thrust::transform` works under the hood ie does it advance the iterator many times or just once etc.

So this iterator implementation calls elem to loc once at dereference time, if we know we will call dereference many times in a kernel we can do work in advance so that we end up doing less work in total like the `LoopedElemToLoc` ([here](https://github.com/ml-explore/mlx/blob/0cae0bdac83bbf5b3d1da3ca53f1f7eb95981d30/mlx/backend/metal/kernels/utils.h#L205)). To be fair that isn't a particularly beautiful implementation but you get the idea.

No need to change anything now, but maybe when optimizing perf in the future.

mlx/backend/cuda/unary.cu
```diff
@@ -0,0 +1,206 @@
...
+    } else {
+      out.set_data(
+          allocator::malloc(in.data_size() * out.itemsize()),
+          in.data_size(),
+          in.strides(),
+          in.flags());
+    }
+  } else {
+    out.set_data(allocator::malloc(out.nbytes()));
+  }
```
angeloskath, line 128: This should probably be refactored out to `common/unary.h` as `set_unary_op_output_data` 🤔

Actually it is already a function in `cpu/unary.h` so it makes sense to go to `common`.

angeloskath (commented):
This looks super!

I left a few minor comments.

mlx/backend/cuda/kernels/utils.cuh
zcbenz: I'm not sure on this, different name can improve code readability, but in the meanwhile I haven't found inconvenience so far: it is impossible to use `std::vector` in kernel code, and non-kernel code all live outside `namespace cu` so they have to use `cu::Shape`.

mlx/backend/cuda/kernels/iterators/general_iterator.cuh
zcbenz: I think thrust dispatches to different kernels depending on the workload, and it dose loop the iterator in many cases. But it seems that thrust passes calculated offset to iterator, instead of advancing iterator:
https://github.com/NVIDIA/cccl/blob/a0a7b3856e9110722fec1d9ed474648251de033d/cub/cub/agent/agent_for.cuh#L65-L78

We might need to do some deep customization or develop our own version of `transform` when optimizing the code in future.

mlx/backend/cuda/unary.cu
zcbenz: There are already `set_copy_output_data` and `set_binary_op_output_data`, it totally makes to move this to a common header.

awni: That's to avoid donating an input buffer which is a lot bigger than the output. Otherwise you could end up slicing a large input and keeping that buffer alive for the output for example.

We should really use the same condition for the CPU and the GPU here... I think that's just a legacy oversight.

zcbenz: Should `set_copy_output_data` also use the `is_donatable` utility? It is currently calling `in.is_donatable()`.

awni: I think so.. but indeed it's more of an art than a science for that... ideally we'd be running a benchmark suite and measuring memory use impact / run time :). 

I think at the very least having consistent behavior across back-ends (when it makes sense) is a good idea though.

zcbenz:
Updated the PR to move `set_unary_output_data` to a common header.

awni:
Are we good to go on this one? @angeloskath @zcbenz ?

zcbenz:
It should be ready, I think eventually I will rewrite some parts for better performance, but this thrust-based implementation served very well for prototyping.

PR #2161, opened 2025-05-09
Fix fft for integer overflow with large batches

awni:
As title. Added the following test to our longer running CI which passes after this diff.

```python
def fft():
    x = mx.random.normal(shape=(5524996, 800))
    y = mx.fft.rfft(x)
    y_end = np.fft.rfft(np.array(x[-1000:]))
    diff = (y[-1000:] - y_end).abs().max()
    assert diff < 1e-3
```

No noticeable difference in benchmarks I ran.


angeloskath (approved):
🙏

PR #2173, opened 2025-05-11
Add remove_index utility

zcbenz:
It is common code getting a copy of vector with an index removed:

```c++
  auto shape = idx.shape();
  shape.erase(shape.begin() + axis_);
```

which I find hard-to-read because I have to read very carefully to find out which object is actually being operated on.

So I think it is worthwhile to have a `remove_index` utility to improve code readability, with less tying.


awni (approved):
Nice!

Issue #2166, opened 2025-05-10
[BUG] Floating point exception (core dumped) on `mx.hadamard_transform`


Redempt1onzzZZ:
**Describe the bug**
When the input array is empty, it will cause floating point exception on `mx.hadamard_transform` and it repros on both cpu and gpu.

**To Reproduce**

Include code snippet
```python
import mlx.core as mx

a_mx = mx.array([], dtype=mx.float16)

res_gpu = mx.hadamard_transform(a_mx)
print("GPU:", res_gpu)

mx.set_default_device(mx.cpu)
res_cpu = mx.hadamard_transform(a_mx)
print("CPU:", res_cpu)

```

<img width="279" alt="Image" src="https://github.com/user-attachments/assets/f24a6a98-c8f8-4269-b0f7-80ad38b852a0" />

**Expected behavior**
Add check on the input array.

**Desktop (please complete the following information):**
 - OS Version: MacOS 15.1.1
 - MLX: Version 0.24.1

Redempt1onzzZZ:
closed this as completed

PR #2177, opened 2025-05-12
Close a couple edge case bugs: hadamard and addmm on empty inputs

### LINKED ISSUE #2164: [BUG] Floating point exception (core dumped) on `mx.addmm`
**Describe the bug**
I found when three inputs of `mx.addmm` are all empty arrays, the API will cause Floating point exception (core dumped), according to my test, only when c,a,b are all empty or a,b are empty, it will cause core dumped, other situation works normally.

**To Reproduce**

Include code snippet
```python
import mlx.core as mx

c_mx = mx.array([], dtype=mx.float32)
a_mx = mx.array([], dtype=mx.float32)
b_mx = mx.array([], dtype=mx.float32)

res_gpu = mx.addmm(c_mx,a_mx,b_mx)
print("GPU:", res_gpu)

mx.set_default_device(mx.cpu)
res_cpu = mx.addmm(c_mx,a_mx,b_mx)
print("CPU:", res_cpu)

```

Image

 - OS Version: MacOS 15.1.1
 - MLX: Version 0.24.1

awni:
Closes #2164 and #2166

### DIFF
--- a/mlx/backend/cpu/matmul.cpp
+++ b/mlx/backend/cpu/matmul.cpp
@@ -132,6 +132,10 @@ void AddMM::eval_cpu(const std::vector<array>& inputs, array& out) {
     throw std::runtime_error(
         "[AddMM::eval_cpu] Currently only supports float32.");
   }
+  if (out.size() == 0) {
+    out.set_data(allocator::malloc(out.nbytes()));
+    return;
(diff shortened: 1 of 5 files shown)


angeloskath (approved):
🙏

PR #2178, opened 2025-05-13
Fix some complex vjps

angeloskath:
It doesn't fix all of them but it fixes the vjps of multiply, divide and the ffts. At some point we should do another pass for functions like power, scan, reduce etc and add more tests as well.

### DIFF
--- a/mlx/primitives.cpp
+++ b/mlx/primitives.cpp
@@ -1957,47 +1957,46 @@ std::vector<array> FFT::vjp(
   //       isn't part of the graph.
   double n_elements = 1;
   for (auto ax : axes) {
-    n_elements *= primals[0].shape(ax);
+    n_elements *= inverse_ ? cotangents[0].shape(ax) : primals[0].shape(ax);
   }
 
-  if (real_ && inverse_) {
-    auto out = fft::fftn(cotangents[0], axes, stream());
-    auto start = Shape(out.ndim(), 0);
-    auto stop = in.shape();
-    out = slice(out, start, stop, stream());
-    auto mask_shape = out.shape();
-    mask_shape[axes_.back()] -= 2;
-    auto mask = full(mask_shape, 2.0f, stream());
-    auto pad_shape = out.shape();
-    pad_shape[axes_.back()] = 1;
-    auto pad = full(pad_shape, 1.0f, stream());
-    mask = concatenate({pad, mask, pad}, axes_.back(), stream());
-    return {multiply(mask, out, stream())};
-  } else if (real_) {
-    Shape n;
-    for (auto ax : axes_) {
-      n.push_back(in.shape(ax));
-    }
-
+  if (real_) {
     // Make a mask to account for the double use in the forward pass.
-    // Everything except the DC and nyquist frequencies gets halved.
+    // Everything except the DC and nyquist frequencies gets halved or doubled.
+    int N =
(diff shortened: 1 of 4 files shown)


tests/autograd_tests.cpp
```diff
@@ -1149,7 +1149,7 @@ TEST_CASE("test complex gradients") {
     auto cotan = array(complex64_t{2.0, 3.0});
     out = vjp([x](array a) { return multiply(a, x); }, y, cotan).second;
     CHECK_EQ(out.dtype(), float32);
-    CHECK_EQ(out.item<float>(), -8.0);
```
awni, line 1152: I think when I originally wrote this I was using Jax as a reference and it gave `-8.0`. Is that a bug / any ideas about that?

```python
import jax
import jax.numpy as jnp

y = jnp.array(3.0)
x = jnp.array(2.0) + 3.0j
cotan = jnp.array(2.0) + 4.0j

def fun(y):
    return y * x

print(jax.vjp(fun, y)[1](cotan))
```

angeloskath: So Jax computes `dl/dz` and the descent direction is `dl/dz*`.  We can choose to do the same but then we 'd have to make sure that functions like `abs` return `dl/dz` instead of `dl/dz*`. And then the user would have to conjugate any complex gradient before applying it.

To talk specifically about the example above, the difference is that Jax assumes that cotan is `dl/dz` and PyTorch (and now us) assume that the cotton is `dl/dz*`

awni: Gotcha! Thanks for the explanation.

mlx/primitives.cpp
```diff
@@ -1950,30 +1952,62 @@ std::vector<array> FFT::vjp(
...
+    bool odd = cotangents[0].shape(axes_.back()) % 2;
+    Shape c(in.ndim(), 1);
+    c[axes_.back()] = N;
+    array indices = reshape(arange(N, stream()), std::move(c), stream());
+    array first(0, indices.dtype());
+    array last(N - 1 + odd, indices.dtype());
+    array one(1 / n_elements, in.dtype());
+    array two(2 / n_elements, in.dtype());
+    array mask =
+        where((first < indices) & (indices < last), two, one, stream());
```
awni, line 1976: These ops are not on the stream (e.g. `<`, `&`)

mlx/primitives.cpp
```diff
@@ -1950,30 +1952,62 @@ std::vector<array> FFT::vjp(
...
+    bool odd = in.shape(axes_.back()) % 2;
+    Shape c(in.ndim(), 1);
+    c[axes_.back()] = N;
+    array indices = reshape(arange(N, stream()), std::move(c), stream());
+    array first(0, indices.dtype());
+    array last(N - 1 + odd, indices.dtype());
+    array one(1, complex64);
+    array half(0.5, complex64);
+    array mask =
+        where((first < indices) & (indices < last), half, one, stream());
```
awni, line 1996: same

awni (approved):
Nice fixes! Some minor comments otherwise LGTM!

mlx/primitives.cpp
angeloskath: Of course... silly me... I got carried away by how good it looked :-)

PR #2179, opened 2025-05-13
Fix typo in row_reduce_small

zcbenz:
I think this is a typo, tests are passing with and without the change so it is likely untested code path.

### DIFF
--- a/mlx/backend/metal/kernels/reduction/reduce_row.h
+++ b/mlx/backend/metal/kernels/reduction/reduce_row.h
@@ -224,7 +224,7 @@ template <
 
   if ((non_row_reductions < 32 && row_size <= 8) || non_row_reductions <= 8) {
     // Simple loop over non_row_reductions and reduce the row in the thread.
-    IdxT out_idx = tid.x + tsize.y * IdxT(tid.y);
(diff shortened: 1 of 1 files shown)


awni:
Would you mind adding a test which exercises that fix?

zcbenz:
I tried following test and I can verify this PR actually fixes a bug:

```c++
    // Large array with small row size
    int dim0 = pow(2, 28);
    int dim1 = pow(2, 4);
    x = broadcast_to(array({1.0f, 2.0f}), {dim0, dim1, 2});
    auto expected = broadcast_to(array(3.0f), {dim0, dim1});
    CHECK(array_equal(sum(x, 2), expected).item<bool>());
  }
```

However it takes 6s to run, is there a way to construct a faster test?

awni:
That's interesting.. I wonder why it needs such a large size to fail.

zcbenz:
It only fails when the grid is 2d, which needs the input to be larger than UINT_MAX.

awni:
Ohhh got it!! We should put this in our long running tests then. I will add it.

Otherwise good to go here thanks for catching and fixing this!

awni (approved):

PR #2182, opened 2025-05-13
Add random normal distribution for complex numbers

### LINKED ISSUE #2170: [Feature Request] Add random generation for complex64 and view_as_complex function
When I tried `a = mx.random.normal([10, 10], dtype=mx.complex64)`, I got 
```
ValueError: [uniform] Can only generate uniform numbers with real floating point type.
```

which I think originates from the below line

https://github.com/ml-explore/mlx/blob/659a51919fd3d70798e91e9e112075680b95556e/mlx/random.cpp#L127

I think this is not strictly necessary, as we can do something like this 
```python
a = mx.random.normal([10, 10], dtype=mx.float32).astype(mx.complex64)
b = mx.random.normal([10, 10], dtype=mx.float32).astype(mx.complex64) * 1j

c = a + b
```

This can be further simplified and made more efficient if we have something like [`torch.view_as_complex`](https://docs.pytorch.org/docs/stable/generated/torch.view_as_complex.html).

So, I'd suggest to add random generation for complex64 in C++ code and `view_as_complex` function. This is theoretically simple, and I'm happy to make a PR. I'm not very familiar with advanced C++ though.

angeloskath:
As title plus it makes `loc` and `scale` broadcastable arrays to `shape` as discussed in #2170. Closes #2170 .

### DIFF
--- a/mlx/random.cpp
+++ b/mlx/random.cpp
@@ -178,8 +178,8 @@ array uniform(
 
 inline array complex_normal(
     Shape shape,
-    const float loc,
-    const float scale,
+    const std::optional<array>& loc,
+    const std::optional<array>& scale,
     const std::optional<array>& key,
     StreamOrDevice s) {
   auto stream = to_stream(s);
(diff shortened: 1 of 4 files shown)


mlx/random.cpp
```diff
@@ -176,24 +176,50 @@ array uniform(
...
+}
+
 array normal(
     const Shape& shape,
     Dtype dtype,
-    const float loc /* = 0.0 */,
-    const float scale /* = 1.0 */,
-    const std::optional<array>& key /*= nullopt */,
+    const std::optional<array>& loc,
+    const std::optional<array>& scale,
```
awni, line 205: One question on this. If the `dtype` is say `float16` but the `loc` or `scale` are `float32` what should the output type be? I kind of think it should be `float16`.. but I'm not sold on it.

awni (approved):
Nice, thanks! Other than the dtype question this looks great. Depending on what you think there we may want to change the behavior a little.

mlx/random.cpp
angeloskath: Yeah I think you are right. When you explicitly set a `dtype` you expect a `dtype`, it doesn't make much sense otherwise.

PR #2183, opened 2025-05-14
Add set_threadgroup_memory_length to CommandEncoder

Jckwind:
## Summary
- Adds a `set_threadgroup_memory_length` method to the `CommandEncoder` class in Metal backend
- Provides direct access to Metal's setThreadgroupMemoryLength functionality
- Enables better control of threadgroup memory for custom Metal compute kernels

## Test plan
- The method is a simple wrapper around an existing Metal API
- This is a straightforward API addition that doesn't modify existing functionality

### DIFF
--- a/mlx/backend/metal/device.h
+++ b/mlx/backend/metal/device.h
@@ -95,8 +95,8 @@ struct CommandEncoder {
     return enc_->setBytes(&v, sizeof(T), idx);
   }
 
-  void set_threadgroup_memory_length(size_t length, NS::UInteger index) {
-    enc_->setThreadgroupMemoryLength(length, index);
+  void set_threadgroup_memory_length(size_t length, int idx) {
+    enc_->setThreadgroupMemoryLength(length, idx);
   }
 
   ConcurrentContext start_concurrent() {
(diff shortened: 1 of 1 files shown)


mlx/backend/metal/device.h
```diff
@@ -95,6 +95,10 @@ struct CommandEncoder {
     return enc_->setBytes(&v, sizeof(T), idx);
   }
 
+  void set_threadgroup_memory_length(size_t length, NS::UInteger index) {
```
angeloskath, line 98: The index should be an `int idx` like the other functions.

angeloskath (commented):
I am not against adding that but are you actually writing a custom primitive and need it or is it for feature completeness?

Jckwind:
> I am not against adding that but are you actually writing a custom primitive and need it or is it for feature completeness?

Yes! I'm working on a paged attention kernel (it is turning it it will have some aspects of flash attention as well!)

[here is the link to the current repo, to the line where I'm using set_threadgroup_memory_length.](https://github.com/TheProxyCompany/proxy-attention-lab/blob/74e44b9f52dadf33f9baa715a868471015d6f6e1/src/pal_core/src/paged_attention_primitive.cpp#L540)

feel free to poke around the primitive and kernel code!

mlx/backend/metal/device.h
Jckwind: Fixed!

angeloskath (approved):
Looks great! I will merge after the tests pass.

PR #2188, opened 2025-05-15
Non-symmetric eig and eigh

### LINKED ISSUE #2159: Request for support of eigen values and vectors computation for a square matrix
Would it be possible to add a function that computes the eigenvalues and eigenvectors of a square matrix the same way it is implemented in:

- Numpy (https://numpy.org/doc/2.2/reference/generated/numpy.linalg.eig.html)
- Jax (https://docs.jax.dev/en/latest/_autosummary/jax.numpy.linalg.eig.html#jax.numpy.linalg.eig)
- Torch (https://docs.pytorch.org/docs/stable/generated/torch.linalg.eig.html)

This is needed in the Keras MLX backend support. https://github.com/keras-team/keras/issues/19571

Thanks!

awni:
Adds `mx.linalg.eig`, `mx.linalg.eigvals` and the primitive `Eig`.

Closes #2159


angeloskath (approved):
Looks great!

PR #2191, opened 2025-05-16
Add complex eigh

### LINKED ISSUE #2171: [BUG] mlx.core.linalg.eigh not support complex64 matrices
**Describe the bug**
I tried to decompose a complex hermitian matrix, but got `ValueError: [linalg::eigh] Arrays must have type float32 or float64. Received array with type complex64`.

As the documentation mentioned that 

https://github.com/ml-explore/mlx/blob/659a51919fd3d70798e91e9e112075680b95556e/python/src/linalg.cpp#L456

I'd rather consider this as a bug instead of a missing feature

**To Reproduce**

Include code snippet
```python
def rand_hermitian_matrix(dim: int) -> mx.array:
    """
    Generate a random Hermitian matrix of given dimension.

    Args:
        dim (int): Dimension of the matrix.
    Returns:
        mx.array: A Hermitian matrix of the specified dimension.
    """

    # temp walkaround when mlx cannot rand a complex number
    H = mx.random.normal([dim, dim], dtype=mx.float32).astype(mx.complex64)
    H += mx.random.normal([dim, dim], dtype=mx.float32).astype(mx.complex64) * 1j
    H = H + mx.conj(H.transpose())
    return H

H = rand_hermitian_matrix(3)
print(f"Random Hermitian Matrix:\n{H}")

# To decompose a *Hermitian* matrix, use the `eigh` function
eigenvalues, u = mx.linalg.eigh(H, stream=mx.cpu)
```

**Expected behavior**
This code runs without errors

 - OS Version: MacOS 15.5
 - Version: 0.25.2

angeloskath:
Implements `eigh` for complex numbers. Closes #2171 .

The one notable change is the move constructor in `array::Data`. I can't think of a reason not to have it but let me know.

I am not too happy with the templatization but the number of arguments changes from `syevd` to `heevd` so not sure what I could have done better.

### DIFF
--- a/mlx/array.h
+++ b/mlx/array.h
@@ -224,6 +224,10 @@ class array {
     // Not copyable
     Data(const Data& d) = delete;
     Data& operator=(const Data& d) = delete;
+    Data(Data&& o) : buffer(o.buffer), d(o.d) {
+      o.buffer = allocator::Buffer(nullptr);
+      o.d = [](allocator::Buffer) {};
+    }
     ~Data() {
       d(buffer);
     }
--- a/mlx/backend/cpu/eigh.cpp
+++ b/mlx/backend/cpu/eigh.cpp
@@ -12,6 +12,133 @@ namespace mlx::core {
 
 namespace {
 
+template <typename T, class Enable = void>
+struct EighWork {};
+
+template <typename T>
+struct EighWork<
+    T,
+    typename std::enable_if<std::is_floating_point<T>::value>::type> {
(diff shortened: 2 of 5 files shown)


awni (approved):
Looks great! I'm not sure why the linux test is failing.. 

The lapack complex config documentation is pretty sparse https://www.netlib.org/lapack/lapacke_config.h. maybe we need to `#define LAPACK_COMPLEX_CPP` 🤔

PR #2192, opened 2025-05-16
Include cuda_bf16.h for bfloat16 overloads

zcbenz:
The operators of `__nv_bfloat16` are defined `cuda_bf16.h` and there is no need to define our own. Refs #2075.


awni (approved):
Thanks!!

PR #2193, opened 2025-05-16
reduce vjp for all and any

awni:
As title. Send zeros through for convenience / consistency with other ops.. though one should usually use stop gradient.


angeloskath (approved):
👍

PR #2202, opened 2025-05-19
Nearest

### LINKED ISSUE #2186: [BUG] `mlx.nn.Upsample` produces different result from `torch.nn.functional.interpolate` with `nearest` mode
**Describe the bug**
When using `mlx.nn.Upsample` to downsample an array with shape BxCxHxW using `nearest` interpolate mode and float scale factor, the result is different from `torch.nn.functional.interpolate`.

**To Reproduce**

Include code snippet
```python
import torch
import numpy as np
import torch.nn.functional as F
import mlx.core as mx
import mlx.nn as nn

# Define input tensor
h, w = 30, 40
scale_factor = (2.5, 2.5)
out_h, out_w = (int(h * scale_factor[0]), int(w * scale_factor[1]))
input_tensor = torch.randn((1, 1, h, w), dtype=torch.float32).to("mps")  # BxCxHxW


# Using torch.functional.interpolate
resized_tensor = F.interpolate(input_tensor, size=(out_h, out_w), mode="nearest")

# Using mlx.nn.Upsample
input_array = mx.array(
    input_tensor.permute(0, 2, 3, 1).cpu().numpy()
)  # BxCxHxW -> BxHxWxC for mlx
print("input_array", input_array.shape)
upsample = nn.Upsample(scale_factor, mode="nearest", align_corners=False)
resized_mlx = upsample(input_array)

# Compare the results
resized_tensor = resized_tensor.permute(0, 2, 3, 1)
print("resized_tensor:")
print(resized_tensor.shape)
print("resized_mlx:")
print(resized_mlx.shape)

# Check if they are the same

diff = torch.sum(
    torch.square(resized_tensor.cpu() - torch.Tensor(np.array(resized_mlx)))
)
print("torch and mlx diff", diff)

assert torch.allclose(
    resized_tensor.cpu(), torch.Tensor(np.array(resized_mlx)), atol=1e-6
), "torch and mlx are not the same!"

```

**Expected behavior**
Expect no error. But got
```
AssertionError: torch and mlx are not the same!
```

 - OS Version: [MacOS 15.2]

angeloskath:
Closes #2186.

There isn't only one way to map pixels to the nearest pixel for instance PT has `nearest-exact` and `nearest`. I am not feeling strongly either way but I guess since it matches PT now we should merge it. If this keeps cropping up it might make sense to match nearest exact which might have an algorithm description rather than going by matching certain outputs.

### DIFF
--- a/python/tests/test_upsample.py
+++ b/python/tests/test_upsample.py
@@ -51,6 +51,7 @@ class TestUpsample(mlx_tests.MLXTestCase):
                     align_corners=align_corner,
                 )(in_mx)
                 mode_pt = {
+                    "nearest": "nearest",
                     "linear": "bilinear",
                     "cubic": "bicubic",
                 }[mode]
@@ -58,7 +59,7 @@ class TestUpsample(mlx_tests.MLXTestCase):
                     in_pt,
                     scale_factor=scale_factor,
                     mode=mode_pt,
-                    align_corners=align_corner,
+                    align_corners=align_corner if mode != "nearest" else None,
                 )
                 out_pt = torch.permute(out_pt, (0, 2, 3, 1)).numpy(force=True)
                 self.assertEqual(out_pt.shape, out_mx.shape)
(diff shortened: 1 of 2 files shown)


python/tests/test_upsample.py
```diff
@@ -76,14 +77,17 @@ def run_upsample(
...
                     ((10, 10), (0.2, 0.2)),
+                    ((10, 10), (0.3, 0.3)),
                     ((11, 21), (3.0, 3.0)),
                     ((11, 21), (3.0, 2.0)),
                 ):
                     # only test linear and cubic interpolation
                     # there will be numerical difference in nearest
                     # due to different indices selection.
-                    for mode in ("cubic", "linear"):
+                    for mode in ("cubic", "linear", "nearest"):
```
awni, line 87: Nit: I guess we should remove that comment?

awni (approved):
Looks great, thanks!

PR #2206, opened 2025-05-19
Fix large arg reduce

awni:
Added the following to long running tests:

```python
def arg_reduce():
    x = mx.random.uniform(shape=(151936 * 64, 16, 16))
    out = mx.argmin(x, axis=-1)

    for idx in [0, 10000, -10000, -1]:
        expected = mx.argmin(x[idx], axis=-1)
        assert mx.array_equal(out[idx], expected)
```


angeloskath (approved):
🙏

PR #2209, opened 2025-05-21
Build for compute capability 70 instead of 75 in CUDA backend

zcbenz:
V100 is compute capability 7.0, it is the oldest architecture that supports synchronization between CPU/GPU with unified memory.


awni (approved):
Nice, thanks!

PR #2213, opened 2025-05-21
Fix out-of-bounds default value in logsumexp/softmax

zcbenz:
The `_looped` version is slightly different from the simple version that fails to handle the `-inf` inputs.


awni (approved):
Nice, thanks!

PR #2215, opened 2025-05-21
check relative path of metallib file from dylib instead of binary

DavitGrigoryan132:
## Proposed changes

When using MLX as a dynamically-linked library, we only ever looked for
default.metallib next to the *executable*. If the metallib wasn’t found
there (or in the SwiftPM bundle), we’d error out—even if it sat beside
the .dylib itself.

This change adds an extra lookup step: if no metallib is found in the
binary’s directory, we now use dladdr() to locate the MLX dylib on disk
and attempt to load metallib relative to that folder. 

## Checklist

### DIFF
--- a/mlx/backend/metal/device.cpp
+++ b/mlx/backend/metal/device.cpp
@@ -3,6 +3,7 @@
 #include <cstdlib>
 #include <filesystem>
 #include <sstream>
+#include <dlfcn.h>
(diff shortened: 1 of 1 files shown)


awni:
Closing as I believe we already do this on main. Feel free to comment if I missed something.

PR #2216, opened 2025-05-22
Use fallbacks in fast primitives when eval_gpu is not implemented

zcbenz:
The fast primitives have fallback implementations that are used for CPU and vjp/vmap, they can also be used as placeholders for GPU backends before having a fast GPU implementation.

This PR adds a `use_fallback()` method to the fast primitives to use the fallbacks in absence of a GPU implementation.

I also cleaned up some unused properties in fast primitives.

### DIFF
--- a/mlx/fast.cpp
+++ b/mlx/fast.cpp
@@ -9,7 +9,6 @@
 #include "mlx/fast_primitives.h"
 #include "mlx/ops.h"
 #include "mlx/transforms.h"
-#include "mlx/transforms_impl.h"
 
 namespace mlx::core::fast {
 
@@ -112,7 +111,8 @@ array rms_norm(
 
   auto passed_weight =
       (has_weight) ? astype(*weight, out_type, s) : array(1, out_type);
-  if (s.device == Device::gpu) {
+
+  if (!RMSNorm::use_fallback(s)) {
     return array(
         x.shape(),
         out_type,
@@ -256,7 +256,7 @@ array layer_norm(
   auto passed_bias =
       (has_bias) ? astype(*bias, out_type, s) : array(0, out_type);
 
-  if (s.device == Device::gpu) {
+  if (!LayerNorm::use_fallback(s)) {
     return array(
         x.shape(),
         out_type,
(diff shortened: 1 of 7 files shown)


angeloskath:
That is a great start! I think this should also take over the stream check and shape checks but if you want as a first step to unblock you I think it is fine to merge. Basically `use_fallback` should have something like the following signature imho

```cpp
static bool Primitive::use_fallback(
    const std::vector<array>& outputs,
    const Stream& s);
```

zcbenz:
I think it is good idea adding parameters to `use_fallback`, though different primitives need to have different parameters.

For example `scaled_dot_product_attention` needs the mask parameters to determine whether to use fallback:
https://github.com/ml-explore/mlx/blob/55b4062dd8c71d4499a430012b49f676da91818a/mlx/fast.cpp#L742-L743

I have updated `use_fallback` to take parameters, and moved some code to `ScaledDotProductAttention::use_fallback`.

mlx/fast.cpp
```diff
@@ -770,7 +746,9 @@ array scaled_dot_product_attention(
     mask_shape.back() = keys.shape(-2);
     inputs.push_back(broadcast_to(mask_arr, mask_shape, stream));
   }
-  if (!detail::in_grad_tracing() && implementation_supports_use_case) {
+  if (!detail::in_grad_tracing() &&
```
angeloskath, line 749: I think `!detail::in_grad_tracing()` should go in `use_fallback` as well. It is there because we don't have a a custom primitive for `ScaledDotProductAttentionVJP` and it is faster to use the fallback if we are computing gradients.

mlx/fast_primitives.h
```diff
@@ -65,7 +67,6 @@ class RMSNorm : public Custom {
   }
 
  private:
-  std::function<std::vector<array>(std::vector<array>)> fallback_;
```
angeloskath, line 68: Thanks for that.

angeloskath (approved):
I think it looks great. I left a minor comment.

mlx/fast.cpp
zcbenz: That makes sense, I have updated the PR.

PR #2217, opened 2025-05-22
Remove unused defines

zcbenz:
Remove the `mlx_EXPORTS` define which is added by cmake automatically when building shared library. Also remove the unused `MLX_USE_CUDA` define.

With this change ccache can share more build cache when switching branches.


awni (approved):
LGTM thanks!

PR #2221, opened 2025-05-24
Add memory cache to CUDA backend

zcbenz:
This PR moves `BufferCache` out of Metal backend and reuses it in CUDA backend.

There are a few changes to `BufferCache`:
* It is decoupled from Metal so page size, buffer type and buffer methods are now passed via constructor.
* The `remove_from_list` method is now responsible for freeing the linked list node, the node was leaked in `release_cached_buffers`.
* The `release_cached_buffers` method did not remove the released node in `buffer_pool_`, I added cleaning code.

### DIFF
--- a/mlx/backend/common/buffer_cache.h
+++ b/mlx/backend/common/buffer_cache.h
@@ -1,0 +1,163 @@
+// Copyright © 2025 Apple Inc.
+
+#pragma once
+
+#include <functional>
+#include <map>
+
+namespace mlx::core {
+
+template <typename T>
+class BufferCache {
+ public:
+  BufferCache(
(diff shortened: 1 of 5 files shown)


zcbenz:
I realized that much code complexity came from the fact that buffer could be null, which only happened because `release_cached_buffers()` set the buffer to null and left the node in `buffer_pool_`. I pushed another commit to simplify the `BufferCache` code, especially the `reuse_from_cache` and `release_cached_buffers` code.

This change needs some cautions, but I think the code becomes easier to understand and verify.

awni (approved):
Looks great to me!

PR #2230, opened 2025-05-31
Do not check event.is_signaled() in eval_impl

zcbenz:
Checking `event.is_signaled()` in CUDA backend can be expensive because it is reading a value in CPU which is usually updated by GPU, and it would involve host-device memory copy and possible locks.

It also seems to have positive effects in Metal backend, I ran `./build/examples/cpp/logistic_regression` for a few times and the average it/s seemed to improve from 2140 to 2160.


awni (approved):
Looks good!

PR #2232, opened 2025-05-31
Avoid invoking allocator::malloc when creating CUDA event

zcbenz:
Use raw `cudaMallocManaged` API instead of `allocator::malloc` to allocate memory in CUDA backend's `SharedEvent`, also slightly refactor `CudaAllocator::cuda_free` so it takes a raw pointer instead of `CudaBuffer` wrapper.

This change is needed because `allocator::malloc` could create `SharedEvent` (caused by `buffer_cache_.release_cached_buffers` → `CudaAllocator::free` → `new Worker`) which would then call `allocator::malloc` again, there is no recursion call but the mutex would be locked twice and results in dead lock. We can work around it by using a recursive mutex, but I think it is better breaking the circle.

The downside is `SharedEvent` no longer benefits from buffer cache, but we should have a separate event cache instead if event creation is costly.

### DIFF
--- a/mlx/backend/cuda/allocator.cpp
+++ b/mlx/backend/cuda/allocator.cpp
@@ -18,7 +18,10 @@ CudaAllocator::CudaAllocator()
     : buffer_cache_(
           getpagesize(),
           [](CudaBuffer* buf) { return buf->size; },
-          [this](CudaBuffer* buf) { cuda_free(buf); }) {
+          [this](CudaBuffer* buf) {
+            cuda_free(buf->data);
+            delete buf;
+          }) {
   // TODO: Set memory limit for multi-device.
   size_t free, total;
   CHECK_CUDA_ERROR(cudaMemGetInfo(&free, &total));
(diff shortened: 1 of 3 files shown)


awni:
I didn't quite follow the design actually. Can you explain why `cuda_free` might create a new worker?

zcbenz:
`CudaAllocator::cuda_free` needs to schedule the `cudaFree` call to worker if it is called in CPU stream thread, otherwise a deadlock would happen when GPU stream waits on a CPU stream with `event_wait_kernel`, while CPU stream calls `cudaFree` which waits on the `event_wait_kernel`.

awni:
@zcbenz it makes sense to me that the `cuda_free` needs to put the call in the worker queue. I'm just wondering... should the workers not already exist at that point? Or when are they first constrcuted? For a given CPU stream is it the first time `cuda_free` is called?

zcbenz:
The `Worker` is lazily created when it is used, i.e. the first time a CPU stream calls `cuda_free`.

The deadlock happened that: a CPU stream calls `malloc` → buffer cache triggered to do some cleanup → `cuda_free` gets called in the same CPU stream thread → `Worker` gets created as `cuda_free` is called for the first time in CPU stream → `SharedEvent` gets created which calls `malloc` again.

awni (approved):
LGTM!

PR #2239, opened 2025-06-03
Default strict mode for module `update` and `update_modules`

### LINKED ISSUE #2233: [BUG] Improvements to `Module.update` and `Module.update_modules`
Some of the following should probably not work.. and maybe the last one should.

```python
import mlx.nn as nn

m = nn.Sequential(nn.Linear(3, 3), nn.Linear(3, 3))
print(m.leaf_modules())

# Updating non-existent parameters should not be allowed by default
updates = {'layers': [{'value': 0}]}
m.update(updates)

# Updating non-existent modules should not be allowed by default
m = m.update_modules({"values": [0, 1]})

# This works.. should it?
m = m.update_modules({"layers": [0, 1, 2]})

# This doesn't work.. should it?
m = m.update_modules({"layers": [nn.ReLU()]})
```

awni:
`Module.update` and `Module.update_modules` take a `strict` kwarg which defaults to `True`. This makes it harder to accidentally not update parameters or modules.

Closes #2233

### DIFF
--- a/python/mlx/nn/layers/base.py
+++ b/python/mlx/nn/layers/base.py
@@ -193,7 +193,7 @@ class Module(dict):
                     )
 
         if len(weights) != 0:
-            self.update(tree_unflatten(weights))
+            self.update(tree_unflatten(weights), strict=False)
         return self
 
     def save_weights(self, file: str):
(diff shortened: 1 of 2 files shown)


angeloskath (approved):
That's great! Thank you for adding that

PR #2240, opened 2025-06-03
Share more common code in Compiled

zcbenz:
Move some code in `meta/compiled.cpp` to 2 utility functions in `common/compiled.h`: `compiled_collapse_contiguous_dims` and `compiled_use_large_index`.

Add a `is_constant_` function object in `Compiled` to replace the direct uses of `inputs_` and `constant_ids_`, this is to improve code readability so I don't have to distinguish between `inputs` and `inputs_`.

Remove some unused code.

### DIFF
--- a/mlx/compile.cpp
+++ b/mlx/compile.cpp
@@ -5,6 +5,7 @@
 #include <unordered_set>
 
 #include "mlx/allocator.h"
+#include "mlx/backend/common/compiled.h"
 #include "mlx/compile.h"
 #include "mlx/compile_impl.h"
 #include "mlx/fast_primitives.h"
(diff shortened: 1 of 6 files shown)


awni (approved):
Looks good!

mlx/compile.cpp
```diff
@@ -82,7 +83,11 @@ Compiled::Compiled(
       inputs_(std::move(inputs)),
       outputs_(std::move(outputs)),
       tape_(std::move(tape)),
-      constant_ids_(std::move(constant_ids)) {}
+      constant_ids_(std::move(constant_ids)),
+      is_constant_([this](size_t i) {
+        return constant_ids_.find(inputs_[i].id()) != constant_ids_.end();
+      }),
+      kernel_lib_(build_lib_name(inputs_, outputs_, tape_, constant_ids_)) {}
```
awni: This is kind of a nit.. but maybe it makes sense to just move that function into the constructor?  Then we wouldn't need to include the backend header. And it's nice to have the code that this file uses centralized.

awni: It's only used here now.. so it doesn't really need to be a free function in the backend anymore..

zcbenz: Yeah it makes sense moving the code, I have updated the PR. The backend header is still needed though since utils like `print_constant` are used.

PR #2241, opened 2025-06-03
CUDA backend: matmul

zcbenz:

Implement `Matmul` and `AddMM` with cublasLt.

There are a few limitations in this approach:

* cublasLt requires the last 2 dimensions of matrix to be contiguous, so copy happens a lot.
* cublasLt can only handle batches with at most one stride, I'm using a loop for batches with two more strides.


mlx/backend/common/matmul.h
```diff
@@ -0,0 +1,78 @@
+// Copyright © 2025 Apple Inc.
+
+#pragma once
```
awni, line 3: Maybe we should move this to `gpu/matmul.h` and include the `check_transpose` function there?

mlx/backend/cuda/matmul.cpp
```diff
@@ -0,0 +1,474 @@
...
+    return std::make_tuple(false, stx, arr);
+  } else if (stx == 1 && sty == arr.shape(-2)) {
+    return std::make_tuple(true, sty, arr);
+  } else {
+    array arr_copy(arr.shape(), arr.dtype(), nullptr, {});
+    copy_gpu(arr, arr_copy, CopyType::General, s);
+    copies.push_back(arr_copy);
+    return std::make_tuple(false, arr.shape(-1), arr_copy);
+  }
+}
```
awni, line 294: Could move this to `gpu/matmul.h`?

awni: I see it's slightly different than the one we have in Metal actually..

mlx/backend/cuda/matmul.cpp
```diff
@@ -0,0 +1,474 @@
...
+  if (batch_count > 1 && !a_transposed && batch_shape.size() == 1 &&
+      a.strides()[a.ndim() - 2] == K && a_batch_strides.back() == M * K &&
+      b_batch_strides.back() == 0) {
+    M *= batch_shape.back();
+    batch_count = 1;
+
+    a_batch_strides = {0};
+    b_batch_strides = {0};
+    batch_shape = {1};
+  }
```
awni, line 350: I'm a little confused by this. Is it safe because B is broadcast along the batch dimension?

awni:
I think we can merge this without it.. but I'm wondering about the loop over batches. Is that a limitation of the cuBLAS light library? Or is there no good solution for it in general?

mlx/backend/common/matmul.h
zcbenz: Can cpu backend benefit from `collapse_batches`? It is not used there but I think it might help someday.

mlx/backend/cuda/matmul.cpp
zcbenz: The one in metal backend does not copy for `gemv`. This one is same with the one in cpu backend, they should be moved to a common header but they need to get a better name.

mlx/backend/cuda/matmul.cpp
zcbenz: This was copied from metal backend.

zcbenz:
> but I'm wondering about the loop over batches. Is that a limitation of the cuBLAS light library? Or is there no good solution for it in general?

MatX does the same thing:
https://github.com/NVIDIA/MatX/blob/7103574680291c8a03df998a2c2db48c8f138995/include/matx/transforms/matmul/matmul_cuda.h#L859-L880

Considering it is from NVIDIA, I think there is likely no better way with cublasLt.

awni:
The test failure here is unrelated to the diff.

PR #2242, opened 2025-06-03
Fix unintuitive metal kernel caching

awni:
Previously if you build a kernel with `mx.fast.metal_kernel` and then do it again with a change in source code.. you would silently get the wrong result. This adds a time based key to the library name so we make a new library every time you make a new kernel.

Open to other suggestions for how to do this... this seems like a good compromise since one should usually build the kernel once and then call it many times.

Also:

- Updated the docs to reflect the above + some nits
- Added a check on max threads per thread group and throw if it's exceeded

Closes #2224


angeloskath:
So this looks great and my only concern has to do with the metal cache. Basically this way we will be recompiling every single time and the kernel will never be reused since there is a timestamp in the cache. That is not bad (it is the desired behavior) but the lib will still be cached won't it?

angeloskath (approved):

awni:
> Basically this way we will be recompiling every single time and the kernel will never be reused since there is a timestamp in the cache. That is not bad (it is the desired behavior) but the lib will still be cached won't it?

Indeed it will be. My first thought was to punt on it because otherwise we'd need to add a way to remove a library from the cache from the front-end which is a little messy. But maybe we should fix that here 🤔 .

awni:
Ok, I made it work so that now it removes the library / kernels from the cache if the source code changes. It ended up being a bigger change but I think hygienically it's a good change regardless of the problem this is trying to solve.

Basically before kernel names had to be globally unique which could cause collisions especially since users can define their own kernel names. Now kernel names are specific to a library, so they only have to be unique in the library.

The reason I needed that for this is because we have to remove the kernels from the kernel cache for the custom library in addition to removing the custom library.

Let me know what you think.

mlx/backend/metal/custom_kernel.cpp
```diff
@@ -39,9 +354,24 @@ void CustomKernel::eval_gpu(
...
 
   auto& d = metal::device(s.device);
-  const auto& lib_name = name_;
-  auto lib =
-      d.get_library(lib_name, [this] { return metal::utils() + source_; });
+
+  {
+    // Clear kernels from the device library cache if needed
+    auto& kernel_cache = cache();
+    std::unique_lock lock(kernel_cache.mtx);
```
angeloskath: It is not clear that we need shared access but given that the `kernel_cache.mtx` is a `shared_mutex` did you mean for this to be

```
bool emplace = false;
bool edit = false;
{
  std::shared_lock lock(kernel_cache.mtx);
  // check and set the flags
}
if (emplace || edit) {
  std::unique_lock lock(kernel_cache.mtx);
  // do the things (need to recheck of course)
}
```

angeloskath (commented):
This looks great! I much prefer this to filling the system cache.

Maybe I missed it or misunderstand so let me try to walk it step by step.

We have a script that registers and runs 2 custom metal kernels under the same name one after the other, we run this script twice.

1. Create the runtime cache entry and the system cache entry when compiled
2. Clear the runtime cache entry and recompile a new kernel because the source changed (the system cache is updated accordingly)
3. The system cache recompiles and updates the entry because the source changed. The runtime cache adds the entry.
4. Clear the runtime cache entry and recompile a new kernel and so on.

Basically, I am trying to confirm that the system will also recompile because the source changed.

awni:
Yea you're outline of the steps sounds right if I'm understanding the terms. By "system cache" I think you mean the OS managed JIT binary cache.

> Basically, I am trying to confirm that the system will also recompile because the source changed.

**System cache**: Indeed that part works. The system always compiles if it gets an unseen source string (doesn't matter what the library or kernel names are). There is no change to that here and it's not managed by MLX at all.

**MLX device library/kernel cache**: In MLX we cache the loaded libraries and loaded kernels on the device. That part is changed so that each library is now associated to a unique kernels cache (so kernel names are scoped to their library). It also let's us remove the kernels for a given library (which is important for the next step).

**Custom kernel cache**: This diff also adds a higher level cache for the custom metal kernels. This caches the source based on the kernel name. And if there is an entry in the cache with different source then it clears the relevant entries from the lower level **MLX device/library kernel cache**.

mlx/backend/metal/custom_kernel.cpp
awni: No indeed, it should just be a regular mutex.. copy paste gone awry.

awni: I mean I guess we could use the shared style but I think it's overkill for this.

angeloskath: Agreed the string equality would probably be the bulk of the cost anyway 🤷‍♂️

PR #2243, opened 2025-06-04
Perf regression fix

angeloskath:
This fixes the performance regression introduced from #2230 by reverting it.

The commit just before has another "fix" which I don't think works because an array may be used in several independent streams or evals before the computation finishes in which case the later ones wouldn't sync with it.

I think a possible better fix that we can look into introducing in a later commit (to get this out quickly to fix the bug) is to have the event itself do the `is_signaled` check when it is cheap to do so or schedule a wait if it is better.

@zcbenz


awni (approved):
Nice!

PR #2247, opened 2025-06-06
Improve metal elementwise kernels

awni:
I ended up going with the two kernel option + dispatch based on size.

In benchmarks pretty much everything is faster:

This is on an M4 max but improvements are similar on M2 Ultra

N | Pre | Post
---| --- | ---
1024    |  0.08511 | 0.0659
2048    |  0.06478 | 0.0555
4096    |  0.06928 | 0.0559
8192    |  0.06935 | 0.0574
16384   |  0.06982 | 0.0570
32768   |  0.07120 | 0.0673
65536   |  0.07111 | 0.0663
131072  |  0.07207 | 0.0687
262144  |  0.08111 | 0.0763
524288  |  0.11154 | 0.1058
1048576 |  0.17414 | 0.1689
2097152 |  0.35799 | 0.3572
4194304 |  0.62967 | 0.6183
8388608 |  1.17261 | 1.1109

### DIFF
--- a/mlx/backend/metal/jit_kernels.cpp
+++ b/mlx/backend/metal/jit_kernels.cpp
@@ -42,8 +42,10 @@ MTL::ComputePipelineState* get_unary_kernel(
     concatenate(kernel_source, metal::unary_ops(), metal::unary());
     kernel_source +=
         get_template_definition("v_" + lib_name, "unary_v", in_t, out_t, op, 1);
-    kernel_source +=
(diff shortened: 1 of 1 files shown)


angeloskath (approved):
🚀

awni:
Our metal library went from 88 -> 91 MB with this change. Not a big bump. But it's bigger in general than I thought it was.. we may need to do a round of optimization on it in the near future.

PR #2248, opened 2025-06-06
Fix linux linking error

awni:
As title.. not sure why this stopped working (or ever did work before). I think $ORIGIN  is the right token to use for linux machines.


angeloskath (approved):
🙏

PR #2259, opened 2025-06-10
CUDA backend: binary ops

zcbenz:

Implements binary ops in CUDA backend, the design is almost identical to the Metal kernels, with a few changes:

* Each thread currently only handles one element, there are some simple optimizations possible but I'll implement them in future PRs with analysis from profiler.

* The Metal kernels rely on a Metal feature that you can specify the exact number of threads in a grid with the [`dispatchThreads` API](https://developer.apple.com/documentation/metal/mtlcomputecommandencoder/dispatchthreads(_:threadsperthreadgroup:)?language=objc) and therefore use grid size as tensor dimensions. There is no such convenient API in so the CUDA kernels for strided data are simplified into 2 versions:
  * `binary_g_nd` which optimizes the case when `ndim` is known at compile time.
  * `binary_g` which handles dynamic `ndim` >= 4.

* New versions of `elem_to_loc` utilities:
  * `elem_to_loc_nd` which optimizes for compile time `ndim`.
  * `elem_to_loc_4d` which optimizes for dynamic `ndim` >= 4.

* New utilities added to `kernel_utils.cuh`:
  * `MLX_SWITCH_BOOL` to turn a bool variable into constexpr.
  * `MLX_SWITCH_1_2_3` to turn an integer between 1~3 to constexpr, this is used for dispatching to `binary_g_nd` kernel with `ndim` <= 3.
  * `get_launch_args` which returns the grid and block dims that maximize occupancy for a kernel and input array.


awni (approved):
This looks really nice! And very self-contained. So I'm good to merge it.

PR #2260, opened 2025-06-10
CUDA backend: copy ops

zcbenz:
This PR is split from https://github.com/ml-explore/mlx/pull/1983. This PR relies on changes in https://github.com/ml-explore/mlx/pull/2259 and must be merged later.

Implements copy ops in CUDA backend, which uses the same strategy with the Metal kernels. Similar to the binary ops (#2259), the implementation currently only handles one element per thread.

The implementations of copy kernels are split into multiple files because of slow compilation speed. Combining them into one file would remove a lot of duplicate code, but would also increase compile time to 4x, usually 15 ~ 20 minutes.

There is also a `CastOp` helper added, which handles conversions between complex numbers and real numbers when doing copies.


awni (approved):
Looks great, thanks! I resolved the conflict and will merge shortly.

mlx/backend/cuda/copy.cu
```diff
@@ -0,0 +1,89 @@
...
+            ctype,
+            in,
+            out,
+            offset_in,
+            offset_out,
+            shape_collapsed,
+            strides_vec[0],
+            strides_vec[1],
+            dynamic_offset_in.value_or(array(0, int64)),
+            dynamic_offset_out.value_or(array(0, int64)));
```
awni: I think the ternary operator is a better choice here because it won't make an array if the default condition is chosen.. whereas here we may make the array and never use it.

awni:
I'm seeing this warning a lot building the cuda back-end:

```
/mnt/task_runtime/mlx/mlx/backend/cuda/copy/copy_general.cu(30): warning #20281-D: in whole program compilation mode ("-rdc=false"), a __global__ function template instantiation or specialization ("mlx::core::cu::copy_gg<    ::__nv_bfloat16,     ::__half, unsigned int> ") will be required to have a definition in the current translation unit, when "-static-global-template-stub" will be set to "true" by default in the future. To resolve this issue, either use "-rdc=true", or explicitly set "-static-global-template-stub=false" (but see nvcc documentation about downsides of turning it off)
```

zcbenz:
This change in `mlx/backend/cuda/CMakeLists.txt` meant to suppress this warning:

```cmake
# CUDA 12.8 emits warning #20280-D for copy kernels which is a false positive.
# Explicitly pass this flag to suppress the warning, it is safe to set it to
# true but the warning wouldn't be suppressed.
if(CMAKE_CUDA_COMPILER_VERSION VERSION_GREATER_EQUAL 12.8)
  target_compile_options(
    mlx
    PRIVATE "$<$<COMPILE_LANGUAGE:CUDA>:--static-global-template-stub=false>")
endif()
```

Can you try changing `12.8` to `12.8.0`?

awni:
@zcbenz not sure the warning above seems to have gone away.. though I still see this one:

```
nvcc warning : Support for offline compilation for architectures prior to '<compute/sm/lto>_75' will be removed in a future release (Use -Wno-deprecated-gpu-targets to suppress warning).
```

zcbenz:
The error means nvcc is going to discard support for targeting 70 in future, which is used by V100. I think we can just disable the warning for now.

PR #2261, opened 2025-06-10
CUDA backend: random

zcbenz:

This is a direct translation of the Metal random kernel to CUDA, the kernel accepts `grid_dims` as a parameter which certainly needs some rework, but it is not leaving debts in other files and we can revisit this later.


awni (approved):
Looks great!

PR #2263, opened 2025-06-10
Fix export to work with gather/scatter axis

awni:
As title.


angeloskath (approved):
👍

awni:
I've noticed our linux tests fail from time to time due to a segfault in the compiler.. seems like a g++ bug..

PR #2269, opened 2025-06-11
CUDA backend: reduce

zcbenz:

Implements reduce ops in CUDA backend. The implementation is mixed with thrust/CUB and custom kernels:

* For filling output with init values, `thrust::fill_n` is used.
* For reducing all and reducing the last dimension of contiguous input, the `cub::DeviceReduce` and `cub::DeviceSegmentedReduce` are used.
* For other types of reduce the Metal kernels are ported to CUDA:
  * The `row_reduce_small` and `row_reduce_looped` from `reduce_row.h`.
  * The `col_reduce_small` and `col_reduce_looped` from `reduce_col.h`.

There are also some changes made in the ported kernels:

* The kernels are split into different files to speed up compilation.
* `cub::LoadDirectBlocked` and `cub::StoreDirectBlocked` are used for loading data, they also have vectorized versions that can be used for optimization in future.
* `cub::BlockReduce` is used for reducing across the thread block, which however requires the block dims to be known on compile time, so I added a `MLX_SWITCH_BLOCK_DIM` macro to help dispatch `nthreads` to constexpr block dim.
* The `Limits` utility uses methods instead of static variables due to many CUDA standard libraries lack constexpr support. Similarly the `Reduce::init` variable in Metal is split into a method in separate `ReduceInit` class.


awni (approved):
LGTM, very nice!

PR #2280, opened 2025-06-12
[CUDA] RMSNorm and VJP

awni:
As title.

I might do a little refactor before we land this to reduce some of the common functions between layernorm and rms norm. But o/w straight-forward diff.

### DIFF
--- a/mlx/backend/cuda/rms_norm.cu
+++ b/mlx/backend/cuda/rms_norm.cu
@@ -1,0 +1,345 @@
+// Copyright © 2025 Apple Inc.
+
+#include "mlx/backend/cuda/device.h"
+#include "mlx/backend/cuda/iterators/strided_iterator.cuh"
+#include "mlx/backend/cuda/kernel_utils.cuh"
+#include "mlx/backend/cuda/reduce/reduce.cuh"
+#include "mlx/backend/gpu/copy.h"
+#include "mlx/dtype_utils.h"
+#include "mlx/fast_primitives.h"
+
+#include <cooperative_groups.h>
+#include <cooperative_groups/reduce.h>
+#include <nvtx3/nvtx3.hpp>
+#include <cub/block/block_load.cuh>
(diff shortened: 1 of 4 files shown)


mlx/backend/cuda/rms_norm.cu
```diff
@@ -0,0 +1,345 @@
...
+    } else {
+      auto x_copy = array(x.shape(), x.dtype(), nullptr, {});
+      copy_gpu(x, x_copy, CopyType::General, s);
+      out.copy_shared_buffer(x_copy);
+      return x_copy;
+    }
+  };
+
+  array o = set_output(inputs[0]);
+  const array& x = o.data_shared_ptr() ? o : out;
```
awni: @zcbenz I'm struggling to understand the purpose of this condition.  Shouldn't `o` always have data? And if shouldn't the output just be assigned to `x`? 

That's how it is in the [metal version](https://github.com/ml-explore/mlx/blob/main/mlx/backend/metal/normalization.cpp#L50-L51).

Note I copied that bit from the LayerNorm version you wrote which I have the same question about.

zcbenz: The code originated from here:

https://github.com/ml-explore/mlx/blob/c2dd81a8aa95e04d16c20c587a9faedec10c48ed/mlx/backend/metal/normalization.cpp#L86-L87

I meant to make the donation more clear, but looking at it again I wonder if I made it more confused, or if it might not be needed at all 🤔.

awni: Ahh .. right .. that is outdated. We no longer move data pointers out of arrays inside the primitives. All of that is managed by the eval run-time.

awni:
> Out of curiosity why do the read twice if the row fits in registers? Is it a first step or am I missing something?

Yea there is no "looped" kernel for large rows. The row may not fit in register. Maybe we'll want to change that.. for now I just went with how the @zcbenz implemented the LayerNorm

PR #2283, opened 2025-06-12
[CUDA] ternary with select op

awni:
As title.


awni:
CC @zcbenz

mlx/backend/cuda/ternary.cu
```diff
@@ -0,0 +1,181 @@
...
+  }
+}
+
+} // namespace cu
+
+template <typename Op>
+void ternary_op_gpu_inplace(
+    const std::vector<array>& inputs,
+    array& out,
+    std::string_view op,
```
zcbenz: This parameter is not needed here, binary_op used it for throwing errors.

angeloskath (approved):
Looks lovely!

PR #2288, opened 2025-06-14
Fix cuda gemm for bf16

awni:
There is a good table [here](https://docs.nvidia.com/cuda/cublas/#cublasltmatmul) that shows all the combinations of compute types, data types and scale types that are supported.


angeloskath (approved):
👍

PR #2291, opened 2025-06-14
Fix cuda arg reduce

awni:
Fix arg reduce.

FYI @zcbenz and others: `blockdim.z` cannot be larger than 64.

I also added a `MLX_ENABLE_TF32` which by default is on. And use tensor cores for fp32 when it's on.

Also fix the fp16 gemm. I think it needs the higher precision reduction otherwise it is not very good. It still uses tensorcores.. (checked with a benchmark).


angeloskath (approved):
🙏

PR #2294, opened 2025-06-15
Add ReLUSquared Activation Function

johnmai-dev:
## Proposed changes

Add `ReLUSquared` & `relu_squared`

Applicable to [`bitnet` ](https://github.com/ml-explore/mlx-lm/pull/219) and [`nemotron`](https://github.com/ml-explore/mlx-lm/blob/19287dc922cadb19c8ec50c29e031218b9ff6ce9/mlx_lm/models/nemotron.py#L50-L51)

Reference: 
- https://arxiv.org/abs/2109.08668v2
- https://github.com/huggingface/transformers/blob/d5d007a1a0f0c11a726a54c8f00bd71825f84d02/src/transformers/activations.py#L170-L178

## Checklist


awni:
Thanks for the contribution! We prefer not to add minor variations on activation functions to the core, especially if they are only used by one or two models and are relatively easy to implement.

I see we merged the equivalent PR in Swift which is not so great as there will be an inconsistency 🤔 

Nevertheless, I'm going to close this and we can revisit on a case by case basis in the future.

PR #2298, opened 2025-06-16
Cuda bug fixes 2

awni:
- Fixes more bugs. 
- Mostly rearranged remaining tests into bugs and otherwise grouped by NYI

### DIFF
--- a/mlx/backend/cuda/device/binary_ops.cuh
+++ b/mlx/backend/cuda/device/binary_ops.cuh
@@ -1,7 +1,7 @@
 // Copyright © 2025 Apple Inc.
 
-#include "mlx/backend/cuda/device/fp16_math.cuh"
 #include "mlx/backend/cuda/device/cucomplex_math.cuh"
+#include "mlx/backend/cuda/device/fp16_math.cuh"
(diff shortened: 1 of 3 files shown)


angeloskath (approved):
🚀

mlx/backend/cuda/device/utils.cuh
```diff
@@ -187,8 +187,8 @@ inline __host__ __device__ cuda::std::tuple<IdxT, IdxT, IdxT> elem_to_loc_nd(
 template <typename IdxT = int64_t>
 inline __host__ __device__ IdxT
 elem_to_loc_4d(IdxT elem, const int* shape, const int64_t* strides, int ndim) {
-  IdxT loc = elem_to_loc_nd<3>(elem, shape, strides);
-  for (int i = ndim - 1; i >= 3; --i) {
+  IdxT loc = 0;
+  for (int i = ndim - 1; i >= 0; --i) {
```
zcbenz, line 191: Is `elem_to_loc_4d` still needed as it is now almost the same with `elem_to_loc`?

awni: I think you're right it looks the same.. we should remove it.

PR #2302, opened 2025-06-17
[CUDA] Divmod, Partition, and sort fixes

awni:
- More fixes to sort
- Increase max dims to 10
- Add divmod
- Add partition and argpartition (just dispatch to sort for now like in metal)


angeloskath (approved):
👍

PR #2304, opened 2025-06-17
Make ptx cache settable by environment variable

angeloskath:
As title. Makes it easy to clear between updates to jitted kernels.


awni:
I was thinking we might need a way to recognize these as stale.. like maybe put the MLX version in the file path or something 🤔

awni (approved):
LGTM, thanks!

PR #2306, opened 2025-06-18
Build CUDA release in Circle

awni:
Right now it won't auto-build. You have to trigger it via Circle. We can switch that when ready.

The package gets uploaded to `mlx-cuda`.


mlx/backend/cuda/device.cpp
```diff
@@ -114,7 +114,7 @@ void CommandEncoder::synchronize() {
   std::future<void> f = p->get_future();
   add_completed_handler([p = std::move(p)]() { p->set_value(); });
   worker_.end_batch();
-  worker_.commit();
```
awni, line 117: Need to commit on the actual cuda stream otherwise there is a race between this commit and any commits on the stream.

PR #2307, opened 2025-06-20
Cuda perf tuning

awni:
A few minor fixes which also have a tiny impact on perf (~1-2 tok/sec) but are mostly the right thing to do.

### DIFF
--- a/mlx/backend/cuda/allocator.cpp
+++ b/mlx/backend/cuda/allocator.cpp
@@ -1,5 +1,6 @@
 // Copyright © 2025 Apple Inc.
 
+#include "mlx/utils.h"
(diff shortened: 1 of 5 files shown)


mlx/backend/cuda/allocator.cpp
```diff
@@ -14,9 +15,11 @@ namespace mlx::core {
 
 namespace cu {
 
+constexpr int page_size = 16384;
+
```
awni, line 19: The default page size on linux is 4096.. which is quite small. I use the same value we use in macos here.. which results in better cache hits and reuse (since we round up to the page size).

angeloskath (approved):
👍

PR #2308, opened 2025-06-20
Fix `update_modules()` when providing a subset

angeloskath:
`update_modules()` like `update()` should allow a subset of modules but when it comes to lists it requires an entry for each element in the list.

The change matches the behavior of `update()`.


awni (approved):
Thanks!!

PR #2314, opened 2025-06-24
[CUDA] Fix reductions

angeloskath:
I am not too happy with this but I think it is a start.

I attempted an (almost certainly premature) "optimization" that I think complicated the code a bit and is probably not worth it. When the input is transposed instead of reading out of order and writing in order, I opted to transpose the output the same way so that we can read and write in order. 

Here are some speed comparisons with PT on matrices `MxN` of type bfloat16.

row_reductions
col_reductions

It is obvious that our column reductions need work and we need to make our performance more consistent across sizes as well.

There are more weird issues to clear up like for instance the maximum complex number being `-inf - inf j` which is weird but also there is no real ordering for complex numbers so 🤷‍♂️


mlx/backend/cuda/reduce/init_reduce.cu
```diff
@@ -0,0 +1,51 @@
...
+    cu::CommandEncoder& encoder,
+    const array& in,
+    array& out,
+    Reduce::ReduceType reduce_type) {
+  // Allocate if needed
+  if (out.data_shared_ptr() == nullptr) {
+    out.set_data(allocator::malloc(out.nbytes()));
+  }
+
+  encoder.set_input_array(in);
```
awni: I don't think you need to add the input here as the following kernel does not depend on it.

awni: In fact maybe it makes sense to remove the input from the function signature entirely?

mlx/backend/cuda/reduce/reduce_ops.cuh
```diff
@@ -3,48 +3,89 @@
...
+  template <typename T>
+  __device__ void atomic_update(T* x, T y) {
+    atomic_reduce<T, Min>(x, y);
+  }
 };
 
 struct Max {
   template <typename T>
-  __device__ T operator()(T a, T b) {
+  __device__ __forceinline__ T operator()(T a, T b) {
```
awni, line 81: Curious, what motivated the `__forceinline__`? Just curious if we should be adding it in our other functors.

mlx/backend/cuda/reduce/all_reduce.cu
```diff
@@ -0,0 +1,141 @@
...
+
+    return std::make_tuple(blocks, threads, block_step);
+  };
+
+  int blocks, threads;
+  size_t block_step;
+  array x = in;
+
+  // Large array so allocate an intermediate and accumulate there
+  std::tie(blocks, threads, block_step) = get_args(x.size(), N_READS);
```
awni:
```suggestion
  array x = in;

  // Large array so allocate an intermediate and accumulate there
  auto [blocks, threads, block_step] = get_args(x.size(), N_READS);
```

mlx/backend/cuda/reduce/col_reduce.cu
```diff
@@ -152,127 +85,178 @@ template <
...
+  // - col_reduce_looped
+  //
+  //   It is a general strided reduce. Each threadblock computes the output for
+  //   a subrow of the fast moving axis. For instance 32 elements.
+  //
+  // Notes: As in row reduce we opt to read as much in order as possible and
+  // leave
+  //        transpositions as they are (contrary to our Metal backend).
+  //
+  //        Moreover we need different kernels for short rows and tuning
```
awni: Tiny nit: the formatting of this comment came out kind of weird..

mlx/backend/cuda/reduce/col_reduce.cu
```diff
@@ -152,127 +85,178 @@ template <
...
-
+    const ReductionPlan& plan,
+    cu::ColReduceArgs args) {
+  // Allocate data for the output using in's layout to access them as
+  // contiguously as possible.
+  allocate_same_layout(out, in, axes);
+
+  // Just a way to get out of the constness because cub doesn't like it ...
+  // (sigh)
+  array x = in;
```
awni: That is a weird restriction. But I would do that as a const cast on the pointer.. it's free vs a shared pointer copy.

mlx/backend/cuda/reduce/reduce_utils.cuh
```diff
@@ -0,0 +1,158 @@
...
+    out.set_data(allocator::malloc(out.nbytes()));
+    return;
+  }
+
+  if (out.ndim() < in.ndim()) {
+    throw std::runtime_error(
+        "Reduction without keepdims only supported for row-contiguous inputs");
+  }
+
+  // Calculate the transpositions applied to in in order to apply them to out.
```
awni, line 121:
```suggestion
  // Calculate the transpositions applied to in order to apply them to out.
```

mlx/backend/cuda/reduce/reduce_utils.cuh
```diff
@@ -0,0 +1,158 @@
...
+  }
+
+  // Calculate the resulting contiguity and do the memory allocation
+  auto [data_size, rc, cc] = check_contiguity(out.shape(), final_strides);
+  auto fl = in.flags();
+  fl.row_contiguous = rc;
+  fl.col_contiguous = cc;
+  fl.contiguous = data_size == out.size();
+  out.set_data(
+      allocator::malloc(out.nbytes()),
```
awni: That condition should always be true right? As in the output is always `contiguous`?

awni (approved):
Looks awesome! Left a few minor comments.

Lot's of green! More passing tests! 🚀

mlx/backend/cuda/reduce/init_reduce.cu
angeloskath: Unfortunately `ReduceInit` takes `T` not `U` for it's type so we need `x`, but yeah it isn't an input to the kernel... silly copy-paste.

Btw I haven't decided if `ReduceInit` should simply take `U`, I can't think of a place where it makes sense to look at `T` .

mlx/backend/cuda/reduce/reduce_ops.cuh
angeloskath: Probably not. I didn't see any perf difference. CUB uses it so I used it to see if it makes a difference 🤷‍♂️.

mlx/backend/cuda/reduce/all_reduce.cu
angeloskath: The thing is that capturing structured bindings is a C++26 feature. I also changed the one in `col_reduce` to remove the warnings.

awni: > The thing is that capturing structured bindings is a C++26 feature

Ahh 😞  I forgot about that.

PR #2316, opened 2025-06-25
Fix get 2d grid dims

### LINKED ISSUE #2315: [BUG] RuntimeError in scan ops
**Describe the bug**
Scan operations throw RuntimeError on certain input shape/dim combinations:
```
RuntimeError: Unable to safely factor shape.
```

**To Reproduce**
```python
import mlx.core as mx

tensor = mx.ones(shape=(2, 3, 4, 5), dtype=mx.float32)
result = mx.cumsum(tensor, axis=1)
print(result)
```
**Desktop:**
 - OS Version: [MacOS 15.5]
 - Version: [mlx==0.26.1]

**Additional context**
Error is thrown in [get_2d_grid_dims_common](https://github.com/ml-explore/mlx/blob/81bb9a2a9e21a54b9658a59f06d8f8b3d677f5e5/mlx/backend/common/utils.cpp#L203)

angeloskath:
Fixes #2315.

I checked that the divisor option is used in several situations but we never assume that the dimensions will be exactly divided just `ceildiv` so I have no idea why this was conservative and giving up instead of ceildividing 🤷‍♂️

### DIFF
--- a/mlx/backend/common/utils.cpp
+++ b/mlx/backend/common/utils.cpp
@@ -199,11 +199,14 @@ Dims get_2d_grid_dims_common(
       }
     }
   }
-  if (grid_y > UINT32_MAX || grid_x > UINT32_MAX || divisor > 1) {
+  if (grid_y > UINT32_MAX || grid_x > UINT32_MAX) {
     throw std::runtime_error("Unable to safely factor shape.");
   }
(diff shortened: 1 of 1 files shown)


awni (approved):
🙏

PR #2317, opened 2025-06-26
[CUDA] Switch to CUDA graphs

awni:
So far inference is a bit faster +5-7 tok/sec for Llama 7B on A100.

Further optimization is definitely possible and warranted. Any reduction in graph construction time which includes time spent in the primitive's `eval_gpu` and time building the graph itself should yield better perf as we are still bottlenecked by that. Ideally we get it low enough to hide the latency of graph building entirely for that benchmark.

Synchronization issues are fixed by committing the open graphs prior to encoding events. A future optimization would be to encode the synchronization events into the graph itself. This will be important if there is a lot of CPU/GPU intermingling to avoid splitting the graphs too frequently.

Transformer training is up to 7.3 it/sec for the 150M model on A100 with the following command:

```
python main.py --num_blocks 12 --batch_size 8 --dtype bfloat16 --gpu
```

### DIFF
--- a/mlx/backend/cuda/device.cpp
+++ b/mlx/backend/cuda/device.cpp
@@ -2,37 +2,22 @@
 
 #include "mlx/backend/cuda/device.h"
 #include "mlx/backend/cuda/worker.h"
-#include "mlx/backend/metal/metal.h"
+#include "mlx/utils.h"
 
 #include <fmt/format.h>
 #include <nvtx3/nvtx3.hpp>
 #include <future>
+#include <unordered_set>
 
 namespace mlx::core {
 
+// Can be tuned with MLX_MAX_OPS_PER_BUFFER
+// This should be less than 255
+constexpr int default_max_nodes_per_graph = 20;
+
+constexpr int max_graph_cache_size = 100;
+
 namespace cu {
-
-DeviceStream::DeviceStream(Device& device) : device_(device), stream_(device) {}
-
-void DeviceStream::synchronize() {
-  cudaStreamSynchronize(stream_);
-}
-
-cudaStream_t DeviceStream::schedule_cuda_stream() {
-  // TODO: Return a stream that maximizes parallelism.
-  return stream_;
-}
-
-cudaStream_t DeviceStream::last_cuda_stream() {
-  return stream_;
-}
-
-CommandEncoder& DeviceStream::get_encoder() {
-  if (!encoder_) {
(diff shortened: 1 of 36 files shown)


awni:
I'm moving this out of draft as the tests are clearing and it's faster for both the training and inference cases I tested. CC @zcbenz for your review as well.

mlx/backend/cuda/device.cpp
```diff
@@ -67,49 +52,253 @@ void Device::make_current() {
...
+    enc.graph_key_ += empty.id;
+    enc.graph_key_ += empty.node_type;
+  }
+
+  // Insert the input -> concurrent node dependencies without updating output
+  // nodes
+  auto outputs = std::move(enc.active_outputs_);
+  enc.insert_graph_dependencies(std::move(enc.concurrent_nodes_));
+
+  // Update output node to be the emtpy node
```
angeloskath: Nit: typo emtpy -> empty

angeloskath (approved):
I think this is pretty brilliant. I like it a lot. The graph key basically encodes all edges in the order they are created by `set_input_array()`. Concurrent context is handled by adding a NULL node that produces all the outputs so all subsequent nodes depend on that.

As a general comment for later, I do think that some logic is a bit spread out but as always it may be inevitable given that it is a complex bit of machinery. An example would be that `insert_graph_dependencies(node)` actually creates the node id but `insert_graph_dependencies(nodes)` doesn't. Similarly it isn't clear if/why we need the empty node count in the key and the subgraph node count. I would guess not but maybe I missed sth.

awni:
> An example would be that insert_graph_dependencies(node) actually creates the node id but insert_graph_dependencies(nodes) doesn't

Indeed, I don't love that either. I think that part of the design should / could use some refinement. 

> Similarly it isn't clear if/why we need the empty node count in the key and the subgraph node count. I would guess not but maybe I missed sth.

It's necessary because you also need the node types to match for the graph update to work. As a simple example consider a graph with a single node (which could be a kernel node, a subgraph node, or an empty node). We need to distinguish between each of those graphs to update the right one. So the node keys will be distinct (e.g. `1.0.0`, `1.1.0`, or `1.0.1`).

angeloskath:
> As a simple example consider a graph with a single node (which could be a kernel node, a subgraph node, or an empty node)

Oh interesting I hadn't thought about disconnected nodes. Cause if all nodes have edges, then you don't need it since the node type is in the edges.

PR #2318, opened 2025-06-27
Split broadcast so it is always fused in compile

angeloskath:
What this does becomes clear with an example. For the following function

```python
def f(a, b, c):
    bb = mx.broadcast_to(b, c.shape)
    return (a + bb * c), bb.sum()
```

the compiled graph changes as follows

broadcast_small

Benchmarking on transformer training the speedup is consistent but minimal (~3%-4%) across M2, M2 Ultra and A100.

Splitting broadcast is safe because it is in principle a noop, but we could consider splitting some copies (astype) and redoing them if it will allow us to fuse more things.

@awni I tried to think if there is a possibility of messing up the graph but I think it is ok since to split a broadcast means we certainly are going to produce a fused primitive.

### DIFF
--- a/mlx/compile.cpp
+++ b/mlx/compile.cpp
@@ -245,6 +245,30 @@ void merge(array& dst, array& src, ParentsMap& parents_map) {
   }
 }
 
+// Any parent in the divider will continue to refer to `x` but any parent not
+// in the divider will refer to a copy of the operation.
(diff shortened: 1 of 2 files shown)


awni:
That's a very elegant solution.. I didn't think of it.

PR #2320, opened 2025-06-29
MLX_SWITCH macros to templates

angeloskath:
This changes all macros used to switch types or values to constexprs into templates.

The main benefit is that we will now have proper compiler errors instead of one huge line from the preprocessor macros. There is very little change to binary size. When we move to C++20 we can make the lambda get the type directly so we won't have to call `MLX_GET_TYPE` on the argument.

@zcbenz let me know if you can think of a reason not to switch to this pattern instead.

The binary was previously 219MB and now 222MB which I think is worth it for the quality of life improvement.

### DIFF
--- a/mlx/backend/cuda/utils.cpp
+++ b/mlx/backend/cuda/utils.cpp
@@ -25,16 +25,38 @@ void check_cuda_error(const char* name, cudaError_t err) {
 }
 
 const char* dtype_to_cuda_type(const Dtype& dtype) {
-  if (dtype == float16) {
-    return "__half";
+  switch (dtype) {
+    case bool_:
+      return "bool";
+    case int8:
+      return "int8_t";
+    case int16:
+      return "int16_t";
+    case int32:
+      return "int32_t";
+    case int64:
+      return "int64_t";
+    case uint8:
+      return "uint8_t";
+    case uint16:
+      return "uint16_t";
+    case uint32:
+      return "uint32_t";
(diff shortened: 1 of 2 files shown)


zcbenz:
Getting rid of the macros would make dev experience much better, thanks for coming up with this!

awni (approved):
Looks awesome! Thanks for the update, really useful

zcbenz:
> When we move to C++20 we can make the lambda get the type directly so we won't have to call `MLX_GET_TYPE` on the argument.

This sounds an exciting new feature, where can I learn more about it?

angeloskath:
The [docs](https://en.cppreference.com/w/cpp/language/lambda.html) have the following.

> If the lambda definition uses an explicit template parameter list, that template parameter list is used with operator(). For each parameter in params whose type is specified as auto, an additional invented template parameter is appended to the end of that template parameter list:
> ```cpp
> // generic lambda, operator() is a template with two parameters
> auto glambda = []<class T>(T a, auto&& b) { return a < b; };
>
> // generic lambda, operator() is a template with one parameter pack
> auto f = []<typename... Ts>(Ts&&... ts)
> {
>     return foo([std::forward](https://en.cppreference.com/w/cpp/utility/forward.html)<Ts>(ts)...);
> };
> ```

The interesting bit is that since the template parameter is explicit we don't really need an argument for it. So no need for tag types.

PR #2321, opened 2025-06-29
Fix module update in strict mode

awni:
Relevant test case shows the issue. If there is a non module type in a container then we have an issue during strict module because `leaf_modules` uses empty dictionaries for placeholders.

```python
        class MyModel(nn.Module):
            def __init__(self):
                super().__init__()
                self.stuff = [nn.Linear(2, 2), 0, nn.Linear(2, 2)]

        m = MyModel()
        m.update_modules(m.leaf_modules())
```


angeloskath (approved):
🙏

PR #2322, opened 2025-06-30
Use fp32 for testing, add more complex ops

awni:
As title.

### DIFF
--- a/mlx/backend/cuda/device/unary_ops.cuh
+++ b/mlx/backend/cuda/device/unary_ops.cuh
@@ -27,6 +27,8 @@ struct ArcCos {
   __device__ T operator()(T x) {
     return acos(x);
   }
+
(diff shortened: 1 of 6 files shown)


python/mlx/nn/layers/positional_encoding.py
```diff
@@ -100,8 +100,8 @@ def __init__(
 
     def __call__(self, x):
         y = x[..., None] * self._sigmas
-        cosy = mx.cos(y)
-        siny = mx.sin(y)
+        cosy = mx.cos(y, stream=mx.cpu)
+        siny = mx.sin(y, stream=mx.cpu)
```
angeloskath: I guess these are forgotten here?

awni: Good catch, that was for debugging.

PR #2326, opened 2025-07-02
Fix compilation error from integral_constant

zcbenz:
Some environments give following compilation errors for CUDA backend:

```
codes/mlx/mlx/backend/cuda/rms_norm.cu(323): error: expression must have a constant value
                    has_w_constant(),
                    ^
Note #2767-D: the value of *this cannot be used as a constant
          detected during:
            instantiation of function "lambda [](auto)->auto [with <auto-1>=std::integral_constant<int, 32>]" at line 49 of /home/zhao_cheng/codes/mlx/mlx/backend/cuda/kernel_utils.cuh
            instantiation of "void mlx::core::dispatch_block_dim(int, F &&) [with F=lambda [](auto)->auto]" at line 71 of /home/zhao_cheng/codes/mlx/mlx/dtype_utils.h
            instantiation of "void mlx::core::dispatch_float_types(mlx::core::Dtype, std::string_view, F &&) [with F=lambda [](auto)->auto]" at line 314
```

I don't know whether it is a bug of CUDA or host compiler, but it can be worked around by using the `.value` member instead.

### DIFF
--- a/mlx/backend/cuda/layer_norm.cu
+++ b/mlx/backend/cuda/layer_norm.cu
@@ -364,7 +364,7 @@ void LayerNormVJP::eval_gpu(
               using DataType = cuda_type_t<MLX_GET_TYPE(type_tag)>;
               auto kernel = cu::layer_norm_vjp<
                   DataType,
-                  has_w_constant(),
+                  has_w_constant.value,
                   block_dim(),
                   N_READS>;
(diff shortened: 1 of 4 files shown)


awni (approved):
LGTM!

angeloskath:
@zcbenz Why does it not have the same issue with `block_dim()` for instance? I am pro consistency here I think ...

PR #2327, opened 2025-07-02
Remove unused `#define inf`

zcbenz:
Seems to be a leftover, tests still pass after removing it.


awni (approved):
LGTM!

awni:
Seems like we need this based on the test failure..

zcbenz:
Sorry I forgot to clear compile cache and the test falsely passed locally.

PR #2329, opened 2025-07-03
[CUDA] Add MLX_CUDA_GRAPH_CACHE_SIZE env for setting graph cache size

zcbenz:
When running tests on a platform with CUDA 12.2 I observed random crashes like this:

```
test_opt_einsum_test_cases (test_einsum.TestEinsum) ... malloc_consolidate(): invalid chunk size

    frame #15: 0x00007f4f787a7295 libmlx.so`libcudart_static_c252ed9cb831aa5dcf8b9f755d8bd06045640941 + 37
    frame #16: 0x00007f4f787f43d0 libmlx.so`cudaGraphExecDestroy + 432
    frame #17: 0x00007f4f78521529 libmlx.so`mlx::core::cu::clear_graphs(std::unordered_map<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, CUgraphExec_st*, std::hash<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > >, std::equal_to<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > >, std::allocator<std::pair<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const, CUgraphExec_st*> > >&) + 41
    frame #18: 0x00007f4f78526a8c libmlx.so`mlx::core::cu::CommandEncoder::commit() + 1468
    frame #19: 0x00007f4f7852a5df libmlx.so`mlx::core::cu::CudaEvent::record(mlx::core::Stream) + 95
    frame #20: 0x00007f4f7852b4f2 libmlx.so`mlx::core::Event::signal(mlx::core::Stream) + 82
    frame #21: 0x00007f4f7730ff1a libmlx.so`mlx::core::eval_impl(std::vector<mlx::core::array, std::allocator<mlx::core::array> >, bool) + 5322
    frame #22: 0x00007f4f77310e8f libmlx.so`mlx::core::eval(std::vector<mlx::core::array, std::allocator<mlx::core::array> >) + 239
    frame #23: 0x00007f4f77210d21 libmlx.so`mlx::core::array::eval() + 177
```

which seems to be memory corruption happened inside CUDA.

With some testings I found that the memory corruption happens with following steps:

1. Create a CUDA graph
2. Instantiate it and launch
3. Create a different CUDA graph
4. Destroy the previous instantiated executable graph

Without access to CUDA's source code I have no idea whether it is a bug of CUDA or we have misusing of CUDA APIs, but the tests run well on CUDA 12.4 and I think we have good reason to blame CUDA.

This PR disables graph cache for CUDA < 12.4, and it also ensures the executable graph is destroyed immediately after launching when graph cache is disabled.

### DIFF
--- a/mlx/backend/cuda/device.cpp
+++ b/mlx/backend/cuda/device.cpp
@@ -15,7 +15,12 @@ namespace mlx::core {
 // This should be less than 255
 constexpr int default_max_nodes_per_graph = 20;
 
-constexpr int max_graph_cache_size = 100;
+int cuda_graph_cache_size() {
+  static int cache_size = []() {
+    return env::get_var("MLX_CUDA_GRAPH_CACHE_SIZE", 100);
+  }();
+  return cache_size;
+}
 
 namespace cu {
 
@@ -252,10 +257,6 @@ void CommandEncoder::commit() {
       CHECK_CUDA_ERROR(cudaGraphAddDependencies(
           graph_, from_nodes_.data(), to_nodes_.data(), from_nodes_.size()));
     }
-    // TODO smarter cache policy
-    if (graph_cache_.size() > max_graph_cache_size) {
-      clear_graphs(graph_cache_);
-    }
 
     graph_key_ += ".";
     graph_key_ += std::to_string(node_count_);
@@ -280,6 +281,11 @@ void CommandEncoder::commit() {
           cudaGraphInstantiate(&graph_exec, graph_, NULL, NULL, 0));
     }
     CHECK_CUDA_ERROR(cudaGraphLaunch(graph_exec, stream_));
+
+    // TODO smarter cache policy
+    if (graph_cache_.size() > cuda_graph_cache_size()) {
+      clear_graphs(graph_cache_);
+    }
 
     // Reset state
     node_count_ = 0;


awni (commented):
This is a nice clean fix and I am good to merge it since it unblocks you. But

- Not using the graph cache is a big hit for performance. Probably worse than not using cuda graphs entirely. So we may consider disabling cuda graphs entirely for < 12.4. This would be fairly easy to do since the kernel launch API is basically the same as the graph node API. And the context manager for stream capture can be a no-op.
- Given the above it would be pretty useful to know why it doesn't work. It sounds like the graph update mechanism is causing the issue. It's not implausible it's a CUDA bug there... but maybe there is a simple work around that doesn't involve removing the cache entirely.

zcbenz:
I agree we should definitely seek better workarounds, but before that I would like to verify the scope of the crash: like whether it happens for other environments with CUDA 12.2, whether CUDA 11 works, whether OS and compiler matters.

Having this change would make it easier for me to further investigate the crash, and I can create an issue to make sure this is not forgotten behind.

zcbenz:
I have tested on a different platform, with exact same OS (ubuntu 22.04) and exact same cuda toolkit version (12.2), and the crash did not happen on it.

The only differences are cuda driver version (535.183.06 vs 550.90.12) and hardware (V100 vs A100), which is controlled by the platform and unconfigurable. So I think it is likely the bug lies in the cuda driver instead of cuda runtime, however it is quite hard to find another platform using old cuda driver and it is hard to verify this theory.

Since the bug is not related to cuda toolkit version, I have changed this PR to make the graph cache size configuration with an environment variable, so for machines affected by it we can set `MLX_CUDA_GRAPH_CACHE_SIZE=0` to work around it.

awni (approved):

PR #2331, opened 2025-07-04
Fix compilation with CUDA 11

zcbenz:
While it is not sure whether we are going to support CUDA 11 eventually (which still needs some efforts), it does not take too much trouble making MLX compile with CUDA 11 for now.

Most compilation error came from conversations between CUDA's half types and integers, which are solved with the `CastOp`/`cast_to` helpers. The `__cast` helper in `reduce_utils.cuh` is also replaced with `cast_to` as they are doing the same thing.

### DIFF
--- a/mlx/backend/cuda/device.cpp
+++ b/mlx/backend/cuda/device.cpp
@@ -264,19 +264,26 @@ void CommandEncoder::commit() {
     graph_key_ += std::to_string(graph_node_count_);
     graph_key_ += ".";
     graph_key_ += std::to_string(empty_node_count_);
-    auto [it, _] = graph_cache_.emplace(graph_key_, nullptr);
-    auto& graph_exec = it->second;
 
-    if (graph_exec != NULL) {
-      cudaGraphExecUpdateResultInfo update_result;
-      cudaGraphExecUpdate(graph_exec, graph_, &update_result);
-      if (update_result.result != cudaGraphExecUpdateSuccess) {
-        cudaGetLastError();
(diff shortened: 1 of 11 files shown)


mlx/backend/cuda/device.cpp
```diff
@@ -263,19 +263,27 @@ void CommandEncoder::commit() {
...
+
+    cudaGraphExec_t& graph_exec = graph_cache_[graph_key_];
+
+    if (graph_exec != nullptr) {
+      cudaGraphExecUpdateResult update_result = cudaGraphExecUpdateError;
+#if CUDART_VERSION >= 12000
+      cudaGraphExecUpdateResultInfo info;
+      if (cudaGraphExecUpdate(graph_exec, graph_, &info) == cudaSuccess) {
+        update_result = info.result;
+      }
```
awni: Does it work to simplify this a bit?

```suggestion
      cudaGraphExecUpdateResult update_result;
#if CUDART_VERSION >= 12000
      cudaGraphExecUpdateResultInfo info;
      cudaGraphExecUpdate(graph_exec, graph_, &info)
      update_result = info.result;
```

awni (approved):
Looks good! I left a minor comment, lmk if it's applicable. O/w good to merge!

mlx/backend/cuda/device.cpp
zcbenz: Yeah it looks better, thanks!

awni:
Can you run the format hooks? Sorry I may have given you a badly formatted suggestion.

zcbenz:
Updated.

PR #2335, opened 2025-07-05
MoE backward improvements

angeloskath:
This PR improves the gradient computation for our `gather_mm` operations used predominantly for MoEs.

- Adds a new op `segmented_mm`
  The op computes a matmul where the reduced over dimension is segmented and each result is saved separately. When the indices are sorted this allows us to compute the gradients wrt the weights without having to instantiate one matrix per token.
- Removes an unnecessary copying in the computation of the jacobian. When the indices are sorted there is no need to scatter the result, it is already in the correct order.
- Implements gradients for the scales and biases in the quantized gather matmul.
- Fixes the scales and biases gradient for batched qmm.


mlx/ops.cpp
```diff
@@ -4649,6 +4649,49 @@ array gather_mm(
...
+  if (a.ndim() != 2 || b.ndim() != 2) {
+    throw std::invalid_argument("[segmented_mm] Batched matmul not supported");
+  }
+
+  if (segments.ndim() < 1 || segments.shape().back() != 2) {
+    std::ostringstream msg;
+    msg << "[segmented_mm] The segments should have shape (..., 2) but "
+        << segments.shape() << " was provided.";
+    throw std::invalid_argument(msg.str());
+  }
```
awni, line 4666: Probably good to require the segments be integral as it's consistent with how we treat other index-like variables.

mlx/backend/metal/indexing.cpp
```diff
@@ -1,5 +1,6 @@
 // Copyright © 2023-2024 Apple Inc.
 #include <fmt/format.h>
+#include <iostream>
```
awni: forgotten

awni:
This is very cool!

One question (to which the answer is probably no, but I'll ask anyway). Does it make sense to make this a fast op vs a core op? On the one-hand it doesn't currently have a gradient and could easily have a fallback implementation.  On the other hand.. maybe it should have a gradient (at some point) and be a full primitive?

awni (approved):

mlx/backend/metal/indexing.cpp
angeloskath: Yeah it is interesting how that got there.

I spent about half a day in contiguous time trying to figure out why the reference gradient in the test was wrong on CI. It appears that the CI allocator is not failing outright but misbehaving in weird ways. At least this is my conclusion as it was failing on CI and succeeding on any computer I tried.

awni: Curious.. but it's passing now? Did you fix something (other than the metal validation stuff below)?

angeloskath: I reduced the size of the matrices.

The weird part was that I was getting two matrices `128x1024x1024` both contiguous, both non-0 and their multiplication was all 0s. I was also evaluating everything just before so it wasn't some weird interaction. I should check but it should be repeatable completely outside of testing I think.

angeloskath:
> Fast op or not

Good point. My plan was to add the gradient, it should be fairly simple as it amounts to a sorted gather matmul. Otoh it is going to be a fairly seldomly used op. I will address the comments and merge unless you think we should make it a fast op (which I don't think from your comment above).

awni:
> I will address the comments and merge

Sounds good!

PR #2336, opened 2025-07-06
Build with all cpu cores by default

zcbenz:
It is tedious having to set `CMAKE_BUILD_PARALLEL_LEVEL` every time when setting up a new environment.


awni:
Nice! Would you mind removing it from the relevant commands in [.circleci/config](https://github.com/ml-explore/mlx/blob/main/.circleci/config.yml)

zcbenz:
Updated.

Issue #577, opened 2024-01-29
Different outputs of mlx.core.max using cpu/gpu


Redmept1on:
**Describe the bug**
Different(maybe wrong) outputs of mlx.core.max using cpu/gpu compared to numpy

**To Reproduce**

```python
import mlx.core as mx
import numpy as np
array_mlx = mx.array([[1, 2, 3], [float('nan'), 5, 6]], dtype=mx.float32)

mx.set_default_device(mx.gpu)
max_gpu = mx.max(array_mlx, axis=0)
print("Max on GPU:", max_gpu)

mx.set_default_device(mx.cpu)
max_cpu = mx.max(array_mlx, axis=0)
print("Max on CPU:", max_cpu)

a_np = np.array([[1, 2, 3], [float('nan'), 5, 6]], dtype=np.float32)
max_np = np.max(a_np, axis=0)
print("Max using NumPy:", max_np)
```
![image](https://github.com/ml-explore/mlx/assets/84373897/58b9b684-1f8e-461f-8398-b2b816352b0f)

**Desktop (please complete the following information):**
 - OS Version: MacOS 14.2.1
 - Version 0.0.10

awni:
Actually this one is not so easy to fix as we use the `simd` instructions in Metal. The `NaN` are being suppressed there.

zcbenz:
This issue had been fixed by https://github.com/ml-explore/mlx/pull/2339.

PR #2339, opened 2025-07-06
Align max op nan propagation with NumPy

jhavukainen:
## Proposed changes

Propagating NaNs in the `max` op according to numpy expectations. Circumvents MSLs tendency to avoid addressing NaNs with explicit checks. Addresses issue: #577 

Adds template specializations to `operator()` member in Max to address NaNs for non-integral types and `complex64` type separately. `simd_reduce_impl` uses the inequality property outlined in MSL specification section 3.1 §6 to check if any thread has data containing NaN and if so returns NaN.

Added unittests for functionality matching expected NaN propagation rules for both C++ and python side.

Added benchmarking both on python and C++ sides to allow verification of the effects from the additional NaN checking. Using the following benchmark before and after the changes in this PR produces the results below, indicating the effect should be relatively small.

```
void time_max() {
  auto a = mx::random::normal({10000, 10000});
  auto indices = mlx::core::array({1});
  auto updates = mlx::core::reshape(mlx::core::array({NAN}), {1, 1, 1});
  std::vector<int> axes{0};

  auto b = scatter(a, {indices}, updates, axes);
  mx::eval(b);

  auto max_along_0 = [&b]() { return mx::max(b, 0, false); };
  TIME(max_along_0);

  auto max_along_1 = [&b]() { return mx::max(b, 1, false); };
  TIME(max_along_1);
}
```

Produced values for three repeated measurements with minimal background activity on an M2 Air device as follows

```
// Before changes 
Benchmarks for Device(gpu, 0)
Timing max_along_0 ... 7.0018 msec
Timing max_along_1 ... 4.9778 msec
Benchmarks for Device(gpu, 0)
Timing max_along_0 ... 7.012 msec
Timing max_along_1 ... 4.9994 msec
Benchmarks for Device(gpu, 0)
Timing max_along_0 ... 6.983 msec
Timing max_along_1 ... 5.0911 msec

// After changes
Benchmarks for Device(gpu, 0)
Timing max_along_0 ... 7.015 msec
Timing max_along_1 ... 5.0204 msec
Benchmarks for Device(gpu, 0)
Timing max_along_0 ... 7.023 msec
Timing max_along_1 ... 5.0087 msec
Benchmarks for Device(gpu, 0)
Timing max_along_0 ... 7.0377 msec
Timing max_along_1 ... 4.9938 msec
```

Possible missing topics that need addressing:
I did not address the `atomic_update` nan-propagation as I didn't manage to find a readily available test case exercising the code path and wasn't exactly sure of what the expected behavior should be there.

If the changes look good, it should be trivial to extend the same treatment for min op to bring it to parity with numpy as well.

## Checklist

- [ x ] I have updated the necessary documentation (if needed)

### DIFF
--- a/benchmarks/cpp/single_ops.cpp
+++ b/benchmarks/cpp/single_ops.cpp
@@ -192,6 +192,17 @@ void time_reductions() {
 
   auto argmin_along_1 = [&a]() { return mx::argmin(a, 1, false); };
   TIME(argmin_along_1);
+
+  auto indices = mlx::core::array({1});
+  auto updates = mlx::core::reshape(mlx::core::array({NAN}), {1, 1, 1});
+  std::vector<int> axes{0};
+  auto b = scatter(a, {indices}, updates, axes);
+  mx::eval(b);
+
+  auto max_along_0 = [&b]() { return mx::max(b, 0, false); };
+  TIME(max_along_0);
+  auto max_along_1 = [&b]() { return mx::max(b, 1, false); };
+  TIME(max_along_1);
 }
 
 void time_gather_scatter() {
--- a/mlx/backend/metal/kernels/reduction/ops.h
+++ b/mlx/backend/metal/kernels/reduction/ops.h
@@ -187,7 +187,10 @@ struct Max {
 
   template <typename T>
   T simd_reduce_impl(T val) {
-    return simd_max(val);
+    if(simd_any(val != val)) {
+      return static_cast<T>(NAN);
+    }
+      return simd_max(val);
   }
 
   static constexpr constant U init = Limits<U>::min;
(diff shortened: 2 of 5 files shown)


benchmarks/cpp/single_ops.cpp
```diff
@@ -192,6 +192,17 @@ void time_reductions() {
 
   auto argmin_along_1 = [&a]() { return mx::argmin(a, 1, false); };
   TIME(argmin_along_1);
+
+  auto indices = mlx::core::array({1});
+  auto updates = mlx::core::reshape(mlx::core::array({NAN}), {1, 1, 1});
```
awni: nit: `mx::` instead of `mlx::core`

mlx/backend/metal/kernels/reduction/ops.h
```diff
@@ -187,6 +187,9 @@ struct Max {
 
   template <typename T>
   T simd_reduce_impl(T val) {
+    if (simd_any(val != val)) {
```
angeloskath, line 195: I would also only add this for non integral types.

angeloskath (commented):
Looks great! Just adding that the check for the simd op should also be enabled only for non-integral types.

jhavukainen:
@awni, @angeloskath  Thanks for the initial review and comments! I implemented the fix in cleaning up the namespace using the alias and the check for non-integral types.

I gave the CPU implementation the same treatment and tested that it passes the new nan propagation tests for max when running with `DEVICE=cpu` env variable. While testing I found a small fib in the int8 and int16 dispatch dtypes for max on the CPU that made the max tests fail if the input array contained both negative and positive values, so I added a fix there.

jhavukainen:
If all looks good, this can be merged in and I'll prepare another one for the min op.

awni:
> While testing I found a small fib in the int8 and int16 dispatch dtypes for max on the CPU that made the max tests fail if the input array contained both negative and positive values, so I added a fix ther

Good catch!

awni:
Couple things:
- The tests are failing from a syntax error which I think may be a Python 3.9 thing
- The CUDA backend is also failing the NaN propagation test. Are you comfortable fixing that? (I'd imagine it's similar fix to what you did for the Metal backend). If not we can also add those tests to be [skipped](https://github.com/ml-explore/mlx/blob/main/python/tests/cuda_skip.py) for now

jhavukainen:
Changed the tuple unpacking to use more friendly syntax for earlier python versions, that should fix the syntax issue.

I think I'll need to brush up on cuda a bit before implementing it on that side as well. Added the tests to the cuda skiplist for now.

PR #2340, opened 2025-07-07
Fix layernorm race condition

### LINKED ISSUE #2333: [BUG] Faulty behaviour for 0.26.2
**Describe the bug**
Broken behaviour in downstream mflux library when using v0.26.2 (for more info see https://github.com/filipstrand/mflux/issues/229). 

Image

**To Reproduce**
Lacking a self-contained mlx-only example right now, but in mflux running the following command

```
mflux-generate --model schnell --prompt "Luxury food photograph" --steps 2 --seed 2 -q 8
```

should produce an image like in 0.26.1 shown above. Instead we get a corrupted image. 
 

 - OS Version: MacOS 15.5
 - MLX Version: 0.26.2

**Additional context**
Currently, mflux mostly have high-level "integration tests" for the image generation, so I have not been able to easily pinpoint exactly where things go wrong. My guess is that this is related to https://github.com/ml-explore/mlx/pull/2246 (2e8cf0b4), since resetting to the commit before that, (24f89173), resolves the issue. Let me know if there is something I could assist with or try out on my end!

angeloskath:
Closes #2333.

There is a race in `threadgroup_sum` because we are doing two passes and using the same shared memory to accumulate in. So basically when we write the normalizer all threads must have finished reading the mean.

### DIFF
--- a/mlx/backend/metal/kernels/layer_norm.metal
+++ b/mlx/backend/metal/kernels/layer_norm.metal
@@ -31,6 +31,7 @@ inline void threadgroup_sum(
   for (int i = 0; i < N; i++) {
     x[i] = simd_sum(x[i]);
   }
+  threadgroup_barrier(mem_flags::mem_threadgroup);
   if (simd_lane_id == 0) {
     for (int i = 0; i < N; i++) {
       xs[N * simd_group_id + i] = x[i];


awni (approved):
Thanks!!

PR #2342, opened 2025-07-08
[CUDA] Do vectorized store/load in contiguous elementwise ops

zcbenz:
Apply the optimization of https://github.com/ml-explore/mlx/pull/2330 to all contiguous elementwise ops, I only profiled the unary and copy ops and both showed similar improvements on memory throughput (770 GB/s to 810 GB/s).

Also with a few small fixes:
* Use `int32_t` instead of `uint32_t` as `IdxT`.
* Rename `binary_xxx` to `binary_two_xxx` in `binary_two.cu` to avoid confusions and possible conflicts.

### DIFF
--- a/mlx/backend/cuda/copy/copy_contiguous.cu
+++ b/mlx/backend/cuda/copy/copy_contiguous.cu
@@ -13,7 +13,7 @@ namespace cg = cooperative_groups;
 template <typename In, typename Out, typename IdxT, int N_READS>
 __global__ void copy_s(const In* in, Out* out, IdxT size) {
   IdxT index = cg::this_grid().thread_rank();
-  int remaining = size - index * N_READS;
+  IdxT remaining = size - index * N_READS;
   if (remaining <= 0) {
     return;
   }
@@ -37,7 +37,7 @@ __global__ void copy_s(const In* in, Out* out, IdxT size) {
 template <typename In, typename Out, typename IdxT, int N_READS>
 __global__ void copy_v(const In* in, Out* out, IdxT size) {
   IdxT index = cg::this_grid().thread_rank();
-  int remaining = size - index * N_READS;
+  IdxT remaining = size - index * N_READS;
   if (remaining <= 0) {
     return;
   }
--- a/mlx/backend/cuda/copy/copy_contiguous.cu
+++ b/mlx/backend/cuda/copy/copy_contiguous.cu
@@ -13,21 +13,16 @@ namespace cg = cooperative_groups;
 template <typename In, typename Out, typename IdxT, int N_READS>
 __global__ void copy_s(const In* in, Out* out, IdxT size) {
   IdxT index = cg::this_grid().thread_rank();
-  IdxT remaining = size - index * N_READS;
-  if (remaining <= 0) {
-    return;
-  }
 
-  if (remaining < N_READS) {
-    for (int i = 0; i < remaining; ++i) {
-      IdxT offset = index * N_READS + i;
-      out[offset] = CastOp<In, Out>{}(in[0]);
+  if ((index + 1) * N_READS > size) {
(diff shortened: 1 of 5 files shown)


mlx/backend/cuda/copy/copy_contiguous.cu
```diff
@@ -37,16 +71,23 @@ void copy_contiguous(
     int64_t out_offset) {
   dispatch_all_types(in.dtype(), [&](auto in_type_tag) {
     dispatch_all_types(out.dtype(), [&](auto out_type_tag) {
-      dispatch_bool(out.data_size() > UINT32_MAX, [&](auto large) {
+      dispatch_bool(out.data_size() > INT32_MAX, [&](auto large) {
```
awni: I think there is a bug here when using `int` that we should address before it gets to messy. In `get_2d_grid_dims_common` we use `UINT32_MAX` as the largest possible size for a grid element (see [here](https://github.com/ml-explore/mlx/blob/main/mlx/backend/common/utils.cpp#L151-L153)). So I think you could end up with a case where you overflow `int32`.

awni: In the Metal back-end we always use `UINT32_MAX` for the condition for contiguous elementwise ops and `uint` for the index type.. so we don't have this issue. But if we are going to use `int` for the index type then we need a version of that function which let's you specify the maximum allowed value.

Alternatively we could use `uint` for the index type with CUDA and rework the contiguous kernels so they don't rely on signed numbers.

awni: For example you could change the kernels like below which would also be a bit simpler and probably just as fast and allow for unsigned indexing:


```c++
  IdxT remaining = size - index * N_READS;
  if (remaining <= 0) {
    return;
  }

  if (remaining < N_READS) {
    for (int i = 0; i < remaining; ++i) {
      IdxT offset = index * N_READS + i;
      out[offset] = Op{}(a[offset], b[offset], c[offset]);
    }
  } else {
```

To 

```c++
  if (size >= (index + 1) * N_READS) {
    for (int i = index * N_READS; i < size; ++i) {
      out[i] = Op{}(a[i], b[i], c[i]);
    }
  } else {
```

awni:
Thanks for the update! I left a comment re a bug with signed vs unsigned maximum size for the contiguous kernels. I think my preferred option is to rework the change (if it is doable) to enable unsigned index types for the contiguous kernels. But lmk what you think.

zcbenz:
Thanks explaining the bug and I get why uint should be used as index now. I have updated the code as suggested.

awni (approved):
Thanks! Will merge when the tests clear.

awni:
I'm not sure if the test failure is related to this diff:

```
test_multi_output_leak (test_array.TestArray.test_multi_output_leak) ... terminate called after throwing an instance of 'std::runtime_error'
  what():  cudaEventDestroy(event_) failed: an illegal memory access was encountered

Received "aborted" signal
```

zcbenz:
I can reproduce similar crashes locally.

zcbenz:
So the crash happened because I changed `data_size() > INT32_MAX` to `data_size() > UINT32_MAX`, but I'm still a bit confused of the reason. Also the `copy_contiguous` also uses `data_size() > UINT32_MAX`, should it be changed?

awni:
> So the crash happened because I changed data_size() > INT32_MAX to data_size() > UINT32_MAX

Which one are you referring to?

zcbenz:
> Which one are you referring to?

Sorry that turns out not to be related, the crash does not happen every time and it is a bit hard to locate the cause.

zcbenz:
I think my mistake was accidentally using `uint32_t` as index for `unary_g`. I have updated the PR and tests are passing on my local machines.

PR #2346, opened 2025-07-09
Align mlx::core::min op nan propagation with NumPy

jhavukainen:
## Proposed changes

Follow-up to https://github.com/ml-explore/mlx/pull/2339, extending the same treatment to the min op so NaNs get propagated.

Adds the similar NaN returns to non-integral op types and the specialized treatment for complex types. Extends the reduce nanpropagation tests to also test the min case. Adds the benchmarks for min.

## Checklist

- [ x ] I have updated the necessary documentation (if needed)


awni (approved):
Awesome, thanks!

awni:
The test failure isn't from this diff.  We have a flaky quantization test that NaN propagation exposes. But we'll need to fix that in a separate PR.

PR #2347, opened 2025-07-10
[CUDA] Implement Scan kernel

zcbenz:
This PR ports the Scan kernels from Metal to CUDA.

There are also a few changes to fix the `LogAddExp` op for complex numbers which caused the scan tests to fail:
* A few C++ tests are added to show the failed case.
* A `cexpf.cuh` file is added to port the complex number `exp` from thrust, we can not use `thrust::exp` directly unfortunately because it does not compile in JIT kernels when building with old CUDA versions (<= 12.2).
* The complex number `log1p` is moved from `utils.cuh` to `unary_ops.cuh` so it is put together with other complex number functions.
* The `LogAddExp`'s implementation is updated to use PyTorch's algorithm.


awni:
Looks awesome! I'm wondering if those logaddexp tests will fail on Metal / CPU as well?

awni (approved):

zcbenz:
The exp test fails on Metal, should I port the `cexpf`?

awni:
>  The exp test fails on Metal, should I port the cexpf?

Yes that would be super useful.. it looks like a bug as is.

zcbenz:
Updated to fix the exp on Metal.

PR #2351, opened 2025-07-10
[CUDA] Set current device before cudaGraphLaunch

zcbenz:
This makes sure the graph is launched in the correct device under a multi-device environment.

Also moves the code of `Device::get_command_encoder` up as it is in the middle of `CommandEncoder` method definitions.


awni:
Test failure is unrelated.

PR #2352, opened 2025-07-10
[CUDA] Put version in ptx cache dir path

zcbenz:
This makes sure the compilation cache gets invalidated when MLX is updated. The downside is when doing local development the `/tmp/mlx/` could be filled with lots of dirs since date and commit are part of the version too.

Also renames the `MLX_PTX_CACHE` env to `MLX_PTX_CACHE_DIR` make it more intuitive, since it sets the cache dir path instead of cache options.


awni:
Test failure is unlrelated.

PR #2354, opened 2025-07-10
[CUDA] Affine quantize

awni:
Add affine quantization.

Benchmark runs at 900 gb/s on an A100. So about 45% of peak. It's a good start.. maybe more room to optimize if needed.

```python
import time
import mlx.core as mx

n = 8192
x = mx.random.normal((n, n))
w, s, b = mx.quantize(x, bits=4, group_size=64)

def fun(x):
    for _ in range(100):
        w, s, b = mx.quantize(x, bits=4, group_size=64)
        x = mx.dequantize(w, s, b, bits=4, group_size=64)
    return x

for _ in range(10):
    mx.eval(fun(x))


tic = time.time()
for _ in range(10):
    mx.eval(fun(x))
toc = time.time()

sec = toc - tic
gb = (10 * 100 * (x.nbytes + w.nbytes + s.nbytes + b.nbytes) * 2) / 1e9

print(gb / sec)
```

### DIFF
--- a/mlx/backend/cuda/quantized.cu
+++ b/mlx/backend/cuda/quantized.cu
@@ -1,0 +1,365 @@
+// Copyright © 2023-2024 Apple Inc.
+
+#include <cassert>
+
+#include "mlx/backend/cuda/device.h"
+#include "mlx/backend/cuda/device/binary_ops.cuh"
+#include "mlx/backend/cuda/kernel_utils.cuh"
+#include "mlx/backend/gpu/copy.h"
+#include "mlx/dtype_utils.h"
+#include "mlx/fast_primitives.h"
+
+#include <cooperative_groups.h>
+#include <cooperative_groups/reduce.h>
+#include <nvtx3/nvtx3.hpp>
+
+namespace mlx::core {
+namespace cu {
+
+namespace cg = cooperative_groups;
+
+template <int bits, int wsize = 8>
+inline constexpr __device__ short get_pack_factor() {
+  return (bits == 3 || bits == 5) ? 8 : (bits == 6 ? 4 : wsize / bits);
+}
+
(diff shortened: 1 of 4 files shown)


angeloskath (approved):
🚀

I guess the dequantization benchmark would be more interesting eg for quantized embeddings and I also think it would be the faster one so it looks great also in terms of perf.

mlx/backend/cuda/quantized.cu
```diff
@@ -0,0 +1,383 @@
...
+__global__ void
+affine_quantize(const T* w, uint8_t* out, T* scales, T* biases, size_t size) {
+  auto block_size = cg::this_thread_block().dim_threads();
+  auto block_idx = cg::this_thread_block().group_index();
+  auto idx_in_block = cg::this_thread_block().thread_index();
+
+  auto tidx = block_idx.x * block_size.x + idx_in_block.x;
+  auto tidy = block_idx.y * block_size.y + idx_in_block.y;
+
+  auto grid_dim = cg::this_grid().dim_threads();
```
zcbenz, line 39: `dim_threads` is not available in CUDA 12.2:

```
/home/zhao_cheng/codes/mlx/mlx/backend/cuda/quantized.cu(39): error: class "cooperative_groups::__v1::grid_group" has no member "dim_threads"
    auto grid_dim = cg::this_grid().dim_threads();
                                    ^

/home/zhao_cheng/codes/mlx/mlx/backend/cuda/quantized.cu(156): error: class "cooperative_groups::__v1::grid_group" has no member "dim_threads"
    auto grid_dim = cg::this_grid().dim_threads();
```

It is also not documented somehow:
https://docs.nvidia.com/cuda/cuda-c-programming-guide/#grid-group

PR #2355, opened 2025-07-10
Fix edge check in QuantizedBlockLoader for qmm_n

angeloskath:
As title. We were checking the wrong edges.


awni (approved):
Wow that's subtle!!

PR #2356, opened 2025-07-11
Install linux with mlx[cuda] and mlx[cpu]

awni:
This changes the way we install backends:

We build four packages:
- mlx-cpu
- mlx-cuda
- mlx-metal
- mlx 

The top-level `mlx` package is basically a shell for the dependency management. So you can do:

- `pip install mlx` to install MLX for macOS with Metal backend
- `pip install "mlx[cpu]"` to install MLX for linux on the CPU.
- `pip install "mlx[cuda]"` to install MLX for linux with the CUDA back-end

For local builds nothing should change. For PyPi release the build now goes in two stages:
- Build the back-end package. This is python version independent but platform and back-end dependent.
- Build the mlx core python binding package. This is platform and version dependent, but back-end independent.

### DIFF
--- a/docs/src/install.rst
+++ b/docs/src/install.rst
@@ -38,8 +38,16 @@ and SM 7.0 (Volta) and up. To install MLX with CUDA support, run:
 
 .. code-block:: shell
 
-    pip install mlx-cuda
+    pip install "mlx[cuda]"
 
+CPU only (Linux)
+^^^^^^^^^^^^^^^^
+
+For a CPU-only version of MLX that runs on Linux use:
+
+.. code-block:: shell
+
+    pip install "mlx[cpu]"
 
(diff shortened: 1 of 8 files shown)


docs/src/install.rst
```diff
@@ -23,13 +23,6 @@ To install from PyPI you must meet the following requirements:
     MLX is only available on devices running macOS >= 13.5
     It is highly recommended to use macOS 14 (Sonoma)
 
-
-MLX is also available on conda-forge. To install MLX with conda do:
```
altaic, line 27: Are conda builds being dropped?

awni: We never officially supported conda so I removed from the docs. The source of truth distribution will be the one on PyPi.

Issue #2348, opened 2025-07-10
[CUDA] Ship CCCL in python distribution


zcbenz:
The CUDA backed is currently built with a fixed version of CCCL:

https://github.com/ml-explore/mlx/blob/85873cb162d0802c925350cbc68e1410bce3f1ad/mlx/backend/cuda/CMakeLists.txt#L92-L97

which makes sure the code compiles across different CUDA versions.

However for JIT kernels the code are built with the CCCL shipped by system-installed CUDA toolkit, which could be many years old and result in a few problems:

* JIT currently does not work with CUDA 11 at all.
* We have to provide a few wrappers in the case of an old CCCL, for example [the atomic ops](https://github.com/ml-explore/mlx/blob/main/mlx/backend/cuda/device/atomic_ops.cuh).
* We can't use some CCCL features, which leads to complex workarounds like in #2347.

A solution is to ship source code of CCCL as part of MLX's python distribution, and make JIT compiler search headers from it.

The size increased for distribution would not be very large:

```console
$ du -hs build/temp.linux-x86_64-cpython-310/mlx.core/_deps/cccl-src/
17M	build/temp.linux-x86_64-cpython-310/mlx.core/_deps/cccl-src/
```

PR #2357, opened 2025-07-11
[CUDA] Bundle CCCL for JIT compilation

zcbenz:
Close https://github.com/ml-explore/mlx/issues/2348.

Bundle CCCL as part of python distribution (installed in the `include` directory along with MLX headers), and make JIT compiler search headers from it. This ensure the JIT compilation gets the same result with host compilation.

Currently only [libcudacxx](https://github.com/NVIDIA/cccl/blob/main/libcudacxx) (i.e. `cuda/std/xxx` headers) are shipped since the JIT code (under `backend/cuda/device`) only uses it. It would be very easy to also ship CUB headers if needed. However thrust currently is not compatible with JIT (https://github.com/NVIDIA/cccl/issues/4305) and we should avoid using it in JIT code.

This PR also removes code that was used to work around old CCCL versions.


mlx/backend/common/utils.cpp
```diff
@@ -11,6 +13,17 @@ std::string get_primitive_string(Primitive* primitive) {
   return op_t.str();
 }
 
+std::filesystem::path current_binary_dir() {
```
awni, line 16: There was a comment about this needing to stay inline. I think the purpose of that is to make sure upstream dependencies of MLX get the right binary path. Could you make it inline in the header?

awni: Maybe a better change would be to have a wrapper around the inline version which sets the static variable so it gets properly cached. But the one used by `register_library` in metal device.h should be inline in a header.

zcbenz: The inline version of `register_library` got removed in https://github.com/ml-explore/mlx/pull/2055 (not sure if it is intentional), so I think there is no longer needs making this inline? The only place using it is `device.cpp` so it only returns the dir containing `libmlx.so`.

awni: Hmm.. good point. `get_library` can be called by third-party extensions. If it is called without a path then it will attempt to look for a metal library where the current binary is located. Presumably that will be the `libmlx` now.. which I would have thought is a bug 🤔

awni: Either way this PR doesn't change that behavior so let's move forward with it.

PR #2361, opened 2025-07-11
lower memory uniform sampling

awni:
```python
import mlx.core as mx
vocabSize = 10000
hiddenSize = 2048
w = mx.random.uniform(shape=[vocabSize, hiddenSize], dtype=mx.bfloat16)
mx.eval(w)
print(mx.get_peak_memory() / 2**20)
```

Uses 117MB insteed of 156MB. The uniform sampling is basically equivalent to sampling then `asytpe`. But this is pretty important to get well distributed samples. Updated normal and laplace to do the same.

### DIFF
--- a/mlx/random.cpp
+++ b/mlx/random.cpp
@@ -82,6 +82,16 @@ array split(const array& key, int num, StreamOrDevice s /* = {} */) {
   return bits({num, 2}, 4, key, s);
 }
 
+// Get the next representable value below 1.0 for half precision
+// floating point types (fp16, bf16)
+template <typename T>
+T below_one() {
+  T f = T(1.0);
+  uint16_t* m = (uint16_t*)&f;
+  *m -= 1;
+  return f;
+}
+
 array uniform(
     const array& low,
     const array& high,
@@ -106,7 +116,23 @@ array uniform(
         << " from broadcasted shape " << out_shape << ".";
     throw std::invalid_argument(msg.str());
   }
-  auto upper = array(std::nextafter(1.0f, 0.0f), float32);
+
+  // Get random values between [0, nextafter(1.0, 0.0)] since samples must
+  // be in [low, high)
+  auto get_upper = [&dtype]() {
+    switch (dtype) {
+      case float32:
+        return array(std::nextafter(1.0f, 0.0f), float32);
+      case float16:
+        return array(below_one<float16_t>(), float32);
+      case bfloat16:
+        return array(below_one<bfloat16_t>(), float32);
+      default:
+        throw std::runtime_error("[uniform] Unsupported type.");
+    }
+  };
+
+  auto upper = get_upper();
   auto maxval = array(std::numeric_limits<uint32_t>::max(), float32);
   auto out = bits(shape, size_of(float32), key, stream);
   out = divide(out, maxval, stream);
@@ -154,6 +180,10 @@ array normal(
     StreamOrDevice s /* = {} */) {
   if (dtype == complex64) {
     return complex_normal(shape, loc, scale, key, s);
+  } else if (!issubdtype(dtype, floating)) {
+    throw std::invalid_argument(
+        "[normal] Can only generate uniform numbers with "
+        "floating point type.");
   }
 
   auto stream = to_stream(s);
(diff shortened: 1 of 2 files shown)


angeloskath (approved):
I was trying to do this with bit manipulation but it is a bit annoying to write with ops I think. The hard part is in choosing the exponent with a probability proportional to the bin width. For any specific bin it is pretty easy to make the float by shifting and or-ing with the exponent.

So let's merge this :-)

awni:
Wdyt about something like [this](https://github.com/jax-ml/jax/blob/main/jax/_src/random.py#L437-L474)?

angeloskath:
Yeah exactly but I think that's not good enough. It is easy to do but for bfloat16 for instance it only gives 128 numbers in [0, 1) but there are something like 127 * 128. Even for float32 it gives 2**23 vs 2**23 * 127.

Do you think I am overthinking it?

awni:
> Even for float32 it gives 223 vs 223 * 127.

Wow really? That seems like a major bug.

> Do you think I am overthinking it?

This uniform sampling has been a recurrent issue. I think if we could find a proper long term fix with a bit of extra effort and put it to bed that would be nice because I don't think I fully fixed it.

angeloskath:
> Wow really? That seems like a major bug.

Sigh I meant to write `2^23` and somehow it became 223. But yeah basically you only randomize the mantissa which gives you uniformly spread random numbers but with the largest possible spread between them.

angeloskath:
Btw just to make it clear why what Jax does is a bit weird, the following generates 2.6k bfloat16 numbers when casting from float32. 382 for MLX and 128 for Jax.

```python
import mlx.core as mx
import jax

def count_unique_mlx(dta, dtb):
    x = mx.random.normal((100000,), dtype=dta).astype(dtb)
    return len(set(x.tolist()))

k = jax.random.PRNGKey(0)
def count_unique_jax(dta, dtb):
    x = jax.random.normal(k, (100000,), dtype=dta).astype(dtb)
    return len(set(x.tolist()))

print(count_unique_mlx(mx.float32, mx.float32))  # 99768
print(count_unique_mlx(mx.float32, mx.float16))  # 14739
print(count_unique_mlx(mx.float32, mx.bfloat16)) # 2680
print()
print(count_unique_mlx(mx.float32, mx.float32))   # 99775
print(count_unique_mlx(mx.float16, mx.float16))   # 3054
print(count_unique_mlx(mx.bfloat16, mx.bfloat16)) # 382
print()
print(count_unique_jax(jax.numpy.float32, jax.numpy.float32))  # 99442
print(count_unique_jax(jax.numpy.float32, jax.numpy.float16))  # 14704
print(count_unique_jax(jax.numpy.float32, jax.numpy.bfloat16)) # 2682
print()
print(count_unique_jax(jax.numpy.float32, jax.numpy.float32))   # 99442
print(count_unique_jax(jax.numpy.float16, jax.numpy.float16))   # 1024
print(count_unique_jax(jax.numpy.bfloat16, jax.numpy.bfloat16)) # 128
```

Doing something like the following (I have to check my float math a bit but it seems correct in that the distribution is uniform and in 0-1) gives 1024 numbers 🤷‍♂️

```python
def int2float(x, ebits, mbits):
    exponent = (x >> mbits) & (2**(ebits - 1) - 1)
    ldz = ebits
    while exponent != 0:
        ldz -= 1
        exponent >>= 1
    new_exponent = 2**(ebits - 1) - 1 - ldz
    new_number = (new_exponent << mbits) | (x & (2**mbits - 1))
    return new_number

x = mx.random.randint(0, 65536, (100000,), dtype=mx.uint16)
y = list(int2float(xi, 8, 7) for xi in x.tolist())
v = mx.array(y, dtype=mx.uint16).view(mx.bfloat16)
print(len(set(v.tolist())))
```

I don't know if it is worth it tbh. The idea is literally sample an exponent with geometric distribution by counting the leading zeros of the 8 bits of exponent. This obviously doesn't give us all possible numbers either since we got 127 values and we can get 8 of them. But the widths of the rest are also quite small hence the improvement.

We should probably choose between
- Always use float32 to generate random numbers and then cast to dtype (best accuracy - worst memory)
- Generate only the mantissa with a fixed exponent (worst accuracy - best memory)
- Create an int2float op and generate using that (medium accuracy - best memory)

I would vote for number 1 tbh.

awni:
I'm on board with 1. It does seem a lot better in terms of randomness. It's also better than what we have now in terms of memory use.

PR #2362, opened 2025-07-12
[CUDA] Do not put kernels in annoymous namespace

zcbenz:
In the profiler the kernels would have `<unnamed>::` in their names, making them hard to read.


awni (approved):
LGTM!

PR #2363, opened 2025-07-12
Only call get_primitive_string on error

zcbenz:
The `get_primitive_string` function is very expensive to call and can delay kernel launch by 0.5~4µs.

### DIFF
--- a/mlx/backend/cuda/binary.cu
+++ b/mlx/backend/cuda/binary.cu
@@ -177,7 +177,7 @@ template <typename Op>
 void binary_op_gpu_inplace(
     const std::vector<array>& inputs,
     array& out,
-    std::string_view op,
+    Primitive* primitive,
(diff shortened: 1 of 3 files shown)


zcbenz:
Note that I can not apply the same optimization on Metal backend because there are code passing op name to `binary_op_gpu`:

```
/Users/zcbenz/codes/mlx/mlx/backend/metal/fft.cpp:403:3: error: no matching function for call to 'binary_op_gpu'
  403 |   binary_op_gpu({temp, w_k_broadcast}, temp1, "Multiply", s);
      |   ^~~~~~~~~~~~~
```

A better (but much larger) fix would be making it cheap to get primitive's name.

zcbenz:
Close in favor of https://github.com/ml-explore/mlx/pull/2365.

PR #2365, opened 2025-07-13
Add Primitive::name and remove Primitive::print

zcbenz:
Remove the `Primitive::print()` method that takes `std::ostream`, and replace it with a `Primitive::name()` method that returns `const char*` directly. This can reduce launch time by 0.5~4µs for some kernels.

This would be a breaking change since it requires rewrites in extensions.


angeloskath (approved):
This looks great and is so much better than before!

We can merge after the tests pass.

zcbenz:
Updated to fix the `MLX_METAL_JIT=ON` build.

awni:
I believe the test failure is unrelated to the diff (we're seeing flaky linux builds with the compiler erroring for unknown reasons).

PR #2368, opened 2025-07-14
[CUDA] Add work per thread to compile

angeloskath:
Adds a work per thread of 4 to compile. It is added as a template parameter so we can tune it but 4 seemed to be the fastest and doesn't increase the compilation time given that we don't increase the number of kernels.

Training a 153M transformer bs=8 seq_len=1024 before 7.4 it/s now 8.4 it/s .

Next steps should include using vectorized loads for contiguous kernels and maybe vectorized stores for all kernels.


mlx/backend/cuda/compiled.cpp
```diff
@@ -227,6 +291,7 @@ void Compiled::eval_gpu(
 
   auto kernel = mod.get_kernel(kernel_name);
   auto [num_blocks, block_dims] = get_launch_args(kernel, outputs[0], large);
```
awni: `get_launch_args` takes `work_per_thread` as the last variable. Maybe better to use than dividing `num_blocks.x` for the large kernels.

PR #2370, opened 2025-07-14
update linux build

awni:
Update linux test to use newer ubuntu.

This uses the default python on the machine which is 3.13. Seems ok to me since we test 3.9 in all the other tests.. maybe even good to run it on a different version. But I'm happy to change it to 3.9 if you prefer.


angeloskath (approved):
Thanks! We are testing 3.9 on macOS. It should be fine.

Issue #2349, opened 2025-07-10
[CUDA] Use a specialized complex number class instead of cuComplex


zcbenz:
The CUDA backed currently uses `cuComplex` for complex number, which is just an alias of `float2` and leads to a few problems:

* We have to define custom operators for `cuComplex`.
* The custom operators would then affect the other uses of `float2`.
* The interface of `cuComplex` is hard-to-read: i.e. `cuCrealf` and `cuCimagf`.
* `cuComplex` can not have implicit conversations between `thrust::complex` or `cuda::std::complex` which makes it hard to use the math functions provided by CCCL.

I think we have 3 choices:

1. Define our own complex number class (PyTorch does this but probably for historical reasons).
2. Use [`thrust::complex`](https://nvidia.github.io/cccl/thrust/api/structthrust_1_1complex.html).
3. Use [`cuda::std::complex`](https://nvidia.github.io/cccl/libcudacxx/standard_api/numerics_library/complex.html) (According to https://github.com/NVIDIA/cccl/issues/1000 it has performance issues due to need of conforming to C++ standard).

awni:
Any downside of option 2? Seems pretty reasonable to me. We could use an alias for the name if needed.

zcbenz:
So far I don't see any downside, but I would have a check of what other frameworks use before working on this.

angeloskath:
One possible downside we have to check is whether vectorization and alignment with `cub::LoadDirectBlockVectorized` works correctly. It shouldn't be the case but the docs say

> The following conditions will prevent vectorization and loading will fall back to cub::BLOCK_LOAD_DIRECT:
>
> - ITEMS_PER_THREAD is odd
> - The data type T is not a built-in primitive or CUDA vector type (e.g., short, int2, double, float2, etc.)

zcbenz:
Unfortunately `thrust::complex` is out of choice because it can not be used in JIT code (`#include <thrust/complex.h>` results in JIT compilation errors and there is an issue: https://github.com/NVIDIA/cccl/issues/4305).

I think a practical solution is to use `cuda::std::complex` for now, and then switch to whatever faster implementation in future.

PR #2372, opened 2025-07-15
[CUDA] Use cuda::std::complex in place of cuComplex

zcbenz:
Close https://github.com/ml-explore/mlx/issues/2349.

Use `cuda::std::complex` for complex numbers and remove utilities built around `cuComplex`. It is not the fastest implementation but we can easily switch to a faster one in future by changing the type alias.


angeloskath (approved):
Awesome!

PR #2377, opened 2025-07-16
[CUDA] Fix complex reduce + nan propagation in min and max

awni:
- Fix complex reductions
- Fix nan propagation for cuda min and max reductions
- Enable working tests


angeloskath (approved):
🙏

PR #2379, opened 2025-07-16
Add contiguous_copy_gpu util for copying array

zcbenz:
Add an utility function to avoid copy-pasting the array copy code.

### DIFF
--- a/mlx/backend/gpu/copy.h
+++ b/mlx/backend/gpu/copy.h
@@ -43,4 +43,7 @@ void copy_gpu_inplace(
 // Fill the output with the scalar val
 void fill_gpu(const array& val, array& out, const Stream& s);
 
+// Return a contiguous array with same shape that copies the data of |arr|.
+array contiguous_copy_gpu(const array& arr, const Stream& s);
+
 } // namespace mlx::core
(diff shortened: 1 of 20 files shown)


mlx/backend/gpu/copy.h
```diff
@@ -43,4 +43,7 @@ void copy_gpu_inplace(
 // Fill the output with the scalar val
 void fill_gpu(const array& val, array& out, const Stream& s);
 
+// Return a contiguous array with same shape that copies the data of |arr|.
+array make_contiguous_array_gpu(const array& arr, const Stream& s);
```
awni: A bit of a nit.. wdyt about the name `contiguous_copy_gpu` instead? It kind of fits better as a copy utility. Or `make_contiguous_gpu` (having `array` in the name seems superfluous..)

zcbenz: `contiguous_copy_gpu` sounds a better name, I have updated the PR.

awni (approved):
Thanks for the refactor! Very nice!

zcbenz:
Updated to fix building in CUDA backend.

PR #2382, opened 2025-07-17
[CUDA] Add more ways finding CCCL headers in JIT

zcbenz:
The JIT compilation needs a certain version of CCCL headers, and the python binding just ships the headers inside the distribution. However for pure C++ programs there is no universal way to ship the headers with the program, in order to make JIT compilation work in C++ programs, there are new ways added to find CCCL headers:

* A `MLX_CCCL_DIR` environment variable.
* A `MLX_CCCL_DIR` compile-time define.

This will fixes some C++ tests failing with old CUDA versions.


awni (approved):
Lgtm

PR #2385, opened 2025-07-18
[CUDA] Initial implementation of Convolution with cuDNN

zcbenz:
With this PR most of the C++ convolution tests pass, however running some of the python convolution tests would result in crashes like this:

```
* thread #1, name = 'tests', stop reason = signal SIGSEGV: invalid address (fault address: 0x4)
  * frame #0: 0x00007fffc1c4bc34 libcuda.so.1`___lldb_unnamed_symbol3755 + 20
    frame #1: 0x00007fffc1c4d139 libcuda.so.1`___lldb_unnamed_symbol3756 + 1753
    frame #2: 0x00007fffc1f184eb libcuda.so.1`___lldb_unnamed_symbol9811 + 347
    frame #3: 0x00007fffc1c65d3f libcuda.so.1`___lldb_unnamed_symbol3824 + 511
    frame #4: 0x00007fffc1c66376 libcuda.so.1`___lldb_unnamed_symbol3825 + 278
    frame #5: 0x00007fffc1c55d66 libcuda.so.1`___lldb_unnamed_symbol3787 + 374
    frame #6: 0x00007fffc1b783b4 libcuda.so.1`___lldb_unnamed_symbol2185 + 356
    frame #7: 0x00007fffc1d14b7d libcuda.so.1`___lldb_unnamed_symbol4932 + 509
    frame #8: 0x00007fffeb271885 libmlx.so`libcudart_static_3e45c59a2e8f97a77d8f5a68e115126b1450c788 + 53
    frame #9: 0x00007fffeb2bfd76 libmlx.so`cudaGraphInstantiate + 486
    frame #10: 0x00007fffea8d3b65 libmlx.so`::cudaGraphInstantiate(pGraphExec=0x0000555564cf91b8, graph=0x0000555556f9de10, pErrorNode=0x0000000000000000, pLogBuffer=0x0000000000000000, bufferSize=0) at cuda_runtime.h:439:32
    frame #11: 0x00007fffea8ddceb libmlx.so`mlx::core::cu::CommandEncoder::commit(this=0x0000555555e22770) at device.cpp:316:7
```

which only happens for certain dtypes and shapes of inputs, and also depends on hardware.

I'll continue looking into this, but I think current work is in good shape and can be checked in.


awni:
FYI: I added the cudnn dep for Python as well.

zcbenz:
Can you rerun CI to check if my last commit fixes the build?

zcbenz:
For some settings the conv op expects the input to have pointer alignment to be passed, which is not documented anywhere and the API is not even available in cuDNN's recommended frontend APIs.

I have updated the PR to use the backend APIs, and it should function the same with PyTorch's cuDNN convolution backend now.

There are still many convolution settings not accepted by the forward conv op of cuDNN, the remaining works are at least:
* Add backward conv primitive.
* Add transpose conv primitive.
* Provide fallback kernels for settings not supported by cuDNN.

PyTorch's [_select_conv_backend function](https://github.com/pytorch/pytorch/blob/d36afac83b01c3de214db91f3d4b3f447f9a77b7/aten/src/ATen/native/Convolution.cpp#L1187-L1199) provides some hints on what's not handled by cuDNN.

zcbenz:
The updated code is ready for review.

mlx/backend/cuda/device.cpp
```diff
@@ -41,9 +52,12 @@ Device::Device(int device) : device_(device) {
   // The cublasLt handle is used by matmul.
   make_current();
   cublasLtCreate(&lt_);
+  // The cudnn handle is used by Convolution.
+  CHECK_CUDNN_ERROR(cudnnCreate(&cudnn_));
 }
 
 Device::~Device() {
+  CHECK_CUDNN_ERROR(cudnnDestroy(cudnn_));
   cublasLtDestroy(lt_);
```
awni: I guess we should also check cublas error when making and destroying the cublas handles right?

zcbenz: Yeah we should do that, I updated the PR.

mlx/backend/cuda/conv.cpp
```diff
@@ -0,0 +1,344 @@
...
+
+#include <cassert>
+#include <numeric>
+
+namespace mlx::core {
+
+namespace {
+
+// Not all engines support it so can not use this API now.
+#define MLX_USE_CUDNN_NATIVE_CUDA_GRAPH_API 0
```
awni, line 26: What do you mean by "engines" here? Which "engines" don't support this?

awni: It looks super convenient.. so I'm wondering why we aren't using it and what it would take to switch it on.

mlx/backend/cuda/conv.cpp
```diff
@@ -0,0 +1,344 @@
...
+      auto plan = cudnn_frontend::ExecutionPlanBuilder()
+                      .setHandle(encoder.device().cudnn_handle())
+                      .setEngineConfig(config, op_graph_tag)
+                      .build();
+      if (execute_plan(encoder, plan, in, wt, out)) {
+        conv_cache().emplace(cache_key, std::move(plan));
+        return true;
+      }
+    } catch (cudnn_frontend::cudnnException&) {
+    }
```
awni, line 232: Does this really need a try catch? I feel like there should be a better way to figure out if a plan will work or not?

mlx/backend/cuda/conv.cpp
```diff
@@ -0,0 +1,344 @@
...
+      fixed_vector(padding_lo_),
+      fixed_vector(padding_hi_),
+      fixed_vector(kernel_strides_),
+      fixed_vector(kernel_dilation_),
+      groups_,
+      get_alignment(in),
+      get_alignment(wt),
+      get_alignment(out)};
+  auto it = conv_cache().find(cache_key);
+  if (it != conv_cache().end()) {
```
awni: nit: C++ has a nice idiom to scope variables to the if condition:

```suggestion
  if (auto it = conv_cache().find(cache_key); it != conv_cache().end()) {
```

mlx/backend/cuda/conv.cpp
```diff
@@ -0,0 +1,344 @@
...
+  return cache;
+}
+
+template <typename T, typename U>
+inline std::vector<T> convert_vector(const std::vector<U>& vec) {
+  return std::vector<T>(vec.begin(), vec.end());
+}
+
+template <typename T>
+inline std::array<T, MAX_NDIM> fixed_vector(const std::vector<T>& vec) {
```
awni, line 56: Why not reuse: https://github.com/ml-explore/mlx/blob/main/mlx/backend/cuda/kernel_utils.cuh#L103-L112 ?

mlx/backend/cuda/conv.cpp
```diff
@@ -0,0 +1,344 @@
...
+
+  cudnn_frontend::EngineConfigList filtered_configs;
+  cudnn_frontend::filter(configs, filtered_configs, [dtype](auto c) {
+    if (cudnn_frontend::hasNumericalNote<
+            CUDNN_NUMERICAL_NOTE_DOWN_CONVERT_INPUTS>(c)) {
+      return true;
+    }
+    // In PyTorch, tf32 seems to be always turned off for convolution even with
+    // "torch.backends.cudnn.allow_tf32 = True", so we are disabling tf32 too to
+    // keep results same.
```
awni: I don't get this.. why would you not want to use tensor cores for convolutions? Should we just rely on the env variable to [enable TF32](https://github.com/ml-explore/mlx/blob/main/mlx/utils.h#L152-L155)?

awni: >  tf32 seems to be always turned off for convolution even with "torch.backends.cudnn.allow_tf32 = True"

That can't be true can it? It would mean PyTorch never uses tensor cores for convolutions which I don't think is right. This [article](https://docs.pytorch.org/docs/stable/notes/cuda.html#tensorfloat-32-tf32-on-ampere-and-later-devices) would be pretty misleading otherwise.

At any rate.. I think a default in MLX of using TF32 where appropriate and being able to shut it off with the env var is reasonable. We can always change the default to off if it makes sense.

mlx/backend/cuda/conv.cpp
jagrit06: Engines are basically cudnn implementation providers - so for a given conv / shape, you have a few valid ways of doing the op - each of those represent different engines with different attributed 
We are able to filter the available engines to only consider those that are able to natively build on the graph API - but I'm not sure if the left over sets will be enough

zcbenz: "engine" is the word cudnn uses to refer to kernels I think. When I experimented with the native cuda graph API lots of conv tests failed to find a working engine.

mlx/backend/cuda/conv.cpp
zcbenz: The C++ `cudnn_frontend::ExecutionPlanBuilder` API throws exceptions on error. We can use the C APIs to check the error code instead but the code would become tediously long.

mlx/backend/cuda/conv.cpp
zcbenz: The array must be filled with 0 for unused dims here, also using `kernel_utils.cuh` would turn this file into a `.cu` file which slows down compilation.

mlx/backend/cuda/conv.cpp
zcbenz: I made a silly mistake testing PyTorch on CPU instead of GPU 🤦. I'll use the env.

zcbenz:
Updated the PR to fix tf32 support.

mlx/backend/cuda/conv.cpp
awni: That's interesting.. so if we do want to switch this on .. then we would need a fallback for when an engine isn't found?

zcbenz: It is possible that an engine that is compatible with native cuda graph API is not the most performant one, so switching this on could end up using a slower engine. I think we would only want to use this API when all engines get supported, and fallback wouldn't be needed then.

zcbenz:
> test_torch_conv_2D (test_conv.TestConv.test_torch_conv_2D) ... terminate called after throwing an instance of 'std::runtime_error'
>  what():  Unable to find an engine for convolution.

I actually saw this failure before when building with cuDNN 8, but I confirm the CI builds with cuDNN 9.

zcbenz:
The `test_torch_conv_2D` only failed for one specific setting:

```python
print(N, C, O, idim, kdim, stride, padding, dtype)
> 1 6 1 (1, 1) (1, 1) (1, 1) (0, 0) bfloat16
```

But PyTorch succeeded to run conv for the that setting.

At current point I'm unable to tell whether it is an unsupported setting of cuDNN on this specific hardware, or I missed something when using cuDNN APIs. Usually I would do a custom build of PyTorch to check whether it is actually using cuDNN backend for this conv op and what parameters it is passing, but it is challenging to do in the CI machine (there is a 2h time limit).

The test passed on a A100. I also tested on V100 however the machine only works with cuDNN 8 which [does not support conv2d for half types](https://github.com/pytorch/pytorch/blob/392fa75411a1f293e891395f005615b257c03eda/aten/src/ATen/native/Convolution.cpp#L443-L447).

@awni What do you think if we skip `test_torch_conv_2D` for now?

awni:
> @awni What do you think if we skip test_torch_conv_2D for now?

Sounds good to me. Let's land this PR and continuing from there.

PR #2389, opened 2025-07-19
[CUDA] speedup handling scalars

awni:
Speedup using scalars in CUDA managed memory back-end.

#### LLM generation

```bash
mlx_lm.generate --model meta-llama/llama-2-7b-chat-hf --prompt "Write a story about Einstein" -m 240
```
Pre: 79.93 tok/sec
Post: 93.2 tok/sec

#### Training:
```bash
python main.py --num_blocks 12 --batch_size 8 --dtype bfloat16 --gpu
```

Pre: 8.60 it/sec
Post: 8.65 it/sec

### DIFF
--- a/mlx/backend/cuda/allocator.cpp
+++ b/mlx/backend/cuda/allocator.cpp
@@ -25,8 +25,7 @@ constexpr int small_block_size = 8;
 constexpr int small_pool_size = 4 * page_size;
 
 SmallSizePool::SmallSizePool() {
-  CHECK_CUDA_ERROR(
-      cudaMallocManaged(&buffer_, small_block_size * small_pool_size));
+  CHECK_CUDA_ERROR(cudaMallocManaged(&buffer_, small_pool_size));
   end_ = reinterpret_cast<void*>(
       reinterpret_cast<char*>(buffer_) + small_pool_size);
   next_free_ = reinterpret_cast<Block*>(buffer_);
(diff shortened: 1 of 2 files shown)


mlx/backend/cuda/allocator.cpp
```diff
@@ -17,6 +17,53 @@ namespace cu {
 
 constexpr int page_size = 16384;
 
+// Any allocations smaller than this will try to use the small pool
+constexpr int small_block_size = 8;
+
+// The small pool size in bytes. This should be a multiple of the host page
```
angeloskath, line 23: nit: It is actually the size in blocks given that you multiply with `small_block_size` below.

angeloskath (approved):
Fantastic!

mlx/backend/cuda/allocator.cpp
```diff
@@ -17,6 +17,53 @@ namespace cu {
...
+// size and small_block_size.
+constexpr int small_pool_size = 4 * page_size;
+
+SmallSizePool::SmallSizePool() {
+  CHECK_CUDA_ERROR(
+      cudaMallocManaged(&buffer_, small_block_size * small_pool_size));
+  end_ = reinterpret_cast<void*>(
+      reinterpret_cast<char*>(buffer_) + small_pool_size);
```
angeloskath, line 30: btw I just noticed that should be either `small_block_size * small_pool_size` or the allocation above should be `small_pool_size`.

awni: Oops that's a bug.

PR #2392, opened 2025-07-19
Cuda simplify allocator

awni:


mlx/backend/cuda/allocator.cpp
```diff
@@ -138,9 +140,7 @@ void CudaAllocator::free(Buffer buffer) {
   if (get_cache_memory() < max_pool_size_) {
     buffer_cache_.recycle_to_cache(buf);
   } else {
-    lock.unlock();
-    cuda_free(buf->data);
-    delete buf;
+    cuda_free(buf);
```
angeloskath, line 143: This is a change of behavior in that we are now feeing under lock. Presumably it was on purpose just not sure why. Is it for the scalar pool?

mlx/backend/cuda/allocator.h
```diff
@@ -24,13 +21,14 @@ struct CudaBuffer {
 
 class SmallSizePool {
  private:
-  struct Block {
+  union Block {
```
angeloskath, line 24: I didn't see that in my first pass. Do we really need to save 100KB by making this a union? If it were a struct we wouldn't need the little bit of math in `malloc` 🤷‍♂️ It's probably fine as it is.

angeloskath (approved):
Looks good. I left only one nitpick comment.

mlx/backend/cuda/allocator.cpp
```diff
@@ -25,52 +24,58 @@ constexpr int small_block_size = 8;
 constexpr int small_pool_size = 4 * page_size;
 
 SmallSizePool::SmallSizePool() {
-  CHECK_CUDA_ERROR(cudaMallocManaged(&buffer_, small_pool_size));
-  end_ = reinterpret_cast<void*>(
-      reinterpret_cast<char*>(buffer_) + small_pool_size);
-  next_free_ = reinterpret_cast<Block*>(buffer_);
-
   auto num_blocks = small_pool_size / small_block_size;
+  buffer_ = new Block[num_blocks];
```
zcbenz, line 28: Since `small_pool_size / small_block_size` is constptr, it should be possible to make `buffer_` a static array:

```
 private:
  Block buffer_[small_pool_size / small_block_size];
```

mlx/backend/cuda/worker.h
```diff
@@ -24,45 +27,33 @@ class Worker {
...
   // finish running.
   void commit(cudaStream_t stream);
 
-  // Return how many batches have been added but not committed yet.
-  size_t uncommited_batches() const {
-    return uncommited_batches_;
-  }
-
  private:
+  friend void signal_worker(void*);
```
zcbenz: I think a static method should be enough here:

```
static void signal_worker(void*);
```

mlx/backend/cuda/worker.h
```diff
@@ -24,45 +27,33 @@ class Worker {
...
+
+  void signal_();
```
zcbenz: Rename `signal_` => `signal` for method name.

mlx/backend/cuda/allocator.cpp
awni: Indeed.. I considered it. It's a bit more complicated since we'd have to move that bit to the header. And I'm not sure it's a great idea / buys us much to put such a large array on the stack.

If you have any insights to best practices there lmk.

zcbenz: That is reasonable, heap allocation sounds a better choice here.

mlx/backend/cuda/allocator.cpp
awni: It's mostly a matter of convenience. As far as I know the way the locking was done is inherited from how it was done in Metal (we unlock before calling into the system malloc/free). The purpose there was performance (it is faster). But here it doesn't seem to matter and since it's more convenient to grab the lock outside this function.. I didn't bother changing it.

mlx/backend/cuda/allocator.h
awni: Maybe I'm missing something.. but I think the math to figure out `i` in `malloc` is necessary even if I make this a `struct` cause I need to get the index into the underlying cuda data buffer.

mlx/backend/cuda/worker.h
awni: Good suggestion.. that simplifies things a bit.

PR #2397, opened 2025-07-21
Add contiguous_copy_cpu util for copying array

zcbenz:
This is the CPU version of https://github.com/ml-explore/mlx/pull/2379, it does not save much lines of code but I think it is worth doing for consistency.


awni:
LGTM, thanks!

PR #2399, opened 2025-07-21
use size option in binary

awni:
Use size minimizing build. Reduces libmlx.so from ~250mb -> 66mb.

Loading time seems relatively unaffected:
```
time python -c "import mlx.core as mx"
```

Gives about ~0.8s for both.


awni:
No.. it seems to be the same run time. And library loading time is also unaffected so win over all.

PR #2400, opened 2025-07-21
Add cuda gemv

awni:
Adds a cuda gemv kernel.

Benchmarks forthcoming.


angeloskath (approved):
Beautiful! 🤩

awni:
For square power of 2 matrices this is clearly quite superior - even in the non-overhead regime. I don't have an explanation for the drop at 4096.. it's very odd. I suppose it is related to caching at the smaller sizes.

Screenshot 2025-07-22 at 7 46 39 AM

awni:
Generation time for 7B llama is up to `Generation: 240 tokens, 102.137 tokens-per-sec` (that's rebased on #2392) so up from `98.4` of so.

awni:
Benchmark script for future reference:

```python
import mlx.core as mx
import time

for k in range(8, 15):
    n = 2 ** k
    a = mx.random.normal(shape=(n, n)).T
    b = mx.random.normal(shape=(1, n))
    mx.eval(a, b)

    def fun(b):
        for _ in range(100):
            b = b @ a
        return b

    for _ in range(10):
        mx.eval(fun(b))

    tic = time.time()
    for _ in range(100):
        mx.eval(fun(b))
    toc = time.time()

    secs = toc - tic
    gb = 100 * 100 * (a.nbytes + 2 * b.nbytes) / 1e9

    print(n, f"{gb / secs:.3f}")
```

PR #2404, opened 2025-07-22
[CUDA] Always use batched matmul

awni:
Basically it was quite bad to hit that loop in the matmul (which showed up for irregularly strided batched matmuls). This came up for models GQA for example. Switch to using a proper batched matmul in all cases which requires using kernel to set the device pointers for the batches.

```
mlx_lm.generate --model meta-llama/Llama-3.2-3B-Instruct --prompt "Write a story about Einstein" -m 240
```

Pre: `Generation: 240 tokens, 56.995 tokens-per-sec`
Post: `Generation: 240 tokens, 136.946 tokens-per-sec`

### DIFF
--- a/mlx/backend/cuda/gemms/cublas_gemm.cpp
+++ b/mlx/backend/cuda/gemms/cublas_gemm.cpp
@@ -7,14 +7,6 @@
 
 #include <fmt/format.h>
 
-void check_cublas_error(const char* name, cublasStatus_t err) {
-  if (err != CUBLAS_STATUS_SUCCESS) {
-    // TODO: Use cublasGetStatusString when it is widely available.
-    throw std::runtime_error(
-        fmt::format("{} failed with code: {}.", name, static_cast<int>(err)));
-  }
-}
-
 namespace mlx::core::cu {
 
 struct CublasPreference {
--- a/mlx/backend/cuda/gemms/cublas_gemm.h
+++ b/mlx/backend/cuda/gemms/cublas_gemm.h
@@ -7,10 +7,6 @@
 #include <cublasLt.h>
 #include <optional>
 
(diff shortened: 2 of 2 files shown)


angeloskath (approved):
Beautiful!

awni:
Changed this to fallback to the old unrolled implementation if we are on < 12.9, using cmake to pull in the right source

mlx/backend/cuda/gemms/cublas_gemm.cpp
```diff
@@ -0,0 +1,290 @@
+// Copyright © 2025 Apple Inc.
+
+#include "mlx/backend/cuda/gemms/cublas_gemm.h"
+#include "mlx/backend/cuda/device.h"
+#include "mlx/dtype_utils.h"
+#include "mlx/utils.h"
+
+#include <fmt/format.h>
+
+void check_cublas_error(const char* name, cublasStatus_t err) {
```
zcbenz: This is no longer needed, it has been moved to `mlx/backend/cuda/utils.h` by https://github.com/ml-explore/mlx/pull/2385.

mlx/backend/cuda/gemms/cublas_gemm.h
```diff
@@ -0,0 +1,104 @@
...
+
+#include "mlx/array.h"
+#include "mlx/backend/cuda/device.h"
+
+#include <cublasLt.h>
+#include <optional>
+
+void check_cublas_error(const char* name, cublasStatus_t err);
+
+#define CHECK_CUBLAS_ERROR(cmd) check_cublas_error(#cmd, (cmd))
```
zcbenz: Same here.

zcbenz (approved):
The improvement is amazing.

PR #2412, opened 2025-07-23
[Metal] Release metal events

awni:
Closes https://github.com/ml-explore/mlx-lm/issues/318

The below test now passes.. it's a bit slow for CI though. I could add it to the long running CI.

```python
import mlx.core as mx

a = mx.zeros((100_000,))
mx.eval(a)
bs = []
for i in range(len(a)):
    bs.append(a[i])
    mx.eval(bs[-1])
```


angeloskath (approved):
👍

PR #2413, opened 2025-07-24
Test on cuda 12.2 and 12.9

awni:
Run the tests on the oldest and newest versions of the cuda toolkit that we build with.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -203,8 +203,12 @@ jobs:
               python -m xmlrunner discover -v python/tests -o test-results/gpu_jit
 
   cuda_build_and_test:
+    parameters:
+      image_date:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -419,7 +423,10 @@ workflows:
             parameters:
               macosx_deployment_target: ["13.5", "14.0"]
       - linux_build_and_test
-      - cuda_build_and_test 
+      - cuda_build_and_test:
+          matrix:
+            parameters:
+              image_date: ["2023.11.1", "2025.05.1"]
```
awni, line 429: These dates correspond to images with 12.2 and 12.9.

zcbenz (approved):
Nice!

PR #2414, opened 2025-07-24
Use ccache in CI

zcbenz:
Use ccache in `cuda_build_and_test` job, which saves 20 minutes build time.

I also tried to use ccache in linux and mac build but somehow the cache miss rate is very high, and it would not really save much time, so this PR only adds cache to `cuda_build_and_test`.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -214,7 +214,7 @@ jobs:
       - checkout
       - restore_cache:
           keys:
-            - mlx-v1-cuda-{{ arch }}-
+            - cuda-<< parameters.image_date >>-{{ arch }}-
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -555,6 +575,9 @@ workflows:
           requires: [ hold ]
       - cuda_build_and_test:
           requires: [ hold ]
+          matrix:
+            parameters:
+              image_date: ["2023.11.1", "2025.05.1"]
```
awni, line 580: Good catch.

awni (approved):
Awesome! Thanks for adding that!

PR #2417, opened 2025-07-24
[CUDA] Implement Convolution VJP

zcbenz:
This is tricky to implement because for Metal/CPU the `Convolution::vjp` just reuses `conv_general` while for CUDA I have to use a different backend.

In this PR I added a `backend_type` parameter to `Convolution` which is only used by CUDA backend to choose the correct cuDNN backend. Another option would be adding a CUDA-only `ConvolutionVJP` primitive. I don't really know which solution is better but current one is easier to implement.

With backward convolution implemented it would be trivial to implement transposed convolution, which I also found tricky because of the `flip` parameter used by Metal/CPU backends. I'll add transposed convolution in another PR.


awni:
A backward conv is just a convolution with different parameters. I don't think it makes sense to add a `backend` member to the primitive. Instead, we should dispatch to the right CUDNN backend based on the other parameters of the Convolution itself. Do you think we can do that here?

zcbenz:
I think it is totally doable, I'll give it a try. (Though I need to spend some time sorting my head out how convs transform between each other)

PR #2424, opened 2025-07-26
[CUDA] Fix segfault on exit

awni:
If the program raises an error in eval then the cuda can segfault on exit because it unloads jit modules before the graph is done using them.

For example:

```bash
mlx_lm.generate --prompt "What is CUDA?" --model mlx-community/Llama-3.2-1B-Instruct-4bit -m 256
```

### DIFF
--- a/mlx/backend/cuda/device.cpp
+++ b/mlx/backend/cuda/device.cpp
@@ -56,7 +56,8 @@ Device::Device(int device) : device_(device) {
   // The cudnn handle is used by Convolution.
   CHECK_CUDNN_ERROR(cudnnCreate(&cudnn_));
 
-  // Ensure the jit module cache is initialized
+  // Initialize the jit module cache here ensures it is not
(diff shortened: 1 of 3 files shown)


angeloskath (approved):
👍

mlx/backend/cuda/device.cpp
```diff
@@ -54,6 +55,9 @@ Device::Device(int device) : device_(device) {
   CHECK_CUBLAS_ERROR(cublasLtCreate(&lt_));
   // The cudnn handle is used by Convolution.
   CHECK_CUDNN_ERROR(cudnnCreate(&cudnn_));
+
+  // Ensure the jit module cache is initialized
```
zcbenz: Maybe clarifying the root reason here like "ensure jit module is not unloaded until evaluation is done".

zcbenz (approved):
Nice catch

PR #2425, opened 2025-07-26
[CUDA] add shared mem to kernel node API

awni:
Add shared memory to the kernel node graph API. I'm (and others certainly) will be writing kernels that use shared memory.

I thought about adding this with overloads .. and can do if you all prefer. It makes the API a bit verbose.. particularly because of the templated parameter pack version. This seems a bit cleaner from an API standpoint though every call is required to specify shared memory. Lmk.. happy to change either way.


awni:
I changed my mind.. I don't think we need this yet as most use of shared mem is static. But we can revisit if needed.

PR #2426, opened 2025-07-27
[CUDA] No occupancy query for launch params

awni:
It's surprisingly expensive as @zcbenz pointed out before:

Most non-trivial kernels set the block parameters manually.. so this is mostly used by elementwise kernels with little shared memory / register pressure and the block size is always 1024.

```
mlx_lm.generate --prompt "What is CUDA?" --model mlx-community/Llama-3.2-1B-Instruct-bf16 -m 200
```

On an A100:

Pre: 216.315 tokens-per-sec
Post: 224.240 tokens-per-sec

### DIFF
--- a/mlx/backend/cuda/kernel_utils.cuh
+++ b/mlx/backend/cuda/kernel_utils.cuh
@@ -120,20 +120,6 @@ dim3 get_2d_grid_dims(
     size_t divisor);
 std::pair<dim3, dim3> get_grid_and_block(int dim0, int dim1, int dim2);
 
-// Return a block size that achieves maximum potential occupancy for kernel.
-template <typename T>
-inline uint max_occupancy_block_dim(T kernel) {
-  int _, block_dim;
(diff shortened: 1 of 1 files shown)


mlx/backend/cuda/kernel_utils.cuh
```diff
@@ -145,7 +131,7 @@ inline std::tuple<dim3, uint> get_launch_args(
     bool large,
     int work_per_thread = 1) {
   size_t nthreads = cuda::ceil_div(size, work_per_thread);
-  uint block_dim = max_occupancy_block_dim(kernel);
```
zcbenz, line 148: The `kernel` parameter can be removed from `get_launch_args` now.

zcbenz (commented):
> It's surprisingly expensive as @zcbenz pointed out before

Somehow it dropped off my radar, thanks making this PR.

angeloskath (approved):
🙏

It's also in the gemv PR.

PR #2427, opened 2025-07-27
Add more CUDA architectures for PyPi package

awni:
- Local builds use `native` to only build for the current arch
- Pypi package specifies list of architectures with real for all and virtual for the last arch for forward compat

### DIFF
--- a/mlx/backend/cuda/device.cpp
+++ b/mlx/backend/cuda/device.cpp
@@ -92,23 +92,6 @@ CommandEncoder::CaptureContext::~CaptureContext() {
   if (discard) {
     return;
   }
-
-  // Extract and add as single kernel node when possible.
-  size_t num_nodes;
-  CHECK_CUDA_ERROR(cudaGraphGetNodes(graph, NULL, &num_nodes));
-  if (num_nodes == 1) {
-    cudaGraphNode_t captured_node;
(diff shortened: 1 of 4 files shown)


mlx/backend/cuda/device.cpp
```diff
@@ -92,23 +92,6 @@ CommandEncoder::CaptureContext::~CaptureContext() {
...
-    CHECK_CUDA_ERROR(cudaGraphGetNodes(graph, &captured_node, &num_nodes));
-    cudaGraphNodeType type;
-    CHECK_CUDA_ERROR(cudaGraphNodeGetType(captured_node, &type));
-    if (type == cudaGraphNodeTypeKernel) {
-      CUDA_KERNEL_NODE_PARAMS params;
-      CHECK_CUDA_ERROR(cuGraphKernelNodeGetParams(captured_node, &params));
-      enc.add_kernel_node(params);
-      return;
-    }
-  }
```
awni, line 110: Removing this optimization since it breaks on H100 .. note this will make it even more important to avoid stream capture for low latency use cases.

awni:
This increases the uncompressed binary from 74M to 78M. So not too bad.

angeloskath (approved):
👍

PR #2429, opened 2025-07-28
[CUDA] More sizes for gemv

awni:
Adds specializations for blocks of 32 and 64 which covers more common sizes.

Nice speed up in generation for LLama 3.1 1B since we don't hit the cublas kernel for part of the SDPA:

```bash
mlx_lm.generate --prompt "What is CUDA?" --model mlx-community/Llama-3.2-1B-Instruct-bf16 -m 200
```
Benchmarks on top of #2426 on A100:

Pre: 224.240 tokens-per-sec 
Post: 241.482 tokens-per-sec


angeloskath (approved):
Nice!

PR #2431, opened 2025-07-28
[CUDA] Backward convolution

zcbenz:
This PR implements backward/transposed convolution by automatically choosing the proper cuDNN convolution backend based on the settings. Convolution with groups are currently not handled properly, which I will solve in following PRs.

The heuristics used for choosing the backend looks fragile, which was written with reference to [JAX's implementation](https://github.com/openxla/xla/blob/b44722f702038714c9d888358df048c0890aa60e/xla/service/gpu/transforms/conv_rewriter.cc#L776-L812). I think a more robust way is to fallback to forward convolution when a backward backend throws error, which I will try in another PR as it requires some tricky refactorings.


mlx/backend/cuda/conv.cpp
```diff
@@ -263,78 +361,104 @@ void Convolution::eval_gpu(const std::vector<array>& inputs, array& out) {
 
   encoder.set_input_array(in);
   encoder.set_input_array(wt);
-  encoder.set_output_array(out);
-
-  auto backend_type = CUDNN_BACKEND_OPERATION_CONVOLUTION_FORWARD_DESCRIPTOR;
-  auto cudnn_type = dtype_to_cudnn_type(in.dtype());
+  encoder.set_output_array(out_);
```
zcbenz: Should `out` or `out_` be passed here? 🤔 I think the final output `out_` should be passed but want to be sure.

awni: They have the same data so it actually doesn't matter in this case. But in general the rule is set the array (or arrays) whose data you write to in the subsequent kernel.

zcbenz:
This test is a bit flaky in CI:

```
FAIL: test_conv2d (test_nn.TestLayers.test_conv2d)
  File "/home/circleci/project/python/tests/test_nn.py", line 758, in test_conv2d
    self.assertTrue(np.allclose(y[:, 0, 0, 0], x.mean(axis=(1, 2, 3))))
AssertionError: False is not true

```

The flaky test uses the `CONV_BACKWARD_WEIGHT` backend, before this PR it was using `CONV_FORWARD`, both backends should deliver the same result but the test being flaky indicates problems with the `CONV_BACKWARD_WEIGHT` code path, could it be the trick used in `out` and `out_`?

zcbenz:
I made a change to the conv backend choosing code that, instead of using heuristics to choose a backend and then execute it, a guess would be made first and different backends will then be tried until there is a working one.

So when a wrong guess is made we will still get the correct backend in the end, at the cost of some redundant copies, and with caching it only affects the first run.

Also in cases that a conv setting may run in multiple backends, we can confidently prefer the backends that involve less copies without worrying about a wrong guess, i.e. forward convs will no longer accidentally use `CONV_BACKWARD_WEIGHT` backend.

awni:
> could it be the trick used in out and out_?

I don't see any issues with that.. but it's hard to say. One thing you can do to double check is [export the cuda graph to dot](https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__GRAPH.html#group__CUDART__GRAPH_1gbec177c250000405c570dc8c4bde20db) format to make sure all the dependencies look right. If there is an kernel is supposed to be after the conv output is done you should see it in the dot graph.

Basically before you commit the graph run something like:

```
cudaGraphDebugDotPrint(graph, "graph.dot", 0);
```

Then you can make it viewable with:

```
dot -Tpdf graph.dot -o graph.pdf
```

And `open graph.pdf`.

awni (approved):
Looks great! What did you end up doing with the flaky test?

zcbenz:
> One thing you can do to double check is [export the cuda graph to dot](https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__GRAPH.html#group__CUDART__GRAPH_1gbec177c250000405c570dc8c4bde20db) format to make sure all the dependencies look right

This is really useful tip, thanks!

> What did you end up doing with the flaky test?

The flaky test only happened once even after a few reruns, also with last change the special condition (using `CONV_BACKWARD_WEIGHT` for forward conv) is gone so I'm letting it go.

PR #2435, opened 2025-07-29
Cuda faster softmax

awni:
Training transformer is faster:

Pre (0.27.1): 8.599
Post: It/sec 8.953


angeloskath (approved):
🚀

Massive difference!

PR #2438, opened 2025-07-30
Move arange to its own file

zcbenz:
A few benefits:

* The implementation of arange gets self contained.
* One less file under `backend/cuda/device/`.
* The `primitives.cu` can be converted to `.cpp` file.


awni (approved):
LGTM!

PR #2440, opened 2025-07-30
Make CI faster

zcbenz:
Reduce the time spent on installing dependencies by:

* removing redundant `apt-install`s
* using uv to install python packages

Reduced time:

* `cuda_build_and_test`:
  * Install python package: 3m14s => 1m12s
* `linux_build_and_test`:
  * Install dependencies: 3m9s => 25s
  * Install python package: 4m17s => 3m26s
* `mac_build_and_test`:
  * Install dependencies: 1m19s => 22s


angeloskath (approved):
That's neat! I also like that you enabled the verbose build no idea why we didn't use that before ...

PR #2441, opened 2025-07-30
[CUDA] Matmul utils initial commit

angeloskath:
This is very far from complete but for now it will a) allow us to collaborate on building the utils b) iterate on a matmul instead of the qmm that I was using so far.

The included `A @ B.T` kernel is doing about 140 TFLOPS on the A100 which is worse than I had measured compared to CUBLAS. I have pushed a branch `simple-gemm` for anyone who wants to play with this kernel. I don't think we 'll ever route to that kernel so maybe I should move it to the `simple-gemm` branch to experiment on instead of main 🤷‍♂️

This also adds explicit addition of shared memory to the command encoder cause dynamic shared memory is necessary for using more than 48KB of shared memory which is practically every matmul kernel ever.

Tentative TODOs:
- Integrate CuTe layouts to get rid of some int math and swizzle math
- Add streaming `cp_async` variants for matrix vector products
- Add fallbacks (this should probably be done last after we have an optimized matmul)
- Add TMA and Hopper matmuls (also after we have an optimized matmul)


angeloskath:
Yeah I think you 're right it completely slipped my mind since I was expecting it to break in CI but it didn't. When time comes the specific PTX will have to be guarded by architecture unfortunately.

jagrit06 (approved):
Will you just add architecture guards inside the function and then they do nothing if it can’t be compiled correctly ? 
Other than that, looks good to me for now!

PR #2443, opened 2025-07-30
fix circular reference

awni:
Fixes an issue with circular references causing us to hold data longer than necessary when updating a module's submodules.


angeloskath (approved):
🙏

PR #2444, opened 2025-07-30
[CUDA] Vectorize generated kernels

angeloskath:
This vectorizes reads for the contiguous case and writes for all cases. Some numbers below:

- Compiled `x**2 + 3 * y`: **1700 GB/s** from 1340 GB/s
- Transformer train: **9.6 it/s** from 8.9 it/s

I also saw no perf regression on Qwen generation but that probably needs checking after we re-fix the gemv.


mlx/backend/cuda/compiled.cpp
```diff
@@ -205,28 +249,19 @@ void Compiled::eval_gpu(
...
       kernel_names.push_back(fmt::format(
-          "mlx::core::cu::{}_contiguous<uint32_t, {}>",
-          lib_name(),
-          work_per_thread));
+          "mlx::core::cu::{}_contiguous<uint32_t, {}>", lib_name(), wpt));
       kernel_names.push_back(fmt::format(
-          "mlx::core::cu::{}_contiguous<int64_t, {}>",
-          lib_name(),
-          work_per_thread));
+          "mlx::core::cu::{}_contiguous<int64_t, {}>", lib_name(), wpt));
```
awni: I don't think we need the `work_per_thread = 1` for the contiguous kernels? Like you could move the start of this loop down a couple lines and only build the contiguous kernels for `work_per_thread`?

awni (approved):
Very very nice!!

angeloskath:
Briefly tried the `unsafe_store/load` stuff it didn't make a difference.

The compiler is pretty ok at moving the checks outside and reusing them across the stores and loads. Possibly some room for improvement by moving them outside the kernel completely but I think this is ok to merge after the tests pass.

PR #2446, opened 2025-07-30
Fix custom metal extension

awni:
Previously (a long time ago) we relied on some implicit behavior of keeping the library loading in a header so it would resolve the right path for the binary. Rather than rely on that which was evidently quite brittle, this makes it explicit in the extension code.


.circleci/config.yml
```diff
@@ -167,6 +167,7 @@ jobs:
             cd examples/extensions
             uv pip install -r requirements.txt
             uv run --no-project setup.py build_ext -j8
+            python test.py
```
angeloskath: Probably `uv run --no-project test.py`

PR #2448, opened 2025-07-31
Use LRU cache for cuda graph

zcbenz:
The cache policy does not affect performance in most cases as rare programs exceeds graph cache size, but I think using a LRU cache would help in many corner cases.

### DIFF
--- a/mlx/backend/cuda/device.cpp
+++ b/mlx/backend/cuda/device.cpp
@@ -189,8 +189,6 @@ CommandEncoder::CommandEncoder(Device& d)
   CHECK_CUDA_ERROR(cudaGraphCreate(&graph_, 0));
 }
 
-CommandEncoder::~CommandEncoder() = default;
-
 void CommandEncoder::add_completed_handler(std::function<void()> task) {
   worker_.add_task(std::move(task));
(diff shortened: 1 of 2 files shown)


awni:
It would be good to check the inference benchmark is unaffected here. Did you try running it?

zcbenz:
I didn't observe any change in inference speed with this PR, the overhead added by LRU cache is moving a node in linked list when a key is requested, which should be nothing.

mlx/backend/cuda/device.cpp
```diff
@@ -184,20 +184,12 @@ void CommandEncoder::insert_graph_dependencies(std::vector<GraphNode> nodes) {
...
-  for (auto& [_, graph_exec] : graphs) {
-    CHECK_CUDA_ERROR(cudaGraphExecDestroy(graph_exec));
-  }
-  graphs.clear();
-}
-
-CommandEncoder::~CommandEncoder() {
-  clear_graphs(graph_cache_);
-}
+CommandEncoder::~CommandEncoder() = default;
```
awni: Why not just remove that from the header (and from here)?

PR #2449, opened 2025-07-31
[CUDA] Save primitive inputs faster

zcbenz:
Avoid constructing expensive std objects in `gpu::eval`, which directly slows down kernel launching.

For command `mlx_lm.generate --model meta-llama/Llama-3.2-1B-Instruct --prompt 'Write a story about Einstein' -m 240`, generation speed increases from 230 t/s to 260.

### DIFF
--- a/mlx/backend/cuda/eval.cpp
+++ b/mlx/backend/cuda/eval.cpp
@@ -43,9 +43,7 @@ void eval(array& arr) {
     }
   }
   for (auto& s : arr.siblings()) {
-    if (s.data_shared_ptr() != arr.data_shared_ptr()) {
-      encoder.add_temporary(s);
-    }
+    encoder.add_temporary(s);
   }
   encoder.maybe_commit();
 }


mlx/backend/cuda/eval.cpp
```diff
@@ -36,18 +36,17 @@ void eval(array& arr) {
...
+      encoder.add_temporary(in);
+    }
   }
   for (auto& s : arr.siblings()) {
-    buffers.insert(s.data_shared_ptr());
-  }
-  // Remove the output if it was donated to by an input.
-  if (auto it = buffers.find(arr.data_shared_ptr()); it != buffers.end()) {
-    buffers.erase(it);
+    if (s.data_shared_ptr() != arr.data_shared_ptr()) {
```
awni: I don't think this check is necessary.. the array and the sibling don't have the same data in any case I know of ..

awni:
Wow I'm floored that that makes such a big difference!

mlx/backend/cuda/quantized/quantized.cpp
```diff
@@ -42,6 +44,7 @@ inline array ensure_row_contiguous_matrix(
 void fast::AffineQuantize::eval_gpu(
     const std::vector<array>& inputs,
     std::vector<array>& outputs) {
+  nvtx3::scoped_range r("AffineQuantize::eval_gpu");
```
awni, line 47: 🙏

awni (approved):
Awesome!!

PR #2454, opened 2025-08-02
Use SmallVector for shapes and strides

zcbenz:
The SmallVector class is modified from [V8's implementation](https://source.chromium.org/chromium/chromium/src/+/main:v8/src/base/small-vector.h), which is battle tested and concise enough to be fully understood.

The result is not super impressing — for command `mlx_lm.generate --model meta-llama/Llama-3.2-1B-Instruct --prompt 'Write a story about Einstein' -m 240` the tokens per second goes from 264 to 278. But I think we will keep seeing performance improve when gradually replacing more code using `std::vector`.

This PR also fixes some out-of-bounds reads caught by the boundary checks of SmallVector.

### DIFF
--- a/mlx/backend/cpu/sort.cpp
+++ b/mlx/backend/cpu/sort.cpp
@@ -8,7 +8,7 @@
 #include "mlx/backend/common/utils.h"
 #include "mlx/backend/cpu/copy.h"
 #include "mlx/backend/cpu/encoder.h"
-
+#include "mlx/dtype_utils.h"
 #include "mlx/primitives.h"
 
 namespace mlx::core {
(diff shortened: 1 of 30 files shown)


mlx/backend/cpu/sort.cpp
```diff
@@ -333,47 +333,24 @@ void Sort::eval_cpu(const std::vector<array>& inputs, array& out) {
...
   auto& in = inputs[0];
 
+  int axis = axis_;
+  if (axis < 0) {
+    axis += in.ndim();
+  }
+
   // Copy input to output
-  CopyType ctype = (in.flags().contiguous && in.strides()[axis_] != 0)
+  CopyType ctype = (in.flags().contiguous && in.strides()[axis] != 0)
```
awni, line 342: yikes.. that's a good catch from the small vector!

python/src/small_vector.h
```diff
@@ -0,0 +1,17 @@
...
+#include "mlx/small_vector.h"
+
+#include <nanobind/stl/detail/nb_list.h>
+
+NAMESPACE_BEGIN(NB_NAMESPACE)
+NAMESPACE_BEGIN(detail)
+
+template <typename Type, size_t Size, typename Alloc>
+struct type_caster<mlx::core::SmallVector<Type, Size, Alloc>>
+    : list_caster<mlx::core::SmallVector<Type, Size, Alloc>, Type> {};
```
awni: Can we default cast it to a tuple instead of a list? I feel like that's what we want it to be in most cases (and we could avoid a copy in e.g. `array.shape`.

awni:
> the tokens per second goes from 264 to 278.

That's pretty awesome that it helps, that's not too bad by any means!

python/src/small_vector.h
zcbenz: Yeah it is possible, though we need to define our own caster, nanobind's tuple caster can not be reused for vector types. Saving a copy for `array.shape` is nice.

PR #2460, opened 2025-08-04
Add missing algorithm header to jit_compiler.cpp for Linux builds

### LINKED ISSUE #2459: Add `#include <algorithm>` to `jit_compiler.cpp` for Linux builds
Can we add #include <algorithm> to [`/backend/cpu/jit_compiler.cpp`](https://github.com/ml-explore/mlx/blob/main/mlx/backend/cpu/jit_compiler.cpp#L121) because the file uses std::find_if on line 121 but wasn't including the header that defines it. https://github.com/ml-explore/mlx/blob/aaf78f4c6b618daa0fd74697a47cc467920e958f/mlx/backend/cpu/jit_compiler.cpp#L121

The specific error was:
```
mlx/backend/cpu/jit_compiler.cpp:121:12: error: no member named 'find_if' in namespace 'std'
```

Looking at the code, `std::find_if` is used in the compilation process. While this might work on some platforms where <algorithm> is transitively included through other headers, it's not guaranteed by the C++ standard. On Linux with GCC/Clang, the transitive inclusion doesn't happen, causing the build to fail. (it'd be great

zamderax:
## Summary

This PR fixes issue #2459 by adding the missing algorithm header to mlx/backend/cpu/jit_compiler.cpp.

## Problem

The file uses std::find_if on line 121 but wasn't including the algorithm header that defines it. While this might work on some platforms where algorithm is transitively included through other headers, it's not guaranteed by the C++ standard. On Linux with GCC/Clang, the transitive inclusion doesn't happen, causing build failures.

## Solution

Added include algorithm to the includes section of jit_compiler.cpp to explicitly include the header that defines std::find_if.

## Testing

- Built successfully on macOS with Clang  
- All existing tests pass (225 test cases, 3290 assertions)
- The change is minimal and safe - only adds a required header

## Related Issues

Fixes #2459

### DIFF
--- a/mlx/backend/cpu/jit_compiler.cpp
+++ b/mlx/backend/cpu/jit_compiler.cpp
@@ -2,6 +2,7 @@
 
 #include "mlx/backend/cpu/jit_compiler.h"
 
+#include <algorithm>
 #include <sstream>
 #include <vector>
 


awni (approved):
Looks great, thanks!

PR #2463, opened 2025-08-05
Faster gather qmm sorted test

awni:
Reduces the sizes of the test cases so the test takes 16 seconds on the CPU instead of 7 minutes.


angeloskath (approved):
Thanks! That makes perfect sense :-)

PR #2465, opened 2025-08-05
revert default cuda install

awni:
The solution doesn't in fact work :( so we're back to `pip install mlx[cuda]`.


angeloskath (approved):
👍

PR #2471, opened 2025-08-06
Fix typo in metal command encoder

### LINKED ISSUE #2467: [BUG] CommandEncoder::outputs() returns copy instead of reference
Looks like `outputs()` should return a reference (similar to `inputs()` above it) rather than a copy:

https://github.com/ml-explore/mlx/blob/db5c7efcf60770995f3083629ab158becb3c4b2c/mlx/backend/metal/device.h#L107-L109

given its usages, like:

https://github.com/ml-explore/mlx/blob/db5c7efcf60770995f3083629ab158becb3c4b2c/mlx/backend/metal/device.cpp#L419

^ erases from a transient copy

https://github.com/ml-explore/mlx/blob/db5c7efcf60770995f3083629ab158becb3c4b2c/mlx/backend/metal/device.cpp#L447
^ move on a copy

angeloskath:
Closes #2467 .

Thanks @ethereon for pointing it out!

The encoder is made anew when we commit it so we aren't really leaking anything but the intermediate optimization of removing temps and moving the outputs into the handler is indeed moot given it returns a copy.

### DIFF
--- a/mlx/backend/metal/device.h
+++ b/mlx/backend/metal/device.h
@@ -104,7 +104,7 @@ struct CommandEncoder {
   };
 
   // Outputs of all kernels in the encoder including temporaries
-  std::unordered_set<const void*> outputs() {
+  std::unordered_set<const void*>& outputs() {
     return all_outputs_;
   };
 


awni (approved):
🙏

PR #2472, opened 2025-08-07
Faster general unary op

awni:
- Use vectorized store and vectorized load with stride for:
  - unary
  - binary
  - ternary
  - copy
- Adds an extra specialization for work per thread = 1 or 4

Benchmarks on an H100:

Bench | GB/s Pre | GB/s Post
--- |---- | ----
strided unary | 1425 | 2021
transposed binary | 1341 | 2380
transposed ternary | 1469 | 2276
transposed copy | 943 | 1600

Bench | It/s Pre | It/s Post
--- |---- | ----
transformer training | 18.7 | 21.3 

No detectable change for Lama 3.2 1B Generation

### DIFF
--- a/mlx/backend/cuda/copy/copy_general.cu
+++ b/mlx/backend/cuda/copy/copy_general.cu
@@ -10,37 +10,80 @@ namespace cu {
 
 namespace cg = cooperative_groups;
 
-template <typename In, typename Out, typename IdxT, int NDIM>
+template <typename In, typename Out, typename IdxT, int NDIM, int N_READS>
 __global__ void copy_gg_nd(
     const In* in,
     Out* out,
-    IdxT size,
+    IdxT size_rest,
     const __grid_constant__ cuda::std::array<int32_t, NDIM> shape,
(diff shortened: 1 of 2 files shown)


awni:
I split binary.cu and unary.cu into a bunch of little files (one for each primitive) which speeds up compilation time a lot.. it's not the most elegant.. but relatively clean.

awni:
In terms of binary size we go from 52M to 57M for the local build.

mlx/backend/cuda/unary/sqrt.cu
```diff
@@ -0,0 +1,15 @@
+// Copyright © 2025 Apple Inc.
+
+#include "mlx/backend/cuda/unary/unary.cuh"
+
+namespace mlx::core {
+void Sqrt::eval_gpu(const std::vector<array>& inputs, array& out) {
+  nvtx3::scoped_range r("Sort::eval_gpu");
```
angeloskath: Typo on the nvtx range

angeloskath (approved):
Fantastic speed up, very clean one typo!

PR #2473, opened 2025-08-07
[CUDA] Optimize set_mm_device_pointers for small ndim

zcbenz:
Use the same optimization with element-wise kernels that simplifies computations for small dimensions.

I also tried a few other possible optimizations like doing vectorized writes and writing to `pointers` in separate threads, but only got slower.

With this change the time spent on `set_mm_device_pointers` reduced from 3.4% to 2.6% during generation. It is still not ideal but I'm out of ideas.


awni:
Looks great! Btw with the fused attention we won't be using this op nearly as much for generation so I think it's ok to not spend as much time on it.

awni (approved):

PR #2474, opened 2025-08-07
Fix logsumexp/softmax not fused for some cases

zcbenz:
For an array with a shape like `(1, 1234)`, make sure `logsumexp(a)` uses the fused `LogSumExp` primitive.

This could happen when the users forget to pass `axis=-1` like [in mlx-lm](https://github.com/ml-explore/mlx-lm/blob/f1a6359c92827f0ba2185f5264b9ddb313f7360e/mlx_lm/generate.py#L409).

### DIFF
--- a/mlx/ops.cpp
+++ b/mlx/ops.cpp
@@ -2381,9 +2381,20 @@ array logsumexp(
     throw std::invalid_argument(
         "[logsumexp] Received non-empty axes for array with 0 dimensions.");
   }
+  bool reduce_last_dim =
+      !axes.empty() && (axes.back() == a.ndim() - 1 || axes.back() == -1);
+  if (reduce_last_dim) {
+    // For more than 2 axes check if axes is [0, 1, ..., NDIM - 1] and shape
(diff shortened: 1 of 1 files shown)


mlx/ops.cpp
```diff
@@ -2381,9 +2381,18 @@ array logsumexp(
...
+  bool reduce_last_dim =
+      !axes.empty() && (axes.back() == a.ndim() - 1 || axes.back() == -1);
+  for (int i = axes.size() - 2; i >= 0; --i) {
+    // For more than 2 axes check if axes is [0, 1, ..., NDIM - 1] and shape
+    // is [1, 1, ..., N].
+    if ((axes[i] + 1 != axes[i + 1]) || (a.shape(axes[i]) != 1)) {
+      reduce_last_dim = false;
+      break;
+    }
+  }
```
awni: Great! Would you mind updating `softmax` as well to use this logic? I think it is similarly flawed.

zcbenz:
Updated the PR to also fix softmax.

PR #2476, opened 2025-08-07
NCCL backend

nastya236:
This PR contains:

- NCCL backend implementation: initialisation of the distributed group and all_sum reduction
- Minor change to the `mlx.utils.average_gradients` function to have `stream` as an input
- `launch_nccl` in `distributed_run.py` to launch execution with nccl backend
- Tests `mlx/python/tests/nccl_test_distributed.py`

Implementation details:

- host process is responsible for at most one GPU 
- process with rank 0 is always a master host

Notes:

- Other reductions (all_min, all_max) and communication primitives (all_gather, recv, send) are in progress
- For now `mlx.launch` is working only locally so only on 1 node, support  for multiple nodes and non local hosts is in progress
- `TestNCCLDistributed` is not inherited from `MLXDistributedCommonTestCase` for now. 

To test the implementation:

- install [headers and cuda toolkit](https://ml-explore.github.io/mlx/build/html/install.html#id3)
- `git clone https://github.com/nastya236/mlx.git && cd mlx && git checkout nccl_backend && cd mlx`
- `CMAKE_ARGS="-DMLX_BUILD_CUDA=ON" python -m pip install -e ".[dev]"`
- run test: `mlx.launch --nproc-per-node n --backend nccl python/tests/nccl_test_distributed.py`, where `n` -- number of process to run on the host  (= number of GPUs per host).


mlx/distributed/ring/ring.cpp
```diff
@@ -975,7 +975,6 @@ class RingGroup : public GroupImpl {
 
   int rank_;
   int size_;
-
```
angeloskath: nit: probably leave as it were 🤷‍♂️

mlx/distributed/nccl/nccl.cpp
```diff
@@ -0,0 +1,359 @@
...
+  }
+}
+
+template <typename T>
+struct type_identity {
+  using type = T;
+};
+
+template <typename F>
+void dispatch_dtype(const array& arr, F&& f) {
```
angeloskath: You can use the ones defined in `mlx/dtype_utils.h`.

python/mlx/distributed_run.py
```diff
@@ -415,6 +415,45 @@ def launch_mpi(parser, hosts, args, command):
             pass
 
 
+def launch_nccl(parser, hosts, args, command):
+    master_host = hosts[0].ips[0]
```
angeloskath, line 419: Nit: Maybe assert that host is localhost for now?

python/mlx/distributed_run.py
```diff
@@ -769,6 +808,19 @@ def main():
...
         "--cwd", help="Set the working directory on each node to the provided one"
     )
+    parser.add_argument(
+        "--nccl-port",
+        type=int,
+        default=12345,
+        help="The port to use for the NCCL communication (only for nccl backend)",
+    )
+    parser.add_argument(
+        "--nproc-per-node",
```
angeloskath: That's `-n` or `--repeat-hosts`. Basically NCCL for now should support only `mlx.launch --backend nccl -n 8 train.py` the default host is also 127.0.0.1 which should be fine.

python/mlx/distributed_run.py
```diff
@@ -799,8 +851,10 @@ def main():
     # Launch
     if args.backend == "ring":
         launch_ring(parser, hosts, args, rest)
-    elif args.backend == "mpi":
```
angeloskath, line 802: The else is pretty important I think otherwise if we start by ring and finish normally then we 'll start by mpi and finally by nccl.

mlx/distributed/ops.cpp
```diff
@@ -31,8 +31,7 @@ array all_sum(
   return array(
       x.shape(),
       x.dtype(),
-      std::make_shared<AllReduce>(
-          to_stream(s, Device::cpu), group, AllReduce::Sum),
+      std::make_shared<AllReduce>(to_stream(s), group, AllReduce::Sum),
```
angeloskath: That's cool but it breaks the metal backend if the communication is not put explicitly on the cpu. I think it is probably ok but we may need to gather some opinions as it will be a breaking change and also change it to all ops.

@awni what do you think? I am in favor of making it explicit. Anyone using `nn.average_gradients` would be explicitly using the cpu anyway.

angeloskath (commented):
This looks awesome. I left a bunch of comments mostly nitpicks.

The only one to consider is the change of behavior for the ops that will now put things on the gpu stream. I think we should just make it explicit in all ops and let it fail if used with the gpu stream in Metal. It is also kind of nice to make explicit the fact that there will be a CPU/GPU sync when communicating.

mlx/distributed/ops.cpp
awni: One option would be to have a `get_stream` helper in this file which checks if we are on `metal` or `cuda` and picks the right default device (assuming no stream is passed).

- I like having the property that the front-end code needs minimal changes to switch between Metal and CUDA.
- I think I prefer when your on the Metal backend the default `mx.all_sum(x)` should work. Rather than requiring a CPU stream to be specified, even though that is more explicit.

angeloskath: Nice! I like that solution.

awni:
What's left on the todo list before we can get this merged? Looks like tests were still failing?

mlx/distributed/nccl/nccl.cpp
nastya236: I reuse `dispatch_all_types` from `mlx/dtype_utils.py`,  let me know if you don't like it / think that it can be done better:)

python/mlx/distributed_run.py
nastya236: Yes, totally agree, thank you for the suggestion! no need to introduce another argument

python/mlx/distributed_run.py
nastya236: If I understand correctly, `args.backend` will always be only one of `[ring, mpi, nccl]`. In that case, only one of the branches will be executed even without the elif statements, right? I might be missing something
probably `elif` looks cleaner, but I just want to check that I don't miss anything:)

mlx/distributed/ops.cpp
nastya236: Oh right, I dropped it to test that nccl is working and forgot about it.. Now it is so clear why `all_min` was working and `all_max` behaved weirdly.. Helper function sounds like a great option! I wrote a `get_device()` helper that returns the device to use for the reduction. Let me know if you had something different in mind!

python/mlx/distributed_run.py
angeloskath: Yeah it's true I don't know what I was thinking :-)

mlx/distributed/ops.cpp
angeloskath: So I changed this to have the group decide on the stream. The reason is that on CUDA we could still decide to run the MPI backend which should run on the CPU. Similarly for the ring backend. So the NCCL backend will always run on the GPU which also makes more sense. It also enables backends that can do both CPU and GPU etc.

nastya236:
@angeloskath Thank you for fixing the communication stream!!

PR #2477, opened 2025-08-08
Fix incorrect interpretation of unsigned dtypes in reduce ops

### LINKED ISSUE #2475: [BUG] Strange results when calculating the mean and std of integer array
**Describe the bug**
Strange results when calculating the mean and std of integer array.

**To Reproduce**

Include code snippet
```python
a = mx.array([0, 100, 200, 123], dtype=mx.uint8)
print(mx.mean(a)) # array(41.75, dtype=float32)
print(mx.mean(a.astype(mx.float32))) # array(105.75, dtype=float32)

print(mx.std(a)) # array(95.8915, dtype=float32)
print(mx.std(a.astype(mx.float32))) # array(71.4086, dtype=float32)
```

**Expected behavior**
Should be the same value with or without .astype()

 - OS Version: MacOS 15.6
 - Version 0.27.1

**Additional context**
None

abeleinin:
This PR resolves #2475 by updating how unsigned dtypes are handled in the `Reduce::Sum` and `Reduce::Prod` primitives.

For CPU, I changed the dispatch type of `reduce_dispatch_sum_prod<T>` to pass in the input type. When we load the input data in the `contiguous_reduce` function on this line:
https://github.com/ml-explore/mlx/blob/aa7b47481a9188407dd99eaaa456c2b87c621ff0/mlx/backend/cpu/reduce.cpp#L103 
with `simd::load<T, N>`, it needs to read in the input as it's original type (ie `uint8_t`, not `int8_t`) before upcasting to the output type `U`.

For GPU, a similar issue happens during the `static_cast<U>` operation. In `reduce.metal` `U` is defined as `int8_t`, which will interpret unsigned values incorrectly.
https://github.com/ml-explore/mlx/blob/aa7b47481a9188407dd99eaaa456c2b87c621ff0/mlx/backend/metal/kernels/reduction/reduce_all.h#L40 https://github.com/ml-explore/mlx/blob/aa7b47481a9188407dd99eaaa456c2b87c621ff0/mlx/backend/metal/kernels/reduce.metal#L137

### DIFF
--- a/mlx/backend/cpu/reduce.cpp
+++ b/mlx/backend/cpu/reduce.cpp
@@ -491,19 +491,27 @@ void Reduce::eval_cpu(const std::vector<array>& inputs, array& out) {
         switch (in.dtype()) {
           case bool_:
           case uint8:
+            reduce_dispatch_sum_prod<uint8_t>(in, out, reduce_type_, axes_);
+            break;
+          case uint16:
+            reduce_dispatch_sum_prod<uint16_t>(in, out, reduce_type_, axes_);
+            break;
+          case uint32:
+            reduce_dispatch_sum_prod<uint32_t>(in, out, reduce_type_, axes_);
+            break;
+          case uint64:
+            reduce_dispatch_sum_prod<uint64_t>(in, out, reduce_type_, axes_);
+            break;
           case int8:
             reduce_dispatch_sum_prod<int8_t>(in, out, reduce_type_, axes_);
             break;
           case int16:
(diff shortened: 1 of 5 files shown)


angeloskath (approved):
Thanks for the fix!

Initially we were accumulating in the same sized output type in which case the remap made sense and reduced the number of kernels to be compiled. Now it obviously results in an error for unsigned types.

I will merge after the tests pass.

PR #2488, opened 2025-08-12
Rename cu::Matmul to CublasMatmul

zcbenz:
This PR does some refactoring to make it easier adding another matmul backend:

* Rename `cu::Matmul` to `CublasMatmul` to make avoid the confusions between `cu::Matmul` and `Matmul`.
* Rename the file `gemms/cublas_gemm.cpp` to `matmul/cublas_matmul.cpp` so it is easier to find the right file when someone wants to find out the implementation of `CublasMatmul` in `matmul.cpp`.
* Let `CublasMatmul::run` decide whether to use batched mode instead of making the caller decide.


angeloskath:
On that note maybe it makes sense to actually name it `CublasGemm`. It is a gemm anyway.

zcbenz:
`CublasGemm` sounds a good name, I have updated the PR.

angeloskath (approved):
Looks great! I like that you put the batching logic in the `CublasGemm`.

It makes way more sense and it should also be refactored further in the future since the normal run is also batched. Basically `run_batched` only runs non-contiguous batches hence the weird heuristic of `batch_count / batch_shape.back() > 1` .

Merge at will :-)

PR #2489, opened 2025-08-12
Use CMake <4.1 to avoid the nvpl error

angeloskath:
For now use CMake versions <4.1 to avoid the NVPL error so that CI works again.


zcbenz (approved):
👍

PR #2491, opened 2025-08-13
Split cuDNN helpers into a separate header

zcbenz:
I was testing a cuDNN based rms norm implementation, it was 17% faster than our custom kernel, but the overhead of capturing the cuDNN graph kills the performance for inference, and the norm op does not support the native cuda graph APIs of cuDNN so I could not get rid of the overhead.

But the good thing is I have made our cuDNN helpers much more general so it would be much easier adding new things with cuDNN, so I'm opening this PR with all the changes on cuDNN helpers without the rms norm implementation. There is also a new `CudaGraph` helper added for easier caching of cuda graphs.


angeloskath (approved):
This looks great to me. Sorry it took so long to review the RAII PR. I liked that a lot.

I will let @jagrit06 chime in as he 's interacted way more with CUDNN.

awni:
@zcbenz let's merge this?

jagrit06 (approved):
LGTM!

PR #2495, opened 2025-08-14
[CUDA] Fix conv grads with groups

zcbenz:
This PR fixes vjp of grouped convs by correctly transpose the inputs with groups, it introduced a few new utility functions which I put in `gpu/copy.h`.


awni (approved):
Looks good, thanks for fixing that!!

PR #2499, opened 2025-08-16
Add mode parameter for quantization

awni:
Adds a mode parameter for quantization.

Even though there is only one mode right now (`affine`) it might be good to review / merge this in the interest of not making massive diffs.

Either way work is in progress on `mxfp4` as the next supported quant.

### DIFF
--- a/mlx/backend/cpu/quantized.cpp
+++ b/mlx/backend/cpu/quantized.cpp
@@ -862,7 +862,7 @@ void QuantizedMatmul::eval_cpu(const std::vector<array>& inputs, array& out) {
   encoder.set_input_array(w);
   encoder.set_input_array(scales);
   encoder.set_output_array(out);
-  if (mode_ == "affine") {
+  if (mode_ == QuantizationMode::Affine) {
     auto biases = ensure_row_contiguous(inputs[3]);
     encoder.set_input_array(biases);
     encoder.dispatch([out = array::unsafe_weak_copy(out),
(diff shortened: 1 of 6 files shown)


awni:
@angeloskath I'm moving this out of draft. I'd like to do a bit of refactoring to avoid copying some shared helper functions. I'm on the fence if it makes sense to refactor rmore beyond that. 

I could see an argument to refactor the dequantizer to be independent of the kernels themselves. It would dramatically reduce code-duplication but I'm also not sure it's worth it yet.

mlx/backend/metal/kernels/fp4_quantized.h
```diff
@@ -0,0 +1,1791 @@
...
+}
+
+template <typename T, typename U, int values_per_thread>
+inline void load_vector_safe(const device T* x, thread U* x_thread, int N) {
+  for (int i = 0; i < N; i += 4) {
+    x_thread[i] = x[i];
+    x_thread[i + 1] = x[i + 1];
+    x_thread[i + 2] = x[i + 2];
+    x_thread[i + 3] = x[i + 3];
+  }
```
angeloskath, line 55: This reads OOB doesn't it? It's safe cause you overwrite with 0s but it does read up to 3 values oob, just saying.

angeloskath (approved):
This is awesome! So much work!

I think it isn't worth generalizing things too much. Simple and copied wins over complex and shared for me.

One thing that I should probably have commented on sooner is the choice of propagating the string everywhere. Supposing a small strings optimization in the standard library then it isn't a big deal (it's <8 bytes anyway) otherwise it might be incurring a lot of unnecessary heap allocations.

mlx/backend/metal/kernels/fp4_quantized.h
awni: Yikes.. good catch!

awni: Is this also a bug in our regular quant or maybe I am missing something? https://github.com/ml-explore/mlx/blob/main/mlx/backend/metal/kernels/quantized.h#L143-L149

awni: From what I can tell N is always a multiple of `4` so this should be safe

awni:
> One thing that I should probably have commented on sooner is the choice of propagating the string everywhere.

I went back and forth on that for a while.. it seemed a bit superflous to convert it to an enum to make the primitive and then de-convert it in the dispatch. But probably you are right it is a bit more efficient, so I will update it.

PR #2506, opened 2025-08-17
Ensure small sort doesn't use indices if not argsort

angeloskath:
Line 114 would cause an invalid threadgroup load when ran with `MTL_SHADER_VALIDATION=1` . Slightly weird since it should have been removed by the compiler but at least with a bit of help it is now correct.


awni (approved):
Nice, thanks!

PR #2508, opened 2025-08-17
Ensure no oob read in gemv_masked

angeloskath:
Deals with reported OOB read when ran with `MTL_SHADER_VALIDATION=1 MTL_SHADER_VALIDATION_REPORT_TO_STDERR=1`.

I didn't see any perf regression due to the read happening after the check now so I think this is fine to merge.


awni (approved):
👍 👍

PR #2511, opened 2025-08-18
[CUDA] Add GEMM-based fallback convolution kernels

zcbenz:
Add fallback convolution kernels that work by converting convs to matmuls, the unfold kernels are ported from the Metal kernels.


angeloskath (approved):
It looks great! Did you by any chance measure the performance of the fallback? Mostly the bandwidth of the unfold kernel.

Either way it is a great v1 for the fallback kernel!

zcbenz:
> Did you by any chance measure the performance of the fallback? Mostly the bandwidth of the unfold kernel.

Nope, I thought it being a fallback kernel can excuse the slowness.

PR #2513, opened 2025-08-18
Fix wrong strides when inserting a dim with expand_dims

zcbenz:
Normally the stride of a dim with size-one does not matter, but when passing the strides to cuDNN it would lead to wrong results.


awni:
I'm not sure it's a good idea to rely on the stride of a singleton dimension to be 1. It might make more sense to pre-process the strides right before you send to cudnn rather than attempt to catch all the places where a stride could be set for a singleton dimension. Wdyt?

zcbenz:
I don't have a preference on this, handling this in cudnn code sounds good to me.

PR #2517, opened 2025-08-19
Custom cuda kernel

angeloskath:
Implements the same functionality as the `mx.fast.metal_kernel` for CUDA. It also adds the option of `precompiled_cuda_kernel` which is for now experimental but can already be used to run kernels implemented in python from triton (or potentially CuTe DSL).

On top of things it also adds the ability to configure jitted kernels the first time they are run (for things like dynamic shared memory) which can be useful even for our own matmuls if we start jitting them down the road.


mlx/backend/cuda/custom_kernel.cpp
```diff
@@ -0,0 +1,380 @@
...
+  }
+  auto kernel =
+      mod.get_kernel(kernel_name, [smem = shared_memory_](CUfunction kernel) {
+        if (smem > 0 && smem > 48000) {
+          cuFuncSetAttribute(
+              kernel, CU_FUNC_ATTRIBUTE_MAX_DYNAMIC_SHARED_SIZE_BYTES, smem);
+        }
+      });
+  encoder.add_kernel_node(kernel, grid, block, shared_memory_, args.args());
+  encoder.add_completed_handler([args = std::move(args)]() {});
```
angeloskath: @zcbenz shouldn't we generally be keeping the args around until after the kernel has been executed?

The arrays are fine but other arguments like shapes and so on may be freed before the kernel is run, did I misunderstand? If so I should remove the above 🤔

awni: I'm pretty sure the arguments get copied into the kernel node in the graph and that's why we don't keep them around.

From the cuda graph [API](https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__GRAPH.html#group__CUDART__GRAPH_1gcbdee72d9f3b4ce17654ee197fc02651):

> The kernelParams or extra array, as well as the argument values it points to, are copied during this call.

angeloskath: I see. That's also why we make sure that the shapes are resized to `NDIM` because it looks into the kernel signature to copy the full argument values into the node. That makes sense.

mlx/backend/cuda/jit_module.h
```diff
@@ -82,23 +85,30 @@ class JitModule {
...
 
   JitModule(const JitModule&) = delete;
   JitModule& operator=(const JitModule&) = delete;
-  CUfunction get_kernel(const std::string& kernel_name);
+  CUfunction get_kernel(
+      const std::string& kernel_name,
+      std::function<void(CUfunction)> configure_kernel);
+  CUfunction get_kernel(const std::string& kernel_name) {
+    return get_kernel(kernel_name, [](auto k) {});
+  }
```
zcbenz: Constructing an empty lambda is a bit expensive here, I think you can just use nullptr as default arg:

```suggestion
  CUfunction get_kernel(
      const std::string& kernel_name,
      std::function<void(CUfunction)> configure_kernel = nullptr);
```

and then do null check before calling it:

```c++
if (configure_kernel) {
  configure_kernel(it->second.first);
}
```

mlx/fast_primitives.h
```diff
@@ -293,19 +296,25 @@ class CustomKernel : public Primitive {
...
       : Primitive(stream),
         source_(std::move(source)),
         name_(std::move(name)),
         grid_(grid),
         threadgroup_(threadgroup),
         shape_infos_(std::move(shape_infos)),
         ensure_row_contiguous_(ensure_row_contiguous),
-        init_value_(init_value) {}
+        init_value_(init_value),
+        scalar_arguments_(scalar_arguments),
```
zcbenz:
```suggestion
        scalar_arguments_(std::move(scalar_arguments)),
```

mlx/backend/cuda/jit_module.cpp
```diff
@@ -312,21 +312,67 @@ JitModule::JitModule(
...
+  std::vector<char> ptx;
+  std::vector<std::pair<std::string, std::string>> ptx_kernels;
+
+  // Try to load them from the file cache
+  if (!read_cached_ptx(ptx_cache_dir(), module_name, &ptx, &ptx_kernels)) {
+    auto [precompiled, source_code, kernel_names] = builder();
+
+    // Get the PTX or cubin
+    if (precompiled) {
+      ptx.insert(ptx.begin(), source_code.begin(), source_code.end());
```
zcbenz: I think you can change the type of `ptx` from `std::vector<char>` to `std::string` and avoid the copy:

```suggestion
      ptx = std::move(source_code);
```

Most methods of `std::vector` and `std::string` are same and ideally code would just compile.

mlx/backend/cuda/jit_module.h
```diff
@@ -64,13 +65,15 @@ struct KernelArgs {
   std::vector<void*> args_;
 
   // The cuLaunchKernel API requires passing pointers to arguments so store
-  // temporary values untill kernel is launched.
+  // temporary values until the kernel is launched.
```
zcbenz: The comment is outdated (no more `cuLaunchKernel` calls): the args are reserved until added to graph.

mlx/backend/cuda/custom_kernel.cpp
```diff
@@ -0,0 +1,379 @@
...
+
+  // Compile the custom kernel
+  std::string kernel_name =
+      (is_precompiled_) ? name_ : "mlx::core::cu::" + name_;
+  cu::JitModule& mod = cu::get_jit_module(
+      s.device,
+      name_,
+      [&]() {
+        return std::make_tuple(
+            is_precompiled_, source_, std::vector<std::string>{kernel_name});
```
zcbenz: Small C++17 tip that the type is no longer required with deduction guides:

```suggestion
            is_precompiled_, source_, std::vector{kernel_name});
```

mlx/backend/cuda/custom_kernel.cpp
```diff
@@ -0,0 +1,379 @@
...
+      hash += "_";
+      hash += get_type_string(std::get<Dtype>(arg));
+    }
+  }
+
+  return hash;
+}
+
+std::string build_kernel(
+    std::string func_name,
```
zcbenz:
```suggestion
    const std::string& func_name,
```

python/src/fast.cpp
```diff
@@ -384,4 +398,216 @@ void init_fast(nb::module_& parent_module) {
...
+                "def __call__(self, *, inputs: List[Union[scalar, array]], output_shapes: List[Sequence[int]], output_dtypes: List[Dtype], grid: tuple[int, int, int], threadgroup: tuple[int, int, int], template: Optional[List[Tuple[str, Union[bool, int, Dtype]]]] = None, init_value: Optional[float] = None, verbose: bool = false, stream: Union[None, Stream, Device] = None)"),
+            R"pbdoc(
+            Run the kernel.
+
+            Args:
+              inputs (List[array]): The inputs passed to the CUDA kernel.
+              output_shapes (List[Sequence[int]]): The list of shapes for each output in ``output_names``.
+              output_dtypes (List[Dtype]): The list of data types for each output in ``output_names``.
+              grid (tuple[int, int, int]): 3-tuple specifying the grid to launch the kernel with.
+                For compatibility with :func:`metal_kernel` the grid is in threads and not in threadblocks.
```
zcbenz: In Metal when you launch a grid of 200 threads with a threadgroup size of 128, a partial threadgroup would be created for the remaining 72 threads. I think we can not emulate this feature in CUDA?

mlx/backend/cuda/jit_module.h
```diff
@@ -82,23 +85,30 @@ class JitModule {
   JitModule(
       Device& device,
       const std::string& module_name,
-      const KernelBuilder& builder);
+      const KernelBuilder& builder,
+      bool cache);
```
zcbenz, line 89: It is vague what `cache` means here, maybe rename to `use_disk_cache` or add a comment here.

python/src/fast.cpp
angeloskath: Yeah unfortunately we have to write kernels explicitly handling that. We can emulate it but we would need to pass the size of the grid to the kernel and add a check on the top of the kernel something like

```
auto elem = cooperative_groups::this_grid().thread_rank();
if (elem >= total_threads) {
  return;
}
```

My current thesis is that it is moot to try to have consistent behavior across Metal and CUDA for the custom kernels so people should code their custom kernels accordingly.

angeloskath:
@zcbenz thanks for the detailed and useful comments!

python/src/cuda.cpp
```diff
@@ -0,0 +1,19 @@
...
+#include "mlx/backend/cuda/cuda.h"
+
+namespace mx = mlx::core;
+namespace nb = nanobind;
+
+void init_cuda(nb::module_& m) {
+  nb::module_ cuda = m.def_submodule("cuda", "mlx.cuda");
+
+  cuda.def(
+      "is_available",
```
awni, line 14: should we add it to the docs?

python/src/fast.cpp
awni: consistent with below
```suggestion
                For compatibility with :func:`metal_kernel` the grid is in threads and not in threadgroups.
```

python/src/fast.cpp
```diff
@@ -384,4 +398,216 @@ void init_fast(nb::module_& parent_module) {
           b = exp_elementwise(a)
           assert mx.allclose(b, mx.exp(a))
      )pbdoc");
+
+  m.def(
+      "cuda_kernel",
```
awni, line 403: I think this one can be added to the docs.. it's pretty stable I assume.

PR #2518, opened 2025-08-19
Fix docs

russellizadi:
Remove state return from function example in compile documentation.

In the documentation example, the text notes that “returning updated state can be pretty inconvenient.” This snippet shows that when using functools.partial, there’s no need to return the outputs argument explicitly.

```python
from functools import partial

import mlx.core as mx

state = []

# Tell compile to capture state as an output
@partial(mx.compile, outputs=state)
def fun(x, y):
    z = x + y
    state.append(z)
    return mx.exp(z)

fun(mx.array(1.0), mx.array(2.0))
# Prints [array(3, dtype=float32)]
print(state)
```

## Checklist

### DIFF
--- a/docs/src/usage/compile.rst
+++ b/docs/src/usage/compile.rst
@@ -225,7 +225,7 @@ In some cases returning updated state can be pretty inconvenient. Hence,
   def fun(x, y):
       z = x + y
       state.append(z)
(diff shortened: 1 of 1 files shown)


angeloskath (approved):
🙏

PR #2523, opened 2025-08-21
Fix power

awni:
- Integer power was broken on CPU
- Integer power to negative power caused hang on all devices


angeloskath (approved):
🙏

PR #2527, opened 2025-08-22
Fix warning 186-D from nvcc

zcbenz:
Fix warning:

```
/root/codes/mlx/mlx/backend/cuda/device/binary_ops.cuh(208): warning #186-D: pointless comparison of unsigned integer with zero
        if (exp < 0) {
                ^
```


awni (approved):
Thanks! maybe we should try to build warning free in circle like we do for macOS to catch this during PR

PR #2528, opened 2025-08-22
nccl default for backend=any

awni:
- NCCL is the default for backend=any when cuda `is_available`
- NCCL logging is off by default
- Fix release build

### DIFF
--- a/python/mlx/distributed_run.py
+++ b/python/mlx/distributed_run.py
@@ -55,6 +55,11 @@ def parse_hardware_ports(ports_string):
     return ports
 
 
+def get_num_nvidia_gpus():
+    result = run(['nvidia-smi', "-L"], capture_output=True, text=True, check=True)
+    return len(result.stdout.strip().split("\n"))
+
+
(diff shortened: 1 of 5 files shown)


python/mlx/distributed_run.py
```diff
@@ -428,7 +428,7 @@ def launch_nccl(parser, hosts, args, command):
     base_env = os.environ.copy()
     base_env.update(
         {
-            "NCCL_DEBUG": "INFO",
+            "NCCL_DEBUG": base_env.get("NCCL_DEBUG", "DEBUG"),
```
angeloskath: Maybe default to debug only if --verbose is passed to launch 🤔

angeloskath (approved):
👍

python/mlx/distributed_run.py
awni: The default of `DEBUG` is less verbose than the default of `INFO` (as it was). So if we pass `--verbose` I will set it to `INFO` (which is likely what you meant).

PR #2530, opened 2025-08-22
Fix allocation bug in NCCL

awni:
Accidentally introduced this bug in https://github.com/ml-explore/mlx/pull/2528


angeloskath (approved):
👍

PR #2531, opened 2025-08-23
[CUDA] Use ConcurrentContext in concatenate_gpu

zcbenz:


awni:
I actually tried this hoping it would speed up the concatenation in average gradients for distributed training but it was oddly slower

zcbenz:
It also seems to make inference slower. I'll look into what is actually happening and do a proper benchmark.

zcbenz:
For simple tests this branch is 10% faster:

```python
import time
import mlx.core as mx

def func():
  a = mx.arange(0, 1024)
  b = mx.arange(1024, 2048)
  c = mx.arange(2048, 4089)
  return mx.concatenate([a, b, c]) 

for _ in range(100):
  mx.eval(func())

tic = time.time()
for _ in range(1000):
  mx.eval(func())
toc = time.time()

s = toc - tic 
print(1000 / s)
```

And the cuda graph supports the improvement:

Before (node 0 1 2 are arange) :

dot1

After:

dot2

zcbenz:
For inference, `mx.concatenate` is only used when updating kv cache, but after a few more testings I think it is likely not making a difference to inference speed. I should probably do a benchmark of kv cache for sure.

> I actually tried this hoping it would speed up the concatenation in average gradients for distributed training but it was oddly slower

@awni Is there a simple example that I can use to debug this?

awni:
> @awni Is there a simple example that I can use to debug this?

What I ran isn't that simple unfortunately. It was a multi-gpu pre-training run.. it seemed like this exact change made it slower. I can point you to some code offline to reproduce.

PR #2532, opened 2025-08-23
Remove stream from average grads so it uses default

awni:
Remove the stream from `average_gradients` so that it uses the default (and doesn't need to be specified). It could be optional but one can also set it with a context manager..


awni:
Would you prefer we keep the stream and default it to None?

angeloskath:
Yeah I think it makes a bit more sense but probably nobody will use it. It is also not clear that it is the communication stream. 

Eg passing a different cuda stream for NCCL to communicate over would be valid in principle but it is not clear that this is what is happening if one passes `stream=mx.new_stream(mx.gpu)`. So I think maybe remove it or change it to communication stream ?

I am fine either way.

PR #2534, opened 2025-08-24
Enable COMPILE_WARNING_AS_ERROR for linux builds in CI

zcbenz:
Also update gguf-tools to fix compile warnings.


awni (approved):
Nice, thanks!!

PR #2539, opened 2025-08-25
Use nccl header only when nccl is not present

awni:
This downlaods nccl.h if NCCL is not installed. It's really intended for building the MLX CUDA back-end on a CPU-only machine (where it won't be run)so we can do the PyPi release.

It may be an issue if someone build from source without NCCL since I think it will be a link error rather than run without NCCL. So it might be good to require NCCL unless a specific flag is passed..

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -394,7 +394,7 @@ jobs:
         default: ""
     machine:
       image: ubuntu-2204:current
-      resource_class: large
+      resource_class: xlarge
     steps:
       - checkout
       - run:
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -394,7 +394,7 @@ jobs:
         default: ""
     machine:
       image: ubuntu-2204:current
-      resource_class: large
+      resource_class: xlarge
```
awni, line 397: This was taking 50m to build the CUDA wheel 🤯 .. hopefully the larger one will be 2x faster.

PR #2541, opened 2025-08-25
Allow pathlib.Path to save/load functions

### LINKED ISSUE #2522: [feature request] `mlx.core.load` and `mlx.core.save` should accept `pathlib.Path`
The [numpy load](https://numpy.org/doc/stable/reference/generated/numpy.load.html) and save functions allow the caller to pass in a pathlib.Path object. Right now, users of MLX need to cast pathlib objects to string before using load/save.

awni:
Closes #2522


angeloskath (approved):
🙏

PR #2543, opened 2025-08-25
Tests for  save/load with `Path`

awni:
Add tests for loading with Path. Sorry I had these locally and somehow missed them in the PR.. `mx.scatter_brained`


angeloskath (approved):
🙏

PR #2544, opened 2025-08-26
Run CPP tests for CUDA build in CI

zcbenz:
Fix a few C++ tests for CUDA backend and run them in CI.


awni (approved):
Awesome, thanks for adding that!

PR #2545, opened 2025-08-26
Enable cuda graph toggle

awni:
- Add `MLX_USE_CUDA_GRAPHS` to enable easy benchmarking and debugging with and without cuda graphs
- Increase graph cache default size to 400

For training a deep Transformer using a larger graph cache gives a nice speed imporvement.

Benchmarks with and without cuda graphs:

```
mlx_lm.generate --model mlx-community/Meta-Llama-3.1-8B-Instruct-bf16 --prompt "Write a story about Einstein" -m 256
```
No graph: `121.675 tokens-per-sec`
Graph: `147.092 tokens-per-sec`

Transformer training:
```
python main.py --gpu --batch_size 8 --dtype bfloat16
```
No graph: `It/sec 20.158`
Graph: `It/sec 21.532`

### DIFF
--- a/mlx/backend/cuda/device.cpp
+++ b/mlx/backend/cuda/device.cpp
@@ -34,6 +34,14 @@ int cuda_graph_cache_size() {
   return cache_size;
 }
 
+bool use_cuda_graphs() {
+  static bool use_graphs = []() {
+    return env::get_var("MLX_USE_CUDA_GRAPHS", true);
+  }();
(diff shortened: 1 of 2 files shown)


awni:
I don't think we need to keep this option around forever.. but it's pretty convenient while we are figuring out perf issues (sometimes related to cuda graphs) and ironing out other bugs that are sometimes related to graph capture.

PR #2546, opened 2025-08-26
[CUDA] Link with nccl

awni:
This should be the missing piece to get the release to properly link with NCCL.


zcbenz (approved):
Nice trick!

PR #2548, opened 2025-08-26
Separate cpu compilation cache by versions

zcbenz:
Avoid reusing cpu compilation cache generated by other versions, which may crash when there were changes to cpu compilation code.


awni (approved):
Long overdue, thx!

PR #2550, opened 2025-08-27
[CUDA] fix sort

### LINKED ISSUE #2537: [BUG] [CUDA] Error in graph capture for argpartition
The following failed on an H100:

```python
import mlx.core as mx

a = mx.zeros((512, 128))
mx.eval(mx.argpartition(a, kth=-32, axis=-1))
```

With the following error:

```
  what():  cudaStreamEndCapture(stream, &handle_) failed: operation failed due to a previous error during capture
```

awni:
The CUB StableSort doesn't work with graph capture. So this uses the Radix sort instead. Closes #2537.

Unfortunately for some sizes it is noticeably slower for some shapes (I see up to 2x slower) and so we likely need to revisit this in the future.

### DIFF
--- a/mlx/backend/cuda/sort.cu
+++ b/mlx/backend/cuda/sort.cu
@@ -9,7 +9,7 @@
 #include <nvtx3/nvtx3.hpp>
 #include <thrust/device_ptr.h>
 #include <thrust/transform.h>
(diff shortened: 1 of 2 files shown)


zcbenz (approved):
🙏

PR #2552, opened 2025-08-28
Faster contiguous gather for indices in the first axis

awni:
This should speed up many use cases that look like `a[b]` (embedding lookups, gather in MoE, etc).

For Qwen3 MoE prompt processing we get a nice speedup:

2048 tokens on M2U

Pre: `prompt_tps=2359.717`
Post: `prompt_tps=2454.472`

### DIFF
--- a/python/mlx/nn/layers/quantized.py
+++ b/python/mlx/nn/layers/quantized.py
@@ -98,11 +98,10 @@ class QuantizedEmbedding(Module):
         # Initialize the quantized weight
         scale = math.sqrt(1 / dims)
         weight = mx.random.normal(shape=(num_embeddings, dims), scale=scale)
-        self.weight, *scales_biases = mx.quantize(weight, group_size, bits, mode=mode)
-        if mode == "affine":
-            self.scales, self.biases = scales_biases
-        else:
-            (self.scales,) = scales_biases
+        self.weight, self.scales, *biases = mx.quantize(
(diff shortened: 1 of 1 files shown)


mlx/backend/metal/kernels/indexing/gather_front.h
```diff
@@ -0,0 +1,21 @@
...
+    device T* out,
+    const constant int64_t& stride,
+    const constant int& size,
+    uint2 index [[thread_position_in_grid]],
+    uint2 grid_dim [[threads_per_grid]]) {
+  auto idx = offset_neg_idx(indices[index.y], size);
+  LocT src_idx = static_cast<LocT>(stride) * idx;
+  LocT out_idx = static_cast<LocT>(stride) * index.y;
+
+  out[out_idx + index.x] = src[src_idx + index.x];
```
angeloskath: May not matter much but why not add work-per-thread? Especially since this kernel is always jitted it should be fairly trivial and with no real downside.

angeloskath (approved):
🚀

mlx/backend/metal/kernels/indexing/gather_front.h
awni: It’s a good point. I tried it and I didn’t see a meaningful speedup. I am not entirely sure why.. needs a little more care to figure out

PR #2558, opened 2025-08-31
Add sdpa with sinks

awni:
Not yet intended to merge yet (or maybe ever). But I want to make a PR for this so we can debate the trade-offs with something concrete.

See https://github.com/ml-explore/mlx-lm/pull/418 for corresponding change to gpt-oss in mlx-lm.

Prompt-processing time gets a nice speed bump:

For prompt length 2048 on M2U:

Pre: `prompt_tps=1299.417`
Post: `prompt_tps=1475.834`

### DIFF
--- a/mlx/backend/metal/kernels/sdpa_vector.h
+++ b/mlx/backend/metal/kernels/sdpa_vector.h
@@ -9,6 +9,7 @@ constant bool query_transposed [[function_constant(21)]];
 constant bool do_causal [[function_constant(22)]];
 constant bool bool_mask [[function_constant(23)]];
 constant bool float_mask [[function_constant(24)]];
+constant bool has_sinks [[function_constant(25)]];
 
 template <typename T, int D, int V = D>
 [[kernel]] void sdpa_vector(
@@ -31,6 +32,9 @@ template <typename T, int D, int V = D>
     [[buffer(14), function_constant(has_mask)]],
     const constant int& mask_head_stride
     [[buffer(15), function_constant(has_mask)]],
+    const device T* sinks [[buffer(16), function_constant(has_sinks)]],
+    const constant int& num_q_heads
+    [[buffer(17), function_constant(has_sinks)]],
     uint3 tid [[threadgroup_position_in_grid]],
     uint3 tpg [[threadgroups_per_grid]],
     uint simd_gid [[simdgroup_index_in_threadgroup]],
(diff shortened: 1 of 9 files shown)


mlx/backend/metal/kernels/steel/attn/kernels/steel_attention.h
```diff
@@ -232,6 +234,15 @@ template <
     max_score[i] = Limits<AccumType>::finite_min;
   }
 
+  // TODO condition here is wrong
+  if (has_sinks) {
```
awni: This condition is not right, sinks need to be incorporated once per vector of qk scores.. I didn't figure it out yet.

angeloskath: Let me know if you want me to test this more but I think this should be

```cpp
if (has_sinks && sn == 0) {
}
```

basically once per row.

mlx/backend/metal/kernels/sdpa_vector.h
```diff
@@ -268,6 +291,10 @@ template <typename T, int D, int V = D>
         score += q[i] * k[i];
       }
       score = simd_sum(score);
+      if (score < Limits<T>::finite_min) {
```
angeloskath, line 292: Interesting does that fix a bug somewhere?

angeloskath (commented):
This looks great. I think I left a comment that fixes the batch case if not let me know and I will investigate further.

mlx/backend/metal/kernels/sdpa_vector.h
awni: Yea it's very much an edge case but if `q_i^T k_j = -inf` for any pair then previously it would create NaNs, now it will do the right thing. I added a test for it as well.

mlx/backend/metal/kernels/steel/attn/kernels/steel_attention.h
awni: It sounds like it should be right.. but it doesn't pass the test case I added for this 🤔

awni: As it turns out the original condition is correct (it needs to be applied for every thread).. and the problem was the log2(e) factor.

awni:
Ok I think this is ready to land once it's approved. Lmk if there are any other comments.

mlx/backend/metal/kernels/steel/attn/kernels/steel_attention.h
angeloskath: Sorry for the noise then :-)

Issue #2208, opened 2025-05-20
[BUG] Slice does not write through, unlike numpy


hlovatt:
**Describe the bug**
A slice in NumPy writes through to underlying array, MLX doesn't. MLX behaves like Python, not NumPy.

**To Reproduce**

```python
import mlx.core as mx
a = mx.array([1, 2, 3])
b = a[1:]
b[0] = 0
a, b
```
`(array([1, 2, 3], dtype=int32), array([0, 3], dtype=int32))`

Note how 0 is **not** written through b into a.

**Expected behavior**
In NumPy:
```python
import numpy as np
e = np.array([1, 2, 3])
f = e[1:]
f[0] = 0
e, f
```
`(array([1, 0, 3]), array([0, 3]))`

Note how 0 is written through f into e.

**Desktop (please complete the following information):**
 - MacOS 15.5
 - MLX 0.25.2

**Additional context**
Only tested inside Jupyter!

zcbenz:
This is expected as designed, I'm closing this issue as the behave had been documented by https://github.com/ml-explore/mlx/pull/2559.

PR #2559, opened 2025-08-31
chore: Update Docs With Slice Copy Example

krishi-saripalli:
## Proposed changes

Issue #2208 points out that unlike NumPy, modifying a slice does not modify the original array. This update to the documentation in `Indexing Arrays > In Place Updates` clearly describes this behaviour so that people coming from NumPy are aware of the difference

image

## Checklist


awni (approved):
Thanks!

PR #2562, opened 2025-09-01
Fix dequantize python sig

wrmsr:
## Proposed changes

The `dequantize` op was [recently changed](https://github.com/ml-explore/mlx/commit/70560b6bd5324efbfd9f97c9076c884d21ffac34#diff-8a607e80673e9f8e9c1e73a5aba9d6f89ac56ad34765e4022d5130f890776cacR4167-R4274) to have a broken type signature, breaking mypy checks against it:

```
.venvs/default/lib/python3.13/site-packages/mlx/core/__init__.pyi:1653:71: error: Invalid syntax  [syntax]
    def dequantize(w: array, /, scales: array, biases: Optional[array] = = None, group_size: int = 64, bit...
```

## Checklist

### DIFF
--- a/python/src/ops.cpp
+++ b/python/src/ops.cpp
@@ -4271,7 +4271,7 @@ void init_ops(nb::module_& m) {
       nb::kw_only(),
       "stream"_a = nb::none(),
       nb::sig(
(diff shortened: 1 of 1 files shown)


awni (approved):
Thanks!

PR #2564, opened 2025-09-03
Add batch offsets for mx.fast.rope

awni:
This changes `mx.fast.rope` to accept a vector of offsets instead of just a single scalar. This is necessary for batch-decoding where the examples in the batch can be at different positions/ offsets.


ZelinMa557:
May I ask, when can mlx users use this feature via `pip install mlx`? This feature is really important for batch infer engine, looking forward to use it!

awni:
We will aim to get a patch release out this week.

PR #2565, opened 2025-09-04
typing: add type hints to mlx.core.array, linalg, distributed, and random

XXXXRT666:
## Proposed changes

image

image

image

image

image

## Checklist

### DIFF
--- a/python/src/array.cpp
+++ b/python/src/array.cpp
@@ -358,7 +358,7 @@ void init_array(nb::module_& m) {
       .def(
           "tolist",
           &tolist,
-          nb::sig("def tolist(self) -> listscalar"),
+          nb::sig("def tolist(self) -> list_or_scalar"),
           R"pbdoc(
             Convert the array to a Python :class:`list`.
 
--- a/python/src/array.cpp
+++ b/python/src/array.cpp
@@ -320,6 +320,7 @@ void init_array(nb::module_& m) {
       .def_prop_ro(
           "shape",
           [](const mx::array& a) { return nb::cast(a.shape()); },
+          nb::sig("def shape(self) -> tuple[int, ...]"),
           R"pbdoc(
           The shape of the array as a Python tuple.
 
@@ -347,6 +348,7 @@ void init_array(nb::module_& m) {
       .def(
           "item",
           &to_scalar,
+          nb::sig("def item(self) -> scalar"),
           R"pbdoc(
             Access the value of a scalar array.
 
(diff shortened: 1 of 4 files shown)


python/src/random.cpp
```diff
@@ -171,7 +171,7 @@ void init_random(nb::module_& parent_module) {
       "key"_a = nb::none(),
       "stream"_a = nb::none(),
       nb::sig(
-          "def normal(shape: Sequence[int] = [], dtype: Optional[Dtype] = float32, loc: Optional[scalar, array] = None, scale: Optional[scalar, array] = None, key: Optional[array] = None, stream: Union[None, Stream, Device] = None) -> array"),
+          "def normal(shape: Sequence[int] = [], dtype: Optional[Dtype] = float32, loc: Optional[Union[scalar, array]] = None, scale: Optional[Union[scalar, array]] = None, key: Optional[array] = None, stream: Union[None, Stream, Device] = None) -> array"),
```
awni: It's more common / consistent with MLX to write that as `Union[scalar, array, None]`. Mind changing?

python/src/distributed.cpp
```diff
@@ -35,7 +35,7 @@ void init_distributed(nb::module_& parent_module) {
           &mx::distributed::Group::split,
           "color"_a,
           "key"_a = -1,
-          nb::sig("def split(self, color: int, key: int = -1) -> Group"),
+          nb::sig("def split(self, color: int, key: int = -1) -> 'Group'"),
```
awni: Why do we need quotes on `Group`?

XXXXRT666: Sorry, at first I thought the red squiggly line would affect parsing, but it seems it actually doesn’t affect type parsing. I would revert it

python/src/array.cpp
```diff
@@ -356,6 +358,7 @@ void init_array(nb::module_& m) {
       .def(
           "tolist",
           &tolist,
+          nb::sig("def tolist(self) -> listscalar"),
```
awni: The return type is not right because it can be a list or a scalar. E.g. it doesn't have to be a list.

python/src/random.cpp
XXXXRT666: Thanks for the suggestion

python/src/array.cpp
awni: For example:

`mx.array(0).tolist()` has type `int`

XXXXRT666: I have changed it to list_or_scalar

awni: It looks like the return type needs to be updated here.

awni (approved):
Thanks!

PR #2568, opened 2025-09-05
fix: library loading for swift dynamic frameworks

bilousoleksandr:
## Proposed changes

By default, Dynamic Framework doesn't create a resources bundle and packs all resources inside its own `Resources` folder. 

To fix the issue related to SwiftPM, add try loading metal library from `Framework/../Resources/default.metallib` when `mlx` codebase is built as a dynamic framework.

## Checklist

### DIFF
--- a/mlx/backend/metal/device.cpp
+++ b/mlx/backend/metal/device.cpp
@@ -127,12 +127,18 @@ MTL::Library* load_default_library(MTL::Device* device) {
     return lib;
   }
 
+  // Try lo load resources from Framework resources if SwiftPM wrapped as a dynamic framework.
+  std::tie(lib, error[3]) = load_colocated_library(device, "Resources/default");
+  if (lib) {
+    return lib;
+  }
+
   // Finally try default_mtllib_path
-  std::tie(lib, error[3]) = load_library_from_path(device, default_mtllib_path);
+  std::tie(lib, error[4]) = load_library_from_path(device, default_mtllib_path);
   if (!lib) {
     std::ostringstream msg;
     msg << "Failed to load the default metallib. ";
-    for (int i = 0; i < 4; i++) {
+    for (int i = 0; i < 5; i++) {
       if (error[i] != nullptr) {
         msg << error[i]->localizedDescription()->utf8String() << " ";
       }
--- a/mlx/backend/metal/device.cpp
+++ b/mlx/backend/metal/device.cpp
@@ -108,7 +108,7 @@ std::pair<MTL::Library*, NS::Error*> load_swiftpm_library(
 }
 
 MTL::Library* load_default_library(MTL::Device* device) {
-  NS::Error* error[4];
+  NS::Error* error[5];
   MTL::Library* lib;
   // First try the colocated mlx.metallib
   std::tie(lib, error[0]) = load_colocated_library(device, "mlx");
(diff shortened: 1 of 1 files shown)


awni:
@davidkoski do you mind checking this change is useful for MLX Swift?

mlx/backend/metal/device.cpp
```diff
@@ -127,12 +127,18 @@ MTL::Library* load_default_library(MTL::Device* device) {
     return lib;
   }
 
+  // Try lo load resources from Framework resources if SwiftPM wrapped as a dynamic framework.
+  std::tie(lib, error[3]) = load_colocated_library(device, "Resources/default");
+  if (lib) {
+    return lib;
+  }
```
davidkoski, line 135: `load_colocated_library` will get the directory of the binary that contains the `get_binary_directory()`.  In the case of an embedded .framework it would be something like `x.app/Contents/Frameworks/MLX.framework/Versions/A`.  This would then append `Resources/default` and `.metallib` onto that and load the file.

I think this works if the `default.metallib` is hoisted out of the `mlx-swift_Cmlx.bundle/Contents/Resources`.  I am not sure it is because I don't have a .framework wrapping the MLX swiftpm library myself, but such a thing could surely be done.

davidkoski: That said, I am surprised that this does not do the same thing:

```c++
std::pair<MTL::Library*, NS::Error*> load_swiftpm_library(
    MTL::Device* device,
    const std::string& lib_name) {
#ifdef SWIFTPM_BUNDLE
...
  auto bundles = NS::Bundle::allBundles();
  for (int i = 0, c = (int)bundles->count(); i < c; i++) {
    auto bundle = reinterpret_cast<NS::Bundle*>(bundles->object(i));
    library = try_load_bundle(device, bundle->resourceURL(), lib_name);
```

That should scan all bundles (frameworks are considered bundles) looking for a `default.metallib`.

davidkoski:
> @davidkoski do you mind checking this change is useful for MLX Swift?

It might be, but:

1) I don't have a good way to test it as I don't have a framework wrapped MLX and it seems to require a particular layout of the framework
2) the existing `load_swiftpm_library` ought to do something identical when it scans the bundles -- I wonder why this didn't work?

mlx/backend/metal/device.cpp
bilousoleksandr: It doesn't work this way, because bundles doesn't contain MLX framework path. In current configuration, it always looks for for main bundle and tries to load `default.metallib` from `mlx-swift_Cmlx.bundle`. But with dynamic frameworks, app directory looks next way: 
- App.app
  - Contents 
    - Frameworks
      - MLX
        - Resources
           - default.metallib
    - Resources

```
NSBundle <~Work/DerivedData/Eney-dyyiohebtzyjuwbwalhqtrcvkvcn/Build/Products/Debug/App.app> (loaded),
NSBundle </System/Library/Extensions/AGXMetalG13X.bundle> (loaded),
NSBundle </System/Library/Input Methods/PressAndHold.app/Contents/PlugIns/PAH_Extension.appex> (not yet loaded),
NSBundle </System/Library/CoreServices/SystemAppearance.bundle> (not yet loaded)
```

bilousoleksandr: @davidkoski I've covered a questions related to `load_swiftpm_library` above

davidkoski: OK, if the default.metallib is in the Resources directory, I agree this looks good.

bilousoleksandr:
> @bilousoleksandr the `NS::Error error[4]` needs to be `NS::Error error[5]`.

Fixed

bilousoleksandr:
@angeloskath could you re-approve the PR to enable CI run?

PR #2571, opened 2025-09-06
Faster complex matmul

### LINKED ISSUE #2076: Use cblas_cgemm for CPU complex matmul
Should be faster than the current op based approach.

https://developer.apple.com/documentation/accelerate/1513288-cblas_cgemm?language=objc

CC-Yeh:
## Proposed changes

- Introduce `cblas_cgemm`

- `Metal`: Make `gemv` compatible with `complex64_t`

- `Metal`: Add `complex64` `BlockMMA` specialization to simplify `gemm` integration.

- `Cuda` :  Make `gemv` and `gemm` compatible with `complex64_t`.

Only tuned on small chip, will need people with larger chips to tune the tile size.

Closes #2076

##Benchmarks

### Metal
Average 6x faster for new `gemv` (bench_gemv.py)

output

Average 1.5x faster for new `gemm` (bench_gemm.py)

output (1)

### CUDA
Average 6x faster for new `gemv`
output (3)
output (2)

`gemm`
~1.7 times lower, need help from someone with CUDA expertise
output (4)

## Checklist

### DIFF
--- a/mlx/backend/cuda/gemms/cublas_gemm.cpp
+++ b/mlx/backend/cuda/gemms/cublas_gemm.cpp
@@ -52,7 +52,8 @@ cublasComputeType_t dtype_to_compute_type(Dtype dtype) {
     case float64:
       return CUBLAS_COMPUTE_64F;
     case complex64:
-      return CUBLAS_COMPUTE_32F;
+      return mlx::core::env::enable_tf32() ? CUBLAS_COMPUTE_32F_FAST_TF32
+                                           : CUBLAS_COMPUTE_32F;
     default:
       throw std::runtime_error(fmt::format(
           "Unsupported dtype in CublasGemm: {}.", dtype_to_string(dtype)));


mlx/ops.cpp
```diff
@@ -1228,15 +1228,15 @@ array pad(
     if (low_pad_size[i] < 0) {
       std::ostringstream msg;
       msg << "Invalid low padding size (" << low_pad_size[i]
-          << ") passed to pad" << " for axis " << i
-          << ". Padding sizes must be non-negative";
+          << ") passed to pad"
+          << " for axis " << i << ". Padding sizes must be non-negative";
```
awni:
```suggestion
          << ") passed to pad for axis "
           << i << ". Padding sizes must be non-negative";
```

mlx/ops.cpp
```diff
@@ -1228,15 +1228,15 @@ array pad(
...
       throw std::invalid_argument(msg.str());
     }
     if (high_pad_size[i] < 0) {
       std::ostringstream msg;
       msg << "Invalid high padding size (" << high_pad_size[i]
-          << ") passed to pad" << " for axis " << i
-          << ". Padding sizes must be non-negative";
+          << ") passed to pad"
+          << " for axis " << i << ". Padding sizes must be non-negative";
```
awni:
```suggestion
          << ") passed to pad for axis " 
          << i << ". Padding sizes must be non-negative";
```

mlx/ops.cpp
```diff
@@ -3651,15 +3631,15 @@ Shape conv_out_shape(
 
     if (pads_lo[i - 1] < 0 || pads_hi[i - 1] < 0) {
       std::ostringstream msg;
-      msg << "[conv] Padding sizes must be non-negative." << " Got padding "
-          << pads_lo << " | " << pads_hi << ".";
+      msg << "[conv] Padding sizes must be non-negative."
+          << " Got padding " << pads_lo << " | " << pads_hi << ".";
```
awni:
```suggestion
      msg << "[conv] Padding sizes must be non-negative. Got padding " 
      << pads_lo << " | " << pads_hi << ".";
```

mlx/ops.cpp
```diff
@@ -3651,15 +3631,15 @@ Shape conv_out_shape(
...
       throw std::invalid_argument(msg.str());
     }
 
     if (strides[i - 1] <= 0) {
       std::ostringstream msg;
-      msg << "[conv] Stride sizes must be positive." << " Got strides "
-          << strides << ".";
+      msg << "[conv] Stride sizes must be positive."
+          << " Got strides " << strides << ".";
```
awni:
```suggestion
      msg << "[conv] Stride sizes must be positive. Got strides " 
      << strides << ".";
```

mlx/ops.cpp
```diff
@@ -3696,7 +3676,8 @@ run_conv_checks(const array& in, const array& wt, int n_dim, int groups) {
   if (in.ndim() != n_dim + 2) {
     std::ostringstream msg;
     msg << "[conv] Invalid input array with " << in.ndim() << " dimensions for "
-        << n_dim << "D convolution." << " Expected an array with " << n_dim + 2
+        << n_dim << "D convolution."
+        << " Expected an array with " << n_dim + 2
```
awni:
```suggestion
        << n_dim << "D convolution. Expected an array with " 
        << n_dim + 2
```

mlx/ops.cpp
```diff
@@ -3184,8 +3164,8 @@ array scatter_axis(
       (mode == ScatterAxis::None) ? "[put_along_axis]" : "[scatter_add_axis]";
   if (axis + a.ndim() < 0 || axis >= static_cast<int>(a.ndim())) {
     std::ostringstream msg;
-    msg << prefix << " Received invalid axis " << " for array with " << a.ndim()
-        << " dimensions.";
+    msg << prefix << " Received invalid axis "
+        << " for array with " << a.ndim() << " dimensions.";
```
awni:
```suggestion
    msg << prefix << " Received invalid axis for array with " 
            << a.ndim() << " dimensions.";
```

mlx/ops.cpp
```diff
@@ -3146,8 +3126,8 @@ array take_along_axis(
     StreamOrDevice s /* = {} */) {
   if (axis + a.ndim() < 0 || axis >= static_cast<int>(a.ndim())) {
     std::ostringstream msg;
-    msg << "[take_along_axis] Received invalid axis " << " for array with "
-        << a.ndim() << " dimensions.";
+    msg << "[take_along_axis] Received invalid axis "
+        << " for array with " << a.ndim() << " dimensions.";
```
awni:
```suggestion
    msg << "[take_along_axis] Received invalid axis for array with " '
    << a.ndim() << " dimensions.";
```

awni:
This looks like a very nice improvement. There are two issues we need to work through:
1. The JIT tests are broken on macOS. Are you able to resolve that?
2. We need a strategy for CUDA. cuBlas has support for [complex 64-bit type](https://docs.nvidia.com/cuda/cublas/#cudadatatype-t), so maybe we can just route to that?

Are you up for making those two fixes? If you are unable to make the CUDA change we can help with that, let us know.

CC-Yeh:
> This looks like a very nice improvement. There are two issues we need to work through:
> 
> 1. The JIT tests are broken on macOS. Are you able to resolve that?

Hope the latest commit will fix this.

> 2. We need a strategy for CUDA. cuBlas has support for [complex 64-bit type](https://docs.nvidia.com/cuda/cublas/#cudadatatype-t), so maybe we can just route to that?
> 

I’d love to add it. I don’t have access to an NVIDIA GPU, so I can’t test CUDA locally.

awni:
Hi Dan - could you DM me and we can discuss work-arounds? You can reach me on [X](https://x.com/awnihannun) or at awni at apple.com.

raishish:
I was planning to work on this issue as well but @Dan-Yeh beat me to it lol.

plot_nn
plot_nt
plot_tn

Anyway, I benchmarked the CPU backend on my M2 Air (`cblas_gemm` vs the current op-based Karatsuba algorithm).

**Observations:**
- Karatsuba is a bit quicker (20-40%) for some cases for `nn` transpose config. I suppose I can scale this further to see if this trend continues.
- However, in case of `nt` and `tn` configs, the cblas implementation is overall much faster. My guess is since it can intrinsically account for the transpose operation, it makes for a more efficient implementation.

I don't suppose it makes sense to retain the Karatsuba approach for CPU. In hindsight, it makes sense that cblas would run faster than the naive Karatsuba decomposition but I'm glad I could actually benchmark it for my curiosity.

CC: @awni

CC-Yeh:
@awni 

> 2. We need a strategy for CUDA. cuBlas has support for [complex 64-bit type](https://docs.nvidia.com/cuda/cublas/#cudadatatype-t), so maybe we can just route to that?

I think `CUDA_C_32F` is already [supported](https://github.com/ml-explore/mlx/blob/main/mlx/backend/cuda/gemms/cublas_gemm.cpp#L71).

awni:
Oh ok.. in that case we may just need to update the routing conditions so we don't route to the custom gemv when the input has complex type.

CC-Yeh:
Interestingly, ops based solution can outperform `cublasLt` in 18/21 cases, except skinny or small matrix.

output (4)

CC-Yeh:
Hey @awni, can you take a look again? 

I believe the CPU, Metal, and CUDA parts are in good shape. One note: CUDA::gemm runs slower than the prior ops-based route and could use optimization by a CUDA specialist.

awni:
> One note: CUDA::gemm runs slower than the prior ops-based route and could use optimization by a CUDA specialist.

It's most likely because it's no longer hitting tensor cores 🤔 . I am not really sure what's expected there when using complex. Tensor cores do matmuls in lower precision (tf32).

mlx/backend/cuda/matmul.cpp
```diff
@@ -93,7 +95,25 @@ void gemm_and_bias(
...
+          out,
+          a,
+          b,
+          batch_shape,
+          a_batch_strides,
+          b_batch_strides,
+          alpha);
+
+      array bias_arr = astype(*bias, out.dtype(), s);
+      out = add(out, bias_arr, s);
```
awni: This is problematic here. We can't do MLX ops inside a primitive's evaluation.

Instead of doing the fallback this way, just fallback to the full addmm. So the condition could be

```
if (bias && a.dtype() != complex64)
```

awni:
Hey @Dan-Yeh this is looking good. Could you please rebase, address the latest remarks and then I will run the tests?

CC-Yeh:
@awni Done!

CC-Yeh:
Hi @awni 
Just removed some redundancies and rebased again.
Can you take a look? Thanks.

angeloskath:
Thank you @Dan-Yeh . I will do some more checks and merge tonight. Thanks for the awesome PR and for your patience.

angeloskath (approved):
> It's most likely because it's no longer hitting tensor cores 🤔 .

I enabled TF32 for the complex gemm and now it is 2x faster than before 🚀.

Thanks @Dan-Yeh for the awesome PR, I will merge after the tests clear.

PR #2572, opened 2025-09-07
Support pickling array for bfloat16

### LINKED ISSUE #795: [Feature] Support pickling array for bfloat16
Leaving this here in case someone wants to pick it up.

Dan-Yeh:
## Proposed changes

Upcast to `float32` to preserve precision.

closes #795 

## Checklist


awni:
I don't think this is the right fix. When you save an array as bfloat16 you would like to load it back as bfloat16. Also if you double the precision then it will double the storage and possibly memory requirements.

PR #2574, opened 2025-09-08
Some tweaks in cmake files

zcbenz:
* Do proper detection of Metal lib with `find_library`.
* Strip the trailing newline in `MACOS_SDK_VERSION` which adds noise to cmake output.
* Bump doctest's version so we can get rid of the `CMAKE_POLICY_VERSION_MINIMUM` hack.


awni (approved):
Nice, thanks!

PR #2577, opened 2025-09-09
Revert "[CUDA] Use GEMM with epilogue instead of AddMM"

zcbenz:
Reverts ml-explore/mlx#2569

I just realized that the tests do not actually test beta=1 and some cases are getting wrong results.


zcbenz:
Closed in favor of https://github.com/ml-explore/mlx/pull/2578.

PR #2578, opened 2025-09-09
[CUDA] Fix alpha not respected when using bias epilogue

zcbenz:
Fix a bug introduced by https://github.com/ml-explore/mlx/pull/2569, also add `addmm` tests for beta=1.


awni (approved):
Thanks for fixing that!

PR #2581, opened 2025-09-10
Fix flaky addmm tests

zcbenz:
It seems that the test is a bit flaky:

```
FAIL: test_addmm (test_blas.TestBlas.test_addmm)
  File "/home/circleci/project/python/tests/test_blas.py", line 713, in test_addmm
    self.assertTrue(mx.allclose(expected, out))
    ~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^

```

This PR adds `atol=1e-5` which all other `addmm` tests already did.

### DIFF
--- a/python/tests/test_blas.py
+++ b/python/tests/test_blas.py
@@ -702,7 +702,7 @@ class TestBlas(mlx_tests.MLXTestCase):
             b = mx.ones((5, 5))
             out = mx.addmm(a, b, a, beta=beta, alpha=alpha)
             expected = beta * a + alpha * (b @ a)
-            self.assertTrue(mx.allclose(expected, out))
+            self.assertTrue(mx.allclose(expected, out, atol=1e-5))
 
             # Broadcast c
             a = mx.ones((5, 5))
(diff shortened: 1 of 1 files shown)


awni (approved):
🙏

python/tests/test_blas.py
```diff
@@ -702,15 +702,15 @@ def test_addmm(self):
...
+            self.assertTrue(mx.allclose(expected, out, atol=1e-5))
 
             # Broadcast c
             a = mx.ones((5, 5))
             b = mx.ones((5, 5))
             c = mx.ones((1, 5))
             out = mx.addmm(c, a, b, beta=beta, alpha=alpha)
             expected = beta * c + alpha * (a @ b)
-            self.assertTrue(mx.allclose(expected, out))
+            self.assertTrue(mx.allclose(expected, out, atol=1e-5))
```
awni, line 713: It seems like this test is still flaky at `1e-5` 🤔 

See e.g. https://app.circleci.com/pipelines/github/ml-explore/mlx/6904/workflows/acc2f356-95ff-4d0f-8396-4446d200d208/jobs/32466

zcbenz: At least the good news is that it is the only flaky test case and it happens a lot, I'll check what is going wrong.

PR #2586, opened 2025-09-11
Support pickling array for bfloat16

### LINKED ISSUE #795: [Feature] Support pickling array for bfloat16

CC-Yeh:
## Proposed changes

- Store `bfloat16` data as a `uint16` NumPy array.
- Pickle the `dtype` to preserve the `bfloat16` type.

closes #795 

Note that [ml_dtype defines bfloat16](https://github.com/jax-ml/ml_dtypes/blob/main/ml_dtypes/_src/dtypes.cc#L44-L62) as a NumPy dtype, 
but integrating it natively into MLX seems like a big change.

## Checklist

### DIFF
--- a/python/src/array.cpp
+++ b/python/src/array.cpp
@@ -485,14 +485,12 @@ void init_array(nb::module_& m) {
             ND nd = nb::cast<ND>(state[0]);
             auto val = static_cast<mx::Dtype::Val>(nb::cast<uint8_t>(state[1]));
             if (val == mx::Dtype::Val::bfloat16) {
-              std::vector<size_t> shape;
-              for (size_t i = 0; i < nd.ndim(); ++i)
-                shape.push_back((size_t)nd.shape(i));
+              auto owner = nb::handle(state[0].ptr());
               new (&arr) mx::array(nd_array_to_mlx(
                   ND(nd.data(),
                      nd.ndim(),
-                     shape.data(),
-                     {},
+                     reinterpret_cast<const size_t*>(nd.shape_ptr()),
+                     owner,
                      nullptr,
                      nb::bfloat16),
                   mx::bfloat16));
--- a/python/src/convert.h
+++ b/python/src/convert.h
@@ -17,7 +17,7 @@ static constexpr dlpack::dtype bfloat16{4, 16, 1};
 }; // namespace nanobind
 
 struct ArrayLike {
-  ArrayLike(nb::object obj) : obj(obj){};
+  ArrayLike(nb::object obj) : obj(obj) {};
   nb::object obj;
 };
 


python/src/convert.cpp
```diff
@@ -161,6 +161,49 @@ nb::ndarray<NDParams...> mlx_to_nd_array(const mx::array& a) {
   }
 }
 
+mx::array nd_array_to_mlx_as_dtype(
```
awni: `nd_array_to_mlx` already takes a dtype parameter. Can we use that?

python/src/array.cpp
```diff
@@ -466,12 +466,27 @@ void init_array(nb::module_& m) {
...
       .def(
           "__iter__", [](const mx::array& a) { return ArrayPythonIterator(a); })
-      .def("__getstate__", &mlx_to_np_array)
+      .def(
+          "__getstate__",
+          [](const mx::array& a) {
+            return nb::make_tuple(
+                mlx_to_np_array(a),
+                static_cast<uint8_t>(a.dtype().val()),
+                a.dtype().size());
```
awni: I don't think we need to save the `val` and `size`. Just save the `val`.

python/src/array.cpp
```diff
@@ -466,12 +466,27 @@ void init_array(nb::module_& m) {
...
+            if (nb::len(state) != 3) {
+              throw std::invalid_argument(
+                  "Invalid pickle state: expected (ndarray, Dtype::Val, size)");
+            }
+            using ND = nb::ndarray<nb::ro, nb::c_contig, nb::device::cpu>;
+            auto nd = nb::cast<ND>(state[0]);
+            auto val = nb::cast<uint8_t>(state[1]);
+            auto size = nb::cast<uint8_t>(state[2]);
+            mx::Dtype dtype(static_cast<mx::Dtype::Val>(val), size);
+            new (&arr) mx::array(nd_array_to_mlx_as_dtype(nd, dtype));
```
awni: Then here, check the `val`. If it's `bfloat` pass in `bfloat16` as the type. Otherwise `nullopt`.

python/src/convert.cpp
```diff
@@ -149,7 +149,7 @@ nb::ndarray<NDParams...> mlx_to_nd_array(const mx::array& a) {
     case mx::float16:
       return mlx_to_nd_array_impl<mx::float16_t, NDParams...>(a);
     case mx::bfloat16:
-      throw nb::type_error("bfloat16 arrays cannot be converted to NumPy.");
+      return mlx_to_nd_array_impl<uint16_t, NDParams...>(a);
```
awni: This is slightly problematic because now MLX bfloat16 arrays will convert to numpy with e.g. `.numpy`. I don't think we want that behavior, it should still raise an exception for that.

awni: Maybe instead of doing this we can `view()` the array as uint16 before calling this function when its type is bfloat16.

python/src/array.cpp
CC-Yeh: Done, thx

python/src/array.cpp
CC-Yeh: done

python/src/convert.cpp
CC-Yeh: I first considered forcing `nd_array_to_mlx` to use the optional `dtype` whenever it was set. But `to_array` and `create_array` also call this function, and some callers may expect `dtype` inference or want to control casting themselves. To avoid breaking that, I now create a `bfloat16` view of the `ndarray` in `__setstate__` and keep all pickling logic on the pickling side.

python/src/convert.cpp
CC-Yeh: Done, thanks for the suggestion.

CC-Yeh:
@awni Could you take another look? Thanks!

python/src/convert.cpp
CC-Yeh: Improved the view logic, ready for re-review : )

PR #2591, opened 2025-09-15
Fix metal scan

awni:
Fix race condition in metal kernel. It's quite difficult to get this to show up in a unit test.

Here's a test that triggers it every so often.

```python
import mlx.core as mx

mx.random.seed(0)
int_arr = mx.random.randint(0, 4, (4000000,))
arr = mx.array(int_arr)
for i in range(200):
    print(mx.cumsum(arr))
```

And in a separate file:

```python
import subprocess

command_arr = []
for i in range(100):
    command_ = "python test_mlx_cumsum.py"
    command_arr.append(command_)

# start all programs
processes = [subprocess.Popen(command_, shell=True) for command_ in command_arr]
# wait
for process in processes:
    process.wait()
```

### DIFF
--- a/mlx/backend/metal/kernels/scan.h
+++ b/mlx/backend/metal/kernels/scan.h
@@ -306,6 +306,7 @@ template <
     U prev_thread = op.simd_exclusive_scan(values[N_READS - 1]);
 
     // Write simdgroup_sums to SM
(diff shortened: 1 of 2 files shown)


angeloskath (approved):
Great catch thanks!

PR #2600, opened 2025-09-17
Detect cache thrashing in LRUCache

zcbenz:
Add a simple detection for cache thrashing when using LRUCache, by counting if cache misses have exceeded 2 * capacity. When that happens an exception would be thrown:

> RuntimeError: Cache thrashing is happening, please set the environment variable MLX_CUDA_GRAPH_CACHE_SIZE to a larger value than 100 to fix degraded performance.


awni (approved):
Nice, thanks!!

Issue #2599, opened 2025-09-16
Make `zeros_like` work with shapeless compile / export.


josharian:
Hi! Thanks very much for mlx.

I'm trying to use shapeless exports and I'm hitting "cannot infer output shapes" for a few ops that seem like they should support shapeless use.

The first one I hit was Conv. I tried adding Conv::output_shape, and it appears to work. Should I send a PR with this, or is there some reason it isn't implemented?

The next was zeros_like. It uses Full, which doesn't implement output_shape. But zeros_like seems naively like it should be sufficiently shapeless. Should there be a FullLike? Or some other more direct way to do zeros_like and ones_like in a shapeless manner? Or am I missing something here?

Thanks!

awni:
For the convolution, yes feel free to send the PR.


`zeros_like` needs some thought. For example suppose you export / compile a function like:

```python
def fun(x)
  return zeros_like(x)
```

That should infer the output shape from the input shape. But `zeros_like` just calls `full(shape, 0)`. And that doesn't preserve the fact that `x` was an input in the graph which will break shape inference.

josharian:
Conv: Great, thanks! Will send the PR tonight or tomorrow.

zeros_like: That makes sense. I haven't dug much into the mlx internals, but maybe could there be a `full_like`, so that `zeros_like` calls through to `full_like(x, 0)`? Then Full could be straightforwardly implemented in terms of FullLike.

awni:
> zeros_like: That makes sense. I haven't dug much into the mlx internals, but maybe could there be a full_like, so that zeros_like calls through to full_like(x, 0)? Then Full could be straightforwardly implemented in terms of FullLike.

Yea something like that could work to support shapeless `zeros_like`. I think we can do it without adding a new primitive. Basically:

- `zeros_like` and `ones_like` call `full_like`
- `full_like` does a `broadcast_arrays` (which is safe for shapeless)
- Then the `full_like` calls a new `full_impl` implementation which does not broadcast

The original `zeros`, `ones` call `full` which does a `broadcast_to` and then calls the new full implementation helper.

PR #2601, opened 2025-09-17
implement Convolution::output_shape

josharian:
## Proposed changes

Implement `Convolution::output_shape` so that convolutions can be used with shapeless function exports.

Please review carefully: My C++ is not super strong, and I'm new to mlx.

The code in mlx/conv_shape.h is 100% code movement from mlx/ops.cpp, no changes aside from marking `conv_out_shape` as `inline`.

Updates #2599

## Checklist

### DIFF
--- a/mlx/primitives.cpp
+++ b/mlx/primitives.cpp
@@ -1243,6 +1243,114 @@ array conv_weight_backward_patches(
   return grad;
 }
 
+namespace {
+
+// Conv helpers
+inline int conv_out_axis_size(int in_dim, int wt_dim, int stride, int padding) {
+  return ((in_dim + padding - wt_dim) / stride) + 1;
+}
+
+// Conv helpers
+inline int dilate_size(int dim, int dil) {
+  return 1 + dil * (dim - 1);
+}
+
+} // namespace
+
+Shape Convolution::conv_out_shape(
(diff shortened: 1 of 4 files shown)


mlx/conv_shape.h
```diff
@@ -0,0 +1,121 @@
+// Copyright © 2025 Apple Inc.
```
awni: We don't want a new file for this. Instead you can make it static on the `Convolution` primitive and then just use it in the `ops`.

mlx/primitives.cpp
```diff
@@ -1454,6 +1455,47 @@ bool Convolution::is_equivalent(const Primitive& other) const {
...
+      return std::vector<int>(spatial_dims, def);
+    }
+    if (v.size() == 1) {
+      return std::vector<int>(spatial_dims, v[0]);
+    }
+    if (v.size() == spatial_dims) {
+      return v;
+    }
+    throw std::invalid_argument("Convolution param rank mismatch");
+  };
```
awni: I think these parameters should already be expanded once they are in the primitive right?

josharian:
Great feedback. Thanks! Fixed and re-tested and pushed a new commit. PTAL.

PR #2602, opened 2025-09-17
Add gated rms norm

Goekdeniz-Guelmez:
## Proposed changes

added this for Mamba2, NemotronH and Qwen3Next, using the gate_before_norm True for mamba2 and also for qwen3next.

## Checklist


awni:
I don't think we should add this to MLX core. It seems quite natural to do `self.norm(x) * gate` in the parent layer rather than making a sub-layer for this. And it's just as readable / flexible.

Goekdeniz-Guelmez:
yep, on second thought, its a logical move indeed.

PR #2606, opened 2025-09-18
Expose `mx.depends` to Python

awni:
This can be useful to manage dependencies in the graph. Using it to fix pipeline parallel prompt processing in MLX LM:


angeloskath (approved):
🙏

PR #2608, opened 2025-09-19
Avoid producing NaN in attention

awni:
As it turns out.. producing NaNs in the attention when an entire row of keys is masked is quite inconvenient for batching. In that case the NaNs can propagate which breaks everything.

The simplest fix is to avoid producing NaNs when an entire row is masked. This changes the behavior from before.

It also aligns prompt processing for the fallback fast SDPA with the fused SDPA (which already did not produce NaNs).


angeloskath (approved):
👍

PR #2609, opened 2025-09-19
Lower sorted QMM gather threshold

awni:
With GPT OSS for short prompts (256) it's much faster:

Pre: prompt_tps=480.979
Post: prompt_tps=974.984


angeloskath (approved):
👍

PR #2611, opened 2025-09-22
Allocator fix

nastya236:
Currently, in the allocator, when `CudaAllocator::free(buf)` is called, buffers with `buf->size <= small_pool_size` are still recycled into the main `buffer_cache_`. As a result, scalar allocations end up in `buffer_cache_` (if I’m reading the logic correctly). Later, when we call `CudaBuffer* buf = buffer_cache_.reuse_from_cache(size)` in `malloc`, these cached scalars may be returned from the main cache. I wasn’t sure if this is the intended behavior, so this change is a quick fix (based on my current understanding).

### DIFF
--- a/mlx/backend/cuda/allocator.cpp
+++ b/mlx/backend/cuda/allocator.cpp
@@ -1,12 +1,12 @@
 // Copyright © 2025 Apple Inc.
 
 #include "mlx/backend/cuda/allocator.h"
-#include "mlx/backend/cuda/utils.h"
-#include "mlx/utils.h"
-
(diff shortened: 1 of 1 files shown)


awni:
> As a result, scalar allocations end up in buffer_cache_ (if I’m reading the logic correctly)

You're right about that.. but it was intentional. Do you think it is causing an issue?

nastya236:
Ah, I see. No, it doesn’t cause a problem by itself, I just thought it wasn’t intentional. I’ll close this then!

Issue #2067, opened 2025-04-11
[BUG] thread issues with evaluation


acsweet:
**Describe the bug**
There seem to be some thread data access issues introduced in the newest version of mlx.

**To Reproduce**
This doesn't produce the exact same error I was experiencing, but it does produce a similar error (I think).
```python
import mlx.core as mx
import mlx.nn as nn
import mlx.optimizers as optim
import numpy as np
import threading
import time
import traceback
import faulthandler

faulthandler.enable()

class SimpleModel(nn.Module):
    def __init__(self):
        super().__init__()
        self.dense1 = nn.Linear(10, 32)
        self.dense2 = nn.Linear(32, 16)
        self.dense3 = nn.Linear(16, 1)
        
    def __call__(self, x):
        x = mx.maximum(0, self.dense1(x))
        x = mx.maximum(0, self.dense2(x))
        return self.dense3(x)

def loss_fn(model, x, y):
    pred = model(x)
    return mx.mean((pred - y) ** 2)

def train_and_convert(thread_id):
    try:
        print(f"Thread {thread_id} starting")
        
        model = SimpleModel()
        optimizer = optim.Adam(learning_rate=0.001)

        x = mx.random.normal((128, 10))
        y = mx.random.normal((128, 1))
        
        for i in range(10):
            print(f"Thread {thread_id}, Iteration {i} starting")
            loss, grads = mx.value_and_grad(loss_fn)(model, x, y)
            
            print(f"Thread {thread_id}, Iteration {i} - evaluating gradients")
            for g in grads.values():
                mx.eval(g)
            
            print(f"Thread {thread_id}, Iteration {i} - updating model")
            optimizer.update(model, grads)
            
            print(f"Thread {thread_id}, Iteration {i} - evaluating loss")
            mx.eval(loss)
            
            print(f"Thread {thread_id}, Iteration {i} - converting loss to numpy")
            # Convert loss to numpy
            np_loss = np.array(loss)
            
            print(f"Thread {thread_id}, Iteration {i} - converting gradients to numpy")
            for name, g in list(grads.items())[:2]:
                print(f"Thread {thread_id}, Iteration {i} - converting gradient {name}")
                np_grad = np.array(g)
                print(f"Thread {thread_id}, Iteration {i}, Grad {name} shape: {np_grad.shape}")
            
            print(f"Thread {thread_id}, Iteration {i}, Loss: {np_loss}")
            
            # small delay to increase chances of thread overlap
[...]

PR #2614, opened 2025-09-23
[CUDA] Make CudaEvent work with multi-device

zcbenz:
Most APIs of CUDA event have requirements on being used on the same device with the device that the event was created on, this PR makes sure the `CudaEvent` class set current device before invoking the APIs.

There is also an issue that the `CudaEvent` could be destroyed in worker thread when used for waiting on a GPU work in CPU stream, which would cause race conditions when accessing event cache, this PR uses a simple strategy by skipping the cache when event is created or destroyed on worker threads, so we don't have to add thread-safety overhead for those rare cases.

Note that this PR does not guarantee multi-thread or multi-device safety, this is mostly a defensive change.

### DIFF
--- a/mlx/backend/cuda/device.cpp
+++ b/mlx/backend/cuda/device.cpp
@@ -68,8 +68,8 @@ Device::~Device() {
 
 void Device::make_current() {
   // We need to set/get current CUDA device very frequently, cache it to reduce
-  // actual calls of CUDA APIs. This function assumes single-thread in host.
-  static int current = 0;
+  // actual calls of CUDA APIs.
+  static thread_local int current = 0;
   if (current != device_) {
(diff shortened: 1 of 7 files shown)


zcbenz:
> when will events be created and destroyed in a worker thread

Waiting a CPU stream is implemented by copying the event to the CPU thread, so the event could be destroyed there:

https://github.com/ml-explore/mlx/blob/7a6adda1e6538d4f2a70c79511c26c1e5b78b4bf/mlx/backend/cuda/event.cu#L91-L97

Another case is user may run `eval()` in workers, I did that to implement `await eval()` in node-mlx, and I also saw some users doing it: https://github.com/ml-explore/mlx/issues/2067

PR #2618, opened 2025-09-23
Implement probability distribution support via mlx.distributions

shaeantoine:
## Proposed changes

This PR includes the beginning components to support probability distributions within MLX via python.mlx.distributions. Inspired by Jax and PyTorch the component implements key probability distribution methods namely:

- sample()
- log_prob()
- entropy()
- mean()
- variance()

for the normal distribution. It also includes composition utilities like Independent and TransformedDistribution methods. It does so while following MLX's functional paradigm and JIT-first approach. I've also included several tests and pieces of documentation in line with the nn module as much as possible. 

This PR, more than anything, is to validate upon review this approach to an MLX probability distribution component. Before then adding additional distributions, more complex composition patterns and additional utilities. 

## Checklist

- [ x] I have read the [CONTRIBUTING](https://github.com/ml-explore/mlx/blob/main/CONTRIBUTING.md) document
- [x ] I have added tests that prove my fix is effective or that my feature works

### DIFF
--- a/docs/src/python/distributions/composition.rst
+++ b/docs/src/python/distributions/composition.rst
@@ -1,0 +1,16 @@
+.. _distributions_composition:
+
+.. currentmodule:: mlx.distributions
+
+Composition Utilities
+---------------------
+
+These are higher-order functions that create new distributions by composing or
+modifying existing ones.
+
+.. autosummary::
+   :toctree: _autosummary_composition
+   :template: nn-module-template.rst
+
+   Independent
+   TransformedDistribution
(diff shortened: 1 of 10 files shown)


awni:
Thanks for the contribution and this looks nicely done!

However, my suggestions is to develop this a third party package as there is no clear reason to put it in MLX core. I haven't seen anyone asking for this..so I don't think we should add it to MLX core until there are more active use cases for it.

See e.g. https://github.com/ml-explore/mlx/pull/1275 which we closed for similar reasons in the past.

Issue #2585, opened 2025-09-11
JIT compilation prevents optimization of GEMM with bias


zcbenz:
Consider following code, which is partial code of model training:

```python
from functools import partial

import mlx.core as mx
import mlx.nn as nn
from mlx.utils import tree_map

model = nn.Linear(1024, 3072, bias=True)
params = model.trainable_parameters()
mx.eval(params)

state = [mx.random.state]

@partial(mx.compile, inputs=state, outputs=state)
def step(x, params):
    model.update(tree_map(lambda x: x.astype(mx.bfloat16), params))
    return model(x.astype(mx.float16))

x = mx.random.uniform(shape=(4,512,1024), dtype=mx.bfloat16)
y = step(x, params)
mx.eval(y)
```

Which converts the Linear module's weights and bias from f32 to bf16, broadcasts the bias to (4,512,3072), and then calls `addmm`. And after JIT compilation the `astype` and `broadcast` would be fused together.

The problem is that with JIT compilation we would lose some optimization opportunities: if we remove the `mx.compile` line, the `AddMM::eval_gpu` would receive the bias input with data size of 3072, from which we know it is a bias vector to be added to the last dim of output, and can use a faster kernel (https://github.com/ml-explore/mlx/pull/2569). However with JIT compilation the bias input's data size becomes 6291456 (4 *512 *3072) because of op fusion, and we can no longer take the same optimization.

I can think of 2 solutions:

1. Add bias epilogue as part of `Matmul` primitive and make `addmm` redirect to it. The downside is it would be relatively large change and we would also need to apply the same optimization to Metal kernels.
2. Move the broardcasting and reshaping of bias input from `addmm` op to `AddMM::eval_gpu`, the change would be minimal and could also be a bit ugly.

And with both solutions I think the `vjp`/`jvp` would be a bit tricky to get right.

awni:
That's a very interesting observation. It seems suboptimal to me that we fuse broadcast when it's the final op.

For example, if you have a chain of elementwise ops that end with a broadcast:

```python
import time
import mlx.core as mx

@mx.compile
def fun(x):
    return mx.broadcast_to(mx.sin(mx.exp(mx.log(mx.abs(x)))), (16384, 2048, 16))

x = mx.array([1.0])
for _ in range(10):
    x = fun(x)
mx.eval(x)

tic = time.time()
x = mx.array([1.0])
for _ in range(10):
    x = fun(x)
mx.eval(x)
toc = time.time()
[...]

PR #2622, opened 2025-09-25
Compiled should not end in broadcast

angeloskath:
Very good point in #2585 and very easy fix.

We go from

cgraph-old

to 

cgraph

for

graph

### DIFF
--- a/mlx/compile.cpp
+++ b/mlx/compile.cpp
@@ -727,7 +727,7 @@ void compile_fuse(
       }
     };
 
-    if (arr.has_primitive()) {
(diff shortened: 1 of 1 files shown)


mlx/compile.cpp
```diff
@@ -727,7 +727,7 @@ void compile_fuse(
       }
     };
 
-    if (arr.has_primitive()) {
+    if (arr.has_primitive() && !is_broadcast(arr.primitive())) {
```
zcbenz, line 734: Can you add a comment for this line?

angeloskath: Very good point.

PR #2631, opened 2025-09-28
Fix for max block dim

### LINKED ISSUE #2628: [BUG] [CUDA] Compiled kernel fails with resource issue
```python
import mlx.core as mx

inputs = [mx.zeros((16384,)).astype(mx.float16) for _ in range(8)]

@mx.compile
def fun(inputs):
    a = inputs[0] + inputs[1]
    b = inputs[2] + inputs[3]
    c = inputs[4] + inputs[5]
    d = inputs[6] + inputs[7]
    return a * b * c * d

mx.eval(inputs)
mx.eval(fun(inputs))
```

Fails with:

```
RuntimeError: cudaGraphInstantiate(&handle_, graph, nullptr, nullptr, 0) failed: too many resources requested for launch
  what():  cudaGraphInstantiate(&handle_, graph, nullptr, nullptr, 0) failed: too many resources requested for launch
Aborted (core dumped)
```

awni:
Uses the occupancy size as the block dim for compiled kernels to avoid using too large of a block dim.

Closes #2628

### DIFF
--- a/mlx/backend/cuda/compiled.cpp
+++ b/mlx/backend/cuda/compiled.cpp
@@ -332,9 +332,9 @@ void Compiled::eval_gpu(
     encoder.set_output_array(out);
   }
 
-  auto kernel = mod.get_kernel(kernel_name);
+  auto [kernel, max_block_dims] = mod.get_kernel_and_dims(kernel_name);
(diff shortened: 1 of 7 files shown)


angeloskath (approved):
🙏

PR #2633, opened 2025-09-30
Fix: [BUG] Compiled functions return the same number of outputs and pre-commit bump

### LINKED ISSUE #2632: [BUG] Compiled functions return the same number of outputs and types
As title and the following snippet

```python
import mlx.core as mx

def foo(x, y):
    if y is None:
        return x + 1, x
    else:
        return x, None

cfoo = mx.compile(foo)
x = mx.random.normal((128,))
print(cfoo(x, 0))    # (array(...), None)
print(cfoo(x, None)) # (array(...), None)

# Called the other way around even results in segfault
```

NripeshN:
Closes #2632

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,16 +1,16 @@
 repos:
 -   repo: https://github.com/pre-commit/mirrors-clang-format
-    rev: v19.1.7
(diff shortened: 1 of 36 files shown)


angeloskath:
Closing in favor of #2634 .

PR #2634, opened 2025-09-30
Compile now can attach arbitrary data to an entry

### LINKED ISSUE #2632: [BUG] Compiled functions return the same number of outputs and types

angeloskath:
Fixes #2632 .

The only way of fixing #2632 without replicating the compilation rules in python is to attach data on the compiler cache entry.

The only ugly thing I see is that `compile` from C++ will have this horrible double wrapping. If people think it is too ugly or a performance consideration I can try to remove it.

### DIFF
--- a/mlx/compile.cpp
+++ b/mlx/compile.cpp
@@ -296,6 +296,7 @@ class CompilerCache {
     std::vector<array> tape;
     bool empty{true};
     std::vector<uint64_t> constants;
+    std::shared_ptr<void> extra;
   };
 
   // Returns a reference to a CacheEntry which can be updated
@@ -376,8 +377,9 @@ CompilerCache& compiler_cache() {
   return compiler_cache_;
 }
 
-std::pair<std::vector<array>, std::vector<array>> compile_trace(
-    const std::function<std::vector<array>(const std::vector<array>&)>& fun,
+std::tuple<std::vector<array>, std::vector<array>, std::shared_ptr<void>>
+compile_trace(
+    const ArrayFnWithExtra& fun,
     const std::vector<array>& inputs,
     bool shapeless) {
   // Set the global tracing flag.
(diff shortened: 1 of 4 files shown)


mlx/compile.cpp
```diff
@@ -991,8 +995,38 @@ std::function<std::vector<array>(const std::vector<array>&)> compile(
 
     // At this point we must have a tape, now replace the placeholders
     // with real arrays that can be evaluated
-    return compile_replace(
-        entry.tape, entry.inputs, entry.outputs, inputs, shapeless);
+    return std::pair<std::vector<array>, std::shared_ptr<void>>{
```
zcbenz:
```suggestion
    return ArraysAndExtra{
```

mlx/compile.cpp
```diff
@@ -991,8 +995,38 @@ std::function<std::vector<array>(const std::vector<array>&)> compile(
...
+  }
+  if (!fun) {
+    throw std::invalid_argument(
+        "[compile] Cannot compile a function without a target.");
+  }
+
+  ArrayFnWithExtra fun_with_extra =
+      [fun = std::move(fun)](const std::vector<array>& inputs) {
+        return std::pair<std::vector<array>, std::shared_ptr<void>>{
+            fun(inputs), nullptr};
```
zcbenz:
```suggestion
        return ArraysAndExtra{fun(inputs), nullptr};
```

zcbenz (approved):
Nice fix!

python/src/transforms.cpp
```diff
@@ -556,7 +563,6 @@ struct PyCompiledFun {
   ~PyCompiledFun() {
     nb::gil_scoped_acquire gil;
 
-    tree_cache().erase(fun_id);
```
awni, line 559: It's nice to get rid of that cache for the output structure.

PR #2641, opened 2025-10-01
fix cross entropy axis param

awni:
Minor fix to use axis parameter correctly in `losses.cross_entropy_loss`.

### DIFF
--- a/python/mlx/nn/losses.py
+++ b/python/mlx/nn/losses.py
@@ -86,7 +86,9 @@ def cross_entropy(
     if targets_as_probs:
         score = mx.sum(logits * targets, axis=axis)
     else:
-        score = mx.take_along_axis(logits, targets[..., None], axis).squeeze(-1)
+        score = mx.take_along_axis(logits, mx.expand_dims(targets, axis), axis).squeeze(
+            axis
+        )
 
     logsumexp_logits = mx.logsumexp(logits, axis=axis)
     if label_smoothing > 0:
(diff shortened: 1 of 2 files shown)


awni:
Also added a speedup for gradient clipping:

Pre: 3.96
Post: 2.57

Benchmark:

```python
import time
import mlx.core as mx
import mlx.optimizers as opt

grads = [mx.random.uniform(shape=(2048, 2048)) for _ in range(10)]

def fn(grads):
    for _ in range(10):
        grads, norm = opt.clip_grad_norm(grads, 1.0)
    return grads

for _ in range(100):
    mx.eval(fn(grads))

tic = time.time()
for _ in range(100):
    mx.eval(fn(grads))
toc = time.time()
print(toc - tic)
```

python/mlx/optimizers/optimizers.py
```diff
@@ -971,10 +971,6 @@ def clip_grad_norm(grads, max_norm):
...
     norm_squared = tree_reduce(lambda acc, g: acc + g.square().sum(), grads, 0.0)
     total_norm = mx.sqrt(norm_squared)
-    normalizer = max_norm / (total_norm + 1e-6)
-
-    def clipper(g):
-        return mx.where(total_norm < max_norm, g, g * normalizer)
-
-    clipped_grads = tree_map(clipper, grads)
+    normalizer = mx.minimum(max_norm / (total_norm + 1e-6), 1.0)
+    clipped_grads = tree_map(lambda g: g * normalizer, grads)
```
angeloskath, line 975: Why do a multiplication when you can do a multiplication and a useless conditional... 🙈

angeloskath (approved):
👍

PR #2644, opened 2025-10-02
Faster triu, tril, where with scalar

awni:
- Speed up triu, tril
- Add specializations for `mx.where(cond, x, scalar)` and `mx.where(cond, scalar, x)` which are pretty common

Benchmark triu:
Pre: 0.52
Post: 0.194


angeloskath (approved):
Nice!

awni:
Those memory tests on cuda have been flaky as of late.. 🤔

PR #2645, opened 2025-10-02
Configure CMake to export `compile_commands.json`

andportnoy:
This helps enable LSP for code navigation using clangd.


awni (approved):
LGTM

PR #2649, opened 2025-10-03
Speed up compile for node with many parents

### LINKED ISSUE #2643: [BUG] Compile is slow from scalars
There is an N^2 computation where `N` is the number of parents in `compile_simplify`.

So something like the following takes forever to compile because the scalar `0.0` has ~40k parents.

```python
import mlx.core as mx

@mx.compile
def fun(x):
    for _ in range(40000):
        x = x + mx.array(0.0)
    return x

x = mx.array(1.0)
print(fun(x))
```

awni:
Closes #2643

The following benchmark:

```python
import mlx.core as mx

@mx.compile
def fun(x):
    for _ in range(40000):
        x = x + mx.array(0.0)
    return x

x = mx.array(1.0)
print(fun(x))
```

Pre: ~60 seconds
Post: < 1 second

Compiling 100-layer MNIST:

Pre:    Time  33.59 (ms)
Post:  Time 31.177 (ms)

### DIFF
--- a/mlx/compile.cpp
+++ b/mlx/compile.cpp
@@ -460,6 +460,24 @@ std::pair<std::vector<array>, ParentsMap> compile_dfs(
   return {tape, parents_map};
 }
 
+static inline uint64_t splitmix64(uint64_t x) noexcept {
+  x += 0x9e3779b97f4a7c15ull;
+  x = (x ^ (x >> 30)) * 0xbf58476d1ce4e5b9ull;
+  x = (x ^ (x >> 27)) * 0x94d049bb133111ebull;
+  return x ^ (x >> 31);
+}
+
+struct VecU64Hash {
+  size_t operator()(const std::vector<uint64_t>& s) const noexcept {
+    uint64_t h =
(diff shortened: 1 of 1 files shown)


mlx/compile.cpp
```diff
@@ -460,6 +460,24 @@ std::pair<std::vector<array>, ParentsMap> compile_dfs(
...
+struct VecU64Hash {
+  size_t operator()(const std::vector<uint64_t>& s) const noexcept {
+    uint64_t h =
+        0x243f6a8885a308d3ull ^ (uint64_t)s.size() * 0x9e3779b97f4a7c15ull;
+    for (uint64_t x : s) {
+      h = splitmix64(x ^ splitmix64(h + 0x9e3779b97f4a7c15ull));
+    }
+    return (size_t)h;
+  }
+};
```
awni, line 479: I got this hash from chat.. seems reasonable..

awni:
Also closes https://github.com/ml-explore/mlx-lm/issues/516

PR #2652, opened 2025-10-05
Check isnan in maximum / minimum with CPU backend

aisk:
## Proposed changes

## Checklist

### DIFF
--- a/mlx/backend/cpu/simd/accelerate_simd.h
+++ b/mlx/backend/cpu/simd/accelerate_simd.h
@@ -207,13 +207,21 @@ Simd<T, N> atan2(Simd<T, N> a, Simd<T, N> b) {
 
 template <typename T, int N>
 Simd<T, N> maximum(Simd<T, N> a, Simd<T, N> b) {
-  // TODO add isnan
+  if constexpr (!std::is_integral_v<T>) {
+    auto a_is_nan = isnan(a);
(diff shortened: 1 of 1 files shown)


awni:
Could you add a test for this so we know it's working as expected?

aisk:
@aisk tests added.

tests/ops_tests.cpp
```diff
@@ -1060,6 +1060,30 @@ TEST_CASE("test reduction ops") {
...
     CHECK(array_equal(max(x, 0), array({4.0f, NAN, 6.0f}), true).item<bool>());
     CHECK(array_equal(max(x, 1), array({NAN, 6.0f}), true).item<bool>());
+
+    // Test maximum and minimum with NaN values
+    x = array({1.0f, NAN, 3.0f});
+    auto y = array({NAN, 2.0f, 1.0f});
+    auto max_result = maximum(x, y);
+    auto min_result = minimum(x, y);
+    CHECK(array_equal(max_result, array({NAN, NAN, 3.0f}), true).item<bool>());
+    CHECK(array_equal(min_result, array({NAN, NAN, 1.0f}), true).item<bool>());
```
awni: None of these tests will hit the simd implementation because it will only be used for 8 or more elements. So could you add a tests with larger vectors, say 8 or 16?

PR #2656, opened 2025-10-07
[CUDA] Fix jit file cache for large kernel names

angeloskath:
Split the name into directories for large kernel names to overcome the 255 character limit on filenames. The win32 limit is probably useless but I just lifted the limit directly from the CPU compile.


zcbenz (approved):
Didn't know there is a length limitation on filenames 🤯, nice fix!

PR #2657, opened 2025-10-08
Fix fast synch when fence is waited before a command buffer is created

awni:
As title.

I didn't add a test for this because we don't test the fast synch yet (maybe we should?)..


awni:
The test case would be something like:
```python
import mlx.core as mx

def fun():
    return mx.squeeze(mx.array([1.0]), stream=mx.cpu) + mx.array(1.0)

mx.synchronize() # should commit active command buffers
print(fun())
```

PR #2663, opened 2025-10-10
Add Masked Scatter

### LINKED ISSUE #2149: Masked scatter update
It would be very convenient to have a scatter update which takes a boolean mask. This is a common use case in for example VLMs where you update one vector of embeddings with another array of image embeddings with the update locations based on a mask.

Something like:

```
a[mask] = b
```

That might be complex if we wanted to allow mixing boolean and non-boolean indices. But maybe it's fine to allow only a single array index if it's boolean.

Or maybe more like the torch API [masked_scatter](https://pytorch.org/docs/stable/generated/torch.Tensor.masked_scatter.html)

CC-Yeh:
## Proposed changes

Supports masked scatter via indexing.

```python
self[mask] = src
```

For the CPU path: iterate with contiguous iterators over `mask`/`out` (and per-batch when `vmap_axis` == 0), copying the next `src` element whenever `mask` is true.

For the GPU path: It first computes `scatter_offsets` with `cumsum`, then use `scatter_offsets` + `mask` to do the assignment in a second kernel.

Closes #2149

Benchmarks results On Apple M4 chip.

Apple_M4_masked_scatter_float32

## Checklist

### DIFF
--- a/tests/ops_tests.cpp
+++ b/tests/ops_tests.cpp
@@ -7,6 +7,7 @@
 
 #include "doctest/doctest.h"
 
+#include "mlx/backend/cuda/cuda.h"
 #include "mlx/mlx.h"
 
 using namespace mlx::core;
@@ -2422,6 +2423,11 @@ TEST_CASE("test scatter") {
 }
 
 TEST_CASE("test masked_scatter") {
+  if (cu::is_available()) {
+    INFO("Skipping masked_scatter cuda ops tests");
+    return;
+  }
+
   // Wrong mask dtype
   CHECK_THROWS(masked_scatter(array({1, 2}), array({1, 2}), array({1, 2})));
 


CC-Yeh:
I have a couple of questions:

- Can a primitive call other primitives? If not, what’s the best way to reuse kernels/primitives? I saw synchronization issues when launching two kernels inside the masked_scatter Metal primitive (in test_vmap, I need to call eval first to get the correct output).

- What’s the best way for a Metal kernel to report errors or set an error status? I’d like to detect out-of-bounds (OOB) accesses.

- The speedup looks a bit too good to be true—could someone help verify it?

I’m new to autograd and vmap. Please let me know if anything needs to be fixed.

awni:
> Can a primitive call other primitives? If not, what’s the best way to reuse kernels/primitives? I saw synchronization issues when launching two kernels inside the masked_scatter Metal primitive (in test_vmap, I need to call eval first to get the correct output).

You can't call other ops inside the `eval_gpu` (or `eval_cpu`) because that's just building a graph but you want to actually run the graph in the eval. Here's is [an example](https://github.com/ml-explore/mlx/blob/main/mlx/backend/metal/fft.cpp#L399-L403) of how to actually do an operation inside an eval. It's a little bit more complicated (and not every op has a back-end API like that so you may need to make one). Please ask follow-up questions if it's confusing.

> What’s the best way for a Metal kernel to report errors or set an error status? I’d like to detect out-of-bounds (OOB) accesses.

There is a way to [log errors](https://developer.apple.com/documentation/metal/logging-shader-debug-messages?language=objc) and then we would check the error buffer in the completion handler. We don't do this right now..and I would not add it in this PR for simplicity. But it would be a really nice follow on PR to add that functionality in general.

> The speedup looks a bit too good to be true—could someone help verify it?

Will check.

python/src/ops.cpp
```diff
@@ -4054,6 +4054,37 @@ void init_ops(nb::module_& m) {
                of metadata to be saved. The values can be a scalar or 1D
                obj:`array`, a :obj:`str`, or a :obj:`list` of :obj:`str`.
       )pbdoc");
+  m.def(
```
awni: What about doing this with `a[mask] = x` instead of making it a function?

mlx/backend/metal/indexing.cpp
```diff
@@ -641,4 +641,113 @@ void ScatterAxis::eval_gpu(const std::vector<array>& inputs, array& out) {
...
+      batched && src.ndim() > 0 && src.shape(0) == static_cast<int>(batch);
+  const size_t src_block =
+      src_batched && batch > 0 ? (src_flat.size() / batch) : src_flat.size();
+  const size_t src_capacity = src_batched ? src_block : src_flat.size();
+
+  array idx_mask(flat_shape, uint32, nullptr, {});
+  idx_mask.set_data(allocator::malloc(idx_mask.nbytes()));
+
+  Scan cumsum(stream(), Scan::Sum, 0, false, false);
+  cumsum.eval_gpu({mask_flat}, idx_mask);
```
awni: Yea exactly, you shouldn't do the eval here. Instead we'll need a way to call the back-end scan implementation.

mlx/backend/cpu/indexing.cpp
```diff
@@ -747,4 +747,146 @@ void ScatterAxis::eval_cpu(const std::vector<array>& inputs, array& out) {
   });
 }
 
+template <typename T>
+void masked_scatter_impl(
+    const array& mask,
+    const array& src,
+    array& out,
+    int vmap_axis) {
```
leviliangtw: [Readability Review] I'd like to implement it with two additional subfunctions `masked_scatter_impl_batch` and `masked_scatter_impl_nonbatch`, and the helper `scatter_once`.

mlx/primitives.cpp
```diff
@@ -4330,6 +4330,107 @@ bool ScatterAxis::is_equivalent(const Primitive& other) const {
...
 
+std::vector<array> MaskedScatter::vjp(
+    const std::vector<array>& primals,
+    const std::vector<array>& cotangents,
+    const std::vector<int>& argnums,
+    const std::vector<array>&) {
+  auto& s = stream();
+  const array& self = primals[0];
+  const array& mask = primals[1];
+  const array& src = primals[2];
```
leviliangtw, line 4359: [Readability Review] It would be good to have enums for them.

mlx/primitives.cpp
```diff
@@ -4330,6 +4330,107 @@ bool ScatterAxis::is_equivalent(const Primitive& other) const {
...
+  const array mask_b = broadcast_to(mask, self.shape(), s);
+  const array& cotan = cotangents[0];
+
+  std::vector<array> vjps;
+  vjps.reserve(argnums.size());
+
+  for (int arg : argnums) {
+    if (arg == 0) {
```
leviliangtw, line 4367: [Readability Review] Then the `0` here could be replaced with enums.

mlx/primitives.cpp
```diff
@@ -4330,6 +4330,107 @@ bool ScatterAxis::is_equivalent(const Primitive& other) const {
...
+    const std::vector<array>& primals,
+    const std::vector<array>& cotangents,
+    const std::vector<int>& argnums,
+    const std::vector<array>&) {
+  auto& s = stream();
+  const array& self = primals[0];
+  const array& mask = primals[1];
+  const array& src = primals[2];
+  const array mask_b = broadcast_to(mask, self.shape(), s);
+  const array& cotan = cotangents[0];
```
leviliangtw, line 4361: [Readability Review] I understood it might be an interface convention, but I'd like to use `cotangents.front()` here with a brief comment.

mlx/primitives.cpp
```diff
@@ -4330,6 +4330,107 @@ bool ScatterAxis::is_equivalent(const Primitive& other) const {
...
+
+  std::vector<array> vjps;
+  vjps.reserve(argnums.size());
+
+  for (int arg : argnums) {
+    if (arg == 0) {
+      vjps.push_back(where(mask_b, zeros_like(cotan, s), cotan, s));
+    } else if (arg == 2) {
```
leviliangtw, line 4369: [Readability Review] Same as above.

mlx/primitives.cpp
```diff
@@ -4330,6 +4330,107 @@ bool ScatterAxis::is_equivalent(const Primitive& other) const {
...
+  auto broadcast_mask = broadcast_to(mask, self.shape(), stream());
+  auto self_front = moveaxis(self, to_ax, 0, stream());
+  auto mask_front = moveaxis(broadcast_mask, to_ax, 0, stream());
+  auto src_front = moveaxis(src, to_ax, 0, stream());
+
+  auto scattered_front = array(
+      self_front.shape(),
+      self_front.dtype(),
+      std::make_shared<MaskedScatter>(stream(), /*vmap_axis=*/0),
+      {self_front, mask_front, src_front});
```
leviliangtw: [Readability Review] Here, you added `/*vmap_axis=*/` for the `0` argument, and I like it. 
It would be great if you did the same thing for lines like 4420: `moveaxis(self, /*source=*/to_ax, /*destination=*/0, stream())`. 
However, if calling numpy APIs like that is intuitive for you, then it's fine.

python/src/ops.cpp
CC-Yeh: Done, Added support for `mlx` and `numpy` boolean arrays inputs.

What do you think about keeping `masked_scatter` as well?

mlx/backend/metal/indexing.cpp
CC-Yeh: Extracted the logic into `scan_gpu` and added a donation flag to control whether the input may be donated.

python/src/ops.cpp
awni: Great! I don't think we need both. It's harder to remove an API once you add it so my preference is to hold off on adding this until we see an explicit need for it.

awni:
@Dan-Yeh if this is ready for review, let's move it out of draft. Also could you please rebase and resolve the conflicts? Then I can run the tests.

python/src/ops.cpp
CC-Yeh: make sense, removed.

CC-Yeh:
Ready for review! 

@awni & @angeloskath

tests/ops_tests.cpp
```diff
@@ -2422,6 +2422,44 @@ TEST_CASE("test scatter") {
...
+      array({1, 2, 3, 4}, {2, 2}),
+      array({false, true, true, false}, {4, 1}),
+      array({1, 2})));
+
+  // 1D mask
+  {
+    auto self = zeros({4}, int32);
+    auto mask = array({true, true, false, true});
+    auto source = array({1, 2, 4});
+    auto out = masked_scatter(self, mask, source, Device::cpu);
```
awni: Why use the cpu device here only?

awni: (And same for the following tests)

mlx/backend/metal/indexing.cpp
```diff
@@ -641,4 +643,154 @@ void ScatterAxis::eval_gpu(const std::vector<array>& inputs, array& out) {
   compute_encoder.dispatch_threads(grid_dims, group_dims);
 }
 
+void MaskedScatter::eval_gpu(const std::vector<array>& inputs, array& out) {
+  const array& self = inputs[0];
```
awni: This is a nit but I think it makes sense to change the name `self` to `out` (or `dst` or something similar). The term `self` is a bit overloaded from Python so it's slightly confusing.

mlx/backend/metal/indexing.cpp
```diff
@@ -641,4 +643,154 @@ void ScatterAxis::eval_gpu(const std::vector<array>& inputs, array& out) {
...
+
+  const size_t inner = batched ? total / batch : total;
+
+  const bool src_batched =
+      batched && src.ndim() > 0 && src.shape(0) == static_cast<int>(batch);
+  const size_t src_size = src.size();
+  if (src_batched && (src_size % batch != 0)) {
+    throw std::invalid_argument(
+        "[MaskedScatter::eval_gpu] src.size() must be divisible by batch when src is batched.");
+  }
```
awni: Basically all of the error checking here (and any that you do in `eval_cpu`) should be done in the operation and not in the primitive.

In general we want to avoid throwing errors during graph eval because we can't recover from them. And it's much better to throw those errors early during the graph building.

mlx/backend/metal/indexing.cpp
```diff
@@ -641,4 +643,154 @@ void ScatterAxis::eval_gpu(const std::vector<array>& inputs, array& out) {
...
+  const Shape flat_shape{static_cast<ShapeElem>(total)};
+  array mask_flat = reshape_in_eval(mask, flat_shape, s);
+
+  // Prefix (exclusive) of mask → scatter_offsets
+  array scatter_offsets(flat_shape, uint32, nullptr, {});
+  scatter_offsets.set_data(allocator::malloc(scatter_offsets.nbytes()));
+
+  scan_gpu(
+      mask_flat,
+      scatter_offsets,
```
awni: the `scan_gpu` will set the data of the output. So I think you don't want to set it here as it will be a superfluous call to `malloc`. Or change the `scan_gpu` to `scan_gpu_inplace`

mlx/backend/metal/indexing.cpp
```diff
@@ -641,4 +643,154 @@ void ScatterAxis::eval_gpu(const std::vector<array>& inputs, array& out) {
...
+    if (v > std::numeric_limits<uint32_t>::max())
+      throw std::overflow_error(
+          "[MaskedScatter::eval_gpu] uint64 size not implemented.");
+    return static_cast<uint32_t>(v);
+  };
+  set_u32(as_u32(src_size), bind_idx++);
+  set_u32(as_u32(total), bind_idx++);
+  set_u32(is_batched_run ? as_u32(inner) : 0u, bind_idx++);
+  set_u32(src_has_batches ? as_u32(src_block) : 0u, bind_idx++);
+  set_u32(is_batched_run ? as_u32(src_capacity) : 0u, bind_idx++);
```
awni: Nit: I think the `set_u32` is unnecessary. Cleaner to just call `compute_encoder.set_bytes(...) ` directly.

mlx/backend/metal/indexing.cpp
```diff
@@ -641,4 +643,154 @@ void ScatterAxis::eval_gpu(const std::vector<array>& inputs, array& out) {
...
+  if (src_batched && (src_size % batch != 0)) {
+    throw std::invalid_argument(
+        "[MaskedScatter::eval_gpu] src.size() must be divisible by batch when src is batched.");
+  }
+
+  const size_t src_block = src_batched ? (src_size / batch) : src_size;
+  const size_t src_capacity = src_block;
+
+  const Shape flat_shape{static_cast<ShapeElem>(total)};
+  array mask_flat = reshape_in_eval(mask, flat_shape, s);
```
awni: Nit: There is a `flatten_in_eval`

tests/ops_tests.cpp
CC-Yeh: Thanks for the catch!

These were implemented first with cpu implementation only, forgot to remove.

mlx/backend/metal/indexing.cpp
```diff
@@ -641,4 +643,154 @@ void ScatterAxis::eval_gpu(const std::vector<array>& inputs, array& out) {
...
+  set_u32(src_has_batches ? as_u32(src_block) : 0u, bind_idx++);
+  set_u32(is_batched_run ? as_u32(src_capacity) : 0u, bind_idx++);
+
+  // Dispatch
+  auto group_dims = get_block_dims(total, 1, 1);
+  MTL::Size grid_dims(total, 1, 1);
+  compute_encoder.dispatch_threads(grid_dims, group_dims);
+
+  // Lifetimes
+  d.add_temporaries({mask_flat, scatter_offsets}, s.index);
```
awni: I think we only want to store `mask_flat` as a temporary if there was a copy done. So it is probably worth checking that the `mask_flat` data pointer is different than the the `mask` data pointer and adding it if so.

mlx/primitives.cpp
```diff
@@ -4348,6 +4348,107 @@ bool ScatterAxis::is_equivalent(const Primitive& other) const {
...
+          multiply(cotan_flat, astype(mask_flat, cotan_flat.dtype(), s), s);
+
+      array gsrc_flat =
+          zeros({static_cast<int>(src.size())}, cotan_src.dtype(), s);
+      if (src.size() > 0) {
+        const int n = static_cast<int>(idx_src.size());
+        const array idx_vec = reshape(idx_src, {n}, s);
+        const array cotan_updates = reshape(cotan_src, {n, 1}, s);
+        gsrc_flat = scatter_add(gsrc_flat, idx_vec, cotan_updates, 0, s);
+      }
```
awni, line 4384: Can this be done more simply / efficiently with a gather?

mlx/primitives.cpp
```diff
@@ -4348,6 +4348,107 @@ bool ScatterAxis::is_equivalent(const Primitive& other) const {
...
+  for (int arg : argnums) {
+    if (arg == 0) {
+      out = where(mask_b, out, tangents[0], s);
+    } else if (arg == 2) {
+      out = masked_scatter(out, mask_b, tangents[1], s);
+      out = array(
+          self.shape(),
+          tangents[1].dtype(),
+          std::make_shared<MaskedScatter>(stream()),
+          {out, mask_b, tangents[1]});
```
awni: I'm confused by this. It looks like you are applying the same scatter twice? Am I missing something?

mlx/primitives.cpp
```diff
@@ -4348,6 +4348,107 @@ bool ScatterAxis::is_equivalent(const Primitive& other) const {
...
+    } else {
+      throw std::invalid_argument("[masked_scatter] invalid arg index in JVP");
+    }
+  }
+  return {out};
+}
+
+std::pair<std::vector<array>, std::vector<int>> MaskedScatter::vmap(
+    const std::vector<array>& inputs,
+    const std::vector<int>& axes) {
```
awni, line 4428: The vmap of a masked scatter should ideally be a masked scatter with appropriatley transformed inputs. E.g you may need to broadcast / repeat the mask. We should aim to remove the vmap axis from the primitive as I don't think it's necessary. Instead try to do the vmap using a regular masked scatter.

awni (changes requested):

awni:
The results are looking quite impressive! Sorry for the delay in getting it reviewed. I left a few comments. Could you please take a look and address when you have time?

CC-Yeh:
> The results are looking quite impressive! Sorry for the delay in getting it reviewed. I left a few comments. Could you please take a look and address when you have time?

Thanks for the great suggestions!

Will take care of it shortly.

mlx/primitives.cpp
```diff
@@ -1317,15 +1317,15 @@ Shape Convolution::conv_out_shape(
 
     if (pads_lo[i - 1] < 0 || pads_hi[i - 1] < 0) {
       std::ostringstream msg;
-      msg << "[conv] Padding sizes must be non-negative." << " Got padding "
-          << pads_lo << " | " << pads_hi << ".";
+      msg << "[conv] Padding sizes must be non-negative."
+          << " Got padding " << pads_lo << " | " << pads_hi << ".";
```
angeloskath: This is a total nitpick but since it's in the diff would you mind fixing it more correctly by joining the string literals?

mlx/primitives.cpp
CC-Yeh: Removed `vmap_axis_`, and added default batch for general case.
This really simplified the implementation, thanks!

mlx/primitives.cpp
CC-Yeh: Forgot to remove, thanks for the catch.

mlx/backend/metal/indexing.cpp
CC-Yeh: Done, thanks

mlx/backend/metal/indexing.cpp
CC-Yeh: done

mlx/backend/metal/indexing.cpp
CC-Yeh: Removed

mlx/backend/metal/indexing.cpp
CC-Yeh: Removed `malloc`, thanks for the catch!

mlx/backend/metal/indexing.cpp
CC-Yeh: That make senses.
Moved some to ops and the rest are gone by introducing default batching.

mlx/backend/metal/indexing.cpp
CC-Yeh: changed to `dst`, thanks

mlx/primitives.cpp
CC-Yeh: I think a `gather` solution would be slower and more complex because, without `nonzero` or `argwhere` APIs, we'd probably have to perform the same complex `cumsum` and `scatter_add` before `gather` to get compact indices array.

If you'd anything in mind, I'm happy to implement it)

CC-Yeh:
Hey @awni 

Thanks again for the review, learned a lot from it!

I did some improvements. Can you take a look again?

mlx/ops.cpp
```diff
@@ -3442,6 +3442,35 @@ array scatter_min(
...
+  }
+
+  if (a.dtype() != src.dtype()) {
+    throw std::invalid_argument(
+        "[masked_scatter] dtype(a) must match dtype(src).");
+  }
+
+  auto expanded_a = expand_dims(a, 0, s);
+  auto expanded_mask = expand_dims(broadcast_to(mask, a.shape(), s), 0, s);
+  auto expanded_src = expand_dims(src, 0, s);
```
angeloskath: You shouldn't need the expand dims here.

mlx/primitives.cpp
```diff
@@ -4348,6 +4348,117 @@ bool ScatterAxis::is_equivalent(const Primitive& other) const {
...
+      const array cotan_flat = flatten(cotan, s);
+
+      const array mask_i = astype(mask_flat, int32, s);
+      const array prefix = cumsum(mask_i, 0, false, false, s);
+      const array idx_src = multiply(mask_i, prefix, s);
+      const array cotan_src =
+          multiply(cotan_flat, astype(mask_flat, cotan_flat.dtype(), s), s);
+
+      array gsrc_flat =
+          zeros({static_cast<int>(src.size())}, cotan_src.dtype(), s);
```
angeloskath, line 4379:
```suggestion
          zeros_like(src, s);
```

mlx/primitives.cpp
```diff
@@ -4348,6 +4348,117 @@ bool ScatterAxis::is_equivalent(const Primitive& other) const {
...
+      vjps.push_back(where(mask_b, zeros_like(cotan, s), cotan, s));
+    } else if (arg == 2) {
+      const array mask_flat = flatten(mask_b, s);
+      const array cotan_flat = flatten(cotan, s);
+
+      const array mask_i = astype(mask_flat, int32, s);
+      const array prefix = cumsum(mask_i, 0, false, false, s);
+      const array idx_src = multiply(mask_i, prefix, s);
+      const array cotan_src =
+          multiply(cotan_flat, astype(mask_flat, cotan_flat.dtype(), s), s);
```
angeloskath: Although infs and nans are bad news anyway, technically speaking this should be 

```cpp
const array cotan_src = where(mask_flat, cotan_flat, array(0, cotan_flat.dtype()), s);
```

because multiplying 0 with `inf` or `nan` will end up putting infs and nans in `cotan_src` which will end up in the gradient due to `scatter_add`.

mlx/primitives.cpp
```diff
@@ -4348,6 +4348,117 @@ bool ScatterAxis::is_equivalent(const Primitive& other) const {
...
+  for (int arg : argnums) {
+    if (arg == 0) {
+      vjps.push_back(where(mask_b, zeros_like(cotan, s), cotan, s));
+    } else if (arg == 2) {
+      const array mask_flat = flatten(mask_b, s);
+      const array cotan_flat = flatten(cotan, s);
+
+      const array mask_i = astype(mask_flat, int32, s);
+      const array prefix = cumsum(mask_i, 0, false, false, s);
+      const array idx_src = multiply(mask_i, prefix, s);
```
angeloskath: No need for this multiplication which will also spread the contention for the atomics (only slightly, still probably slow) away from index 0.

mlx/primitives.cpp
```diff
@@ -4348,6 +4348,117 @@ bool ScatterAxis::is_equivalent(const Primitive& other) const {
...
+      const array cotan_src =
+          multiply(cotan_flat, astype(mask_flat, cotan_flat.dtype(), s), s);
+
+      array gsrc_flat =
+          zeros({static_cast<int>(src.size())}, cotan_src.dtype(), s);
+      if (src.size() > 0) {
+        const int n = static_cast<int>(idx_src.size());
+        const array cotan_updates = reshape(cotan_src, {n, 1}, s);
+        gsrc_flat = scatter_add(gsrc_flat, idx_src, cotan_updates, 0, s);
```
angeloskath, line 4383: This is quite nice but it should probably be replaced with a `masked_gather` in the future because this will be quite slow since most indices will be 0  with 0 value which will be competing for the atomic add of 0.

mlx/primitives.cpp
```diff
@@ -4348,6 +4348,117 @@ bool ScatterAxis::is_equivalent(const Primitive& other) const {
...
+  for (int arg : argnums) {
+    if (arg == 0) {
+      out = where(mask_b, out, tangents[0], s);
+    } else if (arg == 2) {
+      out = masked_scatter(out, mask_b, tangents[1], s);
+    } else {
+      throw std::invalid_argument("[masked_scatter] invalid arg index in JVP");
+    }
+  }
+  return {out};
```
angeloskath, line 4423: That's very nice :-)

mlx/backend/metal/kernels/indexing/masked_scatter.h
```diff
@@ -0,0 +1,47 @@
...
+    uint idx [[thread_position_in_grid]]) {
+  const bool mask_value =
+      mask[elem_to_loc<uint>(idx, mask_shapes, mask_strides, mask_ndim)];
+
+  if (!mask_value) {
+    return;
+  }
+
+  const uint src_index = scatter_offsets[elem_to_loc<uint>(
+      idx, offsets_shapes, offsets_strides, offsets_ndim)];
```
angeloskath: Unless I am mistaken, since we flattened the mask, the offsets are an output of scan (which is contiguous but we can also check and copy if needed) and the output is a contiguous array we can probably remove all `elem_to_loc` calls except the one for `src`.

angeloskath (changes requested):
Very nice indeed! Thanks for the thorough work!

I left a spattering of comments but a more significant concern is that it doesn't implement the same API as numpy and I think we should. Boolean indices in numpy do not really broadcast and omitted dimensions are taken whole.

Basically
```python
x = np.arange(1000).reshape(10, 10, 10)
x[np.random.randn(10, 10) > 0] = 0    # valid
x[np.random.randn(1, 10, 10) > 0] = 0 # invalid
x[np.random.randn(10, 10, 1) > 0] = 0 # invalid
```

This also allows for a more efficient implementation for the cases where the axes being copied as is are the large ones. I do not think that you need to optimize this but I think we should implement the API at least.

The first and simplest approach would be to expand and broadcast the mask in the op then the primitive should work on `dst` and `mask` that have the same size and shape (which pretty much is already a requirement).

CC-Yeh:
@angeloskath 

Thanks for the insightful suggestions! Learned a lot from it 😃 

I addressed most comments, looking forward to your thoughts, thanks.

mlx/ops.cpp
CC-Yeh: I think this is necessary because `MaskedScatter::eval_gpu` treats every input as batched to simplify the kernel implementation and remove the need for the `vmap_axis` member variable.

mlx/primitives.cpp
CC-Yeh: Opened #2744

mlx/ops.cpp
angeloskath: I stand corrected. It is indeed much nicer this way.

CC-Yeh:
@angeloskath I like the vmap and shape comparison improvements!

I fixed CI errors, can you trigger the tests again?

angeloskath:
Ok finally fixed another misconception in vmap (added a double vmap test that shows what wasn't working). It was there before but also in the vmap I wrote.

Basically the `MaskedScatter` primitive is batched. The 0-th dim is the batch. The `masked_scatter` op is not batched. So in the primitive we can't use the op but we have to use the primitive and also we have to deal with the previous batch.

CC-Yeh:
@angeloskath  Nice catch!

Fixed another test, hope it'll pass now 🙏

CC-Yeh:
Hey @awni @angeloskath,

Tests are all good now. When you get a sec, could you check it again?

PR #2673, opened 2025-10-14
Nccl timeout

nastya236:
## Proposed changes

During multi-node runs, the master host sometimes starts later than other nodes. As a result, other hosts wait until rank 0 binds. In the previous implementation the binding retried 30 times with 500 ms intervals, which was not always sufficient (and hard-coded).
- Added `env::nccl_timeout` -- the binding will fail after the elapsed time exceeds `nccl_timeout` (default = 300 000 ms = 5 minutes, be overridden via the `MLX_NCCL_TIMEOUT`). 
- Removed an `ncclGroupEnd()` call -- this was a bug.

### DIFF
--- a/mlx/distributed/nccl/nccl.cpp
+++ b/mlx/distributed/nccl/nccl.cpp
@@ -184,7 +184,7 @@ inline void bootstrap_unique_id(
     close(sock);
 
   } else {
-    // Here we want to make show that rank 0 has enough time to bind
+    // Here we want to make sure that rank 0 has enough time to bind
     // so we will retry to connect until elapsed time exceeds nccl_timeout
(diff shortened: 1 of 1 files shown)


mlx/distributed/nccl/nccl.cpp
```diff
@@ -181,8 +184,9 @@ inline void bootstrap_unique_id(
     close(sock);
 
   } else {
-    // Here just wanted to make show that rank 0 has enough time to bind
-    // so we will retry to connect until max attempts
+    // Here we want to make show that rank 0 has enough time to bind
```
awni:
```suggestion
    // Here we want to make sure that rank 0 has enough time to bind
```

PR #2682, opened 2025-10-18
Fix KeyError for missing domain_uuid_key in Thunderbolt setup

thechriswebb:
- Check for None before accessing domain_uuid_key and skip invalid entries
- Filter items to only those with domain_uuid_key before extracting connection
- Prevents KeyError when thunderbolt topology data is incomplete

## Proposed changes

Fixes a `KeyError` in `mlx.distributed` when setting up Thunderbolt ring topology.

The error occurred when `domain_uuid_key` was missing from Thunderbolt topology data, which happens when certain devices (such as external monitors) are connected to the Mac. The code attempted to access this key without first checking for its presence.

This PR adds validation to check for `domain_uuid_key` before accessing it and filters topology items to only those with valid keys, preventing the crash when the topology data is incomplete.

## Checklist

### DIFF
--- a/python/mlx/distributed_run.py
+++ b/python/mlx/distributed_run.py
@@ -567,13 +567,16 @@ def prepare_tb_ring(args, hosts):
         name = ""
         ports = []
         for t in c["SPThunderboltDataType"]:
+            uuid = t.get("domain_uuid_key")
+            if uuid is None:
+                continue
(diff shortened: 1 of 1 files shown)


ivanfioravanti:
Thanks! I was experiencing this recently!

PR #2684, opened 2025-10-18
Prevent null release in Metal backend

Jckwind:
## Summary

Skip release() calls when mtl_function or linked functions are null so kernel caching never dereferences a missing handle.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -131,12 +131,12 @@ jobs:
       - uses: actions/download-artifact@v6
         with:
           pattern: linux-wheels-*
-          merge-multiples: true
+          merge-multiple: true
           path: dist
       - uses: actions/download-artifact@v6
         with:
           pattern: mac-wheels-*
-          merge-multiples: true
+          merge-multiple: true
           path: dist
       - name: Display structure of downloaded files
         run: ls -R dist


mlx/backend/metal/device.cpp
```diff
@@ -724,8 +724,12 @@ MTL::ComputePipelineState* Device::get_kernel_(
...
   auto kernel = get_kernel_(hash_name, mtl_function, mtl_linked_funcs);
 
-  mtl_function->release();
-  mtl_linked_funcs->release();
+  if (mtl_function) {
+    mtl_function->release();
+  }
+  if (mtl_linked_funcs) {
+    mtl_linked_funcs->release();
+  }
```
awni: In what scenario woudl these be null and we shouldn't throw an error?

Jckwind: I'm not sure - I was getting errors thrown from the release method being called on nullptr - i didn't pause to find a reproducible case / the root cause. I threw up guards and kept it pushin'.

I can refactor it though - what's the ideal behavior here?

awni: If those are null then the kernel can't really be valid and so program should crash with a helpful error message. And it should have crashed here: https://github.com/ml-explore/mlx/blob/8d746d73d946694ce738fbf2a066194adc0795fb/mlx/backend/metal/device.cpp#L594-L601

It's surprising that it didn't.. would be pretty useful to know how that happened as maybe it could point to a bug elsewhere.

Jckwind: it doesn't throw a runtime crash, which is weird. I can confirm that whatever I'm doing locally with custom kernels will throw these scary looking runtime errors if there's no nullptr guard. Those logs might look like:

```
src/pie_core/external/mlx/mlx/backend/metal/device.cpp:728:21: runtime error: member call on null pointer of type 'NS::Referencing<MTL::LinkedFunctions>'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior src/pie_core/external/mlx/mlx/backend/metal/device.cpp:728:21 
build/_deps/metal_cpp-src/Foundation/NSObject.hpp:112:49: runtime error: member call on null pointer of type 'NS::Referencing<MTL::LinkedFunctions> *'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior build/_deps/metal_cpp-src/Foundation/NSObject.hpp:112:49 
```

It could be my own `metal_cpp` dep - but I'm not sure. I don't have time to dedicate to hunting down *why* this error is being thrown - it's not thrown on eval or even a primitive instantiation. 

Execution happens fine - no crash / obvious performance impact - just the scary ghost runtime error. It disappears with the null guard. I'm not doing anything with the LinkedFunctions. 

My mlx fork is parity with main, so it's not from my own custom behavior.

awni: >  just the scary ghost runtime error

You mean it happens at the very end just before program exit?

Jckwind: No, at the very beginning of the program, and then everything runs as expected without error. The errors don't cause an exit.

awni: Oh that's very odd..

PR #2686, opened 2025-10-20
Fp8 conversion

awni:
- Adds a primitive and CPU / Metal / CUDA conversion routines for e4m3 fp8. 
- Adds C++ only ops to convert to/from fp8.

I didn't expose these to python because I think we haven't settled on API yet for fp8 (basically should it be a type or not). For now I'm leaning towards not since it has very limited support in other ops and it's not clear we need to expose it.

So for now, fp8 is represented as `mx.uint8`. So this op is really meant to be used by quantizations (e.g. nvfp4, mxfp8, other block scaled fp8 quants).

### DIFF
--- a/mlx/backend/cuda/device/unary_ops.cuh
+++ b/mlx/backend/cuda/device/unary_ops.cuh
@@ -2,6 +2,8 @@
 
 #pragma once
 
+#include <cuda_fp8.h>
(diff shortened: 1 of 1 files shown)


awni:
CC @nastya236

Issue #2675, opened 2025-10-15
Improve `mx.split` docs


awni:
> general design nitpick - for the mx.split method, when there are multiple indicies - the documentation says that these will be the starting indexes of the new subarrays - slightly confusing, diverges for the behavior. 

> if i have an array that's 19 items long, and I call split with {0, 18} i'd expect two arrays, one that is items 0-17, and another array with 1 element. But in reality this yields two arrays, one with 0 elements, and one with 19 elements.

> To get the behavior I wanted, the indicies I passed in for this 19 item array was {18, 1} and this resulted in the behavior I expected. 

> Not a bug - just something that might want to be clarified.

Maalvi14:
Will check this out and change the documentation accordingly.

Also would like to know your opinion on probably going with the documentation style like the `Containerization` project which matches Apple's Official documentation [Preview here](https://apple.github.io/containerization/documentation/containerization/)

PR #2689, opened 2025-10-21
Improved mx.split() docs

Maalvi14:
## Proposed changes

Changed the documentation of `mx.split()` based on the comments on #2675. Documentation change is regarding what a list passed as the second arg of the function is used for (a boundary that dictates all previous elements before the index will be split into a subarray).

**Before**: `the list contains the indices of the start of each subarray along the given axis.`
**After**: `the list contains the indices which are the split points or boundaries where the array should be divided.`

## Checklist

- [ ] I have added tests that prove my fix is effective or that my feature works **-> Not necessary**


awni (approved):
LGTM thanks!

Issue #2677, opened 2025-10-15
Better error message for `mx.einsum`


awni:
```python
import mlx.core as mx
y = mx.ones((6, 10))
x = mx.ones((6,))
mx.einsum("...i,ij->...j", x[0], y)
```

That fails with
```
ValueError: basic_string
```

PR #2690, opened 2025-10-21
Einsum error msg improvement

Maalvi14:
## Proposed changes

Modified the error message logic of `einsum()` mentioned on #2677 to clearly identify which operand is problematic, show its shape and the specific subscript pattern that's causing the issue.

``` python
import mlx.core as mx
y = mx.ones((6, 10))
x = mx.ones((6,))
mx.einsum("...i,ij->...j", x[0], y)
```

New Error Message
```
  File "/Users/mv/repos/mlx/einsum_test.py", line 4, in <module>
    mx.einsum("...i,ij->...j", x[0], y)
ValueError: [einsum] Operand 0 with shape {} has insufficient dimensions for subscript '...i'.
The ellipsis requires at least 1 dimensions but the operand has 0 dimensions.
```

Example on the same error in PyTorch:
`einsum(): the number of subscripts in the equation (1) is more than the number of dimensions (0) for operand 0`

## Checklist

- [ ] I have added tests that prove my fix is effective or that my feature works **-> Not necessary**
- [ ] I have updated the necessary documentation (if needed) **-> Not necessary**

### DIFF
--- a/mlx/einsum.cpp
+++ b/mlx/einsum.cpp
@@ -724,23 +724,6 @@ std::pair<std::vector<PathNode>, PathInfo> einsum_path_helper(
         ellipsis_length = max_ellipsis_length;
       }
 
-      if (ellipsis_length < 0) {
-        std::ostringstream msg;
-        msg << "[" << fn_name
-            << "] Ellipsis expansion resulted in negative length ("
-            << ellipsis_length << ") for subscript '" << subscript << "'.";
-        throw std::invalid_argument(msg.str());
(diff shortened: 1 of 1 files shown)


Maalvi14:
@awni noticed usually you have to change some of the format on my PRs, and I think that's caused by `pre-commit run --all-files` before I submit, don't know if you or other are experiencing the same situation.

awni:
Maybe try running `pre-commit install`?

PR #2692, opened 2025-10-21
fix warnings showing up with -Wall

andresy:
## Proposed changes

Fix all warnings occurring when compiling with `-Wall`

## Checklist

- [x ] I have read the [CONTRIBUTING](https://github.com/ml-explore/mlx/blob/main/CONTRIBUTING.md) document
- [ x] I have added tests that prove my fix is effective or that my feature works


awni:
This is awesome, thanks! What do you think about adding `-Wall` as a build option by default so we stay warning free (that way it will be checked in circle).

andresy:
I think with `clang` it would make sense.

With `gcc` one get a lot more warnings -- (possibly what you would get with `-Wall -Wextra` with `clang`), so I think we would not want to have this option enabled right now for `gcc`.

These warnings you get with `gcc` include type mismatch warnings (say a `int` for loop, with a `size_t` limit). I did not fix these yet, I was actually not sure we want to fix them

awni (approved):
LGTM

PR #2702, opened 2025-10-24
optionally load metallib from framework

davidkoski:
## Proposed changes

Support for https://github.com/ml-explore/mlx-swift/pull/289 -- this will also allow `SWIFTPM_BUNDLE` to name a framework identifier.

## Checklist


mlx/backend/metal/device.cpp
```diff
@@ -72,6 +72,19 @@ MTL::Library* try_load_bundle(
   }
   return nullptr;
 }
+
+MTL::Library* try_load_framework(
```
davidkoski, line 76: The logic in `try_load_bundle` isn't appropriate for a framework -- they have a different layout.

PR #2709, opened 2025-10-28
add null check -- the bundleIdentifier is optional

davidkoski:
## Proposed changes

Add null check as function can return null (observed by person testing swift xcodeproj).  See also #2702

## Checklist

### DIFF
--- a/mlx/backend/metal/device.cpp
+++ b/mlx/backend/metal/device.cpp
@@ -122,7 +122,7 @@ std::pair<MTL::Library*, NS::Error*> load_swiftpm_library(
     const auto bundle = reinterpret_cast<NS::Bundle*>(frameworks->object(i));
     const auto identifier = bundle->bundleIdentifier();
     if (identifier != nullptr &&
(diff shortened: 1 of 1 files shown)


mlx/backend/metal/device.cpp
```diff
@@ -119,8 +119,10 @@ std::pair<MTL::Library*, NS::Error*> load_swiftpm_library(
   // if SWIFTPM_BUNDLE is a framework identifier, try loading from that
   auto frameworks = NS::Bundle::allFrameworks();
   for (int i = 0, c = (int)frameworks->count(); i < c; i++) {
-    auto bundle = reinterpret_cast<NS::Bundle*>(frameworks->object(i));
-    if (!strcmp(bundle->bundleIdentifier()->utf8String(), SWIFTPM_BUNDLE)) {
+    const auto bundle = reinterpret_cast<NS::Bundle*>(frameworks->object(i));
+    const auto identifier = bundle->bundleIdentifier();
```
davidkoski, line 123: This can return null, per https://developer.apple.com/documentation/foundation/bundle/bundleidentifier and observed by somebody testing the mlx-swift xcodeproj branch.

PR #2715, opened 2025-10-29
Fix addmm with empty matrices and beta != 1.0

### LINKED ISSUE #2698: [BUG] `mx.addmm` with empty `a` and `b` produces incorrect result
admm with empty `a` and `b` produces incorrect result when `beta != 1.0`

```python
import mlx.core as mx
a = mx.ones((2, 0))
b = mx.ones((0, 2))
c = mx.ones((2, 2))
print(mx.addmm(c, a, b, alpha=0.0, beta=2.0))
```

```
array([[1, 1],
       [1, 1]], dtype=float32)
```

harsh-sutariya:
Fixes #2698 [BUG] mx.addmm with empty a and b produces incorrect result

## Proposed changes

- Fix CPU backend to apply beta scaling for empty matrices (K=0)
- Fix Metal backend to apply beta scaling for empty matrices (K=0)
- Add tests for empty matrix addmm with various beta values

The bug occurred when addmm was called with empty matrices (one dimension is 0) and beta != 1.0. The function would return C instead of beta * C because the early return for empty matrices didn't apply beta scaling.

## Checklist

### DIFF
--- a/mlx/backend/cpu/matmul.cpp
+++ b/mlx/backend/cpu/matmul.cpp
@@ -178,6 +178,7 @@ void AddMM::eval_cpu(const std::vector<array>& inputs, array& out) {
                 "[AddMM::eval_cpu] Unsupported dtype for beta scaling");
         }
       });
+      encoder.add_temporary(std::move(beta_scalar));
     }
     return;
   }


mlx/backend/metal/matmul.cpp
```diff
@@ -925,19 +926,28 @@ void AddMM::eval_gpu(const std::vector<array>& inputs, array& out) {
...
-        stream());
+    auto& c = inputs[2];
+    if (beta_ == 1.0f) {
+      copy_gpu(
+          c,
+          out,
+          c.flags().row_contiguous ? CopyType::Vector : CopyType::General,
+          s);
+    } else {
+      out.set_data(allocator::malloc(out.nbytes()));
```
angeloskath: No need to allocate memory here, `binary_op_gpu` allocates memory and handles donation. 
```suggestion
```

mlx/backend/cpu/matmul.cpp
```diff
@@ -135,15 +135,64 @@ void AddMM::eval_cpu(const std::vector<array>& inputs, array& out) {
...
+              val = complex64_t(val.real() * beta, val.imag() * beta);
+            }
+            break;
+          default:
+            throw std::runtime_error(
+                "[AddMM::eval_cpu] Unsupported dtype for beta scaling");
+        }
+      });
+    }
+    return;
```
angeloskath, line 183: Same as for the gpu this should be done using `binary` when `beta_ != 1.0`. No need for extraneous copy, and it will use vectorization when available.

mlx/backend/metal/matmul.cpp
harsh-sutariya: Done, removed explicit allocation and now rely on `binary_op_gpu` for memory handling as suggested.

mlx/backend/cpu/matmul.cpp
harsh-sutariya: Updated to use binary operation for beta scaling instead of manual copy.

harsh-sutariya:
@angeloskath I’ve addressed your feedback, 
Please take another look when convenient. Thanks!

angeloskath (approved):
Looks good I will merge after the tests pass.

harsh-sutariya:
Temporary array `beta_scalar` was destroyed before the asynchronous computation used it, causing garbage memory reads on Linux CI.

Added `encoder.add_temporary(std::move(beta_scalar));` to keep the array alive until computation completes.

The fix is now passing locally and ready for CI rerun. Could you please approve the workflow when you get a chance?
@angeloskath

PR #2716, opened 2025-10-29
Migrate CircleCI to GitHub Actions

madrob:
## Proposed changes

Port over the CircleCI configuration to equivalent GitHub Actions configuration.

### DIFF
--- a/.github/actions/build-macos/action.yml
+++ b/.github/actions/build-macos/action.yml
@@ -1,0 +1,124 @@
+name: 'Build and Test on macOS'
+description: 'Build and test MLX on macOS'
+
+inputs:
+  build-type:
+    description: 'Build type (debug, release)'
+    required: false
+    default: 'debug'
+    type: choice
+    options:
+      - debug
+      - release
+  run-tests:
+    description: 'Whether to run tests'
+    required: false
+    default: 'true'
+  build-jit:
+    description: 'Whether to build with JIT'
+    required: false
+    default: 'true'
+
+runs:
+  using: "composite"
+  steps:
+    - name: Install dependencies
+      shell: sh
+      env:
+        DEBUG: 1
+        DEV_RELEASE: 1
+      run: |
+        uv pip install --upgrade pip cmake setuptools
+        uv pip install nanobind==2.4.0 \
+          numpy torch tensorflow unittest-xml-reporting
+        uv pip install -e . -v
+
+    - name: Generate package stubs
+      shell: bash
+      run: |
+        uv pip install typing_extensions
+        uv run --no-project setup.py generate_stubs
+
+    - name: Run Python tests
+      if: inputs.run-tests == 'true'
+      shell: bash
+      env:
+        LOW_MEMORY: 1
+      run: |
+        DEVICE=cpu uv run -m xmlrunner discover -v python/tests -o test-results/cpu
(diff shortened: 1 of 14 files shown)


.github/workflows/release.yml
```diff
@@ -0,0 +1,188 @@
...
+          path: artifacts
+      - uses: actions/download-artifact@v6
+        with:
+          pattern: mac-wheels-*
+          merge-multiples: true
+          path: artifacts
+      - name: Display structure of downloaded files
+        run: ls -R artifacts
+      # - name: Publish package distributions to PyPI
+      #  uses: pypa/gh-action-pypi-publish@release/v1
```
madrob, line 131: Note that this will not publish to pypi yet, we will have to do that as a second commit after the workflows exist and we can configure trusted publishing.

.github/actions/setup-linux/action.yml
```diff
@@ -0,0 +1,83 @@
...
+        wget https://developer.download.nvidia.com/compute/cuda/repos/ubuntu2204/x86_64/cuda-keyring_1.1-1_all.deb
+        sudo dpkg -i cuda-keyring_1.1-1_all.deb
+        sudo apt-get update
+        sudo apt-get install -y libcudnn9-dev-cuda-12 libnccl2 libnccl-dev cuda-toolkit-12-9
+      # Note: This installs CUDA 12.9, which is the latest supported by cuDNN 9.x and works with the NVidia 570 drivers
+      # cuda-toolkit by itself installs version 13 (+) and requires updated drives (580+), which require a reboot to function properly.
+      # Compatibility matrix: https://docs.nvidia.com/deeplearning/cudnn/backend/latest/reference/support-matrix.html
+      # This also drops `nvcc` into `/usr/local/cuda-12.9/bin/nvcc` - but it's *not* on the default PATH
+
+    - name: Package and Driver Report
```
madrob, line 69: This is not required, but was useful during debugging when setting up the workflows and has low cost to keeping it in.

.github/actions/build-macos/action.yml
```diff
@@ -0,0 +1,124 @@
...
+    - name: Build macOS 14 package
+      if: inputs.build-type == 'release'
+      uses: ./.github/actions/build-macos-release
+      with:
+        macos-target: 14.0
+    - name: Build macOS 15 package
+      if: inputs.build-type == 'release'
+      uses: ./.github/actions/build-macos-release
+      with:
+        macos-target: 15.0
```
madrob, line 124: We can't embed a matrix inside of a job so manually specify the combinations in series. It's not that many so it's probably ok to maintain this for now.

.github/dependabot.yml
```diff
@@ -0,0 +1,6 @@
+version: 2
+updates:
```
madrob, line 2: python ecosystem maybe deserves to be here as well

.github/actions/setup-linux/action.yml
```diff
@@ -0,0 +1,83 @@
...
+        python-version: ${{ inputs.python-version }}
+        cache: 'pip'
+
+    - name: setup python venv
+      shell: bash
+      run: |
+        python -m venv .venv
+        source .venv/bin/activate
+        echo PATH=$PATH >> $GITHUB_ENV
+        pip install --upgrade pip cmake
```
madrob, line 46: We used system python and added the venv directly to the path because cmake was having issues deciding whether to look at system or uv and it was getting too messy to configure it otherwise.

PR #2720, opened 2025-10-31
Use faster dequant for fp4

awni:
Uses a faster + simpler dequant (instead of the LUT for FP4). Speeds up generation a little but also simplifies the code.

Pre: eneration_tps=121.29
Post: generation_tps=124.01

For QMMs I didn't change it because it's slower. The extra ops are worse than using the LUT.


angeloskath (approved):
Very nice!

awni:
Unfortunately can't take credit for this. The very clever dequant here (putting the bits in an fp16 and scaling) was from Alex Kan

PR #2722, opened 2025-10-31
update: add linux fedora container CI - CPP build test only

incertum:
## Proposed changes

Allow me some storytelling time: When I worked on eBPF projects, I noticed a few times that on Ubuntu, for instance, the eBPF verifier would let some buggy bytecode pass. In contrast, testing on a yum-based distro such as Fedora proved to be very valuable for catching additional bugs.

Therefore, I am proposing to add a formal CI build inside a Fedora container for our C++ code.

Note: We could not pre-test the newer arm runner - let’s see.
CC @madrob 

## Checklist

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -92,7 +92,17 @@ jobs:
           retention-days: 7
 
   linux_fedora_build_cpp:
-    runs-on: ubuntu-22.04
+    name: Linux Fedora CPP Build (${{ matrix.arch }})
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - host: ubuntu-22.04
+            arch: x86_64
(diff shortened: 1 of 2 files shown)


.github/workflows/nightly.yml
```diff
@@ -91,3 +91,26 @@ jobs:
           path: wheelhouse/mlx_cuda-*.whl
           retention-days: 7
 
+  linux_fedora_build_cpp:
+    runs-on: ubuntu-22.04
```
madrob: I would set this up as a single matrix build with something like `strategy: matrix: host: [ubuntu-22.04, ubuntu-22.04-arm]` and `runs-on: ${{ matrix.host }}` to visually reinforce that it's the same build but in two different environments. Syntax might need tweaking as I typed it from memory here.

PR #2725, opened 2025-11-01
[CUDA] Reduce use of managed memory

awni:
This is an attempt to reduce the use of managed memory in our cuda back-end. There are still a few details to figure out / ideas to explore. But for the common use cases of LLM training / inference it works quite well.

Some key differences:
- `malloc` is still the same as before. It gives you managed memory and caches it in the buffer cache
- `cu::malloc_async` gives you async allocated non-managed memory and caches it in the buffer cache
- `eval_gpu` should always prefer `cu::malloc_async(size, stream);` to get the good memory
- pointers to kernels should be gotten with `gpu_ptr<T>(array)`
- Accessing `arr.data<T>()` can do a copy if the data is not in the right place. So don't access it unless you need the data accessible to the CPU

Some improvements to think about:

- All allocations still go in the buffer cache. When doing `malloc_async` or `malloc` we first check the buffer cache. It might be good to prefer the appropriate type of memory (managed or not)  when pulling from the cache.

- When doing the copy to managed memory, we rely on the fact that it does a device synchronize. This can be slow. So it may be good to make it async when possible.
 
### Benchmarks

Command:
```
mlx_lm.benchmark --model meta-llama/Meta-Llama-3.1-8B --p 2049 -g 128 -b 1 -n 4
```

DGX Spark 
Pre: Averages: prompt_tps=744.025, generation_tps=10.301, peak_memory=21.707
Post: Averages: prompt_tps=2093.590, generation_tps=14.553, peak_memory=21.707

B200
Pre: Averages: prompt_tps=4780.945, generation_tps=207.900, peak_memory=21.707
Post: Averages: prompt_tps=37887.799, generation_tps=217.166, peak_memory=21.707

note the prompt length of 2049 is to get an aligned size (2048) for prompt processing, otherwise it's quite a bit slower.

### DIFF
--- a/mlx/backend/cuda/allocator.cpp
+++ b/mlx/backend/cuda/allocator.cpp
@@ -97,15 +97,33 @@ CudaAllocator::CudaAllocator()
 
   int device_count = 0;
   CHECK_CUDA_ERROR(cudaGetDeviceCount(&device_count));
+  int curr;
+  CHECK_CUDA_ERROR(cudaGetDevice(&curr));
   for (int i = 0; i < device_count; ++i) {
     CHECK_CUDA_ERROR(cudaSetDevice(i));
     cudaStream_t s;
     CHECK_CUDA_ERROR(cudaStreamCreateWithFlags(&s, cudaStreamNonBlocking));
     free_streams_.push_back(s);
   }
+  CHECK_CUDA_ERROR(cudaSetDevice(curr));
+}
+
+void copy_to_managed(CudaBuffer& buf) {
+  // TODO maybe make this async on a i/o stream to avoid synchronizing the
+  // device on malloc/and free
+  void* new_data;
+  CHECK_CUDA_ERROR(cudaMallocManaged(&new_data, buf.size));
+  buf.device = -1;
+  CHECK_CUDA_ERROR(cudaMemcpy(new_data, buf.data, buf.size, cudaMemcpyDefault));
+  CHECK_CUDA_ERROR(cudaFree(buf.data));
+  buf.data = new_data;
 }
 
 Buffer CudaAllocator::malloc_impl(size_t size, cudaStream_t stream) {
+  if (size == 0) {
+    return Buffer{new CudaBuffer{nullptr, 0, -1}};
+  }
+
   // Find available buffer from cache.
   std::unique_lock lock(mutex_);
   if (size <= small_block_size) {
@@ -114,6 +132,11 @@ Buffer CudaAllocator::malloc_impl(size_t size, cudaStream_t stream) {
     size = next_power_of_2(size);
   } else {
     size = page_size * ((size + page_size - 1) / page_size);
+  }
+
+  int device = -1;
+  if (size > small_block_size && stream != nullptr) {
+    CHECK_CUDA_ERROR(cudaStreamGetDevice(stream, &device));
   }
 
   CudaBuffer* buf = buffer_cache_.reuse_from_cache(size);
@@ -131,10 +154,6 @@ Buffer CudaAllocator::malloc_impl(size_t size, cudaStream_t stream) {
     }
     lock.unlock();
     if (!buf) {
-      int device = -1;
-      if (stream != nullptr) {
-        cudaStreamGetDevice(stream, &device);
-      }
       buf = new CudaBuffer{nullptr, size, device};
       cudaError_t err;
       if (device == -1) {
@@ -156,6 +175,10 @@ Buffer CudaAllocator::malloc_impl(size_t size, cudaStream_t stream) {
   if (get_cache_memory() > max_pool_size_) {
     buffer_cache_.release_cached_buffers(get_cache_memory() - max_pool_size_);
   }
+  // Copy to managed here if the buffer is not on the right device
+  if (buf->device != device) {
+    copy_to_managed(*buf);
+  }
   return Buffer{buf};
 }
 
@@ -170,6 +193,10 @@ Buffer CudaAllocator::malloc(size_t size) {
 void CudaAllocator::free(Buffer buffer) {
   auto* buf = static_cast<CudaBuffer*>(buffer.ptr());
   if (!buf) {
+    return;
+  }
+  if (buf->size == 0) {
+    delete buf;
     return;
   }
 
@@ -274,15 +301,7 @@ void* Buffer::raw_ptr() {
   }
   auto& cbuf = *static_cast<cu::CudaBuffer*>(ptr_);
   if (cbuf.device != -1) {
-    // TODO maybe make this async on a i/o stream to avoid synchronizing the
-    // device on malloc/and free
-    void* new_data;
-    CHECK_CUDA_ERROR(cudaMallocManaged(&new_data, cbuf.size));
-    cbuf.device = -1;
-    CHECK_CUDA_ERROR(
-        cudaMemcpy(new_data, cbuf.data, cbuf.size, cudaMemcpyDefault));
-    CHECK_CUDA_ERROR(cudaFree(cbuf.data));
-    cbuf.data = new_data;
+    copy_to_managed(cbuf);
   }
   return cbuf.data;
 }
--- a/.github/actions/build-cuda/action.yml
+++ b/.github/actions/build-cuda/action.yml
@@ -26,10 +26,6 @@ runs:
         CMAKE_ARGS: -DMLX_BUILD_CUDA=ON -DCMAKE_COMPILE_WARNING_AS_ERROR=ON -DCMAKE_CUDA_COMPILER=${{ inputs.nvcc-location }}
       run: pip install -e ".[dev]" -v
 
-    - name: Check if build actually worked
-      shell: bash
-      run: python -c "import mlx.core"
-
     - name: Run Python tests - CPU
       if: inputs.run-tests == 'true'
       shell: bash
--- a/mlx/backend/cuda/distributed.cu
+++ b/mlx/backend/cuda/distributed.cu
@@ -74,7 +74,7 @@ void AllGather::eval_gpu(
   };
 
   auto input = ensure_contiguous(inputs[0]);
-  outputs[0].set_data(allocator::malloc(outputs[0].nbytes()));
+  outputs[0].set_data(cu::malloc_async(outputs[0].nbytes(), encoder.stream()));
 
   encoder.set_input_array(input);
   encoder.set_output_array(outputs[0]);
@@ -103,7 +103,7 @@ void ReduceScatter::eval_gpu(
   };
 
   auto input = ensure_contiguous(inputs[0]);
-  outputs[0].set_data(allocator::malloc(outputs[0].nbytes()));
+  outputs[0].set_data(cu::malloc_async(outputs[0].nbytes(), encoder.stream()));
 
   encoder.set_input_array(input);
   encoder.set_output_array(outputs[0]);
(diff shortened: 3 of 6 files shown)


mlx/backend/cuda/allocator.cpp
```diff
@@ -88,15 +90,19 @@ CudaAllocator::CudaAllocator()
...
           [this](CudaBuffer* buf) { cuda_free(buf); }) {
-  // TODO: Set memory limit for multi-device.
   size_t free, total;
   CHECK_CUDA_ERROR(cudaMemGetInfo(&free, &total));
   memory_limit_ = total * 0.95;
   max_pool_size_ = memory_limit_;
+  int loc = 0;
+  CHECK_CUDA_ERROR(cudaDeviceGetDefaultMemPool(&cuda_pool_, loc));
+  CHECK_CUDA_ERROR(cudaMemPoolSetAttribute(
+      cuda_pool_, cudaMemPoolAttrReleaseThreshold, &memory_limit_));
```
zcbenz: I haven't actually ran the code but reading the docs I think the threshold works similar to MLX's buffer cache that CUDA would not release real memory until threshold is exceeded?

If that is the case I think the MLX buffer cache and CUDA memory pool cache do not work together: if used memory is below `memory_limit_` only MLX's buffer cache will be used, and if it is over the threshold neither cache would be used. The only case that CUDA's memory pool cache kicks in is after `clear_cache()` gets called, which is probably against the purpose of `clear_cache()`.

mlx/backend/cuda/allocator.cpp
```diff
@@ -88,15 +90,19 @@ CudaAllocator::CudaAllocator()
           page_size,
           [](CudaBuffer* buf) { return buf->size; },
           [this](CudaBuffer* buf) { cuda_free(buf); }) {
-  // TODO: Set memory limit for multi-device.
   size_t free, total;
   CHECK_CUDA_ERROR(cudaMemGetInfo(&free, &total));
   memory_limit_ = total * 0.95;
   max_pool_size_ = memory_limit_;
+  int loc = 0;
+  CHECK_CUDA_ERROR(cudaDeviceGetDefaultMemPool(&cuda_pool_, loc));
```
zcbenz: We can support multi-device with a for loop:

```c++
for (int device = device_count; device >=0; device++) {
  CHECK_CUDA_ERROR(cudaSetDevice(device));
  ...
}
```

mlx/backend/cuda/allocator.cpp
```diff
@@ -121,16 +127,22 @@ Buffer CudaAllocator::malloc(size_t size) {
...
     if (!buf) {
-      buf = new CudaBuffer{nullptr, size};
-      cudaError_t err = cudaMallocManaged(&buf->data, size);
+      bool managed = stream == nullptr;
+      buf = new CudaBuffer{nullptr, size, managed};
+      cudaError_t err;
+      if (managed) {
+        err = cudaMallocManaged(&buf->data, size);
+      } else {
+        err = cudaMallocFromPoolAsync(&buf->data, size, cuda_pool_, stream);
```
zcbenz: This does not work with multi-device because `cuda_pool_` belongs to device 0. Since only the default memory pool is used, we can just use `cudaMallocAsync` here, which automatically allocates on the device that `stream` belongs to. And there is no longer need to keep `cuda_pool_` which is device dependent.

mlx/backend/cuda/allocator.h
```diff
@@ -17,8 +18,14 @@ using allocator::Buffer;
 struct CudaBuffer {
   void* data;
   size_t size;
+  bool managed;
 };
 
+template <typename T>
+T* gpu_ptr(Buffer buf) {
```
zcbenz:
```suggestion
inline T* gpu_ptr(Buffer buf) {
```

mlx/backend/cuda/allocator.h
```diff
@@ -17,8 +18,14 @@ using allocator::Buffer;
...
+  return static_cast<T*>(static_cast<cu::CudaBuffer*>(buf.ptr())->data);
```
zcbenz: When accessing `gpu_ptr()` it is possible that the buffer is on a different device, I think we can probably:

1. make `gpu_ptr()` take a `stream` arg and do memcpy when buffer belongs to a different device. i.e. moving the buffer to current device.
2. in `eval_gpu()` check the device of inputs and do explicit copies when necessary.

mlx/backend/cuda/allocator.h
```diff
@@ -17,8 +18,14 @@ using allocator::Buffer;
 struct CudaBuffer {
   void* data;
   size_t size;
+  bool managed;
```
zcbenz: For multi-device environment we also need to record the device of buffer. I think we can use -1 to indicate managed memory.

```suggestion
  int device;
```

mlx/array.h
```diff
@@ -352,12 +352,13 @@ class array {
   // Return a raw pointer to the arrays data
```
zcbenz: Can you add a comment on the side effect of `data()` (possible memcpy)?

mlx/backend/cuda/allocator.cpp
```diff
@@ -193,6 +213,9 @@ size_t CudaAllocator::get_memory_limit() {
 size_t CudaAllocator::set_memory_limit(size_t limit) {
   std::lock_guard lock(mutex_);
   std::swap(limit, memory_limit_);
```
zcbenz, line 253:
```suggestion
{
  std::lock_guard lock(mutex_);
  std::swap(limit, memory_limit_);
}
```

mlx/io/load.cpp
```diff
@@ -226,10 +227,7 @@ array load(std::shared_ptr<io::Reader> in_stream, StreamOrDevice s) {
     throw std::runtime_error("[load] Failed to open " + in_stream->label());
   }
 
-  auto stream = to_stream(s, Device::cpu);
-  if (stream.device != Device::cpu) {
-    throw std::runtime_error("[load] Must run on a CPU stream.");
-  }
+  auto stream = to_stream(s, cu::is_available() ? Device::gpu : Device::cpu);
```
zcbenz: `Device::gpu` is always device 0, it should load on the device that `s` is on.

zcbenz:
A future improvement we can do is to use `cudaFreeAsync` to free buffers.

mlx/backend/cuda/allocator.cpp
awni: I don't think we need to worry about supporting multi device right now. I wouldn't actively design agains it.. but we don't have any use cases with it so I wouldn't spend too much time on it right now. For current multi GPU jobs we use multiple processes.

mlx/backend/cuda/allocator.h
awni: Good point, but let's deal with this when we add multi device support. I don't want to add that complexity for something we might not add.

mlx/backend/cuda/allocator.h
awni: Same comment as above.

mlx/io/load.cpp
awni: This is just the default if the stream is not specified. If the stream is specified it will use it's device.

mlx/backend/cuda/allocator.cpp
awni: Trying to remove the pool so this will not be relevant.

mlx/backend/cuda/allocator.cpp
awni: Looking into removing the pool so this will not be relevant 🤞

mlx/backend/cuda/allocator.cpp
awni: One nice thing about the cuda pool is it can do defragmentation which is in fact the main reason we tend to clear the MLX memory cache:

> If a memory allocation request made using cudaMallocAsync can’t be serviced due to fragmentation of the corresponding memory pool, the CUDA driver defragments the pool by remapping unused memory in the pool to a contiguous portion of the GPU’s virtual address space.

It would be really nice if we could replace the MLX buffer cache with a CUDA pool(s). In CUDA 13 it's possible to use a pool with managed memory but not in CUDA 12. 

I'm going to try and remove the pool for now.. it might make things slower and we can consider alternatives / adding it back in a follow up.

awni: Using cudaFreeAsync instead of the pool works well enough (I don't see a slowdown).

awni:
Ok I think this is good to go. @zcbenz, thanks for the review. I addressed your comments. The main thing was the removal of the cuda pool which simplified stuff nicely.

zcbenz:
> It would be really nice if we could replace the MLX buffer cache with a CUDA pool(s). In CUDA 13 it's possible to use a pool with managed memory but not in CUDA 12.

What do you think if we use MLX buffer cache for managed memory and CUDA memory pool for others (as a future work)?

> note the prompt length of 2049 is to get an aligned size (2048) for prompt processing, otherwise it's quite a bit slower.

Can you elaborate on this? I'm trying to understand why an extra token is needed to get aligned size.

awni:
> What do you think if we use MLX buffer cache for managed memory and CUDA memory pool for others (as a future work)?

Worth exploring. We might also want to make the cache in a similar API to make it easy to swap a cuda pool even for managed memory when you are on cuda toolkit 13 and up.

awni:
>  Can you elaborate on this? I'm trying to understand why an extra token is needed to get aligned size.

It's an artifact of how we process the prompt to avoid compute the LM head for the prompt for all but the last token. So basically we process the prompt in two steps: n-1 tokens with no LM head and the last 1 token with the LM head.

Shivansh9000:
No

mlx/backend/cpu/primitives.cpp
```diff
@@ -333,7 +333,7 @@ void Reshape::eval_cpu(const std::vector<array>& inputs, array& out) {
 
 void DynamicSlice::eval_cpu(const std::vector<array>& inputs, array& out) {
   if (out.size() == 0) {
-    out.set_data(nullptr);
+    out.set_data(allocator::malloc(0));
```
angeloskath, line 336: Is it intentional to allocate a scalar for empty outputs instead of nullptr?

awni: This is intentional. But it's not exactly a scalar (unless you are assuming the behavior of the underlying allocator). The allocator just has to return a well-formed `Buffer` with size 0. (Which is a `Buffer{nullptr}` for metal and cpu and a `Buffer` with a valid `CudaBuffer` for CUDA. This simplifies the CUDA backend because it doesn't have to check if the buffer is valid before accessing it's properties.

Also previously `nullptr` would implicitly cast to a `Buffer`, I made the constructor explicit cause it seemed more clear that way.  So even previously that call could be read as `set_data(Buffer{nullptr})`, so for most CPU/Metal it's the same as before.

angeloskath: No I understand the need for a valid `CudaBuffer` but why not `{nullptr, 0, -1}`. Here this hits the small pool and allocates a scalar.

angeloskath: Perhaps I should have phrased it as should we add a check for 0 in the cuda allocator?

awni: > Perhaps I should have phrased it as should we add a check for 0 in the cuda allocator?

Yea I was just being lazy cause it's such an edge case. I'll add that in the allocator.

angeloskath: Heh sorry :-)

yiakwy-xpu-ml-framework-team:
Hi @awni I want to help in supporting this framework in H100/H800 DGX machines with standard IB and nccl support. 

How can I start with it ?

yiakwy-xpu-ml-framework-team:
@awni Is there any body working on VMM , may be this can be the first topic I can contribute :

https://github.com/ruizhang1230/vTensor

PR #2726, opened 2025-11-02
Shapeless support for `zeros/ones_like`

### LINKED ISSUE #2599: Make `zeros_like` work with shapeless compile / export.

CC-Yeh:
## Proposed changes

Closes #2599 

## Checklist

### DIFF
--- a/mlx/ops.cpp
+++ b/mlx/ops.cpp
@@ -285,15 +285,14 @@ array full_impl(array vals, Dtype dtype, StreamOrDevice s /* = {} */) {
       vals.shape(),
       dtype,
       std::make_shared<Full>(to_stream(s)),
-      {astype(std::move(vals), dtype, s)});
+      {astype(vals, dtype, s)});
 }
 
 array full(Shape shape, array vals, Dtype dtype, StreamOrDevice s /* = {} */) {
   if (std::any_of(shape.begin(), shape.end(), [](auto i) { return i < 0; })) {
     throw std::invalid_argument("[full] Negative dimensions not allowed.");
   }
-  return full_impl(
-      broadcast_to(std::move(vals), std::move(shape), s), dtype, s);
+  return full_impl(broadcast_to(vals, std::move(shape), s), dtype, s);
 }
 
 array full(Shape shape, array vals, StreamOrDevice s /* = {} */) {


mlx/ops.h
```diff
@@ -69,6 +69,17 @@ array full(Shape shape, T val, StreamOrDevice s = {}) {
...
 
+array full_like(const array& a, array vals, Dtype dtype, StreamOrDevice s = {});
+array full_like(const array& a, array vals, StreamOrDevice s = {});
+template <typename T>
+array full_like(const array& a, T val, Dtype dtype, StreamOrDevice s = {}) {
+  return full_like(a, array(val, dtype), dtype, to_stream(s));
+}
+template <typename T>
+array full_like(const array& a, T val, StreamOrDevice s = {}) {
+  return full_like(a, array(val), to_stream(s));
```
awni: I think the output should inherit the data type of the input here (e.g. `full_like`). So you would need to set `array(val, a.dtype())`.

mlx/ops.cpp
```diff
@@ -280,37 +280,55 @@ array copy(array a, StreamOrDevice s /* = {} */) {
...
+    array vals,
+    Dtype dtype,
+    StreamOrDevice s /* = {} */) {
+  auto inputs = broadcast_arrays({a, std::move(vals)}, s);
+  return full_impl(std::move(inputs[1]), dtype, s);
+}
+
+array full_like(const array& a, array vals, StreamOrDevice s /* = {} */) {
+  auto dtype = vals.dtype();
+  return full_like(a, std::move(vals), dtype, to_stream(s));
```
awni: The dtype here should be from `a`

python/tests/test_compile.py
```diff
@@ -482,6 +482,18 @@ def fun(x):
...
+        x = mx.zeros((1, 1, 32))
+
+        def zeros_fun(x):
+            return mx.zeros_like(x)
+
+        def ones_fun(x):
+            return mx.ones_like(x)
+
+        self.assertEqual(mx.compile(zeros_fun, shapeless=True)(x).shape, (1, 1, 32))
+        self.assertEqual(mx.compile(ones_fun, shapeless=True)(x).shape, (1, 1, 32))
```
awni: The point of shapeless compile is that it works even if you *change* the shape. So it would be good to add a second call where you change the shape and make sure you get the right shape back.

awni:
This looks nice. Please address the comments then we can merge it.

Also could you add a test for `full_like` especially around making sure the various overloads return the expected types? (as they look incorrect at the moment)

mlx/ops.h
CC-Yeh: Thanks for the suggestion! My intention with `full_like(a, val)` was:

- `a` provides the shape

- `val` provides both the fill `value` and the `dtype`

So `full_like(a, val)` felt like the most direct expression of “give me an array with the same shape as `a`, filled with `val (value + type)`”.

What do you think?

awni: Yea I think it could go either way. When it's somewhat ambiguous (should the type come from val or a) then I would default to the same behavior as other frameworks. Numpy (and Jax) usually are the first I check (and they tend to have the same behavior). In this case Numpy takes the type from `a` and not from `val`.

CC-Yeh: Cool that makes sense! thanks

CC-Yeh:
@awni  Can you take a look again? Thanks!

awni (approved):
Looks great, thanks! Will merge when tests clear.

awni:
@CC-Yeh the tests are failing. Are you able to address that?

CC-Yeh:
> @CC-Yeh the tests are failing. Are you able to address that?

@awni Sorry a bit busy lately, will try to reproduce and fix it on colab.

CC-Yeh:
@awni should be fixed now!

PR #2727, opened 2025-11-03
Nccl reduce scatter, all gather

nastya236:
## Proposed changes

1. Added nccl `all_gather` and `reduce_scatter`
2. Make majority of  tests shared for all backends:

- nccl doesn’t support some dtypes, so there’s an extra `all_reduce` test for ring/MPI. But the following are now shared across all backends: `test_all_reduce`, `test_average_gradients`, `test_donation`, `test_shard_linear`, `test_all_gather`.
In test_shard_linear, since we don’t have quantized matmuls on CUDA yet, the quantized variant runs only when CUDA is not available.

Test: `mlx.launch -n 8 mlx/python/tests/nccl_test_distributed.py`

### DIFF
--- a/mlx/backend/no_cpu/primitives.cpp
+++ b/mlx/backend/no_cpu/primitives.cpp
@@ -138,6 +138,7 @@ NO_CPU_MULTI(AllReduce)
 NO_CPU_MULTI(AllGather)
 NO_CPU_MULTI(Send)
 NO_CPU_MULTI(Recv)
+NO_CPU_MULTI(ReduceScatter)
 } // namespace distributed
 
 } // namespace mlx::core
--- a/mlx/backend/no_gpu/primitives.cpp
+++ b/mlx/backend/no_gpu/primitives.cpp
@@ -164,6 +164,7 @@ NO_GPU_MULTI(AllReduce)
 NO_GPU_MULTI(AllGather)
 NO_GPU_MULTI(Send)
 NO_GPU_MULTI(Recv)
+NO_GPU_MULTI(ReduceScatter)
 } // namespace distributed
 
 } // namespace mlx::core


python/src/distributed.cpp
```diff
@@ -301,4 +301,32 @@ void init_distributed(nb::module_& parent_module) {
         Returns:
           array: The array that was received from ``src``.
       )pbdoc");
+
+  m.def(
+      "reduce_scatter",
```
awni: I'm wondering about the name. We use `all_sum` (instead of `all_reduce`) to indicate it's a sum. Maybe we should use `sum_scatter` here to be more consistent? Wdyt?

python/src/distributed.cpp
```diff
@@ -301,4 +301,32 @@ void init_distributed(nb::module_& parent_module) {
...
+      Currently supported only for NCCL backend.
+      Args:
+        x (array): Input array.
+        group (Group): The group of processes that will participate in the
+          reduction. If set to ``None`` the global group is used. Default:
+          ``None``.
+        stream (Stream, optional): Stream or device. Defaults to ``None``
+          in which case the default stream of the default device is used.
+      Returns:
+        array: The reduced scattered array of shape x.shape // group.size.
```
awni:
```suggestion
        array: The output array with shape ``[x.shape[0] // group.size, *x.shape[1:]]``.
```

I think the shape comment is a bit wrong. It's sliced only on the first dimension right?

python/src/distributed.cpp
```diff
@@ -301,4 +301,32 @@ void init_distributed(nb::module_& parent_module) {
...
+      },
+      "x"_a,
+      nb::kw_only(),
+      "group"_a = nb::none(),
+      "stream"_a = nb::none(),
+      nb::sig(
+          "def reduce_scatter(x: array, *, group: Optional[Group] = None, stream: Union[None, Stream, Device] = None) -> array"),
+      R"pbdoc(
+      Reduce scatter.
+      Reduce scatter the ``x`` arrays from all processes in the group. Size of x should be divisible by group size.
```
awni:
```suggestion

      Sum and scatter ``x``  across all processes in the group. The size of ``x`` should be divisible by the group size.
```

awni: It would be nice to add a little line here like:

This equivalent to `all_sum(x)[rank * chunk_size : (rank+1) * chunk_size]` where `chunk_size = x.shape[0] // group.size`` but can be faster.

mlx/backend/cuda/distributed.cu
```diff
@@ -53,4 +53,64 @@ void AllReduce::eval_gpu(
...
+    const std::vector<array>& inputs,
+    std::vector<array>& outputs) {
+  assert(inputs.size() == 1);
+  assert(outputs.size() == 1);
+
+  auto& s = stream();
+  auto& encoder = cu::get_command_encoder(s);
+
+  auto ensure_contiguous = [&s, &encoder](const array& x) {
+    if (x.flags().contiguous && x.strides()[x.ndim() - 1] == 1) {
```
awni: Can you explain that condition? I'm not sure it's correct given the rest of the code.

python/src/distributed.cpp
nastya236: Yes, I would agree.. I think it will be more consistent.

python/src/distributed.cpp
nastya236: Yes, you are right, changing it!

mlx/backend/cuda/distributed.cu
nastya236: You are right, it is not needed.. I did it accidentally tbh, removed it.

mlx/backend/cuda/distributed.cu
```diff
@@ -53,4 +53,64 @@ void AllReduce::eval_gpu(
...
+    const std::vector<array>& inputs,
+    std::vector<array>& outputs) {
+  assert(inputs.size() == 1);
+  assert(outputs.size() == 1);
+
+  auto& s = stream();
+  auto& encoder = cu::get_command_encoder(s);
+
+  auto ensure_contiguous = [&s, &encoder](const array& x) {
+    if (x.flags().contiguous) {
```
awni: I think it should be `row_contiguous` (which is different than regular contiguous).

python/src/distributed.cpp
nastya236: I did the same as for `AllReduce` by adding a reduction op, let me know if you think that it is not needed and single `sum_scatter` is enough.

python/src/distributed.cpp
```diff
@@ -301,4 +301,36 @@ void init_distributed(nb::module_& parent_module) {
...
+        return mx::distributed::sum_scatter(to_array(x), group, s);
+      },
+      "x"_a,
+      nb::kw_only(),
+      "group"_a = nb::none(),
+      "stream"_a = nb::none(),
+      nb::sig(
+          "def sum_scatter(x: array, *, group: Optional[Group] = None, stream: Union[None, Stream, Device] = None) -> array"),
+      R"pbdoc(
+      Sum and scatter ``x`` across all processes in the group and leave the summed result sharded along the first axis across ranks.
```
awni:
```suggestion
      Sum ``x`` across all processes in the group and shard the result along the first axis across ranks.
```

Nit, make the first sentence a little shorter so it displays nicely in the docs table.

mlx/distributed/ops.cpp
```diff
@@ -157,4 +157,30 @@ array recv_like(
...
+    const array& x,
+    std::optional<Group> group_ /* = std::nullopt */,
+    StreamOrDevice s /* = {} */) {
+  auto group = to_group(group_);
+  if (group.size() == 1) {
+    return x;
+  }
+  if (x.shape()[0] % group.size() != 0) {
+    std::ostringstream msg;
+    msg << "Invalid shape=" << x.shape() << " for a group of size "
```
awni: Nit: put the op in the error message to help locate it.
```suggestion
    msg << "[sum_scatter] Invalid shape=" << x.shape() << " for a group of size "
```

mlx/backend/cuda/distributed.cu
```diff
@@ -53,4 +53,71 @@ void AllReduce::eval_gpu(
...
+      copy_gpu(in, out, CopyType::General, s);
+      return {out, out};
+    } else if (in.is_donatable()) {
+      out.copy_shared_buffer(in);
+      return {in, out};
+    } else {
+      out.set_data(allocator::malloc(out.nbytes()));
+      return {in, out};
+    }
+  };
```
awni: The output is smaller than the input right? So I think the donation and copying conditions might be off here? It might not make sense to copy / donate the input into the output?

mlx/distributed/mpi/mpi.cpp
```diff
@@ -465,6 +465,10 @@ class MPIGroup : public GroupImpl {
     });
   }
 
+  void sum_scatter(const array& input, array& output, Stream stream) override {
+    throw std::runtime_error("[mpi] Reduce scatter not yet implemented.");
```
awni:
```suggestion
    throw std::runtime_error("[mpi] sum_scatter not yet implemented.");
```

mlx/backend/cuda/distributed.cu
nastya236: Yes, you are right, it is broken.. 
It should be just ensure `row_contiguous` for the input, and always malloc the output:
```cpp
  auto ensure_contiguous = [&s, &encoder](const array& x) {
    if (x.flags().row_contiguous) {
      return x;
    } else {
      array x_copy = contiguous_copy_gpu(x, s);
      encoder.add_temporary(x_copy);
      return x_copy;
    }
  };

  auto input = ensure_contiguous(inputs[0]);
  outputs[0].set_data(allocator::malloc(outputs[0].nbytes()));
```

mlx/distributed/mpi/mpi.cpp
nastya236: Thank you!

awni (approved):
Looks great! Will merge when tests clear!

nastya236:
I fixed a typo, sorry about that. It should pass now. Thanks for reviewing!

awni:
The tests failed since your last push. It looks like it's trying to initialize nccl on mac.. can you see the failures?

nastya236:
Finally everything is fixed :)

PR #2731, opened 2025-11-04
only build for macos 14 and up

awni:
Deprecate support for macos 13.5. We now support 14 and up (14, 15, 26). Still no specific build for 26 but we will add one soon.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -127,7 +127,7 @@ if(MLX_BUILD_METAL)
   message(STATUS "Building with macOS SDK version ${MACOS_SDK_VERSION}")
 
   set(METAL_CPP_URL
-      https://developer.apple.com/metal/cpp/files/metal-cpp_macOS15_iOS18.zip)
+      https://developer.apple.com/metal/cpp/files/metal-cpp_26.zip)
 
   if(NOT CMAKE_OSX_DEPLOYMENT_TARGET STREQUAL "")
     if(${CMAKE_OSX_DEPLOYMENT_TARGET} LESS 14.0)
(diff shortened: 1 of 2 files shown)


madrob:
Looks like the build fails because it can't find MTL::LanguageVersion4_0.

Earlier in the build we have

```
+ xcodebuild -showComponent MetalToolchain
Asset Path: /System/Library/AssetsV2/com_apple_MobileAsset_MetalToolchain/4ab058bc1c53034b8c0a9baca6fba2d2b78bb965.asset/AssetData
Build Version: 17A324
Status: installed
Toolchain Identifier: com.apple.dt.toolchain.Metal.32023
Toolchain Search Path: /private/var/run/com.apple.security.cryptexd/mnt/com.apple.MobileAsset.MetalToolchain-v17.1.324.0.HxIuH5
```

Do we need a different version to build for this?

madrob:
Adding Python 3.14 on macOS causes the following error:

https://github.com/ml-explore/mlx/actions/runs/19078758690/job/54501381735#step:5:40
```
  × No solution found when resolving dependencies:
  ╰─▶ Because all versions of tensorflow have no wheels with a matching Python
      ABI tag (e.g., `cp314`) and you require tensorflow, we can conclude that
      your requirements are unsatisfiable.

      hint: Pre-releases are available for `tensorflow` in the requested
      range (e.g., 2.20.0rc0), but pre-releases weren't enabled (try:
      `--prerelease=allow`)

      hint: You require CPython 3.14 (`cp314`), but we only found wheels
      for `tensorflow` (v2.20.0) with the following Python ABI tags: `cp39`,
      `cp310`, `cp311`, `cp312`, `cp313`
```

PR #2733, opened 2025-11-04
Make cpu binary_op easily accessible

angeloskath:
Similar to the GPU side of things this moves the high level `binary(...)` into `mlx::core` renamed as `binary_op_cpu` and `comparison_op_cpu` etc.

The point is to simplify things like the one in `matmul.cpp`. Not many around so maybe we don't want to merge if we thing this pollutes the namespace (but we have all the lower-level stuff in `mlx::core` 🤷‍♂️ )

### DIFF
--- a/mlx/backend/cpu/binary.cpp
+++ b/mlx/backend/cpu/binary.cpp
@@ -179,7 +179,7 @@ void LogAddExp::eval_cpu(const std::vector<array>& inputs, array& out) {
   assert(inputs.size() == 2);
   auto& a = inputs[0];
   auto& b = inputs[1];
-  binary_float_cpu(a, b, out, detail::LogAddExp(), stream());
+  binary_float_op_cpu(a, b, out, detail::LogAddExp(), stream());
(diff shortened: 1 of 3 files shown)


awni (approved):
Nice. I like the consistency.

Seems like we are converging on a scheme like ``op_{cpu|gpu}{_inplace}( ...)` for back end ops that need to be used during eval.

angeloskath:
> Seems like we are converging on a scheme like ``op_{cpu|gpu}{_inplace}( ...)` for back end ops that need to be used during eval.

Yeah exactly. On that note, I should have asked on the main one. Should I rename `binary_float_cpu` to `binary_float_op_cpu` ? Which one do you like best. Rethinking it I think I should ...

awni:
I would vote for consistency for sure. I kind of prefer just `binary` as opposed to `binary_op` (which is what we have for `unary` for CPU. But then for metal it's all `unary_op`, `binary_op`, etc.

PR #2739, opened 2025-11-06
minor debugging for publishing

madrob:
## Proposed changes

Print hashes when uploading to PyPI
Skip already existing artifacts for Test-PyPI

## Checklist

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -15,7 +15,7 @@ jobs:
     outputs:
       pypi_env: ${{ github.event_name == 'push' && 'pypi' || 'test-pypi' }}
       pypi_url: ${{ github.event_name == 'push' && 'https://upload.pypi.org/legacy/' || 'https://test.pypi.org/legacy/' }}
-      pypi_dups: ${{ github.event_name == 'push' && 'true' || 'false' }}
+      skip_duplicates: ${{ github.event_name == 'push' && 'false' || 'true' }}
     steps:
       - name: Set publishing variables
         run: echo "Publishing setup complete"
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -15,6 +15,7 @@ jobs:
     outputs:
       pypi_env: ${{ github.event_name == 'push' && 'pypi' || 'test-pypi' }}
       pypi_url: ${{ github.event_name == 'push' && 'https://upload.pypi.org/legacy/' || 'https://test.pypi.org/legacy/' }}
+      pypi_dups: ${{ github.event_name == 'push' && 'true' || 'false' }}
```
awni: I'm a bit confused by this variable. How is it used to determine if there is already a package of the same name on pypi?

madrob: I actually got it backwards! When we are publishing to pypi (on a tag pushed event) we want to fail on duplicates, when publishing to test-pypi (on a manual workflow launch event) it's fine to skip when there are duplicates.

For regular releases there shouldn't be duplicates, but for the test releases since we're not incrementing the version number in between then the releases could be repeating (until we delete them I guess?)

madrob: Tried to make it more clear in 88a9841

madrob:
I'm pretty sure the failure at https://github.com/ml-explore/mlx/actions/runs/19140474034/job/54704171529 is not related to this change set, but I don't know if it is a hardware issue or machine setup issue or just a flaky test.

awni:
I agree that failure seems unrelated. Very strange. Those kinds of failures that only happen on circle are almost impossible  to debug without ssh access :(

zcbenz:
For GitHub-hosted runners people can use https://github.com/mxschmitt/action-tmate to ssh into the machine. Not sure if it is usable for self-hosted runners or whether Apple's policy allows it.

PR #2740, opened 2025-11-06
prettify CI output

madrob:
## Proposed changes

Use workflow commands to group log lines and make output easier to read.
https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-commands#grouping-log-lines

## Checklist

### DIFF
--- a/.github/actions/build-cuda-release/action.yml
+++ b/.github/actions/build-cuda-release/action.yml
@@ -15,6 +15,7 @@ runs:
         MLX_BUILD_STAGE: 2
         CMAKE_ARGS: -DMLX_BUILD_CUDA=ON -DCMAKE_CUDA_COMPILER=${{ inputs.nvcc-location }}
       run: |
+        echo "::group::Build packages"
         pip install auditwheel build patchelf setuptools
         python setup.py clean --all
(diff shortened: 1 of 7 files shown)


zcbenz:
@madrob Is it possible to split the single step into multiple steps in Actions? i.e. Instead of seeing a single `build-cuda` step that takes 44m, make the Actions show time spent for each step.

Screenshot 2025-11-07 at 8 28 06

It would make things easier to debug if something went wrong. For example we used to have a problem that a single test took 30m to run in CI, with current UI it would be much harder to notice the problem.

madrob:
Yea, there's an open GitHub feature request to make each sub-step in a composite action have it's own timing information and header, but that's been around for a while and doesn't look like it will get any traction. We'd have to split things in the primary workflow which is totally doable but would require somebody with more knowledge of how MLX works to do a better job than I would, I think. Maybe just `build` and `test` make sense as the division for now. I don't plan to make these changes, but they would definitely be appreciated!

awni:
This needs a rebase.

awni:
Closing this since it's quite dated / getting incorporated into subsequent PRs.

PR #2741, opened 2025-11-07
Compatibility with pip-installed openmpi

pcuenca:
## Proposed changes

When installing openmpi using `pip` in a virtual environment, the installation looks like this (incomplete):

```
.venv
├── bin
│   ├── mpirun
│   ├── ompi_info
├── lib
│   ├── libmpi.40.dylib  # <- not libmpi.dylib
```

However, `mpi.cpp` uses a hardcoded name, `libmpi.dylib`, to `dlopen()`. The installation does not provide a symlink.

On the Python side, `distributed_run` was already special-casing homebrew installations by explicitly passing `DYLD_LIBRARY_PATH` to `mpirun`. This PR:

- Extends the existing mechanism so `DYLD_LIBRARY_PATH` is sent in all cases, as long as an mpi library is found.
- Derives the library name by examining `ompi_info` with `otool`. Other common tools, such as `mpirun`, are not linked against the shared library.
- Sends the library name as an envvar for `mpi.cpp` to use.

This should result in a more robust environment detection and increased compatibility.

## Checklist

## Manual Tests

- [x] macOS with pip
- [x] macOS with homebrew
- [x] macOS with conda
- [x] Linux (CPU) with pip
- [x] Linux (CPU) with conda (`conda install openmpi`; `conda install conda-forge::openmpi` did not put an `mpirun` executable anywhere in my environment)

### DIFF
--- a/python/mlx/distributed_run.py
+++ b/python/mlx/distributed_run.py
@@ -419,7 +419,7 @@ def launch_mpi(parser, hosts, args, command):
         dyld = Path(mpirun).parent.parent / "lib"
         args.env = [
             f"DYLD_LIBRARY_PATH={str(dyld)}",
-            f"MPI_LIBNAME={mpi_libname}",
+            f"MLX_MPI_LIBNAME={mpi_libname}",
         ] + args.env
 
     log(args.verbose, f"Using '{mpirun}'")


python/mlx/distributed_run.py
```diff
@@ -386,15 +386,35 @@ def node_thread(rank, host, hostfile, input_queue):
         t.join()
 
 
+def get_mpi_libname():
+    try:
+        ompi_info = run(["which", "ompi_info"], check=True, capture_output=True)
```
pcuenca, line 392: I'm assuming `ompi_info` is also installed with conda, but haven't verified yet. Note that if it were not, we would return `None` and the process would follow without adding the environment vars, so everything would work as before.

pcuenca:
If this solution is acceptable, I can tweak the docs a bit.

madrob:
Does this suggest that instead of `brew install openmpi` in the CI setup we should python install it?

pcuenca:
This works with `brew`, `pip`, and `conda`. Needs some adjustments for Linux (cpu backend).

pcuenca:
As updated in the initial description, I tested the following combinations:

- [x] macOS with pip
- [x] macOS with homebrew
- [x] macOS with conda
- [x] Linux (CPU) with pip
- [x] Linux (CPU) with conda (`conda install openmpi`; `conda install conda-forge::openmpi` did not put an `mpirun` executable anywhere in my environment)

angeloskath (approved):
Thanks! This looks great!

I 'll merge after the tests clear.

angeloskath:
Some random thought. Perhaps we should change the environment variable to `MLX_MPI_LIBNAME` since I don't think it is standard and we could avoid collisions. I am fine with `MPI_LIBNAME` too though.

pcuenca:
> change the environment variable to `MLX_MPI_LIBNAME`

Good call, thank you!

PR #2743, opened 2025-11-07
[CUDA] cuDNN forward attention

zcbenz:
This PR adds cuDNN implementation for the `fast.scaled_dot_product_attention` op, which is currently disabled behind the `MLX_CUDA_USE_CUDNN_SPDA` env.

It is not very useful for inference at the moment because:
1. the overhead of cuDNN graph is very high;
2. the cuDNN graph is currently recreated whenever sequence length increases.

I'll create follow-up PRs to:
1. implement backward attention and use cuDNN attention for training;
2. implement missing features (array masks, sinks, ...);
3. investigate ways that avoid recreating graph when sequence length changes (https://github.com/pytorch/pytorch/pull/155958);
4. use cuDNN for batch inference;
5. remove duplicate code between cuDNN conv and attention code, I probably have to update the conv code to use new APIs.

### DIFF
--- a/mlx/backend/cuda/scaled_dot_product_attention.cpp
+++ b/mlx/backend/cuda/scaled_dot_product_attention.cpp
@@ -3,6 +3,7 @@
 #include "mlx/backend/cuda/cudnn_utils.h"
 #include "mlx/backend/cuda/device.h"
 #include "mlx/backend/cuda/lru_cache.h"
+#include "mlx/backend/gpu/copy.h"
 #include "mlx/fast_primitives.h"
 #include "mlx/transforms_impl.h"
 
@@ -43,6 +44,19 @@ void set_tensor_attrs(
   tensor->set_uid(uid)
       .set_dim({x.shape().begin(), x.shape().end()})
       .set_stride(normalized_strides(x));
+}
+
+array prepare_sdpa_input(const array& x, Stream s) {
+  // SDPA kernel's requirements on inputs:
+  // 1. last dim's stride be 1;
+  // 2. pointer be aligned.
+  if (x.strides(-1) != 1 || get_alignment(x) < 16) {
+    array x_copy = contiguous_copy_gpu(x, s);
+    auto& encoder = cu::get_command_encoder(s);
+    encoder.add_temporary(x_copy);
+    return x_copy;
+  }
+  return x;
 }
 
 constexpr int QKV_NDIM = 4;
@@ -167,11 +181,6 @@ bool supports_sdpa_cudnn(
     return false;
   }
 
-  // TODO: Do contiguous copy for inputs.
-  if (q.strides(-1) != 1 || k.strides(-1) != 1 || v.strides(-1) != 1) {
-    return false;
-  }
-
   Dtype dtype = q.dtype();
   return dtype == float16 || dtype == bfloat16;
 }
@@ -290,9 +299,9 @@ void ScaledDotProductAttention::eval_gpu(
 
   auto& s = stream();
 
-  const auto& q = inputs[0];
-  const auto& k = inputs[1];
-  const auto& v = inputs[2];
+  array q = prepare_sdpa_input(inputs[0], s);
+  array k = prepare_sdpa_input(inputs[1], s);
+  array v = prepare_sdpa_input(inputs[2], s);
   bool has_mask = inputs.size() - has_sinks_ > 3;
   bool has_arr_mask = has_mask && !do_causal_;
 


awni:
Did you time prefill at all? I'd imagine the graph overhead there would not matter as much?

For decode shouldn't we still route to the custom `sdpa_vector` implementation? I imagine that will be faster?

zcbenz:
Ah I totally forgot about prefill. I have updated this PR to use `sdpa_cudnn` for prefill and `sdpa_vector` for decoding, it is enabled by default but can be turned off with `MLX_CUDA_USE_CUDNN_SPDA=0`.

For long prompt it shows 30% improvement on A100:

```console
# mlx_lm.benchmark --model meta-llama/Meta-Llama-3.1-8B --p 2049 -g 128 -b 1 -n 4
Running warmup..
Timing with prompt_tokens=2049, generation_tokens=128, batch_size=1.
Trial 1:  prompt_tps=13290.771, generation_tps=92.388, peak_memory=20.444
Trial 2:  prompt_tps=13329.313, generation_tps=92.252, peak_memory=20.444
Trial 3:  prompt_tps=13370.043, generation_tps=92.463, peak_memory=20.444
Trial 4:  prompt_tps=13303.649, generation_tps=92.389, peak_memory=20.444
Averages: prompt_tps=13323.444, generation_tps=92.373, peak_memory=20.444
# MLX_CUDA_USE_CUDNN_SPDA=0 mlx_lm.benchmark --model meta-llama/Meta-Llama-3.1-8B --p 2049 -g 128 -b 1 -n 4
Running warmup..
Timing with prompt_tokens=2049, generation_tokens=128, batch_size=1.
Trial 1:  prompt_tps=10531.077, generation_tps=92.438, peak_memory=21.182
Trial 2:  prompt_tps=10507.410, generation_tps=92.017, peak_memory=21.706
Trial 3:  prompt_tps=10509.630, generation_tps=92.432, peak_memory=21.707
Trial 4:  prompt_tps=10517.068, generation_tps=92.459, peak_memory=21.707
Averages: prompt_tps=10516.296, generation_tps=92.336, peak_memory=21.576
```

Even for very short prompt there is still performance gain:

```console
# mlx_lm.benchmark --model meta-llama/Meta-Llama-3.1-8B --p 8 -g 128 -b 1 -n 4
Running warmup..
Timing with prompt_tokens=8, generation_tokens=128, batch_size=1.
Trial 1:  prompt_tps=236.468, generation_tps=99.045, peak_memory=16.122
Trial 2:  prompt_tps=237.316, generation_tps=99.044, peak_memory=16.122
Trial 3:  prompt_tps=236.706, generation_tps=99.064, peak_memory=16.122
Trial 4:  prompt_tps=237.126, generation_tps=99.064, peak_memory=16.123
Averages: prompt_tps=236.904, generation_tps=99.054, peak_memory=16.122
# MLX_CUDA_USE_CUDNN_SPDA=0 mlx_lm.benchmark --model meta-llama/Meta-Llama-3.1-8B --p 8 -g 128 -b 1 -n 4
Running warmup..
Timing with prompt_tokens=8, generation_tokens=128, batch_size=1.
Trial 1:  prompt_tps=226.593, generation_tps=99.118, peak_memory=16.113
Trial 2:  prompt_tps=227.747, generation_tps=99.123, peak_memory=16.114
Trial 3:  prompt_tps=227.008, generation_tps=99.126, peak_memory=16.114
Trial 4:  prompt_tps=227.179, generation_tps=98.821, peak_memory=16.114
Averages: prompt_tps=227.132, generation_tps=99.047, peak_memory=16.114
```

However this benchmark warms up things before timing, so the cost of creating cuDNN graph is hidden. If I remove the warmup the result becomes subtle:

```console
# mlx_lm.benchmark --model meta-llama/Meta-Llama-3.1-8B --p 2049 -g 128 -b 1 -n 4
Timing with prompt_tokens=2049, generation_tokens=128, batch_size=1.
Trial 1:  prompt_tps=2890.866, generation_tps=92.512, peak_memory=17.587
Trial 2:  prompt_tps=13196.311, generation_tps=92.495, peak_memory=20.444
Trial 3:  prompt_tps=13182.864, generation_tps=92.404, peak_memory=20.444
Trial 4:  prompt_tps=13169.970, generation_tps=92.475, peak_memory=20.444
Averages: prompt_tps=10610.003, generation_tps=92.472, peak_memory=19.730
# MLX_CUDA_USE_CUDNN_SPDA=0 mlx_lm.benchmark --model meta-llama/Meta-Llama-3.1-8B --p 2049 -g 128 -b 1 -n 4
Timing with prompt_tokens=2049, generation_tokens=128, batch_size=1.
Trial 1:  prompt_tps=3321.275, generation_tps=92.462, peak_memory=18.166
Trial 2:  prompt_tps=10478.404, generation_tps=92.432, peak_memory=21.706
Trial 3:  prompt_tps=10484.616, generation_tps=92.473, peak_memory=21.707
Trial 4:  prompt_tps=10481.885, generation_tps=92.422, peak_memory=21.707
Averages: prompt_tps=8691.545, generation_tps=92.447, peak_memory=20.822
```

Using cuDNN makes time to first output longer, in exchange for faster upcoming prefills. But for common use cases I think they only need to prefill once?

python/tests/test_fast_sdpa.py
```diff
@@ -551,10 +551,15 @@ def test_sdpa_vector_batched(self):
 class TestSDPA(mlx_tests.MLXTestCase):
     @property
     def dtypes(self):
-        return ["float32", "float16"] if mx.metal.is_available() else ["float32"]
+        if mx.metal.is_available():
+            return ["float32", "float16"]
+        elif mx.cuda.is_available():
+            return ["float16"]
+        else:
+            return ["float32"]
```
awni: I'd say let's just check all the types for every backend here. For cases where there is no fused primitive it will check the fallback which is a good thing to check IMO.

python/tests/test_fast_sdpa.py
```diff
@@ -551,10 +551,15 @@ def test_sdpa_vector_batched(self):
...
 
     def test_sdpa(self):
-        if not mx.metal.is_available():
+        if not mx.metal.is_available() and not mx.cuda.is_available():
```
awni: nit: this could be `mx.is_available(mx.gpu)`

awni:
For irregular prefill sizes this gives a massive improvement on B200:

```python
mlx_lm.benchmark --model meta-llama/Meta-Llama-3.1-8B --p 2048 -g 128 -b 1 -n 4
```

Pre: Averages: prompt_tps=27617.508, generation_tps=230.929, peak_memory=20.798
Post: Averages: prompt_tps=58095.697, generation_tps=231.540, peak_memory=19.748

mlx/backend/cuda/scaled_dot_product_attention.cpp
```diff
@@ -0,0 +1,299 @@
...
+    // FIXME: Causal mask generates wrong results when L_Q != L_K.
+    if (q.shape(2) != k.shape(2)) {
+      return false;
+    }
+  }
+
+  // Only use cuDNN for prefilling.
+  if (q.shape(2) != k.shape(2)) {
+    return false;
+  }
```
awni, line 176: That's not a great condition because when we prefill in steps we can have cases where `q.shape != k.shape`. Currently in mlx-lm we prefill in steps of 2048. So for a prompt of 4096, on the second step the k length would be 4096 and the q length would be 2048.

mlx/backend/cuda/scaled_dot_product_attention.cpp
```diff
@@ -0,0 +1,299 @@
...
+
+  if (supports_sdpa_cudnn(q, k, v, has_mask, do_causal_, s)) {
+    sdpa_cudnn(q, k, v, scale_, out, do_causal_, s);
+  } else {
+    if (has_sinks_) {
+      sdpa_vector(q, k, v, scale_, out, do_causal_, inputs.back(), s);
+    } else {
+      sdpa_vector(q, k, v, scale_, out, do_causal_, std::nullopt, s);
+    }
+  }
```
awni: This might be more natural to first check if it supports vector and route if it does. That way you don't need to ensure `supports_sdpa_cudnn` also implicitly contains the condition to not run for vectors.

mlx/backend/cuda/scaled_dot_product_attention.cpp
```diff
@@ -0,0 +1,299 @@
...
+      {fe::BehaviorNote_t::SUPPORTS_CUDA_GRAPH_NATIVE_API});
+  CHECK_CUDNN_FE_ERROR(graph.check_support(handle));
+  CHECK_CUDNN_FE_ERROR(graph.build_plans(handle));
+
+  return graph;
+}
+
+} // namespace
+
+bool supports_sdpa_cudnn(
```
awni, line 145: I feel like this is missing some checks on the head dimension (and maybe other dimensions?)?

mlx/backend/cuda/scaled_dot_product_attention.cpp
zcbenz: I need to make cuDNN work with dynamic sequence length to be able to remove this condition. But I'm still not sure if it is feasible with cuDNN, the API https://github.com/pytorch/pytorch/pull/155958 uses was designed to be be used with ragged tensors so it might not deliver best performance.

None of the popular inference engines seems to be using cuDNN for attention so I think we might have to add another backend for inference, I'll investigate after making training work.

zcbenz:
Thanks for the reviews, I have updated the code.

mlx/backend/cuda/scaled_dot_product_attention.cpp
```diff
@@ -0,0 +1,312 @@
...
+  // D_qk and D_v must be a multiple of 8 with maximum value 128.
+  if ((q.shape(-1) % 8 != 0) || (q.shape(-1) > 128) || (v.shape(-1) % 8 != 0) ||
+      (v.shape(-1) > 128)) {
+    return false;
+  }
+
+  // TODO: Do contiguous copy for inputs.
+  if (q.strides(-1) != 1 || k.strides(-1) != 1 || v.strides(-1) != 1) {
+    return false;
+  }
```
awni: I would not check the strides here as they are set after the graph is evaluated (not when it is built). This condition will basically always return true, even if the inputs might not be contiguous in the last dimension during the eval.

awni: As a corollary I would implement the contiguous copies in the `eval_gpu` if they are needed. Otherwise it will be a bug when you get unsupported strides.

zcbenz: Sorry I forgot strides are set in evaluation, I have updated the code.

This is actually not the first time I was bitten by this, before evaluation the strides is not initialized and it could be anything, do you think if it makes sense adding a debug time assertion in `strides()`?

awni: > do you think if it makes sense adding a debug time assertion in strides()

Yea that seems reasonable to me.

PR #2746, opened 2025-11-10
Fix release builds

awni:
Fixing the github actions release build.

### DIFF
--- a/.github/actions/build-linux-release/action.yml
+++ b/.github/actions/build-linux-release/action.yml
@@ -1,0 +1,33 @@
+name: 'Build Linux wheel'
+description: 'Build Linux wheel'
+
+inputs:
+  build-backend:
+    description: 'Build the backend mlx-cpu package'
+    type: boolean
+    required: false
+    default: false
+
+runs:
+  using: "composite"
+  steps:
+    - name: Generate package stubs
+      shell: bash
+      run: |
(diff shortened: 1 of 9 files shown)


.github/actions/build-linux-release/action.yml
```diff
@@ -0,0 +1,32 @@
...
+    default: 'false'
+
+runs:
+  using: "composite"
+  steps:
+    - name: Generate package stubs
+      shell: bash
+      run: |
+        pip install -e ".[dev]" -v
+        uv pip install typing_extensions
```
madrob: The linux environments weren't using `uv` because it had problems with make finding nanobind. Maybe this will still work, but I'd feel more confident if we're not mixing toolchains.

awni: I didn't change this.. it's just moved

awni: Not moved but copied from the linux-build action

madrob: Are we looking at different versions? https://github.com/ml-explore/mlx/blob/main/.github/actions/build-linux/action.yml#L36-L37 does not use `uv`

awni: No you're right, I was copying from a different file. Fixed.

angeloskath (approved):
🙏

PR #2751, opened 2025-11-11
Linux on arm

awni:
Step 1 to get arm-based machines (DGX spark) in PyPi.

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -17,13 +17,13 @@ jobs:
     strategy:
       matrix:
         runner:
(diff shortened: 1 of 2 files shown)


.github/workflows/pull_request.yml
```diff
@@ -14,7 +14,12 @@ jobs:
       - uses: pre-commit/action@v3.0.1
 
   linux_build_and_test:
-    runs-on: ubuntu-22.04
+    strategy:
```
madrob, line 17: Add `fail-fast: false` and a `name:` key so that we can tell them apart (see fedora build for example)

PR #2757, opened 2025-11-13
[CUDA] Check CUDA error in synchronize

zcbenz:
When there was fatal error running CUDA kernel, further CUDA calls would fail and `synchronize()` would get stuck because it would be waiting for a callback failed to be registered.


angeloskath (approved):
🙏

PR #2758, opened 2025-11-13
Fix slice with negative strides

### LINKED ISSUE #2755: [BUG] Reversing array order multiple times returns zeros or unexpected numbers
**Describe the bug**
I have an mlx array that I want to reverse 4 times to get back to the array again. This works in numpy but in mlx I get very unexpected answers. Below is an example.

**To Reproduce**

Include code snippet
```
import mlx.core as mx

a = mx.arange(17 * 3).reshape(17, 3)
orig = a*1.0

for _ in range(4):
    a = a[::-1]   # reverse across first dimension

print(np.array_equal(a, orig))  # → False


import numpy as np

a = np.arange(17 * 3).reshape(17, 3)
orig = a*1.0

for _ in range(4):
    a = a[::-1]   # reverse across first dimension

print(np.array_equal(a, orig))  # → True
```

**Expected behavior**
I'd expect both the numpy version and mlx version to return the original array after 4 reverses but only numpy does it correctly. MLX returns random numbers and keeps getting more unstable. 

- MLX v0.29.4

awni:
Closes #2755

### DIFF
--- a/mlx/array.cpp
+++ b/mlx/array.cpp
@@ -167,7 +167,7 @@ void array::copy_shared_buffer(
     const Strides& strides,
     Flags flags,
     size_t data_size,
-    size_t offset /* = 0 */) {
+    int64_t offset /* = 0 */) {
   array_desc_->data = other.array_desc_->data;
   array_desc_->strides = strides;
   array_desc_->flags = flags;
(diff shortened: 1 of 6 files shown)


angeloskath (approved):
🙏

PR #2759, opened 2025-11-13
fix release

awni:
A few changes here:

- Use conda python for release builds so that we don't have the issue of the python binary being compiled on a later macos than the target for the build (as we had it in circle ci)
- Remove releases to test pypi as they are not useful
- Add support dev release as we had it in circle ci
-  Test on more than one macos version as we had it in circle ci (now test on 14 and 15)

### DIFF
--- a/.github/actions/build-macos-release/action.yml
+++ b/.github/actions/build-macos-release/action.yml
@@ -20,14 +20,17 @@ runs:
       env:
         MACOSX_DEPLOYMENT_TARGET: ${{ inputs.macos-target }}
       run: |
-        uv pip install build
-        uv run --no-project setup.py clean --all
-        MLX_BUILD_STAGE=1 uv run -m build -w
+        conda activate env
+        pip install build
+        python setup.py clean --all
+        MLX_BUILD_STAGE=1 python -m build -w
+
     - name: Build backend package
       if: ${{ inputs.build-backend }}
       shell: bash
       env:
         MACOSX_DEPLOYMENT_TARGET: ${{ inputs.macos-target }}
       run: |
-        uv run --no-project setup.py clean --all
-        MLX_BUILD_STAGE=2 uv run -m build -w
+        conda activate env
+        python setup.py clean --all
+        MLX_BUILD_STAGE=2 python -m build -w
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -28,12 +28,17 @@ jobs:
 
   mac_build_and_test:
     if: github.repository == 'ml-explore/mlx'
+    strategy:
+      matrix:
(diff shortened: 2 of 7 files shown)


angeloskath (approved):
👍

.github/workflows/pull_request.yml
```diff
@@ -28,12 +28,17 @@ jobs:
...
+      matrix:
+        macos-target: ["14.0", "15.0"]
     runs-on: [self-hosted, macos]
     needs: check_lint
     steps:
       - uses: actions/checkout@v5
       - uses: ./.github/actions/setup-macos
       - uses: ./.github/actions/build-macos
+        with:
+          macos-target: ${{ matrix.macos-target }}
```
madrob, line 41: Is this more useful than setting an env for the entire step?

.github/actions/build-macos-release/action.yml
```diff
@@ -20,14 +20,17 @@ runs:
...
+
     - name: Build backend package
       if: ${{ inputs.build-backend }}
       shell: bash
       env:
         MACOSX_DEPLOYMENT_TARGET: ${{ inputs.macos-target }}
       run: |
-        uv run --no-project setup.py clean --all
-        MLX_BUILD_STAGE=2 uv run -m build -w
+        conda activate env
```
madrob, line 34: Conda needs to be set up for nightlies as well

.github/workflows/release.yml
```diff
@@ -54,6 +55,7 @@ jobs:
     runs-on: ${{ matrix.runner }}
     env:
       PYPI_RELEASE: 1
```
madrob, line 57: Dos this need to be the opposite of DEV_RELEASE?

.github/workflows/release.yml
```diff
@@ -166,9 +174,7 @@ jobs:
       - name: Publish package distributions to PyPI
         uses: pypa/gh-action-pypi-publish@release/v1
         with:
-          repository-url: ${{ needs.setup.outputs.pypi_url }}
-          skip-existing: ${{ needs.setup.outputs.skip_duplicates }}
-          print-hash: true
+          repository-url: https://upload.pypi.org/legacy/
```
madrob, line 177: This is the default and doesn’t need to be set

.github/workflows/release.yml
awni: No, PYPI_RELEASE is always on for this. And if DEV_RELEASE is also on it will do a dev release with a dev tag included.

.github/workflows/pull_request.yml
awni: By no means, I'd rather set it for the full step, just didn't know it was possible.

PR #2760, opened 2025-11-13
Metal Kernels for Paged Attention & Batched KV Writes

Sohailm25:
## Proposed changes

- Integrate the new paged-attention/paged-KV backend: Metal kernels for paged attention and batched KV writes, shared `_fast` shims, and the Python `mlx.nn.paged_kv` plumbing now keep
  Metal, CPU fallback, and quantized paths in sync so `mlx-lm` can depend on a single API surface.
- Harden the Python fallback overlay path so it accepts rank-3 overlays and never mutates the live KV cache; add CPU-side batch/layer writers that mirror the Metal kernels when the
  native ops reject a configuration.
- Extend CI: add the GPU-focused regression suite (`python/tests/test_paged_attention_gpu.py`) and rename the duplicated `paged_attention_bench` workflow job (`paged_attention_sweep`)
  so YAML linting and pre-commit pass cleanly.

  (No open GitHub issue tracks this work; it’s the kernel side needed for the paged KV + continuous batching PR in `mlx-lm`.)

## Checklist

- [x] I have run `pre-commit run --all-files` to format my code / installed pre-commit prior to committing changes *(ran the repo’s pre-commit hooks on the touched files; `--all-
  files` would reformat unrelated bench tests, so I left those alone)*


awni:
Thanks for the contribution! At the moment this looks a bit early stage for a PR. If you want to keep running with it, I'd consider refining in your own fork. Then you could start an issue or discussion about if it makes sense to contribute this back to MLX.

Generally an interesting direction and I'd be curious to see more benchmarks around where something like this could help and by how much.

Sohailm25:
> Thanks for the contribution! At the moment this looks a bit early stage for a PR. If you want to keep running with it, I'd consider refining in your own fork. Then you could start an issue or discussion about if it makes sense to contribute this back to MLX.
> 
> Generally an interesting direction and I'd be curious to see more benchmarks around where something like this could help and by how much.

@awni broke it down to a smaller first pr, performed a bench, proved 6x TTFT reduction for continous batching. this mlx change adds a fast path (which is currently optional and NOT used by the mlx-lm PR aka it is not dependent)
https://github.com/ml-explore/mlx/pull/2814

PR #2762, opened 2025-11-14
[CUDA] cuDNN backward attention

zcbenz:
This PR uses cuDNN backward attention for `fast::ScaledDotProductAttention::vjp`.

* A new `ScaledDotProductAttentionVJP` primitive is added, but only implemented in CUDA backend.
* For training a `stats` output is generated from the forward attention op, which is required by the backward op.
* The array mask has not been implemented yet - so in actual training the new code may not kick in.
* There are some duplicate code which I will clean up later together with the convolution code.

For training a 0.6B model:

before:

RAM usage: 60511MiB /  81920MiB

```
INFO:root:Model has 596049920 parameters.
INFO:root:step: 100, train_loss: 10.0955, grad_norm: 4.4649, its_per_sec: 2.1158, toks_per_sec: 17332.9950, tokens: 819200
INFO:root:step: 200, train_loss: 8.1339, grad_norm: 1.5591, its_per_sec: 2.6921, toks_per_sec: 22053.9309, tokens: 1638400
INFO:root:step: 300, train_loss: 7.6831, grad_norm: 1.5067, its_per_sec: 2.6923, toks_per_sec: 22055.2519, tokens: 2457600
INFO:root:step: 400, train_loss: 7.5040, grad_norm: 1.3949, its_per_sec: 2.6794, toks_per_sec: 21949.9022, tokens: 3276800
INFO:root:step: 500, train_loss: 7.2268, grad_norm: 1.4708, its_per_sec: 2.6820, toks_per_sec: 21970.5679, tokens: 4096000
```

after:

RAM usage: 52593MiB /  81920MiB

```
INFO:root:Model has 596049920 parameters.
INFO:root:step: 100, train_loss: 10.0885, grad_norm: 4.4325, its_per_sec: 3.0472, toks_per_sec: 24962.3116, tokens: 819200
INFO:root:step: 200, train_loss: 8.1385, grad_norm: 1.6766, its_per_sec: 3.4142, toks_per_sec: 27968.7873, tokens: 1638400
INFO:root:step: 300, train_loss: 7.6979, grad_norm: 2.3700, its_per_sec: 3.4130, toks_per_sec: 27959.5179, tokens: 2457600
INFO:root:step: 400, train_loss: 7.5094, grad_norm: 1.4560, its_per_sec: 3.3883, toks_per_sec: 27756.6980, tokens: 3276800
INFO:root:step: 500, train_loss: 7.2361, grad_norm: 1.8131, its_per_sec: 3.3830, toks_per_sec: 27713.4482, tokens: 4096000
```

### DIFF
--- a/mlx/fast.cpp
+++ b/mlx/fast.cpp
@@ -795,6 +795,7 @@ array scaled_dot_product_attention(
           has_mask,
           has_arr_mask,
           do_causal,
+          is_training,
           output_logsumexp,
           stream)) {
     Shape out_shape{q.shape(0), q.shape(1), q.shape(2), v.shape(-1)};
--- a/mlx/fast_primitives.h
+++ b/mlx/fast_primitives.h
@@ -225,6 +225,7 @@ class ScaledDotProductAttention : public Custom {
       bool has_mask,
       bool has_arr_mask,
       bool do_causal,
+      bool is_training,
       bool output_logsumexp,
       Stream s);
 
--- a/mlx/backend/cuda/scaled_dot_product_attention.cpp
+++ b/mlx/backend/cuda/scaled_dot_product_attention.cpp
@@ -460,6 +460,7 @@ bool ScaledDotProductAttention::use_fallback(
     bool has_mask,
     bool has_arr_mask,
     bool do_causal,
+    bool is_training,
     bool output_logsumexp,
     Stream s) {
   if (s.device == Device::cpu) {
--- a/mlx/backend/metal/scaled_dot_product_attention.cpp
+++ b/mlx/backend/metal/scaled_dot_product_attention.cpp
@@ -378,8 +378,14 @@ bool ScaledDotProductAttention::use_fallback(
     bool has_mask,
     bool has_arr_mask,
     bool do_causal,
+    bool is_training,
     bool output_logsumexp,
     Stream s) {
+  if (is_training) {
+    // It's faster for training on Metal to use the unfused SDPA for both
+    // forward and backward.
+    return true;
+  }
   if (output_logsumexp) {
     return true;
   }
--- a/mlx/backend/no_gpu/primitives.cpp
+++ b/mlx/backend/no_gpu/primitives.cpp
@@ -30,6 +30,7 @@ bool fast::ScaledDotProductAttention::use_fallback(
     bool has_mask,
     bool has_arr_mask,
     bool do_causal,
+    bool is_training,
     bool output_logsumexp,
     Stream s) {
   return true;


mlx/backend/cuda/scaled_dot_product_attention.cu
```diff
@@ -666,6 +667,10 @@ bool supports_sdpa_vector(
     bool has_mask,
     bool has_arr_mask,
     bool do_causal) {
+  if (detail::in_grad_tracing()) {
+    return false;
+  }
```
awni: I think it would be better to avoid checking this global state during eval. It should be fine.. but it makes the eval code slightly more coupled to the function transformation code. And I think it makes the code more flexible to have fewer places where we need to access this global state.

mlx/fast.cpp
```diff
@@ -797,20 +798,71 @@ array scaled_dot_product_attention(
...
+    Shape stats_shape{q.shape(0), q.shape(1), q.shape(2), 1};
+    bool generate_stats = detail::in_grad_tracing();
+    return array::make_arrays(
+        {out_shape, stats_shape},
+        {final_type, float32},
         std::make_shared<ScaledDotProductAttention>(
-            stream, fallback, scale, do_causal, has_sinks),
-        std::move(inputs));
+            stream, fallback, scale, do_causal, has_sinks, generate_stats),
+        std::move(inputs))[0];
```
awni: I would just make a single output array if `generate_stats` is `False`. I think there may be assumptions in some parts of the code that the outputs should always have their data set after `eval_*`. But either way it will just be more efficient for graph building. Making all those vectors can add up..

mlx/fast_primitives.h
```diff
@@ -205,16 +205,18 @@ class RoPE : public Custom {
...
-      bool has_sinks)
-      : Custom(stream, fallback),
+      bool has_sinks,
+      bool generate_stats)
+      : Custom(stream, std::move(fallback)),
         scale_(scale),
         do_causal_(do_causal),
-        has_sinks_(has_sinks) {}
+        has_sinks_(has_sinks),
+        generate_stats_(generate_stats) {}
```
awni: A small nit: `generate_stats` is quite a vague name. I might call it `output_logsumexp` or something like that. I think that's what the "stats" are in this case.

awni (approved):
Very nice, LGTM!

zcbenz:
I just noticed a behavior change in Metal backend by replacing `detail::in_grad_tracing()` with `output_logsumexp`: for training, the fast sdpa was fallbacking to unfused ops for both forward and backward passes, but with this change it would use fused sdpa for forward pass.

I think the change is good for performance and tests are passing, but I want to make sure I'm not missing anything?

awni:
> I think the change is good for performance and tests are passing, but I want to make sure I'm not missing anything?

Actually we added that in the first place because it's faster for training on Metal to use the unfused SDPA for both forward and backward (since the forward and backward can share some computation there).

zcbenz:
Thanks for the info, I updated the code to keep behavior unchanged for Metal backend.

PR #2763, opened 2025-11-14
Use std::optional for mask_arr arg

zcbenz:
Both the API and kernel only support one mask so there is no need to store masks in a std::vector.


awni:
My understanding is this was an intentional design choice to enable masks which require more than one array in the future.

CC @jagrit06 or @angeloskath might have an example of that?

If not then yea we should simplify it.

angeloskath:
Yeah I think Jagrit had a specific thing in mind but I am pro simplifying now and breaking the API later.

PR #2765, opened 2025-11-14
Separate test-linux from build-linux/cuda in GitHub Actions

zcbenz:
Move the tests from `build-linux`/`build-cuda` actions into a new `test-linux` action. This makes it easier to view the CI results, and hopefully it would make the CI configs simpler.

Screenshot 2025-11-14 at 19 36 54

### DIFF
--- a/.github/actions/test-linux/action.yml
+++ b/.github/actions/test-linux/action.yml
@@ -30,7 +30,7 @@ runs:
         DEVICE: cpu
       run: |
         echo "::group::Python tests - CPU"
-        python -m tests discover python/tests -v
+        python -m unittest discover python/tests -v
         echo "::endgroup::"
 
     - name: Run Python tests - GPU


.github/actions/test-linux/action.yml
```diff
@@ -0,0 +1,63 @@
...
+        fi
+        echo "::endgroup::"
+
+    - name: Run Python tests - CPU
+      shell: bash
+      env:
+        DEVICE: cpu
+      run: |
+        echo "::group::Python tests - CPU"
+        python -m tests discover python/tests -v
```
madrob: CPU and GPU test runs use different modules (I'm not sure why, but that's what they have now)
```suggestion
        python -m unittest discover python/tests -v
```

zcbenz: `python -m tests` would run `python/tests/__main__.py` which skips some tests for CUDA backend (some features are still WIP).

Thanks for noticing this, code has been updated.

awni (approved):
LGTM, thanks!

PR #2771, opened 2025-11-14
[CUDA] Use arch specific targets when possible

awni:
Use architecture specific targets (e.g. 100a instead of 100) when they exist.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -89,7 +89,16 @@ class CMakeBuild(build_ext):
         ]
         if build_stage == 2 and build_cuda:
             # Last arch is always real and virtual for forward-compatibility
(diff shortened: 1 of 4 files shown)


setup.py
```diff
@@ -89,7 +89,7 @@ def build_extension(self, ext: CMakeExtension) -> None:
         ]
         if build_stage == 2 and build_cuda:
             # Last arch is always real and virtual for forward-compatibility
-            cuda_archs = ";".join(("70-real", "80-real", "90-real", "100-real", "120"))
+            cuda_archs = ";".join(("75-real", "80-real", "90a-real", "100a-real", "120a-real", "120-virtual"))
```
awni: 70 is deprecated.. so I'm limiting to 75 for the release.

PR #2773, opened 2025-11-15
Use ccache in GitHub Actions

zcbenz:
Also remove steps that clean files - we are not in short of disk space and it would slow things down.

### DIFF
--- a/.github/actions/setup-linux/action.yml
+++ b/.github/actions/setup-linux/action.yml
@@ -28,19 +28,19 @@ runs:
       shell: bash
       run: |
         sudo apt-get update
-        sudo apt-get install -y libblas-dev liblapack-dev liblapacke-dev tzdata zip
+        sudo apt-get install -y libblas-dev liblapack-dev liblapacke-dev zip
 
     - uses: actions/setup-python@v6
       with:
         python-version: ${{ inputs.python-version }}
         cache: 'pip'
 
-    - name: setup python venv
+    - name: Setup Python venv
       shell: bash
       run: |
         python -m venv .venv
         source .venv/bin/activate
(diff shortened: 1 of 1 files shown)


madrob:
When first setting up GitHub Actions, @heckj looked at ccache since it was in the Circle builds but I think he said that when he printed stats at the end of the build it wasn't actually getting used. I would take a look at those again before jumping into it and assuming it is helping.

zcbenz:
With Circle CI the ccache was only enabled for CUDA build, for other builds ccache had very high cache miss and I didn't look further. So far for GitHub Actions I'm getting reasonable hit rate:

Screenshot 2025-11-15 at 12 50 14

---
Before:


b1
b2

---

After:

a1
a2

zcbenz:
This PR is ready - we have 100% cache hit rate.

```
  Cacheable calls:   423 / 423 (100.0%)
    Hits:            423 / 423 (100.0%)
      Direct:        423 / 423 (100.0%)
      Preprocessed:    0 / 423 ( 0.00%)
    Misses:            0 / 423 ( 0.00%)
```

awni (approved):
Nice! Thank you!

heckj:
Nice @zcbenz , thank you! Was the key the —no-build-isolation parameter, if not run into that before, but it’s been a while since I did deep python anything. Curious what I missed for getting usage of ccache during the build.

zcbenz:
Apart from `--no-build-isolation` we also use [a few other tricks](https://github.com/ml-explore/mlx/blob/1bf605d56df3312af6c173547ed4c4c199f11cf6/setup.py#L132-L134) to make ccache work, ccache is fragile - I would not be surprised if some slight environment change broke it before.

PR #2775, opened 2025-11-15
Fix MPI distributed tests with CUDA backend

zcbenz:
Fix the error when running `mpi_test_distributed.py` with CUDA backend:

```
__nv_name_map(17): error: Error in parsing name expression for lowered name lookup. Input name expression was: " mlx :: core :: cu :: gather < complex64_t, uint32_t, 1, 6, int64_t >"
  #pragma nv_mangled_name mlx::core::cu::gather<complex64_t, uint32_t, 1, 6, int64_t>
```

Also fix the problem that GPU tests not actually running.

### DIFF
--- a/.github/actions/test-linux/action.yml
+++ b/.github/actions/test-linux/action.yml
@@ -9,13 +9,18 @@ inputs:
 runs:
   using: "composite"
   steps:
+    - name: Run MPI tests
+      shell: bash
+      run: |
+        echo "::group::MPI tests"
+        mpirun --bind-to none --allow-run-as-root -host localhost:8 -np 8 python python/tests/mpi_test_distributed.py
+        echo "::endgroup::"
+
     - name: Run distributed tests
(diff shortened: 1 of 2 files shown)


.github/actions/test-linux/action.yml
```diff
@@ -9,13 +9,18 @@ inputs:
 runs:
   using: "composite"
   steps:
+    - name: Run MPI tests
```
awni, line 12: I'm wondering why we don't run NCCL tests for CUDA 🤔

awni (approved):
Looks great, thanks!

.github/actions/test-linux/action.yml
zcbenz: NCCL tests currently requires N physical GPUs.

awni: Good point

PR #2777, opened 2025-11-15
Run CI for pushes

zcbenz:
In Github Actions the cache created in pull requests are restricted in the pull request, to make cache work we have to run CI for pushes in the main branch to make cache available to all pull requests:
https://docs.github.com/en/actions/reference/workflows-and-actions/dependency-caching#restrictions-for-accessing-a-cache

It would also help to know a merged change did not cause CI to fail due to conflicts.

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -1,6 +1,6 @@
 name: Build and Test
 
-on: pull_request  
+on: [pull_request, push]
 
 permissions:
   contents: read


madrob:
Can we restrict the runs on push to specific branches?

zcbenz:
Restricting to main branch would be fine.

```yaml
on:
  push:
    branches:
      - main
```

But I found it convenient when I needed to verify if something works in CI I could just push to a branch to upstream repo, instead of creating a WIP pull request like I do now.

For most times people pushed changes to their own folks and then create pull request, so it wouldn't cost CI time. You can see that we usually do not create branches in this repo: https://github.com/ml-explore/mlx/branches

awni (approved):
LGTM!

PR #2778, opened 2025-11-16
[WIP] Cache APT packages in GitHub Actions

zcbenz:


zcbenz:
The size of APT packages cache is over 7GB which is large enough to invalidate our compilation cache. Also tried [apt-fast](https://github.com/ilikenwf/apt-fast) which improves nothing in our case. I guess we have to accept the long setup time.

PR #2782, opened 2025-11-17
Fix building with CUDA < 12.8

zcbenz:
The `cudaStreamGetDevice` API is available since CUDA 12.8, adds a `device` arg to `malloc_async` instead so MLX can build with older CUDA versions.

### DIFF
--- a/mlx/backend/cuda/allocator.cpp
+++ b/mlx/backend/cuda/allocator.cpp
@@ -120,7 +120,7 @@ void copy_to_managed(CudaBuffer& buf) {
 }
 
 Buffer
-CudaAllocator::malloc_impl(size_t size, int device, cudaStream_t stream) {
+CudaAllocator::malloc_async(size_t size, int device, cudaStream_t stream) {
   if (size == 0) {
     return Buffer{new CudaBuffer{nullptr, 0, -1}};
   }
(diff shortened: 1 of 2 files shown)


mlx/backend/cuda/allocator.cpp
```diff
@@ -182,12 +182,13 @@ Buffer CudaAllocator::malloc_impl(size_t size, cudaStream_t stream) {
   return Buffer{buf};
 }
 
-Buffer CudaAllocator::malloc_async(size_t size, cudaStream_t stream) {
-  return malloc_impl(size, stream);
+Buffer
+CudaAllocator::malloc_async(size_t size, int device, cudaStream_t stream) {
+  return malloc_impl(size, device, stream);
 }
```
awni: Is this used? If not, I would remove it .. add it if we need it.

zcbenz: It is used by `cu::malloc_async`:

https://github.com/ml-explore/mlx/blob/35f81728f17d1389f8ea9a295db296fb97763bf2/mlx/backend/cuda/allocator.cpp#L280-L281

I think it makes sense remove it and then just expose malloc_impl though, what do you think?

awni: Yea that makes sense. (i.e. rename `malloc_impl` to `malloc_async` and make it public?)

PR #2783, opened 2025-11-17
Merge build-cuda and build-linux actions

zcbenz:
This PR merges the `build-cuda` action into `build-linux`, and adds arm64 CUDA builds (tests are disabled).

The CUDA and CPU linux builds are also merged into one job with a matrix — please let me know if you don't think it is a good idea.

### DIFF
--- a/mlx/backend/cuda/CMakeLists.txt
+++ b/mlx/backend/cuda/CMakeLists.txt
@@ -137,8 +137,6 @@ if(NOT DEFINED MLX_CUDA_ARCHITECTURES)
   elseif(MLX_CUDA_ARCHITECTURES IN_LIST UPGRADABLE_ARCHITECTURES)
     # Use arch-specific compute capability whenever possible.
     set(MLX_CUDA_ARCHITECTURES "${MLX_CUDA_ARCHITECTURES}a")
-  else()
-    set(MLX_CUDA_ARCHITECTURES "native")
   endif()
 endif()
 message(STATUS "CUDA architectures: ${MLX_CUDA_ARCHITECTURES}")


.github/workflows/pull_request.yml
```diff
@@ -19,19 +19,32 @@ jobs:
...
       fail-fast: false
-    runs-on: ${{ matrix.runner }}
+      matrix:
+        arch: ['x86_64', 'aarch64']
+        toolkit: ['cpu', 'cuda-12.6', 'cuda-12.9']
+    # Use gpu runner for cuda/x86_64 builds, otherwise cpu runners.
+    runs-on: >-
+      ${{ (startsWith(matrix.toolkit, 'cuda') && matrix.arch == 'x86_64' && github.repository == 'ml-explore/mlx') &&
+          'gpu-t4-4-core' ||
+          (matrix.arch == 'aarch64' && 'ubuntu-22.04-arm' || 'ubuntu-22.04') }}
```
madrob: This seems very complex with all of the nested conditionals. Would it be possible to express more simply with some an explicit matrix include or exclude? https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#jobsjob_idstrategymatrixinclude

mlx/backend/cuda/detect_cuda_arch.sh
```diff
@@ -1,5 +1,12 @@
 #!/bin/bash
 
+# If nvidia-smi fails (no NVIDIA GPU or driver), default to 90a.
+if ! nvidia-smi >/dev/null 2>&1; then
+    echo "90a"
+    exit 0
+fi
```
awni: What's the purpose of the fallback here? In theory we shouldn't be calling this if there is no GPU present. If there is no GPU one has to set the architecture explicitly.

.github/workflows/pull_request.yml
awni: I think this will also run the same tests 3 times each on a fork which is probably not desirable.

mlx/backend/cuda/detect_cuda_arch.sh
zcbenz: Previously when there is no GPU this script would return `native` which would then fallback to an ancient number in nvcc. I have changed the cmake setup to fail when there is no GPU and no arch set.

.github/workflows/pull_request.yml
zcbenz: The matrix include/exclude lacks support for conditional expressions and array inputs so we would have to repeat things:

```yaml
matrix:
  toolkit: ['cpu', 'cuda-12.6', 'cuda-12.8']
  include:
    - toolkit: 'cpu'
      runner: 'ubuntu-22.04'
    - toolkit: 'cpu'
      runner: 'ubuntu-22.04-arm'
    - toolkit: 'cuda-12.6'
      runner: 'gpu-t4-core'
    - toolkit: 'cuda-12.6'
      runner: 'ubuntu-22.04-arm'
    - toolkit: 'cuda-12.9'
      runner: 'gpu-t4-core'
    - toolkit: 'cuda-12.9'
      runner: 'ubuntu-22.04-arm'
```

I have changed the workflow to generate matrix dynamically, which is a bit overkill for current settings, but it will help a lot when we test more things (more python versions, more cuda versions, more cuda hardware...).

.github/actions/build-linux/action.yml
```diff
@@ -1,25 +1,40 @@
...
+      run: |
+        if ${{ startsWith(inputs.toolkit, 'cuda') && runner.arch == 'arm64' }} ; then
+          # There is no GPU in arm64 runner, use a common arch.
+          CMAKE_ARGS="$CMAKE_ARGS -DMLX_CUDA_ARCHITECTURES=90a"
+          # Can not build tests when the built executables can not run.
+          CMAKE_ARGS="$CMAKE_ARGS -DMLX_BUILD_TESTS=OFF"
+        fi
+        pip install --no-build-isolation -e ".[dev]" -v
+        # Pass the CMAKE_ARGS to following steps.
+        echo CMAKE_ARGS="$CMAKE_ARGS" > $GITHUB_ENV
```
madrob: It’s a little bit more verbose but I think passing through step output and then reading CMAKE arcs in subsequent steps might be safer

.github/actions/setup-linux/action.yml
```diff
@@ -51,22 +51,21 @@ runs:
...
-        # it's *not* on the default toolkit path.
         PACKAGES: |
           {
             "cuda-12.6": "libcudnn9-dev-cuda-12 cuda-toolkit-12-6",
             "cuda-12.9": "libcudnn9-dev-cuda-12 cuda-toolkit-12-9",
             "cuda-13.0": "libcudnn9-dev-cuda-13 cuda-toolkit-13-0"
           }
       run: |
-        export ARCH=${{ runner.arch == 'arm64' && 'arm64' || 'x86_64' }}
+        export ARCH=${{ runner.arch == 'arm64' && 'sbsa' || 'x86_64' }}
```
madrob: Should we leave a comment here why it’s sbsa instead of arm64 or is that well understood by anybody working in this space?

.github/workflows/build_and_test.yml
```diff
@@ -16,28 +16,62 @@ concurrency:
...
+              }
+              for (let toolkit of cuda_toolkits) {
+                matrix.toolkit.push(toolkit)
+                runner[toolkit] = cuda_runner
+              }
+            }
+            core.setOutput('matrix', JSON.stringify(matrix))
+            core.setOutput('runner', JSON.stringify(runner))
 
   linux_build_and_test:
```
madrob, line 26: Can we add a name: so that all of the matrix builds are distinguishable?

zcbenz: Good idea, I'm just using the OS name which feels most readable: `Build and Test / Linux (cpu, aarch64)`.

Also renamed `pull_request.md` to `build_and_test.yml` so the filename and the workflow name matches.

zcbenz:
I reverted the change that merges `cuda_build_and_test` to `linux_build_and_test`, it does look overcomplicated.

PR #2786, opened 2025-11-17
Remove unneeded tests in nightly build

zcbenz:
Since we are now running tests for the main branch, there is no need to repeat those tests again in the nightly build.


awni (approved):
I think it would be fine to just check the release builds and not run any tests on the nightly. That's what we used to do in circle..

But this is a good step!

PR #2788, opened 2025-11-17
Avoid duplicate CI runs when starting a PR from upstream branch

zcbenz:
Also makes the `check_lint` runs faster.

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -1,6 +1,12 @@
 name: Build and Test
 
-on: [pull_request, push]
+on:
+  pull_request:
+  push:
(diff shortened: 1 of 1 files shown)


awni:
It's kind of odd how it shows the skipped tests in this case. Maybe it would be cleaner to just restrict push to main?

zcbenz:
Yeah it does not look good, I have changed to test main branch only.

awni:
We could add a manual trigger like we have for the release workflow if you want to be able to run the CI on a ml-explore/mlx branch without making a PR.

awni:
Actually the way you have it also works well.

PR #2789, opened 2025-11-18
qqmm

nastya236:
This PR adds a new operation `mx.qqmm`. The current structure is probably neither optimal nor final.

## General comment

1. For inference we want to support: `qqmm(quantized weights, bf16 activations)`.
2. For training (vjp) we unfortunately still need bf16 weights for two reasons:
   - We currently do not have 2D scaling for `nvfp4`, so we need to transpose and quantize again along a different dimension.
   - For `mxfp8`, the recommended recipe is to quantize with 1D blocks and keep two views of the weights (normal and transposed).

Therefore, `mx.qqmm` supports: (non-quantised ``x``, non-quantised ``w``) or   (non-quantised ``x``, quantised ``w``,  ``scales_w``)
## Details

1. `scales` are repacked on every call for both weights and activations. In the future, we probably want to:
   - Avoid repacking weight scales for inference.
   - Fuse quantization and repacking, and directly pack into swizzled layout in `fp_quantize`.

2. Batched `qqmm` is currently not supported; inputs must be 2D. For now it is implemented this way because:
   - `CUBLASLT_BATCH_MODE_STRIDED` is not supported for scales.
   - `CUBLASLT_BATCH_MODE_POINTER_ARRAY` is not supported for arrays with block scaling.

3. `qqmm` is always executed in TN layout (`transpose = True`).
   There are several reasons for this, but mainly we always quantize along the reduction dimension, which currently ends up being the last dimension.. I am happy to change this if you think that it is useful to support all layouts for ``mxfp8`` for example. Also, only on B200 only TN layout is supported for `nvfp4` and `mxfp4`.

4. There are some changes to `cublas_gemm.cpp`: I grouped all common cuBLAS-related functions into a separate helper class in `cublas_utils.cpp`.
We almost certainly want to add batching in the future, but for simplicity `batch_count = 1` for now.

- For ``nvfp4`` the output matches exactly for every tested shape.
- The difference is not structured: there is no clear pattern, and the indices of the affected elements change with the seed.
- The mismatch is always exactly 1 ULP.

Therefore, I attribute this to differences in accumulation on tensor cores or other numerical details we do not control. 

## What this PR lacks [these] because I first want to make sure the rest of the API looks reasonable

1. ``addmm`` -- basically `c` is always `nullptr`
2. `nn.QQLinear` 
3. `nn.Linear.to_qqlinear` - or similar method to cast to ``nn.QQLinear`` (naming is questionable)

Examples are in ``python/examples/qqmm.py``.
Happy to iterate and change anything here!

---
To give rough estimation on the quantization + scales packing overhead:
For `(M=4*4096, N=9728, K=2560)` ~72% is matmul. Scales repacking takes ~half of the total quantization time..

Screenshot 2025-12-15 at 18 06 27

### DIFF
--- a/mlx/ops.cpp
+++ b/mlx/ops.cpp
@@ -4305,31 +4305,28 @@ std::pair<int, int> extract_qqmm_dims(
     std::optional<array> scales_w,
     int group_size,
     int bits) {
-  auto [w_inner_dims, w_outer_dims] = [&]() {
-    if (w.dtype() != uint32) {
-      // if w is not quantized, check that last dims match
-      if (x.shape(-1) != w.shape(-1)) {
-        std::ostringstream msg;
-        msg << "[qqmm] Last dimension of first input with shape " << x.shape()
-            << " must match last dimension of"
-            << " second input with shape " << w.shape() << ".";
-        throw std::invalid_argument(msg.str());
-      }
-      return std::make_pair(w.shape(-1), w.shape(-2));
-    } else {
-      // if w is quantized, extract dims from quantized w
-      return extract_quantized_matmul_dims(
-          "qqmm",
-          x,
-          w,
-          *scales_w,
-          std::nullopt,
-          /* transpose = */ true,
-          group_size,
-          bits);
+  if (w.dtype() != uint32) {
+    // if w is not quantized, check that last dims match
+    if (x.shape(-1) != w.shape(-1)) {
+      std::ostringstream msg;
+      msg << "[qqmm] Last dimension of first input with shape " << x.shape()
+          << " must match last dimension of"
+          << " second input with shape " << w.shape() << ".";
+      throw std::invalid_argument(msg.str());
     }
-  }();
-  return {w_inner_dims, w_outer_dims};
+    return std::make_pair(w.shape(-1), w.shape(-2));
+  } else {
+    // if w is quantized, extract dims from quantized w
+    return extract_quantized_matmul_dims(
+        "qqmm",
+        x,
+        w,
+        *scales_w,
+        std::nullopt,
+        /* transpose = */ true,
+        group_size,
+        bits);
+  }
 }
 
 array qqmm(
--- a/mlx/backend/cuda/primitives.cpp
+++ b/mlx/backend/cuda/primitives.cpp
@@ -1,6 +1,7 @@
 // Copyright © 2025 Apple Inc.
 
 #include "mlx/distributed/primitives.h"
+#include <cuda_runtime.h>
 #include "mlx/fast_primitives.h"
 #include "mlx/primitives.h"
 
@@ -22,6 +23,13 @@ namespace mlx::core {
   void func::eval_gpu(const std::vector<array>& inputs, array& out) { \
     throw std::runtime_error(#func " has no CUDA implementation.");   \
   }
+
+#if CUDART_VERSION < 12080
+void QQMatmul::eval_gpu(const std::vector<array>& inputs, array& out) {
+  throw std::runtime_error(
+      "[QQMatmul::eval_gpu] QQMM is only supported with CUDA 12.8 or higher.");
+}
+#endif
 
 NO_GPU(BlockMaskedMM)
 NO_GPU(FFT)


mlx/primitives.h
```diff
@@ -1642,6 +1642,37 @@ class QuantizedMatmul : public UnaryPrimitive {
   bool transpose_;
 };
 
+class DualQuantizedMatmul : public UnaryPrimitive {
```
awni: A bit of a nit but I think it makes sense to rename this to `QuantizedQuantizedMatmul` or `QQMatmul` to better match the name of the op. `Dual` is also kind of an overloaded term.

mlx/primitives.h
```diff
@@ -1642,6 +1642,37 @@ class QuantizedMatmul : public UnaryPrimitive {
...
+  void eval_cpu(const std::vector<array>& inputs, array& out) override;
+  void eval_gpu(const std::vector<array>& inputs, array& out) override;
+
+  // DEFINE_VMAP()
+  DEFINE_GRADS()
+  DEFINE_NAME(DualQuantizedMatmul)
+  bool is_equivalent(const Primitive& other) const override;
+  std::vector<Shape> output_shapes(const std::vector<array>& inputs) override;
+  auto state() const {
+    return std::make_tuple(group_size_, bits_, mode_);
```
awni, line 1666: `transpose_` should be part of the state here.

python/tests/test_quantized.py
```diff
@@ -1015,6 +1023,99 @@ def gmm(s, x, wq):
 
             ds = mx.grad(gmm)(s, x, wq)
 
+    def test_qqmm(self):
```
awni: These tests will should only be run for now if `mx.cuda.is_available()`.

And in fact I'm not sure what the behavior is on older hardware and CUDA toolkits. Do you know what the minimum requirements there are?

mlx/primitives.h
nastya236: Yeah, this is a bit unclear and probably should be changed.. `transpose` is not a member variable, `qqmm` is always executed in TN layout (transpose = True). I did it this way because, at the moment, quantization always produces a row-major tensor with the last dimension packed, and TN is the only layout supported for mxfp4 and nvfp4 on B200.

mlx/primitives.h
nastya236: Yes, I agree. I think `QQMatmul` is better, because then the primitive name and the op name are aligned.

mlx/primitives.h
awni: I see it below in the list under `private:`. Maybe it should be deleted?

mlx/ops.cpp
```diff
@@ -4246,6 +4307,73 @@ array quantized_matmul(
...
+array qqmm(
+    array x,
+    array w_q,
+    array scales_w,
+    std::optional<array> w /* = std::nullopt */,
+    std::optional<int> group_size_ /* = std::nullopt */,
+    std::optional<int> bits_ /* = std::nullopt */,
+    const std::string& mode /* = "nvfp4" */,
+    StreamOrDevice s /* = {} */) {
+  // currently only simetric quantization is supported for qqmm
```
awni:
```suggestion
  // currently only symmetric quantization is supported for qqmm
```

mlx/ops.cpp
```diff
@@ -4246,6 +4307,73 @@ array quantized_matmul(
...
+  // https://docs.nvidia.com/cutlass/4.2.1/media/docs/cpp/blackwell_functionality.html
+  // because w_q should always be quantized along the reduction dimension
+  // and we quantize so that the last dim is packed, we assume that the last dim
+  // always the reduction dim so the firat argument in cubals column major is
+  // (the second argument in qqmm) always transposed
+  if (qmode == QuantizationMode::Affine) {
+    std::ostringstream msg;
+    msg << "[qqmm] Affine quantization is not supported for qqmm.";
+    throw std::invalid_argument(msg.str());
+  }
```
awni: It looks like this was already checked above?

mlx/ops.cpp
```diff
@@ -4246,6 +4307,73 @@ array quantized_matmul(
...
+    std::ostringstream msg;
+    msg << "[qqmm] only 'nvfp4' and 'mxfp8' quantization modes are supported but '"
+        << mode << "' was provided.";
+    throw std::invalid_argument(msg.str());
+  }
+  // for fp4 block scaling the only supported layout is TN
+  // https://docs.nvidia.com/cutlass/4.2.1/media/docs/cpp/blackwell_functionality.html
+  // because w_q should always be quantized along the reduction dimension
+  // and we quantize so that the last dim is packed, we assume that the last dim
+  // always the reduction dim so the firat argument in cubals column major is
```
awni:
```suggestion
  // is always the reduction dim so the first argument in cublas column major is
```

awni: This comment feels like it belongs in the cublas implementation rather than here.

mlx/ops.cpp
```diff
@@ -4246,6 +4307,73 @@ array quantized_matmul(
...
+        << "x.ndim() == " << x.ndim() << " and "
+        << "w_q.ndim() == " << w_q.ndim() << ".";
+    throw std::invalid_argument(msg.str());
+  }
+  auto [group_size, bits] =
+      quantization_params_from_mode(qmode, group_size_, bits_);
+  auto [w_inner_dims, w_outer_dims] =
+      extract_qqmm_dims("qqmm", x, w_q, scales_w, w, group_size, bits);
+
+  // we don't backprope through qunatized w and scales
```
awni:
```suggestion
  // we don't backprop through quantized w and scales
```

mlx/ops.cpp
```diff
@@ -4246,6 +4307,73 @@ array quantized_matmul(
...
+    inputs.push_back(*w);
+  }
+  if (x.ndim() > 2 && w_q.ndim() > 2) {
+    inputs = broadcast_arrays(inputs, {-2, -1}, s);
+  }
+
+  auto out_shape = inputs[0].shape();
+  out_shape.back() = w_outer_dims;
+  auto dtype = bfloat16;
+  // out dtype can be only bf16 for now
```
awni: Why this limitation? It looks like the op can output bf16, fp16 or fp32: https://docs.nvidia.com/cuda/cublas/#id103.

The API should infer the output type from `x`.

mlx/ops.cpp
```diff
@@ -133,6 +154,46 @@ std::pair<int, int> extract_quantized_matmul_dims(
...
+    std::string_view tag,
+    const array& x,
+    const array& w_q,
+    const array& scales_w,
+    const std::optional<array>& w,
+    int group_size,
+    int bits) {
+  // Validate w_q and scales_w
+  validate_quantized_input(
+      tag, w_q, scales_w, "weight matrix", "scales_w", group_size, bits);
```
awni: I think you can remove the strings here since `x` is not quantized. The original error message prior to the diff here makes sense.

mlx/ops.h
```diff
@@ -1409,6 +1409,16 @@ std::vector<array> quantize(
     const std::string& mode = "affine",
     StreamOrDevice s = {});
 
+array qqmm(
+    array x, // input activations
+    array w_q, // quantized weights
+    array w_scales,
+    std::optional<array> w = std::nullopt, // optional bf16 weights for vjp
```
awni: I don't really love this API where sometimes it takes a `w` as input and sometimes not. I wonder if it makes sense to change it to something like:

```suggestion
    array x, // input activations
    array w, // possibly quantized weights
    std::optional<array> scales, // scales for w, if not provided `w` must be unquantized
```

So then it will quantize on the fly if `w` is not quantized and otherwise it will just use `w` as is.

awni: And in order to take a vjp, `w` has to be provided unquantized.

mlx/primitives.cpp
```diff
@@ -3468,6 +3468,84 @@ std::vector<Shape> QuantizedMatmul::output_shapes(
...
+    if (arg == 0) { // gradient wrt to x
+      // We transpose weights -> quantize along N -> qqmm (cotan quantized in
+      // eval_gpu)
+      auto wtq = quantize(
+          transpose(primals[3], {1, 0}, s), // we assume that weights are 2D
+          group_size_,
+          bits_,
+          qmode,
+          s); // (K, N_packed), scales
+      vjps.push_back(qqmm(
```
awni, line 3507: A minor problem here is that this function is only once differentiable. I think changing the API as suggested above migth fix that. You always quantize the inputs on the fly when you want gradients.

mlx/backend/cuda/CMakeLists.txt
```diff
@@ -64,6 +65,11 @@ add_subdirectory(${CMAKE_CURRENT_SOURCE_DIR}/unary)
 # fp4 is not available on < 12.8
 if(CMAKE_CUDA_COMPILER_VERSION VERSION_LESS 12.8.0)
   target_include_directories(mlx PRIVATE ${CMAKE_CURRENT_SOURCE_DIR}/quantized/)
+else()
+  target_sources(mlx
+                 PRIVATE ${CMAKE_CURRENT_SOURCE_DIR}/quantized/qqmm_utils.cu)
+  target_sources(mlx
+                 PRIVATE ${CMAKE_CURRENT_SOURCE_DIR}/quantized/cublas_qqmm.cpp)
```
zcbenz: They can be merged:

```
  target_sources(mlx PRIVATE
                 ${CMAKE_CURRENT_SOURCE_DIR}/quantized/qqmm_utils.cu
                 ${CMAKE_CURRENT_SOURCE_DIR}/quantized/cublas_qqmm.cpp)
```

nastya236: Thank you! Missed this

python/tests/test_quantized.py
nastya236: I put it into examples for now.. But can drop it fully.

mlx/backend/cuda/CMakeLists.txt
```diff
@@ -143,6 +144,27 @@ message(STATUS "CUDA architectures: ${MLX_CUDA_ARCHITECTURES}")
...
+
+if(MLX_CUDA_HAS_SM100 AND CMAKE_CUDA_COMPILER_VERSION VERSION_GREATER_EQUAL
+                          12.8.0)
+  target_sources(
+    mlx
+    PRIVATE ${CMAKE_CURRENT_SOURCE_DIR}/quantized/qqmm.cpp
+            ${CMAKE_CURRENT_SOURCE_DIR}/quantized/cublas_qqmm.cpp
+            ${CMAKE_CURRENT_SOURCE_DIR}/quantized/qqmm_utils.cu)
+else()
+  target_sources(mlx
```
awni: Let's change this a little bit. 
1. Instead of checking SM100 here always build this if the toolkit version can build it
2. Check at runtime if the arch can run QQMM and otherwise raise an error

The rational is that we will build a single binary for CUDA which runs on multiple architectures. So we want to always build this when possible but check at run-time if it is ok to run it.

python/src/ops.cpp
```diff
@@ -5433,4 +5433,54 @@ void init_ops(nb::module_& m) {
...
+      If ``w`` is quantized, ``scales`` must be provided, and ``group_size``,
+      ``bits``, and ``mode`` must match the parameters that were used to quantize
+      ``w``.
+
+      Notes:
+        If ``w`` is expected to receive gradients, it must be provided in
+        non-quantized form.
+
+        ``x`` and `w`` if not-quantized must have float32, float16, or bfloat16
+        data types.
```
awni:
```suggestion
        If ``x`` and `w`` are not quantized, their data types must be ``float32``, ``float16``, or ``bfloat16``.
```

python/src/ops.cpp
```diff
@@ -5433,4 +5433,54 @@ void init_ops(nb::module_& m) {
...
+        If quantized, ``w`` must be packed in uint32 data type with.
```
awni: Something missing at the end here?

python/src/ops.cpp
```diff
@@ -5433,4 +5433,54 @@ void init_ops(nb::module_& m) {
...
+          ``w`` if ``w`` is quantized. Default: ``None``.
+        group_size (int, optional): Number of elements in ``x`` and ``w`` that
+          share a scale. See supported values and defaults in the
+          :ref:`table of quantization modes <quantize-modes>`. Default: ``None``.
+        bits (int, optional): Number of bits used to represent each element of
+          ``x`` and ``w``. See supported values and defaults in the
+          :ref:`table of quantization modes <quantize-modes>`. Default: ``None``.
+        mode (str, optional): The quantization mode. Default: ``"nvfp4"``.
+          See supported values and defaults in the :ref:`table of quantization 
+          modes <quantize-modes>`.
```
awni: I would add what quantization modes are supported for this op somewhere, maybe here. Afaik its nvfp4 and mxfp8 right?

python/src/ops.cpp
```diff
@@ -5433,4 +5433,54 @@ void init_ops(nb::module_& m) {
...
+      nb::arg(), // x
+      nb::arg(), // w_q
+      "scales"_a = nb::none(), // scales w
+      "group_size"_a = nb::none(),
+      "bits"_a = nb::none(),
+      "mode"_a = "nvfp4",
+      nb::kw_only(),
+      "stream"_a = nb::none(),
+      nb::sig(
+          "def qqmm(x: array, w: array, scales: Optional[array] = None, group_size: Optional[int] = None, bits: Optional[int] = None, mode: str = 'nvfp4', *, stream: Union[None, Stream, Device] = None) -> array"),
```
awni, line 5448: I think this is actually a pretty nice API.

mlx/ops.h
```diff
@@ -1420,6 +1420,16 @@ array dequantize(
     std::optional<Dtype> dtype = std::nullopt,
     StreamOrDevice s = {});
 
+array qqmm(
+    array x, // input activations
+    array w, // may be quantized weights
```
awni:
```suggestion
    array w, // maybe quantized weights
```

mlx/ops.cpp
```diff
@@ -4244,6 +4254,128 @@ array quantized_matmul(
...
+    std::optional<int> group_size_ /* = std::nullopt */,
+    std::optional<int> bits_ /* = std::nullopt */,
+    const std::string& mode /* = "nvfp4" */,
+    StreamOrDevice s /* = {} */) {
+// we need t ocheck 2 cases:
+// 1. w is quantized, scales is provided
+// 2. w is not quantized, scales is not provided
+#ifndef MLX_USE_CUDA
+  throw std::invalid_argument("[qqmm] QQMM is only supported on CUDA backend.");
+#endif
```
awni: I would change this to be a runtime check. Something like:

```
if (s.device != Device::gpu || !cuda::is_available()) {
 throw 
}
```

mlx/ops.cpp
```diff
@@ -4244,6 +4254,128 @@ array quantized_matmul(
...
+        << "x.ndim() == " << x.ndim() << " and "
+        << "w.ndim() == " << w.ndim() << ".";
+    throw std::invalid_argument(msg.str());
+  }
+  if (w.dtype() == uint32) {
+    // if w is quantized, scales are provided
+    if (!scales_w.has_value()) {
+      std::ostringstream msg;
+      msg << "[qqmm] Scales must be provided if second argument is quantized.";
+      throw std::invalid_argument(msg.str());
```
awni:
```suggestion
      throw std::invalid_argument("[qqmm] Scales must be provided if second argument is quantized.");
```

mlx/ops.cpp
```diff
@@ -4244,6 +4254,128 @@ array quantized_matmul(
...
+  }
+}
+
+std::pair<int, int> extract_qqmm_dims(
+    array x,
+    array w,
+    std::optional<array> scales_w,
+    int group_size,
+    int bits) {
+  auto [w_inner_dims, w_outer_dims] = [&]() {
```
awni: I am not sure if the lambda has a purpose here. But it looks like you could get rid of it and just return the pairs directly.

mlx/primitives.cpp
```diff
@@ -3468,6 +3468,67 @@ std::vector<Shape> QuantizedMatmul::output_shapes(
...
+          transpose(primals[1], {1, 0}, s), // assuming that w is 2D
+          {},
+          group_size_,
+          bits_,
+          qmode,
+          s));
+    } else if (arg == 1) { // gradient wrt to weights
+      vjps.push_back(qqmm(
+          transpose(cotan, reorder, s), // (N, M)
+          transpose(primals[0], reorder, s), // (K, M)
```
awni: These could probably be simplified with `swapaxes(cotan, -1, -2, s)`

mlx/primitives.cpp
```diff
@@ -3468,6 +3468,67 @@ std::vector<Shape> QuantizedMatmul::output_shapes(
...
+  // primal[0] -- non quantized activations (M, K)
+  // cotan -- non quantized grads (M, N)
+  auto qmode = quantization_mode_to_string(mode_);
+  for (auto arg : argnums) {
+    // TODO: we need a kernel that will quantize columnwise + transpose
+    if (arg == 0) { // gradient wrt to x
+      // We transpose weights -> quantize along N
+      vjps.push_back(qqmm(
+          cotan, //  M X N
+          transpose(primals[1], {1, 0}, s), // assuming that w is 2D
```
awni: For consistency could use the swap axes here as well (one less thing to change if we add support for ND.

mlx/primitives.cpp
```diff
@@ -3468,6 +3468,67 @@ std::vector<Shape> QuantizedMatmul::output_shapes(
...
+  out_shape.back() = w_outer_dims;
+  return {std::move(out_shape)};
+}
+
+std::vector<array> QQMatmul::vjp(
+    const std::vector<array>& primals, // non quantized x, non quantized w
+    const std::vector<array>& cotangents, // non quantized upstream grads
+    const std::vector<int>& argnums,
+    const std::vector<array>&) {
+  std::vector<array> vjps;
```
awni, line 3493: Maybe good to do a runtime check here that `primals.size() == 2` to make sure they aren't taking the VJP while sending qmm a quantized input (seems like a likely mistake).

Right now I think it will crash further upstream with a message about scales missing or an invalid dtype or something.

mlx/backend/cuda/quantized/qqmm.cpp
```diff
@@ -0,0 +1,200 @@
...
+    return {x_q, scales_x};
+  };
+  auto [x_q, scale_x_pre] = quantize(inputs[0], encoder, s);
+  auto [w_q, scale_w_pre] = [&]() {
+    if (inputs[1].dtype() != uint32) {
+      return quantize(inputs[1], encoder, s);
+    } else {
+      return std::make_pair(inputs[1], inputs[2]);
+    }
+  }();
```
awni: That's a nice way to avoid an extra copy of the arrays. You can do it a bit simpler in this case with a ternary:
```suggestion
  auto [w_q, scale_w_pre] = (inputs[1].dtype() != uint32) ?
      quantize(inputs[1], encoder, s) : std::make_pair(inputs[1], inputs[2]);
```

mlx/backend/cuda/quantized/qqmm.cpp
```diff
@@ -0,0 +1,200 @@
...
+    sshape[x.ndim() - 1] = pad_inner;
+    sshape.back() = scales_inner;
+
+    // Allocate outputs
+    const int64_t xq_bytes = x.size() * bits_ / 8;
+    const int64_t batch = x.size() / (x.shape(-2) * x.shape(-1));
+    const int64_t scales_bytes = batch * (pad_outer * pad_inner);
+
+    array x_q(cu::malloc_async(xq_bytes, encoder), xq_shape, uint32);
+    array scales_x(cu::malloc_async(scales_bytes, encoder), sshape, uint8);
```
awni:
```suggestion
    array x_q(cu::malloc_async(xq_bytes, encoder), std::move(xq_shape), uint32);
    array scales_x(cu::malloc_async(scales_bytes, encoder), std::move(sshape), uint8);
```

mlx/backend/cuda/quantized/qqmm.cpp
```diff
@@ -0,0 +1,200 @@
...
+  out.set_data(cu::malloc_async(out.nbytes(), encoder));
+
+  auto out_dtype = out.dtype();
+
+  int M = x_q.shape(-2);
+  int N = w_q.shape(-2); // always transposed
+  int K_packed = x_q.shape(-1);
+  int K = K_packed * (32 / bits_);
+
+  // // Repack scales from linear to tiled layout for tensor cores
```
awni:
```suggestion
  // Repack scales from linear to tiled layout for tensor cores
```

awni:
This is looking awesome! I left a few more comments. Could you address, rebase and then we can run the tests and try and get it merged?

mlx/primitives.cpp
nastya236: Good point! Yes make sense, you are right it will crash with misleading error.

python/src/ops.cpp
nastya236: Yes, you are right! only mxfp8 and nvfp4 are supported.

mlx/backend/cuda/CMakeLists.txt
nastya236: Thank you! It makes much more sense.. I did it like this during compilation because macros `__CUDA_ARCH__` are not available on the host and I missed that it could simply done with:
`auto cc = d.compute_capability_major() * 100 + d.compute_capability_minor() * 10;`

Issue #2724, opened 2025-11-01
[BUG] CUDA random crashes on large sizes


awni:
```python
import mlx.core as mx

a = mx.random.uniform(shape=(1024, 1024, 1024, 3))
mx.eval(a)
```

Fails with:
```
RuntimeError: cudaGraphAddKernelNode(&node, graph_, NULL, 0, &params) failed: invalid argument
```

spichardo:
I think this is related to #2681

sofinvalery:
This bug is gone on 0.31.1, but I found a related one where all_reduce over >INT_MAX elements hits
`RuntimeError: cudaGraphAddKernelNode(&node, graph_, NULL, 0, &params) failed: invalid argument`
Opened a new issue #3602 with PR #3603.

zcbenz:
Thanks for testing this issue!

PR #2798, opened 2025-11-19
[CUDA] Partly fix random for large sizes

awni:
Fix for random generation when the size is < `UINT32_MAX`.


awni:
Closes #2724

angeloskath (approved):
👍

PR #2800, opened 2025-11-20
Fix cuda allocator copy condition

awni:
Also decrease the cache threshold slightly to avoid OOMs when close to the limit.


mlx/backend/cuda/allocator.cpp
```diff
@@ -176,7 +176,7 @@ CudaAllocator::malloc_async(size_t size, int device, cudaStream_t stream) {
     buffer_cache_.release_cached_buffers(get_cache_memory() - max_pool_size_);
   }
   // Copy to managed here if the buffer is not on the right device
-  if (buf->device != device) {
+  if (buf->device >= 0 && buf->device != device) {
     copy_to_managed(*buf);
   }
```
awni, line 181: If the buffer is already managed (device = -1) then don't need a copy here

angeloskath (approved):
👍

PR #2805, opened 2025-11-20
Fix Stream Creation Races (Scheduler + Metal)

Jckwind:
Proposed changes
- Guard scheduler stream/thread vectors with a creation mutex to keep stream IDs unique under concurrency.
- Protect Metal stream_map_ with shared/unique locking and stable pointers to prevent races/rehash invalidation; centralize command buffer creation.
- Add a regression test for concurrent CPU stream creation.

Checklist
- [x]  I have read the CONTRIBUTING document
- [x]  I have updated the necessary documentation (if needed)


Jckwind:
Superseded by #2809 (threadsafe-stream-creation squashed).

PR #2808, opened 2025-11-20
Thunderbolt RDMA communications backend

angeloskath:
This PR adds the JACCL backend that enables low-latency distributed communication via thunderbolt RDMA. The backend is latency optimized for now and only supports full mesh but ring and bandwidth optimization should come later.

The following is a plot that compares the latency of 4-way all reduce on an M3 Ultra using RDMA vs the ring backend that uses TCP over thunderbolt.

m3-ultra-all-reduce

and a closer look on just the RDMA

m3-ultra-rdma-all-reduce

<details>
<summary>How to compute the above plots</summary>

The plots are fully computed as follows:

First we configure the backends and create the hostfiles
```
mlx.distributed_config --verbose --backend jaccl --hosts m3-ultra-1,m3-ultra-2,m3-ultra-3,m3-ultra-4 --over thunderbolt --auto-setup --output m3-ultra-jaccl.json
mlx.distributed_config --verbose --backend ring --hosts m3-ultra-1,m3-ultra-2,m3-ultra-3,m3-ultra-4 --over thunderbolt --auto-setup --output m3-ultra-ring.json
```

then we get the numbers and average per size
```
for bk in ring jaccl; do
  for kb in 1 2 4 8 16 32; do
    for j in {1..4}; do
      mlx.launch --backend $bk --cwd /Users/angelos --hostfile m3-ultra-ring.json --no-verify -- /path/to/python all_sum_bench.py $kb
    done
  done
done
```

and this is the full `all_sum_bench.py`

```python
import time
import mlx.core as mx
import sys

N = int(sys.argv[1])
assert 1024 >= N >= 1, "N should be in [1, 1024]"

group = mx.distributed.init()

x = mx.random.normal((N, 256,))
mx.eval(x)

y = x
for i in range(10):
    y = mx.distributed.all_sum(y)
mx.eval(y)

start = time.time()
for i in range(100):
    y = x
    for i in range(24 * 2):
        y = mx.distributed.all_sum(y)
    if N > 128:
        mx.eval(y)
    else:
        mx.async_eval(y)
stop = time.time()

duration = (stop - start) / 4800
print(duration, x.nbytes / duration / 1_000_000)
```
</details>

The only thing left to do is update a bit the docs on `mlx.launch` and `mlx.distirbuted_config` but otherwise this is ready for review.

### DIFF
--- a/docs/src/usage/distributed.rst
+++ b/docs/src/usage/distributed.rst
@@ -415,7 +415,7 @@ of a gigantic model using MLX LM.
 
    mlx.launch --verbose --backend jaccl --hostfile m3-ultra-jaccl.json \
         --env MLX_METAL_FAST_SYNCH=1 -- \  # <--- important
-        /path/to/remote/python -m mlx_lm chat --model mlx-community/DeepSeek-V3.2-8bit --shard
+        /path/to/remote/python -m mlx_lm chat --model mlx-community/DeepSeek-R1-0528-4bit
 
 .. note::
 
@@ -442,7 +442,7 @@ it takes to run a distributed job is
    mlx.launch -n 8 test.py
 
    # perfect for interactive scripts
-   mlx.launch -n 8 python -m mlx_lm chat --model my-model --shard
+   mlx.launch -n 8 python -m mlx_lm chat --model my-model
 
 You can also use ``mlx.launch`` to ssh to a remote node and launch a script
 with the same ease


Anemll:
Xcode 26.2.2 beta is required with new infiniband/verbs.h for "RDMA"

Anemll:
Thunderbolt RDMA setup ( not-official, TB5 devices only )

In recovery OS/terminal, enable RDMA and reboot
rdma_ctl enable

use 
ibv_devices 
ibv_devinfo

CMakeLists.txt
```diff
@@ -119,6 +119,10 @@ if(MLX_BUILD_METAL)
     COMMAND zsh "-c" "/usr/bin/xcrun -sdk macosx --show-sdk-version"
     OUTPUT_VARIABLE MACOS_SDK_VERSION
     OUTPUT_STRIP_TRAILING_WHITESPACE COMMAND_ERROR_IS_FATAL ANY)
+  execute_process(
+    COMMAND zsh "-c" "/usr/bin/xcrun -sdk macosx --show-sdk-path"
+    OUTPUT_VARIABLE CMAKE_OSX_SYSROOT
+    OUTPUT_STRIP_TRAILING_WHITESPACE COMMAND_ERROR_IS_FATAL ANY)
```
awni: Looks unused?

python/src/distributed.cpp
```diff
@@ -52,9 +52,25 @@ void init_distributed(nb::module_& parent_module) {
...
       Check if a communication backend is available.
+
+      Note, this function returns whether MLX has the capability of
+      instantiating that distributed backend not whether it is possible to
+      create a communication group. For that purpose one should use
+      ``init(strict=True)``.
+
+      Args:
+        backend (str, optional): The name of the backend to check for availability.
+          It takes the same values as ``init()``. Default: ``any``.
```
awni:
```suggestion
          It takes the same values as :func:`init()`. Default: ``"any"``.
```

CMakeLists.txt
andresy: Also, it is supposed to be the way one specifies to CMake which SDK to use?
https://cmake.org/cmake/help/latest/variable/CMAKE_OSX_SYSROOT.html

angeloskath: 🙏 Forgot to remove it.

docs/src/usage/distributed.rst
```diff
@@ -192,16 +201,273 @@ almost identical to the example above:
...
+
+Getting Started with Ring
+-------------------------
+
+The ring backend does not depend on any third party library so it is always
+available. It uses TCP sockets so the nodes need to be reachable via a network.
+As the name suggests the nodes are connected in a ring which means that rank 1
+can only communicate with rank 0 and rank 2, rank 2 only with rank 1 and rank 3
+and so on and so forth. As a result :func:`send` and :func:`recv` with
+arbitrary sender and receiver is not supported in the ring backend.
```
awni:
```suggestion
arbitrary sender and receiver are not supported in the ring backend.
```

docs/src/usage/distributed.rst
```diff
@@ -192,16 +201,273 @@ almost identical to the example above:
...
+  ``192.168.0.2`` respectively to the two nodes. For more details you can see
+  the commands prepared by the utility script.
+
+.. _jaccl_section:
+
+Getting Started with RDMA over Thunderbolt
+------------------------------------------
+
+Starting from version 26.2 RDMA over thunderbolt is available in MacOS and
+enables low-latency communication between Macs with thunderbolt 5. MLX provides
```
awni:
```suggestion
Starting from macOS 26.2, RDMA over thunderbolt is available and
enables low-latency communication between Macs with thunderbolt 5. MLX provides
```

(Make it more clear the version is for macOS, also lowercase `m`)

docs/src/usage/distributed.rst
```diff
@@ -192,16 +201,273 @@ almost identical to the example above:
...
+the JACCL backend that uses this functionality to achieve communication latency
+an order of magnitude lower than the ring backend.
+
+.. note::
+
+   The name JACCL (pronounced Jackal) stands for *Jack and Angelos' Collective
+   Communication Library* and it is an obvious pun to Nvidia's NCCL but also
+   tribute to *Jack Beasley* who led the development of RDMA over Thunderbolt
+   at Apple.
```
awni, line 288: ❤️

docs/src/usage/distributed.rst
```diff
@@ -192,16 +201,273 @@ almost identical to the example above:
...
+All of the above can be done instead via ``mlx.distributed_config``. This helper
+script will
+
+- ssh into each node
+- extract the thunderbolt connectivity
+- check for a valid mesh
+- provide the commands to configure each node (or run them if sudo is available)
+- generate the hostfile to be used with ``mlx.launch``
+
+Putting it All Together
```
awni: Nit, the pronoun is usually capitalized

```suggestion
Putting It All Together
```

docs/src/usage/distributed.rst
```diff
@@ -192,16 +201,273 @@ almost identical to the example above:
...
+        --over thunderbolt --backend jaccl \
+        --auto-setup --output m3-ultra-jaccl.json
+
+And now we are ready to run a distributed MLX script such as distributed inference
+of a gigantic model using MLX-LM.
+
+.. code-block::
+
+   mlx.launch --verbose --backend jaccl --hostfile m3-ultra-jaccl.json \
+        --env MLX_METAL_FAST_SYNCH=1 -- \  # <--- important
```
awni, line 417: Wdyt about setting that by default in the launcher? Seems like 99 - 100% of use cases will set it so it should just be the default for this case.

awni: (I mean at some point we should make it the default for MLX, maybe after the kernel hang bug is fixed)

awni:
For some reason building locally on my M1 (macOS 15.6.1) is failing with:
```
        import mlx.core as mx
    ImportError: dlopen(/Users/awnihannun/repos/mlx/python/mlx/core.cpython-312-darwin.so, 0x0002): Library not loaded: /usr/lib/librdma.dylib
      Referenced from: <871D39B2-AC62-3FB7-B432-9301BC01790C> /Users/awnihannun/repos/mlx/python/mlx/lib/libmlx.dylib
      Reason: tried: '/usr/lib/librdma.dylib' (no such file), '/System/Volumes/Preboot/Cryptexes/OS/usr/lib/librdma.dylib' (no such file), '/usr/lib/librdma.dylib' (no such file, not in dyld cache)
```

mlx/distributed/jaccl/CMakeLists.txt
```diff
@@ -0,0 +1,8 @@
+if(MLX_BUILD_CPU
+   AND ${CMAKE_SYSTEM_NAME} MATCHES "Darwin"
+   AND MACOS_SDK_VERSION GREATER_EQUAL 26.2)
+  target_sources(mlx PRIVATE ${CMAKE_CURRENT_SOURCE_DIR}/jaccl.cpp)
+  target_link_libraries(mlx PRIVATE rdma)
```
awni: I think checking the SDK is the wrong thing here. We probably to check the OS version?

Also I'm a bit uncertain about this setup. One can build the binary on a new OS but if you try to run it on an old OS it won't have the shared library installed.

It's especially tricky cause we do a release for macOS 26.0  and the shared library is presumably only on 26.2. So I don't see a way around making this a run-time check if we want the pypi release to support JACCL.

Wdyt?

docs/src/usage/distributed.rst
```diff
@@ -7,22 +7,29 @@ Distributed Communication
...
+   :header-rows: 1
+
+   * - Backend
+     - Description
+   * - :ref:`MPI <mpi_section>`
+     - A full featured and mature distributed communications library.
+   * - :ref:`RING <ring_section>`
+     - Ring all reduce and all gather over TCP sockets. Always available and
+       usually faster than MPI.
+   * - :ref:`JACCL <ring_section>`
```
awni:
```suggestion
   * - :ref:`JACCL <jaccl_section>`
```

docs/src/usage/distributed.rst
```diff
@@ -192,16 +201,273 @@ almost identical to the example above:
...
+  corresponding to that cable in nodes ``i`` and ``i + 1``.
+* Set up a unique subnetwork connecting the two nodes for the corresponding
+  interfaces. For instance if the cable corresponds to ``en2`` on node ``i``
+  and ``en2`` also on node ``i + 1`` then we may assign IPs ``192.168.0.1`` and
+  ``192.168.0.2`` respectively to the two nodes. For more details you can see
+  the commands prepared by the utility script.
+
+.. _jaccl_section:
+
+Getting Started with RDMA over Thunderbolt
```
awni:
```suggestion
Getting Started with JACCL
```

for consistency with the other headers

docs/src/usage/distributed.rst
```diff
@@ -192,16 +201,273 @@ almost identical to the example above:
...
+to be a thunderbolt cable connecting all pairs of Macs directly. For example, in
+the following topology visualizations, the left one is valid because there is a
+connection from any node to any other node, while for the one on the right M3
+Ultra 1 is not connected to M3 Ultra 2.
+
+.. raw:: html
+
+   <div style="display: flex; text-align: center; align-items: end; font-size: 80%;">
+     <div>
+       <img src="/_static/distributed/m3-ultra-mesh.png" alt="M3 Ultra thunderbolt mesh" style="width: 55%">
```
awni:
```suggestion
       <img src="../_static/distributed/m3-ultra-mesh.png" alt="M3 Ultra thunderbolt mesh" style="width: 55%">
```

docs/src/usage/distributed.rst
```diff
@@ -192,16 +201,273 @@ almost identical to the example above:
...
+       <p>Fully connected mesh of four M3 Ultra.</p>
+     </div>
+     <div>
+       <img src="/_static/distributed/m3-ultra-mesh-broken.png" alt="M3 Ultra broken thunderbolt mesh" style="width: 55%">
```
awni:
```suggestion
       <img src="../_static/distributed/m3-ultra-mesh-broken.png" alt="M3 Ultra broken thunderbolt mesh" style="width: 55%">
```

docs/src/usage/distributed.rst
```diff
@@ -192,16 +201,273 @@ almost identical to the example above:
...
+
+.. code-block::
+
+   mlx.distributed_config --verbose \
+        --hosts m3-ultra-1,m3-ultra-2,m3-ultra-3,m3-ultra-4 \
+        --over thunderbolt --backend jaccl \
+        --auto-setup --output m3-ultra-jaccl.json
+
+And now we are ready to run a distributed MLX script such as distributed inference
+of a gigantic model using MLX-LM.
```
awni: Nit we use no hyphen elsewhere when referring to the repo / project.

```suggestion
of a gigantic model using MLX LM.
```

docs/src/usage/distributed.rst
```diff
@@ -192,16 +201,273 @@ almost identical to the example above:
...
+
+.. code-block::
+
+   mlx.launch --verbose --backend jaccl --hostfile m3-ultra-jaccl.json \
+        --env MLX_METAL_FAST_SYNCH=1 -- \  # <--- important
+        /path/to/remote/python -m mlx_lm chat --model mlx-community/DeepSeek-V3.2-8bit --shard
```
awni: I don't think the example here works since it's v3.2 (not v3.1 or v3).

docs/src/usage/distributed.rst
```diff
@@ -269,78 +535,116 @@ Force MPI to use the most performant network interface by setting ``--mca
...
+**CUDA_VISIBLE_DEVICES** should contain the local index of the gpu that
+corresponds to this process.
+
+Of course any `other environment variable
+<https://docs.nvidia.com/deeplearning/nccl/user-guide/docs/env.html>`_ that is
+used by NCCL can be set.
+
+.. _tips_and_tricks:
+
+Tips and Tricks
```
awni, line 622: Very nice section!

docs/src/usage/launching_distributed.rst
```diff
@@ -7,13 +7,106 @@ Launching Distributed Programs
 
 .. currentmodule:: mlx.core.distributed
 
-Installing the MLX python package provides a helper script ``mlx.launch`` that
-can be used to run python scripts distributed on several nodes. It allows
-launching using either the MPI backend or the ring backend. See the
-:doc:`distributed docs <distributed>` for the different backends.
+Installing the MLX python package provides two utilities to help you configure
```
awni: Nit: 
```suggestion
The MLX python package provides two utilities to help you configure
```

docs/src/usage/launching_distributed.rst
```diff
@@ -7,13 +7,106 @@ Launching Distributed Programs
...
+
+.. code-block::
+
+   mlx.distributed_config --verbose --backend jaccl \
+        --hosts m3-ultra-1,m3-ultra-2,m3-ultra-3,m3-ultra-4 --over thunderbolt \
+        --auto-setup --output m3-ultra-jaccl.json
+
+Let's walk through the steps that the script takes to configure the nodes.
+
+1. Ssh to all nodes to verify that they are reachable
```
awni:
```suggestion
1. ssh to all nodes to verify that they are reachable
```

docs/src/usage/distributed.rst
angeloskath: Yeah probably good point. The reason I didn't make it a default is it might actually be a bad idea when developing because the code will probably have bugs which may result in hangs and require reboots.

Should I make it a flag so it's easier and then should it be a `--fast-synch` flag to enable it or `--no-fast-synch` to disable it?

awni: I don't think it's necessary especially since we plan to default it on once the GPU hang bug is fixed.

awni: We can just leave it as is for now and switch it when possible.

mlx/distributed/jaccl/CMakeLists.txt
angeloskath: For some reason I thought that `librdma` would be available but not the TB implementation. No idea why.

Switching to a completely dynamic dlopen wrapper. Will take some measurements to ensure perf is the same/

angeloskath: I think the SDK check can remain cause we only need the `ibverbs` headers to be there in order to build successfully. Then on any machine with librdma it should work.

docs/src/usage/distributed.rst
angeloskath: I added sharding for that but annoyingly the chat template is changed and apparently not supported by HF. So I switched it to R1.

awni: Yes, it is very annoying. It's on my todo list to try and support their custom chat application.

uebber:
@angeloskath That's really great stuff, thank you so much!

I have followed the guide to enable RDMA in recovery mode on M2 Ultra Mac Studios. Unfortunately I still don't get any rdma devices. Any privacy settings that maybe need to be enabled in addition? Latest 26.2. Or is it limited to TB5 only?

`user@studio1 mlx-lm % rdma_ctl status
enabled
user@studio1 mlx-lm % ibv_devices 
    device          	   node GUID
    ------          	----------------
`

awni:
@uebber it's Tb5 only so it won't work on M2 unfortunately.

yaronsumel:
@angeloskath @awni Amazing - thanks! Im trying this out now - is it worth adding a getting started with rdma on 26.2 ? seems like this feature is not well documented on apple dev docs. for example: 
$ rdma_ctl enable
rdma_ctl: This tool needs to be executed from Recovery OS.
and other things worth mentioning to set it up ?

awni:
@yaronsumel it should be covered in the docs https://ml-explore.github.io/mlx/build/html/usage/distributed.html#getting-started-with-jaccl

awni:
If anything is missing there please open an issue or send a PR and we'll get it fixed!

yaronsumel:
@awni thanks !

yaronsumel:
@awni any clue on how to get around this case 

% ibv_devices
    device node GUID
    ------ ----------------
    rdma_en2 00b08afa0xxxxx
    rdma_en3 01b08afa0xxxxx
    rdma_en4 02b08afa0xxxxx

% rdma_ctl status
enabled

% ifconfig | grep rdma
-- nothing -- 

both nodes m4pro connected with tb5 cable / no thunderbolt bridge

awni:
Would you mind opening an issue with more details of the steps you took and we can continue the conversation there?

IntuitIntelLLC:
> @uebber it's Tb5 only so it won't work on M2 unfortunately.

@awni What limits this from being enabled on TB4? 

Half speed of TB5, but still highly usable if it had RDMA.

PR #2810, opened 2025-11-21
[Metal] Implement eval_gpu for Load primitive

Jckwind:
Proposed changes
- Implement Load::eval_gpu for Metal: allocate via the Metal allocator, read tensor directly into the shared MTL buffer, and perform optional endianness swap.
- Run the file read on the I/O thread pool and block until completion so subsequent GPU kernels see valid data (no staging copy needed).

Checklist
- [x]  I have read the CONTRIBUTING document
- [x]  I have updated the necessary documentation (if needed)

### DIFF
--- a/mlx/backend/metal/primitives.cpp
+++ b/mlx/backend/metal/primitives.cpp
@@ -1,6 +1,7 @@
 // Copyright © 2023-2024 Apple Inc.
 #include <algorithm>
 #include <cassert>
+#include <future>
 #include <numeric>
 #include <sstream>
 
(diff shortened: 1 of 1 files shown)


angeloskath:
MLX is using unified memory so there is no copy being done and the load is done in the CPU anyway. The way it is currently written it is stopping the submission to the GPU and doing the read immediately.

Jckwind:
I am deliberately blocking the dispatch thread to enforce determinism in a concurrent multi-tenant environment where the global StreamContext variable causes race conditions that crash the Metal driver during concurrent tensor loading. Solving the unsafe concurrent load meant separating the loads to its own Stream - moving the load op to it's own stream (i.e eval_gpu) was required for this fix.

I'm fine keeping this closed. Worth noting that the current upstream implementation isn't thread-safe for parallel loading - which is known already - but my fix has proven effective in my personal fork.

awni:
> moving the load op to it's own stream

You can move it to it's own CPU stream. 

Also doing this with `eval_gpu` is not so straight-forward. The way you have it in this PR doesn't synchronize properly with the rest of the GPU stream so I would not do that. It's not in a command buffer and it is a race condition. You could do `load -> gpu_op` and there's a good chance the `gpu_op` will read uninitialized data.

> global StreamContext variable causes race conditions that crash the Metal driver during concurrent tensor loading

A test case which reproduces that behavior would be super useful so we can find a general solution for you.

PR #2811, opened 2025-11-21
Centralize NAX condition

awni:
Centralize the nax checking to make it easier to turn on and off.


awni:
It's not needed. (Empirically it works fine without it)

But also the Metal std should be inherited from the sdk version and we only ever build the nax kernels when the SDK is 26.2.

awni:
I was thinking about adding an environment flag to toggle nax at run-time. Not sure if it's worth it.. but would be easy to add the `is_nax_available`.

PR #2813, opened 2025-11-21
Force cudaGraphExec reinstantiation when clusters are used

andportnoy:
## Proposed changes

Thread block cluster dimensions are not correctly updated by cudaGraphExecUpdate. Therefore, when clusters are used, we reinstantiate a cudaGraphExec rather than updating it.

## Checklist

  The change fixes existing tests.

### DIFF
--- a/mlx/backend/cuda/cuda_utils.h
+++ b/mlx/backend/cuda/cuda_utils.h
@@ -67,7 +67,6 @@ class CudaGraph : public CudaHandle<cudaGraph_t, cudaGraphDestroy> {
   using CudaHandle::CudaHandle;
   explicit CudaGraph(cu::Device& device);
   void end_capture(cudaStream_t stream);
-  bool uses_clusters();
 };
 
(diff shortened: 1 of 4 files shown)


awni:
Some benchmarks on B200

Inference with `meta-llama/Meta-Llama-3.1-8B`

Pre: prompt_tps=59185.146
Post: prompt_tps=59359.110

Pre: generation_tps=282.392
Post: generation_tps=282.671


Pre: toks_per_sec: 97961.0518
Post: toks_per_sec: 97452.2300

awni:
@andportnoy I changed this a bit more than I expected. Basically three changes:

1. we don't bother cache the graph exec if the graph is not updatable since there is no point
2. The condition for allowing the update are less conservative (if there is a non-singleton cluster in the x dimension of a sub graph with a single kernel node we can still update the graph).
3. In order to achieve the above I encoded the cluster x dimension in the graph key which took a bit of rearranging to make it work nicely.

awni (approved):
This resolves outstanding test failures! Thanks for the fix!!

PR #2819, opened 2025-11-22
Fix `mx.core.load` type annotation

### LINKED ISSUE #2812: [BUG] mlx.core.load() type annotation mismatch
**Describe the bug**
`mlx.core.load()` is declared to return `Union[array, dict[str, array]]`, but it returns a `tuple` object, at least in some cases.

**To Reproduce**

```python
import mlx.core as mx

gguf_file = "/path/to/model.gguf"
weights_and_metadata = mx.load(gguf_file, return_metadata=True)

print(type(weights_and_metadata))
# <class 'tuple'>

print(type(weights_and_metadata[1]))
# <class 'dict'>

metadata: dict = weights_and_metadata[1] # Fails type checking (ty, pyright, …)
```

**Expected behavior**
Either `mlx.core.load()` returns an object conforming to the declared return type, or the type annotation should be fixed.

 - OS Version: macOS 26.1
 - Version 0.30.0

CC-Yeh:
## Proposed changes

Close #2812

## Checklist

### DIFF
--- a/python/src/ops.cpp
+++ b/python/src/ops.cpp
@@ -4021,7 +4021,7 @@ void init_ops(nb::module_& m) {
       nb::kw_only(),
       "stream"_a = nb::none(),
       nb::sig(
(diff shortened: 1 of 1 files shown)


awni (approved):
Thanks!

PR #2823, opened 2025-11-23
Improve grouped convolution tile selection for better performance

dzoba:
## Proposed changes

Improve grouped convolution performance by using larger M-dimension blocks (64x8) for workloads with many groups and large spatial dimensions.

Modified tile selection heuristic in `implicit_gemm_conv_2D_gpu` to use `bm=64` instead of `bm=32` when:
- `implicit_M >= 8192` (large spatial output)
- `groups >= 8` (many groups)

This improves GPU utilization for grouped convolutions by increasing parallelism along the M dimension.

### Benchmark Results

Tested on Apple Silicon with `conv_bench.py`:

| Configuration | Before | After | Improvement |
|--------------|--------|-------|-------------|
| (4, 64, 64, 256) groups=2 | -12.02% | -0.59% | **+11.4 pp** |
| (4, 64, 64, 256) groups=16 | -52.77% | -34.43% | **+18.3 pp** |
| (4, 64, 64, 256) groups=64 | +69.32% | +132.42% | **+63.1 pp** |

(Percentages are relative to PyTorch MPS performance; positive = MLX faster)

## Checklist

### DIFF
--- a/mlx/backend/metal/conv.cpp
+++ b/mlx/backend/metal/conv.cpp
@@ -201,9 +201,19 @@ void implicit_gemm_conv_2D_gpu(
   int bk = 16;
 
   if (implicit_N <= 16) {
-    bn = 8;
-    wm = 4;
-    wn = 1;
+    // For large M with many groups, use bigger blocks for better parallelism
+    if (implicit_M >= 8192 && groups >= 8) {
+      // Use 64x8 blocks to maximize M-dimension parallelism while keeping N
+      // small
+      bm = 64;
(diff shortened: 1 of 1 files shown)


awni:
I'm a bit confused by the benchmark. Why is `groups=2` faster? Shouldn't this only apply to groups >= 8?

awni:
I tried this and didn't see a consistent speedup. I think we would need more careful benchmarking here across shapes  and hardware to land it.

PR #2825, opened 2025-11-24
[CUDA] Add debug env to save cuda graphs to dot files

zcbenz:
By setting the `MLX_SAVE_CUDA_GRAPHS_DOT_FILE=graph` env the CUDA graphs will be saved to `graph_1.dot`, `graph_2.dot` ... files.

### DIFF
--- a/mlx/backend/cuda/device.cpp
+++ b/mlx/backend/cuda/device.cpp
@@ -24,10 +24,19 @@ void check_cudnn_error(const char* name, cudnnStatus_t err) {
 }
 
 bool use_cuda_graphs() {
-  static bool use_graphs = []() {
-    return env::get_var("MLX_USE_CUDA_GRAPHS", true);
-  }();
+  static bool use_graphs = env::get_var("MLX_USE_CUDA_GRAPHS", true);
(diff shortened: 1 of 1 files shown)


mlx/backend/cuda/device.cpp
```diff
@@ -421,6 +430,17 @@ void CommandEncoder::commit() {
...
     }
+
+    // Save cuda graph to dot file
+    if (const char* filename = save_cuda_graphs_dot_file(); filename) {
+      static std::unordered_set<std::string> saved_keys;
+      auto [_, new_graph] = saved_keys.insert(graph_key);
+      if (new_graph) {
+        auto path = fmt::format("{}_{}.dot", filename, saved_keys.size());
+        CHECK_CUDA_ERROR(cudaGraphDebugDotPrint(graph_, path.c_str(), 0));
+      }
```
awni: Maybe this should save all the graphs and not just the unique `graph_key`?

zcbenz: Yeah simply saving all graphs should be enough for most cases.

PR #2826, opened 2025-11-24
[CUDA] Output of SDPA should have same layout with inputs

zcbenz:
For following code:

```python
import mlx.core as mx

B, T, N, D = (1, 128, 32, 64) 
scale = D ** -0.5

q = mx.random.normal((B, T, N, D), dtype=mx.float16).transpose(0, 2, 1, 3)
k = mx.random.normal((B, T, N, D), dtype=mx.float16).transpose(0, 2, 1, 3)
v = mx.random.normal((B, T, N, D), dtype=mx.float16).transpose(0, 2, 1, 3)

o = mx.fast.scaled_dot_product_attention(q, k, v, scale=scale).transpose(0, 2, 1, 3)
check_row_contiguous = mx.flatten(o)
mx.eval(check_row_contiguous)
```

Before the change the graph is:

before

After the change the graph becomes:

after

i.e. the `o` becomes row contiguous.

### DIFF
--- a/mlx/backend/cuda/scaled_dot_product_attention.cpp
+++ b/mlx/backend/cuda/scaled_dot_product_attention.cpp
@@ -63,6 +63,38 @@ array prepare_sdpa_input(const array& x, Stream s) {
   return x;
 }
 
+void malloc_with_same_layout(
+    cu::CommandEncoder& encoder,
+    array& o,
+    const array& q) {
+  if (q.flags().row_contiguous) {
+    o.set_data(cu::malloc_async(o.nbytes(), encoder));
+    return;
+  }
+  // fill_order = argsort(q.strides())
+  Shape fill_order(q.ndim());
+  std::iota(fill_order.begin(), fill_order.end(), 0);
+  std::stable_sort(
+      fill_order.begin(), fill_order.end(), [&q](int idx1, int idx2) {
+        auto s1 = q.strides(idx1) > 0 ? q.strides(idx1) : 1;
+        auto s2 = q.strides(idx2) > 0 ? q.strides(idx2) : 1;
+        return s1 < s2;
+      });
+  // Generate o_strides with fill_order
+  Strides o_strides(q.ndim());
+  int64_t stride = 1;
+  for (int i : fill_order) {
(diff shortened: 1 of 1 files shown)


mlx/backend/cuda/scaled_dot_product_attention.cpp
```diff
@@ -63,6 +63,38 @@ array prepare_sdpa_input(const array& x, Stream s) {
...
+        auto s2 = q.strides(idx2) > 0 ? q.strides(idx2) : 1;
+        return s1 < s2;
+      });
+  // Generate o_strides with fill_order
+  Strides o_strides(q.ndim());
+  int64_t stride = 1;
+  for (int i : fill_order) {
+    o_strides[i] = stride;
+    stride *= o.shape(i);
+  }
```
awni, line 89: I'm slightly confused by this. Why not just set the output strides to be the same as the input strides?

awni: Is it because `q` can be broadcast (have stride=0 in some dimension)?

zcbenz: If Q and V have different embedding dimensions then `O` would have a different shape:
Screenshot 2025-11-25 at 8 00 47

And yes `Q` may also be broadcasted (https://github.com/pytorch/pytorch/pull/157145).

awni: Sorry, I have some questions about this still:
- Is it right to put the stride=0 dimensions at  the end?
- Do we need to do anything special with negative strides here?

zcbenz: stride<=0 dims are treated as stride=1 when sorting: `s1 = q.strides(idx1) > 0 ? q.strides(idx1) : 1`, and with `std::stable_sort` they will be put before the contiguous dim. This just copies what PyTorch did for broadcasted dims (https://github.com/pytorch/pytorch/pull/157145) and it feels right, but I don't really know whether it is right.

For negative strides I have no idea, it is just handled like the stride=0 dims.

PR #2830, opened 2025-11-24
[CUDA] Exit on crash and more helpful errors

### LINKED ISSUE #2821: [BUG] [CUDA] Hangs on OOM crash with unhelpful error message
On B200 the following crashes and then hangs.

```python
import mlx.core as mx
arrs = []
while True:
    a = mx.random.normal(shape=(4096, 4096))
    mx.eval(a)
    print(mx.get_peak_memory() / 1e9)
    arrs.append(a)
```

```
RuntimeError: cudaMallocManaged failed: an illegal memory access was encountered.
  what():  cudaFreeAsync(buf->data, free_streams_[buf->device]) failed: an illegal memory access was encountered
```

It shouldn't hang and it should give a more useful error message.

Also, I'm wondering also what the behavior we want is.. should it be a hard crash when it runs out of device memory or should it fall back to managed memory which could use CPU RAM as well?

awni:
- Check error state after `event.wait` to catch errors during kernel execution
- Fix OOM error messaging by properly returning nullptr when alloc fails
- Don't throw in scheduler destructor
- Don't destroy cuda resources when error state is set

Closes #2821

### DIFF
--- a/mlx/backend/cuda/allocator.cpp
+++ b/mlx/backend/cuda/allocator.cpp
@@ -155,7 +155,7 @@ CudaAllocator::malloc_async(size_t size, int device, cudaStream_t stream) {
     lock.unlock();
     if (!buf) {
       cudaError_t err;
-      void* data;
+      void* data = nullptr;
       if (device == -1) {
         err = cudaMallocManaged(&data, size);
       } else {
--- a/mlx/backend/cuda/cuda_utils.h
+++ b/mlx/backend/cuda/cuda_utils.h
@@ -29,6 +29,7 @@ class CudaHandle {
   }
 
   ~CudaHandle() {
+    // Skip if there was an error to avoid throwing in the destructors
     if (cudaPeekAtLastError() != cudaSuccess) {
       return;
     }


mlx/backend/cuda/allocator.cpp
```diff
@@ -154,17 +154,21 @@ CudaAllocator::malloc_async(size_t size, int device, cudaStream_t stream) {
     }
     lock.unlock();
     if (!buf) {
-      buf = new CudaBuffer{nullptr, size, device};
       cudaError_t err;
+      void* data;
```
zcbenz:
```suggestion
      void* data = nullptr;
```

mlx/backend/cuda/cuda_utils.h
```diff
@@ -29,6 +29,9 @@ class CudaHandle {
   }
 
   ~CudaHandle() {
+    if (cudaPeekAtLastError() != cudaSuccess) {
```
zcbenz, line 33: Can you add a comment on the reason of the check?

zcbenz (approved):
Nice!

PR #2833, opened 2025-11-25
[WIP] Fix sorted gather_mm M=1 Metal accumulation drift 

Jckwind:
## Proposed changes

  - Added a Metal GPU regression covering gather_mm with sorted_indices=True and M=1; it currently fails because the sorted path uses BlockMMA accumulation and diverges from the unsorted GEMV order.
  - Repro is deterministic and checks both sorted vs unsorted and against a direct matmul reference.
  - Follow-up: add the M==1 scalar accumulation path so the new test passes without affecting other shapes.
  
  TO-DO: Implement fix, determine best path forward.

## Checklist

### DIFF
--- a/tests/gpu_tests.cpp
+++ b/tests/gpu_tests.cpp
@@ -444,6 +444,34 @@ TEST_CASE("test gpu matmul") {
   }
 }
 
+TEST_CASE("test gpu gather_mm sorted matches unsorted for m=1") {
+  // Deterministic inputs that route M==1 sorted gather through the Metal GEMM
+  // path.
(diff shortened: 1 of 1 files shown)


awni:
@Jckwind it's expected that you get numerical differences here. I'm not seeing anything larger than what's reasonable for bf16. A sequence of gemv will give you numerically different results than a gemm (this isn't specific to the gather_mm). The difference will be larger for bf16 than fp32. For bf16 relative differences on the order of ~1e-2 are expected.

If you think there is something else going on here feel free to comment or open an issue with more details on what you are seeing.

PR #2836, opened 2025-11-26
Fix causal mask alignment in scaled_dot_product_attention (Fixes #2835)

### LINKED ISSUE #2835: [BUG] Result of scaled_dot_product_attention does not match PyTorch when S_Q != S_KV, mask="causal"
For below script, the first test passes but the second test fails. Same result for both Metal and CUDA backends.

```python
import math
import torch
import mlx.core as mx

torch.manual_seed(42)

B = 4 
S_Q = 32
S_KV = 1024
H = 12
D = 64
dtype = torch.half

attn_scale = 1.0 / math.sqrt(D)

q = torch.randn(B, S_Q, H, D, dtype=dtype).transpose(1, 2)
k = torch.randn(B, S_KV, H, D, dtype=dtype).transpose(1, 2)
v = torch.randn(B, S_KV, H, D, dtype=dtype).transpose(1, 2)

o = torch.nn.functional.scaled_dot_product_attention(
    q, k, v, is_causal=False, scale=attn_scale
)
o_mlx = mx.fast.scaled_dot_product_attention(
    mx.array(q), mx.array(k), mx.array(v), scale=attn_scale, mask=None
)
torch.testing.assert_close(o, torch.tensor(o_mlx), atol=5e-3, rtol=3e-3)

o = torch.nn.functional.scaled_dot_product_attention(
    q, k, v, is_causal=True, scale=attn_scale
)
o_mlx = mx.fast.scaled_dot_product_attention(
    mx.array(q), mx.array(k), mx.array(v), scale=attn_scale, mask="causal"
)
torch.testing.assert_close(o, torch.tensor(o_mlx), atol=5e-3, rtol=3e-3)
```

harsh-sutariya:
## Proposed changes

Fixed a bug in `scaled_dot_product_attention` where the causal mask produced incorrect results when query sequence length (`S_Q`) differs from key / value sequence length (`S_KV`). Fixes #2835

**Problem**
When `mask="causal"` was used with `S_Q != S_KV`, MLX used **bottom right aligned causal masking** (where query position `i` could attend to keys `0..(kL - qL + i)`). 

While PyTorch's `is_causal=True` uses **top left aligned masking** (where query position `i` can attend to keys `0..i`). 

This caused significant numerical differences.

**Solution:**
Changed all causal mask implementations across all backends (CPU fallback, Metal vector, Metal full attention and CUDA) to use top left alignment, which matches PyTorch's behavior.

## Checklist

### DIFF
--- a/mlx/backend/cuda/scaled_dot_product_attention.cu
+++ b/mlx/backend/cuda/scaled_dot_product_attention.cu
@@ -115,7 +115,7 @@ __global__ void kernel_sdpav_1pass(
   for (int i = kv_seq_idx; i < params.kL; i += BN) {
     bool use_key = true;
     if constexpr (do_causal) {
-      use_key = i <= (params.kL - params.qL + q_seq_idx);
+      use_key = i <= q_seq_idx;
     }
 
     if (use_key) {
@@ -280,7 +280,7 @@ __global__ void kernel_sdpav_2pass_1(
   for (int i = kv_seq_idx; i < params.kL; i += blocks * BN) {
     bool use_key = true;
     if constexpr (do_causal) {
-      use_key = i <= (params.kL - params.qL + q_seq_idx);
+      use_key = i <= q_seq_idx;
     }
 
     if (use_key) {
--- a/mlx/backend/metal/kernels/sdpa_vector.h
+++ b/mlx/backend/metal/kernels/sdpa_vector.h
@@ -98,7 +98,7 @@ template <typename T, int D, int V = D>
   for (int i = simd_gid; i < N; i += BN) {
     bool use_key = true;
     if (do_causal) {
-      use_key = i <= (N - int(tpg.y) + int(q_seq_idx));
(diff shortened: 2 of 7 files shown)


awni:
I'm a little bit confused by this:

Let's say the query length is 4 and the key length is 8:
- 0th query should be able to attend to keys 0-4
- 1th query should attend to keys 0-5
- 2nd query to keys 0-6
- 3rd query to keys 0-7

So the mask should be of shape 4 x 8 and with the entries:

```
[[ True  True  True  True  True False False False]
 [ True  True  True  True  True  True False False]
 [ True  True  True  True  True  True  True False]
 [ True  True  True  True  True  True  True  True]]
```

We should not change that behavior as we rely on it for correct function in several places.

awni:
Thanks for the contribution! Closing this as per discussion in #2835. Let's revisit if we need to add it as an option in the future.

Issue #2841, opened 2025-11-26
[BUG] Memory leak in function transform with unused container


awni:
If the function being transformed updates a container and doesn't use it, and it takes as input the element from the container, then we can get a memory leak.

The reason is MLX makes copies of the input to trace the graph. The copies have previous copies as the input so the chain keeps growing.

```python
import resource
import mlx.core as mx

container = [mx.array(1.0)]

def fn(p, x):
    container[0] = p
    return x.sum()

x = mx.ones(shape=(128,))
grad_fn = mx.grad(fn)

while True:
    mx.eval(grad_fn(container[0], x))
    mem_usage_kb = resource.getrusage(resource.RUSAGE_SELF).ru_maxrss
    print(mem_usage_kb)
```

harsh-sutariya:
**Root Cause:** In `vjp()`, tracer copies are created via `copy(p, s)` for each primal input. When `container[0] = p` stores the tracer, and the next call uses `container[0]` as input, a new copy is created with the previous tracer as its input. This creates an ever growing chain:

```
Call 1: tracer_1 = copy(original)
Call 2: tracer_2 = copy(tracer_1)  → chain: tracer_2 → tracer_1 → original
Call 3: tracer_3 = copy(tracer_2)  → chain: tracer_3 → tracer_2 → tracer_1 → original
```

Each tracer holds a reference to the previous one, this is what prevents the garbage collection.

**Possible Fix:** In `vjp()` and `jvp()`, before creating the tracer copy, check if the input is a Copy from a previous (completed) tracing call. If so, traverse to the original source and copy from there instead. Use `p.is_tracer()` to distinguish active tracers (from nested transforms, which should NOT be flattened) from stale tracers from previous calls (which have `is_tracer=false` after tracing completes).

```cpp
array source = p;
if (!p.is_tracer()) {
  while (source.has_primitive() &&
         typeid(source.primitive()) == typeid(Copy) &&
         !source.inputs().empty()) {
    source = source.inputs()[0];
  }
}
primals_.push_back(copy(source, s));
```

This is safe because Copy's VJP is identity, flattening `copy(copy(copy(x)))` to `copy(x)` produces identical gradients.

This fixes the memory leak issue for me and passes all the tests, **but I'm not sure if it's the best way to solve this issue** as now every `vjp()`/`jvp()` call will have a while loop traversing potential chains.

@awni, I can create a PR if this approach looks reasonable.

PR #2853, opened 2025-12-02
Fix memory leak in transforms (#2841)

harsh-sutariya:
## Proposed changes

Fixes memory leak in function transforms when a container is updated with a tracer and that container element is reused as input (issue #2841).

**Root cause:** When `vjp()`/`jvp()` create tracer copies and a container stores the tracer, subsequent calls create a chain `copy(copy(copy(...)))` that grows unboundedly, this is what prevents the garbage collection.

**Solution:** Flatten copy chains from previous (completed) tracing calls by traversing to the original source before creating new tracers. Only applies to non active tracers to preserve nested transform behavior.

**Changes:**
- Modified `vjp()` and `jvp()` in `transforms.cpp` to flatten copy chains
- Added test `test_grad_with_container_reuse` to verify fix

Follow the discussion in issue #2841 page for more details about the problem and the solution.

## Checklist


zcbenz:
The added test passes without the change, so this PR likely did not capture the root cause.

PR #2857, opened 2025-12-03
Fix graph updating

awni:
- Fix updating property to accumulate correctly over the graph (that was a bug)
- Enable a structure like graph -> graph -> kernel node to be potentially updatable. This comes up with cudnn for some reason and it was breaking graph updating

Speed up pretraining 0.6B on B200

Pre: 106511.44 tok/s
Post: 112456.45 tok/s


zcbenz (approved):
👍

zcbenz:
> * Enable a structure like graph -> graph -> kernel node to be potentially updatable. This comes up with cudnn for some reason and it was breaking graph updating

This comes from cuDNN's frontend API which creates a temporary CUDA graph for no reason:
https://github.com/NVIDIA/cudnn-frontend/blob/be6c079be8aaffa0fc079fcf039887e637c289c7/include/cudnn_frontend/graph_interface.h#L588-L597

PR #2859, opened 2025-12-03
Make CommandEncoder.add_temporary return reference

zcbenz:
This PR tweaks on how we create temporary arrays.

Before:

```c++
array x = x_pre;
if (needs_copy) {
  x = copy(x);
  encoder.add_temporary(x);
}
```

After:

```c++
const array& x = needs_copy ? encoder.add_temporary(copy(x)) : x_pre;
```

Benefits are:

* 1~2 less copies of `array` per each input.
* Simplified code for lots of cases.

### DIFF
--- a/mlx/backend/cuda/device.h
+++ b/mlx/backend/cuda/device.h
@@ -13,6 +13,7 @@
 #include <cudnn.h>
 #include <thrust/execution_policy.h>
 
+#include <deque>
 #include <unordered_map>
 
 namespace mlx::core::cu {
@@ -79,8 +80,9 @@ class CommandEncoder {
 
   void add_graph_node(cudaGraph_t child);
 
-  void add_temporary(const array& arr) {
-    temporaries_.push_back(arr.data_shared_ptr());
+  array& add_temporary(array arr) {
+    temporaries_.push_back(std::move(arr));
+    return temporaries_.back();
   }
 
   void add_completed_handler(std::function<void()> task);
@@ -126,7 +128,7 @@ class CommandEncoder {
   std::string graph_nodes_key_;
   std::string graph_deps_key_;
   std::vector<GraphNode> concurrent_nodes_;
-  std::vector<std::shared_ptr<array::Data>> temporaries_;
+  std::deque<array> temporaries_;
   LRUCache<std::string, CudaGraphExec> graph_cache_;
   std::vector<std::uintptr_t> active_deps_;
   std::vector<std::uintptr_t> active_outputs_;
--- a/mlx/backend/cpu/distributed.cpp
+++ b/mlx/backend/cpu/distributed.cpp
@@ -9,11 +9,12 @@
 
 namespace mlx::core::distributed {
 
-std::pair<array, bool> ensure_row_contiguous(const array& arr, Stream stream) {
+const array& ensure_row_contiguous(const array& arr, Stream stream) {
   if (arr.flags().row_contiguous) {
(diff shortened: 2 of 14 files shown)


angeloskath:
Hm nice. I might have missed something but won't references be potentially invalidated by pushing to the vector? I know that I was doing the following pattern quite a lot before adding the temporaries to the encoder

```cpp
std::vector<array> tmps;
tmps.reserve(number_of_potential_temporaries);

auto maybe_tmp = [&tmps](...) -> array& {
  tmps.push_back(...);
  return tmps.back();
};
```

angeloskath:
Maybe we should change it to a `std::deque` 🤔

zcbenz:
Yeah you are right reference will get invalidated with `std::vector` 🤦 , I changed to use `std::deque`.

mlx/backend/cuda/device.h
```diff
@@ -126,7 +128,7 @@ class CommandEncoder {
   std::string graph_nodes_key_;
   std::string graph_deps_key_;
   std::vector<GraphNode> concurrent_nodes_;
-  std::vector<std::shared_ptr<array::Data>> temporaries_;
+  std::deque<array> temporaries_;
```
angeloskath, line 131: I think that change might have deeper consequences than what seems obvious. Basically, in `eval.cpp` we use `add_temporary` to hold the buffers of the inputs until they are evaluated. However holding the arrays is quite different. For instance the siblings will be added many times which prevents them from being donated.

I do love the change though 🤔

I think perhaps changing the handling of the buffers like for Metal and then keeping arrays in the temporaries might be all we need to fix this.

zcbenz: Can you elaborate why keeping temporaries as arrays does not prevent them to be donated in Metal?

angeloskath: Yeah good question. So firstly, it might. It is not a problem that we come across often cause the temporary is usually allocated and used in place so the whole point is to allocate it no donation. But 100% having the array and having added it as temporary would break donation (same goes for the buffer though so maybe I was just too cautious).

For instance

```c++
array x = ...
encoder.add_temporary(x);
copy_gpu(x, out, ...); // no donation here
```

Your change actually fixes that
```c++
const array& x = encoder.add_temporary(...);
copy_gpu(x, out, ...); // donation here yay
```

However, if you see the `metal/eval.cpp` there we don't use the `add_temporary` mechanism but we save the data pointers. I initially thought there were deeper reasons for it (there still may be) but the more I look into it it seems that saving the arrays would be fine.

Just as extra context the whole point is for the `eval_impl` in `transforms.cpp` to not keep extraneous copies of data pointers or arrays so that we know if there is only 1 reference to it during our `eval_gpu` then we can donate it. 

The way this is achieved is that we keep references in the `tape` and we move them out as we process the tape. Then the arrays themselves have references to their inputs and siblings which are removed when we call `detach()`. 

So I guess keeping arrays would be the same since the donation requirement is 1 reference to the array AND 1 reference to the data.

angeloskath: Btw the TL;DR is you are probably right and it is the same effect. My fear is that an array might keep references of other arrays but that should be handled by `detach` anyway.

zcbenz: Thanks a lot for the detailed explanations! I think I will need to analyze the actual affects to donations before continuing this PR, I'm closing it for now.

PR #2860, opened 2025-12-03
new[CI]: add linux sanitizer tests

incertum:
## Proposed changes

new[CI]: add linux sanitizer tests

CC @madrob 

## Checklist

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -36,14 +36,6 @@ target_sources(
           linalg_tests.cpp
           ${METAL_TEST_SOURCES})
 
-if(MLX_BUILD_CUDA)
-  # C++ tests are always built from source, so we have to specify where to find
-  # CCCL headers for JIT as they are not installed in system.
-  get_target_property(MLX_CCCL_DIR mlx CCCL_DIR)
-  target_compile_definitions(mlx PRIVATE MLX_CCCL_DIR="${MLX_CCCL_DIR}")
-  message(STATUS MLX_CCCL_DIR="${MLX_CCCL_DIR}")
-endif()
-
 target_link_libraries(tests PRIVATE mlx doctest)
 target_compile_options(tests PRIVATE ${SANITIZER_COMPILE_FLAGS})
 target_link_options(tests PRIVATE ${SANITIZER_LINK_FLAGS})


.github/scripts/build-sanitizer-tests.sh
```diff
@@ -0,0 +1,48 @@
...
+
+  rm -rf build
+  mkdir -p build
+  pushd build > /dev/null
+
+  cmake .. ${BASE_CMAKE_ARGS} ${cmake_sanitizer_flag}
+  make -j $(nproc)
+  ./tests/tests
+
+  popd > /dev/null
```
madrob, line 36: This is fine since there's nothing really happening afterwards and it is in a one-shot CI context, but I think more modern practice is to use a sub shell which will eject you out of the directory even on a command failure:

```
(
  cd build
  cmake .. ${BASE_CMAKE_ARGS} ${cmake_sanitizer_flag}
  make -j $(nproc)
  ./tests/tests
)
```

madrob:
Did this find any bugs?

CMakeLists.txt
```diff
@@ -45,6 +45,9 @@ option(MLX_METAL_JIT "Use JIT compilation for Metal kernels" OFF)
 option(MLX_USE_CCACHE "Use CCache for compilation cache when available" ON)
 option(BUILD_SHARED_LIBS "Build mlx as a shared library" OFF)
 option(USE_SYSTEM_FMT "Use system's provided fmt library" OFF)
+option(USE_ASAN "Enable AddressSanitizer (ASan)" OFF)
+option(USE_UBSAN "Enable UndefinedBehaviorSanitizer (UBSan)" OFF)
+option(USE_TSAN "Enable ThreadSanitizer (TSan)" OFF)
```
awni, line 51: These flags will only affect the test binary right? Will they propagate into the core library as well? 

I feel like this diff is just checking the test executables but not the core library for undefined behavior, etc.. but maybe I'm mistaken..

incertum: @awni 

I  thought so too, but then I got errors when I tried it:

```
target_compile_options(mlx PRIVATE ${SANITIZER_COMPILE_FLAGS})
target_link_options(mlx PRIVATE ${SANITIZER_LINK_FLAGS})

clang++: error: linker command failed with exit code 1 (use -v to see invocation)
```

Today, with a fresh look, I realized I had simply used PRIVATE when it should have been PUBLIC ✅ :

```
target_compile_options(mlx PRIVATE -Wno-psabi)
target_compile_options(mlx PUBLIC ${SANITIZER_COMPILE_FLAGS})
target_link_options(mlx PUBLIC ${SANITIZER_LINK_FLAGS})
```

incertum:
> Did this find any bugs?

Now with the fix on top checkout https://github.com/ml-explore/mlx/actions/runs/19943012002/job/57185456349?pr=2860 ... it seems that we could improve a few things, however note that to be more serious we would need to add fuzzing! Also note that for now I turned of the leak detector etc (see settings).

incertum:
@awni: @madrob and I chatted and are proposing to defer enabling ASAN and TSAN CI checks until follow up PR(s) where issues will get addressed. 
We left UBSAN as it's current configured to not halt on error as UBSAN is generally more tricky to address, but that way we can view the logs if we wanted to. 

What do you think?

awni:
Fine with me. Do you plan do to send some of those follow ups?

incertum:
> Fine with me. Do you plan do to send some of those follow ups?

Yes I'll try to look into it :)

incertum:
@angeloskath and @awni I have just rebased this PR and the ASAN tests (CPU backend) look good now. The todo comment is also up-to-date plus I created https://github.com/ml-explore/mlx/issues/2915.

What would be needed from your perspective to get a v1 of these new checks merged in?

PR #2866, opened 2025-12-03
Do a PyPi release for cuda on arm

awni:
We already release the front-end package for Arm so this is a minimal change.

### DIFF
--- a/.github/actions/setup-linux/action.yml
+++ b/.github/actions/setup-linux/action.yml
@@ -15,6 +15,7 @@ runs:
   using: "composite"
   steps:
     - name: Use ccache
+      if: ${{ runner.arch == 'x86_64' }}
       uses: hendrikmuhs/ccache-action@v1.2
       with:
         key: ccache-${{ runner.os }}-${{ runner.arch }}-${{ inputs.toolkit }}-py${{ inputs.python-version }}
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -128,7 +128,10 @@ jobs:
 
   build_cuda_release:
     if: github.repository == 'ml-explore/mlx'
-    runs-on: ubuntu-22-large
+    strategy:
+      matrix:
+        arch: ['x86_64', 'aarch64']
+    runs-on: ${{ matrix.arch == 'x86_64' && 'ubuntu-22-large' || 'ubuntu-22-large-arm' }}
     env:
       PYPI_RELEASE: 1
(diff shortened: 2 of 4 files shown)


.github/workflows/release.yml
```diff
@@ -128,7 +128,10 @@ jobs:
 
   build_cuda_release:
     if: github.repository == 'ml-explore/mlx'
-    runs-on: ubuntu-22-large
+    strategy:
```
madrob, line 131: These aren’t the larger runners so I think you are going to run out of disk space.

angeloskath (approved):
🚀

.github/workflows/release.yml
awni: I'll do a test release and check.

awni: @madrob is there an arm version of `ubuntu-22-large`?

.github/workflows/release.yml
zcbenz, line 144: The arch is not passed.

```
        with:
            arch: ${{ matrix.arch }}
```

awni: Oops! Good catch.

awni:
@madrob can you make an `ubuntu-22-large-arm` so we can use runners with more disk space?

.github/workflows/release.yml
```diff
@@ -128,7 +128,10 @@ jobs:
...
+      matrix:
+        arch: ['x86_64', 'aarch64']
+    runs-on: ${{ matrix.arch == 'x86_64' && 'ubuntu-22-large' || 'ubuntu-22-large-arm' }}
```
madrob, line 134: I would put the conditional logic in the matrix so it's easier to see:
```suggestion
        include:
          - host: ubuntu-22-large
            arch: x86_64
          - host: ubuntu-22-large-arm
            arch: aarch64
    runs-on: ${{ matrix.host }}
```

.github/actions/setup-linux/action.yml
```diff
@@ -15,6 +15,7 @@ runs:
   using: "composite"
   steps:
     - name: Use ccache
+      if: ${{ runner.arch == 'x86_64' }}
```
zcbenz, line 18: @awni This disables ccache completely as `x86_64` is not a valid value for `runner.arch`. Do you remember why it is disabled for arm build? Maybe we should add an input to just disable ccache for release builds?

awni: I think because it wasn't installed on the ubuntu-large-arm machine (the one we use for the release build).

PR #2869, opened 2025-12-04
Try not to fail when there should be memory available

awni:
There are a couple cases where MLX can fail with OOM when there is actually memory available:

- Sometimes `cudaMallocAsync` returns `nullptr` even when there should be enough RAM outside the cache + used memory. I believe this is due to fragmentation. Instead of failing on this case, we free from the cache then try again.
- Sometimes kernel / graph execution fails due to OOM (very curious here what could cause that). If the OS reported free memory is below a limit then we clear from the cache if possible.

### DIFF
--- a/mlx/backend/cuda/allocator.cpp
+++ b/mlx/backend/cuda/allocator.cpp
@@ -21,14 +21,14 @@ constexpr int page_size = 16384;
 constexpr int small_block_size = 8;
 
 #if CUDART_VERSION >= 13000
-cudaMemLocation cuda_mem_loc(int i) {
+inline cudaMemLocation cuda_mem_loc(int i) {
   cudaMemLocation loc;
   loc.type = cudaMemLocationTypeDevice;
   loc.id = i;
   return loc;
 }
 #else
-int cuda_mem_loc(int i) {
+inline int cuda_mem_loc(int i) {
   return i;
 }
 #endif // CUDART_VERSION >= 13000


awni:
I think there is a performance issue here so moving into draft.

awni:
Evidently calling `CHECK_CUDA_ERROR(cudaMemGetInfo(&free, &total));` in `malloc` is a bad idea :(.

awni:
Ok I fixed this and I don't see a regression in perf.

I think the basic premise for what is happening is that even though the MLX cache + active memory is well under the limit, there is fragmentation and since we are using async free, the device is not able return memory to the OS before every time we call malloc, and so CUDA can fail to allocate free memory even when the total amount of free memory exceeds the requested allocation.

awni:
I uploaded a script that repro's the issue on B200 (and should on H100 for smaller batch size). Just leaving it here for reference.

[run.py](https://github.com/user-attachments/files/23972576/run.py)

mlx/backend/cuda/allocator.cpp
```diff
@@ -20,6 +20,19 @@ constexpr int page_size = 16384;
...
 
+#if CUDART_VERSION >= 13000
+cudaMemLocation cuda_mem_loc(int i) {
+  cudaMemLocation loc;
+  loc.type = cudaMemLocationTypeDevice;
+  loc.id = i;
+  return loc;
+}
+#else
+int cuda_mem_loc(int i) {
```
zcbenz: nitpick: add inline.

mlx/backend/cuda/allocator.cpp
```diff
@@ -169,6 +180,21 @@ CudaAllocator::malloc_async(size_t size, int device, cudaStream_t stream) {
...
+
+    // If any cuda memory pool has too much reserved memory, clear some
+    // memory from the cache. This prevents graph / kernel execution failing
+    // from OOM
+    if (get_cache_memory() > 0) {
+      for (auto p : mem_pools_) {
+        size_t used = 0;
+        CHECK_CUDA_ERROR(cudaMemPoolGetAttribute(
+            p, cudaMemPoolAttrReservedMemCurrent, &used));
+        if (used > (total_memory_ - free_limit_)) {
```
zcbenz, line 192: Why having a `free_limit_`? The code would read easier for me if it is just:

```c++
if (used > memory_limit_) {
  buffer_cache_.release_cached_buffers(total_memory_ - memory_limit_);
}
```

zcbenz (approved):
Looks good to me!

mlx/backend/cuda/allocator.cpp
awni: Good question. `memory_limit_` can change (the user can set the memory limit to be higher or lower). I wanted a value that was fixed based on the total device memory.

zcbenz: What do you think about using `hard_memory_limit_`/`soft_memory_limit_`? (Just being nitpick, I'm good with `free_limit_` too.)

awni: I don't really love `hard_memory_limit` cause it's not a hard limit.

It's more like a soft memory limit on the underlying cuda pool. I'll think a bit more on how to phrase it.

awni: The other thing that is a bit of a mess in our allocator especially is how we deal with multi-device on a discrete setup where each device has it's own memory.

I think at some point it might make sense to have separate buffer cache for each device and one for the managed allocator.

zcbenz: That makes sense!

Issue #2864, opened 2025-12-03
[BUG] compile seems broken in latest mlx version 0.30.0


soumya-batra:
**Describe the bug**
mlx.core.compile is broken in latest update of mlx from 0.29.3 to 0.30.0

In 0.30.0 mlx version, I get the following error when running a compiled model:
    loss, grads = mini_batch_step(x, y)
                  ^^^^^^^^^^^^^^^^^^^^^
IndexError: unordered_map::at: key not found

In 0.29.3 mlx version, I get the expected output as follows (ignore loss values as I'm using random data here):
step 0, loss = 11.323957443237305
step 1, loss = 11.33563232421875
step 2, loss = 11.418798446655273

**To Reproduce**

Here is minimal code to reproduce the bug:

```python
from functools import partial
import mlx.core as mx
import mlx.nn as nn
import mlx.optimizers as optim
from mlx.utils import tree_map
from mlx.nn.losses import cross_entropy

import mlx.core as mx

from dataclasses import dataclass


def cross_entropy_loss(model, input_idx, target_idx):
    logits = model(input_idx).astype(mx.float32)
    loss = cross_entropy(logits.reshape(-1, logits.shape[-1]), target_idx.reshape(-1), reduction="mean")
    loss = loss.astype(mx.float32)
    return loss

def get_random_batch(batch_size, context_length):
    buf = mx.random.randint(low=0, high=50257, shape=(batch_size * context_length + 1,))
    arr1 = buf[:-1].reshape(batch_size, context_length)
    arr2 = buf[1:].reshape(batch_size, context_length)
    return arr1, arr2


@dataclass
class GPT2Config:
    context_length: int = 1024
    vocab_size: int = 50304
    num_layers: int = 12
    num_heads: int = 12
    embedding_dim: int = 768


class MLP(nn.Module):
    def __init__(self, config):
        super().__init__()
        self.c_fc = nn.Linear(config.embedding_dim, 4 * config.embedding_dim)
        self.gelu = nn.GELU(approx="tanh")
        self.c_proj = nn.Linear(4 * config.embedding_dim, config.embedding_dim)

    def __call__(self, input):
        output = self.c_fc(input)
        output = self.gelu(output)
        output = self.c_proj(output)
        return output


class TransformerBlock(nn.Module):
    def __init__(self, config):
        super().__init__()
        self.ln_1 = nn.LayerNorm(config.embedding_dim)
[...]

PR #2871, opened 2025-12-04
fix compile copying

awni:
Fix compile for some cases with primitives with siblings. Closes https://github.com/ml-explore/mlx/issues/2864.


angeloskath (approved):
🙏

PR #2872, opened 2025-12-04
[CUDA] Release build for cuda 13

awni:
For CUDA 12:

- pip install mlx[cuda]
- pip install mlx[cuda12]

For CUDA 13:

- pip install mlx[cuda13]

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -131,7 +131,7 @@ jobs:
     strategy:
       matrix:
         arch: ['x86_64', 'aarch64']
-        toolkit: ['12.9', '13.0']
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -131,6 +131,7 @@ jobs:
     strategy:
       matrix:
         arch: ['x86_64', 'aarch64']
+        toolkit: ['12.9', '13.0']
```
zcbenz: The action needs the `cuda-` prefix to know to install cuda.

```suggestion
        toolkit: ['cuda-12.9', 'cuda-13.0']
```

awni: Good catch!

zcbenz (approved):
👍

PR #2874, opened 2025-12-05
Make allocator::malloc throw on allocation failure

zcbenz:
`Allocator::malloc` returned null buffer when allocation failed, which was designed to make `malloc_or_wait` API work. Since the API had been removed in https://github.com/ml-explore/mlx/pull/1976, this behavior is no longer needed and the code should just throw in case of allocation failure.


mlx/backend/cuda/allocator.cpp
```diff
@@ -157,17 +157,11 @@ CudaAllocator::malloc_async(size_t size, int device, cudaStream_t stream) {
...
       } else {
-        err = cudaMallocAsync(&data, size, stream);
-      }
-      if (err != cudaSuccess && err != cudaErrorMemoryAllocation) {
-        throw std::runtime_error(fmt::format(
-            "cudaMallocManaged failed: {}.", cudaGetErrorString(err)));
-      }
-      if (!data) {
-        return Buffer{nullptr};
+        CHECK_CUDA_ERROR(cudaMallocAsync(&data, size, stream));
```
awni, line 162: Does cudaMallocAsync return an error when the allocation fails? It might just return a nullptr without an error?

zcbenz: I think the API guarantees a non-null result if the call succeeds:

> A pointer to the allocated memory is returned immediately in *dptr. The allocation must not be accessed until the the allocation operation completes. The allocation comes from the memory pool associated with the stream's device.

But yeah we should still check the data pointer, should not trust them on this.

PR #2875, opened 2025-12-05
[Metal] No copy array init

### LINKED ISSUE #2855: [feature] support for no-copy mlx::array initialization
There are a couple use cases where it would be handy to have no-copy initialization of an mlx::array (specifically from mlx-swift but possibly useful elsewhere):

- reading a image from disk
    - currently we rasterize into a temporary buffer (copy) and then create an MLXArray which copies the buffer into MLX owned memory

- receiving an image from the camera
    - using macOS/iOS camera hardware gets you a CVPixelBuffer (which is an IOSurface, which is roughly the same sort of thing as an MTLBuffer)
    - we can potentially get access to the inner pointer of the IOSurface and initialize the MLXArray from that (copy)

It would be nice if there was a way to do this without the extra copy.  Some ideas:

- mlx::array provides a callback mechanism to fill MLX owned memory (works for the first case, not the second)

- (probably better) we expose this constructor:

```
  explicit array(
      allocator::Buffer data,
      Shape shape,
      Dtype dtype,
      Deleter deleter = allocator::free);
```

- continued
    - and have a way to prepare an `allocator::Buffer`
    - some callers would find it convenient to get an MTL compatible buffer (e.g. via the MLX allocator) to fill (case 1)
    - other callers would be able to wrap their own memory and provide a custom `Deleter` (case 2 where the Deleter closure would retain the IOSurface)

awni:
Adds a new constructor which accepts a raw pointer (and shape, dtype and deleter) and will make an array without a copy when possible.

Attempts to close #2855

See the test case for example usage.

@davidkoski I think with this new constructor we should be good to go. 

- If the user wants to get a buffer from MLX and fill it in that is already supported via existing public APIs:

```c++
auto buffer = allocator::malloc(size);
// Do some stuff with buffer.raw_ptr()
auto arr = array(buffer, ...);
```

We may wan't to make the above a little simpler.. for example we could provide a way to construct an empty array:

```c++
auto arr = array(shape, dtype);
// Do some stuff with arr.data<T>()
```

- If the user wants to provide a buffer from elsewhere that is now possible with this new constructor.
```c++
auto arr = array(void_ptr, shape, dtype, deleter);
```

### DIFF
--- a/mlx/array.h
+++ b/mlx/array.h
@@ -57,7 +57,10 @@ class array {
       Shape shape,
       Dtype dtype = TypeToDtype<T>());
 
-  /* Build an array from a raw pointer without a copy */
+  /* Build an array from a raw pointer. The constructor will attempt to use the
+   * input data without a copy. The deleter will be called when the array no
+   * longer needs the underlying memory - after the array is destroyed in the
+   * no-copy case and after the copy otherwise. */
   explicit array(
       void* data,
       Shape shape,
--- a/mlx/backend/metal/allocator.cpp
+++ b/mlx/backend/metal/allocator.cpp
@@ -224,9 +224,6 @@ void MetalAllocator::release(Buffer buffer) {
   std::unique_lock lk(mutex_);
   active_memory_ -= buf->length();
   num_resources_--;
-  if (!buf->heap()) {
-    residency_set_.erase(buf);
-  }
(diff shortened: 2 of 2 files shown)


davidkoski:
@andresy this would be great to add to the mlx-c API when it is available!

PR #2877, opened 2025-12-06
fix: possible heap-buffer-overflow in RandomBits::eval_cpu (follow for new ASAN CI tests)

incertum:
## Proposed changes

Follow for https://github.com/ml-explore/mlx/pull/2860, see ASAN issue in a PR run where we had the new CI check enabled: https://github.com/ml-explore/mlx/actions/runs/19943012002/job/57185456349?pr=2860

---

My understanding of the logic I am attempting to improve:
(@awni  please call me out if I got it wrong)

```
std::tie(ptr[count.first], ptr[count.second]) =
            random::threefry2x32_hash(key, count);
```

… is filling the buffer from both ends inward (threefry2x32_hash always returns a pair).

The loop stops just before the pointers from the start and end would cross over or meet. Therefore, the two if statements after it are for handling the “leftovers”.

The first if statement `(if (count.first < half_size) {)` handles the case where we have 2 slots empty in the middle, e.g. `[ X, X, X, _, _, X, X, X ]`.

both `rb.first` and `rb.second` are used because we are filling the middle pair of slots.

The second if statement `(if (!even) {)` is where `ASan` triggered.

There is an odd number of slots and only one is left to fill in the middle, e.g. `[ X, _, X]`.
Here we need to generate a new final random number on the fly and use it to fill that unique middle slot (safely).

---

Proposed Remediation:

The proposed fix aims to remove unsafe, unconditional 4-byte writes. Got rid of direct assignment -- now first confirming if the write position is even valid, then calculating bytes remaining in the buffer and copying only what fits (ideally full 4 bytes, but less if it wouldn’t fit).

Locally I tested it with the ASan and it seems to be fixed now.

CC @madrob 

## Checklist

### DIFF
--- a/mlx/backend/cpu/primitives.cpp
+++ b/mlx/backend/cpu/primitives.cpp
@@ -291,6 +291,17 @@ void RandomBits::eval_cpu(const std::vector<array>& inputs, array& out) {
                     num_keys,
                     kshape = keys.shape(),
                     kstrides = keys.strides()]() mutable {
+    auto copy_remaining = [&](char* cptr, size_t loc, uint32_t v) {
+      if (4 * loc + 4 <= bytes_per_key) {
+        reinterpret_cast<uint32_t*>(cptr)[loc] = v;
+      } else {
+        std::copy(
+            reinterpret_cast<char*>(&v),
+            reinterpret_cast<char*>(&v) + bytes_per_key - 4 * loc,
+            cptr + 4 * loc);
+      }
+    };
+
     size_t out_skip = (bytes_per_key + 4 - 1) / 4;
     auto half_size = out_skip / 2;
     bool even = out_skip % 2 == 0;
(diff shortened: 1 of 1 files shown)


angeloskath:
@incertum I think the proposed fix is a bit of an overkill. The OOB write happens indeed when `!even` because there is no check for `bytes_per_key % 4 == 0`. That is all that is needed imho similar to the check that was already there in line 313.

angeloskath:
Admittedly it doesn't get much simpler... sorry for the delay. Thanks for the fix, I 'll merge after the tests pass.

incertum:
@angeloskath thanks a bunch for getting this over the finish line while I was out ❤️ . I have just rebased https://github.com/ml-explore/mlx/pull/2860 that adds the new CI.

PR #2882, opened 2025-12-07
Metal logger

CC-Yeh:
## Proposed changes

**Draft PR for discussion** – this introduces an opt-in logging path for Metal kernels using `os_log`, with a small set of kernel-friendly macros. The goal is to make debugging Metal kernels easier while keeping **zero runtime overhead** when logging is disabled.

---

### Motivation

This PR experiments with an `os_log`-based approach that:

* Requires **minimal changes** on the kernel developer side (just include a header and call macros).
* Is **opt-in and debug-only**, and compiled out completely when disabled.
* Integrates with the host side so that faults and errors can surface as C++ exceptions with context.

---

### Design overview

* **Logging backend:**
  Uses `os_log` / `os_log_error` / `os_log_fault` from Metal shaders.

* **Compile-time opt-in:**

  * CMake flag: `MLX_METAL_LOGGER_ENABLED=ON`
  * For JIT builds: pass `MLX_METAL_LOG_ENABLED=1` at run time.

* **Zero-overhead when disabled:**

  * When logging is disabled, the logging macros expand to empty `do { } while (0)` blocks.
  * No extra kernel arguments, no extra host-side checks, no branching in the hot path.

* **Error propagation:**

  * All messages go to the standard log console.
  * If any *error* or *fault* log is emitted during a command buffer, the host side throws an exception and prints all collected messages (across log levels) for that command buffer.

---

### API / Usage

**In a Metal kernel**

1. Include the logging header:

```metal
#include "mlx/backend/metal/kernels/logging.h"
```

2. Use one of the provided macros, for example:

```metal
MLX_METAL_KERNEL_LOG_DEBUG("debug message: value=%d", value);
MLX_METAL_KERNEL_LOG_ERROR("something went wrong: code=%d", code);
MLX_METAL_KERNEL_LOG_FAULT("fatal error in kernel X");
```

**Build configuration**

* CMake:

```bash
- DMLX_METAL_LOGGER_ENABLED=ON
```

* JIT compilation: ensure

```bash
MLX_METAL_LOG_ENABLED=1 python ...
```

is set before compiling the Metal sources.

**Checking logs**

* All log messages appear in the OS log console.
* If an error or fault is logged, the host:

  * Detects it in the completion handler.
  * Throws a C++ exception.
  * Dumps all logs (all levels) associated with the command buffer.

---

### Limitations / Notes

* **Not performant – do not use in production.**
  This is intended purely for debugging. `os_log` has non-trivial overhead.

* **Potential log flooding:**
  `os_log` is called from each thread, so a single log statement inside a hot loop can generate many messages and flood the buffer.

  * A robust solution likely requires a **kernel-wide atomic guard** (e.g., “log only the first N occurrences”), but that would require passing additional arguments or shared state into the kernel, which hurts API simplicity.
  * For this first draft, we deliberately avoid extra kernel arguments and keep the API minimal, at the cost of possible log spam.

* **Scope of this PR:**
  This is mainly to **shape the API and discuss design**. 

---

### Open questions for reviewers

* Is throwing on *any* error/fault log too aggressive? Should we make this configurable?
* Are you okay with the current trade-off (simple macros, no extra kernel args) vs. potentially noisy logs?

---

## Checklist

### DIFF
--- a/mlx/backend/metal/device.cpp
+++ b/mlx/backend/metal/device.cpp
@@ -394,16 +394,11 @@ bool Device::command_buffer_needs_commit(int index) {
 MTL::CommandBuffer* Device::get_command_buffer(int index) {
   auto& stream = get_stream_(index);
   if (stream.buffer == nullptr) {
-    MTL::CommandBuffer* buffer = nullptr;
 #ifdef MLX_METAL_LOG_ENABLED
-    if (logger_) {
-      buffer = logger_->make_buffer_with_logging(stream.queue, device_);
-    }
+    stream.buffer = logger_->create_logged_command_buffer(stream.queue);
+#else
+    stream.buffer = stream.queue->commandBufferWithUnretainedReferences();
 #endif
-    if (buffer == nullptr) {
-      buffer = stream.queue->commandBufferWithUnretainedReferences();
-    }
-    stream.buffer = buffer;
     if (!stream.buffer) {
       throw std::runtime_error(
           "[metal::Device] Unable to create new command buffer");
@@ -417,7 +412,7 @@ MTL::CommandBuffer* Device::get_command_buffer(int index) {
 void Device::commit_command_buffer(int index) {
   auto& stream = get_stream_(index);
 #ifdef MLX_METAL_LOG_ENABLED
-  logger_->register_completion(stream.buffer);
+  logger_->install_completion_handler(stream.buffer);
 #endif
   stream.buffer->commit();
   stream.buffer->release();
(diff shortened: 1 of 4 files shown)


CC-Yeh:
Hey @awni @angeloskath 

I just opened a draft PR to prototype logging from Metal kernels.

Would love your thoughts on the overall API and whether this direction makes sense.

awni:
@CC-Yeh I think this is overkill for what we need for MLX. 

Based on the logger API we can do something a lot simpler:
- Log in kernels using e.g. `os_log_default.log_debug`
- For DEBUG builds add the os log compiler flag (also add it to the JIT compiler for debug builds)
- Run with e.g. the following:

```
MTL_LOG_LEVEL=MTLLogLevelInfo MTL_LOG_TO_STDERR=1
```

awni:
I think it would be cool and useful to add some logging for kernels that gets used in debug builds. If you are up for redoing your PR (or making a new one) that does something more like the above that would be great!

CC-Yeh:
Superseded by #2904

PR #2885, opened 2025-12-09
load metallib not only as resource but in root of a bundle

gin66:
## Proposed changes

This is in relation to mlx-swift. MetalCompilerPlugin does not generate a resource, but stores the lib in the bundle root. This patch adds bundle root to the search locations for the default.metallib.

## Checklist

### DIFF
--- a/mlx/backend/metal/device.cpp
+++ b/mlx/backend/metal/device.cpp
@@ -57,6 +57,16 @@ MTL::Library* try_load_bundle(
     MTL::Device* device,
     NS::URL* url,
     const std::string& lib_name) {
+  std::string path =
+    std::string(url->fileSystemRepresentation()) +
+    "/" + lib_name + ".metallib";
+
+  auto [lib, error] = load_library_from_path(device, path.c_str());
+  if (lib) {
+    return lib;
+  }
+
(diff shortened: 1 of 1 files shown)


davidkoski:
I think we should hold onto this until we discuss on the mlx-swift side.  I am concerned that this works around a build issue in the metal builder plugin -- perhaps we should just figure out how to get that to produce the correct path.

See also https://github.com/ml-explore/mlx-swift/pull/313

gin66:
I would appreciate to get this implemented. It has no negative side effects and the code after my comment could be kept, even though I wonder, if this is a leftover of an older code basis.

zcbenz:
At the moment I prefer adding an API to set the path to metallib directly, i.e. https://github.com/ml-explore/mlx/pull/3597, and I plan to close this PR, please let me know if the API would not be able to solve your use case.

davidkoski:
See comment from #3562:

> I think API to set a path would be fine, but I don't know how early it needs to be called. IIRC the worker threads start via static initializers and there is no way to guarantee being able to get in before that (perhaps that has changed?). Anyway I think the automatic scheme we have is ok and having an override would be useful.

PR #2892, opened 2025-12-09
fp quantize

nastya236:
Quantization kernels for nvfp4 and mxfp8 (for larger shapes). The difference can be noticeable, for example, for activations / activation gradients quantization (e.g., `M = 4096 * 8` corresponds to sequence length 4096 and batch size 8).
Notes:

- Stochastic rounding support is partially implemented in `nvfp4_quantized.cuh`: the quantization function is templated on `USE_SR`, but the logic for random bit generation is not implemented yet, so `USE_SR` is currently always instantiated as `false`. It serves as a placeholder for a future stochastic-rounding implementation. Could be removed if you think it is unnecessary for now.
- For small arrays, vectorization introduces overhead (plot below), so it is better to launch one thread per element (as before). Therefore, there are two paths: for small matrices (`w.size() < 1024 * 512`), we use the non-vectorized kernel. This threshold is based on empirical observation and is ~consistent across types and for both mxfp8 and nvfp4 quantization, but I am happy to adjust this. 
- For mxfp8 no ptx implementation yet.

 Screenshot 2025-12-09 at 21 42 20

### DIFF
--- a/mlx/backend/cuda/quantized/mxfp8_quantize.cuh
+++ b/mlx/backend/cuda/quantized/mxfp8_quantize.cuh
@@ -12,18 +12,12 @@ template <typename T>
 __device__ __forceinline__ uint32_t
 scale_cvt_Tx4_to_fp8x4_fallback(const Vector4_t<T> input, const float scale) {
   uint32_t out_fp8x4 = 0;
-  Vector4_t<float> scaled;
+  float4 scaled;
   scaled.x = static_cast<float>(input.x) * scale;
   scaled.y = static_cast<float>(input.y) * scale;
   scaled.z = static_cast<float>(input.z) * scale;
   scaled.w = static_cast<float>(input.w) * scale;
-  uint8_t q0 = __nv_fp8_e4m3(scaled.x).__x;
-  uint8_t q1 = __nv_fp8_e4m3(scaled.y).__x;
-  uint8_t q2 = __nv_fp8_e4m3(scaled.z).__x;
-  uint8_t q3 = __nv_fp8_e4m3(scaled.w).__x;
-  out_fp8x4 = (static_cast<uint32_t>(q3) << 24) |
-      (static_cast<uint32_t>(q2) << 16) | (static_cast<uint32_t>(q1) << 8) |
-      static_cast<uint32_t>(q0);
+  out_fp8x4 = __nv_fp8x4_e4m3(scaled).__x;
   return out_fp8x4;
 }
 
--- a/mlx/backend/cuda/quantized/nvfp4_quantize.cuh
+++ b/mlx/backend/cuda/quantized/nvfp4_quantize.cuh
@@ -33,7 +33,7 @@ scale_cvt_Tx4_to_fp4x4_fallback(const Vector4_t<T> input, const float scale) {
 }
 
 #if (CUDART_VERSION >= 12080) && (__CUDA_ARCH__ >= 1000) && \
-    (__CUDA_ARCH_FAMILY_SPECIFIC__ >= 1000)
+    defined(__CUDA_ARCH_SPECIFIC__)
 
 __device__ __forceinline__ uint16_t
 scale_cvt_bf16x4_to_fp4x4_rn(const bf16x4 input_bf16x4, const float2 scale) {
--- a/mlx/backend/cuda/quantized/quantized_utils.cuh
+++ b/mlx/backend/cuda/quantized/quantized_utils.cuh
@@ -17,16 +17,12 @@ inline constexpr __device__ short get_bytes_per_pack() {
 
 template <typename T>
 __device__ __forceinline__ void abs_max_x2(T& out, const T& x1, const T& x2) {
-  if constexpr (std::is_same<T, __nv_bfloat162>::value) {
-    asm volatile("max.xorsign.abs.bf16x2 %0, %1, %2;\n"
-                 : "=r"(reinterpret_cast<uint32_t&>(out))
-                 : "r"(reinterpret_cast<const uint32_t&>(x1)),
-                   "r"(reinterpret_cast<const uint32_t&>(x2)));
-  } else if constexpr (std::is_same<T, __half2>::value) {
-    asm volatile("max.xorsign.abs.f16x2 %0, %1, %2;\n"
-                 : "=r"(reinterpret_cast<uint32_t&>(out))
-                 : "r"(reinterpret_cast<const uint32_t&>(x1)),
-                   "r"(reinterpret_cast<const uint32_t&>(x2)));
+  if constexpr (
+      (std::is_same<T, __nv_bfloat162>::value) ||
+      (std::is_same<T, __half2>::value)) {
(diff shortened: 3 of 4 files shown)


mlx/backend/cuda/quantized/nvfp4_quantize.cuh
```diff
@@ -0,0 +1,334 @@
...
+  uint8_t q2 = __nv_fp4_e2m1(scaled.z).__x;
+  uint8_t q3 = __nv_fp4_e2m1(scaled.w).__x;
+  out_fp4x4 = (static_cast<uint16_t>(q3) << 12) |
+      (static_cast<uint16_t>(q2) << 8) | (static_cast<uint16_t>(q1) << 4) |
+      static_cast<uint16_t>(q0);
+  return out_fp4x4;
+}
+
+#if (CUDART_VERSION >= 12080) && (__CUDA_ARCH__ >= 1000) && \
+    (__CUDA_ARCH_FAMILY_SPECIFIC__ >= 1000)
```
awni: Just wondering is it necessary to check both CUDA_ARCH and CUDA_ARCH_FAMILY_SPECIFIC? Is there any case that CUDA_ARCH > 1000 but the family specifc one isn't?

mlx/backend/cuda/quantized/quantized_utils.cuh
```diff
@@ -15,6 +15,26 @@ inline constexpr __device__ short get_bytes_per_pack() {
   return power_of_2_bits ? (wsize / 8) : (bits == 5 ? 5 : 3);
 }
 
+template <typename T>
+__device__ __forceinline__ void abs_max_x2(T& out, const T& x1, const T& x2) {
+  if constexpr (std::is_same<T, __nv_bfloat162>::value) {
+    asm volatile("max.xorsign.abs.bf16x2 %0, %1, %2;\n"
+                 : "=r"(reinterpret_cast<uint32_t&>(out))
+                 : "r"(reinterpret_cast<const uint32_t&>(x1)),
+                   "r"(reinterpret_cast<const uint32_t&>(x2)));
```
awni: Does `__habs2` and `__hmax2` work here? Could be a bit simpler than the inline assembly.

mlx/backend/cuda/quantized/mxfp8_quantize.cuh
```diff
@@ -0,0 +1,38 @@
...
+  scaled.z = static_cast<float>(input.z) * scale;
+  scaled.w = static_cast<float>(input.w) * scale;
+  uint8_t q0 = __nv_fp8_e4m3(scaled.x).__x;
+  uint8_t q1 = __nv_fp8_e4m3(scaled.y).__x;
+  uint8_t q2 = __nv_fp8_e4m3(scaled.z).__x;
+  uint8_t q3 = __nv_fp8_e4m3(scaled.w).__x;
+  out_fp8x4 = (static_cast<uint32_t>(q3) << 24) |
+      (static_cast<uint32_t>(q2) << 16) | (static_cast<uint32_t>(q1) << 8) |
+      static_cast<uint32_t>(q0);
+  return out_fp8x4;
```
awni: `__nv_fp8x4_e4m3` could work here.

mlx/backend/cuda/quantized/nvfp4_quantize.cuh
```diff
@@ -0,0 +1,334 @@
...
+
+__device__ __forceinline__ uint16_t
+scale_cvt_bf16x4_to_fp4x4_rn(const bf16x4 input_bf16x4, const float2 scale) {
+  uint16_t out_fp4x4 = 0;
```
awni, line 40: I'm kind of curious did you try not using assembly first? Is it much faster to use the inline assembly than to use the intrinsics here?

mlx/backend/cuda/quantized/nvfp4_quantize.cuh
```diff
@@ -0,0 +1,334 @@
...
+      "mul.f32x2 x01, x01, %2; \n\t" // scale first pair
+      "mov.b64 x23, {x2, x3}; \n\t"
+      "mul.f32x2 x23, x23, %2; \n\t" // scale second pair
+      "mov.b64 {x0, x1}, x01; \n\t"
+      "mov.b64 {x2, x3}, x23; \n\t"
+      "cvt.rn.satfinite.e2m1x2.f32 q0, x1, x0; \n\t" // convert to fp4x2 first
+                                                     // pair
+      "cvt.rn.satfinite.e2m1x2.f32 q1, x3, x2; \n\t" // convert to fp4x2 second
+                                                     // pair
+      "mov.b16 %0, {q0, q1}; \n\t" // pack to output
```
awni, line 70: Maybe as a future optimization:

But I wonder how important it is to put the scale in fp32 before we multiply by `w`. Since it's a quantized scale it's not super precise.. so maybe we could do the `w * s` in bf16 which would be a lot faster (avoids the conversions and should be double the flops).

Using the intrinsics for converting pairs of bf16 to fp4 should also be faster.

awni:
Very nice improvement! The log scale on time does not do it justice ;).

Regarding the small vs vectorized implementations. I'm not sure it's worth keeping the small based on your results. They look quite close and they are both very fast at that size. Wdyt? I can't quite tell from the plot since they are squished there.

mlx/backend/cuda/quantized/nvfp4_quantize.cuh
nastya236: The main reason why I used asm is because stochastic rounding is not supported through intrinsics (only through `cvt.rs.satfinite.e2m1x4.f32`. I thought that it might be useful to have it (for backward for instance), so I tried to make functions with / without stochastic routing similar. But I can do rn rounding fully without asm and it will be only 1-5% slower (honestly the difference is not big).

mlx/backend/cuda/quantized/nvfp4_quantize.cuh
nastya236: It is a very good point! In the [paper](https://arxiv.org/html/2509.25149v1), scale is converted to float, but it would be very interesting to try keeping it in bf16

mlx/backend/cuda/quantized/nvfp4_quantize.cuh
awni: Seems ok to leave it in that case!

mlx/backend/cuda/quantized/nvfp4_quantize.cuh
nastya236: Yea, it is actually incorrect here, it should be:
```cpp
#if (CUDART_VERSION >= 12080) && (__CUDA_ARCH__ >= 1000) && \
    defined(__CUDA_ARCH_SPECIFIC__)
```
because cvt from fp32 to e2m1 ptx instruction is supported only for sm_100a, sm_101a, sm_120a. On sm_100 if will fallback because `__CUDA_ARCH_SPECIFIC__` will be undefined. 
Thank you!

PR #2893, opened 2025-12-10
do not use simd neon intrinsics on x86

davidkoski:
- see https://github.com/ml-explore/mlx-swift/issues/309
- when MLX_USE_ACCELERATE and building for an x86_64 target do not try to use the accelerate_simd header which requires neon
- this is true for Release builds (Xcode) for mlx-swift

## Proposed changes

Conditional include of `accelerate_simd.h` on `x86_64`, especially for Release builds (Xcode) of mlx-swift.

## Checklist


awni (approved):
Awesome, thanks!

PR #2899, opened 2025-12-10
Fix grad in place updates

### LINKED ISSUE #2898: TestAutograd::test_grad_ids_pre_post passes spuriously
The test:

https://github.com/ml-explore/mlx/blob/937ce79660997a9b445c3f8daa56b92ada404012/python/tests/test_autograd.py#L778-L785

is supposed to check that passing a container to `mx.grad` doesn't mutate the items in the container. As [noted](https://docs.python.org/3/library/functions.html#id) in the docs for `id()`, two objects may have the same ids if they are never alive at the same time.

If I modify the test slightly:
 
 ```python
 import mlx.core as mx
 
 def fun(arrs):
     return arrs[0]
 
 arrs = [mx.array(1.0)]
 arr = arrs[0]
 mx.grad(fun)(arrs)
 assert arr is arrs[0]
 ```
 
That fails. I think that means the test just happens to pass because of re-used ids.

This means https://github.com/ml-explore/mlx/pull/1854 didn't fix the issue the test is testing for, despite the guesswork [here](https://github.com/ml-explore/mlx/pull/1854#discussion_r1951708326).

 _Originally posted by @ngoldbaum in [#2807](https://github.com/ml-explore/mlx/issues/2807#issuecomment-3639045984)_

awni:
Close #2898

### DIFF
--- a/python/src/transforms.cpp
+++ b/python/src/transforms.cpp
@@ -124,14 +124,13 @@ auto py_value_and_grad(
 
     // Collect the arrays
     std::vector<mx::array> arrays;
-    std::vector<nb::handle> array_objects;
-
+    std::vector<nb::object> array_objects;
     auto flatten_with_objects = [&arrays, &array_objects](
                                     auto tree, bool strict) {
       tree_visit(tree, [&](nb::handle obj) {
         if (nb::isinstance<mx::array>(obj)) {
           arrays.push_back(nb::cast<mx::array>(obj));
-          array_objects.push_back(obj);
+          array_objects.push_back(nb::borrow<nb::object>(obj));
         } else if (strict) {
           throw std::invalid_argument(
               "[tree_flatten] The argument should contain only arrays");
@@ -200,7 +199,7 @@ auto py_value_and_grad(
           tree_visit_update(tree, [&](nb::handle node) {
             auto replace_arr = nb::cast<mx::array>(node);
             if (replace_arr.id() == a[index].id()) {
-              return nb::borrow<nb::object>(array_objects[index++]);
+              return array_objects[index++];
             } else {
               index++;
               return nb::cast(replace_arr);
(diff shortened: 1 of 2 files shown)


ngoldbaum:
FWIW, the tests all pass on the free-threaded build on this branch, so merging this will unblock adding a 3.14t CI job.

angeloskath (approved):
👍

python/tests/test_autograd.py
```diff
@@ -780,9 +780,21 @@ def fun(arrs):
             return arrs[0]
 
         arrs = [mx.array(1.0)]
-        init_id = id(arrs[0])
+        arr = arrs[0]
         mx.grad(fun)(arrs)
-        self.assertEqual(init_id, id(arrs[0]))
+        self.assertEqual(id(arr), id(arrs[0]))
```
ngoldbaum, line 785:
```suggestion
        assert arr is arrs[0]
```

`arr is arrs[0]` is a little more idiomatic IMO

python/tests/test_autograd.py
```diff
@@ -780,9 +780,21 @@ def fun(arrs):
...
+            arrs[1] = sum(arrs)
+            return arrs[1]
+
+        arrs = [mx.array(1.0), mx.array(1.0), mx.array(1.0)]
+        a_0, a_1, a_2 = arrs
+
+        mx.grad(fun)(arrs)
+        self.assertEqual(id(a_0), id(arrs[0]))
+        self.assertNotEqual(id(a_1), id(arrs[1]))
+        self.assertEqual(id(a_2), id(arrs[2]))
```
ngoldbaum, line 797:
```suggestion
        assert a_0 is arrs[0]
        assert a_1 is arrs[1]
        assert a_2 is arrs[2]
```

PR #2901, opened 2025-12-12
[CUDA] Add host nodes to subgraph types for graph update

awni:
One more node type to add to the subgraph check.

This one only comes up so far with NCCL on multiple nodes.


angeloskath (approved):
Nice!

PR #2902, opened 2025-12-12
[CUDA] Implement gather_mm_rhs

zcbenz:
This PR implements `gather_mm` for the cases that can be transferred into a grouped GEMM.

The grouped GEMM code uses CUTLASS 2.x API which allows us to choose kernels that do not require large alignment. The performance is not good, running `benchmarks/python/gather_mm_bench.py` shows that it takes 7x time than equivalent matmul (for Metal kernel it takes 1.5x time).

There are a lot of things remaining to be done:

* Implement the cases that lhs indices are passed.
* Implement the cases when indices are broadcasted.
* Implement the cases when indices are not sorted.
* Tune the GEMM tile sizes.
* Enable tensor core for sm80 and later.
* Pad the group sizes so we can use much faster kernels.

But current work can serve as a baseline and a good foundation for progressive improvements.

### DIFF
--- a/mlx/backend/cuda/gemms/grouped_gemm_unaligned.cu
+++ b/mlx/backend/cuda/gemms/grouped_gemm_unaligned.cu
@@ -1,0 +1,288 @@
+// Copyright © 2025 Apple Inc.
+
+#include "mlx/backend/cuda/device.h"
+#include "mlx/backend/cuda/gemms/grouped_gemm.h"
+#include "mlx/backend/cuda/kernel_utils.cuh"
+#include "mlx/dtype_utils.h"
+
+#include <cooperative_groups.h>
+#include <cutlass/gemm/device/default_gemm_configuration.h>
+#include <cutlass/gemm/device/gemm_grouped.h>
+#include <cutlass/gemm/kernel/default_gemm_grouped.h>
+#include <fmt/format.h>
+#include <nvtx3/nvtx3.hpp>
+
+namespace mlx::core {
+
+using ProblemSize = cutlass::gemm::GemmCoord;
+
+namespace cu {
+
+namespace cg = cooperative_groups;
+
+template <int N_READS>
+__global__ void prepare_grouped_mm_data(
+    const uint32_t* indices,
+    size_t size,
+    int group_count,
+    int K,
+    int N,
+    int lda,
+    int ldb,
+    int item_size,
+    int8_t* a_start,
+    int8_t* b_start,
+    int8_t* out_start,
+    int a_batch_stride,
+    int b_batch_stride,
+    int out_batch_stride,
+    ProblemSize* problem_sizes,
+    int64_t* a_lds,
+    int64_t* b_lds,
+    int64_t* out_lds,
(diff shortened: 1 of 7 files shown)


mlx/backend/cuda/matmul.cpp
```diff
@@ -254,4 +311,40 @@ void AddMM::eval_gpu(const std::vector<array>& inputs, array& out) {
...
+  auto& a = inputs[0];
+  auto& b = inputs[1];
+  auto& lhs_indices = inputs[2];
+  auto& rhs_indices = inputs[3];
+
+  // Return 0s if either input is empty.
+  if (a.size() == 0 || b.size() == 0) {
+    array zero(0, a.dtype());
+    encoder.add_temporary(zero);
+    fill_gpu(zero, out, s);
```
awni, line 339: Just a note, not necessary for this PR but we should probably do these with `cudaMemsetAsync`

awni: It would need to go in the graph.. but I think that should be fairly straight-forward.

mlx/backend/cuda/matmul.cpp
```diff
@@ -103,6 +136,30 @@ void gemm_and_bias(
...
 
+void gather_mm_rhs(
+    const array& a_,
+    const array& b_,
+    const array& indices_,
+    array& out,
+    cu::CommandEncoder& encoder,
+    Stream s) {
+  if (a_.size() / a_.shape(-2) / a_.shape(-1) != indices_.size()) {
+    throw std::runtime_error("[gather_mm] Boardcasting lhs is not supported.");
```
awni:
```suggestion
    throw std::runtime_error("[gather_mm] Broadcasting lhs is not supported.");
```

mlx/backend/cuda/matmul.cpp
zcbenz: Cuda graph has a memset node we can use, that would be much better than running a kernel.

mlx/backend/cuda/gemms/grouped_gemm_unaligned.cu
```diff
@@ -0,0 +1,266 @@
...
+      DefaultGemmConfiguration<OpClass, Arch, T, T, T, ElementAccumulator>;
+  using EpilogueOutputOp = typename GemmConfiguration::EpilogueOutputOp;
+
+  using GemmKernel = typename cutlass::gemm::kernel::DefaultGemmGrouped<
+      T,
+      cutlass::layout::RowMajor,
+      cutlass::ComplexTransform::kNone,
+      kAlignment,
+      T,
+      cutlass::layout::RowMajor,
```
angeloskath: This should depend on the passed in matrix. Basically when `b_transposed` in `matmul.cpp` is true this should be `ColMajor`.

zcbenz: Thanks for noticing this! I have fixed it and added tests.

angeloskath (approved):
Nice! Thanks.

PR #2903, opened 2025-12-12
Fix attention for large sizes

### LINKED ISSUE #2894: mx.fast.scaled_dot_product_attention produces incorrect results with boolean masks > 2^31 elements
## Bug Description

`mx.fast.scaled_dot_product_attention` produces incorrect results when using a boolean mask with more than 2^31 elements (i.e., when `seq_len > 46,341`).

This appears to be a 32-bit integer overflow in the mask indexing logic.

## Minimal Reproducer

```python
# /// script
# dependencies = ["mlx"]
# ///
"""
Minimal reproducer for SDPA boolean mask overflow bug in MLX.

To run: uv run sdpa_mask_overflow_bug.py
"""

import mlx.core as mx


def attention_sdpa(q, k, v, mask):
    """Attention using mx.fast.scaled_dot_product_attention with boolean mask."""
    return mx.fast.scaled_dot_product_attention(
        q[:, None], k[:, None], v[:, None], scale=1.0, mask=mask
    ).squeeze(1)


def attention_naive(q, k, v, mask):
    """Naive attention implementation for reference."""
    scores = q @ k.swapaxes(-1, -2)
    scores = mx.where(mask, scores, mx.array(float("-inf")))
    attn = mx.softmax(scores, axis=-1)
    return attn @ v


def test_sdpa_mask(seq_len, sparsity=0.125):
    """Test SDPA with a sparse boolean mask of given sequence length."""
    batch, d_model = 1, 64

    mx.random.seed(42)
    q = mx.random.normal(shape=(batch, seq_len, d_model))
    k = mx.random.normal(shape=(batch, seq_len, d_model))
    v = mx.random.normal(shape=(batch, seq_len, d_model))
    mask_prob = mx.random.uniform(shape=(batch, seq_len, seq_len))
    mask = mask_prob < sparsity
    mx.eval(q, k, v, mask)

    out_sdpa = attention_sdpa(q, k, v, mask)
    out_naive = attention_naive(q, k, v, mask)
    mx.eval(out_sdpa, out_naive)

    return (out_sdpa - out_naive).abs().max().item()


# Test around the 2^31 boundary (sqrt(2^31) ≈ 46,341)
for seq_len in [32768, 46080, 47104, 49152, 65536]:
    diff = test_sdpa_mask(seq_len)
    mask_size = seq_len * seq_len
    status = "OK" if diff < 1e-4 else "BUG"
    print(f"seq_len={seq_len:>6}, mask_size={mask_size:>13,}, diff={diff:.2e}, {status}")
```

## Output

```
seq_len= 32768, mask_size=1,073,741,824, diff=2.97e-05, OK
seq_len= 46080, mask_size=2,123,366,400, diff=2.80e-05, OK
seq_len= 47104, mask_size=2,218,786,816, diff=4.98e+00, BUG
seq_len= 49152, mask_size=2,415,919,104, diff=5.59e+00, BUG
seq_len= 65536, mask_size=4,294,967,296, diff=5.51e+00, BUG
```

## Expected Behavior

SDPA should produce results matching the naive attention implementation for all mask sizes.

## Actual Behavior

When mask size exceeds 2^31 elements (~2.1 billion), SDPA output diverges significantly from the correct result (max diff > 1.0 instead of < 1e-4).

## Analysis

The threshold of 2^31 strongly suggests a 32-bit signed integer overflow somewhere in the mask indexing logic. The bug manifests when:
- `seq_len > 46,341` (since `46341^2 ≈ 2^31`)
- `mask_size > 2,147,483,648`

## Environment

- MLX version: latest (tested Dec 2024)
- Platform: macOS (Apple Silicon)

awni:
Close #2894

### DIFF
--- a/mlx/backend/metal/kernels/steel/attn/kernels/steel_attention.h
+++ b/mlx/backend/metal/kernels/steel/attn/kernels/steel_attention.h
@@ -347,7 +347,7 @@ template <
           MMAFrag_mask_t::load_safe(
               mfrag,
               mask,
-              int(mask_params->M_strides[2]),
+              int64_t(mask_params->M_strides[2]),
               Int<1>{},
               params->qL,
               params->kL,
--- a/mlx/backend/metal/kernels/steel/attn/kernels/steel_attention_nax.h
+++ b/mlx/backend/metal/kernels/steel/attn/kernels/steel_attention_nax.h
@@ -346,7 +346,7 @@ template <
           MSubTile mfrag;
           mfrag.load_safe(
               mask,
-              int(mask_params->M_strides[2]),
+              int64_t(mask_params->M_strides[2]),
               Int<1>{},
               params->qL,
               params->kL,
--- a/mlx/backend/metal/kernels/steel/attn/mma.h
+++ b/mlx/backend/metal/kernels/steel/attn/mma.h
@@ -105,17 +105,20 @@ struct BaseMMAFrag<T, 8, 8> {
       LimY lim_y,
       OffX off_x = Int<0>{},
       OffY off_y = Int<0>{}) {
+    src += off_x * str_x + off_y * str_y;
     STEEL_PRAGMA_UNROLL
     for (short i = 0; i < kElemRows; i++) {
       STEEL_PRAGMA_UNROLL
       for (short j = 0; j < kElemCols; j++) {
         if ((off_x + i) < lim_x && (off_y + j) < lim_y) {
-          dst[i * kElemCols + j] =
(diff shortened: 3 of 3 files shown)


awni:
The change in the mma loader is just to speed it up so we don't lose perf using int64 stride for the mask.

awni:
> I presume you ran some tests to check if there is any regression...

Yes I ran a benchmark for just SDPA and a model prefill benchmark and there is no change.

In fact just changing to int64 without changing the loader was a consistent 1-2% slowdown on M2 Ultra (so not that bad). Changing the loader brought it back.

PR #2904, opened 2025-12-12
Metal logging

CC-Yeh:
## Proposed changes

- Add `-fenable-metal-logging` and `setEnableLogging(true)`(for JIT) to enable `os_log` for debug build.
- Add two examples using `os_log`.

## Checklist

### DIFF
--- a/mlx/backend/metal/kernels/binary_ops.h
+++ b/mlx/backend/metal/kernels/binary_ops.h
@@ -5,8 +5,6 @@
 #include <metal_integer>
 #include <metal_math>
 
-#include "mlx/backend/metal/kernels/logging.h"
-
 constant mlx::os_log logger("mlx", "binary_ops");
 
 struct Add {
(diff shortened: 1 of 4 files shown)


awni (approved):
Very nice, thank you!

mlx/backend/metal/kernels/binary_ops.h
```diff
@@ -3,8 +3,11 @@
 #pragma once
 
 #include <metal_integer>
+#include <metal_logging>
 #include <metal_math>
 
+constant os_log logger("mlx", "binary_ops");
```
awni: The test failed building for macOS 14 cause this API is not available. It might make sense to make a stub for lower version so we don't have to use ifdef everywhere.

CC-Yeh: Thanks — I added a no-op os_log stub for older macOS/SDKs

Issue #2842, opened 2025-11-26
Search CUDA toolkit from python packages


zcbenz:
Currently mlx-cuda requires a system install of CUDA toolkit:

https://github.com/ml-explore/mlx/blob/704fd1ae284036908ede151062452a0f190e8221/mlx/backend/cuda/jit_module.cpp#L30-L51

It would be nice to be able to search from installed python packages as well so users can do:

```
uv run --with 'mlx[cuda]' --with 'cuda-toolkit[all]==12.9' python test_mlx.py
```

leofang:
Hello from CUDA Python! 👋  The `cuda-pathfinder` package was designed for precisely this use case, and it has been adopted by CuPy, PyTorch, and many other Python CUDA packages!
https://nvidia.github.io/cuda-python/cuda-pathfinder/latest/generated/cuda.pathfinder.load_nvidia_dynamic_lib.html
Any issues can be reported to https://github.com/NVIDIA/cuda-python 🙂 Let us know how we can help you.

zcbenz:
@leofang This is super helpful, thanks a lot!

PR #2906, opened 2025-12-13
Use CUDA runtime headers from local python package

zcbenz:
Close https://github.com/ml-explore/mlx/issues/2842.

Search `../../nvidia/cuda_runtime/include` for CUDA runtime headers, which is available when the `cuda-toolkit` python package is installed together with `mlx`. This follows the behavior of where CUDA libraries are searched:

https://github.com/ml-explore/mlx/blob/bedefed784115eb233a22c0e192b957de92b2a85/python/scripts/repair_cuda.sh#L20-L22

Also slightly reorganizes the cmake file — especially put the CCCL related configurations together.


awni (approved):
Very nice, thanks for fixing that!

PR #2910, opened 2025-12-13
DOC : Add compile state example

Satyam12singh:
## Proposed changes

This PR addresses an inconsistency in the "Pure Functions" subsection of the `mlx.compile` documentation.

The guide currently mentions two methods for capturing mutable external state, but only provides a code example for the second method.

**Problem:** The text states, "The first option is to simply pass `state` as input to the function," but this method was not demonstrated with code.

**Solution:** This change inserts the missing code block, which explicitly shows the technique of passing `state` as an argument. This provides a clear contrast between the two options and improves the overall clarity and completeness of the documentation section.

## Checklist

### DIFF
--- a/docs/src/usage/compile.rst
+++ b/docs/src/usage/compile.rst
@@ -258,18 +258,23 @@ constants. For example:
 In order to have the change of state reflected in the outputs of ``fun`` you
 again have two options. The first option is to simply pass ``state`` as input
 to the function.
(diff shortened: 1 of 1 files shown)


Satyam12singh:
Thanks for the approval, @awni ! Is there anything else you need from my side before this is ready to be merged?

PR #2917, opened 2025-12-16
Faster copy for col contig to row contig

awni:
Adds a specialized copy from col-contiguous to row-contiguous arrays.

Benchmark on an H100:

Pre: 818 GB/s
Post: 2191.62 GB/s

The benchmark is below:

```python
import time
import mlx.core as mx

t = mx.bfloat16
D = 4096
x = mx.random.normal(shape=(D, D)).astype(t)

def fun(x):
    for _ in range(20):
        x = mx.contiguous(x.T)
    mx.eval(x)

for _ in range(20):
    fun(x)

tic = time.time()
for _ in range(20):
    fun(x)
toc = time.time()

gb = 20 * 20 * x.nbytes * 2 / 1e9
s = toc - tic
gbps = gb / s
print(f"{gbps=:.3f}")
```

### DIFF
--- a/mlx/backend/cuda/copy/copy_general_input.cu
+++ b/mlx/backend/cuda/copy/copy_general_input.cu
@@ -77,7 +77,8 @@ __global__ void copy_g(
 template <typename In, typename Out, int N_READS>
 __global__ void
 copy_col_row(const In* in, Out* out, int64_t rows, int64_t cols) {
-  __shared__ Out tile[N_READS * TILE_SIZE][N_READS * TILE_SIZE + 1];
+  __shared__ Out
+      tile[N_READS * TILE_SIZE][N_READS * TILE_SIZE + 4 / sizeof(Out)];
 
   auto block = cg::this_thread_block();
   auto grid = cg::this_grid();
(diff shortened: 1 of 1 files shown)


mlx/backend/cuda/copy/copy_general_input.cu
```diff
@@ -73,6 +74,52 @@ __global__ void copy_g(
   store_vector(out + shape_x * index_rest, index_x, out_vec, shape_x);
 }
 
+template <typename In, typename Out, int N_READS>
+__global__ void
+copy_col_row(const In* in, Out* out, int64_t rows, int64_t cols) {
+  __shared__ Out tile[N_READS * TILE_SIZE][N_READS * TILE_SIZE + 1];
```
angeloskath: It already works great afaict but this should probably be 

```
__shared__ Out tile[N_READS * TILE_SIZE][N_READS * TILE_SIZE + 4/sizeof(Out)];
```

mlx/backend/cuda/copy/copy_general_input.cu
```diff
@@ -86,15 +133,37 @@ void copy_general_input(
...
     dispatch_all_types(out.dtype(), [&](auto out_type_tag) {
+      using InType = cuda_type_t<MLX_GET_TYPE(in_type_tag)>;
+      using OutType = cuda_type_t<MLX_GET_TYPE(out_type_tag)>;
+      const InType* in_ptr = gpu_ptr<InType>(in) + offset_in;
+      OutType* out_ptr = gpu_ptr<OutType>(out) + offset_out;
+      int ndim = shape.size();
+
+      // Column contiguous to row contiguous specialization
+      if (ndim == 2 && strides_in[0] == 1 && strides_in[1] > 1) {
+        constexpr int work_per_thread = 16 / sizeof(OutType);
```
angeloskath: I think you either have to limit it to sth like

```
constexpr int work_per_thread = sizeof(OutType) > 1 ? 16 / sizeof(OutType) : 8;
```

or request dynamic shared memory cause for uint8 it will go more than 48KB which will fail.

angeloskath: Which is also the test error now that I looked :-D

angeloskath (approved):
Looks great! Left a couple of comments.

PR #2920, opened 2025-12-17
View-only layout ops mover

CC-Yeh:
## Proposed changes

View-only layout ops (e.g. `expand_dims`, `squeeze`) can sit between elementwise ops and reduce compile’s fusion opportunities. 

This PR adds a small “layout mover” step in compile_simplify that (conservatively) rewrites view(op(x)) -> op(view(x)), and runs it to a bounded fixpoint so the view can move up multiple elementwise layers.

This is an early draft to get feedback on whether MLX wants this optimization before I refine/extend it (e.g. consider `reshape`, `flatten` ...).

Before:
fused_old

After:
fused_new

## Checklist

### DIFF
--- a/mlx/compile.cpp
+++ b/mlx/compile.cpp
@@ -60,6 +60,14 @@ bool is_ternary(const Primitive& p) {
   return typeid(p) == typeid(Select);
 }
 
+bool is_expand_dims(const Primitive& p) {
+  return typeid(p) == typeid(ExpandDims);
+}
+
+bool is_squeeze(const Primitive& p) {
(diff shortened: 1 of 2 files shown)


CC-Yeh:
@awni @angeloskath

What do you think about adding this to `compile`?
It's a bit similar to [XLA ReshapeMover](https://openxla.org/xla/hlo_passes#reshape_mover).

zcbenz:
Sorry for the very late response, I don't think we would be able to review it, reviewing PRs like this basically takes as much time as writing it ourselves and at the moment this feature is not on our roadmap.

PR #2926, opened 2025-12-18
fix cuda release part 2

awni:
As title

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -198,7 +198,7 @@ jobs:
     steps:
       - uses: actions/download-artifact@v7
         with:
-          name: mlx-cuda
+          name: mlx-cuda-*
           path: dist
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
zcbenz: Should this be changed too?

awni: Good catch

.github/workflows/release.yml
```diff
@@ -198,7 +198,7 @@ jobs:
     steps:
       - uses: actions/download-artifact@v7
         with:
-          name: mlx-cuda
+          name: mlx-cuda-*
```
zcbenz, line 201: I think it should be `pattern:`

zcbenz: And also `merge-multiple: true`

PR #2928, opened 2025-12-18
Allow dry run for PyPI release workflow

zcbenz:
Add an option to build the releases without publishing them, also remove the `setup` step which does nothing now.

Screenshot 2025-12-18 at 20 21 46


zcbenz:
Sample dry run: https://github.com/ml-explore/mlx/actions/runs/20332298791

PR #2931, opened 2025-12-18
QQ linear

nastya236:
`QQLinear` layer

### DIFF
--- a/python/mlx/nn/layers/quantized.py
+++ b/python/mlx/nn/layers/quantized.py
@@ -344,30 +344,37 @@ class QQLinear(Module):
             f"group_size={self.group_size}, bits={self.bits}, mode={self.mode}"
         )
 
-    def _update_mode(self):
+    def quantize(self):
+        if not self._quantized:
+            self.weight, self.scales = mx.quantize(
+                self.weight,
+                self.group_size,
+                self.bits,
+                mode=self.mode,
+            )
+            self._quantized = True
+
+    def dequantize(self):
+        if self._quantized:
+            self.weight = mx.dequantize(
+                self.weight,
+                scales=self.scales,
+                group_size=self.group_size,
+                bits=self.bits,
+                mode=self.mode,
+            )
+            self.__delattr__("scales")
+            self._quantized = False
+
+    def _set_training_mode(self, mode: bool):
+        super()._set_training_mode(mode)
+
         if self._training:
-            if self._quantized:
-                self.weight = mx.dequantize(
-                    self.weight,
-                    scales=self.scales,
-                    group_size=self.group_size,
-                    bits=self.bits,
-                    mode=self.mode,
-                )
-                self.__delattr__("scales")
-                self._quantized = False
+            self.dequantize()
         else:
-            if not self._quantized:
-                self.weight, self.scales = mx.quantize(
-                    self.weight,
-                    self.group_size,
-                    self.bits,
-                    mode=self.mode,
-                )
-                self._quantized = True
+            self.quantize()
 
     def __call__(self, x):
-        self._update_mode()
         x = mx.qqmm(
             x,
             self["weight"],
@@ -392,6 +399,6 @@ class QQLinear(Module):
             raise NotImplementedError("QQLinear does not support bias yet.")
         ql = cls(input_dims, output_dims, group_size, bits, mode=mode)
         ql.weight = linear_layer.weight
-        ql._training = linear_layer._training
+        ql.train(linear_layer.training)
 
         return ql
--- a/python/mlx/nn/layers/base.py
+++ b/python/mlx/nn/layers/base.py
@@ -559,6 +559,9 @@ class Module(dict):
             _unfreeze_impl("", self)
         return self
 
+    def _set_training_mode(self, mode: bool) -> None:
+        self._training = mode
+
     def train(self, mode: bool = True) -> Module:
         """Set the model in or out of training mode.
 
@@ -573,10 +576,8 @@ class Module(dict):
             The module instance after updating the training mode.
         """
 
-        def _set_train(_, m):
-            m._training = mode
+        self.apply_to_modules(lambda _, m: m._set_training_mode(mode))
 
-        self.apply_to_modules(_set_train)
         return self
 
     def eval(self) -> Module:


python/mlx/nn/layers/quantized.py
```diff
@@ -268,3 +268,93 @@ def from_linear(
...
+
+    Note: This layer does not support a bias term yet.
+
+    Args:
+        input_dims (int): The dimensionality of the input features.
+        output_dims (int): The dimensionality of the output features.
+        group_size (int, optional): The group size to use for the quantized weight.
+            See :func:`~mlx.core.quantize`. Default: ``16``.
+        bits (int, optional): The bit width to use for the quantized weight.
+            See :func:`~mlx.core.quantize`. Default: ``4``.
```
awni: I think it makes sense to default these to `None` so that the right default for the given `mode` is chosen.

python/mlx/nn/layers/quantized.py
```diff
@@ -268,3 +268,93 @@ def from_linear(
             ql.bias = linear_layer.bias
 
         return ql
+
+
+class QQLinear(Module):
```
awni, line 282: Does it make sense to hijack `Module::train()` and `Module::eval()` to update the weights between quantize and not-quantize?

I think at the very least it would be nice to have a method which turns it into the inference version since I expect it would be a common use case.

CC @angeloskath might have some ideas.

python/mlx/nn/layers/quantized.py
awni: (I realize that it's like this for the other Quantize modules and I think I just forgot to change those when I changed the op defaults, but we should change them as well)

python/mlx/nn/layers/quantized.py
```diff
@@ -268,3 +268,93 @@ def from_linear(
...
+            mode=self.mode,
+        )
+        return x
+
+    @classmethod
+    def from_linear(
+        cls,
+        linear_layer: Module,
+        group_size: int = 16,
+        bits: int = 4,
```
awni: Same comment regarding defaults.

python/mlx/nn/layers/quantized.py
awni: Btw I changed the defaults for the other layers in #2946 .. I think you can follow the same pattern for this layer as well

python/mlx/nn/layers/quantized.py
nastya236: I added `QQLinear.train()` and `QQLinear.eval()`. Let me know if you had something different in mind.
For inference, I think we should support two cases: (1) weights are loaded already quantized (with loading handled in mlx-lm, or a similar for cuda), or (2) weights are bf16/fp32, in which case we call .eval() to quantize them.

python/mlx/nn/layers/quantized.py
```diff
@@ -277,3 +277,120 @@ def from_linear(
...
+    Args:
+        input_dims (int): The dimensionality of the input features.
+        output_dims (int): The dimensionality of the output features.
+        group_size (Optional[int]): The group size to use for the quantized weight.
+            See :func:`~mlx.core.quantize`. Default: ``None``.
+        bits (Optional[int]): The bit width to use for the quantized weight.
+            See :func:`~mlx.core.quantize`. Default: ``None``.
+        mode (Optional[str]): The quantization method to use (see
+            :func:`mlx.core.quantize`). Currently, only ``"nvfp4"`` and ``"mxfp8"``
+            are supported. Default: ``nvfp4``.
```
awni:
```suggestion
            are supported. Default: ``"nvfp4"``.
```

python/mlx/nn/layers/quantized.py
```diff
@@ -277,3 +277,120 @@ def from_linear(
...
+    def from_linear(
+        cls,
+        linear_layer: Module,
+        group_size: int = None,
+        bits: int = None,
+        mode: str = "nvfp4",
+        train: bool = True,
+    ):
+        """Create a :obj:`QQLinear` layer from a :obj:`Linear` layer."""
+        output_dims, input_dims = linear_layer.weight.shape  # (N,K)
```
awni, line 397: Probably worth checking the linear layer does not have a bias here.

python/mlx/nn/layers/quantized.py
```diff
@@ -277,3 +277,120 @@ def from_linear(
...
+        return x
+
+    @classmethod
+    def from_linear(
+        cls,
+        linear_layer: Module,
+        group_size: int = None,
+        bits: int = None,
+        mode: str = "nvfp4",
+        train: bool = True,
```
awni: Maybe that should be inherited from the `linear_layer` rather than an argument? Keeps the API a bit more consistent with the other quantize layers.

python/mlx/nn/layers/quantized.py
```diff
@@ -277,3 +277,120 @@ def from_linear(
...
+    def eval(self):
+        if self.weight.dtype != mx.uint32:
+            self.weight, self.scales = mx.quantize(
+                self.weight,
+                self.group_size,
+                self.bits,
+                mode=self.mode,
+            )
+
+    def train(self):
```
awni: Can you make that API consistent with [`Module::train`](https://github.com/ml-explore/mlx/blob/main/python/mlx/nn/layers/base.py#L562)?

awni: I think if you do that we can get rid of the `eval` override above and just use the base class eval.

python/mlx/nn/layers/quantized.py
```diff
@@ -277,3 +277,120 @@ def from_linear(
...
+        weights are stored in higher precision and are quantized on
+        the fly during computation.
+
+    Compared to the :class:`mlx.nn.QuantizedLinear` layer, this layer
+    quantizes the input as well and includes weights in gradient computations.
+
+    :obj:`QQLinear` also provides:
+     -  the class method :meth:`from_linear` to convert :class:`mlx.nn.Linear`
+     layers to :obj:`QQLinear` layers. If the layer is created in evaluation mode,
+     the weights will be quantized.
```
awni: I think that indentation needs to be fixed or it will look a bit funny / give an error.

python/mlx/nn/layers/quantized.py
```diff
@@ -277,3 +277,120 @@ def from_linear(
...
         return ql
+
+
+class QQLinear(Module):
+    """Quantizes the input and applies an affine transformation using quantized weights.
+
+    Two use cases are supported:
+
+    1) **Inference / frozen weights**: weights are stored in quantized form together with
+       their scales (``self.weight`` is quantized and ``self.scales`` is provided).
```
awni:
```suggestion
    1) **Eval**:  The weights are frozen and stored in quantized form together with
       their scales (``self.weight`` is quantized and ``self.scales`` is provided).
```

python/mlx/nn/layers/quantized.py
```diff
@@ -277,3 +277,120 @@ def from_linear(
...
+
+    2) **Training / weights are included in gradient computation**:
+        weights are stored in higher precision and are quantized on
+        the fly during computation.
```
awni:
```suggestion
    2) **Train**: The weights are stored in higher precision and are quantized on
         the fly during computation so that gradients with respect to the weights
         can be computed.
```

python/mlx/nn/layers/quantized.py
```diff
@@ -277,3 +277,120 @@ def from_linear(
...
+
```
awni, line 294: Maybe a line about how to toggle between the two use cases:

```
To switch between the two cases, use ``layer.eval()`` and ``layer.train()`` respectively.
```

python/mlx/nn/layers/quantized.py
nastya236: Of course, this is a very good point. Now the weights will be quantized on the first `qq_linear.__call__()` after calling `Module::eval()`, and likewise dequantized on the first `qq_linear.__call__()`after calling `Module::train()`. This seems to be the only way to keep it consistent with the current API without changing `Module::train()`..

awni: I see why you had to do it this way but I'm not crazy about how it works... I'm wondering if there is a better way to do it.

Basically the behavior that would be good to have is if we do:

```
qq_module.eval()
qq_module.parameters() # should give me the quantized params
qq_module.load_weights(quantized_weights) # should be able to load the quantized params
```

I think that should work and right now it won't.

I think we have some other options:
1. Break away from the `train`/`eval` API and have something like `QQLinear.quantize`/`QQLinear.dequantize` which either quantizes/dequantize the module in-place (or maybe returns a copy that is quantized/dequantized)
2. Change the base class `Module` to make it easier to override `train` (e..g call the submodules `train` as well as setting the local module's `_training`.

Wdyt?

nastya236: Fully agree with the desired behavior you described. Between the options, I prefer (2). I would expect `model.train()` / `model.eval()` to recursively propagate mode changes through the tree and adding a separate quantize() / dequantize() API would create a parallel mode system that would likely need to be aligned with `train` / `eval` as well..

awni (approved):
Awesome, thanks!!

PR #2932, opened 2025-12-19
Set rpath with cmake for CUDA build

zcbenz:
Set RPATH with cmake directly when building python bindings instead of editing the wheel file with a script.

The RPATH is slightly changed though, before:
**$ORIGIN/../../mlx_cuda_13.libs**:$ORIGIN/../../nvidia/cublas/lib:$ORIGIN/../../nvidia/cuda_nvrtc/lib:$ORIGIN/../../nvidia/cudnn/lib:$ORIGIN/../../nvidia/nccl/lib

after:
$ORIGIN/../../nvidia/cublas/lib:$ORIGIN/../../nvidia/cuda_nvrtc/lib:$ORIGIN/../../nvidia/cudnn/lib:$ORIGIN/../../nvidia/nccl/lib:**$ORIGIN/../../mlx_cuda_13.libs**

which I think should be fine.

The `repair_linux.sh` can also be removed similarly, which I will start another PR.


mlx/backend/cuda/CMakeLists.txt
```diff
@@ -156,6 +156,18 @@ message(STATUS "CUDA architectures: ${MLX_CUDA_ARCHITECTURES}")
...
+  set_property(
+    TARGET mlx
+    APPEND
+    PROPERTY INSTALL_RPATH
+             # The paths here should match the install_requires in setup.py.
+             "$ORIGIN/../../nvidia/cublas/lib"
+             "$ORIGIN/../../nvidia/cuda_nvrtc/lib"
+             "$ORIGIN/../../nvidia/cudnn/lib"
+             "$ORIGIN/../../nvidia/nccl/lib")
+endif()
```
awni, line 169: If I'm building python locally and using the system installed libraries that will still work right?

zcbenz: Yeah it still works, setting RPATH does not remove other lib search paths.

Issue #2935, opened 2025-12-19
[BUG] A couple issues with mlx.launch


awni:
Run a basic script locally with:

```
mlx.launch -n 2 test.py
```

it prints the temp files that the process IDs are written to.

```
/var/folders/55/wx45hr350m94txzyxrqkfc900002xx/T/tmp.J8VEkhX0Gy
/var/folders/55/wx45hr350m94txzyxrqkfc900002xx/T/tmp.cvG5d4OtF2
```

Also when trying to debug with pdb inside a script, the enter command doesn't work. It prints `^M` instead of sending the command, so one cannot interact with the terminal.

angeloskath:
Huh the temp files I have noticed when running locally I didn't care too much to fix them cause they don't really cause an issue afaict but I should.

The 2nd part though is weird. I have tested it many times with pdb and I don't have this issue. Is this on Mac ?

awni:
The 2nd part is working now. Must have been an issue in my terminal.. not sure. Sorry for the fuzz.

PR #2936, opened 2025-12-19
Fix pid in local launch

angeloskath:
Fixes part of #2935 . Will update with the 2nd part when more info becomes available.


awni (approved):
Ty!!

PR #2937, opened 2025-12-19
ci: add macOS 26 target

madrob:
## Proposed changes

Add macOS 26 target to all GitHub build targets

## Checklist

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -61,7 +61,7 @@ jobs:
           use-ccache: false
       - uses: ./.github/actions/build-linux-release
         with:
-          build-backend: ${{ matrix.python-version == '3.10' }}
+          build-backend: ${{ matrix.python_version == '3.10' }}
           arch: ${{ matrix.arch }}
       - name: Upload MLX artifacts
         uses: actions/upload-artifact@v6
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -115,6 +115,11 @@ jobs:
         with:
           macos-target: 15.0
           build-backend: ${{ matrix.python-version == '3.10' }}
+      - name: Build macOS 26 package
+        uses: ./.github/actions/build-macos-release
+        with:
+          macos-target: 26.0
+          build-backend: ${{ matrix.python-version == '3.10' }}
```
awni: We will do the 26.0 release manually for now until we can get on macOS 26.2 (which is needed to build neural accelerator and RDMA support).

So please remove that part.

madrob: To explicitly confirm, we'll still leave it for both the nightly and the PR builds, yes?

awni: Yes I think it's great to start testing 26.0 👍 

For the nightly it's fine.. the only downside is burning compute.. but it's also not necessary so we can remove it and add it back once we do the 26.0 release in CI.

PR #2940, opened 2025-12-20
refactor: use perf_counter for accurate benchmarking

Satyam12singh:
## Proposed changes

This PR replaces `time.time()` with `time.perf_counter()` in the logistic regression example. 

**Problem:** `time.time()` is non-monotonic and susceptible to system clock adjustments, which can lead to inaccurate throughput (it/s) measurements.
**Solution:** `time.perf_counter()` provides a high-resolution, monotonic clock ideal for benchmarking.

## Checklist

### DIFF
--- a/examples/python/logistic_regression.py
+++ b/examples/python/logistic_regression.py
@@ -1,6 +1,6 @@
 # Copyright © 2023 Apple Inc.
 
-from time import perf_counter
+import time
 
 import mlx.core as mx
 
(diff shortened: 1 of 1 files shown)


examples/python/logistic_regression.py
```diff
@@ -30,13 +30,13 @@ def loss_fn(w):
 
 grad_fn = mx.grad(loss_fn)
 
-tic = time.time()
+tic = perf_counter()
```
awni: Nit: use `time.perf_counter` instead of importing it into the global namespace. Follows the style of the rest of the examples / codebase.

Satyam12singh: Fixed, thanks!

Satyam12singh:
Hi @awni , I believe this is ready to merge. Please let me know if there’s anything else to address.

PR #2948, opened 2025-12-22
Fix float64 size in data_types.rst

pdevine:
## Proposed changes

Corrected the size of float64 from 4 to 8 bytes.

## Checklist


awni (approved):
Good catch !

PR #2949, opened 2025-12-23
Fetch nanobind with cmake

zcbenz:
This is mostly a build isolation v.s. `ccache` thing: with build isolation the python packages (i.e. `nanobind`) and cmake build dir would be put under different temporary dirs, and `ccache` would not able to cache files using `nanobind` because the they are including headers from a temporary dir. That's why we use `--no-build-isolation` to fix cache misses, and also why we always have cache misses when using uv because it always builds things in temporary dir.

There are also other benefits:
* We don't have to hard-code `nanobind==2.10.2` in a lot of places.
* Stubs generation would be part of cmake build rather than a separate command.
* Simpler CI setup after we remove `--no-build-isolation`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -41,6 +41,7 @@ option(MLX_ENABLE_X64_MAC "Enable building for x64 macOS" OFF)
 option(MLX_BUILD_GGUF "Include support for GGUF format" ON)
 option(MLX_BUILD_SAFETENSORS "Include support for safetensors format" ON)
 option(MLX_BUILD_BLAS_FROM_SOURCE "Build OpenBLAS from source code" OFF)
+option(MLX_BUILD_PYTHON_STUBS "Build stub files for python bindings" ON)
 option(MLX_METAL_JIT "Use JIT compilation for Metal kernels" OFF)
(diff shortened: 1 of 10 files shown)


CMakeLists.txt
```diff
@@ -41,6 +41,7 @@ option(MLX_ENABLE_X64_MAC "Enable building for x64 macOS" OFF)
 option(MLX_BUILD_GGUF "Include support for GGUF format" ON)
 option(MLX_BUILD_SAFETENSORS "Include support for safetensors format" ON)
 option(MLX_BUILD_BLAS_FROM_SOURCE "Build OpenBLAS from source code" OFF)
+option(MLX_BUILD_STUBS "Build stub files for python bindings" ON)
```
awni: I might rename that `MLX_BUILD_PYTHON_STUBS` just to make it a little more clear it's specific to Python.

Also we could just always build the stubs.. since it defaults to on seems unlikely anyone will ever turn it off.

awni (approved):
Looks like a great change. I left a minor comment. Feel free to merge when ready!

CMakeLists.txt
zcbenz: Building stubs requires the binary to be runnable, so for arm64 build on CI we have to disable stubs.

PR #2964, opened 2026-01-03
Fix scheduler exception propagation to Python

### LINKED ISSUE #2888: [BUG] Singular matrix
**Describe the bug**
Kernel gets killed when solving linear problem with singular matrix.

**To Reproduce**
```python
import mlx.core as mx
mx.linalg.solve(mx.ones((2, 2)), mx.ones((2,)), stream=mx.cpu)
# libc++abi: terminating due to uncaught exception of type std::runtime_error: [LUF::eval_cpu] sgetrf_ failed with code 2 because matrix is singular
# Process finished with exit code 134 (interrupted by signal 6:SIGABRT)
```

**Expected behavior**
Have an exception maybe similar to `np.linalg.LinAlgError`, or at least not have the kernel terminated.

**Desktop**
MacOS 26.1, Python 3.13.9, MLX 0.30.0

Anri-Lombard:
## Summary
- Fixed exception propagation from CPU scheduler worker threads to Python
- Added `std::exception_ptr` storage in `StreamThread` to capture exceptions thrown during task execution
- Exceptions are now re-thrown during `synchronize()` so they propagate as `RuntimeError` to Python

## Problem
When `mx.linalg.solve` is called with a singular matrix, the LU factorization throws a `std::runtime_error` inside the scheduler worker thread. Previously, this exception was not caught, causing the process to crash with SIGABRT instead of raising a catchable Python exception.

## Test plan
- [x] Added regression test in `test_linalg.py::test_solve` that verifies singular matrix raises `RuntimeError`
- [x] All existing linalg tests pass (16 tests, 181 subtests)
- [x] Manual verification:
```python
import mlx.core as mx
result = mx.linalg.solve(mx.ones((2, 2)), mx.ones((2,)), stream=mx.cpu)
result.tolist()  # Now raises RuntimeError instead of crashing
```

Fixes #2888

### DIFF
--- a/mlx/scheduler.cpp
+++ b/mlx/scheduler.cpp
@@ -44,6 +44,7 @@ void synchronize(Stream s) {
     std::future<void> f = p->get_future();
     scheduler::enqueue(s, [p = std::move(p)]() { p->set_value(); });
     f.wait();
+    scheduler::check_exception(s);
   } else {
     gpu::synchronize(s);
   }
--- a/mlx/scheduler.h
+++ b/mlx/scheduler.h
@@ -20,8 +20,9 @@ struct StreamThread {
   std::condition_variable cond;
   bool stop;
   std::thread thread;
+  std::exception_ptr eptr;
 
-  StreamThread() : stop(false), thread(&StreamThread::thread_fn, this) {}
+  StreamThread() : stop(false), eptr(nullptr), thread(&StreamThread::thread_fn, this) {}
 
   ~StreamThread() {
     {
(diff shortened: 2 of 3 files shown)


awni:
The challenge with this is not so much propagating the exception but making sure the MLX is in a sane state when the exception is raised. Otherwise we don't want to give the client to the opportunity to catch the exception because they may assume it's safe to continue.

Also I think catching the exception only on synchronization is sub-optimal. We should check for exceptions more frequently (like maybe after a `wait` or something like that.

I'm going to close this for now as I don't think it's at the point that we can consider it. But if you want to send a new PR with the above feedback in mind feel free to do so.

PR #2965, opened 2026-01-04
Fix float64 dtype loaded as complex64 from .npy files

### LINKED ISSUE #2958: [BUG] `mx.load` misinterprets dtype as `complex64` instead of  `float64`
**Describe the bug**
`mx.load()` appears to misinterpret the datatype of `.npy` files with 64-bit data as `complex64` when it should be `float64`.

**To Reproduce**

Include code snippet
```console
>>> import numpy as np
>>> import mlx.core as mx
>>> data = np.load("my_data.npy")
>>> print(f"Dtype: {data.dtype}")
Dtype: float64
>>> mx_data = mx.load("my_data.npy")
>>> print(f"Dtype: {mx_data.dtype}")
Dtype: mlx.core.complex64
>>> 

```

**Expected behavior**
Like numpy, the data should be correctly loaded as `float64`

 - OS Version: MacOS 15.6
 - Version 0.30.1

**Additional context**

Anri-Lombard:
## Summary
- Fix `dtype_from_array_protocol()` to handle float64 (size == 8)
- The 'f' case was missing the float64 handler, causing fall-through to 'c' case

## Test plan
- [x] Added `float64` to the dtypes test list in `test_load.py`
- [x] Added regression test `test_load_float64_dtype` for issue #2958

Fixes #2958

### DIFF
--- a/mlx/io/load.cpp
+++ b/mlx/io/load.cpp
@@ -98,6 +98,8 @@ Dtype dtype_from_array_protocol(std::string_view t) {
           return float16;
         else if (size == 4)
           return float32;
+        else if (size == 8)
+          return float64;
       }
       case 'c': {
         return complex64;
--- a/python/tests/test_load.py
+++ b/python/tests/test_load.py
@@ -22,6 +22,7 @@ class TestLoad(mlx_tests.MLXTestCase):
         "int64",
         "float32",
         "float16",
+        "float64",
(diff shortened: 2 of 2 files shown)


awni:
Thanks for the contribution. We already have a fix out for that here: https://github.com/ml-explore/mlx/pull/2960/files

PR #2967, opened 2026-01-04
Use lower-right causal mask alignment consistently

Anri-Lombard:
## Summary
- Document that MLX's `mask="causal"` uses lower-right alignment
- Clarify the difference from PyTorch's default `is_causal=True` (upper-left)

When `T_q != T_kv`, this distinction matters:
- **MLX (lower-right)**: Last query aligns with last key
- **PyTorch default (upper-left)**: First query aligns with first key

References:
- [PyTorch causal_lower_right](https://docs.pytorch.org/docs/stable/generated/torch.nn.attention.bias.causal_lower_right.html)
- [PyTorch causal_upper_left](https://docs.pytorch.org/docs/stable/generated/torch.nn.attention.bias.causal_upper_left.html)

Relates to #2835

### DIFF
--- a/python/tests/test_fast_sdpa.py
+++ b/python/tests/test_fast_sdpa.py
@@ -620,8 +620,12 @@ class TestSDPA(mlx_tests.MLXTestCase):
                             # Compare only the valid portion
                             if mask_str == "causal" and qL > kL:
                                 offset = qL - kL
-                                out_ref = out_ref[:, :, offset:, :]
-                                out_fst = out_fst[:, :, offset:, :]
+                                if t:  # transpose=True: shape is (B, qL, qH, D)
+                                    out_ref = out_ref[:, offset:, :, :]
+                                    out_fst = out_fst[:, offset:, :, :]
+                                else:  # transpose=False: shape is (B, qH, qL, D)
+                                    out_ref = out_ref[:, :, offset:, :]
+                                    out_fst = out_fst[:, :, offset:, :]
 
                             atol = 2e-5 if dtype == "float32" else 3e-4
 


Anri-Lombard:
Hey @zcbenz, it does have causal_lower_right since 2.3 and can be used with SDPA via the attn_mask parameter. I ran a script with:
```
from torch.nn.attention.bias import causal_lower_right
bias = causal_lower_right(T_q, T_kv)
F.scaled_dot_product_attention(q, k, v, attn_mask=bias)
```
to verify.

Here is the tutorial that documents this explicitly: https://docs.pytorch.org/tutorials/intermediate/scaled_dot_product_attention_tutorial.html.

I also verified masks are mathematically identical. For example with T_q=2, T_kv=4:
```
  MLX's mask (using q_off = max(0, kL - qL)):
        k0  k1  k2  k3
  q0 [  1   1   1   0  ]
  q1 [  1   1   1   1  ]

  PyTorch's causal_lower_right(2, 4):
        k0  k1  k2  k3
  q0 [  1   1   1   0  ]
  q1 [  1   1   1   1  ]

  PyTorch's is_causal=True (upper_left):
        k0  k1  k2  k3
  q0 [  1   0   0   0  ]
  q1 [  1   1   0   0  ]
  
  ```
  
The first two are identical; the third is different. This is also consistent with MLX's CUDA backend which uses cuDNN's set_causal_mask_bottom_right.

Is there something specific about the description you think is incorrect? if your concern is that causal_lower_right isn't a direct SDPA parameter (like is_causal=True) but rather a separate utility class, I could clarify the wording to use the full module path torch.nn.attention.bias.causal_lower_right.

**Comment by zcbenz** [comment]
Thanks for linking the docs, this is a new learn for me. On the behavior, it actually depends on whether T_q is larger or smaller than T_kv:

https://github.com/ml-explore/mlx/blob/9052f678b38330b3c98080ca64c2c946a593d841/mlx/backend/cuda/scaled_dot_product_attention.cpp#L204-L208

**Comment by Anri-Lombard** [comment]
Thanks! Fixed to describe the conditional alignment behavior 🙏

**Comment by awni** [comment]
The comment definitely makes sense. But I also find it a bit strange that we switch from lower right to upper left depending on if query is longer or shorter than the keys. It's quite rare for the query to be longer than the keys which is why we never really looked at it carefully.

I'm wondering if we should change the behavior in that case rather than documenting something that is a bit unusual? Or maybe it's a good idea to keep it this way?

**Comment by zcbenz** [comment]
I agree current behavior is unusual, and using lower right for all should be a better choice.

**Comment by awni** [comment]
@Anri-Lombard what do you think about changing the behavior to always be lower right even when QL > KL? Do you want to send a patch to this PR / send a new one instead of this?

**Comment by Anri-Lombard** [comment]
Hey @awni, always lower-right makes sense. The change is minimal (unless I'm missing somethign) - just two cuDNN locations (forward/backward) and the CPU fallback offset calculation. I'll update this PR to make the behavior change instead of just documenting it 👍

**Comment by awni** [comment]
Yes the change should be pretty straight-forward. We may also need to update the mask index calculation in the Metal kernels. If you add a test for this case as well (qL > kL) that would be great. I can help with the metal kernels if needed.

**Comment by Anri-Lombard** [comment]
@awni and @zcbenz updated and took a stab at the Metal kernels as well - feel free to push changes directly or point out where I deviated if you don't mind the extra time so I can learn the convention preferences more 🙏

For qL > kL, early queries have no keys to attend. Softmax of all-masked values gives uniform weights (exp(finite_min - finite_min) = 1), not zeros. Following PyTorch's https://github.com/pytorch/pytorch/issues/108108 convention, we explicitly zero these rows... I think this is the only "big" change.

**Comment by Anri-Lombard** [comment]
@awni you mentioned the tests, the existing test shapes (127, 65, ...) with mask="causal" cover the qL > kL case. Would you prefer an explicit test that verifies early queries output zeros? 🙏

**Comment by awni** [comment]
Nope if it's already tested that is fine!

**Comment by awni** [comment]
I don't think we should ensure 0s in the qL > kL case. It's a problem I've looked at in the past is what to do if every key position for a given query is masked. And right now it's not consistent. For now let's leave it as undefined behavior and then look into a more principled fix if necessary. I also would rather not reduce performance overall to handle an edge case we don't really care much about).

**Comment by Anri-Lombard** [comment]
@awni done - removed the zero-row handling. The qL > kL + causal case is now undefined behavior as suggested. Tests skip that case.

File: mlx/fast.cpp [hunk 1]
```diff
@@ -720,8 +720,8 @@ array scaled_dot_product_attention(
         if (do_causal) {
           int kL = k.shape(-2);
           int qL = q.shape(-2);
-          int q_off = (kL - qL) < 0 ? 0 : (kL - qL);
-          auto q_idx = arange(q_off, q_off + qL, s);
+          int offset = kL - qL;
+          auto q_idx = add(arange(0, qL, s), array(offset, int32), s);
```
awni:
```suggestion
          int offset = kL - qL;
          auto q_idx = arange(offset, qL + offset, s);
```

python/tests/test_fast_sdpa.py
```diff
@@ -582,6 +585,9 @@ def test_sdpa(self):
             for t in transposes:
                 for mask_str in masks:
                     for B, qL, kL, D, qH, kH in shapes:
+                        # Skip causal tests when qL > kL (undefined behavior)
+                        if mask_str == "causal" and qL > kL:
+                            continue
```
awni: Rather than skipping this test, could you add a little step after the computation which checks that the parts which should match do match?

awni: So basically slice off the initial `qL-kL` from the result if it's greater than 0 and then compare.

Anri-Lombard: I'm struggling to implement this, I'm getting a bug where the fast path is outputting garbage since the fallback is not being applied when it should. I'll need some time to figure this out 🙏

awni: You could revert this change for now which would likely fix it. 

```
  const bool sdpa_full_supported_mask = !has_mask || has_arr_mask ||
      (query_sequence_length <= key_sequence_length && do_causal); # <- add that back
```

awni: So it won't dispatch to the fused implementation for that case.. but that's ok for now. And if you want to fix it in a follow on that would be great!

Anri-Lombard: Thanks! All tests now pass locally and tests slice undefined rows instead of skipping 🙏 Happy to do a follow up pr for a proper fix when gL > kL 👍

awni: Great, I will merge it when the CI tests clear.

zcbenz:
It turns out cuDNN does not like this configuration ☹️ :

```
ERROR: test_sdpa (test_fast_sdpa.TestSDPA.test_sdpa) (B=1, qsl=127, ksl=65, head_dim=64, n_q_heads=32, n_kv_heads=8, mask='causal', transpose=False, dtype='float16')
  File "C:\cygwin64\home\cheng\codes\mlx\python\tests\test_fast_sdpa.py", line 637, in test_sdpa
    self.assertLessEqual(mx.max(diff).item(), atol)
                         ~~~~~~~~~~~~~~~~~^^
RuntimeError: graph.prepare() failed: Bottom right causal mask does not support max_s_q > max_s_kv. Please virtually slice the Q tensor and pass it as max_s_q == max_s_kv.
```

(The test does not run in CI as the hardware is not supported by cuDNN)

I'm going to disable cuDNN SDPA for `T_q > T_kv` with `mask='causal'` for now.

Anri-Lombard:
Dang! Sorry to see this @zcbenz 🙏 I can have a look later to see how we could make cuDNN happy, but disabling it for now makes sense

Issue #2930, opened 2025-12-18
[Docs] Simple examples of using mlx.distributed


awni:
It would be nice to have 2-3 really simple standalone examples of using MLX distributed:
 
-  Data parallel training with a simple model
- Tensor parallel inference with a simple model

Maybe some others.

tylergibbs1:
i’m going to start with a runnable data parallel training script under `examples/python`. the docs added in #2973 explain the pieces, but the examples directory still has no script users can launch. i’ll base it on `linear_regression.py`, shard the synthetic data by rank, average gradients with `nn.average_gradients`, and test it locally with `mlx.launch -n 2`. i’ll keep tensor parallel inference out of this pr.

PR #2973, opened 2026-01-08
[Docs] Simple example of using MLX distributed

stefpi:
related issue: #2930 

- added distributed layers into python/nn/layers for auto summarization
- added tensor parallelism inference example in docs
    - described purpose and usage of ShardedToAllLinear and AllToShardedLinear layers
    - showed simple example of combining layers together and its benefit
    - TP applied simple Llama model inference script ([#1403](https://github.com/ml-explore/mlx-examples/pull/1403) on mlx-examples)
 - extracted data parallel example and moved it to dedicated doc

### DIFF
--- a/docs/src/examples/data_parallelism.rst
+++ b/docs/src/examples/data_parallelism.rst
@@ -3,7 +3,8 @@
 Data Parallelism
 ================
 
-MLX enables efficient data parallel distributed training through its distributed communication primitives.
+MLX enables efficient data parallel distributed training through its
+distributed communication primitives.
 
 .. _training_example:
 
@@ -15,7 +16,7 @@ distributed training. Namely, we will average the gradients across a set of
 hosts before applying them to the model.
 
 Our training loop looks like the following code snippet if we omit the model,
-dataset and optimizer initialization.
+dataset, and optimizer initialization.
 
 .. code:: python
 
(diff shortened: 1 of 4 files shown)


stefpi:
@awni where would be the ideal place for `layers.distributed.shard_linear` and `layers.distributed.shard_inplace` functions to be indexed in the docs? I placed the distributed layers under nn/layers, but I can’t find a natural place for the shard functions.

awni:
I think we should make a new subsection under "Neural Networks", call it say "Distributed" and add all the helper routines and maybe also distributed layers there.

docs/README.md
```diff
@@ -38,3 +38,17 @@ the docs. Then force add the `build/html` directory:
...
+
+Install sphinx autobuild
+```
+pip install sphinx-autobuild
+```
+
+Run auto build on docs/src folder
+```
+sphinx-autobuild ./src ./build/html
+```
```
awni, line 54: That's pretty cool.

stefpi:
For a data parallel training example, should I pull the existing example from `usage/distributed` doc into an `example/data parallelism` doc and expand on it? Or should I leave that and make something new?

awni:
We might not need an additional example given we already have something in the usage section. Wdyt?

stefpi:
I think the existing example is good. Maybe we could put the existing example in a dedicated doc under examples section and then reference both DP and TP examples in the `Distributed Communication` doc? That way the DP and TP examples are next to each other and it is easier to expand on it in the future without cluttering the usage doc.

stefpi:
I have extracted the data parallel training example into `examples/data parallelism` and then referenced both DP and TP examples in `usage/distributed communication`. Let me know if thats fine, I can revert it back if not.

PR #2974, opened 2026-01-08
Add ternary quantization (1.6/2 bit packing)

### LINKED ISSUE #1851: Add support for 1.5bit quantization
1.5-bit quantization would be great to try out large models, but MLX currently doesn't support it:
> [!NOTE]
> MLX is able to read most quantization formats from GGUF directly. However,
> only a few quantizations are supported natively: `Q4_0`, `Q4_1`, and `Q8_0`.
> Unsupported quantizations will be cast to `float16`.
(from https://github.com/ml-explore/mlx-examples/blob/main/llms/gguf_llm/README.md?plain=1)

I would like to run the Unsloth DeepSeek R1 version with 1.5-bit quantization on MLX (I can run it with llama.cpp, but I want to try it with MLX so I can also use my old MacBook).

https://unsloth.ai/blog/deepseekr1-dynamic

2-bit would also be nice to have probably.

CC-Yeh:
## Proposed changes

Closes #1851

Add ternary quantization support for weights (w -> {-1, 0, 1}). 
If this change is acceptable, I can add CUDA 2‑bit kernels to this PR, matmul/qmv will follow in the next PR.

## Checklist

### DIFF
--- a/mlx/backend/cuda/CMakeLists.txt
+++ b/mlx/backend/cuda/CMakeLists.txt
@@ -58,6 +58,7 @@ target_sources(
           ${CMAKE_CURRENT_SOURCE_DIR}/quantized/fp_quantize.cu
           ${CMAKE_CURRENT_SOURCE_DIR}/quantized/quantized.cpp
           ${CMAKE_CURRENT_SOURCE_DIR}/quantized/convert_fp8.cu
+          ${CMAKE_CURRENT_SOURCE_DIR}/quantized/ternary_quantize.cu
           ${CMAKE_CURRENT_SOURCE_DIR}/worker.cpp)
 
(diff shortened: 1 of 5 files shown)


awni:
This is very cool but realistically unlikely to merge into MLX for a couple reasons:

- It's not a commonly used quantization format
- It doesn't have hardware support and I don't see that coming. In MLX we are trending towards supporting quantization which is supported in the hardware as it is much faster and otherwise minimizing surface area of quantization to avoid it getting to complex.

I'm going to close it since we don't plan to merge support for this.

If you are looking to hack on quantization playing around with CUDA support on different platforms for ops that we currently have is high a priority.

PR #2975, opened 2026-01-08
[WIP] Remove conda from setup

angeloskath:


.github/actions/setup-macos/action.yml
```diff
@@ -18,7 +18,6 @@ runs:
       shell: bash
       run: xcodebuild -showComponent MetalToolchain
 
-    - uses: conda-incubator/setup-miniconda@v3
+    - uses: actions/setup-python@v6
```
awni: In case you don't recall, it's going to be more complicated. We rely on conda to get the right macos version when building the release. It's something like if the python binary was built with an old macos then it just ignores the minimum target and we end up with duplicate wheels for the same OS.

angeloskath: Oh thanks. I didn't remember that 😬

zcbenz:
Closing in favor of https://github.com/ml-explore/mlx/pull/3491.

PR #2978, opened 2026-01-09
Fix RandomBits::is_equivalent to include width

### LINKED ISSUE #2977: [BUG] RandomBits::is_equivalent() ignores width_
**Describe the bug**
RandomBits::is_equivalent() currently compares only shape_, which incorrectly treats two RandomBits primitives with the same shape but different width_ as equivalent. Since width_ affects output dtype/bit-width, this is a correctness bug and can lead to incorrect reuse decisions in optimization/caching code that relies on primitive equivalence.

**To Reproduce**

I added a minimal doctest regression that constructs two RandomBits primitives with identical shape and different widths and checks they are not equivalent.

Create a file poc.cpp in the tests directory with the following code:

```cpp
#include <doctest/doctest.h>

#include "mlx/primitives.h"
#include "mlx/stream.h"

using namespace mlx::core;

TEST_CASE("RandomBits::is_equivalent must consider width") {
  Stream s(0, Device::cpu);

  Shape shape{32, 32};

  RandomBits rb_u8(s, shape, /*width=*/1);
  RandomBits rb_u32(s, shape, /*width=*/4);

  CHECK_FALSE(rb_u8.is_equivalent(static_cast<const Primitive&>(rb_u32)));

  RandomBits rb_u8_2(s, shape, /*width=*/1);
  CHECK(rb_u8.is_equivalent(static_cast<const Primitive&>(rb_u8_2)));
}
```

Update CMakeLists.txt target sources to include poc.cpp
```yaml
target_sources(
  tests
  PRIVATE allocator_tests.cpp
          array_tests.cpp
          arg_reduce_tests.cpp
          autograd_tests.cpp
          blas_tests.cpp
          compile_tests.cpp
          custom_vjp_tests.cpp
          creations_tests.cpp
          device_tests.cpp
          einsum_tests.cpp
          export_import_tests.cpp
          eval_tests.cpp
          fft_tests.cpp
          load_tests.cpp
          ops_tests.cpp
          random_tests.cpp
          scheduler_tests.cpp
          utils_tests.cpp
          vmap_tests.cpp
          linalg_tests.cpp
          poc.cpp
          ${METAL_TEST_SOURCES})
```

Rebuild and rerun the tests.  When you do you will see: 
```bash
/Users/daniel/Code/C++/mlx/tests/poc.cpp:8:
TEST CASE:  RandomBits::is_equivalent must consider width

/Users/daniel/Code/C++/mlx/tests/poc.cpp:16: ERROR: CHECK_FALSE( rb_u8.is_equivalent(static_cast<const Primitive&>(rb_u32)) ) is NOT correct!
  values: CHECK_FALSE( true )
```

**Expected behavior**
Creating a RandomBits object uses a shape and width.  Shape is the dimensions of the array, while width is the size of each element, and the comparison is_equivalent should compare **both** shape **and** width because the domain values available for a 1 byte width vs a 4 byte width is significantly different 

 - OS Version: MacOS 26.2 
 - MacBook Air M2

MillaFleurs:
Fixes #2977

- RandomBits::is_equivalent now includes width.
- Adds regression test in tests/random_tests.cpp.

### DIFF
--- a/mlx/primitives.cpp
+++ b/mlx/primitives.cpp
@@ -3716,11 +3716,9 @@ std::pair<std::vector<array>, std::vector<int>> RandomBits::vmap(
 }
 
 bool RandomBits::is_equivalent(const Primitive& other) const {
-  // updated to check state() for equivalence instead of
-  // just shape because RandomBits have *both* shape and width
   const RandomBits& r_other = static_cast<const RandomBits&>(other);
   return shape_ == r_other.shape_ && width_ == r_other.width_;
-}  
+}
 
 std::vector<array> Real::vjp(
     const std::vector<array>& primals,
--- a/tests/random_tests.cpp
+++ b/tests/random_tests.cpp
@@ -5,7 +5,6 @@
 #include "doctest/doctest.h"
 
 #include "mlx/mlx.h"
-#include "mlx/primitives.h"
 
 using namespace mlx::core;
 
@@ -714,18 +713,4 @@ TEST_CASE("test laplace") {
     CHECK(all(less(abs(out), array(inf))).item<bool>());
     CHECK(abs(float(mean(out).item<bfloat16_t>())) < 0.1);
   }
-}
-
-TEST_CASE("RandomBits::is_equivalent must consider width") {
-  Stream s(0, Device::cpu);  // <-- FIX: Stream has no default ctor in this repo
-
-  Shape shape{32, 32};
-
-  RandomBits rb_u8(s, shape, /*width=*/1);
-  RandomBits rb_u32(s, shape, /*width=*/4);
-
-  CHECK_FALSE(rb_u8.is_equivalent(static_cast<const Primitive&>(rb_u32)));
-
-  RandomBits rb_u8_2(s, shape, /*width=*/1);
-  CHECK(rb_u8.is_equivalent(static_cast<const Primitive&>(rb_u8_2)));
 }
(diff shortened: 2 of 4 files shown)


tests/random_tests.cpp
```diff
@@ -714,3 +715,17 @@ TEST_CASE("test laplace") {
     CHECK(abs(float(mean(out).item<bfloat16_t>())) < 0.1);
   }
 }
+
+TEST_CASE("RandomBits::is_equivalent must consider width") {
+  Stream s(0, Device::cpu);  // <-- FIX: Stream has no default ctor in this repo
```
angeloskath: auto s = default_stream(Device::cpu);

mlx/primitives.cpp
```diff
@@ -3716,9 +3716,11 @@ std::pair<std::vector<array>, std::vector<int>> RandomBits::vmap(
 }
 
 bool RandomBits::is_equivalent(const Primitive& other) const {
+  // updated to check state() for equivalence instead of
+  // just shape because RandomBits have *both* shape and width
   const RandomBits& r_other = static_cast<const RandomBits&>(other);
-  return shape_ == r_other.shape_;
-}
+  return state() == r_other.state();
```
angeloskath: This is probably a good idea but for consistency I would just do what every other primitive does and use the following.

```suggestion
  return shape_ == r_other.shape_ && width_ == r_other.width_;
```

angeloskath (changes requested):
Thanks for the find and the fix. Left a couple of minor comments.

MillaFleurs:
Thank you for the comments, changes made

awni (approved):
Ty!

PR #2979, opened 2026-01-09
Swizzle scales

nastya236:
Improvements to scales swizzling kernel.

Overview:
- Each thread loads 4 rows × 4 bytes = 16 bytes with stride 32, so a warp covers one full 128×4 tile.
- If  K % 4 == 0: vectorized int32 loads 
- If not: byte-by-byte loads with zero-padding ( reduction dim is typically divisible by group_size x 4, so we won't be there often).
- load: global-> registers -> shared
- coalesced store: shared -> global 
- XOR swizzle to avoid bank conflicts

Some numbers for `qqmm` with new swizzling (variance is <0.5%):

| M     | N     | K     | Improvement (%) |
|------:|------:|------:|----------------:|
| 16384 |  4096 | 11008 |           10.19 |
| 32768 |  4096 | 11008 |            9.40 |
| 65536 |  4096 | 11008 |            9.41 |
| 16384 | 11008 |  4096 |           13.54 |
| 32768 | 11008 |  4096 |            9.47 |
| 65536 | 11008 |  4096 |           13.79 |
| 16384 |  4096 | 12288 |            5.59 |
| 32768 |  4096 | 12288 |            5.06 |
| 65536 |  4096 | 12288 |            6.44 |
| 16384 | 12288 |  4096 |           13.48 |
| 32768 | 12288 |  4096 |           11.86 |
| 65536 | 12288 |  4096 |           15.02 |
| 16384 |  5120 | 27648 |            7.59 |
| 32768 |  5120 | 27648 |            5.12 |
| 65536 |  5120 | 27648 |            6.09 |
| 16384 | 27648 |  5120 |           13.18 |
| 32768 | 27648 |  5120 |           12.42 |
| 65536 | 27648 |  5120 |           11.34 |

Probably can be improved further!

### DIFF
--- a/mlx/backend/cuda/quantized/qqmm_utils.cu
+++ b/mlx/backend/cuda/quantized/qqmm_utils.cu
@@ -10,6 +10,11 @@ namespace mlx::core {
 
 namespace cg = cooperative_groups;
 
+constexpr int TILE_ROWS = 128;
+constexpr int TILE_COLS = 4;
+constexpr int TILES_PER_LANE = 1;
+constexpr int LANES_PER_BLOCK = 32;
+
 // To pass scales to tensor cores, they need to be repacked into a tiled layout
 // https://docs.nvidia.com/cuda/cublas/index.html#d-block-scaling-factors-layout
 // Tiled layout for scale factors is very well described in CUTLASS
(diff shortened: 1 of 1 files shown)


mlx/backend/cuda/quantized/qqmm_utils.cu
```diff
@@ -43,121 +43,184 @@ namespace cg = cooperative_groups;
...
 
+  auto [num_blocks, block_dims, shared_mem_bytes] =
+      get_swizzle_launch_args(output_rows, output_cols);
   enc.add_kernel_node(
-      cu::repack_scales,
+      cu::swizzle_scales,
       num_blocks,
       block_dims,
-      0,
+      shared_mem_bytes,
```
awni: I don't think you need that, unless I'm missing something, you can keep it at 0. You only have to set it if the shared memory is changes from call to call.

mlx/backend/cuda/quantized/qqmm_utils.cu
```diff
@@ -43,118 +43,180 @@ namespace cg = cooperative_groups;
...
-
-  constexpr size_t tile_size = rows_per_tile * cols_per_sub_block;
-  return tile_idx * tile_size + offset_in_tile;
+
+inline std::tuple<dim3, dim3> get_swizzle_launch_args(
+    size_t M_swizzled,
+    size_t K_swizzled,
+    int tile_rows = 128,
+    int tile_cols = 4,
+    int tiles_per_lane = 1) {
```
awni: It might be worth making these constexpr at the top of the file (unless you intend to change them), but I see the values hardcoded in the kernel so could just define them once at the top instead.

mlx/backend/cuda/quantized/qqmm_utils.cu
```diff
@@ -43,118 +43,180 @@ namespace cg = cooperative_groups;
...
+  const int grid_dim_y = cg::this_grid().dim_blocks().y;
+
+  int remaining = K_int - bid_x * max_tiles_per_block;
+  int tiles_in_block = min(remaining, max_tiles_per_block);
+  bool valid_tile = tidx * num_tiles_per_thread < tiles_in_block;
+
+  __shared__ int4 strided_scales_thread[max_tiles_per_block * tile_stride];
+
+  // Initialize to zero for padding
+  int thread_tile_rows[num_tile_rows_per_thread] = {0, 0, 0, 0};
```
awni: Nit:

```suggestion
  int thread_tile_rows[num_tile_rows_per_thread] = {0};
```

Should zero-initialize the whole array and doesn't require hard coding the size.

mlx/backend/cuda/quantized/qqmm_utils.cu
nastya236: You are not missing anything, memory size is known at compile time!

nastya236:
Thanks for the review! Regarding tests, this is a very good point. I tested it with `examples/python/qqmm.py` on B200. I’ll implement a fallback so we can test correctness

PR #2980, opened 2026-01-09
Fix `grid_dim_x` calculations

CC-Yeh:
## Proposed changes

Fix `grid_dim_x` calculations which only shows error when launching 2D grids.

## Checklist

### DIFF
--- a/mlx/backend/cuda/quantized/affine_quantize.cu
+++ b/mlx/backend/cuda/quantized/affine_quantize.cu
@@ -23,8 +23,7 @@ affine_quantize(const T* w, uint8_t* out, T* scales, T* biases, size_t size) {
   auto tidx = block_idx.x * block_size.x + idx_in_block.x;
   auto tidy = block_idx.y * block_size.y + idx_in_block.y;
 
-  auto grid_dim_x =
-      cg::this_grid().dim_blocks().x * cg::this_grid().block_index().x;
+  auto grid_dim_x = cg::this_grid().dim_blocks().x * block_size.x;
   constexpr float eps = 1e-7;
   constexpr int simd_size = WARP_SIZE;
   constexpr float n_bins = (1 << bits) - 1;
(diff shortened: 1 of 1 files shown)


awni:
Looks good but is there a test case for this? Or do we only launch 2D grids for very large inputs (which we can't really test in CI)?

CC-Yeh:
> Looks good but is there a test case for this? Or do we only launch 2D grids for very large inputs (which we can't really test in CI)?

Yeah, I think it would be too big.

```cpp

bool large = w.size() > UINT_MAX;
------
std::tuple<dim3, uint> get_launch_args(
    size_t size,
    const Shape& shape,
    const Strides& strides,
    bool large,
    int work_per_thread /* = 1 */,
    uint max_block_dim /* = 1024 */) {
  size_t nthreads = cuda::ceil_div(size, work_per_thread);
  uint block_dim = max_block_dim < nthreads ? max_block_dim : nthreads;
  dim3 num_blocks;
  if (large) {
    num_blocks = get_2d_grid_dims(shape, strides, work_per_thread);
    num_blocks.x = cuda::ceil_div(num_blocks.x, block_dim);
  } else {
    num_blocks.x = cuda::ceil_div(nthreads, block_dim);
  }
  return std::make_tuple(num_blocks, block_dim);
}
```

awni:
@nastya236 do you mind double checking this? I thought you changed it recently to fix a bug I had written but maybe the change was also not exact?

nastya236:
Thanks @CC-Yeh! @awni you remember correctly, but I fixed only in `fp_quanitze` / `fp_dequantize`. Also @CC-Yeh `qqmm_utils.cu` would be fixed by [this](https://github.com/ml-explore/mlx/pull/2979) PR.

awni:
Do you mind rebasing to resolve the conflict, then we can run tests + merge?

CC-Yeh:
> Do you mind rebasing to resolve the conflict, then we can run tests + merge?

Done!

PR #2981, opened 2026-01-09
Allow some non 2D inputs in qqmm

awni:
Allow x to be ND when the weights are 2D since it can be flattened or expanded to 2D.


awni:
CC @nastya236

PR #2982, opened 2026-01-09
Don't try to use NAX at run-time if kernels aren't there

awni:
In some cases an MLX binary built for macOS 14 or 15 can end up running on a machine with OS 26.2. It will try to load the NAX kernels since it's a run-time check and it will crash.

### DIFF
--- a/mlx/backend/metal/device.h
+++ b/mlx/backend/metal/device.h
@@ -266,6 +266,9 @@ Device& device(mlx::core::Device);
 std::unique_ptr<void, std::function<void(void*)>> new_scoped_memory_pool();
 
 inline bool is_nax_available() {
+#ifdef MLX_METAL_NO_NAX
+  return false;
+#else
   auto _check_nax = []() {
     bool can_use_nax = false;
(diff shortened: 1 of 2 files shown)


mlx/backend/metal/device.h
```diff
@@ -266,6 +266,9 @@ Device& device(mlx::core::Device);
 std::unique_ptr<void, std::function<void(void*)>> new_scoped_memory_pool();
 
 inline bool is_nax_available() {
+#ifdef MLX_METAL_NO_NAX
```
andresy, line 269: Wouldn't it be better to positively define `MLX_METAL_NAX` instead when NAX is present?

awni: Well this is the minimum edit distance version since there is one path in the cmake that we want to turn off the NAX support for but multiple in the cmake that we want it on for.

Over time I think this makes more sense as by default we will want to build to support neural accelerators... but I'm not so attached to it.. I can rearrange it if needed.

andresy (approved):

PR #2984, opened 2026-01-11
Fp qmv

awni:
- Adds a basic qmv kernel for fp quants for CUDA.
- Adds a simple quantize-dequantize kernel for CUDA, Metal, CPU
- Routes the qqmv to the quantize-dequantize + qmv for all backends


awni:
Moving out of draft.

There is a nice speedup over qqmm with cublas for the qmv case. On a Spark:

quant | GB/s pre | GB/s post
--- | --- | ---
nvfp4 | 164.163 |  232.178
mxfp8 | 178.034 |  221.105

awni:
I think we can optimize the fp_qmv a bit more.. but it's a good start so probably worth landing and hill-climbing.

angeloskath (approved):
It looks great and the perf seems already great... I guess the hill will be small :-)

PR #2986, opened 2026-01-12
replace MLX_IBV_COORDINATOR with MLX_JACCL_COORDINATOR

Evanev7:
## Proposed changes

Expect MLX_JACCL_DEVICES instead of MLX_IBV_DEVICES in the JACCL distributed backend, consistent with the JACCL coordinator env var.

Please close if this was intended - but I couldn't find any discussion explicitly intending this behaviour.

## Checklist

- [ ] (N/A) I have added tests that prove my fix is effective or that my feature works

### DIFF
--- a/docs/src/usage/distributed.rst
+++ b/docs/src/usage/distributed.rst
@@ -584,7 +584,7 @@ the process.
 **MLX_JACCL_COORDINATOR** should contain the IP and port that rank 0 can listen
 to all the other ranks connect to in order to establish the RDMA connections.
 
-**MLX_IBV_DEVICES** should contain the path to a json file that contains the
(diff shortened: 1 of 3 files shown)


awni (approved):
LGTM!

angeloskath:
Sorry for not being able to get to this sooner but that was intentional as the devices are infiniband verbs devices. Moreover this PR breaks the launcher currently.

I will make a PR to revert it but if people prefer this for some reason the I will adapt the PR to fix the launcher.

angeloskath:
Well what I meant is breaks old launchers. Not the launcher on main.

Evanev7:
ah my bad! I think IBV devices makes sense in that case.

PR #2987, opened 2026-01-12
fix distributed all_to_sharded bias shard axis from -2 to -1

gufengc:
## Proposed changes

## Checklist

### DIFF
--- a/python/mlx/nn/layers/distributed.py
+++ b/python/mlx/nn/layers/distributed.py
@@ -87,7 +87,7 @@ def _all_to_sharded(segments):
 
     def _shard_fn(path, weight):
         if path.endswith("bias"):
-            return -1
+            return -1, segments
         return max(weight.ndim - 2, 0), segments
 
     return _shard_fn


python/mlx/nn/layers/distributed.py
```diff
@@ -86,6 +86,8 @@ def _all_to_sharded(segments):
     representation becomes a sharded representation."""
 
     def _shard_fn(path, weight):
+        if path.endswith("bias"):
+            return -1, segments
```
awni, line 90: The axis will be 0 `max(1 - 2, 0) = 0` which is the same as `-1` for the bias since it's 1D.

awni:
I don't think this fix is doing anything. But if there is a broken test case you observed then please share and we can revisit it.

python/mlx/nn/layers/distributed.py
gufengc: Hi, thank you for review. The broken test case is I am tring add gpt-oss tensor parallelism on mlx-lm.
See [https://github.com/ml-explore/mlx-lm/pull/757](https://github.com/ml-explore/mlx-lm/pull/757).

When shard_inplace mlp.experts.gate_proj, the bias shape is 2D due to multi local expert.
Apply this change, I can run mlx-lm pr success.

PR #2995, opened 2026-01-14
[Metal] Add Flash Attention VJP for training

Brooooooklyn:
## Summary

Implements fused backward pass (VJP) for `scaled_dot_product_attention` on Metal GPU. This enables efficient gradient computation during training without falling back to unfused (decomposed) attention operations.

### New Files
- `mlx/backend/metal/kernels/sdpa_vector_vjp.h` - Vector VJP kernel for short sequences (L ≤ 8)
- `mlx/backend/metal/kernels/steel/attn/kernels/steel_attention_vjp_dq.h` - STEEL dQ gradient kernel
- `mlx/backend/metal/kernels/steel/attn/kernels/steel_attention_vjp_dkv.h` - STEEL dK/dV gradient kernel

### Modified Files
- `mlx/backend/metal/scaled_dot_product_attention.cpp` - VJP dispatch logic (+840 lines)
- `mlx/fast.cpp` / `mlx/fast_primitives.h` - Logsumexp caching, VJP routing
- `python/tests/test_fast_sdpa.py` - Comprehensive VJP tests (+220 lines)

## Implementation Notes

Uses a **two-kernel approach** to avoid atomic operations:

1. **dQ kernel** (`steel_attention_vjp_dq.h`): 
   - Computes query gradients via outer loop over KV blocks
   - Uses log2 domain for numerical stability
   - Proper clamping to prevent overflow (exp2 arg clamped to [-88, 0])

2. **dK/dV kernel** (`steel_attention_vjp_dkv.h`):
   - Uses K-row ownership model where each simdgroup owns exclusive rows
   - Eliminates race conditions in GQA where multiple query heads share KV
   - No atomic operations needed

3. **Vector VJP** (`sdpa_vector_vjp.h`):
   - Optimized path for short sequences (L ≤ 8)
   - Uses float32 accumulators for half/bfloat16 precision
   - Shared memory reduction for efficiency

### Key Features
- Float32 accumulators for half/bfloat16 precision
- Logsumexp caching from forward pass for VJP reuse
- Proper GQA (grouped query attention) support
- Causal mask support

## Limitations
- Falls back to unfused attention for mask/sinks gradients (per existing design)
- Requires logsumexp from forward pass (training mode only)
- Head dimension D=256 not supported in vector VJP (32KB threadgroup memory limit)

## Test Plan
- [x] Existing `test_sdpa_grad` passes
- [x] New comprehensive VJP tests added:
  - `test_sdpa_grad_vector_path` - short sequences (L=1,4,7,8)
  - `test_sdpa_grad_steel_path` - longer sequences (L=16,32,128,256)
  - `test_sdpa_grad_head_dims` - head dimensions (D=32,64,96,128)
  - `test_sdpa_grad_gqa` - GQA configurations (4:1, 8:1, 16:1, MHA)
  - `test_sdpa_grad_dtypes` - float16, bfloat16, float32
  - `test_sdpa_grad_edge_cases` - L=1, non-power-of-2, large batch, qL≠kvL

All 21 SDPA tests pass (1 skipped for unrelated disabled feature).

### DIFF
--- a/mlx/backend/cuda/scaled_dot_product_attention.cpp
+++ b/mlx/backend/cuda/scaled_dot_product_attention.cpp
@@ -555,7 +555,6 @@ bool ScaledDotProductAttention::use_fallback(
     bool has_mask,
     bool has_arr_mask,
     bool do_causal,
-    bool is_training,
     bool output_logsumexp,
     Stream s) {
   if (s.device == Device::cpu) {
@@ -618,7 +617,12 @@ void ScaledDotProductAttention::eval_gpu(
   }
 }
 
-bool ScaledDotProductAttentionVJP::use_fallback(const array& q, Stream s) {
+bool ScaledDotProductAttentionVJP::use_fallback(
+    const array& q,
+    Stream s,
+    bool has_mask,
+    bool has_sinks,
+    int /* n_kv_heads */) {
   // The frontend adds a padding mask when sequence length is not a multiple of
   // tile size.
   if (q.shape(2) % 128 != 0) {
--- a/mlx/backend/metal/kernels/steel/attn/kernels/steel_attention.h
+++ b/mlx/backend/metal/kernels/steel/attn/kernels/steel_attention.h
@@ -14,6 +14,7 @@ constant bool align_K [[function_constant(201)]];
 constant bool has_mask [[function_constant(300)]];
 constant bool do_causal [[function_constant(301)]];
 constant bool has_sinks [[function_constant(302)]];
+constant bool output_logsumexp [[function_constant(303)]];
 
 struct MaxOp {
   template <typename T>
@@ -76,6 +77,7 @@ template <
     const constant AttnMaskParams* mask_params [[buffer(5), function_constant(has_mask)]],
     const device MaskType* mask [[buffer(6), function_constant(has_mask)]],
     const device T* sinks [[buffer(7), function_constant(has_sinks)]],
+    device float* LSE [[buffer(8), function_constant(output_logsumexp)]],
     uint simd_lane_id [[thread_index_in_simdgroup]],
     uint simd_group_id [[simdgroup_index_in_threadgroup]],
     uint3 tid [[threadgroup_position_in_grid]],
@@ -454,6 +456,32 @@ template <
   // Normalize output
   Otile.template row_bin_op<DivOp>(sum_score);
   threadgroup_barrier(mem_flags::mem_none);
+
+  // Output logsumexp if requested for VJP backward pass
+  // LSE = max_score + log2(sum_score) in log2 domain (matches STEEL convention)
+  // Physical storage shape: [B*H, qL], laid out as linear array indexed by (B*H
+  // + head)*qL + query_pos LSE_strides[0] = qL (stride between (batch, head)
+  // rows) LSE_strides[1] = 1 (stride between query positions within a row)
+  if (output_logsumexp) {
+    // Compute linear index for (batch, head) combination
+    // This matches the VJP kernel's indexing: (tidl.z * H + tidl.y) *
+    // LSE_strides[0]
+    device float* lse_out =
+        LSE + (tidl.z * params->H + tidl.y) * params->LSE_strides[0];
+
+    // Write one logsumexp per query position in this tile
+    // Each thread handles kRowsPT query positions
+    // align_Q=true means query length is aligned (all blocks full), so always
+    // write align_Q=false means last block is partial, so check bounds
+    STEEL_PRAGMA_UNROLL
+    for (short i = 0; i < kRowsPT; ++i) {
+      int row_pos = tid.x * BQ + tm + sm + (i * decltype(Stile)::kFragRows);
+      if (align_Q || row_pos < params->qL) {
+        AccumType lse_val = max_score[i] + fast::log2(sum_score[i]);
+        lse_out[row_pos * params->LSE_strides[1]] = static_cast<float>(lse_val);
+      }
+    }
+  }
 
   // Store results
   O += (tm + sm) * params->O_strides[2] + sn;
--- a/mlx/backend/metal/kernels/steel/attn/kernels/steel_attention_vjp_dkv.h
+++ b/mlx/backend/metal/kernels/steel/attn/kernels/steel_attention_vjp_dkv.h
@@ -1,0 +1,1004 @@
+// Copyright © 2024-25 Apple Inc.
+
+#pragma once
+
+///////////////////////////////////////////////////////////////////////////////
+// STEEL VJP dKV Kernel for Scaled Dot-Product Attention
+//
+// Part of the two-kernel backward pass optimization that eliminates atomic
+// operations. This kernel computes dK and dV gradients using a REVERSED loop
+// structure compared to the dQ kernel.
+//
+// Grid: [NK, num_kv_heads, B] - one threadgroup per (kv_block, kv_head, batch)
+// Loop: Over gqa_factor query heads, then over Q blocks to accumulate dK/dV
+//
+// For each KV block kb:
+//   1. Load K, V blocks ONCE (resident for entire kernel)
+//   2. Initialize dK, dV accumulators
+//   3. For each Q block qb (that can attend to this KV block):
+//      a. Load Q, O, dO blocks
+//      b. Load LSE for this Q block
+//      c. Compute delta = sum(dO * O) per row
+//      d. Recompute S = scale * Q @ K^T
+//      e. Reconstruct P = exp2(S - LSE)
+//      f. Compute dP = dO @ V^T
+//      g. Compute dS = P * (dP - delta)
+//      h. Accumulate dK += dS^T @ Q (no atomics - local to threadgroup)
+//      i. Accumulate dV += P^T @ dO (no atomics - local to threadgroup)
+//   4. Write dK, dV to output (each simdgroup writes its owned K-rows directly)
+//
+// Key insight: By reversing the loop structure (outer=KV, inner=Q), each
+// threadgroup owns its dK/dV output entirely, eliminating atomic operations.
+//
+// See companion kernel steel_attention_vjp_dq.h for dQ computation.
+///////////////////////////////////////////////////////////////////////////////
+
+using namespace mlx::steel;
+
+// Function constants (match forward kernel indices)
+constant bool align_Q_vjp_dkv [[function_constant(200)]];
+constant bool align_K_vjp_dkv [[function_constant(201)]];
+constant bool do_causal_vjp_dkv [[function_constant(301)]];
+
+///////////////////////////////////////////////////////////////////////////////
+// Transform for scaling (replicated from steel_attention.h)
+///////////////////////////////////////////////////////////////////////////////
+
+// Lower clamp for exp2 arguments to prevent underflow (exp2(-88) ≈ 3.2e-27)
+constexpr constant float kExp2MinArg = -88.0f;
+
+template <typename T>
+struct TransformScaleVJPdKV {
+  T scale;
+  METAL_FUNC TransformScaleVJPdKV(T scale_) : scale(scale_) {}
+
+  METAL_FUNC T apply(T x) const {
+    return scale * x;
+  }
+};
+
+///////////////////////////////////////////////////////////////////////////////
+// STEEL Attention VJP dKV Kernel
+///////////////////////////////////////////////////////////////////////////////
+
+// clang-format off
+template <
+    typename T,
+    int BQ,           // Query block size (32)
+    int BK,           // KV block size (16)
+    int BD,           // Head dimension (64, 96, 128)
+    int WM,           // Warps in M dimension (4)
+    int WN,           // Warps in N dimension (1)
+    typename AccumType = float>
+[[kernel, max_total_threads_per_threadgroup(WM * WN * 32)]]
+void attention_vjp_dkv(
+    // Forward inputs
+    const device T* Q [[buffer(0)]],
+    const device T* K [[buffer(1)]],
+    const device T* V [[buffer(2)]],
+    const device T* O [[buffer(3)]],
+    const device T* dO [[buffer(4)]],
+    const device float* LSE [[buffer(5)]],
+    // Gradient outputs (dK and dV)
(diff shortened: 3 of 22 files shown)


Brooooooklyn:
Notes: I'm working on https://github.com/mlx-node/mlx-node and trying to port some features in [trl](https://github.com/huggingface/trl).
This pull request was generated by Claude Code. I am trying to reduce the computation and memory usage of GRPO training by utilizing the full flash attention feature.

mlx/backend/metal/scaled_dot_product_attention.cpp
```diff
@@ -731,25 +771,793 @@ void ScaledDotProductAttention::eval_gpu(
...
+
+      // Allocate the actual output arrays (will be written after kernel)
+      d_k.set_data(allocator::malloc(d_k.nbytes()));
+      d_v.set_data(allocator::malloc(d_v.nbytes()));
+    } else {
+      // Float32 dtype: can use dK/dV directly with zero-initialization
+      // Must allocate memory before fill_gpu
+      d_k.set_data(allocator::malloc(d_k.nbytes()));
+      d_v.set_data(allocator::malloc(d_v.nbytes()));
+      array zero = array(0.0f, q.dtype());
```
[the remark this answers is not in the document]
Brooooooklyn: The observation is correct, but using `float32` explicitly would be incorrect since `d_k` may have a different `dtype`. Changed to `d_k.dtype()` which is semantically correct - the zero scalar now matches the output array being filled. This follows the pattern used in `matmul.cpp` and avoids unnecessary `dtype` conversions.

mlx/backend/metal/kernels/steel/attn/kernels/steel_attention_vjp_dkv.h
```diff
@@ -0,0 +1,817 @@
...
+
+  // O_smem reuse safety: O_smem is reinterpreted as dS_transposed after delta computation.
+  // O_smem size:        BQ * LDV_tgp * sizeof(T) = BQ * (BD + padV) * sizeof(T)
+  // dS_transposed size: BK * LD_dST * sizeof(AccumType) = BK * BQ * sizeof(float)
+  // Constraint: BK * BQ * 4 <= BQ * (BD + padV) * sizeof(T)
+  //          => BK * 4 <= (BD + padV) * sizeof(T)
+  // For bf16/fp16 (sizeof(T)=2): BK * 4 <= (BD + 8) * 2 => BK * 2 <= BD + 8
+  static_assert(BK * sizeof(AccumType) <= LDV_tgp * sizeof(T),
+      "O_smem buffer too small for dS_transposed reinterpretation: "
+      "BK * sizeof(AccumType) must be <= LDV_tgp * sizeof(T)");
```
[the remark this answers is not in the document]
Brooooooklyn: The alignment is already guaranteed by alignas(sizeof(AccumType)) at line 142. This enforces 4-byte alignment at the O_smem base address. Since all accesses to dS_transposed use integer indexing (byte offset = index × 4), alignment is mathematically guaranteed. The suggested static_assert is redundant because it would verify what the alignas directive already enforces. However, if you prefer explicit documentation, I can add a comment at the reinterpret_cast site explaining the alignment guarantee.

mlx/backend/metal/scaled_dot_product_attention.cpp
```diff
@@ -731,25 +759,864 @@ void ScaledDotProductAttention::eval_gpu(
...
+
+    // Dispatch dQ kernel - computes only dQ gradients
+    sdpa_steel_vjp_dq_dispatch(
+        s, d, q, k, v, o, dO, lse, d_q, scale_, do_causal);
+
+    // Dispatch dKV kernel - computes dK and dV gradients directly
+    sdpa_steel_vjp_dkv_dispatch(
+        s, d, q, k, v, o, dO, lse, d_k, d_v, scale_, do_causal);
+  }
+
```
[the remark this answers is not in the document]
Brooooooklyn: This is a false positive from Copilot. The vector VJP code (lines 1541-1643) is NOT dead code and should NOT be removed. You can safely resolve/dismiss this comment.

Brooooooklyn:
@awni @zcbenz Do you have any interesting review this PR?

mlx/backend/cuda/scaled_dot_product_attention.cpp
```diff
@@ -460,7 +459,16 @@ void ScaledDotProductAttention::eval_gpu(
...
 }
 
-bool ScaledDotProductAttentionVJP::use_fallback(const array& q, Stream s) {
+bool ScaledDotProductAttentionVJP::use_fallback(
+    const array& q,
+    Stream s,
+    bool has_mask,
+    bool has_sinks,
+    int /* n_kv_heads */) {
+  // Force unfused attention when masks/sinks present
```
zcbenz: The fused sdpa vjp does work when masks/sinsks present, we just don't compute vjp for them.

https://github.com/ml-explore/mlx/blob/0b25c9c06ceec68df0bfed7d6b1288bb0a6d5b34/mlx/fast.cpp#L881-L890

mlx/backend/metal/kernels/steel/attn/kernels/steel_attention.h
```diff
@@ -460,6 +462,31 @@ template <
   Otile.template row_bin_op<DivOp>(sum_score);
   threadgroup_barrier(mem_flags::mem_none);
 
+  // Output logsumexp if requested for VJP backward pass
```
zcbenz, line 460: Is logsumexp used in the metal vjp implementation? if not there is no need to compute it. It was there only because the cuDNN SDPA requires it.

mlx/fast_primitives.h
```diff
@@ -258,6 +257,21 @@ class ScaledDotProductAttention : public Custom {
   bool do_causal_;
   bool has_sinks_;
   bool output_logsumexp_;
+  // Cache logsumexp for VJP backward pass
```
zcbenz: There is no need caching it, while the output is not available to user it is passed to vjp primitive as input:

https://github.com/ml-explore/mlx/blob/0b25c9c06ceec68df0bfed7d6b1288bb0a6d5b34/mlx/fast.cpp#L873-L877

mlx/backend/cuda/scaled_dot_product_attention.cpp
Brooooooklyn: I've removed the `has_mask || has_sinks` check from the CUDA `use_fallback`. 
I've kept the parameters in the signature for API consistency with Metal, but marked them as unused.

mlx/backend/metal/kernels/steel/attn/kernels/steel_attention.h
Brooooooklyn: Yes, `logsumexp` is used in the Metal VJP implementation. All three backward kernels use it to reconstruct attention probabilities:

- steel_attention_vjp_dq.h: `P = exp2(S - LSE)` (lines 359-383)
- steel_attention_vjp_dkv.h: `P = exp2(S - LSE)` (lines 459-487)
- sdpa_vector_vjp.h: `P = exp2(score - lse)` (line 236)

This is mathematically necessary for the backward pass since we need to reconstruct P to compute `dS = P * (dP - delta)`

mlx/fast_primitives.h
Brooooooklyn: I initially added it not fully understanding MLX's sibling array design. As you pointed out, `array::make_arrays()` links outputs as siblings, and `array::outputs()` reconstructs the full list during VJP traversal - so outputs[1] (logsumexp) is automatically available even though only outputs[0] is returned to the user.

Removed:
- `cached_logsumexp_` member
- `get_cached_logsumexp()` / `set_cached_logsumexp()` methods
- The `set_cached_logsumexp()` call in `eval_gpu()`

Brooooooklyn:
Benchmark Summary for PR
```                                                                                                                                                           
┌──────────────────────────┬────────────────────┬──────────────────────────────────────────┐                                                                                              
│         Category         │   Speedup Range    │                  Notes                   │                                                                                              
├──────────────────────────┼────────────────────┼──────────────────────────────────────────┤                                                                                              
│ Forward Only             │ 1.0x - 1.8x faster │ Consistently faster across all configs   │                                                                                              
├──────────────────────────┼────────────────────┼──────────────────────────────────────────┤                                                                                              
│ Backward Only (float16)  │ 0.8x - 1.3x        │ Mixed, mostly break-even to faster       │                                                                                              
├──────────────────────────┼────────────────────┼──────────────────────────────────────────┤                                                                                              
│ Backward Only (bfloat16) │ 0.95x - 2.0x       │ Generally faster, especially short STEEL │                                                                                              
├──────────────────────────┼────────────────────┼──────────────────────────────────────────┤                                                                                              
│ Full VJP (fwd+bwd)       │ 0.8x - 1.8x        │ Vector path fastest, STEEL break-even    │                                                                                              
└──────────────────────────┴────────────────────┴──────────────────────────────────────────┘                                                                                              
```
Key observations:                                                                                                                                                                                                                                                                                      
- Vector path (L≤8): Generally 1.1x-1.8x faster for VJP                                                                                                                                   
- STEEL path (L>8): Mostly break-even (~1.0x) for backward, which is expected since the main benefit is memory reduction                                                                  
- Long sequences (L≥1024): VJP is ~1.0x (break-even on speed, but saves memory by not materializing attention matrix)

zcbenz:
I ran the benchmark script on a M4 air and it seems that fused VJP is slower for most cases:

```console
$ python benchmarks/python/sdpa_vjp_bench.py 
SDPA VJP Benchmark - dtype=float16
=====================================================================================

[Forward + Backward (VJP)]
  B  H_q  H_kv      L    D |    unfused      fused  speedup     path
-------------------------------------------------------------------------------------
  2    8     8      1   64 |      0.07ms      0.07ms    1.09x   vector
  2    8     8      4   64 |      0.07ms      0.07ms    1.03x   vector
  2    8     8      8   64 |      0.07ms      0.07ms    1.05x   vector
  2    8     8      8  128 |      0.08ms      0.08ms    0.99x   vector
  2    8     8     32   64 |      0.10ms      0.11ms    0.89x    STEEL
  2    8     8     64   64 |      0.18ms      0.21ms    0.84x    STEEL
  2    8     8    128   64 |      0.45ms      0.54ms    0.82x    STEEL
  2    8     8    128  128 |      0.59ms      0.69ms    0.85x    STEEL
  2    8     8    256  128 |      1.93ms      2.32ms    0.83x    STEEL
  1   32     8    512   64 |     11.45ms     14.91ms    0.77x    STEEL
  1   32     8    512  128 |     14.26ms     17.73ms    0.80x    STEEL
  1   32     8   1024   64 |     50.49ms     64.39ms    0.78x    STEEL
  1   32     8   1024  128 |     56.06ms     69.96ms    0.80x    STEEL
  1   32     8   2048  128 |    244.84ms    299.90ms    0.82x    STEEL
  2   32     8    256   64 |      6.16ms      8.06ms    0.76x    STEEL
  2   32     4    256   64 |      6.23ms      8.02ms    0.78x    STEEL
```

Brooooooklyn:
Interesting, slight difference in the results between the M4 Air and my MacBook Pro 16 with the M3 Max and 128GB of RAM.
It appears to be affected by memory bandwidth limitations on the M4 Air; The VJP kernels need to: 
- Recompute attention scores 
- Load logsumexp from memory
- Compute `P = exp(S - LSE)`
- Compute gradients

The primary benefit of Flash Attention VJP is memory, not speed:
- Unfused: O(N²) memory for attention matrix
- Fused: O(N) memory - no attention matrix materialization 

For long sequences (4K+), memory savings can be critical even if speed is similar or slightly slower.

zcbenz:
I tested on a M3 Max and the results are still mixed:

```console
$ python benchmarks/python/sdpa_vjp_bench.py 
SDPA VJP Benchmark - dtype=float16
=====================================================================================

[Forward + Backward (VJP)]
  B  H_q  H_kv      L    D |    unfused      fused  speedup     path
-------------------------------------------------------------------------------------
  2    8     8      1   64 |      0.30ms      0.26ms    1.17x   vector
  2    8     8      4   64 |      0.34ms      0.24ms    1.43x   vector
  2    8     8      8   64 |      0.28ms      0.25ms    1.14x   vector
  2    8     8      8  128 |      0.29ms      0.33ms    0.88x   vector
  2    8     8     32   64 |      0.33ms      0.36ms    0.92x    STEEL
  2    8     8     64   64 |      0.43ms      0.28ms    1.55x    STEEL
  2    8     8    128   64 |      0.29ms      0.33ms    0.86x    STEEL
  2    8     8    128  128 |      0.31ms      0.31ms    0.98x    STEEL
  2    8     8    256  128 |      0.39ms      0.40ms    0.97x    STEEL
  1   32     8    512   64 |      1.02ms      1.01ms    1.02x    STEEL
  1   32     8    512  128 |      1.30ms      1.34ms    0.97x    STEEL
  1   32     8   1024   64 |      3.75ms      3.75ms    1.00x    STEEL
  1   32     8   1024  128 |      4.81ms      4.81ms    1.00x    STEEL
  1   32     8   2048  128 |     16.65ms     16.66ms    1.00x    STEEL
  2   32     8    256   64 |      0.63ms      0.66ms    0.95x    STEEL
  2   32     4    256   64 |      0.65ms      0.67ms    0.97x    STEEL
```

I'm not really familiar with how VJP works for flash attention but generally speaking reduced memory usage should make op faster when it is memory-bound, which is usually the case for mac, so I think fused op being slower likely means something is off.

But anyway the result on vector path looks promising, maybe separate that part into an independent PR first? It would also make reviewing much easier.

Brooooooklyn:
Done, I only preserve the `vector path` in this pull request, while I continue to investigate the reasons behind the slow performance of the STEEL path.

mlx/backend/cuda/scaled_dot_product_attention.cpp
```diff
@@ -460,7 +459,16 @@ void ScaledDotProductAttention::eval_gpu(
...
+  if (has_mask || has_sinks) {
+    return true;
```
zcbenz: This can be removed.

Brooooooklyn:
@awni, can you take a look at this PR?

Brooooooklyn:
Completely rewritten; it is now closed and will reopen as a new one.

PR #2997, opened 2026-01-14
Fix FFT failure for array sizes 2^21 and 2^22

### LINKED ISSUE #1800: [BUG] FFT fails on certain array lengths
**Describe the bug**
Doing an FFT on array lengths 2**21 and 2**22 results in a kernel failure, but larger array sizes works.

**To Reproduce**

A simple script to reproduce:

```python
import mlx.core as mx
size = int(2**21)
x = mx.ones(size)
mx.eval(mx.fft.fft(x, stream=mx.gpu))
```
This will result in the following error:

```bash
Terminating due to uncaught exception: [metal::Device] Unable to load function four_step_mem_8192_float2_float2_0_false
Function four_step_mem_8192_float2_float2_0_false was not found in the library

Abort trap: 6
```

A similar thing happens for an array that is `2**22` long. However, the code succeeds for arrays that have length `2**23`, `2**24`, `2**25`, etc., up to `2**28`. (I don't have enough memory to test beyond that.) By "succeed" I mean the function runs without failure. I haven't checked that the output is actually correct.

**Expected behavior**
The FFT should work for `2**21` and `2**22` if larger array sizes work. At the very least, the error should be caught appropriately with a more graceful exit.

 - OS Version: MacOS 15.1.1
 - Version 0.22.0

**Additional context**
Digging into the code a bit I can see why it's failing. For a size of 2**21, `plan.n1` [here](https://github.com/ml-explore/mlx/blob/d0f471cff734889ecae45a81e42d46048f791dbf/mlx/backend/metal/fft.cpp#L128) will get set to 2048. Later on, that will cause `threadgroup_mem_size` to get set to 8192 [here](https://github.com/ml-explore/mlx/blob/d0f471cff734889ecae45a81e42d46048f791dbf/mlx/backend/metal/fft.cpp#L632-L639). However, I don't know why that doesn't cause the assert at [line 641](https://github.com/ml-explore/mlx/blob/d0f471cff734889ecae45a81e42d46048f791dbf/mlx/backend/metal/fft.cpp#L641) to raise an error.

I see the comment at line 640 that `// FFTs up to 2^20 are currently supported`, so I'm not sure why the 2^23 FFTs are running. Even if the assert worked properly, why the limit of 2^20? In the research application we're trying to use this for we will be evaluating arrays of 2^21 - 2^25, so it would be ideal if these array sizes could be handled.

Anri-Lombard:
Fixes #1800 by lowering the four-step FFT threshold from 4096 to 1024, forcing recursive decomposition to avoid Metal threadgroup memory limits. Added test case for 2^21, 2^22, 2^23.

### DIFF
--- a/mlx/backend/metal/fft.cpp
+++ b/mlx/backend/metal/fft.cpp
@@ -20,6 +20,7 @@ namespace mlx::core {
 using MTLFC = std::tuple<const void*, MTL::DataType, NS::UInteger>;
 
 #define MAX_STOCKHAM_FFT_SIZE 4096
+#define MAX_SAFE_FFT_SIZE 1024
 #define MAX_RADER_FFT_SIZE 2048
 #define MAX_BLUESTEIN_FFT_SIZE 2048
 // Threadgroup memory batching improves throughput for small n
@@ -121,7 +122,7 @@ FFTPlan plan_fft(int n) {
   int remaining_n = n;
 
   // Four Step FFT when N is too large for shared mem.
-  if (n > MAX_STOCKHAM_FFT_SIZE && is_power_of_2(n)) {
+  if (n > MAX_SAFE_FFT_SIZE && is_power_of_2(n)) {
     // For power's of two we have a fast, no transpose four step implementation.
     plan.four_step = true;
     // Rough heuristic for choosing faster powers of two when we can
--- a/python/tests/test_fft.py
+++ b/python/tests/test_fft.py
@@ -318,6 +318,14 @@ class TestFFT(mlx_tests.MLXTestCase):
             dgdx = torch.func.grad(g)(torch.tensor(x))
             self.assertLess((dfdx - dgdx).abs().max() / dgdx.abs().mean(), 1e-4)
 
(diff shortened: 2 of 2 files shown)


zcbenz:
@barronalex Can you please check if this is the correct fix?

davidopascual:
I tested this patch comparing the fft for different array sizes comparing the MLX GPU output vs NumPy CPU and this patch appears not to work. The error blows up past 2^20:
fft_differences

Here is the code I used to generate the data for that plot: 
```
for i, array_size in enumerate(array_sizes):
        power = powers[i]
        
        # Generate test data
        np.random.seed(42)
        x = np.random.randn(array_size).astype(np.complex64)
        
        # NumPy reference
        numpy_fft = np.fft.fft(x)
        
        # MLX GPU
        mx.set_default_device(mx.gpu)
        mlx_gpu_fft = np.array(fft(mx.array(x)))
        gpu_diff = np.max(np.abs(numpy_fft - mlx_gpu_fft))
```

zcbenz:
@davidopascual Thank you so much for testing the branch! 🙏

PR #2999, opened 2026-01-14
Reverts changing the MLX_IBV_DEVICES to MLX_JACCL_DEVICES

angeloskath:
Reverts #2986 .


awni (approved):
Oops should have waited for you on that one

angeloskath:
Nah no worries, we should probably do a patch release soon-ish cause it might break other people's workflows see exo for instance.

Issue #3000, opened 2026-01-15
[Enhancement] Enable include-what-you-use


MillaFleurs:
** Issue **
Best Practices for C++ Development recommend that includes are not implicit.  What that means is every file should have all the include statements necessary.

See https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#rs-implicit

** Proposed Enhancement ** 

There is a clang tool called include-what-you-use or IWYU that will parse files and report what includes need to be added or deleted.  

Installing IWYU is:
```bash
 brew install include-what-you-use
```

Building with the IWYU flag is:
```bash
mkdir build && cd build
cmake .. -DMLX_ENABLE_IWYU=ON
 make -j
```

Attached is the output of IWYU for the current codebase.

I'm including a git pull request to make this enhancement

[iwyu.txt](https://github.com/user-attachments/files/24630022/iwyu.txt)

zcbenz:
While "C++ Core Guidelines" was created by very well known developers (I have read several books by them) it is not widely adopted as far as I know. For include-what-you-use I think unless it brings actual benefits adopting it only adds overheads.

MillaFleurs:
Well the overhead is related to linting which is something that should be done and doesn't add any actual overhead during use of the program.  It's just something to enable as a test and for code review.

The benefit is you will never be in a situation where you include header A but really you needed to include header C for some reason.  Meaning you reduce bugs / compile errors.

If you add an explicit #include statement it shouldn't add any additional overhead because an implicit include means it's already in the build.

Thoughts?

zcbenz:
Sorry by overhead I meant maintenance overhead: an extra lint that requires you adding a new include when building and testing pass. I worked with a few projects linting IWYU before, it was really annoying and never caught a bug.

Regarding the benefits:

> The benefit is you will never be in a situation where you include header A but really you needed to include header C for some reason.

I don't think the linting fixes this situation, it just asks you to add a header and is not capable of detecting unneeded includes. 

> Meaning you reduce bugs / compile errors.

I'm not aware of any bug would be caused by an implicit include. For compilation error, it is already tested by CI.

Issue #2993, opened 2026-01-14
[BUG] array.h mixes signed and unsigned ints, causing potential issues


MillaFleurs:
**Describe the bug**
The following function adds a signed + unsigned number together without explicitly casting, causing potential issues.

```Cpp
// Line 108
  /** The number of dimensions of the array. */
  size_t ndim() const {
    return array_desc_->shape.size();
  }

// Line 118
  /**
   *  Get the size of the corresponding dimension.
   *
   *  This function supports negative indexing and provides
   *  bounds checking. */
  auto shape(int dim) const {
    return shape().at(dim < 0 ? dim + ndim() : dim);
  }

```

**To Reproduce**

dim is signed ndim is unsigned so there's the possibility of unexpected behavior as the below code demonstrates.

```cpp
// test.cpp
#include <cstdio>
#include <cstddef>
#include <limits>
#include <initializer_list>

int main() {
  std::size_t ndim = 3;

  for (int dim : {-1, -2, -3, -4}) {
    std::size_t idx = (dim < 0) ? (ndim + dim) : static_cast<std::size_t>(dim);
    std::printf("ndim=%zu dim=%d -> idx=%zu\n", ndim, dim, idx);
  }

  std::printf("SIZE_MAX=%zu\n", std::numeric_limits<std::size_t>::max());
  return 0;
}

// compile and run using the following command:
// clang++ -std=c++20 -Wall -Wextra -Wsign-conversion test.cpp && ./a.out
```

**Expected behavior**
This is a quick patch and I'm not sure why this wasn't caught in linting.  As you can see if comiled with -Wall it comes up as a warning when this happens.

**Desktop (please complete the following information):**
 - OS Version: Tahoe 26.2

MillaFleurs:
Currently working on this

zcbenz:
As the title says, it is potential but not really causing issues, because the `at()` method checks bounds. Explicit casting does not fix anything, and explicit boundary check only adds overhead.

MillaFleurs:
@zcbenz you make a very interesting point and indeed at() already throws an out of range.  However, the test I included failed prior to applying the patch I made, and this triggers -Wsign-conversion (and becomes a hard failure under -Werror). I’m aiming to keep the code warning-free under stricter builds.

[...]

PR #3001, opened 2026-01-15
Add iwyu to cmake

MillaFleurs:
## Proposed changes

Adding include-what-you-use per https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#rs-implicit PR 3000

## Checklist

- [X ] I have run `pre-commit run --all-files` to format my code / installed pre-commit prior to committing changes

### DIFF
--- a/mlx/array.h
+++ b/mlx/array.h
@@ -16,20 +16,21 @@
 
 namespace mlx::core {
 
-  namespace detail {
-    inline std::size_t normalize_dim(int dim, std::size_t ndims) {
-      using S = std::ptrdiff_t;
-      S d = static_cast<S>(dim);
-      S n = static_cast<S>(ndims);
+namespace detail {
+inline std::size_t normalize_dim(int dim, std::size_t ndims) {
+  using S = std::ptrdiff_t;
+  S d = static_cast<S>(dim);
+  S n = static_cast<S>(ndims);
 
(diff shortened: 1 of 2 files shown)


MillaFleurs:
> I think replacing `dim < 0 ? dim + ndim() : dim` with a utility function decreases code readability without actual benefits. Per IWYU please discuss at #3000.

See PR 2993 on that question.  The line you're referring to fixed a bug which is the actual benefits.  Sorry I was not clear on that part. I thought I had reset my git repo to match the current version on GitHub so that it would be a different patch but I appear not to have done that correctly.  Mea Culpa.

Regarding the changes in this PR (besides PR 2993), they are arguably bugs in that if you were to compile and there is a missing an include it would throw an error.  Yes it's a minor inconvenience as presumably one would see the compile error and add the missing include but IMHO good practice is to reduce potential defects where possible...  Although I understand that may be a philosophical discussion for some.  I am just trying to be helpful while spotting some issues, trying to familiarize myself with the codebase....

zcbenz:
The `at()` method of `std::vector` already throws for out-of-range access.

MillaFleurs:
@zcbenz Maybe discuss in #2993 ?

awni:
Closing this as it was fixed in #2994 and the iwyu is not planned

PR #3003, opened 2026-01-16
Type Enhancement for Func Transforms and Bug Fix

XXXXRT666:
## Proposed changes

1. Functions lose precise type hints after transformation. For example, mx.compile only annotates its input and output as Callable, so the original function signature and return types are not preserved in the type system.
2. Errors occur during the stub gen.

截屏2026-01-16 22 51 04

Type Hints After This PR:

截屏2026-01-16 20 23 34

截屏2026-01-16 20 23 09

## Checklist

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -169,15 +169,10 @@ class CMakeBuild(build_ext):
             self.copy_tree(regular_dir, inplace_dir)
 
         # Build type stubs.
-        ext_fullpath = Path.cwd() / self.get_ext_fullpath(ext.name)  # type: ignore[no-untyped-call]
-        extdir = ext_fullpath.parent.parent.resolve()
         build_temp = Path(self.build_temp) / ext.name
-        env = os.environ.copy()
-        env["PYTHONPATH"] += f":{extdir}"
         subprocess.run(
             ["cmake", "--install", build_temp, "--component", "core_stub"],
             check=True,
-            env=env,
         )
 
 
--- a/setup.py
+++ b/setup.py
@@ -169,10 +169,15 @@ class CMakeBuild(build_ext):
             self.copy_tree(regular_dir, inplace_dir)
 
         # Build type stubs.
+        ext_fullpath = Path.cwd() / self.get_ext_fullpath(ext.name)  # type: ignore[no-untyped-call]
+        extdir = ext_fullpath.parent.parent.resolve()
         build_temp = Path(self.build_temp) / ext.name
+        env = os.environ.copy()
+        env["PYTHONPATH"] += f":{extdir}"
         subprocess.run(
             ["cmake", "--install", build_temp, "--component", "core_stub"],
             check=True,
+            env=env,
         )
 
 
--- a/docs/src/python/transforms.rst
+++ b/docs/src/python/transforms.rst
@@ -11,6 +11,7 @@ Transforms
    eval
    async_eval
    compile
+   checkpoint
    custom_function
    disable_compile
    enable_compile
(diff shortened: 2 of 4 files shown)


XXXXRT666:
Also, `python setup.py generate_stubs` no longer works, despite being mentioned [here](https://ml-explore.github.io/mlx/build/html/install.html#python-api).

zcbenz:
> Also, `python setup.py generate_stubs` no longer works, despite being mentioned [here](https://ml-explore.github.io/mlx/build/html/install.html#python-api).

Thanks for noticing this, the stubs generation is now part of standard build command after https://github.com/ml-explore/mlx/pull/2949. Can you remove that in document?

setup.py
```diff
@@ -169,10 +169,15 @@ def run(self):
             self.copy_tree(regular_dir, inplace_dir)
 
         # Build type stubs.
+        ext_fullpath = Path.cwd() / self.get_ext_fullpath(ext.name)  # type: ignore[no-untyped-call]
+        extdir = ext_fullpath.parent.parent.resolve()
         build_temp = Path(self.build_temp) / ext.name
+        env = os.environ.copy()
+        env["PYTHONPATH"] += f":{extdir}"
```
zcbenz: I'm not sure if this actually fixes the problem you saw, the exact path has already been passed in the stubgen target:

https://github.com/ml-explore/mlx/blob/d2bef3c6bbb7acd8b311aeab5deeefc4dff3c246/python/src/CMakeLists.txt#L35-L40

Can you share the exact command that you used?

XXXXRT666: the command is `pip install git+https://github.com/ml-explore/mlx.git -vv` and `pip install -vv git+https://github.com/XXXXRT666/mlx.git@type-fix`

XXXXRT666:
> Thanks for noticing this, the stubs generation is now part of standard build command after #2949. Can you remove that in document?

I didn't found it in the repo, maybe it's has been modified

zcbenz:
> I didn't found it in the repo, maybe it's has been modified

Sorry my bad, it was already removed in https://github.com/ml-explore/mlx/pull/2949, the website has not been updated yet.

setup.py
XXXXRT666: `CMAKE_CURRENT_SOURCE_DIR = ~/mlx/python/src` in my environment, but mlx.core is generated in `~/mlx/build/lib.macosx-11.0-arm64-cpython-310/mlx/core.cpython-310-darwin.so`

zcbenz: Thanks a lot for the information! Can you check if https://github.com/ml-explore/mlx/pull/3009 fixes the error for you?

XXXXRT666: > Thanks a lot for the information! Can you check if #3009 fixes the error for you?

It doesn't fixed the error, I think the correct path is `$<TARGET_FILE_DIR:core>/..` as `<TARGET_FILE_DIR:core>` is the path to the directory of the target `.so` file, which is `~/mlx/build/lib.macosx-11.0-arm64-cpython-310/mlx`

zcbenz: Oh I only tested local install and I guess it searched from the other path, thanks for testing. Do you mind checking the updated PR again? This time it passes the remote install for me. `pip install -v git+https://github.com/zcbenz/mlx.git@stubgen-python-path`

zcbenz (approved):
I don't have much experience with python typing but after reading the docs I think this is just standard usage. Still needs a check from @awni .

setup.py
XXXXRT666: Yes no error occurs, stubs are generated correctly

awni:
The typing is nice, thanks for fixing that. I added a couple more for completeness.

awni (approved):
LGTM!

Issue #3007, opened 2026-01-18
[BUG] nullpointer in TCPSocket::connect causes seg fault


MillaFleurs:
**Describe the bug**
A nullpointer in TCPSocket::connect causes seg fault in some situations

**To Reproduce**

Compile and run the following code:

Include code snippet
```cpp
#include <iostream>
#include "mlx/distributed/utils.h"

int main() {
  using namespace mlx::core::distributed::detail;

  // Pick an address that is almost certainly not listening.
  // Port 1 on localhost is typically closed -> connect() fails immediately.
  auto addr = parse_address("127.0.0.1:1");

  std::cout << "About to call TCPSocket::connect with default cb=nullptr...\n";

  // num_retries=1, wait=0, cb omitted -> cb is nullptr.
  // On connect failure, utils.cpp calls cb(attempt, wait) unconditionally -> segfault.
  auto sock = TCPSocket::connect("repro", addr, /*num_retries=*/1, /*wait=*/0);

  // We should never get here.
  (void)sock;
  std::cout << "Unexpectedly connected.\n";
  return 0;
}
```

Compile and running shows a seg fault:
```bash
c++ -std=c++17 -I. repro_null_cb.cpp mlx/distributed/utils.cpp -o repro_null_cb
./repro_null_cb
#output:
#About to call TCPSocket::connect with default cb=nullptr...
#libc++abi: terminating due to uncaught exception of type std::__1::bad_function_call: std::exception
#fish: Job 1, './repro_null_cb' terminated by signal SIGABRT (Abort)
```

The offending line is here in line 59 of utils.h

```cpp
      std::function<void(int, int)> cb = nullptr);
```

Because we are explitily allowing a nullptr but then in utils.cpp we never check if there indeed is a nullptr passed in:

```cpp
// line 187
    cb(attempt, wait);
```

**Expected behavior**
The default for this argument should NOT be a nullptr in order to prevent potential segfaults.  We can update the header to fix this.  Alternatively, if there is some reason to expect a nullptr to get passed in, utils.cpp should check if cb is null before calling it

**Desktop (please complete the following information):**
 - OS Version: 26.2

PR #3008, opened 2026-01-18
PR 3007 Fix Seg Fault

MillaFleurs:
## Proposed changes

Fixes a potential seg fault wherein utils.h defaults to a nullptr but utils.cpp doesn't check to see if it's actually passed a nullptr issue #3007 .

## Checklist

- [ X] I have run `pre-commit run --all-files` to format my code / installed pre-commit prior to committing changes
- [ X] I have updated the necessary documentation (if needed)

### DIFF
--- a/mlx/distributed/utils.cpp
+++ b/mlx/distributed/utils.cpp
@@ -184,7 +184,9 @@ TCPSocket TCPSocket::connect(
       break;
     }
 
-    cb(attempt, wait);
+    if (cb != nullptr) {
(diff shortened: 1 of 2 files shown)


zcbenz (commented):
I think we should check null before calling the callback as caller may pass null themselves.

angeloskath:
Thanks @MillaFleurs for the find and fix and @zcbenz for the fix of fix :-) Let's merge after the tests pass.

PR #3009, opened 2026-01-18
Fix python package install path in stubgen

zcbenz:
We need to pass the path where the mlx package is installed to `stubgen`, previously it was hard coded to `~/mlx/python` but would fail for some cases like https://github.com/ml-explore/mlx/pull/3003#discussion_r2701997478. This PR fixes it by also searching the actual path of extension with the ` $<TARGET_FILE_DIR>` expression.


awni (approved):
Thx!

PR #3010, opened 2026-01-18
Only clear disk when building for release

zcbenz:
IO is very slow in CI so removing lots of files would add 1~2 minutes to run time. Only do it for release build.

Also removes `required: false` from the inputs, it literally does nothing and adds cognitive costs when reading code.

### DIFF
--- a/.github/actions/setup-linux/action.yml
+++ b/.github/actions/setup-linux/action.yml
@@ -4,20 +4,25 @@ description: 'Install dependencies for Linux builds'
 inputs:
   toolkit:
     description: 'Which toolkit to install'
-    required: false
     default: 'cpu'
   python-version:
     description: 'Version of python to set up'
-    required: false
     default: '3.14'
+  clear-disk:
+    description: 'Remove unused software to free some disk space'
+    default: 'false'
   use-ccache:
     description: 'Whether to enable ccache'
-    required: false
     default: 'true'
 
 runs:
   using: "composite"
   steps:
+    - name: Clear disk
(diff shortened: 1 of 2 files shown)


.github/actions/setup-linux/action.yml
```diff
@@ -27,7 +32,6 @@ runs:
             zip \
             libblas-dev liblapack-dev liblapacke-dev \
             openmpi-bin openmpi-common libopenmpi-dev
-        sudo rm -rf /usr/share/dotnet /usr/local/lib/android /opt/ghc || true
```
awni, line 30: I added that cause we kept getting disk space usage errors in CI for linux (see e.g. https://github.com/ml-explore/mlx/actions/runs/20985183462/job/60317833989)

awni: And it was mostly for the PR tests not the release IIRC. We might be able to get away with only clearing one of them or clearing something else that doesn't take quite as long?

zcbenz: For the linked job I think it was mostly because of downloading the CUDA python packages for a CPU build. I will fix that part instead.

awni: I see think I fixed it by downloading only the cpu only torch. That may have been enough actually in which case we could just remove this line that clears packages and be done with it

zcbenz: Let me do some experiments in this PR to see if it is enough.

PR #3011, opened 2026-01-18
[CUDA] Faster grouped mm

zcbenz:
Make grouped mm faster that it now takes 4x time than equivalent matmul. It is still too slow but a lot better than original https://github.com/ml-explore/mlx/pull/2902.

Tricks used:
* Use tensor cores for sm80 and later.
* Set alignment for output.
* Encode the kernel function to CUDA graph directly instead of capturing.

We can keep improving this kernel by tuning block sizes, but what we really need to do is pad the data to get higher alignment so we can make use of hardware features like TMA.


awni (approved):
Looks great!

PR #3012, opened 2026-01-18
Update CCCL to v3.1.3

zcbenz:
I didn't notice any change on performance, but might still need some caution since it is a major version bump.


awni (approved):
🙏

PR #3013, opened 2026-01-18
Do not clear disk space in setup-linux

zcbenz:
After the building there are still 20G disk space so there is no need to clear disk space which takes 1~2 minutes.

```
Run df -h
Filesystem      Size  Used Avail Use% Mounted on
/dev/root        73G   54G   20G  74% /
```

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -36,6 +36,7 @@ jobs:
       - uses: ./.github/actions/setup-linux
       - uses: ./.github/actions/build-linux
       - uses: ./.github/actions/test-linux
+      - run: df -h
 
   cuda_build_and_test:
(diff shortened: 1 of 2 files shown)


awni:
Cool, thanks for checking that!

.github/workflows/build_and_test.yml
```diff
@@ -36,6 +36,7 @@ jobs:
       - uses: ./.github/actions/setup-linux
       - uses: ./.github/actions/build-linux
       - uses: ./.github/actions/test-linux
+      - run: df -h
```
awni, line 39: Do you want to keep that?

zcbenz: Yeah it would be helpful when out-of-disk-space error happens next time.

PR #3014, opened 2026-01-18
Do not give workflow boolean inputs default values

zcbenz:
There is a pitfall in GitHub Actions that the default value of workflow_dispatch inputs can only be string, so when setting `default: true` we might get a string "true" or a boolean false as default value (it seems that we ran into the latter case).

This PR just removes the default values so inputs would all default to boolean `false` for pushes.

Also note that I'm using a non-exist env in `name: ${{ inputs.dry_run && 'dry-run' || 'pypi' }}` because `inputs.dry_run && '' || 'pypi'` would always evaluate to `pypi` regardless of the input value, which is another pitfall.

Checked a few cases and they all seem to work as expected:
* ✅ [Manually triggered dry run](https://github.com/ml-explore/mlx/actions/runs/21111585367)
* ❌ [Manually triggered publish](https://github.com/ml-explore/mlx/actions/runs/21111587901)
* ❌ [Push triggered publish](https://github.com/ml-explore/mlx/actions/runs/21111546178)

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -4,18 +4,18 @@ on:
   push:
     tags:
       - 'v*'
+    branches:
+      - 'test-publish/*'
   workflow_dispatch:
     inputs:
-      publish:
-        description: 'Publish to PyPI (uncheck for dry run)'
+      dry_run:
+        description: 'Dry run (do not publish to PyPi)'
         required: false
         type: boolean
-        default: true
       dev_release:
         description: 'Development release (DEV_RELEASE=1)'
         required: false
         type: boolean
-        default: false
 
 permissions:
   contents: read
@@ -29,7 +29,7 @@ jobs:
       - uses: ./.github/actions/build-docs
 
   deploy_documentation:
-    if: ${{ github.event_name == 'push' || inputs.publish }}
+    if: ${{ !inputs.dry_run }}
     needs: build_documentation
     permissions:
       pages: write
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -29,7 +29,7 @@ jobs:
       - uses: ./.github/actions/build-docs
 
   deploy_documentation:
-    if: ${{ github.event_name == 'push' || inputs.publish }}
```
awni, line 32: I'm not sure about removing that: `github.event_name == 'push'`

I added it because these steps weren't getting triggered because the inputs were not defined when it's a push (vs when it's a manual trigger).

awni: If it's not defined will `!inputs.dry_run` evaluate to true?

zcbenz: Yes for pushes the `!inputs.dry_run` would evaluate to true so it defaults to publish run, unlike previous condition `inputs.publish` which defaults to dry run. 

I did a push-triggered run to verify it:
https://github.com/ml-explore/mlx/actions/runs/21111546178
(it ran the `deploy_documentation` job, unlike the [dry run](https://github.com/ml-explore/mlx/actions/runs/20976892857) that did not.)

It is still possible that another pitfall would cause the push-triggered publish to fail, but when that happens I think we can just do a manually-triggered publish and fix it later.

awni: Got it thanks for clarifying!

awni (approved):

PR #3022, opened 2026-01-20
Tensor scale nvfp4

nastya236:
#### Add per tensor scale for `nvfp4` quantization for `cuda` and `cpu`.
`qqmm`, `quantize`, `dequantize` inputs optional 1D `float32` array (`global_scale`) if `mode == "nvfp4"`.

Also some files related to `qqmm` were refactored. 

Important details:

- [`qqmm`] currently if `global_scale` is provided for the first input, it must be provided for the second input as well. This is because we pass global scales as inputs in `QQMatmul::eval_gpu()` and we can't distinguish between `global_scale_x` and `global_scale_w`. 
- `alpha` and `beta` both should be  device or host ptrs. Therefore, if `alpha` is a device ptr, 
Tensor scale will help with small inputs:

```python
import mlx.core as mx

x = mx.random.uniform(shape=(2, 16)) / 1e5
xq_ns, scales_ns = mx.quantize(x, mode="nvfp4")
global_scale=mx.absmax(x).astype(mx.float32)
xq_s, scales_s = mx.quantize(x, mode="nvfp4", global_scale = global_scale)

print(mx.allclose(scales_ns, mx.zeros_like(scales_ns)))
print(mx.allclose(scales_s, mx.zeros_like(scales_s)))

```

TODO: 
- we probably want to support `global_scale` in metal as well but it requires changing all quantized operations
- it is not yet clear what might be the best strategy for scale computation during training (for `x`, `w` as well as for `cotan`), therefore `QQLinear` does not have global scale support. I will add it after some exploration as a separate PR. 
- `fp_qmv_impl` was not updated yet to support global scale

### DIFF
--- a/mlx/backend/cuda/quantized/qqmm_impl.cpp
+++ b/mlx/backend/cuda/quantized/qqmm_impl.cpp
@@ -39,7 +39,7 @@ void qqmm_impl(
       out.dtype(),
       qmode);
 
-  if (scalars.uses_device_pointers()) {
+  if (scalars.has_values()) {
     qqmm.run(
         encoder,
         out,
--- a/mlx/backend/cuda/quantized/qqmm_impl.h
+++ b/mlx/backend/cuda/quantized/qqmm_impl.h
@@ -12,7 +12,7 @@ struct GemmScalars {
   std::optional<array> alpha_device;
   std::optional<array> beta_device;
 
-  bool uses_device_pointers() const {
+  bool has_values() const {
     return alpha_device.has_value();
   }
 };


python/src/ops.cpp
```diff
@@ -4288,6 +4288,8 @@ void init_ops(nb::module_& m) {
 
           * w_q (array): The quantized version of ``w``
           * scales (array): The quantization scales
+          * tensor_scale (array): The per-tensor float32 absolute max
```
awni: This is a bit of a nit, but I don't really love the name `tensor_scale`. Partly because we call them `array` in MLX. What do you think about `global_scale`, or maybe `array_scale`?

python/src/ops.cpp
```diff
@@ -5501,7 +5510,10 @@ void init_ops(nb::module_& m) {
         mode (str, optional): The quantization mode. Default: ``"nvfp4"``.
           Supported modes are ``nvfp4`` and ``mxfp8``. See the
           :ref:`table of quantization modes <quantize-modes>` for details.
-
+        global_scale (array, optional): The per-input float32 scale used for x
+            ``"nvfp4"`` quantization if provided. Default: ``None``.
```
awni:
```suggestion
        global_scale (array, optional): The per-input float32 scale used for x
            with ``"nvfp4"`` quantization. Default: ``None``.
```

python/src/ops.cpp
```diff
@@ -5501,7 +5510,10 @@ void init_ops(nb::module_& m) {
...
+        global_scale_w (array, optional): The per-input float32 scale used for w
+            ``"nvfp4"`` quantization if provided. Default: ``None``.
```
awni:
```suggestion
        global_scale_w (array, optional): The per-input float32 scale used for w
            with ``"nvfp4"`` quantization. Default: ``None``.
```

mlx/ops.cpp
```diff
@@ -4380,6 +4427,15 @@ array qqmm(
   if (scales_w.has_value()) {
     inputs.push_back(*scales_w);
   }
+  // if
```
awni: stray comment?

mlx/ops.cpp
```diff
@@ -4380,6 +4427,15 @@ array qqmm(
...
+  if (global_scale_x.has_value()) {
+    // Stop gradient through global scales
+    inputs.push_back(stop_gradient(*global_scale_x));
+  }
+  if (global_scale_w.has_value()) {
+    // Stop gradient through global scales
+    inputs.push_back(stop_gradient(*global_scale_w));
+  }
```
awni: I'm wondering why you added it? Is it necessary?

mlx/ops.cpp
```diff
@@ -4532,19 +4589,28 @@ std::vector<array> fp_quantize(
...
     auto wq = reshape(w, {-1, group_size}, s);
     auto scales =
-        divide(max(abs(wq, s), -1, true, s), array(maxval, w.dtype()), s);
+        divide(max(abs(wq, s), -1, true, s), array(maxval, w.dtype()), s) *
+        scale_encode;
     if (group_size == 16) {
       // convert to e4m3
       scales = to_fp8(scales, s);
-      wq = divide(wq, from_fp8(scales, w.dtype(), s), s);
+      wq = divide(wq, from_fp8(scales, w.dtype(), s), s) * scale_encode;
```
awni: Common gotcha but all the ops here should be done explicitly on the inputs stream `s`. So instead of `a * b` do `multiply(a, b, s)`.

mlx/ops.cpp
```diff
@@ -4798,6 +4877,9 @@ array fp_dequantize(
        s](const std::vector<array>& inputs) mutable -> std::vector<array> {
     auto out = inputs[0];
     auto scales = inputs[1];
+    array inv_scale_enc = inputs.size() > 2
+        ? divide(inputs[2], array(448.0f * 6.0f, out_type), s)
+        : array(1.0f, out_type);
```
awni: Does it make sense to only make that in the `group_size == 16` case?

mlx/ops.cpp
```diff
@@ -4532,19 +4589,28 @@ std::vector<array> fp_quantize(
...
+    auto scale_encode =
+        inputs.size() > 1 ? 448.0f * 6.0f / inputs[1] : array(1.0f, float32);
     float maxval = (bits == 4) ? 6.0f : 448.0f;
     auto new_shape = w.shape();
     new_shape.back() = -1;
     auto wq = reshape(w, {-1, group_size}, s);
     auto scales =
-        divide(max(abs(wq, s), -1, true, s), array(maxval, w.dtype()), s);
+        divide(max(abs(wq, s), -1, true, s), array(maxval, w.dtype()), s) *
+        scale_encode;
```
awni: I would refactor that slightly so you only need to use and compute `scale_encode` in `group_size==16`.

mlx/primitives.cpp
```diff
@@ -3484,21 +3485,39 @@ std::vector<Shape> QQMatmul::output_shapes(const std::vector<array>& inputs) {
...
+    const std::vector<array>& primals, // non quantized x, non quantized w, if
+                                       // nvfp4 global_scale_x, global_scale_w
     const std::vector<array>& cotangents, // non quantized upstream grads
     const std::vector<int>& argnums,
     const std::vector<array>&) {
-  if (primals.size() != 2) {
-    throw std::runtime_error(
-        "[QQMatmul::vjp] Expected exactly 2 non-quantized primal inputs (x, w).");
+  bool is_nvfp4 = (mode_ == QuantizationMode::Nvfp4);
+  auto expected_size = is_nvfp4 ? 4 : 2;
```
awni: I thought the global scales were optional even for nvfp4?

nastya236: Thanks, it is optional, it is a bug

mlx/ops.cpp
nastya236: It's not necessary, you're right. Better to just return `zeros_like` in the vjp for inputs that don't need gradients (like in other primitives).

mlx/primitives.cpp
```diff
@@ -3484,21 +3485,30 @@ std::vector<Shape> QQMatmul::output_shapes(const std::vector<array>& inputs) {
...
+
   std::vector<array> vjps;
   auto& cotan = cotangents[0];
   auto& s = stream();
   // primal[1] -- non quantized w (N, K)
   // primal[0] -- non quantized activations (M, K)
   // cotan -- non quantized grads (M, N)
   auto qmode = quantization_mode_to_string(mode_);
+  std::optional<array> cotan_amax = (primals.size() == 4)
+      ? std::make_optional(astype(max(abs(cotan, s), s), float32))
```
awni: `astype` should also be on the stream `s`.

mlx/backend/cuda/quantized/qqmm_impl.cpp
```diff
@@ -41,10 +36,22 @@ void qqmm_impl(
       1, // batch_count
       0, // a_batch_stride
       0, // b_batch_stride
-      out_dtype,
+      out.dtype(),
       qmode);
 
-  qqmm.run(encoder, out, a, b, a_scale, b_scale, alpha);
+  if (scalars.uses_device_pointers()) {
```
awni: The name of that method is a little confusing. Maybe it would make more sense to call it `has_values()` or something?

PR #3023, opened 2026-01-20
Faster two pass sdpa

awni:
This PR does two things:
- Adds an option to use more than 32 blocks in the 2pass vector SDPA (it is set via function constant)
- Changes the work distribution so that one thread group is responsible for all the queries for a given key to maximize locality 

I tuned the block sizes on m3u, m4m and m5

### DIFF
--- a/mlx/backend/metal/kernels/sdpa_vector.h
+++ b/mlx/backend/metal/kernels/sdpa_vector.h
@@ -215,7 +215,7 @@ template <typename T, int D, int V = D>
 
   // Adjust positions
   const int kv_head_idx = tid.x;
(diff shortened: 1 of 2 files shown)


awni:
Here's the pre-post on an M3 Ultra with KV heads = 8, head dim = 128, varying sequence length and gqa factor:
Screenshot 2026-01-22 at 12 40 28 PM


Other cases I tested have similar results and I think the tuning is mostly generalizable (as I tuned the block sizes for head dim 128 per machine).

awni:
Closes https://github.com/ml-explore/mlx-lm/issues/763

PR #2972, opened 2026-01-07
WIP Windows CUDA support

dhiltgen:
## Proposed changes

This change partially addresses windows build issues in MLX.  The remaining errors are related to `add_kernel_node` which may require a deeper change to keep MSVC+NVCC happy on windows.

To repro, on a windows system, install:
- Visual Studio 2022
- [CUDA 13.0](https://developer.nvidia.com/cuda-13-0-0-download-archive)
- [cuDNN](https://developer.nvidia.com/cudnn-downloads)

I've been building with:
```
$env:CUDNN_INCLUDE_PATH="C:\Program Files\NVIDIA\CUDNN\v9.15\include\13.0"
$env:CUDNN_LIBRARY_PATH="C:\Program Files\NVIDIA\CUDNN\v9.15\lib\13.0\x64"

cmake -B build -DMLX_BUILD_CUDA=on -DMLX_BUILD_GGUF=OFF -DBUILD_SHARED_LIBS=ON -DMLX_BUILD_PYTHON_STUBS=OFF -DMLX_BUILD_METAL=OFF  -DMLX_BUILD_TESTS=OFF .
cmake --build build --parallel 2>&1 | % ToString | Tee-Object build.log
```

## Checklist

Put an `x` in the boxes that apply.

- [X] I have read the [CONTRIBUTING](https://github.com/ml-explore/mlx/blob/main/CONTRIBUTING.md) document
- [X] I have run `pre-commit run --all-files` to format my code / installed pre-commit prior to committing changes
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have updated the necessary documentation (if needed)

PR #3024, opened 2026-01-20
win: symbol exports and minor fixes

dhiltgen:
## Proposed changes

This PR extracts the non-CUDA portions of #2972 for Windows support.

Test results (known GGUF library incompatibility on Windows)
```
(.venv) PS C:\Users\danie\code\mlx> .\build\tests\Release\tests.exe 
[doctest] doctest version is "2.4.12"
[doctest] run with "--help" for options
===============================================================================
C:\Users\danie\code\mlx\tests\load_tests.cpp(43):
TEST CASE:  test gguf

C:\Users\danie\code\mlx\tests\load_tests.cpp(43): ERROR: test case THREW exception: [save_gguf] Compile with MLX_BUILD_GGUF=ON to enable GGUF support.

===============================================================================
C:\Users\danie\code\mlx\tests\load_tests.cpp(97):
TEST CASE:  test gguf metadata

C:\Users\danie\code\mlx\tests\load_tests.cpp(97): ERROR: test case THREW exception: [save_gguf] Compile with MLX_BUILD_GGUF=ON to enable GGUF support.

===============================================================================
[doctest] test cases:  221 |  219 passed | 2 failed | 0 skipped
[doctest] assertions: 3045 | 3045 passed | 0 failed |
[doctest] Status: FAILURE!
```
## Checklist

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -155,6 +155,7 @@ target_link_options(mlx PUBLIC ${SANITIZER_LINK_FLAGS})
 
 if(MLX_BUILD_CUDA)
   enable_language(CUDA)
+  find_package(CUDAToolkit REQUIRED)
 endif()
 
 if(MLX_BUILD_METAL)
--- a/mlx/backend/cuda/CMakeLists.txt
+++ b/mlx/backend/cuda/CMakeLists.txt
@@ -202,7 +202,6 @@ FetchContent_MakeAvailable(nvtx3)
 target_link_libraries(mlx PUBLIC $<BUILD_INTERFACE:nvtx3-cpp>)
 
 # Make cuda runtime APIs available in non-cuda files.
-find_package(CUDAToolkit REQUIRED)
 target_include_directories(mlx PRIVATE ${CUDAToolkit_INCLUDE_DIRS})
 
 # Use cublasLt.
--- a/mlx/distributed/nccl/CMakeLists.txt
+++ b/mlx/distributed/nccl/CMakeLists.txt
@@ -9,24 +9,17 @@ if(MLX_BUILD_CUDA)
       STATUS
         "NCCL not found, using stubs. To run distributed with NCCL backend, install NCCL."
     )
-
-    include(ExternalProject)
-    ExternalProject_Add(
+    file(
+      DOWNLOAD
+      "https://raw.githubusercontent.com/NVIDIA/nccl/refs/tags/v2.27.5-1/src/nccl.h.in"
+      "${CMAKE_CURRENT_BINARY_DIR}/nccl.h")
+    add_library(nccl_stub OBJECT
+                ${CMAKE_CURRENT_SOURCE_DIR}/nccl_stub/nccl_stubs.cpp)
+    target_include_directories(
       nccl_stub
-      SOURCE_DIR "${CMAKE_CURRENT_SOURCE_DIR}/nccl_stub"
-      BUILD_COMMAND ${CMAKE_COMMAND} --build . --config Release
-      INSTALL_COMMAND "")
-    # Ensure nccl_stub is built before mlx compiles nccl.cpp (needs nccl.h)
-    add_dependencies(mlx nccl_stub)
-    set(NCCL_PATH
-        "${CMAKE_CURRENT_BINARY_DIR}/nccl_stub-prefix/src/nccl_stub-build/")
-    if(WIN32)
-      # On Windows, shared libraries produce .dll and .lib (import library)
-      target_link_libraries(mlx PRIVATE ${NCCL_PATH}/Release/nccl.lib)
-    else()
-      target_link_libraries(mlx PRIVATE ${NCCL_PATH}/libnccl.so)
-    endif()
-    target_include_directories(mlx PRIVATE ${NCCL_PATH})
+      PRIVATE ${CUDAToolkit_INCLUDE_DIRS}
+      PUBLIC ${CMAKE_CURRENT_BINARY_DIR})
+    target_link_libraries(mlx PRIVATE $<BUILD_INTERFACE:nccl_stub>)
   endif()
 else()
   target_sources(mlx PRIVATE ${CMAKE_CURRENT_SOURCE_DIR}/no_nccl.cpp)
--- a/mlx/distributed/nccl/nccl_stub/CMakeLists.txt
+++ b/mlx/distributed/nccl/nccl_stub/CMakeLists.txt
@@ -1,16 +1,1 @@
-cmake_minimum_required(VERSION 3.25)
 
-project(nccl LANGUAGES C CXX)
-
-file(
-  DOWNLOAD
-  "https://raw.githubusercontent.com/NVIDIA/nccl/refs/tags/v2.27.5-1/src/nccl.h.in"
-  "${CMAKE_CURRENT_BINARY_DIR}/nccl.h")
-
-add_library(nccl SHARED nccl_stubs.cpp)
-set_target_properties(nccl PROPERTIES SOVERSION 2)
-# On Windows, export all symbols so that an import library (.lib) is generated
-set_target_properties(nccl PROPERTIES WINDOWS_EXPORT_ALL_SYMBOLS ON)
-find_package(CUDAToolkit REQUIRED)
-target_include_directories(nccl PRIVATE ${CUDAToolkit_INCLUDE_DIRS})
-target_include_directories(nccl PRIVATE ${CMAKE_CURRENT_BINARY_DIR})


CMakeLists.txt
```diff
@@ -260,9 +269,14 @@ if(MLX_BUILD_CPU)
       GIT_REPOSITORY https://github.com/OpenMathLib/OpenBLAS.git
       GIT_TAG v0.3.28
       EXCLUDE_FROM_ALL)
+    # Save BUILD_SHARED_LIBS before OpenBLAS modifies it (OpenBLAS's CMake
+    # disables BUILD_SHARED_LIBS on MSVC when BUILD_STATIC_LIBS is ON)
+    set(_MLX_SAVED_BUILD_SHARED_LIBS ${BUILD_SHARED_LIBS})
```
zcbenz: This can be done with `block()` and `endblock()` instead.

mlx/random.cpp
```diff
@@ -83,12 +83,13 @@ array split(const array& key, int num, StreamOrDevice s /* = {} */) {
...
-// floating point types (fp16, bf16)
+// floating point types (fp16, bf16).
+// Access the bits_ member directly to avoid strict aliasing issues
+// that can cause incorrect optimizations on MSVC.
 template <typename T>
 T below_one() {
   T f = T(1.0);
-  uint16_t* m = (uint16_t*)&f;
-  *m -= 1;
+  f.bits_ -= 1;
```
zcbenz: This does not work on mac:

```
  /Users/runner/actions-runner/_work/mlx/mlx/mlx/random.cpp:92:4: error: member reference base type '__fp16' is not a structure or union
     92 |   f.bits_ -= 1;
        |   ~^~~~~~
  /Users/runner/actions-runner/_work/mlx/mlx/mlx/random.cpp:128:22: note: in instantiation of function template specialization 'mlx::core::random::below_one<__fp16>' requested here
    128 |         return array(below_one<float16_t>(), float32);
        |                      ^
```

CMakeLists.txt
```diff
@@ -224,6 +220,19 @@ if(WIN32)
     set(MLX_BUILD_GGUF OFF)
     # There is no prebuilt OpenBLAS distribution for MSVC.
     set(MLX_BUILD_BLAS_FROM_SOURCE ON)
+    # Prevent Windows.h from defining min/max macros that break
+    # std::min/std::max and CCCL headers (cuda::std::min, cuda::std::max).
+    target_compile_definitions(mlx PRIVATE NOMINMAX)
+    # Windows DLLs have a 65535 symbol export limit. We use explicit exports via
+    # MLX_API macro (__declspec(dllexport)) on public API functions only. This
+    # avoids exporting internal template instantiations.
+    if(BUILD_SHARED_LIBS)
```
zcbenz: Can you move the compilation options to `mlx/CMakeLists.txt`? The root `CMakeLists.txt` is a bit overloaded and windows compilation things are usually put there.

CMakeLists.txt
```diff
@@ -224,6 +220,19 @@ if(WIN32)
     set(MLX_BUILD_GGUF OFF)
     # There is no prebuilt OpenBLAS distribution for MSVC.
     set(MLX_BUILD_BLAS_FROM_SOURCE ON)
+    # Prevent Windows.h from defining min/max macros that break
+    # std::min/std::max and CCCL headers (cuda::std::min, cuda::std::max).
+    target_compile_definitions(mlx PRIVATE NOMINMAX)
```
zcbenz: We do not include `windows.h` in MLX unless it is unavoidable (currently only `load.cpp` uses it and we should fix that), so there should be no need of a global `NOMINMAX`.

mlx/CMakeLists.txt
```diff
@@ -22,16 +22,37 @@ target_sources(
...
+  target_compile_options(mlx PRIVATE $<$<COMPILE_LANGUAGE:CXX>:/bigobj>)
 endif()
 
 if(WIN32)
-  # Export symbols by default to behave like macOS/linux.
-  set_target_properties(mlx PROPERTIES WINDOWS_EXPORT_ALL_SYMBOLS TRUE)
+  # On Windows, we use explicit __declspec(dllexport) via MLX_API macro instead
+  # of WINDOWS_EXPORT_ALL_SYMBOLS to avoid exceeding the 65535 symbol export
+  # limit (template instantiations can generate many symbols).
+  set_target_properties(mlx PROPERTIES WINDOWS_EXPORT_ALL_SYMBOLS FALSE)
```
zcbenz: I think we can just delete this as it is the default value and exporting with `MLX_API` is a common practice.

mlx/CMakeLists.txt
```diff
@@ -22,16 +22,37 @@ target_sources(
 # Define MLX_VERSION only in the version.cpp file.
 add_library(mlx_version OBJECT ${CMAKE_CURRENT_SOURCE_DIR}/version.cpp)
 target_compile_definitions(mlx_version PRIVATE MLX_VERSION="${MLX_VERSION}")
+# mlx_version needs access to version.h for MLX_API export macro
```
zcbenz: I think it meant to be this?
```suggestion
# mlx_version needs access to mlx_export.h for MLX_API export macro
```

mlx/backend/cpu/reduce.cpp
```diff
@@ -419,12 +419,37 @@ void reduce_dispatch_sum_prod(
   if (rtype == Reduce::Sum) {
     if constexpr (std::is_integral_v<InT> && sizeof(InT) <= 4) {
       reduction_op<InT, int32_t, SumReduce>(in, out, axes, 0);
+    } else if constexpr (
+        std::is_same_v<InT, float16_t> || std::is_same_v<InT, bfloat16_t>) {
+      // Use float32 accumulator for half-precision types to avoid precision
+      // loss during reductions, then convert back to the output dtype
+      array temp_out(out.shape(), float32, nullptr, {});
+      temp_out.set_data(allocator::malloc(temp_out.nbytes()));
```
zcbenz: Allocating a temporary array would be bad for performance, I think we should fix `reduction_op` instead. It is probably going to be a relative larger change and maybe exclude it from this PR?

zcbenz:
Properly exporting public APIs on Windows requires adding a `MLX_API` macro before every public function, which is a common practice. I'm good with that but also /cc @awni @angeloskath @davidkoski since it would change how we define APIs.

mlx/random.cpp
dhiltgen: I couldn't find a good pattern for this one, so I backed it out.  As a result, there's one known test failure:
```
C:\Users\danie\code\mlx\tests\random_tests.cpp(244):
TEST CASE:  test random uniform

C:\Users\danie\code\mlx\tests\random_tests.cpp(368): ERROR: CHECK( abs(float(mean(out).item<bfloat16_t>()) - 0.5f) < 0.02 ) is NOT correct!
  values: CHECK( 0.244141 <  0.02 )
```

benchmarks/cpp/CMakeLists.txt
```diff
@@ -3,6 +3,14 @@ function(build_benchmark SRCFILE)
   set(target "${src_name}")
   add_executable(${target} ${SRCFILE})
   target_link_libraries(${target} PRIVATE mlx)
+  # On Windows, copy the mlx DLL to the executable directory for runtime loading
```
zcbenz, line 6: I hope there is a better solution but I'm good with what it is now. I have created a issue to track this: https://github.com/ml-explore/mlx/issues/3031.

mlx/random.cpp
zcbenz: I can reproduce this failure on Mac running CPP tests with CPU, and it is also failing on Windows. /cc @awni

benchmarks/cpp/CMakeLists.txt
dhiltgen: I didn't want to restructure how the binaries are written out with CMAKE_RUNTIME_OUTPUT_DIRECTORY  since that would impact all platforms.  I don't believe there's an "rpath" equivalent we can rely on, so somehow the DLLs all need to be in the PATH or "." relative to the binary.  We could probably get cmake to internally use the right PATH so the run of the test tool to get the list of tests would work, but then when the user tries to run the tests it would just exit immediately due to unmet dependencies if they didn't set the PATH manually, but it would be pretty confusing to most users why since Windows no longer pops up a dialog showing the details.

zcbenz: What do you think if we only change the output dir of binaries for Windows, and only when MLX is the top project?

```cmake
if(WIN32 AND (CMAKE_CURRENT_SOURCE_DIR STREQUAL CMAKE_SOURCE_DIR))
  set(CMAKE_LIBRARY_OUTPUT_DIRECTORY "${CMAKE_BINARY_DIR}/bin")
  set(CMAKE_RUNTIME_OUTPUT_DIRECTORY "${CMAKE_BINARY_DIR}/bin")
endif() 
```

It would make binaries path different on Windows, but I think it is still better than having multiple copies of `mlx.dll` and having to add a copy command for every executable target.

mlx/api.h
```diff
@@ -0,0 +1,31 @@
+// Copyright © 2024 Apple Inc.
```
awni, line 1: Can we change the name of this file to `api.h` or `mlx_api.h` if you prefer? `mlx_export.h` is conflated a bit with `export.h` which is for function exports.

tests/linalg_tests.cpp
```diff
@@ -349,8 +349,11 @@ TEST_CASE("test SVD factorization") {
 
   const auto A_again = matmul(matmul(U_slice, diag(S)), Vt);
 
+  // Note: When TF32 is enabled (default on CUDA/Metal for performance),
+  // matmul has ~1e-3 precision instead of ~1e-7. Use 1e-3 tolerance to
+  // accommodate this. The test verifies SVD correctness, not matmul precision.
```
awni: We usually run the tests with TF32 disabled. It's basically the main reason we have environment flags for that. Where did you notice this failing?

awni:
I think I'm ok with the MLX_API macro.  One thing I observed is that it is infront of most functions in the top-level headers. Is that necessary even for functions which we don't really intend to be part of the API?

zcbenz:
> Is that necessary even for functions which we don't really intend to be part of the API?

No, we only need to export functions that we expect users to explicitly use.

mlx/dtype.h
```diff
@@ -94,21 +95,21 @@ inline constexpr Dtype::Category integer = Dtype::Category::integer;
...
 inline uint8_t size_of(const Dtype& t) {
   return t.size();
 }
 
-Dtype::Kind kindof(const Dtype& t);
+MLX_API Dtype::Kind kindof(const Dtype& t);
 
 template <typename T>
-struct TypeToDtype {
+struct MLX_API TypeToDtype {
```
awni, line 112:
```suggestion
struct TypeToDtype {
```

mlx/event.h
```diff
@@ -5,11 +5,12 @@
 #include <memory>
 #include <stdexcept>
 
+#include "mlx/mlx_export.h"
```
awni:
```suggestion
```

mlx/event.h
```diff
@@ -5,11 +5,12 @@
...
 #include "mlx/stream.h"
 
 namespace mlx::core {
 
-class Event {
+class MLX_API Event {
```
awni:
```suggestion
class Event {
```

mlx/fence.h
```diff
@@ -24,7 +25,7 @@ namespace mlx::core {
  * `MLX_METAL_FAST_SYNCH=1`. Fast mode also requires Metal 3.2+ (macOS 15+,
  * iOS 18+).
  */
-class Fence {
+class MLX_API Fence {
```
awni:
```suggestion
class Fence {
```

mlx/fence.h
```diff
@@ -3,6 +3,7 @@
 #include <vector>
 
 #include "mlx/array.h"
+#include "mlx/mlx_export.h"
```
awni:
```suggestion
```

mlx/primitives.h
```diff
@@ -349,7 +350,7 @@ class ArgPartition : public UnaryPrimitive {
   int axis_;
 };
 
-class ArgReduce : public UnaryPrimitive {
+class MLX_API ArgReduce : public UnaryPrimitive {
```
awni, line 353:
```suggestion
class ArgReduce : public UnaryPrimitive {
```

mlx/primitives.h
```diff
@@ -45,7 +46,7 @@
 namespace mlx::core {
 
 // Abstract base class
-class Primitive {
+class MLX_API Primitive {
```
awni:
```suggestion
class Primitive {
```

mlx/primitives.h
```diff
@@ -7,6 +7,7 @@
 #include "mlx/array.h"
 #include "mlx/device.h"
 #include "mlx/io/load.h"
+#include "mlx/mlx_export.h"
```
awni:
```suggestion
```

mlx/scheduler.h
```diff
@@ -162,7 +163,7 @@ void Scheduler::enqueue(const Stream& stream, F&& f) {
   threads_[stream.index]->enqueue(std::forward<F>(f));
 }
 
-Scheduler& scheduler();
+MLX_API Scheduler& scheduler();
```
awni, line 166: Did you add that one for a specific reason? Like would some test fail if we remove it?

mlx/utils.h
```diff
@@ -136,7 +135,7 @@ inline int next_power_of_2(int n) {
 
 namespace env {
 
-int get_var(const char* name, int default_value);
+MLX_API int get_var(const char* name, int default_value);
```
awni: Probably shouldn't expose that one.

mlx/utils.h
```diff
@@ -83,23 +84,21 @@ inline Dtype result_type(const array& a, const array& b) {
...
-    int ndim,
-    const std::string& msg_prefix = "");
-
-std::ostream& operator<<(std::ostream& os, const Device& d);
-std::ostream& operator<<(std::ostream& os, const Stream& s);
-std::ostream& operator<<(std::ostream& os, const Dtype& d);
-std::ostream& operator<<(std::ostream& os, const Dtype::Kind& k);
-std::ostream& operator<<(std::ostream& os, array a);
+MLX_API int
+normalize_axis_index(int axis, int ndim, const std::string& msg_prefix = "");
```
awni, line 95:
```suggestion
int
normalize_axis_index(int axis, int ndim, const std::string& msg_prefix = "");
```

awni:
I went through and suggested we remove a few of the MLX_API. If they were needed for a test failure or something then feel free to disregard the suggestion.  So basically two comments:

- Let's change `mlx_export.h` to `api.h`
- Remove several of the `MLX_API` for stuff that's not really supposed to be part of the public API even though it's in the main headers .. there are a few cases we are a little sloppy there.

awni:
Otherwise LGTM! Thanks for the change!

mlx/dtype.h
dhiltgen: The template specializations are defined in dtype.cpp and called from Python bindings across the DLL boundary.

mlx/primitives.h
dhiltgen: Used directly by the C++ test suite (arg_reduce_tests.cpp)

mlx/scheduler.h
dhiltgen: Used directly by the C++ test suite (scheduler_tests.cpp)

mlx/utils.h
dhiltgen: Used directly by the C++ test suite (utils_tests.cpp)

awni: Gotcha - ok for those ones I might do a follow up PR to make the tests a little less intrusive. But it's also not a big deal so let's 🚢

mlx/primitives.h
dhiltgen: MLX_API used elsewhere in the file.

zcbenz:
@awni Would you need another review or are we good to merge?

awni:
Good to go on my end. Thanks for making the changes @dhiltgen. We can merge as soon as the tests clear.

PR #1515, opened 2024-10-23
Quantized SDPA

barronalex:
First pass at adapting @angeloskath's flash attention to support quantized keys and values.

Still needs some optimization work since it's currently faster to write out the `quantized_matmul`s rather than use this fused version.

E.g. 4 bit on M2 Ultra for `L=32768`:
```
Timing sdpa ... 2.51938 msec
Timing quant_sdpa ... 0.97137 msec
Timing attention ... 1.31419 msec
Timing quant_attention ... 0.92342 msec
```

PR #3026, opened 2026-01-20
Quantized SDPA

CC-Yeh:
## Proposed changes

Add Metal quantized SDPA vector kernels based on #1515

### Speedup vs fp16

speedup_main

TODO:
- [x] Support `Affine` and `NVFP4`
- [x] Adapt #3023 once it's merged.
- [x] Cleanup

What improve performance:
- Removed thread storage k, v to reduce register pressure (was waiting on synchronization).
- Fused computation with dequantization
- Tuned reading size ('uint16_t'/'uin32_t') for quantized `k`/`v`
- Manual unroll better than `clang` loop optimizer

## Checklist


awni:
The numbers seem quite good.. a little too good to be true 😅 

What's the difference between SDPA and Attention in the benchmark? Also what's the query sequence length used for the benchmark?

CC-Yeh:
> The numbers seem quite good.. a little too good to be true 😅

Totally agree, must be missing something 🤔

> What's the difference between SDPA and Attention in the benchmark? Also what's the query sequence length used for the benchmark?

Attention is a simple reference implementation built from `matmul` + `softmax` + `matmul` (Maybe too naive?).
SDPA uses `mx.fast.scaled_dot_product_attention`, which hits the `sdpa_vector_2pass` kernels when Lq ≤ 8 (this case).

The query sequence length here is 1 (q.shape = (1, 32, 1, 128)), so this benchmark is measuring the single-token decode case, where one new token attends to a long KV cache (L = 32768).

CC-Yeh:
@awni 
Fixed some bugs in dequantizing 8bit and benchmark(unneccessary dequantization steps).
Now the numbers make more sense 😃

awni:
So if I’m understanding correctly the fused implementation is slower in the quantized case than the unfused ops-based one?

CC-Yeh:
Fused SDPA is faster: `MXFP4` 15.33 ms vs 24.71 ms, and `MXFP8` 26.09 ms vs 46.48 ms to decode a single query.

awni:
Very nice!!

mlx/fast.cpp
```diff
@@ -844,6 +845,273 @@ array scaled_dot_product_attention(
...
+  auto qmode = string_to_quantization_mode(
+      mode, "quantized_scaled_dot_product_attention");
+  if (qmode == QuantizationMode::Affine) {
+    throw std::invalid_argument(
+        "[quantized_scaled_dot_product_attention] Only fp quantization modes are supported.");
+  }
+  if (qmode == QuantizationMode::Nvfp4) {
+    throw std::invalid_argument(
+        "[quantized_scaled_dot_product_attention] Mode 'nvfp4' is not supported for fast attention.");
+  }
```
awni: Why not nvfp4?

mlx/fast.cpp
```diff
@@ -844,6 +845,273 @@ array scaled_dot_product_attention(
...
+      throw std::invalid_argument(msg.str());
+    }
+  }
+
+  auto qmode = string_to_quantization_mode(
+      mode, "quantized_scaled_dot_product_attention");
+  if (qmode == QuantizationMode::Affine) {
+    throw std::invalid_argument(
+        "[quantized_scaled_dot_product_attention] Only fp quantization modes are supported.");
+  }
```
awni: Why not affine?

mlx/fast.cpp
CC-Yeh: It’s on the way! I just wanted to make sure the PR structure was okay first.

mlx/fast.cpp
awni: Btw not suggesting we necessarily do it. Maybe it's better to be more limited in the quants we support here. Maybe fp8, fp4 are fine to start?

For example I don't think it's necessary to support every bit width because in practice no-one will ever use 2, 3 for KV cache quantization.

awni:
@CC-Yeh I'm interested in this PR moving forward. Let me know if you have questions. Also no need to support everything on a first pass. I think doing one 8-bit (fp8 / int8) quant well for Metal / CUDA is already probably good enough to start.

mlx/fast.cpp
CC-Yeh: Added support

mlx/fast.cpp
CC-Yeh: Added initial support, still has more room for tuning bit 2/3/5/6

CC-Yeh:
@awni 

I’ve added the Metal paths for `mxfp4/8`, `nvfp4`, and `affine(2/3/4/5/6/8)` (affine is not optimized).
Further tuning likely needs validation on other machines. 

For the CUDA path (maybe next PR), Colab doesn’t support NVFP4, so would need help for that.

quant_sdpa_speedup_vs_seqlen

awni:
> affine(2/3/4/5/6/8)

What group sizes did you do for that? I"m not convinced we need broad support for bitwidth X group size. I expect bits < 4 to be used rarely if ever.

CC-Yeh:
> > affine(2/3/4/5/6/8)
> 
> What group sizes did you do for that? I"m not convinced we need broad support there. I expect < 4 to be used rarely.

What group sizes do you think we should support for affine? Currently it's templated so it can handle various      
  sizes, but I can limit the instantiations if there's a specific set that's practical.
  
```cpp
template <typename T, int D, QuantMode mode, int group_size, int bits>
[[kernel]] void quant_sdpa_vector_2pass_1(
```

awni:
Yes totally. I think it's good to keep it generic. But probably better to limit initial support and grow than vice versa.

I would maybe start with bits  = {4, 6, 8} and just group_size = 32. I think 32 is most flexible for the head dimension right?

CC-Yeh:
> Yes totally. I think it's good to keep it generic. But probably better to limit initial support and grow than vice versa.
> 
> I would maybe start with bits = {4, 6, 8} and just group_size = 32. I think 32 is most flexible for the head dimension right?

Limited the affine support.

Yeah, 32 is most flexible for head dim.

CC-Yeh:
Hey @awni 

Just fine-tuned the block sizes and GQA factors, and switched from template kernels to a function_constant approach to trade some cold-start latency for reduced binary size. Ready for review!

speedup_main

mlx/backend/metal/scaled_dot_product_attention.cpp
```diff
@@ -656,6 +656,8 @@ void quant_sdpa_vector_2pass(
   kname += get_type_string(q.dtype());
   kname += "_";
   kname += std::to_string(q.shape(-1));
+  kname += "_";
+  kname += std::to_string(q.shape(-1));
```
altaic, line 660: I think you meant `v.shape(-1)` here?

CC-Yeh: Yeah, in 2 pass kernels both values are the same.

CC-Yeh:
@awni

Just wanted to gently bump this PR in case it got buried. 
Happy to split it up or adjust anything if that helps with review.

Thanks!

CC-Yeh:
Looping in @angeloskath @jagrit06 for visibility.

Happy to split it up or adjust anything if that helps with review.

Thanks!

Thump604:
This is exciting — fusing KV dequantization into SDPA is exactly what large-context quantized inference needs.

I run Qwen3.5-122B (5-bit, 10B active MoE) daily on M2 Ultra 128GB with 192K enforced context. Long-context decode is the primary bottleneck once SpecPrefill handles TTFT. The 1.6–1.8x decode speedup at 32K KV cache would be transformative for my use case.

Happy to benchmark this on M2 Ultra once it's ready for testing — I can provide before/after numbers at 16K, 32K, 64K, and 128K KV cache lengths on the 122B model. Let me know if there's anything specific you'd like measured.

+1 for merge — this fills a significant performance gap vs llama.cpp for quantized long-context inference.

Thump604:
Tested on M2 Ultra (128GB, macOS 26.3.1).

Build succeeds and FP16 SDPA works, but all 30 quantized SDPA tests fail with:

```
RuntimeError: [metal::Device] Unable to load function quant_sdpa_vector_2pass_1_float_128_16
```

The issue is in `scaled_dot_product_attention.cpp` around line 660. The kernel name is built using `v.shape(-1)`, which gives the packed uint32 dimension (16 for 4-bit quantized 128-dim values) instead of the logical value dimension (128). The `.metal` instantiation uses the logical dimension, so the dispatch looks for `..._128_16` but the compiled kernel is `..._128_128`.

Happy to apply a fix if you'd like, or I can just run M2 Ultra benchmarks (head_dim=128 and 256, GQA, 32K-128K context) once you've patched it.

Nicolas-nwb:
Hi @CC-Yeh and @awni,

Been following this PR with interest and recently ported it locally to get quantized SDPA working on Gemma 4 31B dense. Wanted to share two findings that might help move things forward.

**Supporting @Thump604's report on `v.shape(-1)`**

I ran into the same dispatch bug described in https://github.com/ml-explore/mlx/pull/3026#issuecomment-4173476256. The existing inline review from @altaic on 2026-02-07 had flagged the same asymmetry; the reasoning that "both values are the same in 2-pass kernels" holds for the FP pass-2 path but not for the quantized dispatch: `v` is stored as `uint32` packed (8 values/uint32 at 4-bit, 4 values/uint32 at 8-bit), so `v.shape(-1)` returns the packed dim (e.g. 16 for head_dim=128 / 4-bit), while the `.metal` instantiations use the logical dim.

The fix is the one-liner @Thump604 hinted at:

```cpp
// mlx/backend/metal/scaled_dot_product_attention.cpp (quant_sdpa_vector_2pass)
-  kname += std::to_string(v.shape(-1));
+  kname += std::to_string(q.shape(-1));
```

In this code path `query_head_dim == value_head_dim` is already enforced by `use_fallback()`, so using `q.shape(-1)` gives the correct logical dim without changing behavior.

**Structural confirmation**: after applying the fix, `xcrun metal-readobj --symbols python/mlx/lib/mlx.metallib` shows compiled kernels with the logical dim naming (e.g. `quant_sdpa_vector_2pass_1_float16_t_128_128`). Without the fix, the dispatch path builds `..._128_16` which never matches a compiled symbol.

**Proposal: add `head_dim=512` support (Gemma 4 31B dense use case)**

Gemma 4 31B dense uses `head_dim=512` on its 10 `full_attention` layers (out of 60; the other 50 are sliding with head_dim=256, GQA 8:1 on the full-attention path). These are currently routed to the FP16 fallback, losing the KV-cache memory benefit on the layers that matter most for long-context reasoning.

The blocker is that naively instantiating `instantiate_quant_sdpa_vector(type, 512, 512)` makes the existing BD=4 layout use `elem_per_thread = D/BD = 128`, which spills registers and hangs the GPU on Apple Silicon. A dynamic BD in `quant_sdpa_vector_2pass_1_impl` fixes it:

```cpp
constexpr int BD = (D > 256) ? 8 : 4;
constexpr int BN = 32 / BD;
```

With BD=8, `elem_per_thread = 64` (same as D=256 with BD=4, well within limits). The score and output reductions need a small adjustment to walk strides by BD instead of hard-coded 4:

```cpp
// score reduction, after quad_sum(score):
for (int stride = 4; stride < BD; stride <<= 1) {
  score += simd_shuffle_xor(score, stride);
}

// output reduction:
for (int stride = BD; stride < 32; stride <<= 1) {
  val += simd_shuffle_xor(val, stride);
}
```

Plus `instantiate_sdpa_vector_aggregation(type, 512)` in both FP instantiations (pass-1 and pass-2): on Apple Silicon `select_sdpa_blocks()` always returns ≥32 blocks for head_dim≥128, so the 2-pass path is always taken; a missing pass-2 instantiation hangs the GPU silently rather than producing a clean "kernel not found" error.

And finally updating `QuantizedScaledDotProductAttention::use_fallback()` to include 512 alongside 64/128/256.

**Validation**

Build and testing on Apple M4 Max (128 GB, macOS 26.4):

I also wrote a dedicated C++ doctest (`tests/quantized_sdpa_tests.cpp`, 3 TEST_CASE / 68 assertions) that exercises the quantized dispatch path directly — since `quantized_scaled_dot_product_attention` isn't exposed in the Python bindings by this PR, Python-level tests aren't possible yet. The test covers head_dim ∈ {64, 128, 256, 512} × bits ∈ {4, 8} × dtype ∈ {float16, bfloat16} with GQA 8:1, comparing against a dequantize + matmul + softmax + matmul reference. All 68 assertions pass with both patches applied; max_err vs. reference stays well below conservative tolerances:

| head_dim | bits | max_err | atol |
|---------:|-----:|--------:|-----:|
|       64 |    4 |  0.0038 |  0.5 |
|      128 |    4 |  0.0020 |  0.5 |
|      256 |    4 |  0.0027 |  0.5 |
|  **512** |    4 |  0.0036 |  0.5 |
|       64 |    8 |  0.0028 |  0.1 |
|      128 |    8 |  0.0028 |  0.1 |
|      256 |    8 |  0.0028 |  0.1 |
|  **512** |    8 |  0.0022 |  0.1 |

A minimal `CHECK_NOTHROW(eval(out))` case at head_dim=128 / 4-bit is included specifically to reproduce the dispatch mismatch — without Patch 1 it raises `Unable to load function quant_sdpa_vector_2pass_1_float16_t_128_16` (the exact pattern @Thump604 reported).

Additional checks:

- `python/tests/test_fast_sdpa.py`: 14 passed, 1 pre-existing skip, 592 subtests. No regressions on the FP path.
- `python/tests/test_fast.py`: 22/22 passed.
- `xcrun metal-readobj --symbols mlx.metallib` confirms all 12 expected quantized kernels are present and correctly named (float/bfloat16/float16 × {64, 128, 256, 512}).

End-to-end sanity check: ran Gemma 4 31B dense MXFP4 on an image-description + thinking workload (512 generated tokens, GQA 8:1, head_dim=512 on 10 full-attention layers). Without the patches: either dispatch error or FP16 fallback. With both patches: the quantized kernel path is active and generation is coherent.

Full local diff is ~30 lines across 3 files plus the new test (~290 lines). Happy to post the full diffs here, or open a follow-up PR after this one merges — whichever is less work for you. Thanks for the great effort on this PR.

CC-Yeh:
@Nicolas-nwb, @Thump604 thanks for the suggestions!

Improvements pushed.

Nicolas-nwb:
Hi @CC-Yeh and @awni — we've been building on top of this PR for
transformer inference on Apple Silicon and wanted to share a few
additions from our fork that might be useful upstream.

**Multi-row dispatch (qsl up to 32)** — commit f0ac19b9 on our fork:
- Moves `qsl` from the threadgroup z-dim to the grid z-dim. Threadgroup
  becomes `(32, gqa_factor, 1)` = 256 threads (well under the 1024 hw
  limit); grid z becomes `blocks * qsl`. The kernel decodes
  `block_idx = tid.z / qsl` and `q_seq_idx = tid.z % qsl` via a new
  `function_constant(31)` for `q_seq_len`.
- Removes the `qsl * gqa <= 32` constraint from the `use_fallback` gate,
  allowing qsl up to 32. This unblocks speculative decoding verifier
  batches (e.g. GQA 8:1 with K=8 → qsl=9) without falling back to FP16.
- Adds `head_dim == 512` to the fast-path gate (instantiations were
  already present in `scaled_dot_product_attention.metal`).
- Backports dynamic `BD=8` head_dim selection to `sdpa_vector.h`.

**New tests added:**
- TEST 4: qsl ∈ {2, 4, 8, 16, 32} × D ∈ {128, 256, 512} × {fp16, bf16}
  with `CHECK_FALSE(use_fallback)` + `max_err < 0.06` vs FP32 ref.
- TEST 5: pure gate-logic contract for qsl=1..8, GQA 8:1, post-patch.

**Debug spy:** `MLX_DEBUG_QUANT_SDPA=1` env var logs q/k shapes and
KERNEL/FALLBACK decisions to stderr — useful for Swift integration tests.

The changes live on the `quantized-sdpa` branch of our forks:
- mlx (submodule): f0ac19b9
- mlx-swift: 509a729

Happy to split these into a follow-up PR if that makes review easier, or
to rebase on top of the final merged state of this PR. Let us know what
works best.

joelnishanth:
**Benchmarks: Quantized SDPA + TurboQuant modes on M3 Pro (18 GB)**

Built from the `quantized_sdpa` branch with @dedalien's TurboQuant3/4 modes (CC-Yeh/mlx#3) cherry-picked on top. All 12 quantized SDPA tests pass. Benchmark script measures single-token decode (qsl=1) with realistic GQA configs.

**Setup:** Apple M3 Pro, 18 GB, macOS 26, MLX `0.31.2.dev+f983ccad`

### Llama-3-8B config (D=128, GQA 4:1, 32 query / 8 KV heads)

| Context | FP16 | affine-4b | affine-8b | turbo3 | turbo4 |
|--------:|-----:|----------:|----------:|-------:|-------:|
| 1K | 0.224ms | 0.307ms (0.7x) | 0.322ms (0.7x) | 0.313ms (0.7x) | 0.332ms (0.7x) |
| 2K | 0.246ms | 0.193ms (1.3x) | 0.197ms (1.3x) | 0.403ms (0.6x) | 0.442ms (0.6x) |
| 4K | 0.301ms | 0.266ms (1.1x) | 0.270ms (1.1x) | 0.609ms (0.5x) | 0.378ms (0.8x) |
| 8K | 0.493ms | 0.359ms (1.4x) | 0.395ms (1.2x) | 0.487ms (1.0x) | 0.503ms (1.0x) |
| 16K | 0.881ms | 0.550ms (1.6x) | 0.603ms (1.5x) | 0.544ms (1.6x) | 0.864ms (1.0x) |
| 32K | 1.740ms | 0.946ms (1.8x) | 1.084ms (1.6x) | 0.943ms (1.8x) | 1.267ms (1.4x) |

### Gemma-4 config (D=256, GQA 6:1, 24 query / 4 KV heads)

| Context | FP16 | affine-4b | affine-8b | turbo3 | turbo4 |
|--------:|-----:|----------:|----------:|-------:|-------:|
| 1K | 0.170ms | 0.247ms (0.7x) | 0.267ms (0.6x) | 0.426ms (0.4x) | 0.657ms (0.3x) |
| 4K | 0.340ms | 0.487ms (0.7x) | 0.501ms (0.7x) | 0.975ms (0.3x) | 0.646ms (0.5x) |
| 8K | 0.598ms | 0.731ms (0.8x) | 0.734ms (0.8x) | 1.068ms (0.6x) | 0.794ms (0.8x) |
| 32K | 1.429ms | 2.272ms (0.6x) | 2.385ms (0.6x) | 3.798ms (0.4x) | 2.710ms (0.5x) |

### Memory compression (32K context)

| Mode | Llama (128MB fp16) | Gemma (128MB fp16) |
|------|-------------------:|-------------------:|
| affine-4b | 40.0MB (3.2x) | 40.0MB (3.2x) |
| affine-8b | 72.0MB (1.8x) | 72.0MB (1.8x) |
| turbo3 | 25.0MB (**5.1x**) | 24.5MB (**5.2x**) |
| turbo4 | 33.0MB (**3.9x**) | 32.5MB (**3.9x**) |

### Key takeaways

1. **Affine-4b is the speed winner for decode.** 1.8x speedup at 32K on Llama config — the fused dequant kernel clearly benefits from reduced memory bandwidth.

2. **TurboQuant turbo3 matches affine-4b speed at long context (32K)** while delivering **60% more compression** (5.1x vs 3.2x). The crossover happens around 16K tokens.

3. **Short-context overhead is real for TurboQuant** — the codebook lookup path has higher per-dispatch cost. This is consistent with @Landon-Molt's findings on mlx#3404 about codebook vs scalar dequant.

4. **Gemma-4 (D=256) shows weaker speedups across the board** — affine is slower than fp16 at all context lengths. This may be a tuning opportunity for the D=256 kernel specialization.

5. **TurboQuant's value proposition is memory, not speed.** For memory-constrained inference (long context on 8-16 GB devices), turbo3's 5x compression enables contexts that wouldn't fit in fp16. The speed cost is manageable above 8K tokens.

Happy to run additional configs (D=512 for Gemma-4 31B, different batch sizes, M-series comparisons). The benchmark script is at [joelnishanth/mlx-swift-turboquant](https://github.com/joelnishanth/mlx-swift-turboquant) and I'll push the Python version to a branch.

This PR + @dedalien's turbo modes together provide a compelling quantized attention stack. Would love to see it land.

— [Joel Nishanth](https://offlyn.ai) · [offlyn.AI](https://offlyn.ai)

RNT56:
Hi, I noticed this PR is currently out of date with `main`, so I tried a local rebase in case it helps:

https://github.com/RNT56/mlx/tree/support/3026-quantized-sdpa-rebase

It only resolves the current rebase conflict and keeps the existing quantized SDPA changes. I also ran `python.tests.test_quantized` locally and it passed.

No need to use this if you prefer a different path; just sharing it as a reference.

PhilipJohnBasile:
**Independent reproduction at `bfed86b` on Apple M5 Max** (macOS 27.0, build 26A5378n; Xcode 27.0 / 27A5209h, Apple clang 21.0.0).

`python/tests/test_quantized.py`: 37/37 passed (1,696 subtests), including all `test_quantized_sdpa*`.

I couldn't reproduce the reported "Unable to load function `quant_sdpa_vector_2pass_1_...`" failure. On this head, `QuantizedScaledDotProductAttention::use_fallback` checks the supported head-dim set *before* the Metal kernel lookup, so unsupported dimensions are routed away rather than attempting (and failing) a symbol load. My dispatch probe matched that: the instantiated dims `{64,128,256,512}` reach the fused kernel; other dims are routed to the fallback composition (correct output, no crash); asymmetric Q/K-vs-V shapes raise a `ValueError` rather than crashing.

Application check: Qwen3-32B-4bit, GQA 8:1, `q_len=1`, 8-bit KV -- FP16 argmax matched at every tested position across 512-32K context.

Happy to rerun on a rebased head or a requested head-dim/device matrix. Thanks for the work!

PhilipJohnBasile:
Measured data for this PR from an Apple M5 Max (128 GB), posted as promised over in [mlx-lm#1587](https://github.com/ml-explore/mlx-lm/issues/1587) — we benched this kernel head-to-head earlier this month, and that issue has since produced evidence that changes what this PR is *for*. Data first, framing second.

**Head-to-head (decode-shaped — scoping caveat below).** Qwen3-32B-4bit, head_dim 128 (which this PR's symbol table instantiates — kernel-not-fallback confirmed via the `hasattr` gate), two N=9 interleaved passes plus a real-model correctness pass; `group_size=64`, `bits=8`.

- **Correctness: clean.** 100% greedy-argmax parity across fp16 / unfused int8 composition / this PR's fused kernel on the correctness pass.
- **This kernel beats the unfused composition** — it is simply the better int8-KV implementation.
- **Neither beats fp16 decode on this model at any context tested** (fused: −1.23% kernel-micro to −4.07% decode-sim at 16K projection; best case a tie). The byte accounting says why: on a 4-bit dense 32B, KV is only ~19% of decode-step reads — weights dominate — so halving KV bandwidth can't buy net decode speed here. Quality of int8-KV storage itself measured separately and clean (greedy-argmax agreement vs fp16 *rising* with context to 99.9% at 16K, perplexity within 0.4%; public receipts in [trailbrake's README](https://github.com/PhilipJohnBasile/trailbrake#readme)).

**Scoping caveat, stated precisely: every number above is `query_sequence_length = 1`.** Nothing in our data exercises qsl > 8 — which matters because of the next point.

**What changed since this PR stalled: it now has a live consumer, but for a different reason than decode speed.** mlx-lm#1587 established — two independent rigs, pre-registered discriminator, mechanism pinned quantitatively (predicted scores-tensor delta 1.208 GB vs 1.216 GB measured) — that the unfused `quantized_matmul → softmax → quantized_matmul` path's *untiled scores matrix* causes a real prefill peak-memory inversion: quantized caches transiently cost **more** memory than fp16 on long prompts, up to +73% at 32K in one report, reproducing from an M1/16GB laptop to an M5 Max. The unfused path this PR replaces is the direct cause.

**But — verified against this PR's own diff — the kernel as written cannot fix that:** `QuantizedScaledDotProductAttention::use_fallback()` returns true at `query_sequence_length > 8` (credit @rajanshxrma's read on #1587, which we confirmed line-for-line), so a 2048-token prefill chunk falls back to exactly the unfused composition that causes the inversion. The May fork patch lifting the gate to qsl≈32 helps, but chunked prefill needs hundreds-to-thousands — and per the analysis on #1587 (including a correctness-verified blockwise prototype whose Python form is *worse* under lazy evaluation, so the tiling must live in a fused kernel), prefill-shaped quantized attention reads as a differently-tiled problem: a quantized flash-style forward, not a wider dispatch of this structure.

**So the honest framing is two separate work items:**
1. **This PR, revived roughly as-is: the decode path.** Correct at head_dim 128, strictly better than the unfused composition it replaces, and its value case is memory-capacity regimes (long context, batch, KV-dominated models) rather than dense-model decode speed. The thread history reads as maintainer bandwidth, not rejected design.
2. **A new, separately-scoped kernel: quantized flash-style prefill attention.** That one — not this PR — is what closes mlx-lm#1587. Conflating the two would leave that issue "fixed" by a merge that never fires on its reproduction.

The kernel itself benched exactly as designed at head_dim 128 — the fp16 decode losses above are the physics of this model's shape, not implementation quality. Raw run JSON available on request for either work item.

PR #3028, opened 2026-01-21
Fix/cholesky psd error handling

bjornefisk:
## Proposed changes

This PR addresses a correctness issue in the CPU Cholesky decomposition where non-positive semi-definite (non-PSD) matrices would silently fail without raising an error, potentially returning incorrect results to users.

The fix adds proper error handling for the case when LAPACK's `potrf` function returns `info > 0`, which indicates the input matrix is not positive semi-definite. The error message now clearly identifies which leading minor failed, helping users diagnose issues with their input matrices.

This resolves the TODO comment at line 47 in [mlx/backend/cpu/cholesky.cpp

## Checklist


zcbenz:
I think this is clearly not what the TODO means.

Issue #3031, opened 2026-01-21
Built DLLs in the same directory with EXEs on Windows


zcbenz:
When building on Windows with `-DBUILD_SHARED_LIBS=ON`, the dynamic libraries like `mlx.dll` are put under a sub-directory and the executables would not be able to find the DLLs. This can be work-arounded by copying DLLs to the dir of executables, but we should find a way to force cmake to build them in the same dir.

ccking870:
https://github.com/ml-explore/mlx/issues/3031#issue-3836462983

Issue #3020, opened 2026-01-20
[Enhancement] Link with prebuilt binaries of OpenBLAS on Windows


zcbenz:
Currently for Windows build we are building OpenBLAS from source code with `MLX_BUILD_BLAS_FROM_SOURCE=ON`, but OpenBLAS actually provides [prebuilt binaries for Windows on GitHub Releases](https://github.com/OpenMathLib/OpenBLAS/releases/tag/v0.3.31) which we can link with directly, and it would save a lot of build time.

PR #3036, opened 2026-01-22
Link with prebuilt OpenBLAS and fix shared libs build on Windows

zcbenz:
Close https://github.com/ml-explore/mlx/issues/3031.
Close https://github.com/ml-explore/mlx/issues/3020.

This PR links MLX with the prebuilt dynamic library of OpenBLAS on Windows instead of building it from source code, which can save a lot of build time. (Note that we can not do static-linking, check the comments for more.)

Which then requires changing the locations where libraries and executables are generated, as Windows requires the DLL and EXE files be in the same directory (as there is no equivalent of RPATH on Windows):
* For building: the DLL and EXE would be written in the root build dir (previously EXE are put in sub-dirs like `tests/test.exe`)
* For installing: the DLL would be put in the same dir with extension file (previously they are put in the `bin` sub-dir)

Note that the locations are only changed for Windows, and only when MLX is the top project, for projects that include MLX as a third party library there is nothing changed.

/cc @dhiltgen

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -40,7 +40,6 @@ option(MLX_METAL_DEBUG "Enhance metal debug workflow" OFF)
 option(MLX_ENABLE_X64_MAC "Enable building for x64 macOS" OFF)
 option(MLX_BUILD_GGUF "Include support for GGUF format" ON)
 option(MLX_BUILD_SAFETENSORS "Include support for safetensors format" ON)
-option(MLX_BUILD_BLAS_FROM_SOURCE "Build OpenBLAS from source code" OFF)
 option(MLX_BUILD_PYTHON_STUBS "Build stub files for python bindings" ON)
 option(MLX_METAL_JIT "Use JIT compilation for Metal kernels" OFF)
 option(MLX_USE_CCACHE "Use CCache for compilation cache when available" ON)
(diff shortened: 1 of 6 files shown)


CMakeLists.txt
```diff
@@ -249,20 +251,25 @@ if(MLX_BUILD_CPU)
...
+    # link with the dynamic library, the prebuilt binaries were built with MinGW
+    # so static-linking would require linking with MinGW's runtime.
     FetchContent_Declare(
       openblas
-      GIT_REPOSITORY https://github.com/OpenMathLib/OpenBLAS.git
-      GIT_TAG v0.3.28
-      EXCLUDE_FROM_ALL)
-    set(BUILD_STATIC_LIBS ON) # link statically
-    set(NOFORTRAN ON) # msvc has no fortran compiler
+      URL "https://github.com/OpenMathLib/OpenBLAS/releases/download/v0.3.31/OpenBLAS-0.3.31-x64.zip"
```
dhiltgen, line 261: fwiw, I've also been playing around with trying to get things building on Windows ARM64, which needed patching of the OpenBLAS cmake files to get all the variables set up properly for LLVM-MinGW.

zcbenz: OpenBLAS provides binaries for Windows ARM64 (the woa64 files), can we just link with that?

PR #3039, opened 2026-01-22
CUDA gather mv

angeloskath:
Adds a CUDA implementation for `gather_mv`, the `gather_vm` is missing but using this and @zcbenz grouped mm for the sorted version we can run unquantized MoEs on CUDA.

```
$ mlx_lm.benchmark --model Qwen/Qwen3-30B-A3B -p 2048 -g 128
Running warmup..
Timing with prompt_tokens=2048, generation_tokens=128, batch_size=1.
Trial 1:  prompt_tps=2101.020, generation_tps=32.027, peak_memory=61.821
Trial 2:  prompt_tps=2088.784, generation_tps=32.078, peak_memory=61.821
Trial 3:  prompt_tps=2101.889, generation_tps=32.072, peak_memory=61.822
Trial 4:  prompt_tps=2095.918, generation_tps=32.046, peak_memory=61.822
Trial 5:  prompt_tps=2112.402, generation_tps=30.730, peak_memory=61.822
Averages: prompt_tps=2100.002, generation_tps=31.791, peak_memory=61.822
```

### DIFF
--- a/mlx/backend/cuda/matmul.cpp
+++ b/mlx/backend/cuda/matmul.cpp
@@ -356,18 +356,15 @@ void GatherMM::eval_gpu(const std::vector<array>& inputs, array& out) {
 
   auto [transposed_a, lda, a_] = check_transpose(encoder, s, a);
   auto [transposed_b, ldb, b_] = check_transpose(encoder, s, b);
-  if (M == 1) {
-    if (transposed_b) {
-      gather_mv(b_, a_, rhs_indices, lhs_indices, out, N, K, encoder);
-      return;
-    }
+  auto use_gemv = cu::can_use_gemv(M, N, K, transposed_a, transposed_b);
+  if (M == 1 && use_gemv) {
+    gather_mv(b_, a_, rhs_indices, lhs_indices, out, N, K, encoder);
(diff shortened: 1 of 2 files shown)


mlx/backend/cuda/matmul.cpp
```diff
@@ -354,6 +354,22 @@ void GatherMM::eval_gpu(const std::vector<array>& inputs, array& out) {
     return;
   }
 
+  auto [transposed_a, lda, a_] = check_transpose(encoder, s, a);
+  auto [transposed_b, ldb, b_] = check_transpose(encoder, s, b);
+  if (M == 1) {
+    if (transposed_b) {
+      gather_mv(b_, a_, rhs_indices, lhs_indices, out, N, K, encoder);
```
awni: There is a function `can_use_gemv` which determines if it's safe to route to the custom gemv vs cublas matmul.

I think we might need to use that here to gaurd against unsupported cases (reduction dimension has to be a multiple of 32) or alternatively fix the gemv to support non-32 multiples.

(I think it's fine to throw for now in the unsupported case and we can do it in a follow on).

PR #3046, opened 2026-01-23
Allow take on empty array when it makes sense

awni:
This also matches the behavior in numpy and closes https://github.com/ml-explore/mlx-lm/issues/798


awni:
There is a flaky test in our gemv.. slightly concerning.

PR #3049, opened 2026-01-23
Build and test python package on Windows CI

zcbenz:
Also fixed a few failed tests that caused by `long` being 32bit on Windows.


python/tests/test_memory.py
```diff
@@ -18,8 +18,8 @@ def test_memory_info(self):
         self.assertEqual(mx.set_cache_limit(old_limit), old_limit)
 
         old_limit = mx.set_memory_limit(10)
-        self.assertTrue(mx.set_memory_limit(old_limit), 10)
-        self.assertTrue(mx.set_memory_limit(old_limit), old_limit)
+        self.assertEqual(mx.set_memory_limit(old_limit), 10)
```
awni, line 21: Interesting.. I wonder what that was testing before 😅

PR #3050, opened 2026-01-23
Use C++20

zcbenz:
We only used C++17 features but to make some code (mainly `dispatch_xxx` utils) compile with MSVC we have to enable C++20.

### DIFF
--- a/tests/einsum_tests.cpp
+++ b/tests/einsum_tests.cpp
@@ -5,6 +5,15 @@
 
 using namespace mlx::core;
 
+namespace std {
+
+// Required to make doctest compile.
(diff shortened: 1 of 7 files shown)


tests/einsum_tests.cpp
```diff
@@ -5,6 +5,15 @@
 
 using namespace mlx::core;
 
+namespace std {
+
+// Required to make doctest compile.
+ostream& operator<<(ostream& os, const vector<vector<int>>&) {
+  return os;
```
angeloskath, line 12: Hmm and this shouldn't print anything?

zcbenz: We should print the content of vector but it is only used when the test fails so I'm getting lazy here.

PR #3055, opened 2026-01-24
GPU discovery

dhiltgen:
## Proposed changes

Enhanced GPU discovery leveraging existing device_info.  In particular for CUDA this tries to use NVML (if present) which provides an overall system view of VRAM consumption compared to cudaMemGetInfo which is scoped just to the current process when calculating consumed memory.  This also exposes PCI IDs and UUIDs for devices which can help correlate with other APIs for deduplication or additional queries.

### Simple example program
```python
#!/usr/bin/env python3
import mlx.core as mx

def main():
    print("="*70)
    print("MLX GPU Device Discovery Demo")
    print("="*70)

    # Check backends
    print(f"\nCUDA available: {mx.cuda.is_available()}")
    print(f"Metal available: {mx.metal.is_available()}")

    # Device count
    count = mx.device_count()
    print(f"\nTotal GPU devices: {count}")

    if count == 0:
        print("No GPU devices found.")
        return

    # Enumerate devices
    for i in range(count):
        info = mx.device_info(i)
        print(f"\n{'='*70}")
        print(f"Device {i}")
        print(f"{'='*70}")
        for key, value in sorted(info.items()):
            if 'memory' in key and isinstance(value, int):
                print(f"  {key:30s}: {value:15,} bytes ({value/(1024**3):.2f} GB)")
            else:
                print(f"  {key:30s}: {value}")

if __name__ == "__main__":
    main()
```

### Output on a dual GPU linux system
```
% ./test_device_discovery.py 
MLX GPU Device Discovery Demo

CUDA available: True
Metal available: False

Total GPU devices: 2

Device 0
  architecture                  : sm_120
  compute_capability_major      : 12
  compute_capability_minor      : 0
  device_name                   : NVIDIA RTX PRO 6000 Blackwell Workstation Edition
  free_memory                   : 101,384,192,000 bytes (94.42 GB)
  pci_bus_id                    : 0000:43:00.0
  total_memory                  : 102,641,958,912 bytes (95.59 GB)
  uuid                          : GPU-e0be9369-843f-a56c-2def-efc2848da030

Device 1
  architecture                  : sm_120
  compute_capability_major      : 12
  compute_capability_minor      : 0
  device_name                   : NVIDIA RTX PRO 6000 Blackwell Workstation Edition
  free_memory                   : 101,384,192,000 bytes (94.42 GB)
  pci_bus_id                    : 0000:6f:00.0
  total_memory                  : 102,641,958,912 bytes (95.59 GB)
  uuid                          : GPU-c681a258-86c7-673f-3f2e-77c50fd2ae0c
```

### Output on a Mac
```
% test_device_discovery.py 
MLX GPU Device Discovery Demo

CUDA available: False
Metal available: True

Total GPU devices: 1

Device 0
  architecture                  : applegpu_g15s
  device_name                   : Apple M3 Max
  max_buffer_length             : 77309411328
  max_recommended_working_set_size: 103079215104
  memory_size                   : 137,438,953,472 bytes (128.00 GB)
  resource_limit                : 499000
```

## Checklist

### DIFF
--- a/mlx/backend/cuda/device_info.cpp
+++ b/mlx/backend/cuda/device_info.cpp
@@ -227,11 +227,6 @@ bool is_available() {
   return true;
 }
 
-const std::unordered_map<std::string, std::variant<std::string, size_t>>&
-device_info(int device_index) {
-  return device_info_impl(device_index);
-}
-
 } // namespace cu
 
 } // namespace mlx::core
--- a/mlx/backend/cuda/eval.cpp
+++ b/mlx/backend/cuda/eval.cpp
@@ -3,7 +3,6 @@
 #include "mlx/backend/gpu/eval.h"
 #include "mlx/backend/cuda/allocator.h"
 #include "mlx/backend/cuda/device.h"
-#include "mlx/backend/gpu/device_info.h"
 #include "mlx/primitives.h"
 #include "mlx/scheduler.h"
 
--- a/mlx/backend/metal/CMakeLists.txt
+++ b/mlx/backend/metal/CMakeLists.txt
@@ -139,5 +139,5 @@ endif()
 
 add_subdirectory(${CMAKE_CURRENT_SOURCE_DIR}/kernels)
 
-target_compile_definitions(
-  mlx PRIVATE METAL_PATH="${MLX_METAL_PATH}/mlx.metallib" MLX_USE_METAL)
+target_compile_definitions(mlx
+                           PRIVATE METAL_PATH="${MLX_METAL_PATH}/mlx.metallib")
--- a/mlx/backend/metal/eval.cpp
+++ b/mlx/backend/metal/eval.cpp
@@ -1,7 +1,6 @@
 // Copyright © 2023-2024 Apple Inc.
 #include <memory>
 
-#include "mlx/backend/gpu/device_info.h"
 #include "mlx/backend/gpu/eval.h"
 #include "mlx/backend/metal/device.h"
 #include "mlx/backend/metal/utils.h"
--- a/mlx/backend/metal/allocator.cpp
+++ b/mlx/backend/metal/allocator.cpp
@@ -1,5 +1,6 @@
 // Copyright © 2023-2024 Apple Inc.
 #include "mlx/backend/metal/allocator.h"
+#include "mlx/backend/gpu/device_info.h"
 #include "mlx/backend/metal/metal.h"
 #include "mlx/backend/metal/resident.h"
 #include "mlx/memory.h"
@@ -43,16 +44,17 @@ MetalAllocator::MetalAllocator()
           }),
       residency_set_(device_) {
   auto pool = metal::new_scoped_memory_pool();
-  auto memsize = std::get<size_t>(device_info().at("memory_size"));
+  const auto& info = gpu::device_info(0);
+  auto memsize = std::get<size_t>(info.at("memory_size"));
   auto max_rec_size =
-      std::get<size_t>(device_info().at("max_recommended_working_set_size"));
-  resource_limit_ = std::get<size_t>(device_info().at("resource_limit"));
+      std::get<size_t>(info.at("max_recommended_working_set_size"));
+  resource_limit_ = std::get<size_t>(info.at("resource_limit"));
   block_limit_ = std::min(1.5 * max_rec_size, 0.95 * memsize);
   gc_limit_ = std::min(static_cast<size_t>(0.95 * max_rec_size), block_limit_);
   max_pool_size_ = block_limit_;
   device(mlx::core::Device::gpu)
       .set_residency_set(residency_set_.mtl_residency_set());
-  bool is_vm = std::get<std::string>(device_info().at("device_name")) ==
+  bool is_vm = std::get<std::string>(info.at("device_name")) ==
       "Apple Paravirtual device";
   if (is_vm) {
     return;
@@ -249,8 +251,8 @@ size_t get_memory_limit() {
   return metal::allocator().get_memory_limit();
 }
 size_t set_wired_limit(size_t limit) {
-  if (limit > std::get<size_t>(metal::device_info().at(
-                  "max_recommended_working_set_size"))) {
+  if (limit > std::get<size_t>(
+                  gpu::device_info(0).at("max_recommended_working_set_size"))) {
     throw std::invalid_argument(
         "[metal::set_wired_limit] Setting a wired limit larger than "
         "the maximum working set size is not allowed.");
--- a/mlx/backend/metal/device_info.cpp
+++ b/mlx/backend/metal/device_info.cpp
@@ -1,6 +1,9 @@
 // Copyright © 2026 Apple Inc.
 
+#include <sys/sysctl.h>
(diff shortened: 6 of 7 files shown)


mlx/backend/cuda/cuda.cpp
```diff
@@ -2,10 +2,207 @@
...
+const std::unordered_map<std::string, std::variant<std::string, size_t>>&
+device_info(int device_index) {
+  // Cache per device (static info doesn't change, memory refreshed)
+  static std::unordered_map<
+      int,
+      std::unordered_map<std::string, std::variant<std::string, size_t>>>
+      cache;
+  static std::mutex cache_mutex;
+
+  std::lock_guard<std::mutex> lock(cache_mutex);
```
zcbenz: We can avoid using mutex by doing something like below (a common pattern we use in MLX):

```c++
auto devices_info() {
  // return all devices info.
}

const auto& device_info(int index) {
  static auto all_devices = devices_info();
  return all_devices.at(index);
}
```

It can also avoid checking initialized state of NVML since the code will only need to read once.

zcbenz:
What should we do with `mx.metal.device_info`, leave it be or deprecate it?

awni:
Yes I do think we should deprecate `mx.metal.device_info` and pull it into the high level `mx.device_info`.

awni:
This is a really nice addition. I also agree with the above comments from @angeloskath regarding the organization / API. Otherwise looking forward to merging it!

dhiltgen:
Thanks for the pointers!  I'll take a refactoring pass to adjust accordingly.

mlx/backend/cuda/cuda.cpp
```diff
@@ -2,10 +2,237 @@
...
 #include "mlx/backend/cuda/cuda.h"
 
+#include <cuda_runtime.h>
+#include <string>
+#include <unordered_map>
+#include <variant>
+#include <vector>
+
+#ifdef _WIN32
+#include <windows.h>
```
zcbenz: You can use the dlfcn APIs on Windows, we are already using a compatibility layer (dlfcn-win32).

mlx/backend/gpu/device_info.cpp
```diff
@@ -0,0 +1,39 @@
...
+#include "mlx/backend/metal/metal.h"
+#elif defined(MLX_USE_CUDA)
+#include <cuda_runtime.h>
+#include "mlx/backend/cuda/cuda.h"
+#endif
+
+namespace mlx::core::gpu {
+
+int device_count() {
+#if defined(MLX_USE_METAL)
```
zcbenz: We don't use macros to define per-backend code, the pattern we use is:

```c++
// mlx/backend/metal/device_info.cpp

#include "mlx/backend/metal/metal.h"

namespace mlx::core::gpu {

int device_count() {
  return 1;
}

auto device_info(int device_index) {
  return ...;
}

}
```

```c++
// mlx/backend/cuda/device_info.cpp

#include <cuda_runtime.h>

namespace mlx::core::gpu {

int device_count() {
  return ...;
}


auto device_info(int device_index) {
  return ...;
}

}
```

```c++
// mlx/backend/no_gpu/device_info.cpp
namespace mlx::core::gpu {

int device_count() {
  return 0;
}


auto device_info(int device_index) {
  throw ...;
}

}
```

And we would also need a `cpu::` version in `mlx/backend/no_cpu/device_info.cpp` as the cpu backend could be turned off.

(Note that the `MLX_USE_CUDA` macro is a special case, we don't have a good alternative and it is not used for used for dispatching backend code.)

mlx/device.cpp
```diff
@@ -1,4 +1,4 @@
-// Copyright © 2023 Apple Inc.
+// Copyright © 2023-2025 Apple Inc.
```
zcbenz: It is 2026 now 😄

mlx/backend/cuda/eval.cpp
```diff
@@ -3,18 +3,14 @@
 #include "mlx/backend/gpu/eval.h"
 #include "mlx/backend/cuda/allocator.h"
 #include "mlx/backend/cuda/device.h"
-#include "mlx/backend/gpu/available.h"
+#include "mlx/backend/gpu/device_info.h"
```
zcbenz: Can be removed.

mlx/backend/metal/CMakeLists.txt
```diff
@@ -138,5 +139,5 @@ endif()
 
 add_subdirectory(${CMAKE_CURRENT_SOURCE_DIR}/kernels)
 
-target_compile_definitions(mlx
-                           PRIVATE METAL_PATH="${MLX_METAL_PATH}/mlx.metallib")
+target_compile_definitions(
+  mlx PRIVATE METAL_PATH="${MLX_METAL_PATH}/mlx.metallib" MLX_USE_METAL)
```
zcbenz: `MLX_USE_METAL` is no longer needed.

mlx/backend/cuda/device_info.cpp
```diff
@@ -0,0 +1,237 @@
...
+} // namespace gpu
+
+namespace cu {
+
+bool is_available() {
+  return true;
+}
+
+const std::unordered_map<std::string, std::variant<std::string, size_t>>&
+device_info(int device_index) {
```
zcbenz: We don't need a `cu::device_info`, the `metal::device_info` is also going to be removed.

mlx/backend/metal/eval.cpp
```diff
@@ -1,7 +1,7 @@
 // Copyright © 2023-2024 Apple Inc.
 #include <memory>
 
-#include "mlx/backend/gpu/available.h"
+#include "mlx/backend/gpu/device_info.h"
```
zcbenz: Can be removed.

mlx/backend/no_gpu/eval.cpp
```diff
@@ -2,15 +2,11 @@
 
 #include <stdexcept>
 
-#include "mlx/backend/gpu/available.h"
+#include "mlx/backend/gpu/device_info.h"
```
zcbenz, line 5: Can be removed.

mlx/backend/cpu/device_info.cpp
```diff
@@ -0,0 +1,59 @@
...
+#include <sys/sysctl.h>
+#endif
+
+namespace mlx::core::cpu {
+
+namespace {
+
+// Get CPU architecture string
+std::string get_cpu_architecture() {
+#if defined(__aarch64__) || defined(__arm64__)
```
zcbenz, line 15: This implementation returns the architecture that the program was compiled against rather than the real hardware. For example a x64 binary running on arm64 platform through a compatibility layer would report the CPU architecture as x64.

I don't think it is an important thing and not needed to be fixed in this PR, but I think we should exclude it from `device_info`'s result, in case users rely on the behavior which would make it hard to change.

dhiltgen: I'll work on a fast-follow to improve this.

dhiltgen: #3068

PR #3057, opened 2026-01-24
Hide symbols by default for mac/linux

zcbenz:
Change the default symbol visibility to hidden for mac/linux, and require using `MLX_API` macro for exporting symbols. This makes all platforms use the same symbol visibility, which can reduce the cases that code works for mac/linux but fails for windows.

Hiding symbols by default also has some benefits:
* Smaller binary size: `libmlx.so` shrinks from 81M to 79M.
* Faster loading of shared library.

### DIFF
--- a/mlx/backend/cpu/encoder.h
+++ b/mlx/backend/cpu/encoder.h
@@ -12,7 +12,7 @@ namespace mlx::core::cpu {
 // Number of dispatches per scheduler task
 constexpr int DISPATCHES_PER_TASK = 10;
 
-struct CommandEncoder {
+struct MLX_API CommandEncoder {
   CommandEncoder(Stream stream) : stream_(stream) {}
 
   CommandEncoder(const CommandEncoder&) = delete;
@@ -62,6 +62,6 @@ struct CommandEncoder {
   int num_ops_{0};
 };
 
-CommandEncoder& get_command_encoder(Stream stream);
+MLX_API CommandEncoder& get_command_encoder(Stream stream);
 
 } // namespace mlx::core::cpu
--- a/mlx/backend/metal/device.h
+++ b/mlx/backend/metal/device.h
@@ -20,7 +20,7 @@ using MTLFCList =
 
 struct DeviceStream;
 
-struct CommandEncoder {
+struct MLX_API CommandEncoder {
   explicit CommandEncoder(DeviceStream& stream);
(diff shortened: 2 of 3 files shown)


bear24rw:
This unfortunately breaks my extension where I was using a bunch of non-public helpers. These were not officially public to begin with but it was very convenient that I could still access them. Whats the best path forward for my use case? Seems my two options now are either maintain a fork of mlx or copy a lot of code into my extension. In my ideal world everything would be accessible and all header files would be shipped in the .whl but some would be under an `internal/` namespace indicating it's technically private and prone to change without warning (this is similar to what https://github.com/ocornut/imgui does).

```
Undefined symbols for architecture arm64:
        "mlx::core::gather_qmm_rhs(mlx::core::array const&, mlx::core::array const&, mlx::core::array const&, std::__1::optional<mlx::core::array> const&, mlx::core::array const&, mlx::core::array&, bool, int, int, int, int, int, mlx::core::metal::Device&, mlx::core::Stream const&, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>>)"
        "mlx::core::broadcast_arrays(mlx::core::array const&, mlx::core::array const&, std::__1::variant<std::__1::monostate, mlx::core::Stream, mlx::core::Device>)"
        "mlx::core::swapaxes_in_eval(mlx::core::array const&, int, int)"
        "mlx::core::contiguous_copy_gpu(mlx::core::array const&, mlx::core::Stream const&)"
        "mlx::core::validate_mode_with_type(std::__1::basic_string_view<char, std::__1::char_traits<char>>, mlx::core::array const&, std::__1::optional<mlx::core::array> const&, std::__1::optional<mlx::core::Dtype>, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char>> const&)"
        "mlx::core::quantization_mode_to_string(mlx::core::QuantizationMode)"
        "mlx::core::quantization_params_from_mode(mlx::core::QuantizationMode
```

zcbenz:
@bear24rw I think for now we can just export those functions with `MLX_API` since you are using them in extension — we are already exporting a few non-public APIs for users' convenience.

bear24rw:
@zcbenz I think it might be a bit of game of whack a mole since those functions I listed are just what I happened to be using today but my extension is still in development. I've decided to just go down the fork mlx route with the hope that one day it would be upstreamed anyway :)

PR #3058, opened 2026-01-24
Fix double precision for exp, sin, cos, erf, erfinv on CPU

### LINKED ISSUE #3047: [BUG] [CPU] Some ops fallback to fp32 implementation even fro fp64
Some ops (exp, sin, cos, erf, erfinv) use the fp32 approximation for double and are hence not so accurate for double precision. Would be good to have specializations of these for double as well.

Anri-Lombard:
## Proposed changes

Fixes #3047 - These functions were using float32 polynomial approximations for all inputs, causing precision loss for float64. Added proper double precision paths using element-wise `std::` calls for exp/sin/cos/erf, and Julia's SpecialFunctions.jl rational approximation for erfinv.

**Questions for @awni:**
1. For `erfinv`, I used coefficients from [Julia's SpecialFunctions.jl](https://github.com/JuliaMath/SpecialFunctions.jl/blob/master/src/erf.jl) (Blair et al. 1976, 3-region rational approximation, ~1e-13 relative error). Is this acceptable?
2. I added float64 variants and a `--cpu` flag to the benchmark. Is it useful to include this or rather remove these changes?

## Checklist

### DIFF
--- a/mlx/backend/cpu/simd/math.h
+++ b/mlx/backend/cpu/simd/math.h
@@ -26,8 +26,16 @@ constexpr float inf = std::numeric_limits<float>::infinity();
  */
 template <typename T, int N>
 Simd<T, N> exp(Simd<T, N> in) {
-  if constexpr (is_complex<T>) {
-    return Simd<T, 1>{std::exp(in.value)};
+  if constexpr (is_complex<T> || std::is_same_v<T, double>) {
+    if constexpr (N == 1) {
+      return Simd<T, 1>{std::exp(in.value)};
+    } else {
+      Simd<T, N> result;
+      for (int i = 0; i < N; ++i) {
+        result[i] = std::exp(in[i]);
+      }
+      return result;
+    }
   } else {
     Simd<float, N> x_init = in;
     auto x = x_init * 1.442695f; // multiply with log_2(e)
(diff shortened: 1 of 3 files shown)


mlx/backend/cpu/simd/math.h
```diff
@@ -26,8 +26,16 @@ constexpr float inf = std::numeric_limits<float>::infinity();
  */
 template <typename T, int N>
 Simd<T, N> exp(Simd<T, N> in) {
-  if constexpr (is_complex<T>) {
-    return Simd<T, 1>{std::exp(in.value)};
+  if constexpr (is_complex<T> || std::is_same_v<T, double>) {
+    if constexpr (N == 1) {
+      return Simd<T, 1>{std::exp(in.value)};
+    } else {
```
awni, line 32: Is the check for `N=1` here necessary?

awni:
The purpose of the issue (#3047), perhaps implicitly, was to add actual vectorized implementations rather than unrolled element-wise implementations. Are you interested in making that update? So basically we'd need to find good approximations for exp, sin/cos, erf and use simd functions to implement them. And similarly for erfinv it should be translated into a simd version.

Anri-Lombard:
Hey @awni, so my original approach was using Cephes polynomials, but I ran the benchmark and it was slower for everything except exp (where it was only ~4% faster). For sin and cos it was about 50% slower. But I'm happy to change the implementation if a vectorized implementation is preferred 👍 Perhaps there are reasons I'm not thinking of as well?

mlx/backend/cpu/simd/math.h
Anri-Lombard: Right now Simd<T, 1> (in base_simd.h) doesn't have operator[] - only the general Simd<T, N> in accelerate_simd.h does. So we need the N=1 check to use .value instead of [i]... I think adding operator there could remove this check so that's probably cleaner 🙏

awni:
> but I ran the benchmark and it was slower for everything except exp (where it was only ~4% faster). For sin and cos it was about 50% slower.

That's interesting and unexpected. What machine were you running benchmarks on?

Anri-Lombard:
> > but I ran the benchmark and it was slower for everything except exp (where it was only ~4% faster). For sin and cos it was about 50% slower.
> 
> That's interesting and unexpected. What machine were you running benchmarks on?

I'm using an M1 Macbook Pro 🙏
image

zcbenz (changes requested):
Let's use vectorized implementations for cleaner code.

zcbenz:
Closing due to lack of activity for a long time.

PR #3059, opened 2026-01-24
Update pre-commit hooks and versions for clang-format, black, and isort

NripeshN:


awni:
Mind rebasing and then we can run the tests?

awni:
I might recommend starting from a fresh branch and then:
1. Applying the version bump to the linters
2. Run the linters again

Rather than trying to patch the format changes.

NripeshN:
> I might recommend starting from a fresh branch and then:
> 
> 1. Applying the version bump to the linters
> 2. Run the linters again
> 
> Rather than trying to patch the format changes.

Ah I usually take the lazier approach, copy the conflicting files from main and paste it into my branch manually and run pre-commit😅.

PR #3061, opened 2026-01-25
Delay load CUDA libs and resolve DLL paths at runtime

zcbenz:
Make it possible to dynamically search for CUDA libs on Windows during run time:

1. Delay loading `cudnn.dll` so it is lazy-loaded until cuDNN APIs are used.
2. Add a hook which is used by Windows to locate DLL files when they are not found in system DLL directory or `$PATH`.
3. In the hook:
    1. Load from `nvidia-cudnn-cu13` package (`"../nvidia/cudnn/bin"`) if it is a python package.
    2. Otherwise load from a fixed path which was set during build time.

A downside of this approach is that, when an application uses MLX as a static library, they would need to setup the delay load hook themselves (though the cmake configuration sets up it automatically). But it should work without extra setup if MLX is used as a shared library. /cc @dhiltgen

### DIFF
--- a/mlx/backend/cuda/delayload.cpp
+++ b/mlx/backend/cuda/delayload.cpp
@@ -1,0 +1,80 @@
+// Copyright © 2026 Apple Inc.
+
+#include "mlx/backend/common/utils.h"
+
+// clang-format off
+#include <windows.h> // must be included first
+#include <delayimp.h>
+// clang-format on
+
+namespace mlx::core {
+
+namespace fs = std::filesystem;
+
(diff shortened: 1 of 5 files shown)


mlx/backend/cuda/delayload.cpp
```diff
@@ -0,0 +1,84 @@
+// Copyright © 2026 Apple Inc.
+
+#if defined(_WIN32)
```
awni: ifdef may not be necessary here since it's conditionally added in the cmake?

PR #3062, opened 2026-01-25
Do not require ConcurrentManagedAccess when not used

zcbenz:
Do the check of `ConcurrentManagedAccess` capacity in `AtomicEvent` so MLX can work on consumer hardwares when CPU/GPU synchronization is not required.

Also did some refactoring:
* `cu::Device` becomes cheap to create — the creations of cudnn and cublasLt are now lazy.
* Create all `cu::Device`s during initialization and store them in a vector — this is more efficient than using hash map since we have code iterating all devices (`SmallSizePool`).


zcbenz:
Calling `cu::device` in `SmallSizePool` caused `Device` to be created before `Scheduler` which broke the initialization order. I fixed it by ensuring `Scheduler` is always created before `CudaAllocator`.

PR #3065, opened 2026-01-26
Fix long cache file path on Windows

zcbenz:
By adding `\\?\` before file paths we can make `std::filesystem` APIs support long file path on Windows.
https://learn.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation

Also add `.ilk` file (output of MSVC) to `.gitignore`, and slightly organize the `.gitignore` file.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,16 +3,12 @@ __pycache__/
 *.py[cod]
 *$py.class
 
-# C extensions
-*.so
-
 # tensor files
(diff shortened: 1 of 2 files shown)


zcbenz:
The `max_file_name_length` only limits the length of filename, which works perfectly for mac, but for Windows the there is a limitation on the whole path rather than the filename part.

Issue #3064, opened 2026-01-25
[CUDA][Performance] Topk is slow


awni:
MLX topk on CUDA is pretty slow in some cases (especially compared to PyTorch).

Here is a benchmark:

```python
import time
import mlx.core as mx

b = 2048
v = 8192
k = 32

q = mx.random.normal(shape=(b, v)).astype(mx.bfloat16)

def fun(q):
    for _ in range(50):
        idx = mx.argpartition(-q, kth=k-1, axis=-1)[:, :k]
        values = mx.take_along_axis(q, idx, axis=-1)
        q = mx.put_along_axis(q, idx, values, axis=-1)
    mx.eval(q)

for _ in range(20):
    fun(q)

tic = time.time()
for _ in range(20):
    fun(q)
toc = time.time()
ms = 1e3 * (toc - tic)
print(f"MLX {ms=:.3f}")

import torch


q = torch.randn(size=(b, v)).to("cuda").to(torch.bfloat16)
def topk_old(q):
    return idx, values


def fun(q):
    for _ in range(50):
        values, idx = torch.topk(q, k=k, axis=-1)
        q = torch.scatter(q, -1, idx, values)
    torch.cuda.synchronize()

for _ in range(20):
    fun(q)

tic = time.time()
for _ in range(20):
    fun(q)
toc = time.time()
ms = 1e3 * (toc - tic)
print(f"PyTorch {ms=:.3f}")
```

On a spark:

```
MLX ms=2975.014
PyTorch ms=919.764
```

NripeshN:
At sort.cu, we use gpu_sort() which performs a radix sort which could be inefficient(O(n log n)) compared to even our CPU backend where we use std::nth_element which is O(n).

We could use NVIDIA's RAFT library which already has an efficient top-k selection

awni:
The sort in `sort.cu` is a merge sort not a radix sort. Also a radix sort is not `O(n log n))` but `O(n)` with a constant based on the bit width. For an efficient top-k (esp for half precision) a radix-based selection probably makes sense. Though we will have to play around with that.

NripeshN:
> The sort in `sort.cu` is a merge sort not a radix sort. Also a radix sort is not `O(n log n))` but `O(n)` with a constant based on the bit width. For an efficient top-k (esp for half precision) a radix-based selection probably makes sense. Though we will have to play around with that.

[...]

PR #3069, opened 2026-01-27
[Metal] [Performance] Add radix select implementation for efficient partition operations

NripeshN:
Implementation for metal #3064

### DIFF
--- a/mlx/backend/metal/sort.cpp
+++ b/mlx/backend/metal/sort.cpp
@@ -567,6 +567,7 @@ void gpu_radix_partition(
     bool arg_partition) {
   int axis = axis_ < 0 ? axis_ + in.ndim() : axis_;
   int size_sorted_axis = in.shape(axis);
+  int n_rows = out.size() / size_sorted_axis;
 
   // Normalize kth
   if (kth < 0) {


NripeshN:
@awni 

On metal I am getting these speedup

```
Results for the benchmark case (b=2048, v=8192, k=32):
bfloat16: 0.923ms vs 4.466ms (sort) = 4.84x speedup
float16: 0.822ms vs 4.182ms (sort) = 5.08x speedup
float32: 1.266ms vs 5.280ms (sort) = 4.17x speedup
```

NripeshN:
@awni 

image

NripeshN:
The code in issue #3064 gives the following output with metal and mps backed torch:

MLX ms=967.725
PyTorch ms=7261.384

awni:
A few shape/type combinations are a lot slower (esp a long vector with batch size 1):

Pre:
(mlx.core.bfloat16, b=1, v=128000): ms=99.249
(mlx.core.bfloat16, b=1, v=512): ms=21.032
(mlx.core.bfloat16, b=2048, v=4096): ms=1399.125
(mlx.core.bfloat16, b=2048, v=8192): ms=3214.426
(mlx.core.float32, b=1, v=128000): ms=100.609
(mlx.core.float32, b=1, v=512): ms=20.622
(mlx.core.float32, b=2048, v=4096): ms=1631.850
(mlx.core.float32, b=2048, v=8192): ms=3833.621

Post:

(mlx.core.bfloat16, b=1, v=128000): ms=317.180
(mlx.core.bfloat16, b=1, v=512): ms=21.306
(mlx.core.bfloat16, b=2048, v=4096): ms=386.231
(mlx.core.bfloat16, b=2048, v=8192): ms=721.148
(mlx.core.float32, b=1, v=128000): ms=479.955
(mlx.core.float32, b=1, v=512): ms=35.616
(mlx.core.float32, b=2048, v=4096): ms=761.181
(mlx.core.float32, b=2048, v=8192): ms=1352.038


We could fallback to sort for those cases or try and make the radix kernel faster.

NripeshN:
> A few shape/type combinations are a lot slower (esp a long vector with batch size 1):

Yea this kinda makes sense, Radix select uses a single threadgroup per row, so if number of rows are less and number of elements are higher then it might just be inefficient. Let me see if there is any way I could optimize the kernel but my best bet might just be to fallback to merge sort in cases like this

NripeshN:
@awni 
Does metal have global synchronization between threadgroups?

awni:
The way to do that would be with a second kernel

NripeshN:
The overhead of multiple kernel dispatches might not be worth it tbh. merge sort might just be efficient in this cases.

awni:
For very long vectors (especially half precision) it might be worth it. Sorting is quite a bit more expensive than the radix select.

But in the interest of keeping this PR manageable I think a fall back is good for now and we can follow up if it makes sense

NripeshN:
Out of curiosity have you already started working on cuda on this branch or shall we create a different PR for cuda implementation? I might have to spin up an EC2 instance to work on the cuda implementation, but we could quite definitely get some performance boost with cuda

awni:
I am not working on CUDA. A separate PR would indeed be better.

awni:
Let me know if you plan to work on it, otherwise I will look into it.

NripeshN:
I have honestly gotten way too invested in this😅 I will create a draft PR in a bit and we could work on it together

NripeshN:
@awni 
I am not too sure what heuristics I should set for the fallback tbh. [This](https://github.com/ml-explore/mlx/pull/3069/changes#diff-3cc99a346916b6379fd129f6431da935a9b2ab6fed408c06ce980d4275dc37d8R576-R626) is what I have done for now

awni:
I removed the benchmark as well. I will update the one in the issue which is what we should use. It's more representative of the workload we want to optimize.

NripeshN:
@awni 
~~I made a push for cuda in this branch itself, if you do have a cuda machine could you maybe quickly run the test on the issue so we know if this already works? If we need a complete implementation I will just create a new branch~~

NVM it would not do great, we might have to implement everything ourselves as topk from cuda does not have batch support

NripeshN:
@awni 
I believe this PR is ready to be merged.

angeloskath (commented):
Hi @NripeshN ,

Sorry to come to this quite late. It is a big PR and it looks great on the first look. Thanks for the good work!

However, I think it needs both more work and simplification before it is in a mergeable state. There are issues with missing arguments, the code doesn't compile as is, unused variables and arguments etc.

More importantly there are performance issues. For instance on the shapes that matter for the MoEs there is a big slow down up to 3x for radix select vs sorting. Even more importantly these shapes are easy to deal with and would only require a simpler implementation. Shapes you can test with are `(1, 160)`, `(8, 160)`, `(1024, 160)`, `(8192, 160)`.

I suggest that you make a much simpler PR. Most of the cases should still route to sort/argsort but let's say that you can implement radix select only for the cases where the axis fits in the thread group. You don't really need any atomics either and it should at the very least be faster than sort for this cases. That would be a much nicer PR, easier to review and would have a clearly positive impact.

zcbenz:
I'm closing this PR per comment above.

PR #3070, opened 2026-01-27
[CUDA] Fallback Event impl when there is no hardware cpu/gpu coherency

zcbenz:
For 4090 there is no hardware cpu/gpu coherency, so we can not synchronize between CPU and GPU by sharing atomics. This PR provides a fallback implementation of `AtomicEvent` for this case: we store the atomic in device-only memory, and do CPU waiting with a busy-wait loop that reads the value with `cudaMemcpy`.

This is a very inefficient fallback but it mostly serves to make tests pass, `AtomicEvent` has always been a slow fallback from `CudaEvent` for infrequent use cases.

Also includes some small changes:
* Use uint32 instead of uint64 as the latter has more hardware requirements for atomic ops.
* Use `atomic_ref` instead of `atomic` to make initialization easier for device-only memory.

### DIFF
--- a/mlx/backend/cuda/event.cu
+++ b/mlx/backend/cuda/event.cu
@@ -174,62 +174,94 @@ class CopyableCudaEvent {
 // AtomicEvent implementations
 ///////////////////////////////////////////////////////////////////////////////
 
-__host__ __device__ void event_wait(AtomicEvent::Atomic* ac, uint64_t value) {
-  uint64_t current;
-  while ((current = ac->load()) < value) {
-    ac->wait(current);
+__host__ __device__ void event_wait(uint32_t* ptr, uint32_t value) {
+  cuda::atomic_ref<uint32_t> ac(*ptr);
+  uint32_t current;
+  while ((current = ac.load()) < value) {
+    ac.wait(current);
   }
 }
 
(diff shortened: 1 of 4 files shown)


mlx/backend/cuda/event.cu
```diff
@@ -174,62 +174,94 @@ class CopyableCudaEvent {
...
-        "Device does not support synchronization in managed memory.");
+  void* buf;
+  cudaError_t (*cuda_free)(void*);
+  // There are 3 kinds of systems we are implementing for:
+  // 1. concurrentManagedAccess == true
+  //    => use cuda::atom_ref on managed memory
+  // 2. hostNativeAtomicSupported == true
+  //    => use cuda::atom_ref on pinned host memory
+  // 2. no hardware cpu/gpu coherency
+  //    => use cuda::atom_ref on device memory
```
awni, line 225: Which category (2 or 3) do typical consumer gpus fall into (4090, 5090 etc)?

zcbenz: 4090 is cat 3, I don't know about 5090. Cat 2 is probably only theoretically possible, but it is very simple to support.

PR #3071, opened 2026-01-27
Fix ALiBi slopes for non-power-of-2 num_heads

vovw:
## Proposed Changes

What the diff addresses

The old create_alibi_slope function used a simple formula: x = (2**8) ** (1 / num_heads). The output was mx.power(x, -mx.arange(1, num_heads + 1)). This formula generated a uniform geometric sequence of slopes, which was only accurate when num_heads was a power of 2 (e.g., 1, 2, 4, 8, 16…). However, for non-power-of-2 heads (e.g., 3, 5, 6, 7, 12), it produced incorrect slopes.

The new code employs the algorithm from the https://github.com/ofirpress/attention_with_linear_biases. This algorithm handles non-power-of-2 heads by:

1. Calculating slopes for the nearest lower power of 2.
2. Recursively calculating slopes for 2 times that value and interleaving every other slope for the remaining heads.

Additionally, the diff:

- Removes broken class-level caching (_alibi_mask_key / _alibi_mask), which stored state on the class itself. This was problematic for correctness with varying inputs.
   
## Checklist

### DIFF
--- a/python/mlx/nn/layers/positional_encoding.py
+++ b/python/mlx/nn/layers/positional_encoding.py
@@ -133,40 +133,19 @@ class ALiBi(Module):
         return alibi_mask
 
     @staticmethod
-    def alibi_get_slopes(num_heads: int):
-        """Get the slopes for different attention heads defined in ALiBi paper.
-
-        This is a direct copy from ALiBi codebase.
-        Ref:
-        https://github.com/ofirpress/attention_with_linear_biases/tree/3b7c2eca/fairseq/models/transformer.py#L742-L752
-
-        Args:
-            num_heads: An integer for the number of attention heads.
-
-        Returns:
-            A tensor of slopes with shape of [num_heads]. Each value represents
-            a slope for one attention head.
-        """
-
-        def get_slopes_power_of_2(n: int):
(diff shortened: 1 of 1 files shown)


vovw:
solves ##343

vovw:
cc: @zcbenz

python/mlx/nn/layers/positional_encoding.py
```diff
@@ -115,47 +115,59 @@ def __call__(self, x):
...
+        x2 = mx.arange(0, k_sequence_length)
+        distance_matrix = -mx.abs(
+            mx.expand_dims(x1[:, None] - x2[None, :], axis=(0, 1))
+        )
+        alibi_slope = ALiBi.create_alibi_slope(num_heads=num_heads, dtype=dtype)
+        alibi_mask = (distance_matrix * alibi_slope).astype(dtype)
+        return alibi_mask
+
+    @staticmethod
+    def alibi_get_slopes(num_heads: int):
```
awni: I don't think we need a new method for this. Let's put the implementation in `create_alibi_slope`

vovw: thanks did that :)

vovw:
yay

PR #3073, opened 2026-01-27
Faster vector SDPA in CUDA

awni:
A pretty straight forward port of the changes in https://github.com/ml-explore/mlx/pull/3023 for CUDA. The main difference is the use of shared memory to store keys and values.


awni:
Closing this given https://github.com/ml-explore/mlx/pull/3113 is a much better solution.

PR #3075, opened 2026-01-28
Fallback to pinned host memory when managed memory is not supported

zcbenz:
For 4090 the managed memory does not work on Windows (crashed when accessed on CPU) even though the API reports it being supported, so I'm just disabling managed memory for Windows unless there is hardware unified memory support.

The memory allocation code looks a little messy with all the conditions but I think it is fine for now.

### DIFF
--- a/mlx/backend/cuda/allocator.cpp
+++ b/mlx/backend/cuda/allocator.cpp
@@ -22,6 +22,49 @@ constexpr int page_size = 16384;
 // Any allocations smaller than this will try to use the small pool
 constexpr int small_block_size = 8;
 
+// The small pool size in bytes. This should be a multiple of the host page
+// size and small_block_size.
+constexpr int small_pool_size = 4 * page_size;
+
+bool supports_managed_memory() {
+  static bool managed_memory = []() {
+    int device_count = gpu::device_count();
+    for (int i = 0; i < device_count; ++i) {
+      auto& d = cu::device(i);
+      if (!d.managed_memory()) {
+        return false;
+      }
+#if defined(_WIN32)
+      // Empirically on Windows if there is no concurrentManagedAccess the
+      // managed memory also does not work.
+      if (!d.concurrent_managed_access()) {
+        return false;
+      }
+#endif
+    }
+    return true;
+  }();
+  return managed_memory;
+}
+
+inline void* unified_malloc(size_t size) {
+  void* data = nullptr;
+  if (supports_managed_memory()) {
+    CHECK_CUDA_ERROR(cudaMallocManaged(&data, size));
+  } else {
+    CHECK_CUDA_ERROR(cudaMallocHost(&data, size));
+  }
+  return data;
+}
+
+inline void unified_free(void* data) {
+  if (supports_managed_memory()) {
+    CHECK_CUDA_ERROR(cudaFree(data));
+  } else {
+    CHECK_CUDA_ERROR(cudaFreeHost(data));
+  }
+}
+
 #if CUDART_VERSION >= 13000
 inline cudaMemLocation cuda_mem_loc(int i) {
   cudaMemLocation loc;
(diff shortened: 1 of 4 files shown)


mlx/backend/cuda/allocator.cpp
```diff
@@ -106,32 +134,15 @@ CudaAllocator::CudaAllocator()
...
-    CHECK_CUDA_ERROR(cudaSetDevice(i));
-    cudaStream_t s;
-    CHECK_CUDA_ERROR(cudaStreamCreateWithFlags(&s, cudaStreamNonBlocking));
-    free_streams_.push_back(s);
-
-    cudaMemPool_t mem_pool;
-    CHECK_CUDA_ERROR(cudaDeviceGetDefaultMemPool(&mem_pool, i));
-    mem_pools_.push_back(mem_pool);
+    auto& d = device(i);
+    free_streams_.emplace_back(d);
```
awni: I feel like I did this without accessing the MLX device intentionally. Maybe there was some initialization order thing that was causing problems.. I wish I had left a comment :/.

But then again maybe it's fixed now .. if the tests clear then they clear..

mlx/backend/cuda/allocator.cpp
```diff
@@ -319,6 +340,31 @@ Buffer malloc_async(size_t size, CommandEncoder& encoder) {
...
+    return;
+  }
+  void* new_data;
+  cudaMemcpyKind kind;
+  if (supports_managed_memory()) {
+    CHECK_CUDA_ERROR(cudaMallocManaged(&new_data, buf.size));
+    kind = cudaMemcpyDefault;
+  } else {
+    CHECK_CUDA_ERROR(cudaMallocHost(&new_data, buf.size));
+    kind = cudaMemcpyDeviceToHost;
```
awni: I thought the API could infer that so default should work in both cases.

mlx/backend/cuda/allocator.cpp
```diff
@@ -171,7 +182,11 @@ CudaAllocator::malloc_async(size_t size, int device, cudaStream_t stream) {
     if (!buf) {
       void* data = nullptr;
       if (device == -1) {
-        CHECK_CUDA_ERROR(cudaMallocManaged(&data, size));
+        if (supports_managed_memory()) {
+          CHECK_CUDA_ERROR(cudaMallocManaged(&data, size));
+        } else {
+          CHECK_CUDA_ERROR(cudaMallocHost(&data, size));
+        }
```
awni: There are a few cases of `if (supports_managed) do x else y fi`

It might make sense to refactor to a `unifiedMalloc` and `unifiedFree` to keep the code a little more readable.

mlx/backend/cuda/allocator.cpp
```diff
@@ -106,32 +134,15 @@ CudaAllocator::CudaAllocator()
...
-    free_streams_.push_back(s);
-
-    cudaMemPool_t mem_pool;
-    CHECK_CUDA_ERROR(cudaDeviceGetDefaultMemPool(&mem_pool, i));
-    mem_pools_.push_back(mem_pool);
+    auto& d = device(i);
+    free_streams_.emplace_back(d);
+    if (d.memory_pools()) {
+      CHECK_CUDA_ERROR(cudaDeviceGetDefaultMemPool(&mem_pools_[i], i));
+    }
```
awni, line 156: What's the purpose of that check here? Some devices do not support memory pools?

mlx/backend/cuda/allocator.cpp
zcbenz: I think you ran into https://github.com/ml-explore/mlx/pull/3062#issuecomment-3797832133.

mlx/backend/cuda/allocator.cpp
zcbenz: Yeah according to https://github.com/ml-explore/mlx/pull/2972/changes#diff-3e8aaaff4c1529bbcf6ea804df3793a6c354f2812ff63377dffec82b8ca4321d some devices do not have memory pools. Also I just realized that `cudaMallocAsync` should not be used when memory pools is not available, will make change.

mlx/backend/cuda/allocator.cpp
zcbenz: The document "However, cudaMemcpyDefault is only allowed on systems that support unified virtual addressing." is vague whether the flag would still work when unified memory is not available.

awni: Better safe than sorry.

PR #3076, opened 2026-01-28
Fix failing python tests on Windows

zcbenz:
Some small fixes making all tests pass on Windows:

* Initialize `ArrayDesc::data_size` otherwise empty arrays which usually do not call `array::set_data` would get garbage data.
* Only use tensor ops in grouped mm when tf32 is enabled.
* Avoid accessing `outputs[1]` in SDPA when `output_logsumexp` is false.
* Do not try to get vjp for mask in `test_sdpa_grad`.

### DIFF
--- a/mlx/backend/cuda/scaled_dot_product_attention.cpp
+++ b/mlx/backend/cuda/scaled_dot_product_attention.cpp
@@ -140,7 +140,7 @@ DnnGraph build_sdpa_graph(
     const std::optional<array>& mask_arr,
     bool output_logsumexp,
     const array& o,
-    const array& stats) {
+    const std::optional<array>& stats) {
   DnnGraph graph(handle, q.dtype());
 
   auto q_ = graph.tensor("Q", Q, q);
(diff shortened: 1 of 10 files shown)


mlx/backend/cuda/scaled_dot_product_attention.cpp
```diff
@@ -428,8 +428,8 @@ void ScaledDotProductAttention::eval_gpu(
   array q = prepare_sdpa_input(inputs[0], s);
   array k = prepare_sdpa_input(inputs[1], s);
   array v = prepare_sdpa_input(inputs[2], s);
-  auto& out = outputs[0];
-  auto& stats = outputs[1];
+  array& out = outputs[0];
+  array* stats = output_logsumexp_ ? &outputs[1] : nullptr;
```
awni: Why not `std::optional`? Seems more in line with the rest of the code.

zcbenz: We need to set data on the original array i.e. `outputs[1].set_data()` but `std::optional<array>` would make a copy.

awni: Technically that's fine. `set_data` modifies the underlying array descriptor.

```
auto x = y;
x.set_data(..) # sets the data for both x and y
```

But avoiding the shared pointer copy is reasonable.

PR #3088, opened 2026-02-01
tests: harden memory leak check in test_siblings_without_eval

booxter:
When running the test suite on darwin as part of nixpkgs package build,
we often hit the following error:

```
       >         mx.synchronize()
       >         t()
       >         gc.collect()
       >         expected = get_mem()
       >         for _ in range(100):
       >             t()
       >         used = get_mem()
       > >       self.assertEqual(expected, used)
       > E       AssertionError: 273678336 != 273694720
```

You can find an example of the full failure log here:
https://hydra.nixos.org/build/320731525

The difference between actual and expected is 16k (one page size on
aarch64-darwin).

The test seems brittle as it depends on *peak* memory, not real memory
at the moment of measurement, so anything during the run of the process
that could push the allocator to allocate another page can break the
test.

This patch switches the test case to measure real memory using psutil on
all platforms.

Signed-off-by: Ihar Hrachyshka <ihar.hrachyshka@gmail.com>

## Proposed changes

## Checklist

### DIFF
--- a/.github/actions/build-macos/action.yml
+++ b/.github/actions/build-macos/action.yml
@@ -12,12 +12,12 @@ runs:
       run: |
         pip install --upgrade pip
         pip install cmake setuptools typing_extensions
-        pip install -e . -v
+        pip install -e ".[dev]" -v
 
     - name: Install tests dependencies
       shell: bash -l {0}
       run: |
-        pip install numpy torch tensorflow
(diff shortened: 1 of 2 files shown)


booxter:
Looks like I broke something that was not present before I rebased to master. Will fix.

zcbenz (approved):

Issue #206, opened 2023-12-18
Possible mx.array bug? 


sck-at-ucy:
I think the `mx.array` is not initializing arrays the way one would expect and does not raise an error when one references array elements that are out of bounds, which can be a trap when implementing complex logic, especially for guys like me, who have been trained in fortran and always have to be conscious about the indexing 😄 ... 

Here's a simple example to highlight what I think is a bug:

```
import mlx.core as mx
import numpy as np

# Initializing the temperature field and copy on the GPU
T_mx = mx.ones([4, 4])
T_np = np.ones([4, 4])

print(T_mx)
print(T_np)
print(T_mx[0,0], T_mx[0,1], T_mx[0,2], T_mx[0,3], T_mx[0,4], T_mx[0,5], T_mx[0,6])
print(T_np[0,0], T_np[0,1], T_np[0,2], T_np[0,3], T_np[0,4], T_np[0,5], T_np[0,6])
```

and here's the output 

```
array(1, dtype=float32) array(1, dtype=float32) array(1, dtype=float32) array(1, dtype=float32) array(1, dtype=float32) array(1, dtype=float32) array(1, dtype=float32)
Traceback (most recent call last):
  File "/Users/m2/PycharmProjects/pythonProject_MXL/bug.py", line 11, in <module>
    print(T_np[0,0], T_np[0,1], T_np[0,2], T_np[0,3], T_np[0,4], T_np[0,5], T_np[0,6])
                                                      ~~~~^^^^^
IndexError: index 4 is out of bounds for axis 1 with size 4

Process finished with exit code 1
```

MLX is happy to putting 1 into the out of bounds elements, while Numpy properly raises error for this elements.

jagrit06:
Thanks for flagging this! I'll get working on a fix right away

awni:
Actually, this is expected behavior. This is in general once place where we might end up being different than NumPy and go more like Jax. You can try the same example in Jax, I believe the default behavior is to use the last value for out of bounds indexing (but it doesn't throw).

One reason we go this route in general is that's we don't raise exceptions from Metal kernels and we don't want to sanitize inputs before the kernel as that is costly / doesn't make sense with the execution model.

There maybe some simple cases (like Python integer based indexing) that we can properly throw (or at least make it an option) for. But when you start to do more fancy indexing with other `mx.array`, we may have to follow Jax.

Another thing we should do @jagrit06 is document this a little more clearly so it's not a surprise.

awni:
[...]

PR #3091, opened 2026-02-02
Add bounds checking on GPU

nhey:
## Proposed changes

I propose to slow things down a bit. This is a draft implementation of array bounds checking in the Metal backend. The scatter and gather benchmarks show a 2% median runtime overhead on an 8-core M4 GPU (see results below). Is this something you'd be interested in merging? I'll probably be working on it for fun, so I'd appreciate your thoughts either way :)

The low overhead is due to @athas's strategy described in [Bounds Checking on GPU](https://futhark-lang.org/publications/hlpp20.pdf): errors are propagated from the GPU to the CPU by writing a single global 32-bit integer. This is just a draft to play around with the essence of it. I've matched the behaviour against pytorch's MPS backend, which appears to prioritize performance over safety. I don't yet report the out-of-bounds index value or the axis size, but it'd be straightforward to add this. Also I assume that gathers and scatters are never fused. A non-draft PR should add bounds checking to the CPU and CUDA backends too, of course.

Details
---

The strategy/implementation has limitations:
1. Only one error is recorded and, if there are multiple errors at runtime, it's indeterministic which one.
2. Handling index errors is disallowed.

The global error is written/read/thrown asynchronously and really means "there is at least one error", so handling the thrown error may leave us in a bad state. I check the global error in `array::wait()` and reset it right before throwing. (Otherwise the error would persist, being thrown on every eval when MLX is used in the Python REPL.) So the program must terminate on index errors.

3. My implementation doesn't make indexing safe.

The paper prevents memory corruption by checking global failure in the prelude of every kernel. I didn't implement this, so kernels that have been enqueued before the global failure was set may operate on invalid inputs such as uninitialized arrays. This means _all_ arrays have to be considered invalid on an index error. Still, there can be no out-of-bounds reading or writing and an index error will get reported to the user. I'm unsure whether pytorch allows memory corruption, I did not study their implementation.

Possible improvements:
* Better error information.
* Optimization: disable bounds checking for safe gather/scatter calls (e.g., I imagine some `jvp`/`vjp` operations are safe if the corresponding primal has been checked).
* Bounds checking specialized fused kernels like `gather_mm`.
* Preventing memory corruption by checking for global failure in every GPU kernel. The paper demonstrates low overhead.
* As the paper mentions, you can use this method to report any kind of error (e.g., division by zero).

Related to #206.

Evaluation
---

I ran `benchmarks/python/gather_bench.py` and `benchmarks/python/scatter_bench.py` to get the below results.

Benchmark                       | main (ms) | bounds (ms) | Overhead
--------------------------------|-----------|-------------|---------
Gather: X(100, 64)              |     3.331 |       3.409 |   +2%
Gather: X(100, 1024)            |     4.903 |       4.928 |   +0.5%
Gather: X(4, 1000000)           |     0.312 |       0.311 |   -0.3%
Scatter: Dst(10, 64)            |     3.800 |       4.127 |   +8%
Scatter: Dst(100000, 64)        |     7.373 |       7.595 |   +3%
Scatter: Dst(1000000, 64)       |     1.185 |       1.209 |   +2%
Scatter: Dst(100000,)           |     0.408 |       0.403 |   -1%
Scatter: Dst(200000,)           |     3.030 |       3.000 |   -1%
Scatter: Dst(20000000,)         |    31.614 |      31.682 |   +0.2%
Scatter: Dst(10000, 64)         |     4.071 |       4.315 |   +6%
Scatter: Dst(100, 64)           |    35.631 |      38.184 |   +7%
Scatter: Dst(100, 10000, 64)    |    49.503 |      52.397 |   +6%
Scatter: Dst(10, 100, 100, 21)  |   210.873 |     219.162 |   +4%
Scatter: Dst(1000, 1000, 10)    |     0.910 |       0.761 |  -16%

Summary _disregarding the -16% test_ (which I guess is just noise):

Average Overhead: 2.85%
Max Overhead: 8.61%
Median Overhead: 2.34%

## Checklist

### DIFF
--- a/mlx/CMakeLists.txt
+++ b/mlx/CMakeLists.txt
@@ -7,6 +7,7 @@ target_sources(
           ${CMAKE_CURRENT_SOURCE_DIR}/dtype_utils.cpp
           ${CMAKE_CURRENT_SOURCE_DIR}/export.cpp
           ${CMAKE_CURRENT_SOURCE_DIR}/einsum.cpp
+          ${CMAKE_CURRENT_SOURCE_DIR}/failure.cpp
           ${CMAKE_CURRENT_SOURCE_DIR}/fast.cpp
           ${CMAKE_CURRENT_SOURCE_DIR}/fft.cpp
           ${CMAKE_CURRENT_SOURCE_DIR}/ops.cpp
--- a/mlx/array.cpp
+++ b/mlx/array.cpp
@@ -3,6 +3,7 @@
 #include <unordered_map>
 
 #include "mlx/array.h"
+#include "mlx/failure.h"
 #include "mlx/ops.h"
 #include "mlx/primitives.h"
 #include "mlx/transforms.h"
@@ -132,7 +133,7 @@ bool array::is_available() const {
   if (status() == Status::available) {
     return true;
   } else if (
-      status() == Status::evaluated &&
+      status() == Status::evaluated && !global_failure() &&
       (!event().valid() || event().is_signaled())) {
     set_status(Status::available);
     return true;
@@ -145,6 +146,10 @@ void array::wait() {
     if (event().valid()) {
       event().wait();
       detach_event();
+    }
+    if (global_failure()) {
+      reset_global_failure();
+      throw std::out_of_range("Array index out of bounds");
     }
     set_status(Status::available);
   }
--- a/mlx/backend/gpu/failure.h
+++ b/mlx/backend/gpu/failure.h
@@ -1,0 +1,10 @@
+// Copyright © 2026 Apple Inc.
+
+#pragma once
+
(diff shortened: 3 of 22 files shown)


awni:
I think a better and much simpler compromise would be to use Metal logging and only enable this in debug mode. See e.g https://ml-explore.github.io/mlx/build/html/dev/metal_logging.html

nhey:
It's a different game if it's a debug only feature. It seems straightforward to change my commit to use logging. I could do that if you want me to.

zcbenz:
Considering the overhead this is not going to be adopted by industry, for debugging it feels useful but adds a lot of complexity to code, so I think unless there is popular request for it we are not going to support this.

PR #3092, opened 2026-02-02
Fix for NAX overflow.

awni:
As title.


awni:
Closes https://github.com/ml-explore/mlx-swift/issues/344

angeloskath (approved):
Very nice catch. It took me a second reading it...

jagrit06 (approved):
Thanks for the fix

PR #3094, opened 2026-02-03
JACCL update

angeloskath:
This update adds the following:

- Large buffers for JACCL that fix the BW ( >= macOS 26.3 h/t Jack Beasly)
- JACCL ring for arbitrarily large number of connected Macs and larger bandwidth (albeit with slightly higher latency for small sizes)
- Adds the backend and environment variables to the hostfile so we don't have to define them when calling `mlx.launch`

The following shows the achieved bandwidth with the ring and mesh across 4 M3 Ultras.
jaccl-bw

The code needs a cleaning pass for sure but I think it is in a mergeable state. The main issue is I am reusing a lot of logic with slight changes which may (or may not if we lose performance) be possible to simplify.

Here are some distributed generation benchmarks comparing before and after this PR. They are with a 2048 tokens prompt, generating 128 tokens and a batch size of 1.

| Model                  | Phase      | Nodes |         | JACCL before | JACCL now (mesh) | JACCL now (ring) | Delta |
|------------------------|------------|------:|--------:|-------------:|-----------------:|-----------------:|------:|
| Devstral2 123B (4-bit) | Prompt     | 1     | 91.5    | —            | —                | —                | —     |
| Devstral2 123B (4-bit) | Prompt     | 2     | —       | 119          | 166              | 167              | +40.3%|
| Devstral2 123B (4-bit) | Prompt     | 4     | —       | 135          | 255              | 303              | +124.4%|
| Devstral2 123B (4-bit) | Generation | 1     | 9.0     | —            | —                | —                | —     |
| Devstral2 123B (4-bit) | Generation | 2     | —       | 15.4         | 15.8             | 15.5             | +2.6% |
| Devstral2 123B (4-bit) | Generation | 4     | —       | 24.5         | 26.2             | 24.1             | +6.9% |
|                        |            |       |         |              |                  |                  |       |
| GLM 4.7 358B (6-bit)   | Prompt     | 1     | 265     | —            | —                | —                | —     |
| GLM 4.7 358B (6-bit)   | Prompt     | 2     | —       | 305          | 445              | 446              | +46.2%|
| GLM 4.7 358B (6-bit)   | Prompt     | 4     | —       | 309          | 599              | 719              | +132.8%|
| GLM 4.7 358B (6-bit)   | Generation | 1     | 16.9    | —            | —                | —                | —     |
| GLM 4.7 358B (6-bit)   | Generation | 2     | —       | 22.2         | 22.3             | 21.9             | +0.5% |
| GLM 4.7 358B (6-bit)   | Generation | 4     | —       | 27.4         | 27.7             | 25.2             | +1.1% |

It is also very impressive to compare batched inference performance where the MoE which traditionally doesn't shard very well with batch size 1. With batch size 8 we **go from a speedup of 1.6x with 4 machines to a speedup of 2.6x** !

### DIFF
--- a/mlx/distributed/jaccl/ring.cpp
+++ b/mlx/distributed/jaccl/ring.cpp
@@ -167,8 +167,8 @@ void RingGroup::all_gather(const array& input, array& output, Stream stream) {
 
     // Counters to maintain the state of transfers
     int in_flight = 0;
-    size_t send_offset[2];
-    size_t recv_offset[2];
+    int64_t send_offset[2];
+    int64_t recv_offset[2];
     int64_t limits[2];
     int send_count[2 * MAX_CONNS] = {0};
     int recv_count[2 * MAX_CONNS] = {0};
@@ -187,12 +187,13 @@ void RingGroup::all_gather(const array& input, array& output, Stream stream) {
         post_recv_all(sz, buff);
         for (int lr = 0; lr < 2; lr++) {
           for (int lw = 0; lw < n_wires; lw++) {
-            int offset = lw * N +
+            int64_t offset = lw * N +
                 send_count[lr * MAX_CONNS + lw] * n_wires * N +
                 lr * n_wires * n_bytes_per_wire;
             std::copy(
                 out_ptr + send_offset[lr] + offset,
-                out_ptr + send_offset[lr] + std::min(offset + N, limits[lr]),
+                out_ptr + send_offset[lr] +
+                    std::max(offset, std::min(offset + N, limits[lr])),
                 send_buffer(sz, buff, lr, lw).begin<char>());
             send_count[lr * MAX_CONNS + lw]++;
           }
@@ -219,11 +220,12 @@ void RingGroup::all_gather(const array& input, array& output, Stream stream) {
           in_flight--;
 
           if (work_type == SEND_WR && send_count[wire] < n_steps) {
-            int offset = lw * N + send_count[wire] * n_wires * N +
+            int64_t offset = lw * N + send_count[wire] * n_wires * N +
                 lr * n_wires * n_bytes_per_wire;
             std::copy(
                 out_ptr + send_offset[lr] + offset,
-                out_ptr + send_offset[lr] + std::min(offset + N, limits[lr]),
+                out_ptr + send_offset[lr] +
+                    std::max(offset, std::min(offset + N, limits[lr])),
                 send_buffer(sz, buff, lr, lw).begin<char>());
             send_to(sz, buff, lr, lw);
             in_flight++;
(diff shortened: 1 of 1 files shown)


blightbow:
Has send/recv been working well for you in live testing? While implementing P2P transfers of files between M3 Ultras using JACCL, I found that tight synchronization between send_like and recv_like was an absolute must in order to avoid silent data corruption. Did your test suite include validation that the data received matched what was originally sent? all_reduce is an inherently synchronized operation, so I would caution additional testing if you have not confirmed this already.

angeloskath:
@blightbow with JACCL in general or the JACCL ring now? With JACCL yeah and I just tested a file transfer script it works fine and with this PR gets quite good bandwidth (basically disk is the bottleneck).

I tested the ring send/recv but indeed now in my tests it seems to have some issue so I will update that. But JACCL should work fine.

You can find a script to do the file transfer at https://gist.github.com/angeloskath/3d63afc50bd122527b426174fa739b7e . It is basically bottlenecked by disk-read speed at ~6 GB/s.

angeloskath:
@blightbow I had only ran the test with 4 nodes hence the bug went unnoticed. With 2 nodes left and right are the same node so it was being selected incorrectly. Works as expected now and allows to scale to multiple cables for bigger bandwidth (although it won't help for the file transfer case as the disk is the bottleneck).

blightbow:
Mitigating factors I'm observing:

- mx.eval(send(size)) on the sender and recv(size).item() on the remote creates a two step blocking handshake that imposes synchronization
- f.read(chunk_size) of 100Mb combined with disk I/O constraints creates fewer race opportunities

In my case I was streaming model weights into the memory of the peer (Rank 1), without anything touching the disk on the remote side. The disk I/O only happens when it's time to read another chunk from Rank 0.

To answer your earlier question, this was all base JACCL with no ring. Just providing insights -- I've been tracking this branch because I've been wanting public access to SideChannel.

angeloskath:
I don't think the disk io plays any role here. The fact that each communication is on its own eval may be helping though. Either way the issue is not with the distributed backend but there could be some other issue that has to do with the order that multiple sends/recvs are ran on different nodes.

I would suggest to start an issue with a reproduction (as minimal as you can make it). Keep in mind that communication runs on the CPU and our CPU runs ops 1 at a time so there can't really be a race between two sends on a single node.

blightbow:
Agreed. Once this PR hits I'll create a branch with tests that recreate the issue, and use SideChannel as a mechanism to avert it. My own tests are very ugly right now and started off with using an OOB coordinator process to mitigate, but a minimal test is better and this really belongs in SideChannel regardless.

PR #3095, opened 2026-02-03
Disable managed memory on WSL when concurrentManagedAccess is not supported

jessegross:
Extend the Windows managed memory check from #3075 to also apply to WSL, as the underlying behavior is the same.

Otherwise, on WSL we get a segfault on load


zcbenz:
Can you fix the lint error?

jessegross:
Just did :)

PR #3096, opened 2026-02-04
Fix qmv_impl for small N

manuelcandales:
## Proposed changes

Fixes two bugs in `qmv_impl`:
 
* Fix 1: Changed `qdot` to `qdot_safe` when handling remaining elements in the `N < 8` case
Notice that in `qmv_impl`, there are two code paths based on whether `N < 8`:                                                                             
  - When `N >= 8`: Correctly uses `qdot_safe` for remaining elements                                                                          
  - When `N < 8`: Incorrectly used `qdot` instead of `qdot_safe` for remaining elements (fixed in this PR)
 
* ~Fix 2: Changed `in_vec_size_g` from floor division to ceiling division to correctly handle cases where `K` is not a multiple of group size~
EDIT: I have reverted Fix 2, since it seems that MLX never allows K to be non-multiple of group_size

## Checklist

### DIFF
--- a/python/tests/cuda_skip.py
+++ b/python/tests/cuda_skip.py
@@ -50,6 +50,7 @@ cuda_skip = {
     "TestQuantized.test_fp_qvm",
     "TestQuantized.test_qvm",
     "TestQuantized.test_qvm_splitk",
+    "TestQuantized.test_qmv_small_non_multiples",
     "TestQuantized.test_small_matrix",
     "TestQuantized.test_throw",
     "TestQuantized.test_vjp_scales_biases",


awni:
I would imagine we have the same bug in `fp_quantized.h` : https://github.com/ml-explore/mlx/blob/main/mlx/backend/metal/kernels/fp_quantized.h#L442

Do you mind fixing it there as well and updating the test to include another quantization mode like `nvfp4`?

python/tests/test_quantized.py
```diff
@@ -681,6 +681,68 @@ def test_non_multiples(self):
         self.assertEqual(y_q.shape, y_hat.shape)
         self.assertLess((y_q - y_hat).abs().max(), 1e-3)
 
+    def test_small_non_multiples(self):
```
awni: Nit: could you change this function name to `tests_qmv_small_non_multiple` so it's a bit more descriptive.

awni:
Good catch! We can merge this as soon as comments are addressed!

awni (approved):
LGTM!

PR #3098, opened 2026-02-05
Proposal to add dynamic backends to MLX

dhiltgen:
### Dynamic Backend Loading for MLX

Today MLX compiles backends statically via CMake flags (`MLX_BUILD_METAL`, `MLX_BUILD_CPU`, `MLX_BUILD_CUDA`).  Each build of `libmlx` is locked to the exact set of backends chosen at compile time.  As new hardware targets emerge — SIMD-tuned x86 CPU variants (SSE, AVX2, AVX-512), multiple CUDA Compute Capabilities, vendor-specic accelerators — it becomes impractical to ship a single binary that covers all targets.

Dynamic backend loading solves this by:

1. Letting a single `libmlx` load backend shared libraries at runtime.
2. Enabling independent versioning and distribution of backends.
3. Allowing the runtime (or the application) to select the best backend for the hardware it discovers at startup.
4. Supporting multiple GPU backends simultaneously (e.g. CUDA alongside future backends for non-NVIDIA GPUs).
5. Enabling third-party/out-of-tree backends without rebuilding MLX.
6. Decouples transitive dependencies from the primary application linked to libmlx which allows startup and graceful detection of missing dependencies (e.g., CUDA)

## Checklist

### DIFF
--- a/dynamic_backend_proposal.md
+++ b/dynamic_backend_proposal.md
@@ -1,0 +1,914 @@
+# Dynamic Backend Loading for MLX
+
+## Motivation
+
+Today MLX compiles backends statically via CMake flags (`MLX_BUILD_METAL`,
+`MLX_BUILD_CPU`, `MLX_BUILD_CUDA`).  Each build of `libmlx` is locked to the
+exact set of backends chosen at compile time.  As new hardware targets emerge
+— SIMD-tuned x86 CPU variants (SSE, AVX2, AVX-512), multiple CUDA Compute
+Capabilities, vendor-specific accelerators — it becomes impractical to ship
+a single binary that covers all targets.
+
+Dynamic backend loading solves this by:
+
+1. Letting a single `libmlx` load backend shared libraries at runtime.
+2. Enabling independent versioning and distribution of backends.
+3. Allowing the runtime (or the application) to select the best backend for
+   the hardware it discovers at startup.
+4. Supporting multiple GPU backends simultaneously (e.g. CUDA alongside
+   future backends for non-NVIDIA GPUs).
+5. Enabling third-party/out-of-tree backends without rebuilding MLX.
+6. Decouples transitive dependencies from the primary application linked
+   to libmlx which allows startup and graceful detection of missing
+   dependencies (e.g., CUDA)
+
+---
+
+## Design
+
+### 1. Backend Plugin Interface
+
+Each backend shared library implements a C++ abstract interface and exports
+C entry points for `dlopen`-based loading.
+
+#### 1.1 `mlx/backend_interface.h` — The Backend Contract
+
+```cpp
+#pragma once
+
+#include "mlx/api.h"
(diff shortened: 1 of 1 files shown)


jessegross:
Overall, I think fleshing out support for backends will be great - here are a few comments mostly focusing on the high level "user" (client API) experience:

**Device Interface**

MLX does have support for different concurrent backends now in the sense that you can use CPU or GPU together. If we expand this to additional types of backends like Vulkan, I would expect that the experience should be pretty similar. That includes mixing Vulkan and CUDA but less so SIMD backends since there would only be one loaded at a time in that case.

If we apply the current experience for mixing CPU and GPU to the "User Experience" section at the end of the doc, it would look like this:

```
with mx.stream(mx.default_stream(device=mx.gpu(0))): # CUDA
    a = mx.ones(1000, 1000)
with mx.stream(mx.default_stream(device=mx.cpu)):    # CPU
    b = mx.ones(1000, 1000)

c = mx.matmul(a, b) # No error
```

Conceptually, with a second backend I would expect it to look the same. For example, if you switch the CPU for a Vulkan device, the same pattern should still work. That implies that you likely need to handle cross backend transfers with an automatic bounce through system memory. However, that seems better than having multiple ways of working with devices - the current proposal treats the new backends differently by requiring operations that allocate memory to specify the device and explicit copies.

**Out of tree backends**

I would be cautious about supporting plugins and out of tree backends. Officially supporting plugins requires maintaining some level of backwards compatibility, which makes it more difficult to evolve interfaces rather than just updating the source in the tree. This can both add complexity and slow down improvements. This is especially true in the beginning when the interfaces are likely to change more and the use case for plugins isn't completely clear. Some of the complexity already surfaces in the section around C++ ABI compatibility.

angeloskath:
Thanks Daniel for starting this discussion and the very well thought out proposal. We were debating amongst ourselves lately whether we should go down a similar route. Especially with respect to facilitating people to implement "backends" for MLX.

Now having said that, I think the proposal ignores a major part of MLX and as a result underestimates the undertaking a little bit. The implementation of the operations are in the `Primitive`s and are different per "backend".

I see two ways of actually achieving what you have proposed.

**Case 1: Abstract away all the ops**

As the title says. The backend actually abstracts away the ops completely (or some ops) and it can use any method to implement them. A bit like the distributed backends do basically (mind you with a very simple and limited API though).

**Case 2: Abstract away primitive creation**

The ops shouldn't create the primitive directly by `std::make_shared<FooOperation>(...)` but they call on the "backend" to provide the primitive for the `foo` operation.

Both of the above are quite a bit of an undertaking with deep changes to MLX. I think the second one is the way to go if we ever decide to go down that route because it allows us to share at least some of the high level logic of the operations, of compile etc.

dhiltgen:
Closing for now.  I think dynamic loading between multiple variants of the library is sufficient for my use-cases.

PR #3099, opened 2026-02-05
Fix 2pass sdpa on < M2

awni:
Closes https://github.com/ml-explore/mlx-lm/issues/844

Honestly I think this is a lower level issue. For some reason using blocks = 128 when blocks is a function constant makes the kernel support < 1024 threads per thread group. Other values (64, 256, etc) do not have that issue. This is just for bfloat16 on M1 and M2 🤷‍♂️


awni:
No but at first I tried getting rid of the blocks function constant entirely from the first pass as well and there was a very note-able regression.. which was unexpected.

PR #3106, opened 2026-02-06
Quantize module to QQLinear

nastya236:
Extended `Linear.to_quantize(..)` and `nn.quantize()` to support quantization to `QQLinear`.

Added a flag `quantize_input` that defines the behaviour (default is `False`, so `QuantizedLinear` is used).

```python
class DuckModel(nn.Module):
    def __init__(self, dim, n_layers):
        super(DuckModel, self).__init__()
        self.block = nn.Sequential(
            nn.Linear(dim, dim, bias=False),
            nn.ReLU(),
            nn.Linear(dim, dim, bias=False),
            nn.ReLU()
        )

        self.layers = [
            self.block for _ in range(n_layers)
        ]

    def __call__(self, x):
        for layer in self.layers:
            x = layer(x)
        return x

model = DuckModel(128, 4)
x = mx.random.uniform(shape=(128, 128))
y_hat = model(x) 
nn.quantize(model, quantize_input=True, mode='mxfp8', class_predicate=lambda path, module: isinstance(module, nn.Linear))
y = model(x) 
mx.allclose(y, y_hat, atol=1e-3)
```

### DIFF
--- a/python/mlx/nn/layers/embedding.py
+++ b/python/mlx/nn/layers/embedding.py
@@ -49,5 +49,5 @@ class Embedding(Module):
     ):
         """Return a :obj:`QuantizedEmbedding` layer that approximates this embedding layer."""
         if quantize_input:
-            raise ValueError("[Embedding] Quantized input is not supported.")
+            raise ValueError("Quantized input is not supported.")
         return QuantizedEmbedding.from_embedding(self, group_size, bits, mode)
--- a/python/mlx/nn/layers/linear.py
+++ b/python/mlx/nn/layers/linear.py
@@ -102,7 +102,7 @@ class Linear(Module):
         if quantize_input:
             if mode not in ["nvfp4", "mxfp8"]:
                 raise ValueError(
-                    f"[Linear] Quantized activations are only supported for 'nvfp4' and 'mxfp8' modes, got {mode}."
+                    f"Quantized activations are only supported for 'nvfp4' and 'mxfp8' modes, got {mode}."
                 )
             return QQLinear.from_linear(self, group_size, bits, mode)
         return QuantizedLinear.from_linear(self, group_size, bits, mode)
--- a/python/tests/test_nn.py
+++ b/python/tests/test_nn.py
@@ -204,6 +204,27 @@ class TestBase(mlx_tests.MLXTestCase):
         self.assertTrue(isinstance(m.layers[2], nn.QuantizedLinear))
         self.assertTrue(isinstance(m.layers[2].scales, mx.array))
 
+        m = nn.Sequential(
+            nn.Embedding(5, 256), nn.ReLU(), nn.Linear(256, 256, bias=False)
+        )
+        nn.quantize(
+            m,
+            group_size=32,
+            mode="mxfp8",
+            quantize_input=True,
+            class_predicate=lambda path, module: isinstance(module, nn.Linear),
+        )
+        self.assertTrue(isinstance(m.layers[0], nn.Embedding))
+        self.assertTrue(isinstance(m.layers[1], nn.ReLU))
+        self.assertTrue(isinstance(m.layers[2], nn.QQLinear))
+
+        # Check that Embedding does not support quantize_input
+        m = nn.Sequential(
+            nn.Embedding(5, 256), nn.ReLU(), nn.Linear(256, 256, bias=False)
+        )
+        with self.assertRaises(ValueError) as context:
+            nn.quantize(m, group_size=32, mode="mxfp8", quantize_input=True)
+
     def test_quantize_freeze(self):
         lin = nn.Linear(512, 512)
         qlin = lin.to_quantized()


python/mlx/nn/layers/quantized.py
```diff
@@ -41,22 +53,35 @@ def quantize(
...
+    is_embedding = lambda m: m.__class__.__name__ == "Embedding"
+    class_predicate = class_predicate or (
+        lambda _, m: hasattr(m, "to_quantized")
+        and (not quantize_input or not is_embedding(m))
+    )
 
     def _maybe_quantize(path, m):
         if bool_or_params := class_predicate(path, m):
             if hasattr(m, "to_quantized"):
                 if isinstance(bool_or_params, bool):
```
angeloskath, line 72: Oh that seems like a bug that went unnoticed for a while.
```suggestion
                if isinstance(bool_or_params, bool):
                    if not bool_or_params:
                        return m
```

python/mlx/nn/layers/quantized.py
```diff
@@ -41,22 +53,35 @@ def quantize(
...
+                    return m.to_quantized(
+                        group_size=group_size,
+                        bits=bits,
+                        mode=mode,
+                        quantize_input=quantize_input,
+                    )
                 elif isinstance(bool_or_params, dict):
-                    return m.to_quantized(**bool_or_params)
+                    return m.to_quantized(
+                        **bool_or_params, quantize_input=quantize_input
```
angeloskath: No `quantized_input` here. The point is the predicate returned the quantization options. If they want `quantize_input=True` they should set it.

python/mlx/nn/layers/quantized.py
```diff
@@ -41,22 +53,35 @@ def quantize(
...
         class_predicate (Optional[Callable]): A callable which receives the
           :obj:`Module` path and :obj:`Module` itself and returns ``True`` or a
           dict of params for `to_quantized` if it should be quantized and
           ``False`` otherwise. If ``None``, then all layers that define a
           ``to_quantized(group_size, bits)`` method are quantized.
           Default: ``None``.
     """
-    class_predicate = class_predicate or (lambda _, m: hasattr(m, "to_quantized"))
+    # qqmm is not supported for Embedding layer
+    is_embedding = lambda m: m.__class__.__name__ == "Embedding"
```
angeloskath: Probably better to do `lambda m: isinstance(m, Embedding)`.

awni:
> We could do the same for quantize_input or we could make it optional and nn.quantize only passes it if it is True.

For the quantize-able modules we have now `Embedding`, `Linear`, `SwitchLinear`, it makes sense for all of them to support `quantized_input`. (For the embedding it would only be used when you call it as a linear layer `Embedding.as_linear`). So I'm ok with the option of just passing to all modules (but maybe only if it's `True` to avoid breaking the API for others).

If we don't intend to have all modules accept the parameter then ideally we should generalize the `is_embedding` check. Some options there are:
- Inspect the signature for the kwarg `quantize_input`. Hacky.. but maybe not too bad.
- Have a different method `to_qq` (or something) and check for it's existence

angeloskath:
Hadn't thought of `Embedding.as_linear` at all. Good point.

I think passing it all the time makes sense. We did break the api for `mode` it doesn't seem that this will keep happening so let's do it.

python/mlx/nn/layers/quantized.py
nastya236: Ah of course.. thank you!

python/mlx/nn/layers/quantized.py
nastya236: If I am not mistaken, it is not a bug because of  `if bool_or_params := class_predicate(path, m):`

angeloskath: Very good point indeed.

nastya236:
> For the embedding it would only be used when you call it as a linear layer `Embedding.as_linear`

Good point. Do you think that it is needed? I had a feeling that typically embedding stays in `bf16` for convergence. If we do want to quantize it, then we probably need either:
- A new module (something like `QQEmbedding`) that uses `qqmm` for `as_linear`. And we will change `Embedding.to_quantize(quantize_input= quantize_input)`
- Alternatively, we can use `quantize_input` in `Embedding.as_linear` directly, but it feels wrong..
- We skip embedding layers if `quantize_input=True` (at least for now).

Did you have some other option in mind?

awni:
> We skip embedding layers in quantize_input=True (at least for now).

I think that option is perfectly fine for now. 

The main question is if we intend to require modules handle the argument `quantize_input` themselves in which case it's on the user to make sure they only request certain modules to use `quantize_input`, or do we choose for the user by only passing it to certain modules (e.g. `nn.Linear`).

I maybe just slightly prefer the first version. So in that version `nn.Embedding.to_quantized` should take `quantize_input`. And maybe at some point we implement `QQEmbedding` but also fine to just raise an exception if it's `True` for now. 

The alternative would be to only pass `quantize_input` to modules that we choose. It's more opinionated and less flexible. But also probably easier to use for now.

I'm really okay with either way... we can always adapt later. It's quite a new feature so it's ok to be flexible based on how it's used.

python/mlx/nn/layers/embedding.py
```diff
@@ -45,6 +45,9 @@ def to_quantized(
         group_size: Optional[int] = None,
         bits: Optional[int] = None,
         mode: str = "affine",
+        quantize_input: bool = False,
     ):
         """Return a :obj:`QuantizedEmbedding` layer that approximates this embedding layer."""
+        if quantize_input:
+            raise ValueError("[Embedding] Quantized input is not supported.")
```
awni: Style nit: python gives helpful back traces so we don't really need to tag error messages like that.
```suggestion
            raise ValueError("Quantized input is not supported.")
```

python/mlx/nn/layers/linear.py
```diff
@@ -75,8 +75,36 @@ def to_quantized(
...
+            QuantizedLinear or QQLinear: A quantized version of this layer.
+
+        Notes:
+            Quantized input is only supported for ``"nvfp4"`` and ``"mxfp8"``
+            modes.
+        """
+        if quantize_input:
+            if mode not in ["nvfp4", "mxfp8"]:
+                raise ValueError(
+                    f"[Linear] Quantized activations are only supported for 'nvfp4' and 'mxfp8' modes, got {mode}."
```
awni:
```suggestion
                    f"Quantized activations are only supported for 'nvfp4' and 'mxfp8' modes, got {mode}."
```

awni:
This is very nice! Minor request: would you mind adding a simple test [in this file](https://github.com/ml-explore/mlx/blob/main/python/tests/test_nn.py) to make sure that quantizing a model with `quantize_input` produces QQ layers as expected?

awni (approved):
Looks good to go other than missing test! I'll run the tests after you add a test and then we can merge when they clear.

PR #3108, opened 2026-02-06
Fix residency set with user provided buffer

awni:
Make sure to remove the buffer from the residency set.


awni:
Thanks @davidkoski for debugging that.

angeloskath (approved):
👍

PR #3110, opened 2026-02-07
[CUDA] Set current device before allocating memory

zcbenz:
This is a followup to https://github.com/ml-explore/mlx/pull/3100 that fixes illegal memory access when using MLX under multi-device environment.


awni (approved):
Nice! Thanks for fixing that!

PR #3113, opened 2026-02-08
[CUDA] Use cuDNN SDPA for decoding when using fixed-size KV cache

zcbenz:
The kv cache in mlx-lm has fixed-size and keys/values passed to `fast.sdpa` are slices, using this information we can create cuDNN graphs with fixed sequence length and use padding masks to set the actual sequence lengths.

This makes decoding (`T_q == 1`) 30%~100% faster for large sequence lengths. For small sequence lengths the cuDNN SDPA is still faster but the overhead of creating cuDNN graphs would eliminate the advantage so we fallback to vector SDPA.

This approach however does not support custom array masks, we can add more options to reduce the uses of array masks but cuDNN does not support left padding masks, so integrating with `BatchKVCache` would require extra efforts.

### DIFF
--- a/mlx/backend/cuda/scaled_dot_product_attention.cpp
+++ b/mlx/backend/cuda/scaled_dot_product_attention.cpp
@@ -57,6 +57,58 @@ void malloc_with_same_layout(
       {true, false, false});
 }
 
+bool use_cudnn_for_decoding(
+    const array& q,
+    const array& k,
+    const array& v,
+    bool has_arr_mask) {
+  if (q.shape(2) != 1) {
+    return false;
+  }
+  if (has_arr_mask) {
+    return false;
+  }
+  // The cuDNN SDPA is faster than vector kernel but for small sequence the
+  // overhead would kill the advantage.
+  constexpr int kv_cache_step = 256; // number is from mlx-lm
+  if (k.shape(2) < kv_cache_step) {
+    return false;
+  }
+  // When called during graph building the strides is not available, and we
+  // rely on |supports_sdpa_vector| to decide whether to use fast sdpa since
+  // we can fallback to |sdpa_vector|.
+  if ((k.status() != array::evaluated) || (v.status() != array::evaluated)) {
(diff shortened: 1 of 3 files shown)


mlx/backend/cuda/scaled_dot_product_attention.cpp
```diff
@@ -57,6 +57,57 @@ void malloc_with_same_layout(
...
+    if (kv.status() != array::evaluated) {
+      return (kv.has_primitive() && typeid(kv.primitive()) == typeid(Slice)) ||
+          (kv.shape(2) % kv_cache_step == 0);
+    }
+    // Get pre-sliced sequence length from strides, and check if the buffer
+    // belongs to a contiguous kv cache.
+    int64_t T_kv = kv.strides(1) / kv.strides(2);
+    if (kv.size() / kv.shape(2) * T_kv != kv.buffer_size() / kv.itemsize()) {
+      return false;
+    }
```
awni, line 90: If we assume that the cudnn case is a strict subset of what the vector fallback can handle then this is not really necessary. We could just use this function during low level dispatch (and actually use the strides which is a lot more robust).

zcbenz: Strictly speaking cuDNN supports more head dims, but yeah it should be more robust just making it a subset of vector sdpa. I have updated the code.

awni:
Very nice improvement. On B200 it's more than 2x speedup for long context:

```
mlx_lm.benchmark --model Qwen/Qwen3-4B-Instruct-2507 --prompt 64000 --g 512 -n 3
```

Pre: 72.998 Tok/s
Post: 208.822 Tok/s

I wonder if we should defensively increase the default cache size for the forward op? If you do long generations it will start to thrash after about 16k tokens which is not that many. Maybe we should make it 256 or something?

zcbenz:
Yeah for decoding we definitely need a larger cache size.

Issue #2846, opened 2025-11-27
[BUG] metal_capture does not capture working GPU trace in MacOS 26+ On Mac Mini M1


mrcodechef:
**Describe the bug**
metal_capture does not capture a working GPU trace on MacOS 26.1
It generates the file mlx_trace.gputrace, but opening it results in  multiple errors

**To Reproduce**
With Xcode 26.1.1 on MacOS 26.1 on MacMini M1:
```bash
git clone https://github.com/ml-explore/mlx.git
mkdir build && cd build
cmake .. -DMLX_METAL_DEBUG=ON -G Xcode
open mlx.xcodeproj
```
Enable 'Automatically managing signing' and set Team in 'Signing & Capabilities' for metal_capture
Run metal_capture
```bash
open examples/cpp/Debug/mlx_trace.gputrace
```
enable 'Profile after replay' IMPORTANT: (You won't see the errors unless you enable this, or click the Profile button (in the Performance section) afterwards)
click replay
'Unexpected Replayer Termination' see below [1] for error text

// secondary error, likely partially resolved if the above is fixed. Debugging (somewhat) seems to work if you don't try to profile first. MacOS 15.7 also gives the 'No Function Found' error, so likely a seperate issue.

Switch to Group by Pipeline State
Open the first Compute Pipeline
Click arrangefloat32
Right click in shader source and select 'Debug Function...'
Click OK on 'No Function Found' error
Click Debug
'No Replay Device' see below [2] for error text

[1]
```
=== Crash Info ===
Signal: SIGABRT

=== Error summary: ===
1x DYErrorDomain (512) - guest app crashed (512)
1x com.apple.gputools.MTLReplayer (150) - Metal object creation failed.
25x com.apple.gputools.MTLReplayer (200) - The replayer encountered a Metal validation layer error.
1x com.apple.gputools.MTLReplayer (100) - Abort trap: 6

=== First Error ===
Domain: com.apple.gputools.MTLReplayer
Error code: 150

Description: Metal object creation failed.

=== Setup ===
Capture device:	Macmini9,1 - macOS 26.1 (25B78)
Replay device:	Macmini9,1 - macOS 26.1 (25B78)
Host:			Macmini9,1 - macOS 26.1 (25B78)
Tool:			Xcode (17B100)

Known SDKs:
```

[2]
```text
=== First Error ===
Domain: GPUShaderDebuggerTraceErrorDomain
Error code: 0

Description: Internal error: No replay device

[...]

PR #3114, opened 2026-02-08
Prevent allocator from operating on residency set during metal capture

robertmsale:
## Proposed changes

In issue #2846, it was reported that when profiling a `.gputrace` generated by MLX, Xcode could fail to rebuild the graph due to residency-set replay errors (`command queue residency set limit of 32 exceeded`).

This PR makes Metal capture replay-safe by preventing residency-set queue attachment/mutation while capture is active, and restoring normal behavior after capture ends.

### What changed

- Added capture-aware residency gating in the Metal backend:
  - Introduced `metal::residency_sets_enabled()` / `metal::set_residency_sets_enabled(bool)`.
  - Disabled residency sets at `start_capture(...)` and re-enabled them at `stop_capture()`.
- Updated device residency attachment flow to respect capture state:
  - Queue-level residency sets are detached at capture start.
  - Deferred re-attachment occurs after capture stop.
- Updated allocator wired-limit handling:
  - Residency-set resize/attach operations are deferred while capture is active.
  - Deferred state is applied on capture stop.
- Updated residency-set creation logic:
  - Avoid creating/attaching residency sets during active capture.
  - Keep allocations tracked and apply when safe.

These changes preserve normal runtime behavior outside capture while avoiding replay-time residency-set explosions during Xcode profiling/counters.

## Checklist

### DIFF
--- a/mlx/backend/metal/allocator.cpp
+++ b/mlx/backend/metal/allocator.cpp
@@ -52,8 +52,6 @@ MetalAllocator::MetalAllocator()
   block_limit_ = std::min(1.5 * max_rec_size, 0.95 * memsize);
   gc_limit_ = std::min(static_cast<size_t>(0.95 * max_rec_size), block_limit_);
   max_pool_size_ = block_limit_;
-  device(mlx::core::Device::gpu)
-      .set_residency_set(residency_set_.mtl_residency_set());
   bool is_vm = std::get<std::string>(info.at("device_name")) ==
       "Apple Paravirtual device";
   if (is_vm) {
@@ -96,10 +94,41 @@ size_t MetalAllocator::get_memory_limit() {
 
 size_t MetalAllocator::set_wired_limit(size_t limit) {
   std::unique_lock lk(mutex_);
-  std::swap(limit, wired_limit_);
+  size_t previous = wired_limit_;
+  wired_limit_ = limit;
+  // During active Metal capture, avoid issuing residency-set mutations so the
+  // capture remains replayable for derived counters.
+  if (!metal::residency_sets_enabled() || metal::is_capture_active()) {
+    wired_limit_pending_apply_ = true;
+    return previous;
+  }
+  // Attach the residency set lazily so normal execution (wired limit == 0)
+  // does not add queue residency-set state that is never used.
+  if (wired_limit_ > 0 && !residency_set_attached_ &&
+      residency_set_.mtl_residency_set() != nullptr) {
+    device(mlx::core::Device::gpu)
+        .set_residency_set(residency_set_.mtl_residency_set());
+    residency_set_attached_ = true;
+  }
   residency_set_.resize(wired_limit_);
-  return limit;
+  wired_limit_pending_apply_ = false;
+  return previous;
 };
+
+void MetalAllocator::on_capture_stop() {
+  std::unique_lock lk(mutex_);
+  if (!wired_limit_pending_apply_ || !metal::residency_sets_enabled()) {
+    return;
+  }
+  if (wired_limit_ > 0 && !residency_set_attached_ &&
+      residency_set_.mtl_residency_set() != nullptr) {
+    device(mlx::core::Device::gpu)
+        .set_residency_set(residency_set_.mtl_residency_set());
+    residency_set_attached_ = true;
+  }
+  residency_set_.resize(wired_limit_);
+  wired_limit_pending_apply_ = false;
+}
 
 Buffer MetalAllocator::malloc(size_t size) {
   // Metal doesn't like empty buffers
--- a/mlx/backend/metal/device.cpp
+++ b/mlx/backend/metal/device.cpp
@@ -370,13 +370,15 @@ void Device::new_queue(int index) {
     throw std::runtime_error(
         "[metal::Device] Failed to make new command queue.");
   }
+  attach_residency_set_to_existing_queues_if_needed_();
   stream_map_.emplace(index, q);
-  if (residency_set_ != nullptr) {
-    q->addResidencySet(residency_set_);
+  if (residency_set_ != nullptr && !metal::is_capture_active()) {
+    attach_residency_set_to_stream_(stream_map_.find(index)->second);
   }
 }
 
 MTL::CommandQueue* Device::get_queue(Stream stream) {
+  attach_residency_set_to_existing_queues_if_needed_();
   return get_stream_(stream.index).queue;
 }
 
@@ -387,6 +389,7 @@ bool Device::command_buffer_needs_commit(int index) {
 }
 
 MTL::CommandBuffer* Device::get_command_buffer(int index) {
+  attach_residency_set_to_existing_queues_if_needed_();
   auto& stream = get_stream_(index);
   if (stream.buffer == nullptr) {
     stream.buffer = stream.queue->commandBufferWithUnretainedReferences();
@@ -785,6 +788,28 @@ MTL::ComputePipelineState* Device::get_kernel(
       base_name, default_library_, hash_name, func_consts, linked_functions);
 }
 
+void Device::attach_residency_set_to_stream_(DeviceStream& stream) {
+  if (residency_set_ == nullptr || stream.residency_set_attached ||
+      !metal::residency_sets_enabled() || metal::is_capture_active()) {
+    return;
+  }
+  stream.queue->addResidencySet(residency_set_);
+  stream.residency_set_attached = true;
+}
+
+void Device::attach_residency_set_to_existing_queues_if_needed_() {
+  if (!residency_set_pending_attach_ || residency_set_ == nullptr) {
+    return;
+  }
+  if (!metal::residency_sets_enabled() || metal::is_capture_active()) {
+    return;
+  }
+  for (auto& [_, stream] : stream_map_) {
+    attach_residency_set_to_stream_(stream);
+  }
+  residency_set_pending_attach_ = false;
+}
+
 void Device::set_residency_set(const MTL::ResidencySet* residency_set) {
   if (residency_set_ != nullptr) {
     throw std::runtime_error(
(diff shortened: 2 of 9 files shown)


mlx/backend/metal/allocator.cpp
```diff
@@ -96,11 +94,42 @@ size_t MetalAllocator::get_memory_limit() {
...
+    return previous;
+  }
+  // Attach the residency set lazily so normal execution (wired limit == 0)
+  // does not add queue residency-set state that is never used.
+  if (wired_limit_ > 0 && !residency_set_attached_ &&
+      residency_set_.mtl_residency_set() != nullptr) {
+    device(mlx::core::Device::gpu)
+        .set_residency_set(residency_set_.mtl_residency_set());
+    residency_set_attached_ = true;
+  }
```
robertmsale: Attach queue-level residency lazily at first non-zero limit:
- Production fast path with wired_limit == 0 avoids queue-level residency state entirely.
- Capture path can safely defer this and replay later.

mlx/backend/metal/allocator.cpp
```diff
@@ -96,11 +94,42 @@ size_t MetalAllocator::get_memory_limit() {
...
-  return limit;
+  wired_limit_pending_apply_ = false;
+  return previous;
 };
 
+void MetalAllocator::on_capture_stop() {
+  std::unique_lock lk(mutex_);
+  if (!wired_limit_pending_apply_ || !metal::residency_sets_enabled()) {
+    return;
+  }
```
robertmsale: Nothing to do when:
- No deferred request exists.
- Residency mode is still disabled (defensive check for failed capture startup/teardown order).

mlx/backend/metal/device.cpp
```diff
@@ -794,10 +819,35 @@ void Device::set_residency_set(const MTL::ResidencySet* residency_set) {
...
+    return;
+  }
+  // Avoid queue-level residency state in captures; this can break
+  // derived-counter replay in Xcode.
+  for (auto& [_, stream] : stream_map_) {
+    if (!stream.residency_set_attached) {
+      continue;
+    }
+    stream.queue->removeResidencySet(residency_set_);
+    stream.residency_set_attached = false;
```
robertmsale: Core replay fix:
remove all queue-level residency attachments before capture begins.

Why:
- Counter/profiling replay may re-run command queue state setup and hit queue residency-set limits.
- Detaching up-front avoids encoding queue-level addResidencySet actions in the captured stream.

Runtime impact:
- This only runs for capture sessions (debug/profiling path).
- Normal production execution keeps queue attachments as before.

awni:
Residency sets are off by default so I'm slightly confused. Did you have them turned on when  capturing the workload (e.g. did you set `mx.set_wired_limit`)?

awni:
Oh I guess `addResidencySet` is the issue (which we always do even if it is empty). However, I'm not crazy about adding this complexity especially without knowing why. We should only ever add one residency set to the command queue. The fact that it's adding a lot during replay is quite strange 🤔

robertmsale:
This confused me too! In this repro, I did not call `mx::set_wired_limit(...)`.

The workload is the stock capture example in `examples/cpp/metal_capture.cpp`:
- capture starts at line 14 (`mx::metal::start_capture(...)`)
- then it creates GPU streams at lines 19-20 (`new_stream(mx::Device::gpu)`)

My takeaway was backend behavior during capture/replay causing those errors.

What was happening pre-fix:
1. The Metal allocator path eagerly registered queue residency state during backend setup (even when effective wired limit was 0 / not explicitly set by user).
2. When streams/queues were created, queue-level residency-set operations could still be recorded during capture.
3. Xcode counter replay replays command queue state and eventually hit the per-queue residency-set limit (`addResidencySet` limit exceeded), causing graph/counter profiling failures.

What changed:
- I removed eager constructor-time queue residency attachment and made attachment lazy on first non-zero wired limit request.
- During capture, residency-set mutations/attachments are gated off, and existing queue attachments are detached before capture to keep trace replay stable.
- After capture stops, deferred state is restored.

I tried to piece it together in a way that production workloads do not touch any of the bookkeeping and release builds can optimize out those branches if nobody does Metal GPU capture. I would like to point out that I tried many different strategies in `metal_capture.cpp`, in the hopes that maybe if that test is written a certain way we do not hit the replay issue and wouldn't require any working code edits (e.g. creating streams and running eval before capture start), but to no avail. Freezing the residency set edits during capture is what solved it for me.

robertmsale:
>  We should only ever add one residency set to the command queue. The fact that it's adding a lot during replay is quite strange 🤔

I'm just speculating here, but I think the problem happens when the profiler does derived counter analysis. MLX does indeed only have 1 residency set, everything we're doing is above board. But during derived counter analysis the GPUTools replays passes and if we have any residency attachment operations during capture, it re-issues queue residency attachment over and over because it happened even one time during capture. We hit a limit of 32 adds, and get 24 errors. I think the pattern is we have 3 GPU queues: default stream multiply, s2, s3. The 56 residency sets comes from 3 queues and number of replay counter passes. Some passes touch all queues, some only touch s2, some s3, and you end up with 56 passes total from that.

So I think this PR is mostly a workaround for the GPU profiling limitation and is not indicative of a deeper issue, but I can't debug the profiler 😅

robertmsale:
I'm converting this to draft because, after thinking about this, I think this is a workaround for a bug in the GPU profiler. Something didn't feel right about it.

I wrote a toy benchmark:
```swift
import Foundation
import Metal

enum ResidencySetLimitError: Error {
    case noDevice
    case noQueue
    case noCommandBuffer
    case noEncoder
    case noOutputBuffer
    case invalidSetCount(String)
}

@available(macOS 15.0, *)
func makePipeline(device: MTLDevice) throws -> MTLComputePipelineState {
    let source = """
    #include <metal_stdlib>
    using namespace metal;

    kernel void write_index(device uint *out [[buffer(0)]],
                            uint gid [[thread_position_in_grid]]) {
        out[gid] = gid;
    }
    """

    let library = try device.makeLibrary(source: source, options: nil)
    guard let function = library.makeFunction(name: "write_index") else {
        fatalError("Could not build runtime kernel function.")
    }
    return try device.makeComputePipelineState(function: function)
}

@available(macOS 15.0, *)
func runExperiment(device: MTLDevice, queue: MTLCommandQueue, setCount: Int) throws {
    let descriptor = MTLResidencySetDescriptor()
    var sets: [any MTLResidencySet] = []
    sets.reserveCapacity(setCount)

    print("Adding \(setCount) residency sets to one command queue...")
    for i in 0..<setCount {
        let set = try device.makeResidencySet(descriptor: descriptor)
        set.requestResidency()
        queue.addResidencySet(set)
        sets.append(set)
        print("  added set \(i + 1)")
    }

    let pipeline = try makePipeline(device: device)
    let elementCount = 64
    let outputSize = elementCount * MemoryLayout<UInt32>.stride

    guard let output = device.makeBuffer(length: outputSize, options: .storageModeShared) else {
        throw ResidencySetLimitError.noOutputBuffer
    }

    guard let commandBuffer = queue.makeCommandBuffer() else {
        throw ResidencySetLimitError.noCommandBuffer
    }
    guard let encoder = commandBuffer.makeComputeCommandEncoder() else {
        throw ResidencySetLimitError.noEncoder
    }

    encoder.setComputePipelineState(pipeline)
    encoder.setBuffer(output, offset: 0, index: 0)

    let width = min(pipeline.maxTotalThreadsPerThreadgroup, elementCount)
    let tg = MTLSize(width: width, height: 1, depth: 1)
    let grid = MTLSize(width: elementCount, height: 1, depth: 1)
    encoder.dispatchThreads(grid, threadsPerThreadgroup: tg)
    encoder.endEncoding()

    commandBuffer.commit()
    commandBuffer.waitUntilCompleted()

    let ptr = output.contents().bindMemory(to: UInt32.self, capacity: elementCount)
    let first = (0..<8).map { ptr[$0] }
    print("Kernel completed. First 8 outputs: \(first)")

    // Keep sets alive through command buffer completion.
    withExtendedLifetime(sets) {}
}

func parseSetCount(_ args: [String]) throws -> Int {
    guard let raw = args.dropFirst().first else {
        return 33
    }
    guard let value = Int(raw), value > 0 else {
        throw ResidencySetLimitError.invalidSetCount(raw)
    }
    return value
}

func main() throws {
    guard let device = MTLCreateSystemDefaultDevice() else {
        throw ResidencySetLimitError.noDevice
    }
    guard let queue = device.makeCommandQueue() else {
        throw ResidencySetLimitError.noQueue
    }

    let setCount = try parseSetCount(CommandLine.arguments)
    print("Device: \(device.name)")
    print("Set count: \(setCount)")

    if #available(macOS 15.0, *) {
        try runExperiment(device: device, queue: queue, setCount: setCount)
    } else {
        print("Residency sets require macOS 15.0+.")
    }
}

do {
    try main()
} catch ResidencySetLimitError.invalidSetCount(let raw) {
    fputs("Invalid set count: \(raw)\n", stderr)
    exit(2)
} catch {
    fputs("Experiment failed: \(error)\n", stderr)
    exit(1)
}
```

I wanted to see if I could hit a runtime assertion by adding 33 residency sets so I ran it with 1 set, and ran it with 33. The run with 33 produced the same exact error as was present in the GPU Tools profiler. Profiling with no residency sets would yield results that are nowhere near production (e.g. cache evictions that would never happen if memory was wired down) so this workaround would be preventing real profiling and smart code decisions from being made.

This leads me to believe that this is actually a bug in Xcode's GPU profiling. If running that `metal_capture` program didn't hit the assertion, then that is sufficient evidence to prove MLX is doing everything correctly, and it is in fact the tooling that is broken. I am considering closing this PR, but keeping it in draft for right now so the information is out there. If MLX is interested in the workaround I can open it back up, or we can close it if the goal is to wait for the Xcode development team to fix the profiler, but I am confident this is a profiler issue. @awni what would you like me to do?

Test run:
```
Device: Apple M1 Ultra
Set count: 33
Adding 33 residency sets to one command queue...
  added set 1
  added set 2
  added set 3
  added set 4
  added set 5
  added set 6
  added set 7
  added set 8
  added set 9
  added set 10
  added set 11
  added set 12
  added set 13
  added set 14
  added set 15
  added set 16
  added set 17
  added set 18
  added set 19
  added set 20
  added set 21
  added set 22
  added set 23
  added set 24
  added set 25
  added set 26
  added set 27
  added set 28
  added set 29
  added set 30
  added set 31
  added set 32
EXIT:134
_addResidencySets:453: failed assertion `IOGPUMetalCommandQueue: command queue residency set limit of 32 exceeded'
```

awni:
I think a simpler work-around would be to change [here](https://github.com/ml-explore/mlx/blob/main/mlx/backend/metal/resident.cpp#L8-L10) and add a check for if capturing is enabled it's in an M1 GPU  then simply don't use residency sets.

But if you are not actively blocked by this then I would also leave this in draft and hopefully the underlying bug will be fixed.

zcbenz:
I'm closing this PR since the related code had been refactored a lot that it would need a lot of changes, and per the comments above we are not going to merge it.

Issue #3086, opened 2026-01-31
[BUG] `mx.quantized_matmul` is slower than fp16 matmul for batch sizes 10-14


zhijian-liu:
**Describe the bug**

`mx.quantized_matmul` is slower than fp16 matmul for batch sizes 10-14.

**To Reproduce**

```python
import mlx.core as mx
import time

def bench(fn, warmup=100, runs=500):
    for _ in range(warmup):
        mx.eval(fn())
    times = []
    for _ in range(runs):
        t0 = time.perf_counter()
        mx.eval(fn())
        times.append(time.perf_counter() - t0)
    times.sort()
    return times[runs // 2] * 1000

W = mx.random.normal((2560, 2560)) * 0.01
W_q = mx.quantize(W, bits=4, group_size=64)
W_dq = mx.dequantize(*W_q, group_size=64, bits=4)
mx.eval(W_q, W_dq)

for N in [1, 8, 10, 12, 14, 16, 32]:
    X = mx.random.normal((N, 2560))
    mx.eval(X)
    t_fp = bench(lambda: X @ W_dq.T)
    t_q = bench(lambda: mx.quantized_matmul(X, *W_q, bits=4, group_size=64))
    print(f"N={N:2d}: fp16={t_fp:.3f}ms, 4bit={t_q:.3f}ms, ratio={t_q/t_fp:.2f}x")
```

Output:
```
N= 1: fp16=0.183ms, 4bit=0.160ms, ratio=0.88x
N= 8: fp16=0.214ms, 4bit=0.199ms, ratio=0.93x
N=10: fp16=0.218ms, 4bit=0.258ms, ratio=1.19x
N=12: fp16=0.216ms, 4bit=0.258ms, ratio=1.19x
N=14: fp16=0.217ms, 4bit=0.404ms, ratio=1.86x
N=16: fp16=0.553ms, 4bit=0.337ms, ratio=0.61x
N=32: fp16=0.559ms, 4bit=0.261ms, ratio=0.47x
```

**Expected behavior**

4-bit `quantized_matmul` should be faster than fp16 matmul for all batch sizes due to reduced memory bandwidth.

**Desktop (please complete the following information):**
- OS Version: macOS 15.3
- MLX Version: 0.30.3
- Hardware: M2 Max

**Additional context**

The regression occurs specifically for batch sizes 10-14. Batch sizes 1-8 and 16+ work as expected (4-bit is faster).

angeloskath:
[...]

PR #3120, opened 2026-02-12
[Metal][Performance]: Add split-K for quantized matmul (small M)

Ziqiao-git:
### Proposed changes
In issue #3086, it was observed that the quantized `qmm` kernel severely underutilizes the GPU for small `M` (e.g., `M=12-32`). For example, a configuration of `D=2560` and `M=12` yields only 80 threadgroups (assuming `BM=BN=32`), which is insufficient to saturate the GPU grid.

This PR introduces a split-K variant (`qmm_t_splitk`) that partitions the `K` dimension across multiple threadgroups. This safely improves GPU occupancy and execution speed for small-batch inference scenarios, while falling back to the standard kernel for larger batches to prevent any performance regression.

### What changed
* Added a split-K variant of the quantized matrix multiplication kernel (`qmm_t_splitk`) in the Metal backend, conceptually similar to the existing `fp16` `steel_gemm_splitk`.
* Updated the dispatch logic in `quantized.cpp` to dynamically calculate the split factor, targeting ~512 threadgroups for optimal occupancy.
* Added a fallback mechanism that automatically routes to the regular `qmm` kernel when `split_k <= 1` (e.g., for large `M`).
* Verified performance gains on Apple M3 Max (4-bit, `group_size=64`):
  * `D=2560, M=12`: 0.079ms -> 0.055ms (~30% faster)
  * `D=4096, M=16`: 0.155ms -> 0.117ms (~25% faster)
  * No regressions observed for large `M` configurations.
* Verified correctness: Existing tests cover the new code path (27 tests, 1639 subtests pass).

### Checklist

### DIFF
--- a/mlx/backend/metal/quantized.cpp
+++ b/mlx/backend/metal/quantized.cpp
@@ -792,6 +792,9 @@ void qmm_splitk(
   int current_tgs = n_tiles * m_tiles;
   int split_k = std::max(1, 512 / current_tgs);
 
+  // Cap split_k by the number of quantization groups
+  split_k = std::min(split_k, K / group_size);
+
   // Ensure K divides evenly by split_k * group_size
   while (split_k > 1 && (K % (split_k * group_size) != 0)) {
     split_k--;


Ziqiao-git:
As a quick note on why these specific dimensions matter: The performance bottleneck for small $M$ sizes ($12 \sim 32$) directly impacts the verification step in Speculative Decoding. 
By fixing the GPU underutilization here, we significantly speed up the time it takes to evaluate draft tokens. Given how important speculative decoding is for pushing the limits of inference speed on edge devices, this change should provide a meaningful boost to overall generation latency, making the backend more robust for future speculative decoding implementations.

angeloskath:
Thanks that is great! I 'll take a look asap.

Ziqiao-git:
Addressed all feedback:

1. Removed unused `qmv_split_k` code (impl, kernel, macro, and dispatch function)
2. Removed `qmm_t_splitk_impl` — `affine_qmm_t_splitk` now pre-offsets pointers and calls `qmm_t_impl` directly (added `K_eff` parameter for loop bound, following the `qvm_splitk` pattern)
3. Added fp quantization support: `fp_qmm_t_impl` also takes `K_eff`, new `fp_qmm_t_splitk` kernel wrapper, and instantiation macros in `fp_quantized.metal`

Build passes, pre-commit clean, benchmark confirms split-K working correctly for both affine and fp paths.

<details>
<summary><b>Benchmark Results (M1/M2/M3 / applegpu_g15s)</b></summary>
Device: applegpu_g15s
Memory: 52 GB

qmv_batch_limit(D=4096, O=4096) = 12

============================================================
D=4096  mode=mxfp8 (bits=8, group_size=32)
   M      fp16     quant   ratio            fp16 kernel      quant kernel
----  --------  --------  ------  --------------------  ---------------
   1    0.475ms    0.036ms   0.08x                  gemv              qmv
   2    0.519ms    0.061ms   0.12x               split-K              qmv
   4    0.533ms    0.111ms   0.21x               split-K              qmv
   8    0.516ms    0.212ms   0.41x               split-K              qmv
  10    0.520ms    0.262ms   0.50x               split-K              qmv
  12    0.518ms    0.121ms   0.23x               split-K       qmm_splitk 
  14    0.521ms    0.122ms   0.23x               split-K       qmm_splitk
  16    0.515ms    0.122ms   0.24x               split-K       qmm_splitk
  20    0.584ms    0.121ms   0.21x               split-K       qmm_splitk
  32    0.526ms    0.125ms   0.24x               split-K       qmm_splitk
2048    6.318ms    6.646ms   1.05x          regular GEMM              qmm

(Tested across affine, mxfp8, and mxfp4 - truncated for brevity but all show similar smooth transitions).

</details>

Let me know if there is anything I missed.

mlx/backend/metal/kernels/fp_quantized.h
```diff
@@ -642,6 +642,7 @@ METAL_FUNC void fp_qmm_t_impl(
     const constant int& K,
     const constant int& N,
     const constant int& M,
+    int K_eff,
```
jagrit06: Nit: Change to match convention 
```suggestion
    const constant int& K_eff,
```

mlx/backend/metal/kernels/quantized.h
```diff
@@ -1102,6 +1102,7 @@ METAL_FUNC void qmm_t_impl(
     const constant int& K,
     const constant int& N,
     const constant int& M,
+    int K_eff,
```
jagrit06: Nit: Change to match convention 
```suggestion
    const constant int& K_eff,
```

mlx/backend/metal/quantized.cpp
```diff
@@ -771,6 +771,98 @@ void qmm(
...
+  int split_k = std::max(1, 512 / current_tgs);
+
+  // Ensure K divides evenly by split_k * group_size
+  while (split_k > 1 && (K % (split_k * group_size) != 0)) {
+    split_k--;
+  }
+  if (split_k <= 1) {
+    return qmm(
+        x, w, scales, biases, out, true, group_size, bits, M, N, K, d, s, mode);
+  }
```
jagrit06, line 805: Could you help me understand why this approach was chosen ? 
I can make some sense of 512 thread groups being the target (if the M or N are larger, we go to the regular matmul, if the K is larger still we just loop more inside the kernel), but would like to know how you ended up at that number 

Also, we do employ checks such that `K` is always divisible by `group_size` - so wouldn't the looping step just then come down to be equivalent to `split_k = std::max(split_k, K / group_size) `

jagrit06 (changes requested):
Requesting a few small changes, thanks for putting this together!

mlx/backend/metal/quantized.cpp
Ziqiao-git: For the 512 target, it was mostly an empirical choice trying to balance GPU occupancy with atomic reduction overhead. My mental math was that on the larger Apple Silicon chips (like the 76-core M2 Ultra), we generally need a handful of active threadgroups per core to effectively hide memory latency (roughly $76 \times 6 \approx 456$). So, aiming for ~512 felt like a safe baseline to ensure good saturation even on the high-end chips. On the flip side, I wanted to avoid pushing the split factor too high to keep the atomic add contention in check during the final accumulation, as over-splitting the K-dimension can quickly offset the occupancy gains.Ideally, this target should be a dynamic function of the active GPU core count. However, since dynamically querying the exact core count in Metal isn't straightforward, tuning for the high-end ceiling acts as a safe catch-all. The hardware scheduler on smaller chips (like the base M1/M2) handles the oversubscription of 512 threadgroups gracefully without much penalty, whereas under-subscribing an Ultra would be catastrophic for latency hiding. If MLX has a preferred internal way to dynamically scale split-K factors based on device tiers, I'm completely open to refactoring this!

Ziqiao-git: I originally avoided just using `split_k = std::min(split_k, K / group_size)` because we need to guarantee **quantization group alignment**, not just an upper bound. 

If we strictly use `std::min`, it might result in a `split_k` that doesn't perfectly divide `K` into multiples of `group_size`. For example, if `K=1024`, `group_size=64`, and the calculated `split_k` is `12`, the chunk size `1024 / 12` does not align with `64`(the quant group). If a threadgroup's chunk starts unaligned with `group_size`, it will read misaligned `scale` and `bias` values, leading to incorrect numerical results. 

The `while` loop ensures we step down until `(K / split_k) % group_size == 0` so that every threadgroup boundary perfectly aligns with the quantization blocks. Let me know if you think there is a more elegant way to enforce this modulo alignment mathematically instead of the loop!

mlx/backend/metal/kernels/quantized.h
Ziqiao-git: addressed

mlx/backend/metal/kernels/fp_quantized.h
Ziqiao-git: addressed

mlx/backend/metal/quantized.cpp
jagrit06: Good point

I think the way around this would be support unaligned `K_eff` for the last partition (that logic doesn't exist in the QMMs but does in the regular MM code) - we certainly don't want it to be the case that you end up with a `K` that divides to 31 quantized groups, but doesn't end up dispatched to the split-K variant because 31 is a prime and we require the number of splits to perfectly divide `K / group_size` 

That said, we can merge this for now and do that in a follow up PR

mlx/backend/metal/quantized.cpp
```diff
@@ -770,6 +771,98 @@ void qmm(
...
+    int K,
+    metal::Device& d,
+    const Stream& s,
+    const std::string& mode) {
+  // Choose split_k to target ~512 threadgroups
+  int bm = 32, bn = 32;
+  int n_tiles = (N + bn - 1) / bn;
+  int m_tiles = (M + bm - 1) / bm;
+  int current_tgs = n_tiles * m_tiles;
+  int split_k = std::max(1, 512 / current_tgs);
```
jagrit06, line 793: For now, just so we are safe, could you add a check here to make sure that K is large enough (in relation to M and N) to warrant going through the loop of splits ? 
We naturally short-circuit if `n_tiles * m_tiles >= 512` so large `M` and `N` are covered - it would be good to similarly short circuit out if the K isn't too large compared to the M and N. 

Till we have a fix for the loop, I would like to avoid to a `64x64x128` matmul as an example, going through a 100 iterations where it could have just short circuited earlier 

After that, we should be ready to merge!

Ziqiao-git: Addressed: added split_k = std::min(split_k, K / group_size) before the alignment loop to cap split_k by the actual number of quantization groups. This avoids unnecessary iterations when K is small relative to M and N (e.g., 64×64×128 would now short-circuit immediately instead of looping ~126 times). Sorry for the late reply!

Thump604:
This directly benefits MTP speculative decoding verification on MoE models — exactly the small-M quantized matmul pattern where GPU underutilization is worst.

I run Qwen3.5-122B (5-bit MoE, 10B active) with MTP on M2 Ultra. The verification step does batched quantized matmuls at M=8-16 where this split-K work would help most. I also have SpecPrefill (attention-based sparse prefill) which uses the draft model for Q@K^T scoring at similar small batch sizes.

Both reviewers have approved and the benchmarks look solid. Happy to provide M2 Ultra numbers once this lands. Would be great to see this merged — it's been approved since early March.

Thump604:
Gentle nudge — this has been approved by both @angeloskath and @jagrit06 for 10 days now with all CI green. Is there anything blocking the merge? Happy to help with any remaining benchmarking if needed.

angeloskath:
Sorry, nothing was really keeping us from merging it, I thought Jagrit would merge it since he was the last to review (perhaps he thought I would merge it 😅)

PR #3121, opened 2026-02-12
Fix donation in sdpa vector

angeloskath:
Sneaky bug that would only show up when vector sdpa would donate the queries to the input and also be large enough to warrant a command encoder commit.

The bug is that arrays registered as temporaries should not span command encoders because they are removed from the list of outputs we keep track of.

I also deleted a few outdated checks for donation because they are pointless since we changed moving the buffer to copying it.

Fixes the issue in https://github.com/ml-explore/mlx-lm/pull/868 .


awni (approved):
Very nice find!!

PR #3123, opened 2026-02-13
feat: add __version__/__doc__/__all__ attributes to top-level package

Vlor999:
## Description
This PR improves the top-level `mlx` package to better align with standard Python practices and libraries like NumPy.

It adds three main features to `python/mlx/__init__.py`:

1.  **`__version__`**: Now dynamically retrieved using `importlib.metadata`. This allows users to programmatically check the version (`mlx.__version__`) without hardcoding it in the source.
2.  **`__doc__`**: Adds a module-level docstring so `help(mlx)` displays a proper description of the framework.
3.  **`__all__`**: Explicitly defines the public API. It exposes submodules (`nn`, `optimizers`, `utils`) and flattens the `core` functions (like `array`, `matmul`) to support safe `from mlx import *` usage.

## Checklist

- [x] I have updated the necessary documentation

### DIFF
--- a/python/mlx/__init__.py
+++ b/python/mlx/__init__.py
@@ -1,0 +1,34 @@
+"""
+MLX: An array framework for Apple silicon.
+
+MLX is an array framework for machine learning on Apple silicon, brought
+to you by the Apple machine learning research team.
+
+It features:
+- A familiar API (similar to NumPy and PyTorch).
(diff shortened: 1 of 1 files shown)


awni:
I think we want to avoid having __init__ to keep mlx as a namespace package.

PR #3124, opened 2026-02-13
[featuring] - add hanning window function

Vlor999:
## Description

This PR implements the Hanning window function (`mlx.core.hanning`), bringing MLX closer to feature parity with NumPy's signal processing capabilities (`numpy.hanning`).

Window functions are essential for audio processing and various signal analysis tasks.

## Implementation Details

- **C++**: Implemented `hanning` in `mlx/ops.cpp` using existing primitives (`arange`, `cos`, `multiply`, `subtract`). Used `M_PI` from `<cmath>` for precision.
- **Python Bindings**: Exposed the function via pybind11 in `python/src/ops.cpp` under `mlx.core`.
- **Documentation**: Added docstrings describing the parameters and return values, matching the NumPy behavior.

## Test Plan

I have verified the implementation locally against NumPy to ensure numerical accuracy.

**Verification script:**
```python
import mlx.core as mx
import numpy as np

M = 50
mx_hanning = mx.hanning(M)
np_hanning = np.hanning(M)

# Check for numerical closeness
assert np.allclose(mx_hanning, np_hanning, atol=1e-6), "Mismatch with NumPy implementation"
print("✅ Verification passed against numpy.hanning")
```
Unit Tests:

Added a test case in python/tests/test_ops.py covering standard usage, edge cases (M=1), and empty input (M=0).

## Checklist

### DIFF
--- a/mlx/ops.cpp
+++ b/mlx/ops.cpp
@@ -2311,6 +2311,19 @@ array argmax(
   return out;
 }
 
+array hanning(int M, StreamOrDevice s /* = {} */) {
+  if (M < 1) {
+    return array({});
+  }
+  if (M == 1) {
+    return ones({1}, float32, s);
+  }
+
+  auto n = arange(0, M, float32, s);
+  array factor(M_PI / (M - 1), float32);
+  return square(sin(multiply(factor, n, s), s), s);
+}
+
 /** Returns a sorted copy of the flattened array. */
 array sort(const array& a, StreamOrDevice s /* = {} */) {
   int size = a.size();
--- a/mlx/ops.h
+++ b/mlx/ops.h
@@ -666,6 +666,9 @@ min(const array& a,
 MLX_API array
 min(const array& a, int axis, bool keepdims = false, StreamOrDevice s = {});
 
+/** Returns the Hanning window of size M. */
+MLX_API array hanning(int M, StreamOrDevice s = {});
+
 /** Returns the index of the minimum value in the array. */
 MLX_API array argmin(const array& a, bool keepdims, StreamOrDevice s = {});
 inline array argmin(const array& a, StreamOrDevice s = {}) {
--- a/python/tests/test_ops.py
+++ b/python/tests/test_ops.py
@@ -1450,6 +1450,18 @@ class TestOps(mlx_tests.MLXTestCase):
         expected = [0]
         self.assertListEqual(a.tolist(), expected)
 
+    def test_hanning_general(self):
+        a = mx.hanning(10)
+        expected = np.hanning(10)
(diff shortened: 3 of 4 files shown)


mlx/ops.cpp
```diff
@@ -2311,6 +2311,28 @@ array argmax(
...
+  auto pi = array(M_PI, float32);
+  auto two = array(2.0f, float32);
+  auto m_minus_one = array(M - 1, float32);
+
+  auto arg = divide(multiply(multiply(two, pi), n), m_minus_one, s);
+
+  auto cos_vals = cos(arg, s);
+  auto half = array(0.5, float32);
+
+  return subtract(half, multiply(half, cos_vals, s), s);
```
angeloskath: Probably simpler to write as

```cpp
auto n = arange(0, M, float32, s);
array factor(M_PI / (M-1), float32);
return square(sin(multiply(factor, n, s), s), s);
```

mlx/ops.h
```diff
@@ -666,6 +666,17 @@ min(const array& a,
...
 min(const array& a, int axis, bool keepdims = false, StreamOrDevice s = {});
 
+/**
+ * Returns the Hanning window.
+ *
+ * The window is defined as:
+ * w(n) = 0.5 - 0.5 * cos(2 * pi * n / (M - 1))
+ *
+ * @param M The number of points in the output window.
+ * @param s The stream or device to run the operation on.
```
angeloskath: As in the rest of the code I would keep this comment simpler and expand the one in the python module. Something like Returns the Hanning window of size M would probably be enough for the C++ op. And in the python module you can define the implementation and even use some math like for example in the `erf` op.

python/tests/test_ops.py
```diff
@@ -1450,6 +1450,26 @@ def test_arange_corner_cases_cast(self):
...
+        # Edge case : M=1
+        a = mx.hanning(1)
+        self.assertEqual(a.item(), 1.0)
+
+    def test_hanning_case_0(self):
+        # Edge case : M=0 (Empty)
+        a = mx.hanning(0)
+        self.assertEqual(a.size, 0)
+
+    def test_hanning_case_typing(self):
```
angeloskath: You can merge the 4 tests in 1.

Vlor999:
Thanks for the feedback!

I've updated the implementation as discussed:
- **Optimization:** Switched to the $\sin^2$ identity (`square(sin(...))`). This reduces the compute graph size and memory usage by avoiding the intermediate arrays needed for the cosine formula.
- **Verification:** Verified that tests still pass and numerical results remain consistent with NumPy.

Ready for another look! 🚀

angeloskath (approved):
Looks good thanks!

PR #3127, opened 2026-02-13
Manage stream placement in import function

### LINKED ISSUE #3125: [BUG] Stream/Device index issue when export from cpu only device and import from device with gpu
**Describe the bug**
When exporting a cpu function from a **cpu-only device**, default stream is cpu with index 0
When importing the function on cpu on a **device with gpu available**, default stream is gpu with index 0
However, when importing, stream is only checked by it's index and not device type:
 1. https://github.com/ml-explore/mlx/blob/v0.30.6/mlx/export.cpp#L476
 2. https://github.com/ml-explore/mlx/blob/v0.30.6/mlx/stream.h#L28-L30
 
After import, cpu is loaded as the device for imported primitives with stream index 0 which will then cause a bad access when `Scheduler::enqueue` https://github.com/ml-explore/mlx/blob/v0.30.6/mlx/scheduler.h#L163 as thread_[0] is nullptr given stream index 0 is gpu and not cpu.

Thus stream should check for both index and device type in `==`

**Related Ask**
Would it make sense to always make cpu as the default stream with index 0 so we won't have the mismatch of cpu stream indices (0 on cpu-only device and 1 on gpu-available device) when exporting from a cpu-only device and importing from a gpu-available device?

awni:
Previously if you export a function on a stream that is not present at import then there would be a crash. This requires the user to make sure all the streams are present which is hard to manage and impossible in some cases as the defualt stream on a CPU-only machine is different than the CPU+GPU machine.

Closes #3125

### DIFF
--- a/mlx/export.cpp
+++ b/mlx/export.cpp
@@ -446,6 +446,7 @@ struct PrimitiveFactory {
       SERIALIZE_PRIMITIVE(ScaledDotProductAttention),
       SERIALIZE_PRIMITIVE(CustomKernel)};
   std::unordered_map<std::string, std::string> name_remap;
+  std::unordered_map<int, Stream> stream_map;
 
   PrimitiveFactory() {
     for (auto& [n, f] : factory) {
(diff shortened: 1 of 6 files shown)


angeloskath (approved):
But just to be clear this still means you can't export on cpu and import on gpu.

Btw I don't think you generally should be able to do that, just saying...

awni:
No that’s not intended to work.

awni:
But exporting for say the cpu in a cpu+gpu machine should probably work in a cpu only machine and this will fix that.

PR #3131, opened 2026-02-15
Fix CPU gather_mm index lifetime

booxter:
## Proposed changes

On CPU we schedule ops onto the stream thread, but GatherMM::eval_cpu
was reading index buffers on the host thread before dispatch. When those
indices come from earlier ops (e.g. arange/broadcast/argsort), the
buffers may not be initialized yet, producing sporadic garbage indices.

This has been observed in Darwin nixpkgs builds, e.g.:
https://hydra.nixos.org/build/321268968

```
  self.assertLess((y1 - y2).abs().max(), tol)
> self.assertLess((y1 - y3).abs().max(), tol)
E AssertionError: array(nan, dtype=float32) not less than 1.5e-05
  python/tests/test_quantized.py:980
```

This patch captures lhs/rhs indices by value in the dispatched task,
take the data pointers inside the task, and register the index arrays as
temporaries so their buffers stay alive until the task completes. This
removes the host‑side race without changing donation behavior.

## Checklist

### DIFF
--- a/mlx/backend/cpu/masked_mm.cpp
+++ b/mlx/backend/cpu/masked_mm.cpp
@@ -419,8 +419,8 @@ void GatherMM::eval_cpu(const std::vector<array>& inputs, array& out) {
     return decltype(v){v.begin(), v.end() - 2};
   };
 
-  auto& lhs_indices = inputs[2];
-  auto& rhs_indices = inputs[3];
+  auto lhs_indices = inputs[2];
+  auto rhs_indices = inputs[3];
 
   auto batch_shape = get_batch_dims(out.shape());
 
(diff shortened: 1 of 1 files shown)


angeloskath:
This change unfortunately doesn't change any behavior. We were keeping a reference to the data buffers anyway and the pointers will be the same inside and outside the lambda.

`add_temporary` is only for arrays that are created within the primitive's eval function. The inputs and outputs are taken care of automatically. In this case it doesn't really have any effect. It holds another reference to these buffers (on top of the one we were already holding).

Regarding the bug, I would advise to start an issue with a reproduction so we can investigate. In the meantime, I will close this.

PR #3133, opened 2026-02-15
fix: propagate quantization mode in QuantizedAllToShardedLinear / QuantizedShardedToAllLinear

### LINKED ISSUE #3132: [BUG] QuantizedAllToShardedLinear / QuantizedShardedToAllLinear don't propagate quantization `mode` — silent garbage output for MXFP8 tensor parallel
### Describe the bug

`QuantizedAllToShardedLinear` and `QuantizedShardedToAllLinear` in `mlx/nn/layers/distributed.py` do not accept, store, or pass the `mode` parameter to `mx.quantized_matmul`. When an MXFP8-quantized `QuantizedLinear` is converted via `shard_linear()`, the `mode` is silently lost. The resulting sharded layer calls `quantized_matmul` without `mode=`, which defaults to `"affine"` — interpreting FP8 packed weights as affine int8.

This produces **garbage output with no error message**, making it very hard to diagnose.

Additionally, MXFP8 does not use biases, but both classes unconditionally pass `biases=self["biases"]`, which raises `ValueError: [quantized_matmul] Biases must be null for quantization mode 'mxfp8'` once the mode fix is applied.

### To Reproduce

```python
import mlx.core as mx
import mlx.nn as nn
from mlx.nn.layers.distributed import shard_linear

# Simulate what happens when shard_linear converts an MXFP8 QuantizedLinear

# 1. Create a QuantizedLinear with mode="mxfp8" (as loaded from an MXFP8 model)
ql = nn.QuantizedLinear(64, 64, bias=False, group_size=32, bits=8)
ql.mode = "mxfp8"
# Requantize weight in mxfp8 mode
ql.weight, ql.scales, *_ = mx.quantize(
    mx.random.normal(shape=(64, 64)), group_size=32, bits=8, mode="mxfp8"
)
ql.freeze()

print(f"Original QuantizedLinear mode: {ql.mode}")  # "mxfp8"

# 2. Convert to sharded layer (this is what Exo/mlx-lm does for tensor parallel)
group = mx.distributed.init()
sharded = shard_linear(ql, "all-to-sharded", group=group)

print(f"Sharded layer type: {type(sharded).__name__}")
# QuantizedAllToShardedLinear

print(f"Sharded layer has mode: {hasattr(sharded, 'mode')}")
# False — mode was lost!

# 3. Forward pass silently uses mode="affine" instead of "mxfp8"
x = mx.random.normal(shape=(1, 64))
y = sharded(x)  # Uses quantized_matmul with default mode="affine"
mx.eval(y)
# Result is garbage because FP8 weights are interpreted as affine int8
```

### Expected behavior

`shard_linear` should propagate the `mode` parameter from the source `QuantizedLinear` to the resulting `QuantizedAllToShardedLinear` / `QuantizedShardedToAllLinear`, so that `quantized_matmul` uses the correct dequantization.

### Root cause

In `mlx/nn/layers/distributed.py`:

1. Neither `QuantizedAllToShardedLinear.__init__` nor `QuantizedShardedToAllLinear.__init__` accept a `mode` parameter. They store `group_size` and `bits` but not `mode`.

2. Both `__call__` methods call `mx.quantized_matmul(...)` without `mode=`:
```python
# Current code:
x = mx.quantized_matmul(
    x, self["weight"], scales=self["scales"],
    biases=self["biases"],  # also wrong for mxfp8
    transpose=True,
    group_size=self.group_size,
    bits=self.bits,
    # mode= is missing, defaults to "affine"
)
```

3. `from_quantized_linear()` copies `group_size` and `bits` but not `mode`.

### Suggested fix

```python
# In __init__ of both classes, add:
self.mode = mode  # new parameter, default "affine"

# In __call__ of both classes:
x = mx.quantized_matmul(
    x, self["weight"], scales=self["scales"],
    biases=None if self.mode == "mxfp8" else self.get("biases"),
    transpose=True,
    group_size=self.group_size,
    bits=self.bits,
    mode=self.mode,
)

# In from_quantized_linear of both classes:
sl = cls(
    ...,
    mode=getattr(quantized_linear_layer, "mode", "affine"),
)
```

### Impact

This blocks tensor parallel inference for **all MXFP8-quantized models**, including large MoE models on multi-node Apple Silicon clusters. We confirmed the fix works: GLM-5 754B (`mlx-community/GLM-5-8bit-MXFP8`, mode=mxfp8, group_size=32, bits=8) runs at ~14 tok/s on 2× M3 Ultra 512GB with tensor parallel after applying the fix.

The same issue likely affects `mxfp4` models if used with `shard_linear`.

### Desktop

- **MLX version:** 0.24.1 (via mlx-lm 0.30.7)
- **macOS:** 26.2 (Tahoe)
- **Hardware:** Apple M3 Ultra, 512 GB unified memory
- **Python:** 3.13

### Additional context

- Regular `QuantizedLinear` correctly passes `mode=` — single-node inference works fine
- The bug only manifests with distributed/sharded layers via `shard_linear`
- MXFP8 Metal kernel support was added recently (#2887), and the distributed layers were not updated
- Discovered while running GLM-5 754B on a 2-node Apple Silicon cluster via [Exo](https://github.com/exo-explore/exo)

vskiwi:
## Summary

Fixes #3132

`QuantizedAllToShardedLinear` and `QuantizedShardedToAllLinear` in `mlx/nn/layers/distributed.py` do not accept, store, or pass the `mode` parameter to `mx.quantized_matmul`. When an MXFP8-quantized `QuantizedLinear` is converted via `shard_linear()`, the `mode` is silently lost. The resulting sharded layer calls `quantized_matmul` without `mode=`, which defaults to `"affine"` — interpreting FP8 packed weights as affine int8, producing **garbage output with no error**.

Additionally, MXFP8 does not use biases, but both classes unconditionally accessed `self["biases"]`, which would raise `ValueError` once the mode fix is applied.

## Changes

- Add `mode` parameter (default `"affine"`) to both `__init__` methods
- Store `self.mode` and pass it to `mx.quantize` and `mx.quantized_matmul`
- Use `*biases` unpacking to handle modes that don't produce biases (mxfp8, mxfp4)
- Use `self.get("biases")` instead of `self["biases"]` for safe access (consistent with `QuantizedLinear`)
- Propagate `mode` from source layer in `from_quantized_linear`
- Include `mode` in `_extra_repr` output
- Add distributed test for mxfp8 quantized `shard_linear`

## Impact

This unblocks tensor parallel inference for **all MXFP8-quantized models** (and likely mxfp4). Confirmed working: GLM-5 754B (`mlx-community/GLM-5-8bit-MXFP8`, mode=mxfp8, group_size=32, bits=8) on 2× M3 Ultra 512GB at ~14 tok/s with tensor parallel.

No changes to the affine (default) code path — full backward compatibility.

## Test plan

- [x] Existing `test_shard_linear` test for affine quantization is unchanged and should still pass
- [x] New mxfp8 test in `test_shard_linear` verifies mode propagation, biases=None, and output correctness
- [x] Formatted with `black`

Made with [Cursor](https://cursor.com)

### DIFF
--- a/python/mlx/nn/layers/distributed.py
+++ b/python/mlx/nn/layers/distributed.py
@@ -371,6 +371,8 @@ class QuantizedAllToShardedLinear(Module):
             weight. See :func:`~mlx.core.quantize`. Default: ``64``.
         bits (int, optional): The bit width to use for the quantized weight.
             See :func:`~mlx.core.quantize`. Default: ``4``.
+        mode (str, optional): The quantization method to use (see
+            :func:`~mlx.core.quantize`). Default: ``"affine"``.
         group (mx.distributed.Group, optional): The sharding will happen across
             this group. If not set then the global group is used. Default is
             ``None``.
@@ -383,6 +385,7 @@ class QuantizedAllToShardedLinear(Module):
         bias: bool = True,
         group_size: int = 64,
         bits: int = 4,
+        mode: str = "affine",
         group: Optional[mx.distributed.Group] = None,
     ):
         super().__init__()
@@ -390,6 +393,7 @@ class QuantizedAllToShardedLinear(Module):
         # Quantization config
         self.group_size = group_size
         self.bits = bits
+        self.mode = mode
 
         # Initialize the quantized weight
         scale = math.sqrt(1.0 / input_dims)
@@ -406,7 +410,10 @@ class QuantizedAllToShardedLinear(Module):
             high=scale,
             shape=(output_dims // N, input_dims),
         )
-        self.weight, self.scales, self.biases = mx.quantize(weight, group_size, bits)
+        self.weight, self.scales, *biases = mx.quantize(
+            weight, group_size, bits, mode=mode
+        )
+        self.biases = biases[0] if biases else None
 
         # And bias if needed
         if bias:
@@ -427,7 +434,7 @@ class QuantizedAllToShardedLinear(Module):
         out_dims *= self.group.size()
         return (
             f"input_dims={in_dims}, output_dims={out_dims}, bias={'bias' in self}, "
-            f"group_size={self.group_size}, bits={self.bits}"
+            f"group_size={self.group_size}, bits={self.bits}, mode={self.mode}"
         )
 
     def __call__(self, x: mx.array) -> mx.array:
@@ -438,10 +445,11 @@ class QuantizedAllToShardedLinear(Module):
             x,
             self["weight"],
             scales=self["scales"],
-            biases=self["biases"],
+            biases=self.get("biases"),
             transpose=True,
             group_size=self.group_size,
             bits=self.bits,
+            mode=self.mode,
         )
         if "bias" in self:
             x = x + self["bias"]
@@ -465,6 +473,7 @@ class QuantizedAllToShardedLinear(Module):
             hasattr(quantized_linear_layer, "bias"),
             group_size=quantized_linear_layer.group_size,
             bits=quantized_linear_layer.bits,
+            mode=getattr(quantized_linear_layer, "mode", "affine"),
             group=group,
         )
         sl.update(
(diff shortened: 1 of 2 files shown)


angeloskath (approved):
Thank you that looks great!

I'll merge after the tests pass.

PR #3144, opened 2026-02-19
Fix fence synchronization accross command buffers

awni:
This fixes a bug @angeloskath found regarding race condition between updating a fence across two or more command buffers.


rltakashige:
Thank you for this change! Will test if this resolves the issue we were seeing.

Update: unfortunately not, but still an improvement! :)

PR #3145, opened 2026-02-19
Export: preserve Dtype state values in export callback arguments

skryl:
## Proposed changes

This PR fixes export callback state serialization for primitives that include `Dtype` in their state (for example `AsType`), so callback `arguments` preserve dtype values instead of dropping them.

Add explicit handling for Dtype during state unpacking:

```
else if constexpr (std::is_same_v<T, Dtype>) {
  unpacked_state.push_back(state);
}
```

It also adds a regression test to validate callback behavior and updates export docs with callback payload details.

### Problem

In mlx/export.cpp, the state unpacking logic did not handle Dtype in tuple/variant flattening.
As a result, primitives with dtype state (for example AsType, NumberOfElements) could emit incomplete arguments during export.

### Why this is needed

Downstream IR/export consumers rely on complete primitive arguments. Missing dtype state breaks robust graph export/lowering paths and forces downstream inference fallbacks.

## Checklist


awni (approved):
Thanks!

PR #3146, opened 2026-02-19
Fix np bfloat16 misinterpreted as complex

### LINKED ISSUE #1075: [BUG] `np.ndarray` of bfloat16 using ml_dtypes is being interpreted as complex64
**Describe the bug**

`np.ndarray` of bfloat16 using ml_dtypes is being interpreted as complex64 by mlx.

**To Reproduce**

```python
>>> from ml_dtypes import bfloat16
>>> import numpy as np
>>> x = np.array(1., dtype=bfloat16)
>>> import mlx.core as mx
>>> mx.array(x)
array(1+0j, dtype=complex64)
>>> x = np.array(1)
>>> mx.array(x)
array(1, dtype=int64)
>>> x = np.array(1., dtype=bfloat16)
>>> x.dtype
dtype(bfloat16)
>>> type(x.dtype)
<class 'numpy.dtype[bfloat16]'>
>>>
```

**Expected behavior**
Conversion to complex should not happen. Should remain as bfloat16

 - OS Version: [e.g. MacOS 14.1.2] MacOS 14.1.1
 - Version [e.g. 0.7.0] 0.11

**Additional context**
_Originally posted by @lkarthee in https://github.com/ml-explore/mlx/issues/1066#issuecomment-2089573368_

kellen-sun:
## Proposed changes

Fixes #1075
Bug: The bug happens when converting ``np.array(1., dtype=ml_dtypes.bfloat16)`` and ``np.array([1.], dtype=ml_dtypes.bfloat16)`` to ``mx.array(x)``. For the former case, it'll silently be caught as a ``std::complex`` as part of ``ArrayInitType`` and get converted as such (see related issue for code). For the latter, it'll be interpreted as an ``ArrayLike``, not be able to make the conversion to ``mx.array()`` and raise a ``ValueError``.
The Fix: We need to catch this case before it gets filtered by ``ArrayInitType``. I made the ``array.__init__`` more generic to catch this and checked the dtype to match bfloat16, then manually construct the array. Otherwise, we fallback to the original ``ArrayInitType`` case.
Note: bfloat16, is the only current ml_dtype that mlx supports.
Verification: Verified locally (macOS 26.2, MLX 0.30.7, Apple M2) with the additional test case I provided. If this is run from main, it raises the bug mentioned above:
```
    @unittest.skipIf(not has_ml_dtypes, "requires ml_dtypes")
    def test_conversion_ml_dtypes(self):
        x_scalar = np.array(1.5, dtype=ml_dtypes.bfloat16)
        a_scalar = mx.array(x_scalar)
>       self.assertEqual(a_scalar.dtype, mx.bfloat16)
E       AssertionError: mlx.core.complex64 != mlx.core.bfloat16
```

## Checklist

### DIFF
--- a/python/src/convert.cpp
+++ b/python/src/convert.cpp
@@ -491,7 +491,7 @@ mx::array create_array(nb::object v, std::optional<mx::Dtype> t) {
     std::optional<nb::dlpack::dtype> nb_dtype;
     // Nanobind does not recognize bfloat16 numpy array:
     // https://github.com/wjakob/nanobind/discussions/560
-    if (nb::str(v.attr("dtype")).equal(nb::str("bfloat16"))) {
+    if (v.attr("dtype").equal(nb::str("bfloat16"))) {
       nd = nb::cast<ContigArray>(v.attr("view")("uint16"));
       nb_dtype = nb::dtype<mx::bfloat16_t>();
     } else {


python/src/array.cpp
```diff
@@ -296,8 +296,40 @@ void init_array(nb::module_& m) {
...
+                  auto res = (ndim == 0)
+                      ? mx::array(*typed_ptr, mx::bfloat16)
+                      : mx::array(typed_ptr, shape, mx::bfloat16);
+                  if (t.has_value())
+                    res = mx::astype(res, *t);
+                  new (aptr) mx::array(res);
+                  return;
+                }
+              }
+            }
```
awni: This should really be done in the `create_array` function. That way other conversions from numpy bfloat16 will work.

kellen-sun: we can probably move the code to create_array, but the fix does require capturing a general nb::object from __init__ because of how the bfloat16 falls into complex, so we'll need to change the signature on create_array as well and how it's called. If so, I've added a fix, with an additional test case (the other valid use for create_array with bfloat16).

Here's the output of that case if run from main:
```
>      a_asarray = mx.asarray(x_vector)
                    ^^^^^^^^^^^^^^^^^^^^
E       ValueError: Invalid type ndarray received in array initialization.

python/tests/test_bf16.py:221: ValueError
```

python/src/convert.cpp
```diff
@@ -467,38 +467,70 @@ mx::array array_from_list(nb::tuple pl, std::optional<mx::Dtype> dtype) {
...
-                 nb::ndarray<nb::ro, nb::c_contig, nb::device::cpu>>(&v);
-             pv) {
-    return nd_array_to_mlx(*pv, t);
-  } else if (auto pv = std::get_if<mx::array>(&v); pv) {
-    return mx::astype(*pv, t.value_or((*pv).dtype()));
-  } else {
-    auto arr = to_array_with_accessor(std::get<ArrayLike>(v).obj);
-    return mx::astype(arr, t.value_or(arr.dtype()));
+  } catch (const std::bad_cast& e) {
+    throw std::invalid_argument("Cannot convert to mlx array.");
```
zcbenz: Do we need to catch and rethrow? I think the original exception should be good enough.

python/src/convert.cpp
```diff
@@ -467,38 +467,70 @@ mx::array array_from_list(nb::tuple pl, std::optional<mx::Dtype> dtype) {
...
+mx::array create_array(nb::object v, std::optional<mx::Dtype> t) {
+  if (nb::hasattr(v, "dtype")) {
+    nb::object dtype_obj = v.attr("dtype");
+    if (nb::str(dtype_obj).equal(nb::str("bfloat16"))) {
+      nb::object module_obj = v.attr("__class__").attr("__module__");
+      auto type_mod = nb::str(module_obj);
+      if (type_mod.equal(nb::str("numpy")) ||
+          type_mod.equal(nb::str("ml_dtypes"))) {
+        auto np = nb::module_::import_("numpy");
+        auto contig_obj = np.attr("ascontiguousarray")(v);
```
zcbenz: Rather than calling `ascontiguousarray()`, would `view()` be a better way as described in https://github.com/wjakob/nanobind/discussions/560#discussioncomment-15942880?

python/src/convert.cpp
```diff
@@ -467,38 +467,70 @@ mx::array array_from_list(nb::tuple pl, std::optional<mx::Dtype> dtype) {
...
+        mx::Shape shape;
+        nb::tuple shape_tuple = nb::cast<nb::tuple>(v.attr("shape"));
```
zcbenz: We have a converter for converting to `Shape` directly in "python/src/small_vector.h".

python/src/convert.cpp
```diff
@@ -467,38 +467,70 @@ mx::array array_from_list(nb::tuple pl, std::optional<mx::Dtype> dtype) {
...
+        auto res = (ndim == 0) ? mx::array(*typed_ptr, mx::bfloat16)
+                               : mx::array(typed_ptr, shape, mx::bfloat16);
+        if (t.has_value())
+          res = mx::astype(res, *t);
+        return res;
+      }
+    }
+  }
+  try {
+    auto v_cast = nb::cast<ArrayInitType>(v);
```
zcbenz: I think the whole `ArrayInitType` type should be gone now, the code would be cleaner using `nb::isinstance<...>` than the indirect `std::get_if`.

python/tests/test_bf16.py
```diff
@@ -15,6 +15,13 @@
 except ImportError as e:
     has_torch = False
 
+try:
+    import ml_dtypes
```
zcbenz, line 19: Can you add ml_dtypes to dev dependencies?

https://github.com/ml-explore/mlx/blob/e353be8235d9ad4a192fbd26b9c2f042dc687268/setup.py#L230-L238

kellen-sun:
thanks for the feedback!

setup.py
```diff
@@ -234,6 +234,7 @@ def get_tag(self) -> tuple[str, str, str]:
             "psutil",
             "torch>=2.9",
             "typing_extensions",
+            "ml_dtypes",
```
zcbenz: Can you sort the list by alphabet?

python/src/convert.cpp
```diff
@@ -467,38 +467,64 @@ mx::array array_from_list(nb::tuple pl, std::optional<mx::Dtype> dtype) {
   return array_from_list_impl(pl, dtype);
 }
 
-mx::array create_array(ArrayInitType v, std::optional<mx::Dtype> t) {
-  if (auto pv = std::get_if<nb::bool_>(&v); pv) {
-    return mx::array(nb::cast<bool>(*pv), t.value_or(mx::bool_));
-  } else if (auto pv = std::get_if<nb::int_>(&v); pv) {
-    auto val = nb::cast<int64_t>(*pv);
+mx::array create_array(nb::object v, std::optional<mx::Dtype> t) {
+  if (nb::hasattr(v, "dtype")) {
```
zcbenz: I think we can avoid the custom conversion code by redirecting to `nd_array_to_mlx`?

```c++
if (nb::isinstance<mx::array>(v)) {
  ...
} else if (nb::ndarray_check(v)) {
  ...
  return nd_array_to_mlx(nd, t);
} else if (...) {
  ...
}
```

kellen-sun: nanobind raises a bad_cast on nb::cast<ContigArray>, if we don't specify a type. I think we still need some sort of extra if clause for bfloat16 potentially with the int16 conversion trick.

zcbenz:
I pushed a change to reuse `nd_array_to_mlx` for conversion, can you check if everything still works correctly? (I will fix the windows CI later)

kellen-sun:
Thanks for the help, this builds and runs fine on my machine (the only tests I had were the ones in `test_bf16.py`). 

Also, I really enjoyed working on the binding and types for this PR. I'm looking for another issue to pick up, but the public ones seem pretty sparse right now. I was wondering if the mlx team had any feature buildouts targeted (maybe internally) that are open to community contribution? Perhaps I could work on something there?

zcbenz:
Thanks for your interests! There are some python related issues that I would be happy to review contributions:
https://github.com/ml-explore/mlx/issues?q=is%3Aissue%20state%3Aopen%20label%3Apython

Issue #3148, opened 2026-02-19
Headless/no-GPU environments crash on MLX import (NSRangeException) instead of failing gracefully


jrp2014:
### Summary

In environments where no Metal device is visible (headless/sandboxed/virtualized/macOS automation sessions),
MLX initialization aborts the process with an uncaught Objective-C exception instead of returning
a recoverable Python error.

This makes downstream tooling fail hard, including quality/lint/test pipelines that do not need GPU execution.

### Reproduction

1. Run in a session with no visible Metal device.
2. Execute:

```bash
python -c "import mlx.core as mx; print(mx.default_device())"
```

(Equivalent failure also occurs during `import mlx` via dependency probes.)

### Actual behavior

Process exits with signal/abort (`-6`) and an uncaught exception similar to:

```text
NSRangeException: -[__NSArray0 objectAtIndex:]: index 0 beyond bounds for empty array
```

### Expected behavior

- No hard abort.
- Either:
  1. raise a clear Python exception (e.g., `RuntimeError: No Metal device available`), or
  2. allow a documented CPU/no-op mode for import-time checks.
- Error path should be machine-detectable so CI tools can handle it gracefully.

### Why this matters

Hard aborts break non-inference workflows (lint/type/test/packaging checks) when MLX is installed
but GPU is unavailable. A recoverable error would allow callers to skip MLX-dependent runtime tests
without crashing the whole process.

### Suggested fix

- Guard the zero-device path before indexing into device arrays.  `load_device()` in `mlx/backend/metal/device.cpp` is the primary fix site (empty-device guard + graceful error path).  
- Convert this failure path to a typed Python exception instead of process termination.
- Optionally provide an env flag to skip Metal probing at import time in CI/headless contexts.

The main code path is:

- [`mlx/backend/metal/device.cpp` `load_device()`](https://github.com/ml-explore/mlx/blob/main/mlx/backend/metal/device.cpp#L1623-L1629)  
  It currently does `devices->object(0)` without checking if `CopyAllDevices()` returned an empty list, which matches the `NSRangeException` we're seeing.

- [`mlx/backend/metal/device.cpp` `Device::Device()`](https://github.com/ml-explore/mlx/blob/main/mlx/backend/metal/device.cpp#L1911-L1917)  
  This calls `load_device()` during backend init.

- [`mlx/device.cpp`](https://github.com/ml-explore/mlx/blob/main/mlx/device.cpp)  
[...]

PR #3150, opened 2026-02-20
activations: lazily compile activation functions to defer GPU init

JakeHillion:
Importing mlx.nn or mlx.optimizers triggers Metal GPU initialization at import time because activations.py applies @partial(mx.compile, shapeless=True) to 25 functions at module level. This causes SIGABRT in environments without Metal GPU access, such as PyInstaller packaging inside a Nix sandbox or macOS VMs.

Replaced the eager mx.compile decorators with a _lazy_compile wrapper that defers compilation to first function call using double-checked locking for free-threaded Python safety.

This allows importing all mlx Python modules without requiring a Metal GPU, while preserving identical compilation behavior at runtime.

Test plan:
- Ran an EXO instance, it worked as intended with no new errors/warnings.

## Checklist

### DIFF
--- a/python/mlx/nn/layers/activations.py
+++ b/python/mlx/nn/layers/activations.py
@@ -1,11 +1,26 @@
 # Copyright © 2023 Apple Inc.
 
 import math
-from functools import partial
+import threading
+from functools import wraps
 from typing import Any
 
 import mlx.core as mx
 from mlx.nn.layers.base import Module
+
+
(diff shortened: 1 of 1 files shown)


awni:
Compilation is already lazy. Could you please open an issue with the errors you are seeing and steps to reproduce it so we can check to see if it needs a fix?

JakeHillion:
Turns out there already is one, it's #3148

PR #3151, opened 2026-02-20
[CUDA] Implement MaskedScatter 

Lyxot:
## Proposed changes

This PR adds CUDA support for `MaskedScatter`.

## Changed files
- `mlx/backend/cuda/indexing.cpp`: implemented CUDA `MaskedScatter::eval_gpu` using the CUDA JIT module path.
- `mlx/backend/cuda/device/scatter.cuh`: added the JIT device kernel `masked_scatter_assign<...>` used by CUDA masked scatter.
- `mlx/backend/cuda/scan.cu`: refactored scan execution into reusable `scan_gpu_inplace(...)` and updated `Scan::eval_gpu` to delegate to it.
- `mlx/backend/cuda/scan.h`: added declaration for `scan_gpu_inplace(...)`.
- `python/tests/cuda_skip.py`, `tests/ops_tests.cpp`, `tests/autograd_tests.cpp`: removed CUDA skip entries for masked-scatter-related tests.

## Validation
- Python:
  - `python -m pytest python/tests/test_ops.py -k masked_scatter -q` passed.
  - `python -m pytest python/tests/test_vmap.py -k vmap_masked_scatter -q` passed.
  - `python -m pytest python/tests/test_array.py -k setitem_with_boolean_mask -q` passed.
- C++:
  - `build/tests/tests -tc="test masked_scatter,test masked_scatter autograd"` passed.

## Checklist


mlx/backend/cuda/indexing.cpp
```diff
@@ -431,4 +432,101 @@ void ScatterAxis::eval_gpu(const std::vector<array>& inputs, array& out) {
...
+  encoder.set_output_array(out);
+
+  std::string kernel_name = fmt::format(
+      "mlx::core::cu::masked_scatter_assign<{}, {}, {}>",
+      dtype_to_cuda_type(out.dtype()),
+      src.flags().row_contiguous ? "true" : "false",
+      large ? "int64_t" : "int32_t");
+  auto kernel = mod.get_kernel(kernel_name);
+  auto [num_blocks, block_dims] = get_launch_args(mask_flat, large);
+  encoder.add_kernel_node(kernel, num_blocks, block_dims, 0, args.args());
```
zcbenz: Can you rebase on main and use `add_kernel_node_raw` instead?

nastya236 (changes requested):
As I said looks great, thanks for your contribution.
Could you please provide bandwidth numbers for masked scatter kernel for a range of shapes?

Lyxot:
@nastya236 bench result from `benchmarks/python/masked_scatter.py`
NVIDIA_GeForce_RTX_4070_SUPER_masked_scatter_float32_page-0001

Lyxot:
@nastya236 
I followed up on the large-shape regression and implemented an optimization pass for `masked_scatter`.

What changed:
- Added a contiguous fast path with a fused masked-scatter kernel.
- Increased per-thread work and reduced memory traffic in the hot path.
- Reworked prefix handling for that path (tile count + prefix offsets) to avoid the previous per-element offset overhead.

On my setup, MLX is now faster than Torch in all benchmark cases, and the prior large-shape degradation trend is resolved.

NVIDIA_GeForce_RTX_4070_SUPER_masked_scatter_float32

Could you please take another look?

nastya236:
Thanks for the update! I will look as soon as possible.

zcbenz:
I think the improvement on performance is impressive but in the meanwhile the new kernel code is complicated and really hard to review, and we also want to avoid using the CUB device APIs as the overhead of graph capture easily eliminates the performance gain.

I would say the initial PR was already good enough, the code was clean and achieved reasonable performance. For further optimizations we should judge on real world needs otherwise we would have a large piece of code that are hard to maintain and never used.

@nastya236 What do you think if we just merge the initial version?

nastya236:
I agree you @zcbenz, I think initial masked scatter is slow for larger shapes because of the scan. Just out of curiosity I checked pytorch masked scatter and it is identical to what was proposed by @Lyxot initially. I think we can merge masked scatter -- first version, and if needed scan improvement should be a separate PR. 

@Lyxot thanks for exploring faster approach.

Lyxot:
@nastya236 @zcbenz Should we preserve e70c1fc9fa124a09e2139b836838a1011e90120e and f1c2a0b8332c464c264b83ac4290a806d416c797? These 2 small commits provide a ~1.1x-1.4x performance improvement over the initial version, while keeping the implementation relatively simple

mlx/backend/cuda/device/scatter.cuh
```diff
@@ -65,4 +65,97 @@ __global__ void scatter(
   Op{}(out + out_idx, upd[upd_loc]);
 }
 
+template <typename T, bool SrcContiguous, bool DstContiguous, typename IdxT>
+__global__ void masked_scatter_fused(
```
zcbenz: To match the naming style of other kernels we shouldn't add a `_fused` suffix.
```suggestion
__global__ void masked_scatter(
```

mlx/backend/cuda/indexing.cpp
```diff
@@ -435,4 +436,125 @@ void ScatterAxis::eval_gpu(const std::vector<array>& inputs, array& out) {
...
+        }
+      }
+    }
+    for (int use_large = 0; use_large <= 1; ++use_large) {
+      kernel_names.push_back(
+          fmt::format(
+              "mlx::core::cu::masked_scatter_fused_vec_contiguous<{}, {}, {}>",
+              dtype_to_cuda_type(out.dtype()),
+              use_large ? "int64_t" : "int32_t",
+              16));
```
zcbenz: Should use a variable instead of raw number everywhere.

mlx/backend/cuda/device/scatter.cuh
```diff
@@ -65,4 +65,97 @@ __global__ void scatter(
...
+__global__ void masked_scatter_fused_vec_contiguous(
+    const T* dst,
+    const bool* mask,
+    const int32_t* scatter_offsets,
+    const T* src,
+    T* out,
+    IdxT size,
+    IdxT src_batch_size,
+    IdxT mask_batch_size,
+    const __grid_constant__ Shape,
```
zcbenz: I think you can remove these parameters and then skip them in `eval_gpu`:

```c++
if (!vectorized) {
  args.append(...);
  args.append(...);
}
```

mlx/backend/cuda/scan.h
```diff
@@ -0,0 +1,20 @@
...
+
+#pragma once
+
+#include "mlx/array.h"
+#include "mlx/primitives.h"
+#include "mlx/stream.h"
+
+namespace mlx::core {
+
+void scan_gpu_inplace(
```
zcbenz: We should move it to a `mlx/backend/gpu/scan.h` header and share with the metal backend.

zcbenz (approved):
The new changes look good to me, thanks for updating the PR!

PR #3155, opened 2026-02-22
feat: adding the bartlett function

Vlor999:
## Description

This PR adds the Bartlett (triangular) window function (`mlx.core.bartlett`), continuing the effort to expand MLX's signal processing capabilities and feature parity with NumPy.

## Implementation Details

- **C++**: Implemented in `mlx/ops.cpp`.
    - **Optimization (No Branching)**: Instead of using conditional masking (`where` or `less_equal`) to build the triangular shape, this implementation leverages the absolute value identity: $w(n) = 1 - \left| \frac{2n}{M-1} - 1 \right|$. This avoids branching in the compute graph, making it significantly more efficient on the GPU.
    - **Constant Folding**: The scalar factor `2.0 / (M-1)` is computed on the CPU prior to graph construction to minimize graph size.
    - **Edge Cases**: Explicitly handles `M < 1` (empty array) and `M = 1` (returns `[1.0]`), matching `numpy.bartlett` behavior.
- **Python Bindings**: Exposed via `nanobind` in `python/src/ops.cpp` with `nb::sig` type hinting and LaTeX documentation.

## Test Plan

Verified locally against `numpy.bartlett` for numerical accuracy and edge cases.

**Verification script:**
```python
import mlx.core as mx
import numpy as np

# 1. Standard Case
M = 50
mx_bartlett = mx.bartlett(M)
np_bartlett = np.bartlett(M)

assert np.allclose(mx_bartlett, np_bartlett, atol=1e-6), "Mismatch in values"

# 2. Edge Case M=1
assert mx.bartlett(1).item() == 1.0
assert np.bartlett(1).item() == 1.0

# 3. Edge Case M=0
assert mx.bartlett(0).size == 0

print("✅ All verifications passed against numpy.bartlett")
```

## Unit Tests:
Added test_bartlett to `python/tests/test_ops.py`.

## Checklist

### DIFF
--- a/mlx/ops.cpp
+++ b/mlx/ops.cpp
@@ -2312,6 +2312,21 @@ array argmax(
   return out;
 }
 
+array bartlett(int M, StreamOrDevice s /* = {} */) {
+  if (M < 1) {
+    return array({});
+  }
+  if (M == 1) {
+    return ones({1}, float32, s);
+  }
+
+  auto n = arange(0, M, float32, s);
+  float factor_val = 2.0f / (M - 1);
+  auto factor = array(factor_val, float32);
+  auto term = subtract(multiply(factor, n, s), array(1.0f, float32), s);
+  return subtract(array(1.0f, float32), abs(term, s), s);
+}
+
 array hanning(int M, StreamOrDevice s /* = {} */) {
   if (M < 1) {
     return array({});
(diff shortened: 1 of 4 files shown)


angeloskath:
Let's merge after the tests pass and conflicts are resolved.

PR #3019, opened 2026-01-19
Add support for x86 SIMD (AVX2)

dhiltgen:
## Proposed changes

Implement AVX2 SIMD support for better performance on CPU-only x86 systems.  Quantized matmul leveraging int8 maddubs for 4-bit/8-bit weights, with FP4, and FP8 support.  Fast implementations for SDPA, RoPE, Norms, softmax and reduce.  Threadpool coordination with OpenBLAS/GCD to utilize all CPU cores.  JIT support for CPU SIMD.

Unless stated otherwise, all benchmarks with  `mlx_lm.benchmark -p 2048 -g 128` (5 trials, averages reported)

### 4-bit Quantized

| Model | Prompt tok/s | Gen tok/s | Peak GB |
|-------|-------------|-----------|---------|
| Llama-3.2-1B-4bit | 375.9 | 23.0 | 1.131 |
| Qwen2.5-1.5B-4bit | 281.9 | 19.7 | 1.193 |
| Llama-3.2-3B-4bit | 134.0 | 8.1 | 2.504 |

### 8-bit Quantized

| Model | Prompt tok/s | Gen tok/s | Peak GB |
|-------|-------------|-----------|---------|
| Llama-3.2-1B-8bit | 366.8 | 17.1 | 1.747 |
| Qwen2.5-1.5B-8bit | 277.9 | 14.4 | 1.965 |

### bf16 (Unquantized)

| Model | Prompt tok/s | Gen tok/s | Peak GB |
|-------|-------------|-----------|---------|
| Llama-3.2-1B-bf16 | 360.0 | 12.4 | 2.817 |
| Qwen2.5-1.5B-bf16 | 268.6 | 10.1 | 3.393 |
| Llama-3.2-3B-bf16 | 113.6 | 4.4 | 7.017 |

### vs Upstream MLX (unoptimized)
Upstream is too slow for p2048/g128, so both sides use `p16/g4`
| Model | Benchmark | Upstream pp tok/s | Upstream gen tok/s |
|-------|-----------|------------------:|-------------------:|
| Llama-3.2-1B-4bit | p16/g4 | 0.020 | 0.022 | 
| Llama-3.2-1B-bf16 | p16/g4 | 3.031 | 1.609 |

### 4-bit Quantized

| Model | Prompt tok/s | Gen tok/s | Peak GB |
|-------|-------------|-----------|---------|
| Llama-3.2-1B-4bit | 237.0 | 24.8 | 1.131 |
[...]

PR #3156, opened 2026-02-22
feat: native i0 (modified Bessel function) and kaiser window

Vlor999:
## Description

This PR introduces native hardware-accelerated support for the modified Bessel function of the first kind, order zero (`mlx.core.i0`), and leverages it to implement the Kaiser window function (`mlx.core.kaiser`).

Implementing `i0` natively allows `kaiser` to be fully compiled and evaluated on Apple Silicon GPUs without relying on slow CPU fallback loops or external libraries.

### 1. Hardware-Accelerated `i0` (Cephes Approximation)
Since $I_0(x)$ cannot be computed via simple composition, it is implemented as a core unary primitive using the industry-standard Cephes polynomial approximation (matching NumPy/SciPy):
- **CPU Backend (`mlx/backend/cpu/simd/math.h`)**: Implemented vectorized evaluation using `Simd<T, N>` and FMA instructions for high-throughput CPU processing.
- **GPU Backend (`mlx/backend/metal/kernels/i0.h`)**: Implemented a native Metal shader using `metal::fma` and `metal::precise::exp`/`sqrt` to ensure numerical stability matching the CPU.
- **Autograd / VJP (`mlx/primitives.cpp`)**: Implemented the analytic derivative. Since $\frac{d}{dx} I_0(x) = I_1(x)$, I implemented a composite $I_1(x)$ function directly in the `jvp` pass using the same two-domain Cephes approximation logic.

### 2. Kaiser Window (`mlx.core.kaiser`)
- Implemented in `mlx/ops.cpp` as a composite function using the new `i0` primitive.
- Formula: $w(n) = I_0\left(\beta \sqrt{1 - \left(\frac{2n}{M-1} - 1\right)^2}\right) / I_0(\beta)$
- **Optimization**: Uses scalar broadcasting for `1.0f` to avoid unnecessary array allocations in the `sqrt` term. Handles edge cases (`M=0`, `M=1`).

### 3. Python Bindings
- Exposed `i0` and `kaiser` in `python/src/ops.cpp` with fully typed `nb::sig` and LaTeX docstrings.

## Test Plan

Added comprehensive tests in `python/tests/test_ops.py`:
1. **NumPy Parity (`test_i0`)**: Checked `mx.i0` against `numpy.i0` across both polynomial domains (`|x| <= 3.75` and `|x| > 3.75`).
2. **Hardware Parity (`test_i0_cpu_gpu_parity`)**: Verified that forcing `mx.stream(mx.cpu)` and `mx.stream(mx.gpu)` yields identical results within `1e-5` tolerance.
3. **Kaiser Parity (`test_kaiser_general`)**: Verified `mx.kaiser(M, beta)` against `np.kaiser` including edge cases (`M=1`) and symmetry.

All tests pass successfully.

## Checklist


zcbenz:
Sorry for the long wait, it had been hard to justify the maintenance costs of adding new primitives against the very few requests of these ops. For the moment I'm going to close this issue because there are some planned improvements on SIMD in https://github.com/ml-explore/mlx/pull/3019, and the discussion of adding new special functions can happen in #3330. Still thanks for the PR!

PR #3157, opened 2026-02-23
[CUDA] columnwise quantize with tma

nastya236:
Columnwise quantization with tma (mxfp8), bfloat16:
| Size | With tma (ms)| Without tma (ms)|
|---|---:|---:|
| 4096×4096 | 68.48 | 77.21 |
| 4096×8192 | 78.74 | 102.58 |
| 8192×4096 | 80.73 | 100.61 |
| 8192×8192 | 100.67 | 145.16 |
| 4096×16384 | 97.08 | 144.45 |
| 16384×4096 | 99.50 | 137.13 |

  This PR:                                                                                                                                                                                                                                                                                     
  - Adds PTX instructions for asynchronous copy with TMA                                                                                                                                                                                                                                       
  - Adds fp_quantize_columnwise_mxfp8 kernel for columnwise MXFP8 quantization using TMA on SM100+                                                                                                                                                                                             
  - Splits fp_quantize.cu into fp_quantize.cu (dispatch) and fp_quantize.cuh (kernels) to reduce file size                                                                                                                                                                                     
  - Moves swizzle_scales constants and `get_swizzle_launch_args` into `cu::` namespace for consistency

The diff seems bigger than it is: since `fp_quantize.cu` already contains all the logic for block quantization, I decided to split it into `fp_quantize.cu` and `fp_quantize.cuh` to separate dispatch from kernel implementation. 

  TODO:
`nvfp4` requires a separate columnwise kernel due to TMA tile size constraints. In the proposed kernel each thread processing a tile of size (N, M) and store a transposed result. M is equal to `group_size` -- 32 bytes for`mcfp8`, but only 8 bytes for `nvfp4`. Since TMA requires the innermost tile dimension to be at least 128 bits (16 bytes), for `nvfp4` kernel would need to load a larger tile and iterate over multiple groups.


mlx/backend/cuda/ptx.cuh
```diff
@@ -0,0 +1,127 @@
...
+#include <cuda_runtime.h>
+
+namespace mlx::core {
+
+namespace ptx {
+
+#if (CUDART_VERSION >= 12080) && (__CUDA_ARCH__ >= 1000) && \
+    defined(__CUDA_ARCH_SPECIFIC__)
+
+__device__ __forceinline__ void mbarrier_init(uint64_t* mbar, uint32_t count) {
```
zcbenz, line 13: Should we use the `cuda::ptx` APIs like `cuda::ptx::mbarrier_init` API instead? They don't have good documentation and you would have to search https://github.com/NVIDIA/cccl to find out API names though.

mlx/backend/cuda/quantized/qqmm_utils.cu
```diff
@@ -10,11 +10,6 @@ namespace mlx::core {
 
 namespace cg = cooperative_groups;
```
zcbenz: This should be moved to `namespace cu` too.

mlx/backend/cuda/quantized/fp_quantize.cuh
```diff
@@ -0,0 +1,629 @@
...
+  using Tx2 = Vector2_t<T>;
+  uint32_t rbits = 0; // reserved bits for future use
+  auto block_size = cg::this_thread_block().dim_threads();
+  auto block_idx = cg::this_thread_block().group_index();
+  auto idx_in_block = cg::this_thread_block().thread_index();
+  auto tidx = block_idx.x * block_size.x + idx_in_block.x;
+  auto tidy = block_idx.y * block_size.y + idx_in_block.y;
+  auto grid_dim_x = cg::this_grid().dim_blocks().x * block_size.x;
+
+  size_t thread_idx = tidx + grid_dim_x * size_t(tidy);
```
zcbenz: I think we can just use `cg::this_grid().thread_rank()`?

mlx/backend/cuda/quantized/fp_quantize.cu
```diff
@@ -14,384 +13,98 @@
...
+    int SCALES_PER_STAGE>
+inline std::tuple<dim3, dim3, size_t> get_columnwise_quantize_mxfp8_launch_args(
+    size_t grid_dim_x_size, // rows
+    size_t grid_dim_y_size, // cols
+    size_t block_size_x, // ROWS_PER_BLOCK
+    size_t block_size_y, // COL_PER_BLOCK
+    int in_size_bytes, // itemsize
+    int bits) {
+  dim3 grid;
+  grid.x = (grid_dim_x_size + block_size_x - 1) / block_size_x;
```
zcbenz: Can you use `cuda::ceil_div` when possible? 

```suggestion
  grid.x = cuda::ceil_div(grid_dim_x_size, block_size_x);
```

mlx/backend/cuda/quantized/fp_quantize.cuh
```diff
@@ -0,0 +1,629 @@
...
+  constexpr size_t BUFF_ELEMS = TILE_M * TILE_K;
+  constexpr size_t in_tile_size = BUFF_ELEMS * sizeof(T);
+  constexpr size_t in_buff_size_aligned =
+      ((in_tile_size * BUFFS_NUM + TMA_SHMEM_ALIGNMENT - 1) /
+       TMA_SHMEM_ALIGNMENT) *
+      TMA_SHMEM_ALIGNMENT;
+
+  constexpr size_t out_tile_elems = BUFF_ELEMS / elem_per_byte;
+  constexpr size_t out_tile_size = out_tile_elems;
+  constexpr size_t out_buff_size_aligned =
```
zcbenz: This is not used anywhere.

mlx/backend/cuda/quantized/fp_quantize.cuh
```diff
@@ -0,0 +1,629 @@
...
+      ((out_tile_size * BUFFS_NUM + TMA_SHMEM_ALIGNMENT - 1) /
+       TMA_SHMEM_ALIGNMENT) *
+      TMA_SHMEM_ALIGNMENT;
+
+  extern __shared__ char shared_mem[];
+  uintptr_t aligned_shared =
+      (reinterpret_cast<uintptr_t>(shared_mem) + TMA_SHMEM_ALIGNMENT - 1) &
+      ~(static_cast<uintptr_t>(TMA_SHMEM_ALIGNMENT - 1));
+
+  T* in_sh = reinterpret_cast<T*>(aligned_shared);
```
zcbenz: You can make sure you get necessary alignment with dynamic allocated shared memory with this:

```cuda
  extern __shared__ uint128_t shared_mem[];
  T* in_sh = reinterpret_cast<T*>(shared_mem);
```

or:

```cuda
extern __shared__ alignas(128) char shared_mem[];
```

Also, I think `in_smem` would be an easier to understand name.

mlx/backend/cuda/quantized/fp_quantize.cu
```diff
@@ -14,384 +13,98 @@
...
+      ((in_tile_size * BUFFS_NUM + TMA_SHMEM_ALIGNMENT - 1) /
+       TMA_SHMEM_ALIGNMENT) *
+      TMA_SHMEM_ALIGNMENT;
+
+  const size_t out_tile_elems = BUFF_ELEMS / elem_per_byte;
+  const size_t out_buff_size_aligned =
+      ((out_tile_elems * BUFFS_NUM + TMA_SHMEM_ALIGNMENT - 1) /
+       TMA_SHMEM_ALIGNMENT) *
+      TMA_SHMEM_ALIGNMENT;
+  const size_t smem_size =
```
zcbenz: It appears that the size of shared memory is static? I don't think you need to use dynamic shared memory in this case, you can ensure alignment with:

```cuda
__shared__ alignas(128) T smem[SIZE];
```

mlx/backend/cuda/ptx.cuh
nastya236: Thanks, great question. My intention with raw PTX was to make sure I understand exactly what each instruction does. I felt that when wrapped in ccc up the details are hidden bit, so writing them in a raw ptx makes them explicit. I agree on docs point. That being said if you have a strong preference, I am happy to substitute it with cccl functions :)

angeloskath:
@nastya236 is there any reason not to merge this? Will do a quick pass but I think it should be good to go, isn't it?

nastya236:
The reason we have not merged it is that i was too slow to respond to Cheng's comment. I think it is mergable, thank you!

angeloskath (approved):
👍

PR #3158, opened 2026-02-23
Add Expert Parallelism for MoE inference

0xDaizz:
> **Umbrella PR** — broken into smaller, focused PRs per maintainer feedback.
> Each sub-PR is independently reviewable and testable.

## Sub-PR Checklist

- [x] **PR1A** — `all_to_all` collective primitive (#3164)
- [ ] **PR1B** — MoE dispatch/combine C++ primitives + blocking comm infra
- [ ] **PR1C** — Metal GPU runtime for MoE primitives
- [ ] **PR1D** — Python `MixtureOfExperts` layer + tests/docs
- [ ] **PR2A** — Production infra: auto backend policy, warmup, metrics
- [ ] **PR2B** — Performance: batched expert FFN + zero-copy combine
- [ ] **PR2C** — Benchmarks (EP diagnostic + EP-vs-TP comparison)

---

_Original description below for reference._

## Summary

Add Expert Parallelism (EP) for MoE inference on Apple Silicon.

EP distributes experts across devices; each device holds a subset and tokens are routed via all-to-all exchange. This enables scaling MoE models (Mixtral, DBRX, Kimi K2.5) beyond single-device memory.

### Key components

1. **`all_to_all` collective** — equal-chunk exchange between all ranks
2. **Fused MoE dispatch/combine primitives** — C++ primitives with CPU + Metal GPU backends
3. **Blocking comm infra** — `blocking_exchange_v` for variable-size packet exchange (RDMA/MPI)
4. **Auto backend policy** — CPU for decode (N≤64), Metal for prefill (N≥320)
5. **Python `MixtureOfExperts` layer** — drop-in replacement with EP support
6. **Production infra** — warmup, metrics, GPU fallback

### Performance (2-rank JACCL, E=384 D=7168 top_k=8)

| Regime | EP vs TP | Interpretation |
|--------|----------|----------------|
| Decode (N≤64) | **0.32x** | EP 3.1x faster |
| Prefill (N≥256) | 1.07x | TP 7% faster |
| Overall geomean | **0.55x** | EP 1.8x faster |

### DIFF
--- a/docs/src/python/distributed.rst
+++ b/docs/src/python/distributed.rst
@@ -17,6 +17,9 @@ made available.
     init
     all_sum
     all_gather
+    all_to_all
+    moe_dispatch_exchange
+    moe_combine_exchange
     send
     recv
     recv_like
--- a/docs/src/python/nn/layers.rst
+++ b/docs/src/python/nn/layers.rst
@@ -45,6 +45,7 @@ Layers
    MaxPool2d
    MaxPool3d
    Mish
+   MixtureOfExperts
    MultiHeadAttention
    PReLU
    QuantizedAllToShardedLinear
--- a/mlx/backend/cpu/distributed.cpp
+++ b/mlx/backend/cpu/distributed.cpp
@@ -1,11 +1,13 @@
 // Copyright © 2024 Apple Inc.
 
 #include <cassert>
+#include <cstring>
 
 #include "mlx/allocator.h"
 #include "mlx/backend/cpu/copy.h"
 #include "mlx/backend/cpu/encoder.h"
 #include "mlx/distributed/primitives.h"
+#include "mlx/types/half_types.h"
 
 namespace mlx::core::distributed {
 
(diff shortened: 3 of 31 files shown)


0xDaizz:
**Follow-up (planned)**: After this PR is merged, a second PR will add production infrastructure:
- Auto backend policy (CPU/Metal selection based on workload size)  
- Batched expert FFN (1.2x decode speedup)
- Zero-copy combine Metal kernel
- RDMA warmup, metrics, and fallback hardening
- EP vs TP comparison benchmarks

The follow-up branch (`feat/ep-pr2`) is ready on the fork and will be rebased onto `main` once this PR merges.

angeloskath:
Thanks for the great work @0xDaizz !

However, in order for this to be evaluated properly and merged I think you need to break it into multiple PRs. You can start for instance with the `all_to_all` implementation. My advice is either close this or turn it into a draft PR that gets constantly rebased on main and becomes smaller and smaller as the constituent PRs get merged.

0xDaizz:
@angeloskath Thanks for the guidance — makes sense.
  I’ll convert this PR to draft and split it into smaller PRs, starting with all_to_all.
  I’ll link each follow-up PR here as they are opened.

hemanth-e-4469:
@0xDaizz I am curious if you had any benchmarks done (end -end on K2.5/2.6) on Mac Studio cluster if so can you share them here

zcbenz:
Sorry for the very late response, I don't think we would be able to review it, reviewing PRs like this basically takes as much time as writing it ourselves and at the moment this feature is not on our roadmap.

PR #3159, opened 2026-02-23
[Metal] Fix event leak

awni:
There was an edge case where events can leak

Should close https://github.com/ml-explore/mlx-lm/issues/887


angeloskath (approved):
🙏

ivanfioravanti:
Thanks 🙏 gonna try it right now!

Issue #2536, opened 2025-08-25
[BUG] RuntimeError: QuantizedMatmul has no CUDA implementation


ehartford:
**Describe the bug**
RuntimeError: QuantizedMatmul has no CUDA implementation

**To Reproduce**

```python
import mlx.core as mx
import mlx_lm
model, tokenizer = mlx_lm.load("mlx-community/Qwen3-30B-A3B-Thinking-2507-4bit")
logits = model(mx.array([[1]]))
mx.eval(logits)
```

**Expected behavior**
The model should load successfully and return a tensor of logits without errors. The mx.eval() call should complete execution and the logits variable should contain the model's output predictions for the input token.

**Desktop (please complete the following information):**
 - OS Version:  Ubuntu 24.04.3 LTS
 - Version: 
   - mlx==0.28.0
   - mlx-cuda==0.28.0
   - mlx-lm==0.26.3

**Additional context**
I was trying to run mlx-bench.py

[mlx-bench.py](https://github.com/user-attachments/files/21959703/mlx-bench.py)

with command:

`$ python ./mlx-bench.py -m mlx-community/Qwen3-30B-A3B-Thinking-2507-4bit`

Full stack trace:

```
Test configuration:
  Prompt tokens: 512 (random)
  Generation tokens: 128 (random)
  Repetitions: 5
  Warmup: True

Running warmup...
  Warmup prompt processing...
Error: QuantizedMatmul has no CUDA implementation.
Traceback (most recent call last):
  File "/home/eric/benchmarks/./mlx-bench.py", line 232, in main
    results = run_benchmark(
              ^^^^^^^^^^^^^^
  File "/home/eric/benchmarks/./mlx-bench.py", line 117, in run_benchmark
    _ = test_prompt(model, tokenizer, min(32, n_prompt))
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/eric/benchmarks/./mlx-bench.py", line 47, in test_prompt
    mx.eval(logits)  # Force evaluation
    ^^^^^^^^^^^^^^^
RuntimeError: QuantizedMatmul has no CUDA implementation.
```

awni:
Quantized matmuls are WIP for the CUDA back-end. It's probably the top priority, hopefully they will be in an upcoming release 🤞 

[...]

PR #3160, opened 2026-02-23
[CUDA] FPxINT quantized matmul for Hopper

zcbenz:

This PR implements QMM for Hopper only, using CUTLASS.

Performance of FP16xINT4 with group size 64 on H100:

| M | N | K | CUTLASS TFlops | CUBLAS TFlops | Speedup |
| 1 | 16384 | 16384 | 6.0 | 2.8 | 2.16 |
| 2 | 16384 | 16384 | 12.0 | 5.6 | 2.15 |
| 4 | 16384 | 16384 | 24.0 | 11.4 | 2.10 |
| 8 | 16384 | 16384 | 48.0 | 21.8 | 2.20 |
| 16 | 16384 | 16384 | 96.1 | 47.8 | 2.01 |
| 32 | 16384 | 16384 | 100.1 | 86.9 | 1.15 |
| 64 | 16384 | 16384 | 102.7 | 178.2 | 0.58 |

[Independent C++ code for profiling the kernel](https://github.com/zcbenz/qmm/blob/4feffe21d757a8045dcff6d0f24ca1e8a92b63b2/qmm_sm90.cu)

For small batch size (<=16) the performance data is similar to (though slower than) [Machete](https://developers.redhat.com/articles/2024/10/14/introducing-machete-mixed-input-gemm-kernel) so I think it is doing good. When M becomes larger than 32 it can no longer surpass cuBLAS, and it can be solved by increasing tile size, which I will fix in following PRs.

Another problem is this kernel behaves badly for small N/K (0.5x for n4096k4096), I'm not sure if it is something TMA kernels are not good at and we might have to fallback to Ampere optimized kernels.

---

There are also a few things need further optimizations:

1. The CUTLASS kernel requires scales/zeros to be N-major but MLX's quant format is K-major so there is a transpose involved for each call. CUDA graph can partly hide this overhead but I haven't profiled how much it hurts performance.
2. By shuffling the quantized weights we can use wider vectorized reads and faster int-to-float conversions, which would bring 17%~20% improvement. The [55_hopper_int4_bf16_gemm](https://github.com/NVIDIA/cutlass/blob/main/examples/55_hopper_mixed_dtype_gemm/55_hopper_int4_bf16_gemm.cu) example implements a `--shuffle` switch that can be used to measure the performance difference.

### DIFF
--- a/mlx/backend/cuda/CMakeLists.txt
+++ b/mlx/backend/cuda/CMakeLists.txt
@@ -56,6 +56,7 @@ target_sources(
           ${CMAKE_CURRENT_SOURCE_DIR}/utils.cpp
           ${CMAKE_CURRENT_SOURCE_DIR}/quantized/affine_quantize.cu
           ${CMAKE_CURRENT_SOURCE_DIR}/quantized/fp_quantize.cu
+          ${CMAKE_CURRENT_SOURCE_DIR}/quantized/qmm_sm90.cu
           ${CMAKE_CURRENT_SOURCE_DIR}/quantized/qmv.cu
           ${CMAKE_CURRENT_SOURCE_DIR}/quantized/quantized.cpp
           ${CMAKE_CURRENT_SOURCE_DIR}/quantized/qqmm.cpp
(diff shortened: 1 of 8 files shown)


zcbenz:
Skimming the Machete code which achieve 3x for batch size 16/32, I think the performance difference mainly comes from a better designed weights layout. (Machete is modified from the cutlass kernel this PR uses.)

PR #3166, opened 2026-02-24
Fix compile_fuse broadcast split aliasing bug

### LINKED ISSUE #3165: compile_fuse crash: unordered_map::at on graphs with >~100 ops
## Bug

`mx.compile` crashes with `unordered_map::at: key not found` when the compiled function's computation graph exceeds ~100 operations. The crash occurs in `compile_replace` during the second call (cached path).

## Reproduction (Python)

```python
import mlx.core as mx
import math

def test_compile_n(n):
    obs = [mx.array(i * 2.0) for i in range(n)]
    xs = [mx.array(i * 1.0) for i in range(n)]
    mu0, sig_prior, sig_obs = mx.array(0.0), mx.array(10.0), mx.array(1.0)
    half = mx.array(0.5)
    log2pi_half = mx.array(0.5 * math.log(2 * math.pi))
    idx0, idx1 = mx.array(0, mx.int32), mx.array(1, mx.int32)

    def gaussian_log_prob(v, mu, sigma):
        z = mx.divide(mx.subtract(v, mu), sigma)
        return mx.negative(
            mx.add(log2pi_half, mx.add(mx.log(sigma), mx.multiply(half, mx.square(z))))
        )

    @mx.compile
    def score_fn(params):
        t = mx.transpose(params)
        slope = mx.take(t, idx0, axis=0)
        intercept = mx.take(t, idx1, axis=0)
        score = mx.add(
            gaussian_log_prob(slope, mu0, sig_prior),
            gaussian_log_prob(intercept, mu0, sig_prior)
        )
        for i in range(n):
            mean = mx.add(mx.multiply(slope, xs[i]), intercept)
            score = mx.add(score, gaussian_log_prob(obs[i], mean, sig_obs))
        return score

    params = mx.random.normal((4, 2))
    r = score_fn(params)
    mx.eval(r)
    # Second call triggers the crash (cached path)
    r2 = score_fn(mx.random.normal((4, 2)))
    mx.eval(r2)

# N=8: OK, N=9+: crash
for n in [8, 9, 10, 20, 50]:
    try:
        test_compile_n(n)
        print(f"N={n}: OK")
    except Exception as e:
        print(f"N={n}: FAIL - {e}")
```

**Output (MLX 0.30.6, also fails on 0.29.4):**
```
N=8: OK
N=9: FAIL - [unordered_map::at: key not found]
N=10: FAIL - [unordered_map::at: key not found]
N=20: FAIL - [unordered_map::at: key not found]
N=50: FAIL - [unordered_map::at: key not found]
```

## Root cause

The bug is in `compile_fuse` in `mlx/compile.cpp`. When a broadcast array has mixed parents (some inside the fused group, some outside), `split_one` creates a **copy** of the broadcast array. This copy:

1. Is not present in the original tape
2. May end up NOT being fused into the compiled group
3. Becomes an "input" to the `Compiled` primitive in `recurse_tape`
4. In `compile_replace`, the lookup in `trace_to_real` fails because the copy was never in the tape

The relevant code path (around line 854 of `compile.cpp`):
```cpp
if (!all_parents_in) {
  if (a.has_primitive() && is_broadcast(a.primitive())) {
    array b = split_one(a, parents_map, cache);  // creates new array not in tape
    recurse(b, depth, s, shape);
  } else {
    input_set.insert(a.id());
  }
  return;
}
```

This was introduced in commit `2c11d10f` ("Split broadcast so it is always fused in compile").

## Workaround

Removing the broadcast split optimization fixes the crash — broadcasts with mixed parents become regular inputs instead of being split:

```cpp
if (!all_parents_in) {
  input_set.insert(a.id());
  return;
}
```

This trades a minor fusion optimization for correctness. A more targeted fix could ensure `split_one` copies are properly tracked in the tape.

## Environment

- macOS 15.5, Apple M4
- MLX 0.30.6 (also reproduces on 0.29.4)
- Python 3.13

robert-johansson:
## Summary

Fixes #3165 — `mx.compile` crashes with `unordered_map::at: key not found` on computation graphs with >~100 operations.

- **Root cause**: `split_one` modifies a parent's input vector, silently replacing the `const array& a` reference in the `recurse` lambda (which is aliased through that vector). When the split copy fails to fuse, the original broadcast is orphaned — never added to `input_set`.
- **Fix**: Save `array orig(a)` before `split_one`. If the copy isn't fused, undo with `merge_one` and use the original as an input. Preserves the broadcast fusion optimization from #2318.
- **Test**: Added `test_compile_large_graph_with_broadcasts` — builds a ~150-op graph (Gaussian scoring with 15 observations), verifies both cached execution and correctness vs uncompiled.

## Test plan

- [x] New test `test_compile_large_graph_with_broadcasts` passes
- [x] Existing `test_shared_broadcast` still passes (optimization preserved)
- [x] Verified N=3 through N=100 all pass (previously failed at N=9+)

### DIFF
--- a/mlx/compile.cpp
+++ b/mlx/compile.cpp
@@ -856,18 +856,10 @@ void compile_fuse(
       // are not fusable except for broadcast which we can split to avoid
       // stopping fusion
       if (!all_parents_in) {
-        if (a.has_primitive() && is_broadcast(a.primitive())) {
-          // Save the original before split_one, because `a` is a reference
-          // into a parent's input vector and split_one will replace it
-          array orig(a);
-          array b = split_one(orig, parents_map, cache);
+        if (a.has_primitive() && is_broadcast(a.primitive()) &&
+            input_set.size() < max_compile_arrays) {
+          array b = split_one(a, parents_map, cache);
           recurse(b, depth, s, shape);
-          if (cache.find(b.id()) == cache.end()) {
-            // Split copy wasn't fused, undo and treat original as input
-            merge_one(orig, b, parents_map);
-            input_set.erase(b.id());
-            input_set.insert(orig.id());
-          }
         } else {
           // Possible input
           input_set.insert(a.id());
--- a/mlx/compile.cpp
+++ b/mlx/compile.cpp
@@ -857,8 +857,17 @@ void compile_fuse(
       // stopping fusion
       if (!all_parents_in) {
         if (a.has_primitive() && is_broadcast(a.primitive())) {
-          array b = split_one(a, parents_map, cache);
+          // Save the original before split_one, because `a` is a reference
+          // into a parent's input vector and split_one will replace it
+          array orig(a);
+          array b = split_one(orig, parents_map, cache);
           recurse(b, depth, s, shape);
+          if (cache.find(b.id()) == cache.end()) {
+            // Split copy wasn't fused, undo and treat original as input
+            merge_one(orig, b, parents_map);
+            input_set.erase(b.id());
+            input_set.insert(orig.id());
+          }
         } else {
           // Possible input
           input_set.insert(a.id());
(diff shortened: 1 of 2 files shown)


angeloskath (approved):
Thanks for finding the bug. I simplified the test and solution and will merge after the tests pass.

Issue #3126, opened 2026-02-13
[BUG] Enable MLX to cleanly exit when used in a sub-thread


awni:
```c++
#include <iostream>
#include <thread>

#include "mlx/mlx.h"

namespace mx = mlx::core;

int main() {
  std::thread t([] {
    auto a = mx::random::normal({2048, 2048});
    std::cout << "START" << std::endl;
    for (int i = 0; i < 1000; ++i) {
      a = mx::matmul(a, a);
      // Eval periodically to avoid building a huge graph
      if (i % 10 == 0) {
        mx::eval(a);
        std::cout << "Step " << i << std::endl;
      }
    }
    mx::eval(a);
    std::cout << "Done: " << a.shape(0) << "x" << a.shape(1) << std::endl;
  });

  sleep(1);
  t.detach();
  std::cout << "Main thread exiting." << std::endl;
  return 0;
}
```

That will crash / exit uncleanly from time to time.

This is important for applications so that when they are quit they don't crash.

zcbenz:
Closed by https://github.com/ml-explore/mlx/pull/3167.

PR #3167, opened 2026-02-24
Fix crashes in multi-threaded process teardown

louen:
This PR is a proposed fix for issue #3126 , where exceptions or segfaults can be triggered during the teardown of a multi-thread process using MLX.

A typical example would be an interactive application where UI runs on the main thread and MLX work is being done in the background on a separate thread. 
If the app exits while the GPU is busy with a compute kernel, that kernel can return, and execute its completion call back  _after_ static objects have been destroyed by the runtime. Unfortunately in many case these callbacks try to access those singleton static objects, either the `Device` or the `Scheduler`. 

## Proposed changes

This PR implements a mitigation strategy in two parts
- intentionally leaking singleton objects, by replacing the static instance with a pointer, which is never `delete`d. This ensure that the objects remain alive even after the main thread's static object destruction has completed.
- ~~adding a global atomic toggle flag to `metal` namespace  which allows a user to disable the metal backend. This prevents additional work from being enqueued on the GPU while the application is shutting down~~.

## Tests

A new test executable has been added with the minimal repro case from #3126.  

Before proposed changes: 
```
> ./build/tests/test_teardown
START
...
Step 200
Step 210
Main thread exiting.
Step 220
libc++abi: terminating due to uncaught exception of type std::out_of_range: vector
zsh: abort      ./tests/test_teardown
```

After:
```
 ./tests/test_teardown
START
...
Step 200
Step 210
Main thread exiting.
```
Showing the effect of the fix
- No more crash at exit
- Step 220 is not being run after the exit signal

## Checklist

### DIFF
--- a/tests/test_teardown.cpp
+++ b/tests/test_teardown.cpp
@@ -14,7 +14,7 @@ namespace mx = mlx::core;
 
 int main() {
   using namespace std::chrono_literals;
-  
+
   std::thread t([] {
     auto a = mx::random::normal({2048, 2048});
     std::cout << "START" << std::endl;


mlx/backend/metal/metal.cpp
```diff
@@ -7,8 +8,15 @@
...
+// Global flag to prevent new GPU operations, for example at app termination.
+static std::atomic<bool> enabled{true};
+
 bool is_available() {
-  return true;
+  return enabled.load(std::memory_order_acquire);
+}
+
+void set_enabled(bool enabled) {
+  metal::enabled.store(enabled, std::memory_order_release);
```
angeloskath: That is probably not the right way to do that. I understand that it changes the default stream from gpu to cpu when metal is disabled but the stream is still there. People could still submit to it and if that causes a failure it can still cause a failure.

Eg what should the following code do?

```cpp
auto s = mx::default_stream(mx::default_device());
mx::metal::set_enabled(false);
auto x = mx::add(mx::array(1), mx::array(2), s);
mx::eval(x)
```

Just so that I don't sound pedantic. It is common to grab a stream specific to a task. For instance in mlx-lm we have `generation_stream` for well generation. That means if I was generating in a thread and the program exited and set the above, it would still submit to the GPU even though it was "disabled".

angeloskath:
@louen how do you feel about removing the flag and merging this?

louen:
@angeloskath Yes. I will do this asap !

angeloskath:
Yeah sorry was waiting for CI to get fixed and then it got forgotten

PR #3174, opened 2026-02-26
JACCL refactor and small update

angeloskath:
This PR refactors the communication implementations outside the groups so we can use the more efficient ring reduce in the mesh group for large sizes.

Comparison of all reduce performance on 4 M3 Ultras.
jaccl-mesh-update

### DIFF
--- a/mlx/distributed/jaccl/mesh.cpp
+++ b/mlx/distributed/jaccl/mesh.cpp
@@ -205,8 +205,9 @@ void MeshGroup::all_reduce(
   encoder.set_input_array(input);
   encoder.set_output_array(output);
   encoder.dispatch([in_ptr, out_ptr, size, this, reduce_op]() {
-    if ((std::is_same_v<T, bfloat16_t> && size > 65536) ||
-        size >= 8 * 1024 * 1024 / sizeof(T)) {
+    if (size_ > 2 &&
+        ((std::is_same_v<T, bfloat16_t> && size > 65536) ||
+         size >= 8 * 1024 * 1024 / sizeof(T))) {
       ring_.all_reduce<2>(in_ptr, out_ptr, size, 1, reduce_op);
     } else {
       mesh_.all_reduce(in_ptr, out_ptr, size, reduce_op);


nastya236:
Looks awesome!! Thanks for adding me as a reviewer, I will look at it today!

mlx/distributed/jaccl/mesh.cpp
```diff
@@ -75,18 +84,27 @@ void MeshGroup::initialize() {
...
+  ring_recv_buffers_.clear();
 
   // Allocate the memory
   for (int k = 0; k < BUFFER_SIZES; k++) {
     for (int i = 0; i < NUM_BUFFERS; i++) {
+      // Mesh buffers
       for (int j = 0; j < size_; j++) {
         buffers_.emplace_back(FRAME_SIZE * (1 << k));
       }
+      // Ring buffers (1 for each direction)
```
nastya236, line 97: Nit and just for me to understand better: in case `size_ = 2`, mesh and ring should be identical and we would never fall into ring all_reduce. I am wondering if it makes sense to allocate ring buffers and init `RingImpl` only if `size_ > 2`? But probably, extra complexity does not justify small memory saving..

mlx/distributed/jaccl/mesh.cpp
```diff
@@ -336,114 +200,17 @@ void MeshGroup::all_reduce(
...
-          if (w + (PIPELINE - 1) * N < total) {
-            recv_from(sz, r, buff);
-            in_flight++;
-          }
-        }
-        completed_recv_begin[r] = s;
-      }
+  encoder.dispatch([in_ptr, out_ptr, size, this, reduce_op]() {
+    if (size_ > 2 &&
+        ((std::is_same_v<T, bfloat16_t> && size > 65536) ||
```
nastya236, line 209: Interesting! Just for my understanding, why for `bfloat16` the cutover is at 65,536 elements = 128KB, but for other 2-byte types (like float16), it is 8MB / 2 = 4M elements = 8MB?

mlx/distributed/jaccl/mesh.cpp
```diff
@@ -139,83 +170,7 @@ void MeshGroup::all_gather(const array& input, array& output, Stream stream) {
...
-              data + rank * n_bytes + write_offset[rank]);
-          write_offset[rank] += N;
-          if (write_offset[rank] + N * (PIPELINE - 1) < total) {
-            recv_from(sz, rank, buff);
-            in_flight++;
-          }
-        }
-      }
-    }
+    mesh_.all_gather(in_ptr, out_ptr, n_bytes);
```
nastya236, line 173: Also just for me to understand better, I am wondering if it makes sense here to have similar conditional `all_gather` as for `all_reduce` if a message is large enough? What do you think?

mlx/distributed/jaccl/mesh.cpp
angeloskath: Yeah that isn't a bad idea. The memory is 4MB so not important but perhaps there are other reasons to not register unnecessary buffers.

We might still want to use a ring even with 2 nodes because the process is different. The mesh sends all at once and reduces all. The ring sends half, reduces half and gathers half. If the reduction is very expensive then the ring will be faster.

mlx/distributed/jaccl/mesh.cpp
angeloskath: Bfloat summation is slow on the M3s (we should put in a TODO to make it faster) so the ring (that sums less) is more efficient. For 8MB is where the ring surpasses the mesh with 4 nodes.

mlx/distributed/jaccl/mesh.cpp
angeloskath: Good pointing it out. It doesn't because there is no bandwidth benefit for all gather as there is for all reduce. All reduce benefits because the summations are shared among nodes so we don't ever send all the data.

So tl;dr all gather will be quite slower via the ring vs the mesh.

PR #3176, opened 2026-02-26
Enable passing in a GPU architecture string via env var

angeloskath:
We tend to use this string for routing to kernels and so on and it can be helpful to test different pathways quickly. See for instance with or without NAX and so on.

### DIFF
--- a/mlx/backend/metal/device_info.cpp
+++ b/mlx/backend/metal/device_info.cpp
@@ -21,9 +21,10 @@ device_info(int device_index) {
   auto init_device_info = []()
       -> std::unordered_map<std::string, std::variant<std::string, size_t>> {
     auto pool = metal::new_scoped_memory_pool();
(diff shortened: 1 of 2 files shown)


mlx/utils.h
```diff
@@ -169,6 +170,11 @@ inline int nccl_timeout(int default_value) {
   return nccl_timeout;
 }
 
+inline const std::string& metal_gpu_arch() {
+  static std::string gpu_arch_ = get_var("MLX_METAL_GPU_ARCH", "");
+  return gpu_arch_;
```
andresy, line 175: Wondering if some sanity checks should be done here? E.g., we use `.back()` on the arch name in a number of places...

angeloskath: Oops overzealous merging.

Well there is the assumption that `device_->architecture()->name()->utf8String()` can't be the empty string. Given that the PR fixes the error of accessing positions -3 and -2 and also assuming these were in 0-9.

PR #3179, opened 2026-02-27
[CUDA] implement Hadamard transform

Lyxot:
## Proposed changes

This PR adds CUDA support for `Hadamard` (`mx.hadamard_transform`) with the same staged decomposition strategy used by the Metal backend.

## Changed files
- `mlx/backend/cuda/hadamard.cu`: implemented CUDA `Hadamard::eval_gpu` and JIT launch flow (`n1`/`n2`/`m` staged execution), reusing `decompose_hadamard(...)`.
- `mlx/backend/cuda/device/hadamard.cuh`: added JIT device kernels `hadamard_n<...>` and `hadamard_m<...>` plus radix helpers.
- `python/tests/cuda_skip.py`: removed CUDA skip entries

## Validation
- `python -m pytest python/tests/test_ops.py -k test_hadamard -q` passed.
- `python -m pytest python/tests/test_ops.py -k test_hadamard_grad_vmap -q` passed.

## Checklist

### DIFF
--- a/mlx/backend/cuda/hadamard.cu
+++ b/mlx/backend/cuda/hadamard.cu
@@ -159,8 +159,8 @@ void hadamard_mn_contiguous(
     args.append(num_transforms);
 
     auto kernel = mod.get_kernel(n1_kernel_name);
-    encoder.add_kernel_node(
-        kernel, num_blocks, n1 / max_radix_1, 0, args.args());
+    encoder.add_kernel_node_raw(
+        kernel, num_blocks, n1 / max_radix_1, {}, 0, args.args());
   }
 
   {
@@ -179,8 +179,8 @@ void hadamard_mn_contiguous(
     args.append(num_transforms);
 
     auto kernel = mod.get_kernel(n2_kernel_name);
-    encoder.add_kernel_node(
-        kernel, num_blocks, n2 / max_radix_2, 0, args.args());
+    encoder.add_kernel_node_raw(
+        kernel, num_blocks, n2 / max_radix_2, {}, 0, args.args());
   }
 
   if (m > 1) {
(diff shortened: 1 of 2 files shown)


mlx/backend/cuda/hadamard.cu
```diff
@@ -0,0 +1,244 @@
...
+    encoder.set_output_array(y);
+
+    cu::KernelArgs args;
+    args.append(x);
+    args.append(y);
+    args.append(scale_n1);
+    args.append(num_transforms);
+
+    auto kernel = mod.get_kernel(n1_kernel_name);
+    encoder.add_kernel_node(
```
zcbenz: Needs to use `add_kernel_node_raw` after https://github.com/ml-explore/mlx/pull/3168, sorry for the breakage.

nastya236:
Looks great, thanks for the contribution.

Could you please share bandwidth numbers for the proposed kernel across a range of shapes? I’m also particularly interested in case where the Hadamard transform is applied to tiled inputs with `N=16` or `N=32`. Something like:

```python
x = mx.random.uniform(shape=(4096, 4096))
mx.hadamard_transform(x.reshape(4096, 4096 // N, N))
```
where `N=16` or `N=32`.

Lyxot:
@nastya236 
> Could you please share bandwidth numbers for the proposed kernel across a range of shapes? I’m also particularly interested in case where the Hadamard transform is applied to tiled inputs with `N=16` or `N=32`. Something like:
> 
> ```python
> x = mx.random.uniform(shape=(4096, 4096))
> mx.hadamard_transform(x.reshape(4096, 4096 // N, N))
> ```
> 
> where `N=16` or `N=32`.

Here's the requested bandwidth test running on 4070 Super
```
shape=(4096, 256, 16) N=16 warmup=5 iters=50 time_ms=0.9020 bw_GBps=148.80
shape=(4096, 128, 32) N=32 warmup=5 iters=50 time_ms=0.6797 bw_GBps=197.46
```

Additionally, bench results from `benchmarks/python/hadamard_bench.py`

bench_float16
bench_float32

Lyxot:
@nastya236 I believe this addresses the bandwidth question. Please let me know if I missed anything.

nastya236:
Looks great, thanks! I will wait for @zcbenz and then we can merge.

PR #3180, opened 2026-02-27
[CUDA] Quantized GEMV

zcbenz:

Implements a qmv kernel using CUTLASS to do vectorized dequantization and fma, which works for all types of quants.

This kernel is fast for small problems for FP32xINT8, measured on A100:

| M | N | K  | QMV (TFlop/s) | CUBLAS (TFlop/s) | QMV (GiB/s) | CUBLAS (GiB/s) | Speedup (x) |
| 1 | 4096 | 4096 | 2.40 | 0.71 | 1351.7 | 1414.2 | 3.39 |
| 1 | 8192 | 8192 | 2.26 | 0.85 | 1272.7 | 1694.4 | 2.67 |
| 1 | 16384 | 16384 | 2.47 | 0.87 | 1389.7 | 1749.5 | 2.82 |

The memory bandwidth is somehow lower for FP16xINT8:

| M | N | K | QMV (TFlop/s) | CUBLAS (TFlop/s) | QMV (GiB/s) | CUBLAS (GiB/s) | Speedup (x) |
| 1 | 4096 | 4096 | 2.10 | 1.47 | 1118.1 | 1466.7 | 1.43 |
| 1 | 8192 | 8192 | 1.95 | 1.49 | 1039.1 | 1494.6 | 1.31 |

[Independent C++ source code for profiling the kernel](https://github.com/zcbenz/qmm/blob/0dedd1073438cb2b742de21149e18013e1e40324/qmv.cu)

The memory bandwidth drops to half for FP8/FP4/INT4 quants unfortunately, which is likely because CUTLASS does not implement fast vectorized conversions for them. We can fix it by writing specializations of `dequant_fma` and I'll continue in followup PRs.

This PR also does some refactoring to dispatch `quantized_mamtul` to the fastest kernel depending on the problems size. For now we still prefer `fp_qmv` over `qmv` for FP8/FP4 quants but eventually I will merge `fp_qmv` into `qmv`.


mlx/backend/cuda/quantized/qmm/qmm.cpp
```diff
@@ -21,6 +21,46 @@ void qmm_impl_sm90(
...
+  }
+  if (group_size < k) {
+    return false;
+  }
+  if (mode != QuantizationMode::Affine) {
+    return false;
+  }
+  if (device.compute_capability_major() != 9) {
+    return false;
+  }
```
jagrit06: Maybe we can move this to the top so we exit early ?

jagrit06:
Just so I understand the comparison - in table 1, the cublas is doing FP32xFP32 and in table 2 cublas is doing FP16xFP16 ?

zcbenz:
> Just so I understand the comparison - in table 1, the cublas is doing FP32xFP32 and in table 2 cublas is doing FP16xFP16 ?

Yeah cublas was measured with activation dtype.

mlx/backend/cuda/quantized/qmm/fp_qmv.cu
```diff
@@ -259,15 +264,17 @@ void fp_qmv(
         dispatch_bool(B > 1, [&](auto batched) {
```
angeloskath: I don't think we need the `dispatch_bool` here do we? It is never really used as a constexpr anywhere afaict.

zcbenz:
There was a mistake that the native dtype (`__half`) instead of cutlass dtype (`cutlass::half_t`) was passed to cutlass, fixing it makes FP16xINT8 a lot faster:

| M | N | K | QMV (TFlop/s) | CUBLAS (TFlop/s) | QMV (GiB/s) | CUBLAS (GiB/s) | Speedup (x) |
| 1 | 4096 | 4096 | 2.64 | 1.42 | 1406.0 | 1420.9 | 1.86 |
| 1 | 8192 | 8192 | 2.37 | 1.48 | 1261.5 | 1477.7 | 1.61 |
| 1 | 16384 | 16384 | 2.55 | 1.69 | 1352.4 | 1691.0 | 1.51 |

The sub-byte quants are still slow though.

mlx/backend/cuda/quantized/qmm/qmv.cu
```diff
@@ -0,0 +1,280 @@
...
+template <int N, typename Q>
+__device__ __forceinline__ void
+dequant_fma(const float* x, const Q* w, float scale, float bias, float* out) {
+  auto x_vec = *(reinterpret_cast<const cutlass::AlignedArray<float, N>*>(x));
+  auto w_vec = *(reinterpret_cast<const cutlass::AlignedArray<Q, N>*>(w));
+  auto* out_vec = reinterpret_cast<cutlass::Array<float, N>*>(out);
+
+  cutlass::NumericArrayConverter<float, Q, N> converter;
+  cutlass::Array<float, N> w_dq = converter(w_vec);
+#pragma unroll
```
angeloskath, line 55: Probably like above `w_dq = w_dq * scale + bias`.

zcbenz: For `float` it is slower, I think it is because there is no vectorized instructions and there would be new registers used.

mlx/backend/cuda/quantized/qmm/fp_qmv.cu
zcbenz: That is right, good catch!

PR #3184, opened 2026-02-28
Skip Hopper-only kernels in CI

zcbenz:
The QMM kernels broke the CI due to OOM, we can work around it by using `-j 1` but it would then increase build time to 3 hours. Disable them in CI until we get more RAM.


angeloskath (approved):
🙏

PR #3194, opened 2026-03-03
Hybrid sharding

nastya236:
Added hybrid sharding to fsdp. 
I refactored the code, because I thought it would be a bit more consistent, but I can revert it back and simply call `mx.distributed.all_sum`.

### DIFF
--- a/python/mlx/nn/utils.py
+++ b/python/mlx/nn/utils.py
@@ -128,14 +128,12 @@ def average_gradients(
 
     if all_reduce_size <= 0:
         return tree_map(
-            lambda x: _comm_fn(
+            lambda x: mx.distributed.all_sum(
                 x,
-                "all_sum",
-                group,
-                communication_type,
-                communication_stream,
-                average=True,
-            ),
+                group=group,
+                stream=communication_stream,
+            )
+            / N,
             gradients,
         )
 
@@ -149,15 +147,9 @@ def average_gradients(
 
         # We can't group them if they have mixed types
         if not all(dt == dtypes[0] for dt in dtypes):
-            return average_gradients(gradients, group, 0, communication_type)
-        itemsize = (
-            communication_type.size
-            if communication_type is not None
-            else dtypes[0].size
-        )
-
+            return average_gradients(gradients, group, 0)
         # Gather the gradients in groups that are just above or equal to all_reduce_size
-        grad_groups = _group_by_size(keys, sizes, itemsize, all_reduce_size)
+        grad_groups = _group_by_size(keys, sizes, dtypes[0].size, all_reduce_size)
 
         # Concatenate-reduce-split
         new_flat_grads = []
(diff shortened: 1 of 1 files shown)


python/mlx/nn/utils.py
```diff
@@ -96,6 +96,18 @@ def _group_by_size(keys, sizes, itemsize, communication_size):
...
+def _comm_fn(x, comm_op, group, type, stream, average=False):
+    N = group.size() if average else 1
+    dt = x.dtype
+    x = x.astype(type) if type is not None else x
+    if comm_op == "all_sum":
+        return mx.distributed.all_sum(x, stream=stream, group=group).astype(dt) / N
+    elif comm_op == "sum_scatter":
+        return mx.distributed.sum_scatter(x, stream=stream, group=group).astype(dt) / N
+    elif comm_op == "all_gather":
+        return mx.distributed.all_gather(x, stream=stream, group=group).astype(dt) / N
```
angeloskath: If you do want to refactor it out I would change this as follows.
1. You don't have to rewrite it for every comm operation
2. You don't divide with 1 when not averaging
3. You don't need a lambda to use it

We could also consider removing the communication dtype. I don't think anyone is using it tbh and it just makes everything uglier 🤷‍♂️ It's a backwards incompatible change but it 's probably fine.

```suggestion
def _comm_fn(x, comm_op, group=None, dtype=None, stream=None, average=False):
    def _inner(x):
        op = getattr(mx.distributed, comm_op)
        group = group or mx.distributed.init()
        dt = x.dtype
        if dtype is not None:
            x = x.astype(dtype)
        x = op(x, stream=stream, group=group).astype(dt)
        if average and group.size() > 1:
            x /= group.size()
        return x
    return _inner
```

python/mlx/nn/utils.py
```diff
@@ -130,13 +142,18 @@ def average_gradients(
     if N == 1:
         return gradients
 
-    def _average(x):
```
angeloskath, line 133: Given the previous comment I would simply

```python
_average = _comm_fn(x, "all_sum", group=communication_group, stream=communication_stream, average=True)
```

I also just noticed that the group was missing from `_average` so that had a bug before :-)

python/mlx/nn/utils.py
```diff
@@ -247,9 +275,14 @@ def fsdp_apply_gradients(
         ... )
         >>> model.update(updated_params)
     """
-    group = group or mx.distributed.init()
-    N = group.size()
-    rank = group.rank()
+    world = mx.distributed.init()
+    N = world.size()
+    fsdp_group = fsdp_group or world
```
angeloskath: That is interestingly incorrect.

`mx.distributed.init()` doesn't return the world communicator but the first one instantiated. There is no guarantee that it would even be the same type ie `world` could be TCP ring and `fsdp_group` could be NCCL.

python/mlx/nn/utils.py
angeloskath: I would probably also use `normalizer: float` instead of `average: bool` cause it is more convenient eg for hybrid fsdp where the normalizer is actually `1 / (fsdp_group.size() * dp_group.size())` rather than doing the division twice.

python/mlx/nn/utils.py
```diff
@@ -260,40 +293,43 @@ def fsdp_apply_gradients(
...
+        grad_slices[group_idx] = _comm_fn(
+            big_grad,
+            "sum_scatter",
+            fsdp_group,
+            communication_type,
+            communication_stream,
+            average=not fsdp_is_full_world,
         )
-        grad_slices[group_idx] = _sum_scatter(big_grad)
+        if not fsdp_is_full_world:
```
angeloskath: Same as above, the comms become way simpler to read if assigned to small functions like `_sum_scatter`. The data parallel average can also be a communication or a noop depending on `dp_group is None` so no ifs needed this will read straight down.

nastya236: I think if is needed unfortunately, because once a backend is initialized `all_sum(x, group=None)` will be a world-wide reduction 🤔

PR #3197, opened 2026-03-04
[CUDA] Use fp16 accumulation for 4-bit quant in GEMV

zcbenz:
Refs https://github.com/ml-explore/mlx/issues/2536, #3180.

Use fp16 for accumulation within threads and fp32 for accumulation between threads, so we can make use of faster half type vectorized instructions and avoid expensive promotions to fp32. I'm not really sure about precision issues but this is what every open source A16W4 kernels do so I guess it is fine.

Also do 64bit loads for 4-bit quant. Doing 128bit load slows things down unfortunately because we would be loading too many activations (32 elements) which uses too many registers.

This brings us decent memory bandwidth for FP16xINT4. It is slower than FP16xINT8 which I think is probably because the latter does 128bit loads.

| M | N | K | QMV (GiB/s) | CUBLAS (GiB/s) | Speedup |
| 1 | 4096 | 4096 | 1031.1 | 1427.5 | 2.57x |
| 1 | 8192 | 8192 | 1158.0 | 1476.1 | 2.79x |
| 1 | 16384 | 16384 | 1313.1 | 1691.5 | 2.76x |

(Speedup is compared to FP16xFP16 matmul.)

Note that I tried an optimization by loading scales/biases into shared memory first with vectorized loads, but unfortunately it made the kernel slower.

A few combinations that are still too slow:

* FP32xINT4: no vectorized int4=> fp32 conversions.
* FP quants: no vectorized fp4/8 => fp16/32 conversions.
* 2-bit quants: because we only do 32bit loads?

### DIFF
--- a/mlx/backend/cuda/quantized/qmm/qmv.cu
+++ b/mlx/backend/cuda/quantized/qmm/qmv.cu
@@ -20,6 +20,36 @@ namespace cg = cooperative_groups;
 // out = fma(x, w_dq, out)
 template <int N, typename T, typename Q>
 __device__ __forceinline__ void
+dequant_fma(const T* x, const Q* w, T scale, T bias, T* out) {
+  // Read x/w into registers.
+  auto x_vec = *(reinterpret_cast<const cutlass::AlignedArray<T, N>*>(x));
+  auto w_vec = *(reinterpret_cast<const cutlass::AlignedArray<Q, N>*>(w));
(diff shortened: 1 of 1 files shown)


angeloskath (approved):
Nice!

This is a prime candidate for going into PTX and looking at the code-gen. Not saying you should do that, just that it is rare to have such a simple case with complicated code-gen.

PR #3202, opened 2026-03-04
Fix shapeless compile eliding reductions on size-1 dimensions

### LINKED ISSUE #3201: [BUG] compile(shapeless=True): Reduce returns stale values on dynamically-shaped inputs from take/gather
## Summary

When a compiled function with `shapeless=True` performs a reduction (`sum`, `mean`) on the output of `mx.take` (GatherAxis), the reduction returns stale results from the first call. The `take` primitive correctly produces dynamic-sized output, but `Reduce`/`ReduceAll` replays the cached first-call result instead of re-executing.

Related: [ml-explore/mlx-c#104](https://github.com/ml-explore/mlx-c/issues/104) — same bug reproduced via the C API, confirming it's in mlx core.

## MLX Version

0.31.0

## Minimal Reproducer

```python
import mlx.core as mx

def _fn(buf, idx):
    taken = mx.take(buf, idx, axis=0)
    return taken.sum()

fn = mx.compile(_fn, shapeless=True)

buf = mx.array([10.0, 20.0, 30.0, 40.0, 50.0])
mx.eval(buf)

for n in [1, 2, 3, 4]:
    idx = mx.arange(n)
    s = fn(buf, idx)
    mx.eval(s)
    expected = sum(buf.tolist()[:n])
    print(f"n={n}: sum={s.item():.0f} expected={expected:.0f} "
          f"{'PASS' if abs(s.item() - expected) < 0.01 else 'FAIL'}")
```

### Output

```
n=1: sum=10 expected=10 PASS
n=2: sum=10 expected=30 FAIL    <-- should be 30
n=3: sum=10 expected=60 FAIL    <-- should be 60
n=4: sum=10 expected=100 FAIL   <-- should be 100
```

## Diagnostic Results

| Variant | shapeless=True | Result |
|---------|----------------|--------|
| `take` only (no reduction) | **PASS** | GatherAxis correctly returns dynamic-sized output |
| `take` → `matmul` | **PASS** | Matmul on dynamic-sized take output works |
| `take` → `sum` | **FAIL** | Sum always returns first-call value (10) |
| `take` → `mean` | **FAIL** | Returns `10/n` — sum part stale, count uses new size |
| `take` → `reshape(-1)` → `sum` | **FAIL** | Reshape doesn't help |
| `take` → `sum` with `shapeless=False` | **PASS** | Re-traces on shape change |
| `take` → `sum` without compile | **PASS** | No compilation, works correctly |

## Root Cause Hypothesis

The `mean` result (`10/n`) is the smoking gun: MLX implements mean as `sum / count`. The count correctly uses the new input size, but the sum returns the stale traced value.

In `compile_replace()` with `shapeless=True`, `Primitive::output_shapes()` is called to infer output shapes for replayed nodes. For `Reduce` with reduce-all, the output is always scalar `[]`, so the shape doesn't change between calls. The reduction kernel appears to either:

1. Not be re-dispatched when input shapes change (since output shape is constant), or
2. Use a cached intermediate buffer sized for the first-call input

Since the output shape is always `[]` regardless of input size, the shapeless machinery may be treating the output as unchanged and reusing the previous buffer without re-executing the kernel.

## Impact

This blocks using `shapeless=True` for compiled functions that include reductions over variable-length inputs — a common pattern in LLM inference (e.g., attention over growing KV cache sequences). We discovered this while trying to use `mlx_compile` via the C API to fuse transformer decode sub-blocks for performance.

ghstrider:
## Summary

Fixes #3201 — `mx.compile(shapeless=True)` returns stale values when a reduction (`sum`, `mean`, `max`, etc.) operates on a dynamically-shaped input from `take`/`gather`.

## Root Cause

`compute_reduce_shape()` reports `is_noop=true` when a reduced dimension is size 1 at trace time. This causes `sum()` et al. to skip creating the `Reduce` primitive entirely. On replay with a larger dimension, the missing reduction means the graph goes from `GatherAxis` → `Squeeze` with no reduction in between, and only the first element survives.

## Fix

Disable the `is_noop` optimisation during `in_dynamic_tracing()`, following the same pattern already used for no-op broadcast elision (lines 1489, 1543 of `ops.cpp`).

**One-line change** in `compute_reduce_shape()`:
```cpp
if (detail::in_dynamic_tracing()) {
    is_noop = false;
}
```

This fixes all 8 affected reduction operations (`sum`, `prod`, `min`, `max`, `all`, `any`, `argmin`, `argmax`) in one place.

## Test Plan

- Added `test_shapeless_compile_reduce_after_gather` covering `sum`, `mean`, and `max` with varying index sizes starting from size 1
- Existing `test_shapeless_compile_with_reduction` continues to pass (it doesn't trigger the bug because its reduced dimensions are never size 1 at trace time)

### DIFF
--- a/mlx/ops.cpp
+++ b/mlx/ops.cpp
@@ -51,6 +51,13 @@ std::tuple<Shape, std::vector<int>, bool> compute_reduce_shape(
     is_noop &= (out_shape.back() == shape[i]);
   }
   std::vector<int> sorted_axes(axes_set.begin(), axes_set.end());
+  // During dynamic (shapeless) tracing, dimensions that happen to be size 1
+  // at trace time may have different sizes on replay.  Never elide the
+  // reduction in that case, otherwise the Reduce primitive is missing from
+  // the traced graph and replays produce wrong results.
+  if (detail::in_dynamic_tracing()) {
+    is_noop = false;
+  }
   return {out_shape, sorted_axes, is_noop};
 }
 
--- a/python/tests/test_compile.py
+++ b/python/tests/test_compile.py
@@ -1017,6 +1017,55 @@ class TestCompile(mlx_tests.MLXTestCase):
         self.assertEqual(out[0].shape, (3, 1, 4, 2))
         self.assertEqual(out[1].shape, (2, 2, 5))
 
+    def test_shapeless_compile_reduce_after_gather(self):
+        # Regression test: when the first call to a shapeless-compiled function
+        # has a size-1 reduced dimension, the reduction was elided as a no-op
+        # during tracing. On replay with larger sizes, the missing reduction
+        # caused stale (first-call) values to be returned.
+        buf = mx.array([10.0, 20.0, 30.0, 40.0, 50.0])
+
+        # take + sum
+        def fn_sum(buf, idx):
+            return mx.take(buf, idx, axis=0).sum()
+
+        cfn = mx.compile(fn_sum, shapeless=True)
+        for n in [1, 2, 3, 4]:
+            idx = mx.arange(n)
+            expected = fn_sum(buf, idx)
+            result = cfn(buf, idx)
+            self.assertTrue(
+                mx.allclose(result, expected),
+                f"sum failed for n={n}: got {result.item()}, expected {expected.item()}",
+            )
+
+        # take + mean
+        def fn_mean(buf, idx):
+            return mx.take(buf, idx, axis=0).mean()
+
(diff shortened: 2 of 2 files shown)


angeloskath (dismissed):
I am not a huge fan of this approach where the graph changes based on whether we are in compile or not but since we are already doing it in other places I think we should merge this for now.

angeloskath:
Well it seems that the fixing is breaking other places.

zcbenz:
Tests are still failing after rebasing on main, so I'm closing this PR.

PR #3204, opened 2026-03-04
win: fix cuda build

dhiltgen:
Adjust recent CUDA changes to build on windows.

## Proposed changes

HEAD on main currently fails to build on Windows with CUDA enabled.  This gets the build working.

## Checklist

### DIFF
--- a/mlx/backend/cuda/scaled_dot_product_attention.cpp
+++ b/mlx/backend/cuda/scaled_dot_product_attention.cpp
@@ -318,7 +318,15 @@ bool supports_sdpa_cudnn(
     bool has_arr_mask,
     bool do_causal,
     Stream s) {
+#ifdef _WIN32
+  // On Windows (WDDM), cuDNN SDPA has severe performance issues due to
+  // high per-kernel-launch overhead in the WDDM driver model. cuDNN's
+  // multi-kernel SDPA amplifies this, making it much slower than the
+  // single-kernel sdpa_vector path for both prefill and generation.
+  static bool enabled = env::get_var("MLX_CUDA_USE_CUDNN_SDPA", 0);
+#else
   static bool enabled = env::get_var("MLX_CUDA_USE_CUDNN_SDPA", 1);
+#endif
   if (!enabled) {
     return false;
   }


mlx/backend/cuda/quantized/qmm/qmm.cpp
```diff
@@ -2,11 +2,10 @@
 
 #include "mlx/backend/cuda/quantized/qmm/qmm.h"
 
-#include <cute/tensor.hpp>
-
 namespace mlx::core {
 
 #if defined(MLX_CUDA_SM90A_ENABLED)
+#include <cute/tensor.hpp>
```
zcbenz: This moves the `include` into a namespace.

dhiltgen:
While testing more deeply I noticed sporadic JIT crashes on Windows CUDA and traced it down to missing device initialization, which I've fixed in a discrete commit.  Let me know if you want that split out to a different PR.

zcbenz:
The new change looks good to me, thanks for fixing it! Still need to wait for the CI environment to get fixed before I can merge.

dhiltgen:
I fixed another performance issue.  I can move it to another PR as well if needed.  It looks like cuDNN should be disabled by default on Windows due to how it interacts with WDDM.  Using mlx-community/Qwen3-0.6B-4bit to test with an RTX 6000 Ada on Win 11:
```
  ┌───────────┬─────────────────┬───────────────────┐
  │   Mode    │ Prefill (p2048) │ Generation (g128) │
  ├───────────┼─────────────────┼───────────────────┤
  │ cuDNN ON  │ 2,022 tok/s     │ 10.4 tok/s        │
  ├───────────┼─────────────────┼───────────────────┤
  │ cuDNN OFF │ 4,822 tok/s     │ 480 tok/s         │
  └───────────┴─────────────────┴───────────────────┘
```

zcbenz:
I'm good disabling cuDNN SDPA for Windows, but can you check with NVIDIA whether this is something that can be fixed? Our fallback kernel is not good for prefill and long context decoding.

Issue #2710, opened 2025-10-28
Support of boolean and symbolic indexing


cetagostini:
Any of the following indexing's are support. Would be nice to get these ones.

```python
import mlx.core as mx
import numpy as np
x = mx.array([1, 2, 3])
x[np.int64(1)]

---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Cell In[42], line 4
      2 import numpy as np
      3 x = mx.array([1, 2, 3])
----> 4 x[np.int64(1)]

ValueError: Cannot index mlx array using the given type.
```
```python
x[np.int64(1):]

ValueError                                Traceback (most recent call last)
Cell In[43], line 4
      2 import numpy as np
      3 x = mx.array([1, 2, 3])
----> 4 x[np.int64(1):]

ValueError: Slice indices must be integers or None.
```

PR #3210, opened 2026-03-05
Support indexing with any type which implmented `__index__`

aisk:
## Proposed changes

Any type which implemented `__index__` can be used to indexing mlx array, so we can use numpy scalar to indexing array.

`bool` is not supported also it implemented `__index__` (actually, `bool` is a subclass of `int`) because it's to quirky.

Close: https://github.com/ml-explore/mlx/issues/2710

## Checklist

### DIFF
--- a/python/src/indexing.cpp
+++ b/python/src/indexing.cpp
@@ -16,7 +16,20 @@ bool is_none_slice(const nb::slice& in_slice) {
 }
 
 bool is_index_scalar(const nb::object& obj) {
-  return !nb::isinstance<nb::bool_>(obj) && PyIndex_Check(obj.ptr());
+  if (nb::isinstance<nb::bool_>(obj)) {
+    return false;
+  }
+  if (!PyIndex_Check(obj.ptr())) {
+    return false;
+  }
+  // Exclude multi-dimensional arrays (mx.array, np.ndarray) by checking ndim
+  if (nb::hasattr(obj, "ndim")) {
+    auto ndim = nb::getattr(obj, "ndim");
+    if (nb::isinstance<nb::int_>(ndim) && nb::cast<int>(ndim) > 0) {
+      return false;
+    }
+  }
+  return true;
 }
 
 int safe_to_int32(nb::object obj) {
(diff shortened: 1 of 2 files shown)


python/src/indexing.cpp
```diff
@@ -15,8 +15,30 @@ bool is_none_slice(const nb::slice& in_slice) {
       nb::getattr(in_slice, "step").is_none());
 }
 
+bool is_index_scalar(const nb::object& obj) {
+  if (nb::isinstance<nb::bool_>(obj)) {
+    return false;
+  }
+  if (!PyIndex_Check(obj.ptr())) {
+    return false;
+  }
```
angeloskath, line 24: I think this should simply be `return PyIndex_Check(obj.ptr()) && !nb::isinstance<nb::bool_>(obj);`

angeloskath (changes requested):
Numpy arrays and mlx arrays do not implement `__index__` so no need for the extra checks.

aisk:
Hi @angeloskath thank you for your review, but numpy has implmeneted `__index__`:

```
>>> np.int64(21).__index__()
21
>>> np.array([1,2]).__index__()
  File "<python-input-11>", line 1, in <module>
    np.array([1,2]).__index__()
    ~~~~~~~~~~~~~~~~~~~~~~~~~^^
TypeError: only integer scalar arrays can be converted to a scalar index
```

Even the `np.array` have `__index__` slots on it, but it will raise error while calling, so we can't simplify the codes by just check `PyIndex_Check(obj.ptr()) && !nb::isinstance<nb::bool_>(obj)`.

And there is also a test case for this:

https://github.com/ml-explore/mlx/blob/main/python/tests/test_array.py#L2065

Simplify the code will break it:

```
ERROR: test_setitem_with_boolean_mask (__main__.TestArray.test_setitem_with_boolean_mask)
  File "/Users/asaka/Codes/mlx/python/tests/test_array.py", line 2073, in test_setitem_with_boolean_mask
    mx.arange(1000).reshape(10, 10, 10)[mask_np] = 0
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^
TypeError: only integer scalar arrays can be converted to a scalar index
```

PR #3212, opened 2026-03-06
Remove custom fp4/fp8 classes

zcbenz:
We have some custom fp4/fp8 classes in `cuda/quantized/cuda_fp4.h` for building with not so recent CUDA toolkits (like 12.6), in which case there is no builtin equivalents.

This PR replaces them with the equivalents in CUTLASS, which generates same code with CUDA's builtin `__nv_fp8_xxs`/`__nv_fp8_xxx` and provides decent fallbacks for old CUDA toolkits.

The main benefit is we can just rely on `cutlass::NumericArrayConverter` to do vectorized number conversions, instead of dealing with various CUDA types.


nastya236:
Thank you, looks like a great alternative to custom header. Do you mind if I put myself as a reviewer?

zcbenz:
It is always a good thing having more people reviewing this 🙏

PR #3225, opened 2026-03-09
fix/jaccl ring 2node wire split

blightbow:
## Summary                                                                                                                                                                                                                                                                 
                                                                                                                                                                                                                                                                             
- Fix `extract_ring_connectivity()` returning duplicate device vectors for 2-node rings where `left_peer == right_peer`, causing all cables to appear in both `left_` and `right_` connection arrays                                                                       
- Add midpoint split so each direction gets dedicated cables (e.g., 3+3 from 6 cables)                                                                                                                                                                                     
- Add `is_valid_ring()` guard requiring even cable `count >= 2` for 2-node rings                                                                                                                                                                                             
- Swap cable halves for rank 1 in `configure_jaccl_ring()` so QP index-based pairing lands on the same physical wire after the split                                                                                                                                       

## Problem

For a 2-node ring (the only viable ring topology for direct link aggregation), `left` and `right` both resolve to the same peer index:

```cpp
int left = (rank + size() - 1) % size();   // size=2: left = 1
int right = (rank + 1) % size();            // size=2: right = 1
```

`extract_ring_connectivity()` returns `(devices_[rank][1], devices_[rank][1])` — the identical vector for both directions. RingGroup then creates N `left_` connections AND N `right_` connections all pointing to the same physical cables. Both directions would contend for the same wires instead of having dedicated cables per direction.

##  Fix

### jaccl.cpp (C++, 21 lines):
  - `is_valid_ring()`: When `size() == 2`, require `num_connections >= 2 && num_connections % 2 == 0`
  - `extract_ring_connectivity()`: When `left == right`, split the device vector at the midpoint — first half → left, second half → right

###  config.py (Python, 10 lines):
  - `configure_jaccl_ring():` For 2-node rings, swap cable halves for rank 1 so that after the C++ midpoint split, rank 0 `left_[k]` pairs with rank 1 `right_[k]` on the same physical wire


angeloskath:
Have you encountered any issue wrt to sharing the wires? I chose to leave it as it is more flexible. For instance it allows connecting two nodes with 1, 3, 5 wires.

I would be inclined to close this unless there is an issue identified...

blightbow:
I see where I went astray on this one. I'm working with two M3 Ultras (6 connections), and `MAX_CONNS = 4` plays poorly with `configure_jaccl_ring()` in a two node setup:

`"rdma": [null, ["rdma_en0", "rdma_en1", "rdma_en2", "rdma_en3", "rdma_en4", "rdma_en5"]]`

This got me thinking that the mistake was the _shaping_ in a two node configuration, but the actual problem lies elsewhere. I'm fine with closing this PR and opening a bug instead; I'm more than happy to defer to a more experienced touch on the fix.

I'll hold off on opening the bug until I get your feedback. It'll save a round of unnecessary feedback in the event that this is a deliberate failure scenario.

angeloskath:
To be fair I haven't tested the configuration above (6 is going to fail as you said). If you encounter problems with 4 file a bug otherwise we could enable 6 if it makes sense although I want to do an optimization path before going overboard with cables :-) .

I will close this one for now.

Issue #3228, opened 2026-03-09
[BUG] float16/bfloat16 assignment from std::vector<bool>::reference stores raw bit 0/1, not numeric 0.0/1.0


MillaFleurs:
**Describe the bug**
The special assignment overload for `std::vector<bool>::reference` writes `bits_ = x;`, which encodes `true` as bit-pattern `0x0001` instead of the numeric value `1.0`.  std::vector<bool> is a notorious C++ special case. Its elements are not real bool& references. They are proxy objects

Evidence:
- `mlx/types/fp16.h:31-33`
- `mlx/types/bf16.h:31-33`

Why this is a bug:
- `true` should map to numeric `1.0`, not a tiny subnormal value.
- This produces incorrect values when assigning bool-proxy values into half/bfloat objects.

**To Reproduce**

```python
import mlx.core as mx

# Direct construction from bool list with explicit dtype
a = mx.array([True, False, True], dtype=mx.bfloat16)
mx.eval(a)
print(a.tolist())
# ACTUAL:   [9.183549615799121e-41, 0.0, 9.183549615799121e-41]
# EXPECTED: [1.0, 0.0, 1.0]

# float16 has the same buggy code but appears to work on Apple clang:
b = mx.array([True, False, True], dtype=mx.float16)
mx.eval(b)
print(b.tolist())
# ACTUAL:   [1.0, 0.0, 1.0]  (happens to work, but code is still wrong)

# Workaround — create as bool first, then cast:
c = mx.array([True, False, True]).astype(mx.bfloat16)
mx.eval(c)
print(c.tolist())
# ACTUAL:   [1.0, 0.0, 1.0]  (correct — astype goes through proper conversion)

```

**Expected behavior**
A boolean value should be 1.0 for true 0.0 for false

**Desktop (please complete the following information):**
 - OS Version: 26.3.1 on M2 Macbook

PR #3229, opened 2026-03-09
Fix assigning bool to float16/bfloat16 

MillaFleurs:
## Proposed changes

The special assignment overload for `std::vector<bool>::reference` writes `bits_ = x;`, which encodes `true` as bit-pattern `0x0001` instead of the numeric value `1.0`.  This fixes that so true values are returned as 1.0 instead of 9.183549615799121e-41

#3228 

## Checklist

### DIFF
--- a/mlx/types/fp16.h
+++ b/mlx/types/fp16.h
@@ -8,6 +8,7 @@
 #include <vector>
 
 #define __MLX_HALF_NAN__ 0x7D00
+#define __MLX_HALF_ONE__ 0x3C00
 
 namespace mlx::core {
 
@@ -29,8 +30,7 @@ struct _MLX_Float16 {
 
   // Appease std::vector<bool> for being special
   _MLX_Float16& operator=(std::vector<bool>::reference x) {
-    // 0x3c00 == TRUE == 1.0
-    bits_ = (x) ? 0x3c00 : 0;
+    bits_ = (x) ? __MLX_HALF_ONE__ : 0;
     return (*this);
   }
 


mlx/types/bf16.h
```diff
@@ -29,8 +29,7 @@ struct _MLX_BFloat16 {
 
   // Appease std::vector<bool> for being special
   _MLX_BFloat16& operator=(std::vector<bool>::reference x) {
-    bits_ = x;
-    return *this;
+    return (*this = _MLX_BFloat16(static_cast<float>(static_cast<bool>(x))));
```
angeloskath: I would write it as follows

```
bits_ = (x) ? 0x3f80 : 0;
```

no need to do all that math just to convert a bool.

mlx/types/fp16.h
```diff
@@ -29,8 +29,7 @@ struct _MLX_Float16 {
 
   // Appease std::vector<bool> for being special
   _MLX_Float16& operator=(std::vector<bool>::reference x) {
-    bits_ = x;
-    return *this;
+      return (*this = _MLX_Float16(static_cast<float>(static_cast<bool>(x))));
```
angeloskath: Similarly 

```
bits_ = (x) ? 0x3c00 : 0;
```

mlx/types/bf16.h
MillaFleurs: Agreed that is much clearer I would just recommend a comment above it, like so...

```cpp
// 0x3f80 == 1.0 == TRUE
bits_ = (x) ? 0x3f80 : 0; 
```

Make sense?

MillaFleurs: Updated!

mlx/types/fp16.h
MillaFleurs: Updated!

angeloskath (approved):
👍

zcbenz:
I made some edits to the PR titles, please feel free to rename to more meaningful ones — it is not helpful when there are commits with titles like "PR#3226 Fix (#3227)".

angeloskath:
> it is not helpful when there are commits with titles like "PR#3226 Fix (https://github.com/ml-explore/mlx/pull/3227)".

Yeah totally. I usually change it in the commit but I think this is much better as it allows for easier searching past merged PRs on GitHub.

MillaFleurs:
Sorry about that I’ll make sure my PR Titles are better going forward...

Issue #3222, opened 2026-03-09
[BUG] einsum_path prints uninitialized optimized metrics for <=2 inputs


MillaFleurs:
**Describe the bug**
 User-visible API output (`einsum_path` info string) can be incorrect and nondeterministic for common equations with one or two operands.  The output produced is garbage and should be fixed.

**To Reproduce**

Include code snippet
```python
import mlx.core as mx

# Two-input einsum — the most common use case (matrix multiply)
a = mx.ones((3, 4))
b = mx.ones((4, 5))
path, info = mx.einsum_path("ij,jk->ik", a, b)
print(info)

# Output (actual, from mlx 0.31.0):
#   Complete contraction:  ij,jk->ik
#          Naive scaling:  3
#      Optimized scaling:  45882692432      <-- GARBAGE
#       Naive FLOP count:  120
#   Optimized FLOP count:  45888039760      <-- GARBAGE
#
# Expected: Optimized scaling = 3, Optimized FLOP count = 120
# (for <=2 inputs, optimized == naive since no contraction reordering is possible)

# Single-input einsum — same bug
path, info = mx.einsum_path("ij->ji", a)
print(info)
# Optimized scaling and FLOP count are also garbage.

# Three-input einsum — works correctly (uses greedy_path which sets the fields)
c = mx.ones((5, 6))
path, info = mx.einsum_path("ij,jk,kl->il", a, b, c)
print(info)
# Optimized scaling = 3, Optimized FLOP count = 384 — correct.


```

**Expected behavior**
For a one of two input einsum te naive scaling and optimized scaling should be the same, as should the naive and optimized FLOP counts.

**Desktop (please complete the following information):**
 - OS Version:26.3.1

**Additional context**
The bug creates incorrect logging info, it does not actually affect the value of the einsums

PR #3232, opened 2026-03-09
Fix return value in einsum_path for simple contractions

MillaFleurs:
## Proposed changes

#3222 

**Describe the bug**
User-visible API output (einsum_path info string) can be incorrect and nondeterministic for common equations with one or two operands. The output produced is garbage and should be fixed.
## Checklist


angeloskath:
Ah sorry, linter again. Will merge after.

MillaFleurs:
Sorry about that I had set up the pre-commit to run on any pushes upstream but it seems it was not working

MillaFleurs:
Sorry I thought that was the technical term?  My old CS professor always joked about being a janitor with all the garbage collection...  🤣

MillaFleurs:
done!

Issue #3233, opened 2026-03-09
[BUG] mx.split causes crash without throwing an error due to lacking validation for non-positive num_splits


MillaFleurs:
**Describe the bug**
`split(const array&, int num_splits, int axis, ...)` does not validate `num_splits > 0` before using it and can cause a crash of python.

  The issue is not whether users intentionally pass 0 — it's that invalid input
  crashes the process instead of raising a clean error. Every other invalid input
  to split() raises a ValueError with a helpful message (bad axis, non-divisible
  shape). num_splits <= 0 is the only case that kills the process.

  NumPy comparison: np.split(arr, 0) raises ValueError cleanly. np.split(arr, -1)
  also raises ValueError. MLX should match this behavior.
  (ref: https://numpy.org/doc/stable/reference/generated/numpy.split.html)

  The fix is a 5-line guard. The cost of NOT fixing is:
  - A user's entire Python process dies with SIGBUS, no traceback, no error message
  - Debugging requires a crash report to even identify which call caused it
  - Any code that computes num_splits dynamically is at risk:
      num_batches = len(data) // batch_size  → 0 if batch_size > len(data)
      num_workers = len(available_gpus)      → 0 in degenerate configs
      num_chunks = num_items // chunk_size   → 0 for small inputs

 ** Security/fuzzing consideration:**
  - This is a process-crashing bug triggered by a single integer argument to a
    public Python API. Fuzzers (e.g., OSS-Fuzz, Atheris) will find this
    immediately — passing boundary values like 0, -1, INT_MIN to integer
    parameters is basic fuzzing strategy.
  - A library that crashes on trivial invalid input rather than raising an
    exception is a red flag in security audits. Even if MLX doesn't process
    untrusted input directly, it may be embedded in services that do (e.g.,
    user-specified batch sizes in ML serving pipelines).
  - The fix is trivially small relative to the risk surface it closes.

**To Reproduce**

Include code snippet
```python
import mlx.core as mx
a = mx.arange(6)

# Case 1: num_splits = 0
# std::ldiv(6, 0) is undefined behavior (divide-by-zero).
# On macOS ARM64 it happens to not crash because ldiv returns a non-zero
# remainder, which the existing remainder check catches. But this is UB
# and not guaranteed on other platforms/compilers.
# uncomment and the crash is a ValueError traceback which we got to accidentally
# by the same bug...
# mx.split(a, 0)
# Result: ValueError (by accident, via UB — not a clean validation)

[...]

PR #3234, opened 2026-03-09
Validate num_splits in split

MillaFleurs:
#3233 

Describe the bug
split(num_splits) lacks validation for non-positive num_splits, enabling divide-by-zero and invalid allocations
split fails to throw an error instead it crashes python3 

Checklist

Put an x in the boxes that apply.

[x] I have read the [CONTRIBUTING](https://github.com/ml-explore/mlx/blob/main/CONTRIBUTING.md) document
[x] I have run pre-commit run --all-files to format my code / installed pre-commit prior to committing changes
[x] I have added tests that prove my fix is effective or that my feature works
[x] I have updated the necessary documentation (if needed)


MillaFleurs:
Updated with linter run now, sorry for the confusion!

PR #3236, opened 2026-03-10
[CUDA] Support 3/5/6-bit quants in QMV

zcbenz:

This PR adds some `cutlass::NumericArrayConverter` specializations for converting 3/5/6-bit ints to float, which I will move into a header (like `cutlass_utils.cuh`) when adding QMM implementations.


angeloskath (approved):
Looks great, very clean!

PR #3237, opened 2026-03-10
Remove quantized_utils.cuh

zcbenz:
We no longer really need a `quantized_utils.cuh` after removing the `__nv_fp8` classes in #3212:

- Move `get_pack_factor`/`get_bytes_per_pack` to `backend/common/quantized.h`: share the code between CPU and CUDA backends.
- Move `dispatch_groups`/`dispatch_bits` to `quantized/affine_quantize.cu`: there are several QMM kernels in CUDA backend optimized for different quantizations so each kernel has their own version of `dispatch_groups`.
- Move `absmax_x2` to `quantized/fp_quantize.cu`: it is only used there.


nastya236:
Looks great, thank you for doing this! It makes a lot of sense to move dispatch to a file related to affine quantization.

nastya236 (approved):

PR #3238, opened 2026-03-10
[CUDA] Implement SegmentedMM

Lyxot:
## Proposed changes

Implement `mx.segmented_mm` for the CUDA backend using CUTLASS grouped GEMM.

## Performance

`MLX_ENABLE_TF32=0`, random segments.

**float32**
```
| Case              | MLX ms | Loop ms | Speedup |
|-------------------|--------|---------|---------|
| 128x128x1024x16   | 0.040  | 0.353   | 8.73x   |
| 128x128x1024x32   | 0.042  | 0.645   | 15.49x  |
| 256x256x2048x16   | 0.130  | 0.566   | 4.35x   |
| 512x512x4096x32   | 0.200  | 0.788   | 3.95x   |
| 1024x1024x4096x32 | 0.596  | 1.509   | 2.53x   |
| 1024x1024x8192x64 | 1.197  | 2.929   | 2.45x   |
```

**bfloat16**
```
| Case              | MLX ms | Loop ms | Speedup |
|-------------------|--------|---------|---------|
| 128x128x1024x16   | 0.041  | 0.184   | 4.54x   |
| 128x128x1024x32   | 0.032  | 0.327   | 10.35x  |
| 256x256x2048x16   | 0.061  | 0.186   | 3.05x   |
| 512x512x4096x32   | 0.161  | 0.395   | 2.46x   |
| 1024x1024x4096x32 | 0.515  | 0.791   | 1.54x   |
| 1024x1024x8192x64 | 1.095  | 1.593   | 1.46x   |
```

`MLX ms` = `mx.segmented_mm` (CUTLASS grouped GEMM), `Loop ms` = MLX loop-of-matmuls baseline.

## Checklist

### DIFF
--- a/python/tests/test_conv.py
+++ b/python/tests/test_conv.py
@@ -1151,7 +1151,7 @@ class TestConv(mlx_tests.MLXTestCase):
             )
             self.assertEqual(grads.shape, k_shape)
 
-    def test_1d_conv_with_2d(self):
+    def test_conv_1d_with_2d(self):
         x = mx.random.uniform(shape=(2, 10, 16))
         y = mx.random.normal(shape=(16, 3, 16))
 


mlx/backend/cuda/gemms/grouped_gemm_unaligned.cu
```diff
@@ -93,6 +93,52 @@ __global__ void prepare_grouped_mm_data(
...
+    int8_t* b_start,
+    int8_t* out_start,
+    ProblemSize* problem_sizes,
+    int64_t* a_lds,
+    int64_t* b_lds,
+    int64_t* out_lds,
+    void** a_ptrs,
+    void** b_ptrs,
+    void** out_ptrs) {
+  int idx = blockIdx.x * blockDim.x + threadIdx.x;
```
zcbenz: We usually use `cg::this_grid().thread_rank()`.

mlx/backend/cuda/gemms/grouped_gemm_unaligned.cu
```diff
@@ -93,6 +93,52 @@ __global__ void prepare_grouped_mm_data(
...
+  if (idx >= num_segments)
+    return;
+
+  uint32_t start = segments[2 * idx];
+  uint32_t end = segments[2 * idx + 1];
```
zcbenz: We can save it as `int64_t` so we don't have to do conversions everywhere.
```suggestion
  int64_t end = segments[2 * idx + 1];
```

mlx/backend/cuda/matmul.cpp
```diff
@@ -370,4 +370,58 @@ void GatherMM::eval_gpu(const std::vector<array>& inputs, array& out) {
...
+  nvtx3::scoped_range r("SegmentedMM::eval_gpu");
+  auto& s = stream();
+  auto& encoder = cu::get_command_encoder(s);
+
+  assert(inputs.size() == 3);
+  auto& a_pre = inputs[0];
+  auto& b_pre = inputs[1];
+  auto& segments_pre = inputs[2];
+
+  out.set_data(cu::malloc_async(out.nbytes(), encoder));
```
zcbenz: This should only be called when we decide to invoke the kernel, see where it is called in `GatherMM::eval_gpu`.

Lyxot:
Can't reproduce test failure on my setup

zcbenz:
For failures that only happened in CI, a plausible explanation is that there are code relying on uninitialized GPU memory being zeros, but we are not going to know since it happened in a cuDNN kernel. It is also possible that we have memory corruptions or out-of-bound writes, but it should be able to reproduce locally. Anyway changing the order of tests by renaming can work around it, and I have went through the code several times and I don't think it is the problem of this PR.

PR #3239, opened 2026-03-10
[CUDA] Use qmv kernel for fp quantizations

zcbenz:
Use the QMV kernel for fp quantizations.

Did a simple benchmarking and it is about 9% faster on A100.

<details>

```python
import time
import mlx.core as mx

M,N,K = (1, 16384, 16384)

x = mx.random.normal(shape=(M, K), dtype=mx.float16)
w = mx.random.normal(shape=(N, K), dtype=mx.float16)

w_q, scales = mx.quantize(w, mode='mxfp4')
y = mx.quantized_matmul(x, w_q, scales, transpose=True, mode='mxfp4')
mx.eval(y)

def fun():
    y = mx.quantized_matmul(x, w_q, scales, transpose=True, mode='mxfp4')
    mx.eval(y)

for _ in range(100):
    fun()

iterations = 1000
tic = time.time()
for _ in range(iterations):
    fun()
toc = time.time()

s = toc - tic
gb = iterations * (x.nbytes + w_q.nbytes + scales.nbytes + y.nbytes) / 1e9

print("{:5.2f}".format(gb / s))
```

</details>

### DIFF
--- a/mlx/backend/cuda/quantized/qmm/qmm.cu
+++ b/mlx/backend/cuda/quantized/qmm/qmm.cu
@@ -108,6 +108,11 @@ bool supports_fp_qmv(
     int group_size,
     QuantizationMode mode,
     cu::Device& device) {
+  // The fp_qmv kernel uses less registers and is faster for sm120. For sm80/90
+  // the qmv kernel is faster. We didn't test sm89/100.
+  if (device.compute_capability_major() <= 9) {
+    return false;
+  }
   bool non_batched = w.ndim() == 2;
   int k = x.shape(-1);
   int n = out.shape(-1);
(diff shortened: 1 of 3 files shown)


zcbenz:
Tested on H100 this is also 9% faster, but on DGX the old fp_qmv is at least 20% faster 😲.

zcbenz:
For sm120/121 the fp_qmv kernel is faster because it uses less registers, I'm not entirely sure whether this is a Blackwell thing or because the consumer hardwares have less registers and do not optimize vectorized instructions (There is no spare B200 for me to test).

Anyway I'm keeping fp_qmv for now for sm >= 100, and I'm going to work on another version of qmv that is implemented in a similar style using less registers, assuming the affine quantization would also benefit from it.

PR #3242, opened 2026-03-11
win: re-enable and fix cuDNN performance

dhiltgen:
## Proposed changes

Populating a fresh cuDNN CUDA graph for each layer, and adding that new graph to the overall MLX CUDA graph is costly with WDDM.  To resolve this we cache the graph (first call does the expensive populate_cuda_graph, subsequent calls only patch pointers via update_cuda_graph), and we cache the subgraph key to avoid the overhead of recomputing (kernel attribute queries that hit WDDM round-trip overhead.)

SDPACacheKey has bool fields adjacent to int64_t arrays which causes padding bytes for alignment.  The BytesKey constructor memsets everything to zero, but the aggregate init cache_key.pod = {...} creates a stack temporary with uninitialized padding, and the compiler's trivial copy-assignment copies the entire struct — including the garbage padding  — over the zeroed bytes. Since BytesKey uses memcmp for equality, every SDPA call produces a unique key and is a cache miss.

**Results (RTX 5090, Windows 11 WDDM, `mlx_lm benchmark -p 2048 -g 128`):**

| Model | Metric | main | After Fix | Change |
|-------|--------|--------------|-----------|--------|
| Llama-3.2-3B-4bit | Prefill (tok/s) | 1,228 | 2,436 | **+98%** |
| Llama-3.2-3B-4bit | Gen (tok/s) | 371 | 385 | +4% |
| Qwen3-8B-4bit | Prefill (tok/s) | 494 | 917 | **+86%** |
| Qwen3-8B-4bit | Gen (tok/s) | 220 | 229 | +4% |
| Llama-3.2-3B-bf16 | Prefill (tok/s) | 19,157 | 19,347 | +1% |
| Llama-3.2-3B-bf16 | Gen (tok/s) | 200 | 199 | ~0% |
| Qwen3-8B-bf16 | Prefill (tok/s) | 9,254 | 9,253 | ~0% |
| Qwen3-8B-bf16 | Gen (tok/s) | 91 | 91 | ~0% |

## Checklist

### DIFF
--- a/mlx/backend/cuda/scaled_dot_product_attention.cpp
+++ b/mlx/backend/cuda/scaled_dot_product_attention.cpp
@@ -150,10 +150,6 @@ inline BytesKey<SDPACacheKey> build_sdpa_cache_key(
     bool decoding = false,
     bool output_logsumexp = false) {
   BytesKey<SDPACacheKey> cache_key;
-  // Set fields individually to preserve zero-initialized padding bytes.
-  // Aggregate init (cache_key.pod = {...}) creates a stack temporary whose
-  // padding is uninitialized, and trivial copy-assignment copies the entire
-  // struct including padding — breaking BytesKey's memcmp-based comparison.
   cache_key.pod.device_id = encoder.device().cuda_device();
   cache_key.pod.cudnn_dtype = dtype_to_cudnn_type(q.dtype());
   cache_key.pod.q_shape = vector_key<QKV_NDIM>(q.shape());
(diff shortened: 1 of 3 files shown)


mlx/backend/cuda/scaled_dot_product_attention.cpp
```diff
@@ -150,21 +150,21 @@ inline BytesKey<SDPACacheKey> build_sdpa_cache_key(
...
-      do_causal,
-      {},
-      {},
-      sinks.has_value(),
-      output_logsumexp,
-  };
+  // Set fields individually to preserve zero-initialized padding bytes.
+  // Aggregate init (cache_key.pod = {...}) creates a stack temporary whose
+  // padding is uninitialized, and trivial copy-assignment copies the entire
+  // struct including padding — breaking BytesKey's memcmp-based comparison.
```
zcbenz: Can you move this comment to `BytesKey` and also update the call in `backend/cuda/conv.cpp`?

Issue #2561, opened 2025-09-01
[BUG] Provide CUDA implementation of FFT


ekamperi:
**Describe the bug**
Attempting an FFT calculation with the CUDA backend throws a runtime error. Not sure if it's a bug or a feature request.

**To Reproduce**
```bash
snapshots-gpu-4000adax1-20gb-tor1# cat reproduce_fft_error.py
#!/usr/bin/env python3
"""
Minimal reproduction of MLX FFT CUDA backend error.

This script demonstrates the error: "FFT has no CUDA implementation."
"""

import mlx.core as mx

def main():
    print("Testing MLX FFT with CUDA backend...")
    print(f"MLX version: {mx.__version__ if hasattr(mx, '__version__') else 'unknown'}")
    print(f"Default device: {mx.default_device()}")

    # Create a simple 1D signal
    x = mx.array([1.0, 2.0, 3.0, 4.0], dtype=mx.float32)
    print(f"Input array: {x}")

    try:
        # Attempt FFT operation that triggers the CUDA error
        result = mx.fft.fft(x)
        print(f"FFT result: {result}")
    except Exception as e:
        print(f"Error occurred: {e}")
        print(f"Error type: {type(e).__name__}")
        raise

if __name__ == "__main__":
    main()
snapshots-gpu-4000adax1-20gb-tor1# uv run python reproduce_fft_error.py
Testing MLX FFT with CUDA backend...
MLX version: 0.29.0
Default device: Device(gpu, 0)
Input array: array([1, 2, 3, 4], dtype=float32)
Error occurred: FFT has no CUDA implementation.
Error type: RuntimeError
Traceback (most recent call last):
  File "/root/logfiles/reproduce_fft_error.py", line 30, in <module>
    main()
  File "/root/logfiles/reproduce_fft_error.py", line 23, in main
    print(f"FFT result: {result}")
                        ^^^^^^^^
RuntimeError: FFT has no CUDA implementation.
snapshots-gpu-4000adax1-20gb-tor1#
```

**Expected behavior**
FFT should work with CUDA backend, or a fallback computation in the CPU should occur on the fly.

**Desktop**
```
snapshots-gpu-4000adax1-20gb-tor1# nvidia-smi
Mon Sep  1 14:06:40 2025
+-----------------------------------------------------------------------------------------+
| NVIDIA-SMI 575.57.08              Driver Version: 575.57.08      CUDA Version: 12.9     |
|-----------------------------------------+------------------------+----------------------+
[...]

PR #3243, opened 2026-03-11
[CUDA] Add FFT support

lucasnewman:
## Proposed changes

This adds CUDA parity with Metal for FFT ops, as requested in https://github.com/ml-explore/mlx/issues/2561

The implementation uses cufft and is inspired by `nvmath-python` which is referenced in that issue. I tried to make this as minimal as possible, and I verified all the tests for both python and cpp pass as expected, including the disabled-in-CI exhaustive one, `test_fft_exhaustive`.

I only had access to an H200 with CUDA 12.4 for testing -- if there's an easy way to test on other hardware, let me know and I can try it.

## Checklist

### DIFF
--- a/mlx/backend/cuda/fft.cu
+++ b/mlx/backend/cuda/fft.cu
@@ -57,9 +57,9 @@ enum class FFTTransformType : uint8_t {
 
 struct FFTPlanKey {
   int device_id;
-  uint8_t transform_type;
-  uint64_t n;
-  uint64_t batch;
+  FFTTransformType transform_type;
+  int64_t n;
+  int64_t batch;
 };
 
 struct CuFFTPlan {
@@ -100,9 +100,9 @@ FFTPlanKey make_plan_key(
     int64_t batch) {
   FFTPlanKey key{};
   key.device_id = device_id;
-  key.transform_type = static_cast<uint8_t>(transform_type);
-  key.n = static_cast<uint64_t>(n);
-  key.batch = static_cast<uint64_t>(batch);
+  key.transform_type = transform_type;
+  key.n = n;
+  key.batch = batch;
   return key;
 }
 
@@ -268,7 +268,7 @@ OrderedArray prepare_input(
   std::vector<int> order = current.order;
   if (!axis_last) {
     auto perm = move_axis_to_back_permutation(current.arr.ndim(), axis_pos);
-    view = transpose_view_in_eval(current.arr, perm);
+    view = transpose_in_eval(current.arr, perm);
     order = apply_permutation(current.order, perm);
   }
 
@@ -439,7 +439,5 @@ void FFT::eval_gpu(const std::vector<array>& inputs, array& out) {
 
   restore_output_layout(current, out);
 }
-
-#undef CHECK_CUFFT_ERROR
 
 } // namespace mlx::core
--- a/mlx/backend/gpu/copy.cpp
+++ b/mlx/backend/gpu/copy.cpp
@@ -4,6 +4,7 @@
 #include "mlx/primitives.h"
 
 #include <cassert>
+#include <numeric>
 
 namespace mlx::core {
 
@@ -39,27 +40,6 @@ array contiguous_copy_gpu(const array& arr, const Stream& s) {
   return arr_copy;
 }
 
-array transpose_view_in_eval(const array& x, const std::vector<int>& axes) {
-  Shape shape(axes.size());
-  Strides strides(axes.size());
-  for (int i = 0; i < axes.size(); ++i) {
-    shape[i] = x.shape(axes[i]);
-    strides[i] = x.strides(axes[i]);
-  }
-
-  auto [data_size, row_contiguous, col_contiguous] =
-      check_contiguity(shape, strides);
-  bool contiguous = x.flags().contiguous && data_size == x.data_size();
-
-  array out(std::move(shape), x.dtype(), nullptr, {});
-  out.copy_shared_buffer(
-      x,
-      std::move(strides),
-      {contiguous, row_contiguous, col_contiguous},
-      x.data_size());
-  return out;
-}
-
 array flatten_in_eval(const array& x, int start_axis, int end_axis, Stream s) {
   int ndim = x.ndim();
   if (start_axis < 0) {
@@ -80,6 +60,27 @@ array reshape_in_eval(const array& x, Shape shape, Stream s) {
   return out;
 }
 
+array transpose_in_eval(const array& x, const std::vector<int>& axes) {
+  Shape shape(axes.size());
+  Strides strides(axes.size());
+  for (int i = 0; i < axes.size(); ++i) {
+    shape[i] = x.shape(axes[i]);
+    strides[i] = x.strides(axes[i]);
+  }
+
+  auto [data_size, row_contiguous, col_contiguous] =
+      check_contiguity(shape, strides);
+  bool contiguous = data_size == x.data_size();
+
+  array out(std::move(shape), x.dtype(), nullptr, {});
+  out.copy_shared_buffer(
+      x,
+      std::move(strides),
+      {contiguous, row_contiguous, col_contiguous},
+      x.data_size());
+  return out;
+}
+
 array swapaxes_in_eval(const array& x, int axis1, int axis2) {
   int ndim = x.ndim();
   if (axis1 < 0) {
@@ -90,11 +91,9 @@ array swapaxes_in_eval(const array& x, int axis1, int axis2) {
   }
 
   std::vector<int> axes(ndim);
-  for (int i = 0; i < ndim; ++i) {
-    axes[i] = i;
-  }
+  std::iota(axes.begin(), axes.end(), 0);
   std::swap(axes[axis1], axes[axis2]);
-  return transpose_view_in_eval(x, axes);
+  return transpose_in_eval(x, axes);
 }
 
 } // namespace mlx::core
--- a/mlx/backend/gpu/copy.h
+++ b/mlx/backend/gpu/copy.h
@@ -53,7 +53,7 @@ void reshape_gpu(const array& in, array& out, Stream s);
 // Like the normal ops but safe to call in eval_gpu.
 array flatten_in_eval(const array& x, int start_axis, int end_axis, Stream s);
 array reshape_in_eval(const array& x, Shape shape, Stream s);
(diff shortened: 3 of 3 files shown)


mlx/backend/cuda/cuda_utils.h
```diff
@@ -6,17 +6,20 @@
...
 // Throw exception if the cuda API does not succeed.
 void check_cublas_error(const char* name, cublasStatus_t err);
+void check_cufft_error(const char* name, cufftResult err);
 void check_cuda_error(const char* name, cudaError_t err);
 void check_cuda_error(const char* name, CUresult err);
 void check_cudnn_error(const char* name, cudnnStatus_t err);
 
 // The macro version that prints the command that failed.
 #define CHECK_CUBLAS_ERROR(cmd) check_cublas_error(#cmd, (cmd))
+#define CHECK_CUFFT_ERROR(cmd) check_cufft_error(#cmd, (cmd))
```
zcbenz: The utility should be put in `fft.cu‎` since there is only one file using it, this header is used by too many files and we don't want to make it too heavy.

mlx/backend/cuda/fft.cu
```diff
@@ -0,0 +1,452 @@
...
+  size_t workspace_size;
+};
+
+struct OrderedArray {
+  array arr;
+  std::vector<int> order;
+};
+
+template <typename T>
+__global__ void scale_fft_output(T* out, T scale, size_t size) {
```
zcbenz: The kernels are usually put in the `cu` namespace, putting it in anonymous namespace would add a  long `<anonymous namespace>` in function name in profiler.

mlx/backend/cuda/fft.cu
```diff
@@ -0,0 +1,452 @@
...
+    }
+    throw;
+  }
+
+  auto plan = std::make_shared<CuFFTPlan>(
+      encoder.device().cuda_device(), handle, workspace_size);
+  return cache.emplace(key, plan).first->second;
+}
+
+array transpose_view(const array& in, const std::vector<int>& axes) {
```
zcbenz: This would be a nice utility to put in `mlx/backend/gpu/copy.h`.

mlx/backend/cuda/fft.cu
```diff
@@ -0,0 +1,452 @@
...
+    out_shape[i] = in.shape(axes[i]);
+    out_strides[i] = in.strides(axes[i]);
+  }
+
+  auto [data_size, row_contiguous, col_contiguous] =
+      check_contiguity(out_shape, out_strides);
+  bool contiguous = in.flags().contiguous && data_size == in.data_size();
+  if (!contiguous) {
+    row_contiguous = false;
+    col_contiguous = false;
```
zcbenz: I think it is possible for `row_contiguous` to be true when `contiguous` is false.

https://github.com/ml-explore/mlx/blob/ce45c52505c8158ea48d2a54e8caae05efd86bfe/mlx/array.h#L249-L266

mlx/backend/cuda/fft.cu
```diff
@@ -0,0 +1,452 @@
...
+} // namespace
+
+void FFT::eval_gpu(const std::vector<array>& inputs, array& out) {
+  nvtx3::scoped_range r("FFT::eval_gpu");
+  auto& s = stream();
+  auto& encoder = cu::get_command_encoder(s);
+  auto& in = inputs[0];
+
+  if (out.size() == 0) {
+    out.set_data(cu::malloc_async(0, encoder));
```
zcbenz: There is no need to set data for empty output.

mlx/backend/cuda/fft.cu
```diff
@@ -0,0 +1,452 @@
...
+};
+
+struct OrderedArray {
+  array arr;
+  std::vector<int> order;
+};
+
+template <typename T>
+__global__ void scale_fft_output(T* out, T scale, size_t size) {
+  auto index = static_cast<size_t>(blockIdx.x) * blockDim.x + threadIdx.x;
```
zcbenz: We usually use `cg::this_grid().thread_rank()`.

mlx/backend/cuda/fft.cu
```diff
@@ -0,0 +1,452 @@
...
+    auto transform_type = step_real
+        ? (inverse_ ? FFTTransformType::C2R : FFTTransformType::R2C)
+        : FFTTransformType::C2C;
+
+    // cuFFT may overwrite the input buffer for C2R, so only use the direct
+    // input when the transform is out-of-place from the library's perspective.
+    auto prepared = prepare_input(
+        current,
+        axis,
+        /* allow_direct= */ transform_type != FFTTransformType::C2R,
```
zcbenz: We can overwrite input if `is_donatable(in, out) == true`.

zcbenz:
Can you enable the FFT tests in `python/tests/cuda_skip.py` and `.github/actions/test-linux/action.yml`?

mlx/backend/cuda/fft.cu
lucasnewman: You're right, thanks!

mlx/backend/cuda/fft.cu
lucasnewman: I moved it and renamed it `transpose_view_in_eval` to match the self-documenting pattern -- let me know if that's what you had in mind. I also had `swapaxes_in_eval` call it directly since that one can be simplified now.

lucasnewman:
@zcbenz Thank you for the review! I think I've addressed everything you mentioned, and tests are passing for me on my CUDA setup. Let me know if there's anything else needed.

mlx/backend/cuda/fft.cu
```diff
@@ -0,0 +1,445 @@
...
+  C2C = 0,
+  R2C = 1,
+  C2R = 2,
+};
+
+struct FFTPlanKey {
+  int device_id;
+  uint8_t transform_type;
+  uint64_t n;
+  uint64_t batch;
```
zcbenz: You can just use original types to avoid castings.

```suggestion
  FFTTransformType transform_type;
  int64_t n;
  int64_t batch;
```

mlx/backend/cuda/fft.cu
```diff
@@ -0,0 +1,445 @@
...
+    case FFTTransformType::C2C:
+    case FFTTransformType::C2R:
+      return CUDA_C_32F;
+    case FFTTransformType::R2C:
+      return CUDA_R_32F;
+  }
+  throw std::runtime_error("[FFT] Unsupported cuFFT input transform type.");
+}
+
+cudaDataType_t output_type(FFTTransformType transform_type) {
```
zcbenz, line 120: I'm not quite familiar with FFT, could a transform have different input/output types? If not then `input_type`/`output_type` should be one function.

mlx/backend/gpu/copy.cpp
```diff
@@ -39,6 +39,27 @@ array contiguous_copy_gpu(const array& arr, const Stream& s) {
   return arr_copy;
 }
 
+array transpose_view_in_eval(const array& x, const std::vector<int>& axes) {
```
zcbenz: The function names should follow the corresponding op's name.

And can you also move the code so the order of implementations is the same with the order of declarations?

```suggestion
array transpose_in_eval(const array& x, const std::vector<int>& axes) {
```

mlx/backend/gpu/copy.cpp
```diff
@@ -39,6 +39,27 @@ array contiguous_copy_gpu(const array& arr, const Stream& s) {
...
+  Shape shape(axes.size());
+  Strides strides(axes.size());
+  for (int i = 0; i < axes.size(); ++i) {
+    shape[i] = x.shape(axes[i]);
+    strides[i] = x.strides(axes[i]);
+  }
+
+  auto [data_size, row_contiguous, col_contiguous] =
+      check_contiguity(shape, strides);
+  bool contiguous = x.flags().contiguous && data_size == x.data_size();
```
zcbenz: The check is redundant, let's just follow other call sites:

```suggestion
  bool contiguous = data_size == x.data_size();
```

mlx/backend/cuda/fft.cu
```diff
@@ -0,0 +1,445 @@
...
+  }
+
+  if (inverse_) {
+    apply_inverse_scale(current.arr, axes_, out, encoder);
+  }
+
+  restore_output_layout(current, out);
+}
+
+#undef CHECK_CUFFT_ERROR
```
zcbenz: This feels unnecessary.

mlx/backend/gpu/copy.cpp
```diff
@@ -68,22 +89,12 @@ array swapaxes_in_eval(const array& x, int axis1, int axis2) {
...
-  out.copy_shared_buffer(
-      x,
-      std::move(strides),
-      {contiguous, row_contiguous, col_contiguous},
-      x.data_size());
-  return out;
+  std::vector<int> axes(ndim);
+  for (int i = 0; i < ndim; ++i) {
+    axes[i] = i;
+  }
```
zcbenz:
```suggestion
  std::iota(axes.begin(), axes.end(), 0);
```

mlx/backend/cuda/fft.cu
angeloskath: Sorry to be pedantic, especially given that both the PR and review are awesome, however we cannot have `row_contiguous` be true without `contiguous` also being true. It is a strictly stronger restriction.

There are many ways to think about these 3 flags but basically `contiguous` means iff I process the memory from offset 0 to offset `data_size` one element at a time, I will have processed all elements exactly once.

`row_contiguous` means iff I process the array starting from offset 0 to offset `size >= data_size` one element at a time I will have processed all elements exactly once and in a row-wise order (`col_contiguous` is the same with column-wise order of course).

zcbenz: Thanks for the explanations, I still got confused about the 3 flags from time to time and you just gave the most intuitive way to understand them! 🙏

To confirm, we don't need to do `if (!contiguous) row_contiguous = false;` as the referenced line did because when we calculated that `row_contiguous` is true then there would be no way for `contiguous` to be false?

mlx/backend/cuda/fft.cu
lucasnewman: Yes, they are identical in the complex-to-complex case, but complex-to-real and real-to-complex transforms have different input and output types since the frequency domain representation includes a phase.

Issue #3205, opened 2026-03-05
[Enhancement] Group.split() support for JACCL and Ring backends (parity with NCCL #3172)


N0xMare:
## Summary

`Group.split()` was recently added to the NCCL backend in #3172 for multi-parallelism support. The JACCL and Ring backends currently throw `std::runtime_error` when `split()` is called. Adding `split()` to these backends would bring parity across all MLX distributed backends and unlock the full range of multi-group distributed patterns on Apple Silicon Thunderbolt 5 clusters.

## Motivation

`Group.split()` is a foundational primitive for creating communication subgroups allowing for other distributed patterns:

- **Hybrid tensor + pipeline parallelism**: TP within small node groups, PP between them.
- **Expert parallelism**: Distributing MoE experts across node subgroups instead of replicating all experts on every node. Reduces per-node memory for models like Kimi K2.5 where the vast majority of weights are unused per inference step.
- **Data parallel replicas**: Running independent model replicas on node subsets for concurrent serving. Each replica has its own TP subgroup for example.
- **Context parallelism**: Splitting long sequences across node subgroups, each handling a chunk of the context window. Relevant for architectures with large KV caches and large context.
- **Speculative decoding**: Draft model on one subgroup, verifier on another, coordinating candidate token generation and batch verification.
- **Multi-model pipelines**: Different models (vision encoder → LLM → reward model) on different subgroups with activation forwarding between them.

With TB5 RDMA and JACCL, fully connected topologies are limited by physical port count, and practical TP group sizes are further constrained by model dimension divisibility (attention heads, hidden sizes must divide evenly), making 2 and 4 the most common TP configurations.

| Machine | Chip | TB5 Ports | Full-Mesh TP Max (N-1 ≤ ports) |
|---------|------|-----------|-------------------------------|
| MacBook Pro | M4 Max | 3 | 4 nodes |
| Mac Studio | M4 Max | 4 | 5 nodes |
| Mac Studio | M3 Ultra | 6 | 7 nodes |

> **Note:** Full-Mesh TP Max is a hardware ceiling.

Subgroup support would allow clusters of Apple Silicon machines to apply a broader spectrum of parallelism strategies such as running large models at full precision across nodes that individually can't hold them, serving concurrent usage with replica groups, or distributing MoE experts efficiently.

MLX already has the higher-level building blocks:
- `shard_linear`/`AllToShardedLinear` for TP. 
- `PipelineMixin` with `send`/`recv` for PP
[...]

PR #3245, opened 2026-03-11
feat: implement Group.split() for JACCL and Ring backends

machiabeli:
## Summary

Implements `Group.split()` (MPI_Comm_split semantics) for the JACCL mesh, JACCL ring, and TCP ring distributed backends. This enables mixed parallelism strategies — such as combining tensor parallelism with pipeline parallelism — on Apple Silicon clusters connected via Thunderbolt 5 RDMA.

Addresses #3205

## Problem

`Group.split()` was unimplemented for all Apple Silicon distributed backends (`throw std::runtime_error("Group split not supported")`), blocking any use of sub-group collectives. This prevented:
- Mixed tensor + pipeline parallelism
- Expert parallelism for MoE models
- Any workflow requiring communicator subdivision (standard in MPI/NCCL)

### 1. TCPGroup Fallback for JACCL Sub-Groups

Apple's Thunderbolt 5 RDMA driver does **not** support multiple `ibv_context` instances on the same physical RDMA device simultaneously. If a sub-group tries to open new RDMA connections on devices already held by the parent group, the driver deadlocks.

**Solution:** Sub-groups created via `split()` on JACCL backends use `TCPGroup` — a new group implementation that performs all collective operations over TCP (via `SideChannel`). This is slower than RDMA but:
- Avoids the `ibv_context` concurrency deadlock entirely
- Keeps RDMA reserved for the parent group's high-bandwidth tensor parallelism
- Matches how MPI implementations handle transport fallback (e.g., MPICH's ch3:sock)

### 2. Strict Collective Synchronization

All `SideChannel::all_gather` calls in `split()` execute **before** any rank makes a branching decision (e.g., returning a `LocalGroup` for size-1 sub-groups). This matches MPI_Comm_split's collective semantics — the MPICH implementation internally uses `MPI_Allgather` with the same constraint.

### 3. LocalGroup for Size-1 Sub-Groups

Ranks that end up alone in their color group get a lightweight no-op implementation (identity for reductions, memcpy for gather). No RDMA or TCP connections opened.

### 4. Coordinator Address Derivation

Sub-group coordinators reuse the parent's coordinator host with port offsets based on the `color` parameter: `parent_port + 1000 + color`. This ensures no port collisions between concurrent sub-groups.

## Changes

| File | What |
|------|------|
| `mlx/distributed/jaccl/mesh.h` | Store `device_names_`, `coordinator_host_`, `coordinator_port_` for split() |
| `mlx/distributed/jaccl/mesh.cpp` | `MeshGroup::split()` + `LocalGroup` class |
| `mlx/distributed/jaccl/ring.h` | Store `all_devices_`, coordinator info for split() |
| `mlx/distributed/jaccl/ring.cpp` | `RingGroup::split()` + `TCPGroup` (full collective ops over TCP) + `RingLocalGroup` |
| `mlx/distributed/ring/ring.cpp` | `RingGroup::split()` using ring all-gather + port-offset sub-ring creation |
| `tests/test_jaccl_split.py` | 5 tests: same-color, two-group, three-group, key ordering, send/recv |

## Testing

Tested on a 4-node Apple Silicon cluster:
- 3× Mac Studio M3 Ultra (512GB, 256GB, 256GB)
- 1× Mac Studio M4 Max (128GB)
- Connected via Thunderbolt 5 with RDMA (jaccl-ring backend)

```bash
mlx.launch --hostfile /tmp/hostfile-jaccl-ring-4node.json tests/test_jaccl_split.py
```

All 5 tests pass: same-color split, even/odd two-group split, three-group split with size-1 sub-groups, key-reversed ordering, and send/recv on sub-groups (gracefully skipped for TCPGroup which doesn't support point-to-point).

## Hardware Context

This was developed and tested against Apple's first-generation Thunderbolt 5 RDMA driver (macOS 26.x, `infiniband/verbs.h` from Xcode SDK). The `ibv_context` concurrency limitation is specific to this driver — standard InfiniBand hardware (Mellanox ConnectX) supports multiple contexts via SR-IOV. The TCPGroup fallback is designed to be replaced with direct RDMA sub-group connections if/when Apple adds multi-context support.

### DIFF
--- a/mlx/distributed/jaccl/mesh.cpp
+++ b/mlx/distributed/jaccl/mesh.cpp
@@ -7,6 +7,58 @@
 
 namespace mlx::core::distributed::jaccl {
 
+// Lightweight group for size-1 sub-groups (no RDMA needed).
+class LocalGroup : public GroupImpl {
+ public:
+  Stream communication_stream(StreamOrDevice s) override {
+    return to_stream(s, Device::cpu);
+  }
+  int rank() override {
+    return 0;
+  }
+  int size() override {
+    return 1;
+  }
+  std::shared_ptr<GroupImpl> split(int color, int key = -1) override {
+    return std::make_shared<LocalGroup>();
+  }
+  void all_sum(const array& input, array& output, Stream stream) override {
+    copy_input(input, output, stream);
+  }
+  void all_max(const array& input, array& output, Stream stream) override {
+    copy_input(input, output, stream);
+  }
+  void all_min(const array& input, array& output, Stream stream) override {
+    copy_input(input, output, stream);
+  }
+  void all_gather(const array& input, array& output, Stream stream) override {
+    copy_input(input, output, stream);
+  }
+  void send(const array&, int, Stream) override {
+    throw std::runtime_error("[jaccl] Cannot send in a size-1 group.");
+  }
+  void recv(array&, int, Stream) override {
+    throw std::runtime_error("[jaccl] Cannot recv in a size-1 group.");
+  }
+  void sum_scatter(const array& input, array& output, Stream stream) override {
+    copy_input(input, output, stream);
+  }
+
+ private:
+  void copy_input(const array& input, array& output, Stream stream) {
+    auto& encoder = cpu::get_command_encoder(stream);
+    encoder.set_input_array(input);
+    encoder.set_output_array(output);
+    encoder.dispatch(
+        [in = input.data<char>(),
+         out = output.data<char>(),
+         n = input.nbytes()]() {
+          if (in != out)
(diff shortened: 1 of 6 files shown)


angeloskath:
Hi! Thanks for the PR.

Group split is something we plan to add for sure. However this PR is not adding it correctly. Firstly, one can reopen an ibv device there is no problem with that and the ring does it already when there are only 2 nodes. Secondly, no communication should ever happen over the side channel, that is for initiating the connection or perhaps metadata if needed in the future.

I will close the PR but feel free to open an issue to track progress for when we add this functionality in the near future.

Issue #3216, opened 2026-03-07
SIGSEGV in QuantizedMatmul::eval_gpu during long token generation on Mac Studio M2 Ultra


Thump604:
## Description

Sporadic SIGSEGV crashes during quantized inference on Mac Studio M2 Ultra (Mac14,14). The crash occurs in Apple's Metal driver (`AGXMetalG14X`) when MLX's `QuantizedMatmul::eval_gpu` creates a new command encoder while the driver is coalescing a previous compute command buffer.

The crash is **not** caused by concurrent API requests — we serialize all inference requests through an asyncio lock (one request at a time). It appears to be an internal race condition between MLX's GPU dispatch threads and Metal's command buffer management.

## Environment

- **Hardware:** Mac Studio M2 Ultra (76-core GPU, 128GB unified memory, wired to 115GB via `MLX_MAX_TBP_SIZE_GB`)
- **macOS:** 26.3 (25D125)
- **MLX:** 0.31.0
- **Python:** 3.12.12
- **Model:** Qwen3.5-27B-8bit (MLX quantized format)
- **Serving:** vllm-mlx 0.2.6

## Reproduction

Crashes happen during **long token generation** (thousands of tokens, e.g. generating large JSON output). Short responses (< 100 tokens) work reliably. 37 crashes collected over 2 days (March 13-14) during normal single-request use. The server auto-restarts via LaunchAgent after each crash.

```bash
# Typical invocation (via vllm-mlx serve)
vllm-mlx serve ~/ai-models/mlx_models/Qwen3.5-27B-8bit \
    --host 127.0.0.1 --port 8080 \
    --tool-call-parser qwen3_coder \
    --reasoning-parser qwen3 \
    --enable-auto-tool-choice --mllm --timeout 2400
```

## Crash Details

```
Exception Type:  EXC_BAD_ACCESS (SIGSEGV)
Exception Subtype: KERN_INVALID_ADDRESS at 0x0065c62f691dadd0 -> 0x0000462f691dadd0
                   (possible pointer authentication failure)

Faulting Thread: 21 (of 52 total)
Threads with MLX frames: [1, 2, 3, 4, 5, 6, 7, 8, 9, 21, 51] (11 total)
```

### Faulting Thread Stack
```
0: libobjc.A.dylib :: objc_msgSend
1: libobjc.A.dylib :: -[NSObject isMemberOfClass:]
[...]

Issue #3078, opened 2026-01-29
Request: Support for concurrent inference of independent models from separate threads


1-ashraful-islam:
### Summary

I'm building an application that runs two independent ML models concurrently on Apple Silicon (both macOS and iOS). Each model processes different inputs and produces different outputs - no shared arrays or dependencies between them. Currently this crashes due to thread-safety issues in the Metal backend.

I've researched the existing issues (#2133, #2067, #2086) and PR #2104. I'd like to discuss whether this use case can be supported and what the best path forward might be.

### Environment

- **MLX version:** 0.30.3
- **Platform:** macOS and iOS (Apple Silicon)
- **API:** C++ (direct MLX C++ API)
- **Build:** Built from source

### Use Case

```
Thread A: model_a → inference(input_a) → output_a
Thread B: model_b → inference(input_b) → output_b
(concurrent execution, completely independent)
```

Both models are:
- Loaded once at startup via `mx::import_function()`
- Wrapped with `mx::compile()` for kernel fusion
- Called from their own dedicated threads

### Current Behavior

Concurrent inference produces Metal assertion failures such as:

```
-[_MTLCommandBuffer addScheduledHandler:]: failed assertion
'Scheduled handler provided after commit call'
```

```
A command encoder is already encoding to this command buffer
```

I can provide more detailed stack traces if helpful.

### What I've Tried

**1. Dedicated streams per model using `StreamContext`**

```cpp
// Construction
dedicated_stream_ = mx::new_stream(mx::Device::gpu);

// Inference
mx::StreamContext ctx(dedicated_stream_);
auto outputs = compiled_model_({input_array});
mx::eval(outputs[0]);
mx::synchronize();
```

**Result:** Crashes. Two threads using `StreamContext` concurrently race on the global `default_streams_` map in `Scheduler::set_default_stream()`.

**2. Default stream (no `StreamContext`)**

Both threads use the default GPU stream.

**Result:** Crashes. Both threads race on the shared `DeviceStream::buffer` and `DeviceStream::encoder` fields in `get_command_buffer()` / `commit_command_buffer()`.

**3. Mutex serialization**

This works but negates the benefit of running models in parallel.

### Root Cause Analysis

I traced this to two issues:

**A. `StreamContext` modifies global state**

`set_default_stream()` writes to `Scheduler::default_streams_`, a non-thread-safe `unordered_map`. Concurrent `StreamContext` usage corrupts this or causes incorrect restore on destruction.

[...]

PR #2104, opened 2025-04-22
Metal thread safety

acsweet:
## Proposed changes

These changes are an attempt to improve thread safety for the metal backend. This is related to #2067
Please let me know what you think. 

## Checklist

- [x] I have read the [CONTRIBUTING](https://github.com/ml-explore/mlx/blob/main/CONTRIBUTING.md) document
- [x] I have run `pre-commit run --all-files` to format my code / installed pre-commit prior to committing changes
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have updated the necessary documentation (if needed)

PR #3247, opened 2026-03-12
fix(metal): per-stream locking for concurrent inference thread safety

rsnow:
## Summary

Replace unsynchronized `DeviceStream` mutable state access with fine-grained per-stream locking to eliminate SIGSEGV/SIGABRT crashes during concurrent Metal inference.

**Related issues:** #3216, #2067, #3078

## Problem

The Metal backend's `DeviceStream` holds mutable state (command buffer, encoder, temporaries, fence counters) that is accessed without synchronization during concurrent GPU evaluations. This causes crashes in Metal's command encoder lifecycle, particularly during multi-stream inference with large models on Apple Silicon.

PR #2104 proposed a global mutex but that's too coarse — it serializes all GPU work and risks deadlocks.

## Approach

Three-domain architecture with narrow critical sections:

1. **Host preparation (unlocked)** — graph traversal, kernel lookup, shape specialization. Pure reads, no stream state mutation.
2. **Stream submission (per-stream `op_mtx`)** — encoder commands, buffer rotation, temporary tracking. Short hold, only Metal mutations.
3. **Dependency state (`fence_mtx`)** — fence signaling and waiting. Separate lock to avoid blocking submitters on sync.

Key design elements:
- **`SubmissionEpoch`** — RAII struct wrapping buffer/encoder/temporaries/sequence for atomic rotation
- **`StreamOpLock`** — `[[nodiscard]]` scoped lock with `with_fence_state()` chaining for ordered acquisition
- **`DebugOwner`** — thread-ID based ownership tracking, replacing the previous `try_lock()` assert (which had UB under POSIX and false negatives under contention)
- **Lock ordering enforced by construction** — `op_mtx` via `StreamOpLock`, then `fence_mtx` via `with_fence_state()`
- `FenceImpl::count` and `Event::value_` made `std::atomic` for lock-free fast paths
- `stream_map_` protected by `std::shared_mutex` for concurrent reads

### Unit tests (included in this PR)
6 new tests in `python/tests/test_concurrent_eval.py`:
- Concurrent matmuls on separate streams (4 threads)
- Mixed operation types (matmul, reduction, elementwise, softmax)
- Numerical correctness verification under concurrency
- Sustained pressure (4 threads × 20 iterations)
- Cross-stream data dependencies (shared input, separate consumer streams)
- High concurrency (8 threads, varied workloads)

### Integration testing (external, on M3 Ultra 256GB)
- **355/355** requests across 3 model architectures (Mamba-2 hybrid, dense transformer, GLM-4) at up to 20 concurrent streams — zero crashes, zero restarts
- Ramp test scaling: 3 concurrent → 15s, 6 → 20s, 9 → 24s, 12 → 29s (linear, no cliff)
- 48-request torture test at 12→16→20 concurrent with 4096 tokens each — clean throughout
- oMLX server PID unchanged across all test sessions

## Files changed (6)
- `mlx/backend/metal/device.h` — `SubmissionEpoch`, `StreamOpLock`, `DebugOwner`, per-stream mutexes
- `mlx/backend/metal/device.cpp` — narrow lock scopes in `new_encoder()`, `commit_command_buffer()`, `end_encoding()`
- `mlx/backend/metal/eval.cpp` — `StreamOpLock` acquisition around Metal submission, host prep outside lock
- `mlx/backend/metal/event.cpp` — atomic `value_` access
- `mlx/backend/metal/fence.cpp` — `fence_mtx` for signal/wait, atomic `FenceImpl::count`
- `mlx/event.h` — `std::atomic` for `Event::value_`

## Notes

- `mx.random` has a separate thread-safety issue (global PRNG state, unprotected). Unit tests use deterministic inputs to isolate the Metal stream behavior from that unrelated bug.
- This does not address #3078 (StreamContext / default stream semantics), which is a separate concern.

## Checklist

### DIFF
--- a/python/tests/test_concurrent_eval.py
+++ b/python/tests/test_concurrent_eval.py
@@ -1,0 +1,283 @@
+# Copyright © 2026 Apple Inc.
+
+import threading
+import unittest
+
+import mlx.core as mx
+import mlx_tests
+
+
+class TestConcurrentEval(mlx_tests.MLXTestCase):
+    """Tests for thread safety of concurrent Metal evaluations.
+
+    Validates that multiple threads can safely perform GPU operations
+    on separate streams without crashes or incorrect results.
+
+    Note: mx.random is not used here because the global PRNG state is
+    not thread-safe. These tests use deterministic inputs to isolate
+    the Metal stream concurrency behavior.
+    """
+
+    @unittest.skipIf(not mx.metal.is_available(), "Metal is not available")
+    def test_concurrent_matmul_separate_streams(self):
+        """Multiple threads doing matmuls on separate GPU streams."""
+        errors = []
+        results = [None] * 4
+
+        def worker(idx, size):
+            try:
+                s = mx.new_stream(mx.gpu)
+                with mx.stream(s):
+                    # Deterministic but varied per-thread inputs
+                    a = mx.full((size, size), float(idx + 1) * 0.01)
+                    b = mx.full((size, size), float(idx + 1) * 0.02)
+                    c = a @ b
+                    mx.eval(c)
+                    results[idx] = c.shape
+            except Exception as e:
+                errors.append((idx, e))
+
+        threads = []
+        for i in range(4):
+            t = threading.Thread(target=worker, args=(i, 256))
+            threads.append(t)
+
+        for t in threads:
+            t.start()
+        for t in threads:
+            t.join(timeout=30)
+
+        self.assertEqual(len(errors), 0, f"Thread errors: {errors}")
+        for i in range(4):
+            self.assertEqual(results[i], (256, 256))
+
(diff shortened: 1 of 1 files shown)


zcbenz:
I have put some thoughts about thread safety in https://github.com/ml-explore/mlx/issues/3078#issuecomment-4061668799: basically I think we should not try to achieve thread safety for arrays in different threads, at least not in the first try, for more practical targets (e.g. #3078, which I think is what projects like vllm-mlx or omlx do) we shouldn't need a complex solution like the one in this PR.

rsnow:
Understandable, it is deep and complex.  We've had good luck with it locally and will likely continue to run it, so if this gets revisited, we'd be happy to report results from our continued soak.

PR #952, opened 2024-04-02
Fix GGUF load bug and GRU layer

awni:
Closes #943 

And h/t @davidkoski for the GRU fix

PR #3250, opened 2026-03-13
Fix GRU hidden=None handling for biased new-gate path

### LINKED ISSUE #3249: nn.GRU skips bhn at the first timestep when hidden=None
## Summary

`mlx.nn.GRU` is not equivalent to an explicit zero initial state when called with the default `hidden=None`.

At the first timestep, the implementation skips the hidden-side contribution to the new gate entirely, including the learned `bhn` bias. That makes:

```python
gru(x)
```

behave differently from:

```python
gru(x, hidden=mx.zeros((batch_size, hidden_size)))
```

for the same weights and input.

## Current status

I first noticed this on MLX `0.31.1` and confirmed it is still present on current `main` at:

- `b0564a91123ab79bd3bdbe1a251d3492a316349c`

This appears to be present since `v0.7.0`, the first release that included `nn.GRU`.

## Why this looks wrong

The GRU update in the docstring is:

```text
n_t = tanh(W_xn x_t + b_n + r_t ⊙ (W_hn h_t + b_hn))
```

If `hidden=None` means an implicit zero initial state, then at `t=0` this should reduce to:

```text
n = tanh(W_xn x + b_n + r * b_hn)
```

But the current code skips the hidden-path work when `hidden is None`, so the first step becomes:

```text
n = tanh(W_xn x + b_n)
```

which drops the `r * bhn` term.

## Repro

```python
import numpy as np
import mlx.core as mx
import mlx.nn as nn

np.random.seed(42)
mx.random.seed(42)

D, H = 4, 3
gru = nn.GRU(D, H, bias=True)
x = mx.array(np.random.randn(1, 3, D).astype(np.float32) * 0.1)

y_none = gru(x)
y_zeros = gru(x, hidden=mx.zeros((1, H)))

mx.eval(y_none, y_zeros)

print("hidden=None: ", np.array(y_none)[0, 0])
print("hidden=zeros:", np.array(y_zeros)[0, 0])
print("max diff:    ", np.abs(np.array(y_none) - np.array(y_zeros)).max())
```

Observed on current `main`:

```text
hidden=None:  [ 0.19130191  0.03415391 -0.12855242]
hidden=zeros: [ 0.12526409  0.06978775 -0.23501271]
max diff:     0.10646029
```

With `bias=False`, the mismatch disappears.

## PyTorch comparison

Using the same weights and input, PyTorch matches the explicit-zero MLX path, not the `hidden=None` MLX path.

Observed max diffs on the validated checkout:

```text
max_diff(PyTorch, MLX hidden=zeros) = 5.96e-08
max_diff(PyTorch, MLX hidden=None)  = 1.0646035e-01
```

## Impact

This affects `nn.GRU` when:

- `bias=True`
- `hidden` is omitted or passed as `None`

That is likely the common call pattern. The first hidden state is wrong, and the error then propagates through the rest of the sequence.

## Test gap

The current GRU coverage in `python/tests/test_nn.py` appears to be shape-only and does not check that `gru(x)` matches `gru(x, hidden=zeros)`.

ssmall256:
## Proposed changes

Fix `nn.GRU` so `hidden=None` is treated the same as an explicit zero initial state.

Previously, the `hidden=None` path skipped the hidden-side new-gate contribution at the first timestep, including `bhn`, so `gru(x)` could differ from `gru(x, hidden=zeros)` when `bias=True`.

This change initializes a zero hidden state before the loop and adds a regression test covering batched and unbatched inputs.

Closes #3249

## Tests

```bash
PYTHONPATH=python pytest python/tests/test_nn.py
```

## Checklist

### DIFF
--- a/python/mlx/nn/layers/recurrent.py
+++ b/python/mlx/nn/layers/recurrent.py
@@ -159,6 +159,9 @@ class GRU(Module):
         else:
             x = x @ self.Wx.T
 
+        if hidden is None:
+            hidden = mx.zeros(x.shape[:-2] + (self.hidden_size,), dtype=x.dtype)
+
         x_rz = x[..., : -self.hidden_size]
         x_n = x[..., -self.hidden_size :]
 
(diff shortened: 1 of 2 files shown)


ssmall256:
I checked prior MLX issues and PRs for duplicates before opening this. I did not find an existing report or fix for this specific `hidden=None` / first-step `bhn` behavior. The closest prior GRU-related change I found was #952, but that patch addressed a different GRU bug in the hidden-state update and did not change the `hidden=None` first-timestep path.

angeloskath:
Hi @ssmall256 , thanks for the fix! I hope it's ok but I will merge #3252 instead because it is a bit more concise and efficient. The efficiency comes from avoiding all the computations with 0s when None is passed.

Thanks again for the fix!

Issue #3081, opened 2026-01-29
[BUG] mlx.JVP returns all boolean answers while mlx.VJP returns the correct floats


kyrollosyanny:
**Describe the bug**
I have a function that Im calculating the jacobian of. It works well in mlx.VJP. However, when using JVP, I first get this error `[negative] Not supported for bool, use logical_not instead.`. Then running it again, the error disappears and I get an all boolean output for the jacobian. 

I assume this could be because some operations are not implemented for the forward pass. I am just not sure which ones. 

 - Version [0.30.4]

awni:
Could you share some code which reproduces the issue you are seeing?

kellen-sun:
@kyrollosyanny 
Was there an update on this?

mm65x:
i ran into this too. here's a minimal repro:

```python
import mlx.core as mx

x = mx.array([1.0, -2.0, 3.0])
t = mx.ones_like(x)

# comparison op tangents come back as bool instead of float32
_, tangents = mx.jvp(lambda x: x > 0.0, [x], [t])
print(tangents[0].dtype)  # mlx.core.bool (should be float32)

# this crashes: "[negative] Not supported for bool"
mx.jvp(lambda x: mx.negative(x > 0.0), [x], [t])
```

all six comparison ops (`greater`, `less`, `equal`, `greater_equal`, `less_equal`, `not_equal`) are affected — their JVP implementations return `zeros(shape, bool_, ...)` instead of preserving the input tangent's dtype.

opened a fix in #3253.

zcbenz:
Closed by https://github.com/ml-explore/mlx/pull/3253.

PR #3253, opened 2026-03-14
fix comparison op JVP returning bool tangents instead of input dtype

mm65x:
## Proposed changes

#3081

the six comparison op JVP implementations (`Equal`, `Greater`, `GreaterEqual`,
`Less`, `LessEqual`, `NotEqual`) return `zeros(shape, bool_, stream())`,
producing bool-typed tangents. this causes downstream ops like `negative` to
crash with `[negative] Not supported for bool, use logical_not instead`.

the fix changes them to `zeros(shape, tangents[0].dtype(), stream())` so the
zero tangent preserves the input tangent's dtype. this matches the pattern
used by `LogicalNot::jvp` which already uses `zeros_like(tangents[0])`.

## Checklist


mm65x:
fixed the failing test, was calling an invalid op on the primal path.

PR #3254, opened 2026-03-14
add causal_upper_left mask option to scaled_dot_product_attention

mm65x:
## Proposed changes

#2835

adds `"causal_upper_left"` and `"causal_lower_right"` as explicit mask options
to `mx.fast.scaled_dot_product_attention`. `"causal"` stays as an alias for
`"causal_lower_right"`, so nothing breaks.

when `S_Q != S_KV`, lower-right aligns the last query with the last key,
while upper-left aligns query `i` with keys `0..i` (matching PyTorch's
`is_causal=True`). when `S_Q == S_KV` they're identical.

on Metal, the full-attention kernels already parameterize the causal diagonal
via `qL_off` in `AttnParams`, so the change is just passing `0` instead of
`kL - qL` for upper-left. the vector kernels previously hardcoded the offset
inline, so a `causal_offset` buffer argument was added instead. on CUDA, the
cuDNN path uses `set_causal_mask` vs `set_causal_mask_bottom_right`, and the
vector kernels use a `causal_offset` field in `AttnParams` (same approach as
Metal).

## Checklist

### DIFF
--- a/mlx/backend/cuda/scaled_dot_product_attention.cpp
+++ b/mlx/backend/cuda/scaled_dot_product_attention.cpp
@@ -638,8 +638,7 @@ void ScaledDotProductAttention::eval_gpu(
         output_logsumexp_,
         s);
   } else {
-    sdpa_vector(
-        q, k, v, scale_, out, do_causal_, causal_upper_left_, sinks, s);
+    sdpa_vector(q, k, v, scale_, out, do_causal_, causal_upper_left_, sinks, s);
   }
 }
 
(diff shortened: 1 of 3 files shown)


zcbenz:
Can you fix the lint error?

angeloskath:
I am inclined to close this until there are use cases that need it. Adding options for the sake of having them is at this point just maintenance overhead imho. @zcbenz what do you think?

zcbenz:
Yeah I agree this is not really going to be used in practice, I will also close the original issue.

PR #3258, opened 2026-03-15
support numpy scalar indexing

mm65x:
## Proposed changes

#2710

indexing with numpy integer scalars (`np.int64(1)`, `np.int32(2)`, etc.) throws
"Cannot index mlx array using the given type" because the indexing code only
checks for Python's built-in `int` via `nb::isinstance<nb::int_>`. numpy
integer scalars aren't `int` subclasses but they do implement `__index__`
(PEP 357).

the fix replaces all `nb::isinstance<nb::int_>` checks in the indexing path
with a helper that uses `PyIndex_Check`, which matches any object implementing
`__index__`. this covers direct indexing (`x[np.int64(1)]`), slicing
(`x[np.int64(1):]`), tuple indexing (`x[np.int64(0), np.int64(1)]`), and
assignment through all the same paths.

## Checklist

### DIFF
--- a/python/src/indexing.cpp
+++ b/python/src/indexing.cpp
@@ -8,6 +8,16 @@
 
 #include "mlx/ops.h"
 
+// Matches any object implementing __index__ (PEP 357):
+// Python int, numpy integer scalars, etc.
+bool is_int_like(const nb::object& obj) {
+  return PyIndex_Check(obj.ptr());
+}
+
+nb::int_ to_nb_int(const nb::object& obj) {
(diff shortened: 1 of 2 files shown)


angeloskath:
There is already a PR implementing the exact same fix #3210, feel free to comment there if there is something that isn't supported or can be improved.

It has also been several PRs that already had PRs submitted with the exact same fix (a lot of which I merged because they were small). Please do a search and contribute to the existing PRs before submitting a new one.

PR #3259, opened 2026-03-15
tests: add bfloat16 support to save/load tests

shivansh023023:
## This PR addresses the long-standing TODO in python/tests/test_load.py regarding bfloat16 support.

While MLX has robust bfloat16 support for computation, the serialization tests were previously skipping this dtype to avoid compatibility issues with NumPy. This change "unlocks" bfloat16 validation for the following formats:

Standard NPY (.npy)

Compressed and Uncompressed NPZ (.npz)

Safetensors (.safetensors)

Technical Implementation:

Added a helper method _to_np_dtype to map bfloat16 to np.float32. This allows us to use NumPy as a ground-truth for file structure and shape without crashing on a missing native np.bfloat16 attribute.

Updated test_save_and_load, test_save_and_load_fs, and test_savez_and_loadz to use this mapping.

Note: GGUF support for bfloat16 remains excluded from this PR as the current C++ backend implementation for GGUF throws a RuntimeError for this dtype.

Checklist

### DIFF
--- a/python/tests/test_load.py
+++ b/python/tests/test_load.py
@@ -23,8 +23,13 @@ class TestLoad(mlx_tests.MLXTestCase):
         "int64",
         "float32",
         "float16",
+        "bfloat16",
         "complex64",
     ]
+
+    def _to_np_dtype(self, dt):
+        """Helper to map MLX dtypes to NumPy-compatible dtypes."""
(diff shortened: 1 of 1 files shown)


zcbenz:
Please take a look at the changes before creating a PR.

PR #3262, opened 2026-03-16
Add output_shapes for AddMM

pHequals7:
## Summary

Adds `output_shapes()` to `AddMM`, enabling `compile(shapeless=True)` for models with biased Linear layers.

### Change

`AddMM::output_shapes` returns `inputs[0].shape()` (the C matrix shape), which is already validated to match the output shape at construction in `ops.cpp`.

### Context

Most transformer models use biased Linear layers, which dispatch through AddMM. Without this, `compile(shapeless=True)` throws "primitive does not have shape inference implemented". This follows the same pattern as #2601 (Convolution::output_shapes) and #1727 (shapeless SliceUpdate + Broadcast).

Discovered while porting [mlx-whisper](https://github.com/ml-explore/mlx-examples/tree/main/whisper) to Swift — whisper-small has 145 biased Linear layers that all fail in shapeless compile without this.

### Files Changed

| File | Lines | Change |
|------|-------|--------|
| `mlx/primitives.h` | +1 | Declaration |
| `mlx/primitives.cpp` | +6 | Implementation |

### Previous scope

Originally included Slice and CustomKernel `output_shapes` — dropped per review feedback. CustomKernel shape inference via a callback function could be a separate discussion/PR.

### DIFF
--- a/python/tests/test_compile.py
+++ b/python/tests/test_compile.py
@@ -904,8 +904,7 @@ class TestCompile(mlx_tests.MLXTestCase):
 
         # With alpha and beta
         fun2 = mx.compile(
-            lambda c, a, b: mx.addmm(c, a, b, alpha=2.0, beta=3.0),
-            shapeless=True
+            lambda c, a, b: mx.addmm(c, a, b, alpha=2.0, beta=3.0), shapeless=True
         )
         c = mx.ones((2, 4))
         a = mx.ones((2, 3))


mlx/primitives.cpp
```diff
@@ -345,6 +345,14 @@ bool AddMM::is_equivalent(const Primitive& other) const {
...
 }
 
+std::vector<Shape> AddMM::output_shapes(const std::vector<array>& inputs) {
+  // out = alpha * (A @ B) + beta * C
+  // Output shape matches C (inputs[0]), with last dim from B (inputs[2])
+  auto out_shape = inputs[0].shape();
+  out_shape.back() = inputs[2].shape(-1);
+  return {std::move(out_shape)};
+}
+
```
jagrit06, line 354: Why wouldn't the out shape just be the c shape ? 

From `ops.cpp`
```cpp
  if (c.shape() != out_shape) {
    throw std::invalid_argument(
        "[addmm] input c must broadcast to the output shape");
  }

  auto out = array(
      std::move(out_shape),
      out_type,
      std::make_shared<AddMM>(to_stream(s), alpha, beta),
      {a, b, c});
```

pHequals7: Good catch — you're right, since `c.shape()` is already validated against `out_shape` at construction, we can just return it directly. Simplified in 95382f9.

mlx/backend/metal/custom_kernel.cpp
```diff
@@ -305,8 +305,30 @@ CustomKernelFunction metal_kernel(
...
+    if (has_dynamic) {
+      size_t total = 0;
+      for (const auto& in : inputs) total += in.size();
+      auto per_out = static_cast<ShapeElem>(total / resolved_shapes.size());
+      for (size_t i = 0; i < resolved_shapes.size(); i++) {
+        for (size_t j = 0; j < resolved_shapes[i].size(); j++) {
+          if (resolved_shapes[i][j] == -1) resolved_shapes[i][j] = per_out;
+        }
+      }
+    }
```
angeloskath: I think this is not doing what you were thinking it was doing.

`resolved_shapes.size()` is the number of outputs. `total` is the sum of number of elements of all inputs. What is `per_out` supposed to represent?

Say for instance I write a custom kernel that adds two arrays. `per_out` would be 2 times the input size 🤷‍♂️

The only way to do this properly is to pass a function that computes the output shapes from the input shapes. If this function is passed then shapeless compilation of the custom kernel will be automatically enabled otherwise not.

mlx/primitives.cpp
```diff
@@ -4775,6 +4781,31 @@ bool Slice::is_equivalent(const Primitive& other) const {
       end_indices_ == s_other.end_indices_ && strides_ == s_other.strides_);
 }
 
+std::vector<Shape> Slice::output_shapes(const std::vector<array>& inputs) {
+  // Re-normalize slice bounds against the runtime input shape.
+  // Works for constant-dimension slices; variable-dimension slices
+  // should use take()/DynamicSlice instead.
```
angeloskath:
```suggestion
```

mlx/backend/metal/custom_kernel.cpp
pHequals7: yeah you're right, the -1 sentinel was a hack that only worked for my specific kv cache concat case. a shape inference function passed to metalKernel() makes way more sense as a general api.

stripped this pr down to just AddMM which is the straightforward one. happy to open a separate issue for the CustomKernel shape inference function if that's useful — or leave it for someone with better context on the compile internals.

mlx/primitives.cpp
pHequals7: done, dropped it. makes sense that constant slices don't need this.

pHequals7:
added a test in `python/tests/test_compile.py` (`test_shapeless_compile_addmm`).

writing the test actually caught a bug in the previous implementation: `AddMM` receives inputs as `{a, b, c}` (ops.cpp:5440), not `{c, a, b}`, so `inputs[0].shape()` was returning `a`'s shape, not `c`'s. it worked by coincidence because `c` is validated to match the output shape at construction, so `a`'s shape and `c`'s shape only differ in the last dim, which was never exercised with varying sizes.

fixed to mirror `Matmul::output_shapes`: take `a`'s shape, replace the last dim with `b.shape(-1)`.

the test covers:
- basic `addmm` inside `compile(shapeless=True)`
- a second call with different shapes (same ranks) to verify shapeless reuse
- custom `alpha`/`beta` parameters

PR #3268, opened 2026-03-17
[CUDA] fp and int4 quants for qmm_sm80

zcbenz:
Followup to https://github.com/ml-explore/mlx/pull/3255 finishing the missing quants, was blocked by https://github.com/NVIDIA/cutlass/issues/3111.

We can't support int2/3/5/6 in this version unfortunately since the lengths supported by `cp.async` instruction are limited.

### DIFF
--- a/mlx/backend/cuda/quantized/qmm/qmm.cu
+++ b/mlx/backend/cuda/quantized/qmm/qmm.cu
@@ -103,10 +103,11 @@ void qmm_impl_sm80(
     const array& x,
     const array& w,
     const array& scales,
-    const array& biases,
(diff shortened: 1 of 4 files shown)


zcbenz:
The memory bandwidth of int4 is about 50% of int8 on A100:

| M | N | K | QMM (GiB/s) | CUBLAS (GiB/s) | QMM (TF/s) | CUBLAS (TF/s) | Speedup (x) |
| 16 | 4096 | 4096 | 174.3 | 1397.5 | 9.65 | 22.19 | 0.43 |

I think it means that the kernel is bottlenecked at loading quantized weights.

PR #3279, opened 2026-03-19
Do not create device when not using device

zcbenz:
- Do not call `default_device()` on initialization.
- Do not clear compile cache on exit which initializes device eventually, it is meaningless since it simply clears a `std::map`.

### DIFF
--- a/python/src/device.cpp
+++ b/python/src/device.cpp
@@ -1,8 +1,10 @@
 // Copyright © 2023-2025 Apple Inc.
 
+#include <optional>
 #include <sstream>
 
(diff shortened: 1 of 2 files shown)


angeloskath:
Hm the cache clearing could actually be a problem. The reason is that the C++ compiler cache can outlive the python interpreter and it will keep python data that when destroyed will try to acquire the GIL which is not available since the python interpreter has exited.

zcbenz:
Ah that makes sense, it would be a lot more tricky to ensure thread safety then, I'll revisit this later.

Issue #2086, opened 2025-04-18
[BUG] c++ compile cache should be thread safe


davidkoski:
**Describe the bug**

The c++ compile cache should be thread safe -- whatever mutable state is behind `mlx::core::detail::compile` should be thread safe so it can safely be called from multiple threads (common in Swift, maybe C++).

See also:

- https://github.com/ml-explore/mlx-swift/pull/226
- https://github.com/ml-explore/mlx-swift/issues/225

**To Reproduce**

See:

- https://github.com/ml-explore/mlx-swift/issues/225

Note that the fix for this was to protect the inner function that mutates the arguments (as that is where the crash was observed), during discussion with @awni we think that the cache of graphs also needs locking (or whatever technique is appropriate) as it is also mutable state.

**Expected behavior**

Inner state of mlx::core should be thread safe or documented as not (like MLXArray).

**Desktop (please complete the following information):**
 - OS Version: macOS15
 - Version 0.23.1

awni:
> Inner state of mlx::core should be thread safe

There are still a few places this is not true still. Off the top of my head:
- rng state
- compiler cache
- transform state (e.g. `InTracing`)

ashthedash2k:
I'd be interested in working on this issue, I have knowledge of threading and caches so I might be able to help out? I have not really ever contributed to open source, the code base is pretty large, I was wondering if I could get some pointers / a starting direction! Thanks so much.

zcbenz:
Closing since compile cache is thread safe after 0.31.2.

PR #3280, opened 2026-03-19
Use thread local storage for frontend compile cache

zcbenz:
Refs #2086, #3078, #3216.

The frontend compile cache uses function ID as key so it is very unlikely functions in different threads would shared cache, and making it thread local should be enough to ensure thread safety.

For backend compile cache, i.e. `Compiled::eval_cpu`/`eval_gpu`, it is reasonable to share cache between threads and they are already guarded with locks (except for the CUDA backend which I will fix in followup PRs).

### DIFF
--- a/mlx/compile.cpp
+++ b/mlx/compile.cpp
@@ -1,4 +1,6 @@
 // Copyright © 2023-2024 Apple Inc.
+
+#include <atomic>
 #include <cstdlib>
 #include <map>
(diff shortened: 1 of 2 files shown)


zcbenz:
That would already be a problem since changing the default stream would invalidate the compile cache. In a multi-thread environment we expect users to use different streams as default streams so hopefully it would not be a problem.

PR #3289, opened 2026-03-21
[Metal] Fix depthwise conv 1D kernel name for large variant

Brooooooklyn:
The kernel name was constructed with the wrong argument order, producing `depthwise_conv_1d__largefloat16` instead of `depthwise_conv_1d_float16_large` which matches the Metal kernel registration.

### DIFF
--- a/mlx/backend/metal/conv.cpp
+++ b/mlx/backend/metal/conv.cpp
@@ -1041,8 +1041,8 @@ void depthwise_conv_1D_gpu(
   concatenate(
       base_name,
       "depthwise_conv_1d_",
-      large ? "_large" : "",
-      type_to_name(out));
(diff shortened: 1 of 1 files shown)


Brooooooklyn:
What I got:
```
❯ node index.js

[MLX] Exception in async_eval: [metal::Device] Unable to load kernel depthwise_conv_1d__largebfloat16

[MLX] Exception in mlx_qwen35_moe_eval_token_and_caches: [metal::Device] Unable to load kernel depthwise_conv_1d__largebfloat16

[MLX] Exception in array_eval: [metal::Device] Unable to load kernel depthwise_conv_1d__largebfloat16

[1] 32912 segmentation fault  node index.js
```

PR #3290, opened 2026-03-21
Fix stale transform copy-chain leaks

Brooooooklyn:
`vjp()` and `jvp()` wrap each primal in `copy(p, s)` to create tracers. When user code stores a tracer into an external container and feeds it back as a primal in the next call, each iteration nests another Copy:

```
  call 1: container[0] = copy(original)
  call 2: container[0] = copy(copy(original))
  call N: container[0] = copy^N(original)
```

The container keeps the head alive, which transitively keeps every intermediate Copy node alive — linear memory growth per call.

Fix: before creating the new tracer copy, `unwrap_stale_copy_wrappers()` peels off Copy nodes that have `is_tracer()=false`, collapsing the chain to depth 1. Active tracers (`is_tracer()=true`, from nested transforms) are never unwrapped, preserving nested transform semantics. Copy's VJP is identity so flattening is gradient-safe.

- Close https://github.com/ml-explore/mlx/issues/2841

### DIFF
--- a/mlx/transforms.cpp
+++ b/mlx/transforms.cpp
@@ -24,6 +24,28 @@ namespace mlx::core {
 
 static constexpr int MAX_ACTIVE_TASKS = 10;
 
+namespace {
+
+// Create a tracer copy of a primal for use in vjp/jvp. If the primal is a
+// stale Copy from a previous transform call (not an active tracer), peel it
+// off to prevent copy-chain accumulation when containers feed tracers back.
+array make_tracer(const array& p) {
+  auto s = p.has_primitive() ? p.primitive().stream()
+                             : default_stream(default_device());
+  auto source = p;
+  if (!p.is_tracer() && p.has_primitive() && !p.inputs().empty()) {
+    auto& prim = p.primitive();
+    if (typeid(prim) == typeid(Copy)) {
+      source = p.inputs()[0];
+    }
+  }
+  auto out = copy(source, s);
+  out.set_tracer(true);
+  return out;
+}
+
+} // namespace
+
 /* This class is only meant to be used in eval
  * for synchronizing with the main thread. */
 class Synchronizer : public Primitive {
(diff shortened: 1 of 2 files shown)


mlx/transforms.cpp
```diff
@@ -24,6 +24,26 @@ namespace mlx::core {
...
 
+namespace {
+
+bool is_stale_copy_wrapper(const array& x) {
+  if (!x.has_primitive() || x.inputs().empty() || x.is_tracer()) {
+    return false;
+  }
+
+  auto& p = x.primitive();
+  return typeid(p) == typeid(Copy);
```
zcbenz: Wouldn't this also return true for intended copies from users?

mlx/transforms.cpp
```diff
@@ -24,6 +24,26 @@ namespace mlx::core {
...
+}
+
+array unwrap_stale_copy_wrappers(array x) {
+  while (is_stale_copy_wrapper(x)) {
```
zcbenz: I think we don't need to check recursively if we are not creating chains of `copy`s?

Brooooooklyn: You are right, I've replaced the two helper functions + while loop with a single inline if that peels exactly one `Copy`

mlx/transforms.cpp
```diff
@@ -337,7 +337,16 @@ std::pair<std::vector<array>, std::vector<array>> vjp(
...
+    // Peel off one stale Copy wrapper to prevent copy-chain accumulation
+    // when a container feeds a previous tracer back as a primal.
+    auto source = p;
+    if (!p.is_tracer() && p.has_primitive() && !p.inputs().empty()) {
+      auto& prim = p.primitive();
+      if (typeid(prim) == typeid(Copy)) {
+        source = p.inputs()[0];
+      }
+    }
+    primals_.push_back(copy(source, s)); // Does not do a deep copy
```
zcbenz: Can you move the code into a helper?

```c++
  for (auto& p : primals) {
    primals_.push_back(make_tracer(p));
  }
```

Brooooooklyn: Done

PR #3292, opened 2026-03-21
Fix vmap + floor_divide: preserve integer dtype

robert-johansson:
## Summary

`Divide::vmap` calls `divide(a, b)` which promotes integers to float via `at_least_float`. But `floor_divide` creates a `Divide` primitive with integer dtype for integer division. Under vmap, the integer dtype is lost.

**Before:**
```python
import mlx.core as mx
f = lambda s: mx.floor_divide(s, mx.array(5, mx.int32))
vf = mx.vmap(f)
result = vf(mx.arange(25, dtype=mx.int32))
# result.dtype == float32, values are [0.0, 0.2, 0.4, ...] (wrong)
```

**After:**
```python
result = vf(mx.arange(25, dtype=mx.int32))
# result.dtype == int32, values are [0, 0, 0, 0, 0, 1, 1, ...] (correct)
```

## Fix

In `Divide::vmap`, check if inputs are integer type and construct the `Divide` primitive directly (preserving dtype) instead of calling the high-level `divide()` which promotes to float.

## Test plan

- [x] New test case `test vmap floor_divide integer` with 3 sub-tests:
  - `floor_divide` preserves int32 dtype under vmap
  - `remainder` preserves int32 dtype under vmap  
  - `floor_divide + remainder` reconstructs original values under vmap
- [x] All existing vmap tests pass (38 assertions)

### DIFF
--- a/mlx/primitives.cpp
+++ b/mlx/primitives.cpp
@@ -1789,10 +1789,6 @@ std::pair<std::vector<array>, std::vector<int>> Divide::vmap(
     const std::vector<array>& inputs,
     const std::vector<int>& axes) {
   auto [a, b, to_ax] = vmap_binary_op(inputs, axes, stream());
-  // The high-level divide() promotes integers to float via at_least_float.
-  // But floor_divide() creates a Divide primitive with integer dtype
-  // for integer division. Preserve that: if inputs are integer, keep
-  // integer semantics by constructing the primitive directly.
   auto out = issubdtype(a.dtype(), integer) ? floor_divide(a, b, stream())
                                             : divide(a, b, stream());
   return {{out}, {to_ax}};
--- a/mlx/primitives.cpp
+++ b/mlx/primitives.cpp
@@ -1793,19 +1793,9 @@ std::pair<std::vector<array>, std::vector<int>> Divide::vmap(
   // But floor_divide() creates a Divide primitive with integer dtype
   // for integer division. Preserve that: if inputs are integer, keep
   // integer semantics by constructing the primitive directly.
-  auto dtype = promote_types(a.dtype(), b.dtype());
-  if (issubdtype(dtype, integer)) {
-    auto bcast = broadcast_arrays(
-        {astype(a, dtype, stream()), astype(b, dtype, stream())}, stream());
-    return {
-        {array(
-            bcast[0].shape(),
-            dtype,
-            std::make_shared<Divide>(stream()),
-            std::move(bcast))},
-        {to_ax}};
-  }
-  return {{divide(a, b, stream())}, {to_ax}};
+  auto out = issubdtype(a.dtype(), integer) ? floor_divide(a, b, stream())
+                                            : divide(a, b, stream());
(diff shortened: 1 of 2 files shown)


zcbenz (changes requested):
I think this would make vmap + divide output wrong dtype for integers? Adding a flag in Divide might be a better solution?

mlx/primitives.cpp
```diff
@@ -1789,6 +1789,22 @@ std::pair<std::vector<array>, std::vector<int>> Divide::vmap(
...
+        {astype(a, dtype, stream()), astype(b, dtype, stream())}, stream());
+    return {
+        {array(
+            bcast[0].shape(),
+            dtype,
+            std::make_shared<Divide>(stream()),
+            std::move(bcast))},
+        {to_ax}};
+  }
   return {{divide(a, b, stream())}, {to_ax}};
```
angeloskath: You can simplify this either via

```suggestion
  auto out = issubdtype(a.dtype(), integer) ? floor_divide(a, b, stream())
                                            : divide(a, b, stream());
  return {{out}, {to_ax}};
```

or via

```suggestion
auto arrs = broadcast_arrays({a, b}, stream());
return {{array(
    arrs[0].shape(),
    arrs[0].dtype(),
    std::make_shared<Divide>(stream()),
    std::move(arrs))}, {to_ax}};
```

angeloskath (changes requested):

mlx/primitives.cpp
zcbenz: @angeloskath I think this approach would make divide(int, int) output int instead of float after vmap?

angeloskath: Yeah this can be confusing but if we walk through vmapping then you see why it doesn't.

Say we have the following code
```python
vdivide = mx.vmap(mx.divide, in_axes=(1, None))
x = mx.arange(100).reshape(50, 2)
y = (mx.arange(50) + 1)
z = vdivide(x, y)
```

what happens is:
1. first we create tracer inputs without the vectorized axes
2. we call the function and get the graph
3. we vectorize the graph by forward propagating the vectorized axes

Step 1 results in two inputs of shape `(50,)`.

Step 2 is where the "magic" happens. The graph of divide is not simply the `Divide` primitive but it is something like

```
A <- AsType(A, float)
B <- AsType(B, float)
A, B <- Broadcast(A, B)
C <- Divide(A, B)
```

so the vmap will always work on arrays that are the correct type and excluding the vectorized axis also broadcasted.

But I am fine going with option 1 if that is more intuitive.

robert-johansson: Thanks @zcbenz and @angeloskath for your input. Do you need any revisions made by me to be submitted?

angeloskath (approved):

mlx/primitives.cpp
zcbenz: Thanks for the explanations! It totally makes sense to me now.

zcbenz (approved):

Issue #3196, opened 2026-03-03
NA/M5 addmm / matmul on MPS ~10–20% slower than PyTorch for 1280×1280 BF16 (hurts Nano Chat training)


Anemll:
Summary
On an Mac M5, mlx.addmm and A @ B are consistently ~1.1–1.2× slower than PyTorch’s addmm / matmul on MPS for 1280×1280 BF16 matrices. This GEMM shape is representative of Nano Chat–style transformer training, so the gap directly reduces end‑to‑end training throughput vs PyTorch.

It sped  up after NA support but its still behind since official Neural Accelerator Support

Repro
Gist with script and logs: https://gist.github.com/Anemll/5420800c3d29c7fae18a2b9b10907b14
Key details:
Shape: (1280, 1280)
Dtype: bfloat16
Device: PyTorch mps, MLX Metal
Warmup: 30, Iterations: 1000
Ops tested:
addmm: beta * C + alpha * (A @ B)
matmul: A @ B
Sync:
PyTorch: torch.mps.synchronize()
MLX: mx.eval(result)
Results (MLX 0.31.0)
PyTorch MPS:
addmm: ~0.596 ms
matmul: ~0.549 ms
MLX:
addmm: ~0.651 ms
matmul: ~0.646 ms
Ratios:
addmm: 1.09× slower (MLX / PyTorch)
matmul: 1.18× slower
Motivation (Nano Chat)
Small LLM (“Nano Chat”) training is GEMM‑bound at these sizes; this 10–20% gap in BF16 GEMM on M‑series Macs translates almost directly into slower training steps vs PyTorch MPS. Closing this gap would make MLX more competitive as the default backend for local Nano Chat training.

jagrit06:
I will get to addressing this issue 
In the meanwhile, just so I also have an idea of the real world use case associated, do the tests you're doing with NanoChat also require the added `C` matrix to be full `1280x1280` ?

Anemll:
Yes, the real-world use case requires full 1280×1280 C matrices (not bias vectors).

**NanoChat model uses `model_dim = 1280`** (20 layers × 64 aspect ratio). The Muon optimizer's Newton-Schulz iteration operates directly on weight matrices:

| Weight Matrix | Shape | NS iteration operates on |
|---|---|---|
| Attention Q/K/V/O | (1280, 1280) | 1280×1280 |
| MLP up-projection | (1280, 5120) | Transposed → 1280×1280 |
[...]

PR #3295, opened 2026-03-22
Extend regular NAX tuning to gen-17 g devices

lentil32:
## Proposed changes

- Extend the shared regular NAX GEMM selector in `mlx/backend/metal/matmul.cpp` while preserving the existing tuned `s/c/d` path.
- Apply the same tuned `64x128x(64|256), wm=2, wn=4, swizzle=2` route to M5-generation `g` devices for `float16` / `bfloat16`, which is the path relevant to #3196.
- Keep tile and swizzle selection in a single helper.

This may address #3196 on M5 devices that route through the `g` regular NAX path.

## Benchmarks

Measured manually on an Apple M5 Pro with the issue-shaped BF16 `1280x1280` `addmm` / `matmul` microbenchmark (`30` warmup iterations, `1000` timed iterations), using explicit BF16 flags.

This machine reports `architecture = applegpu_g17s`, so the real-device path is effectively unchanged because it already takes the existing tuned route:

- `addmm`: `0.3073 ms -> 0.3059 ms` (`-0.5%`)
- `matmul`: `0.3108 ms -> 0.3099 ms` (`-0.3%`)

To validate the new `g` route directly, I reran the same workload with `MLX_METAL_GPU_ARCH=applegpu_g17g`:

- `addmm`: `0.3620 ms -> 0.3132 ms` (`-13.5%`, `1.16x faster`)
- `matmul`: `0.3495 ms -> 0.3208 ms` (`-8.2%`, `1.09x faster`)

These results suggest the change improves the targeted `g` path, but I have not yet validated it on a real device that reports `applegpu_g17g`.

## Checklist

### DIFF
--- a/mlx/backend/metal/matmul.cpp
+++ b/mlx/backend/metal/matmul.cpp
@@ -79,6 +79,67 @@ ensure_batch_contiguous(const array& x, metal::Device& d, const Stream& s) {
   return std::make_tuple(false, x_copy.strides()[x_copy.ndim() - 2], x_copy);
 }
 
+struct RegularNaxGemmTuning {
+  int bm;
+  int bn;
+  int bk;
+  int wm;
+  int wn;
+  int swizzle_log;
+};
+
+inline bool use_regular_nax_tuning(
+    const metal::Device& d,
+    const Dtype& out_dtype) {
+  const auto& arch = d.get_architecture();
+  if (arch.empty()) {
+    return false;
+  }
+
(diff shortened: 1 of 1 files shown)


angeloskath:
Changing the architecture via `MLX_METAL_GPU_ARCH=applegpu_g17g` may change the code that the machine runs but it can't be used to check the performance on that type of architecture. You need the actual machine to be able to tune a kernel for that machine.

PR #3296, opened 2026-03-22
add nn.WeightNorm layer

mm65x:
## Proposed changes

#1888

adds `nn.WeightNorm`, a module wrapper that applies weight normalization to a
parameter of a given module. reparameterizes a weight `w` into a magnitude
`weight_g` and direction `weight_v` such that `w = g * v / ||v||`, recomputed
on each forward pass.

implemented as a pure `nn.Module` layer with no C++ or free functions, as
suggested in #1921. the wrapped module's original weight is frozen so only
`weight_g` and `weight_v` are trainable, using the same freeze/unfreeze pattern
as `BatchNorm`'s running stats.

works with any module that has a weight parameter (Linear, Conv1d, Conv2d, etc.).

## Checklist

### DIFF
--- a/docs/src/python/nn/layers.rst
+++ b/docs/src/python/nn/layers.rst
@@ -72,3 +72,4 @@ Layers
    Tanh
    Transformer
    Upsample
+   WeightNorm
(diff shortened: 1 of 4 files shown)


zcbenz:
Closing since this has been draft for a long time, and the code is not production ready.

PR #3310, opened 2026-03-24
Fix moved-from shape bug in broadcast_arrays causing vmap bus error

### LINKED ISSUE #3309: Bus error in vmap + take_along_axis with in_axes=(None, 0) on higher-rank index
Reproducer (verified on clean upstream/main, no local changes):

```python
import mlx.core as mx
a = mx.arange(4*5*3).reshape(4,5,3)
idx = mx.zeros((2,2,1,3), dtype=mx.int32)
out = mx.vmap(lambda x,y: mx.take_along_axis(x,y,axis=0), in_axes=(None,0))(a, idx)
```

Result: Bus error (core dumped)

Environment:
- MLX: upstream/main @ e40ada3fe (clean worktree)
- macOS 26.3.1 (25D2128)
- Hardware: MacBook Pro, Apple M3 Max, 128 GB RAM

Aristide021:
Fixes a moved-from shape bug in `broadcast_arrays` that caused a bus 
error when using `vmap` with `take_along_axis` and `in_axes=(None, 0)` 
on higher-rank index tensors.

### Root cause
In `mlx/ops.cpp`, `std::move(out_shape)` was passed to the output array 
constructor before `out_shape` was used to construct the `Broadcast` 
primitive. Because argument evaluation order is unspecified in C++, the 
primitive could be constructed with moved-from `out_shape` (observed as 
`shape_ = {}`), causing undefined behavior under vmap.

### Fix
Copy `out_shape` instead of moving it so both the output array and the 
`Broadcast` primitive receive the correct shape.

### Regression tests
- C++: `test vmap take_along_axis with unmapped input and mapped index`
- Python: extended `test_vmap_take_along_axis` with the higher-rank 
  `in_axes=(None, 0)` case

### Reproducer (verified on clean upstream/main before fix)
```python
import mlx.core as mx
a = mx.arange(4*5*3).reshape(4,5,3)
idx = mx.zeros((2,2,1,3), dtype=mx.int32)
out = mx.vmap(lambda x,y: mx.take_along_axis(x,y,axis=0), in_axes=(None,0))(a, idx)
# Previously: Bus error (core dumped)
# After fix: out.shape == (2, 2, 5, 3)
```

Closes #3309.

### DIFF
--- a/mlx/ops.cpp
+++ b/mlx/ops.cpp
@@ -1659,7 +1659,7 @@ std::vector<array> broadcast_arrays(
         outputs.push_back(in);
       } else {
         outputs.push_back(array(
-            std::move(out_shape),
+            out_shape,
             in.dtype(),
             std::make_shared<Broadcast>(to_stream(s), out_shape),
             {in}));
--- a/python/tests/test_vmap.py
+++ b/python/tests/test_vmap.py
@@ -595,6 +595,12 @@ class TestVmap(mlx_tests.MLXTestCase):
         out = mx.vmap(fun, in_axes=(None, 0))(a, idx)
         self.assertEqual(out.shape, (4, 2, 1))
 
+        a = mx.zeros((4, 5, 3))
+        idx = mx.zeros((2, 2, 1, 3), mx.int32)
+
+        out = mx.vmap(fun, in_axes=(None, 0))(a, idx)
+        self.assertEqual(out.shape, (2, 2, 5, 3))
+
     def test_vmap_put_along_axis(self):
         a = mx.zeros((4, 5, 1))
         idx = mx.ones((2, 4, 1), mx.int32)
--- a/tests/vmap_tests.cpp
+++ b/tests/vmap_tests.cpp
@@ -392,6 +392,18 @@ TEST_CASE("test vmap gather") {
   }
 }
 
+TEST_CASE("test vmap take_along_axis with unmapped input and mapped index") {
+  auto fun = [](std::vector<array> inputs) {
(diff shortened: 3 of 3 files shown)


hnshah:
## Validation on M3 Ultra 256GB

I've validated the vmap + take_along_axis fix on M3 Ultra:

**Test Hardware:**
- Mac Studio M3 Ultra (256GB)
- macOS 25.3.0 (Darwin 25.3.0)
- MLX: `0.31.2.dev20260324+63b73e7a` (from your `vmap-take-along-axis-crash` branch)

**Reproducer Test (Issue #3309):**

```python
import mlx.core as mx
a = mx.arange(4*5*3).reshape(4,5,3)
idx = mx.zeros((2,2,1,3), dtype=mx.int32)
out = mx.vmap(lambda x,y: mx.take_along_axis(x,y,axis=0), in_axes=(None,0))(a, idx)
```

**Before PR #3310:**
```
Command aborted by signal SIGBUS
```

**After PR #3310:**
```
✅ No crash! Output shape: (2, 2, 5, 3)
```

**Additional Tests:**

| Test Case | Result |
|-----------|--------|
| Original reproducer (in_axes=(None,0)) | ✅ Pass |
| Both inputs mapped (in_axes=(0,0)) | ✅ Pass |

**Key Finding:**
> The bus error is completely resolved. The moved-from shape bug in `broadcast_arrays` is fixed - both the output array and the `Broadcast` primitive now receive the correct shape.

Successfully fixes Issue #3309! 🎯

PR #3286, opened 2026-03-20
[CUDA] support sorting complex numbers

Lyxot:
## Proposed changes

Add complex64 sort, argsort, partition, and argpartition support to the CUDA backend

## Summary

Remove the `complex64_t` exclusion from the CUDA merge sort kernels. The current merge sort is comparison-based and already supports complex64. The guard was stale.

## Checklist

Put an `x` in the boxes that apply.

- [x] I have read the [CONTRIBUTING](https://github.com/ml-explore/mlx/blob/main/CONTRIBUTING.md) document
- [x] I have run `pre-commit run --all-files` to format my code / installed pre-commit prior to committing changes
- [x] I have added tests that prove my fix is effective or that my feature works

PR #3314, opened 2026-03-25
[Metal] Support sorting complex numbers

Lyxot:
## Proposed changes

Followup to #3286. 

Add complex64 sort, argsort, partition, and argpartition support to the Metal backend

### DIFF
--- a/mlx/backend/metal/kernels/complex.h
+++ b/mlx/backend/metal/kernels/complex.h
@@ -97,10 +97,6 @@ constexpr bool operator<(complex64_t a, complex64_t b) {
   return operator>(b, a);
 }
 
-constexpr bool isnan(complex64_t x) {
-  return isnan(x.real) || isnan(x.imag);
-}
-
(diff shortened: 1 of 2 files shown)


mlx/backend/metal/kernels/complex.h
```diff
@@ -97,6 +97,10 @@ constexpr bool operator<(complex64_t a, complex64_t b) {
   return operator>(b, a);
 }
 
+constexpr bool isnan(complex64_t x) {
```
zcbenz: It seems that the `isnan` in global namespace is only used by `metal/kernels/sort.h`? I think the code would be cleaner by fixing the callee instead:

```c++
bool an = metal::is_same_v<T, complex64_t> ? (metal::isnan(a.real) || metal::isnan(a.imag)) : metal::isnan(a);
```

Otherwise we would have multiple versions of `isnan` that some work for complex while some do not.

Lyxot: fixed

PR #3321, opened 2026-03-25
[CUDA] Add GatherQMM for quantized gather matmul

Lyxot:
## Proposed Changes

Adds CUDA `GatherQMM` implementation, enabling quantized MoE inference on GPU. Also extracts `qmv_impl` device function from `qmv_kernel`, so both `qmv_kernel` and `gather_qmv_kernel` share the same dequant+FMA+reduce compute body.

## What's supported

- `transpose=true`
- All quantization modes: affine (2/3/4/5/6/8-bit), mxfp4, mxfp8, nvfp4
- All float types: float32, float16, bfloat16
- Arbitrary batch dimensions and multi-dimensional gather indices

## What's next

- ~~Dedicated `gather_fp_qmv` to inherit architecture-specific FP-mode optimizations from `fp_qmv`~~
- Matrix-matrix path for large M using CUTLASS-based `gather_qmm` kernel
- Sorted-rhs batching with grouped GEMM optimization for sorted expert indices

## Performance

4-bit affine quantization, M=1:

| N | K | Experts | Batch | Loop (µs) | GatherQMM (µs) | Speedup |
|---:|---:|---:|---:|---:|---:|---:|
| 4096 | 4096 | 8 | 2 | 233 | 91 | 2.6x |
| 4096 | 4096 | 8 | 8 | 894 | 252 | 3.5x |
| 4096 | 4096 | 8 | 32 | 3559 | 1000 | 3.6x |
| 11008 | 4096 | 8 | 32 | 8012 | 2642 | 3.0x |

Loop baseline: individual `quantized_matmul` calls per expert, stacked.


mlx/backend/cuda/quantized/quantized.cpp
```diff
@@ -104,6 +104,95 @@ void QuantizedMatmul::eval_gpu(const std::vector<array>& inputs, array& out) {
...
+        mode_,
+        encoder.device());
+  };
+  bool can_use_fp_qmv = supports(supports_fp_qmv);
+  bool can_use_qmv = supports(supports_qmv) || can_use_fp_qmv;
+
+  auto call_qmv = [&]() {
+    out.set_data(cu::malloc_async(out.nbytes(), encoder));
+    if (can_use_fp_qmv) {
+      // TODO: Add gather_fp_qmv for FP-mode-specific optimizations.
```
zcbenz: We plan to deprecate the `fp_qmv` kernel, see also https://github.com/ml-explore/mlx/pull/3239#issuecomment-4037002990.

mlx/backend/cuda/quantized/qmm/qmv.cu
```diff
@@ -274,6 +261,165 @@ __global__ void qmv_kernel(
...
+    const uint32_t* w,
+    const S* scales,
+    const T* biases,
+    T* out,
+    const uint32_t* lhs_indices,
+    const uint32_t* rhs_indices,
+    int n,
+    int k,
+    int x_batch_ndims,
+    const __grid_constant__ Shape x_batch_shape,
```
zcbenz: We do not need to support non-contiguous x/w/scales, at least not in the initial implementation. This kernel is optimized for LLM inference.

mlx/backend/cuda/quantized/qmm/qmv.cu
```diff
@@ -182,38 +182,23 @@ dequant_fma(const T* x, const Q* w, S scale, T bias, float* out) {
...
-    int rows_per_block,
     int elems_per_thread,
     int group_size,
     bool has_bias,
     bool has_residue_k,
     typename T,
     typename Q,
     typename S>
-__global__ void qmv_kernel(
+__device__ __forceinline__ void qmv_impl(
```
zcbenz:
```suggestion
__device__ __forceinline__ void qmv_kernel_impl(
```

mlx/backend/cuda/quantized/quantized.cpp
```diff
@@ -104,6 +104,78 @@ void QuantizedMatmul::eval_gpu(const std::vector<array>& inputs, array& out) {
...
+        scales,
+        biases,
+        out,
+        transpose_,
+        bits_,
+        group_size_,
+        mode_,
+        encoder.device());
+  };
+  bool can_use_qmv = supports(supports_qmv) || supports(supports_fp_qmv);
```
zcbenz: Should not check `supports_fp_qmv`.

mlx/backend/cuda/quantized/qmm/qmv.cu
```diff
@@ -274,6 +258,114 @@ __global__ void qmv_kernel(
...
+  x += block.group_index().y * k + m * k * l;
+  out += block.group_index().y * n + m * n * l;
+
+  // Advance w/scales/biases for batch dimension.
+  constexpr int bits = cute::sizeof_bits_v<Q>;
+  auto w_step = [&](int idx) { return idx * cuda::std::min(8, bits) / 8; };
+  int groups_per_row = k / group_size;
+  int w_batch = broadcast_w ? 0 : l;
+  w += static_cast<int64_t>(n) * w_batch * w_step(k);
+  scales += static_cast<int64_t>(n) * w_batch * groups_per_row;
```
zcbenz: `w_batch` should be passed as an arg to `_impl` so we don't have to redefine `w_step` / `groups_per_row`.

mlx/backend/cuda/quantized/qmm/qmv.cu
```diff
@@ -274,6 +258,114 @@ __global__ void qmv_kernel(
...
+    return;
+  }
+
+  // Gather: look up batch indices.
+  uint32_t batch_idx = block.group_index().z;
+  uint32_t x_idx = lhs_indices[batch_idx];
+  uint32_t w_idx = rhs_indices[batch_idx];
+
+  // Offset pointers using gathered indices.
+  x += x_idx * x_batch_stride;
```
zcbenz: Code would be easier to maintain to just follow what `qmv_kernel` does:

```c++
x += block.group_index().y * k + m * k * x_idx;
```

i.e. just calculate the batch stride on the fly rather than passing as a parameter.

`gather_qmv_kernel` and `qmv_kernel` should only differ little.

mlx/backend/cuda/quantized/qmm/qmv.cu
```diff
@@ -433,4 +525,106 @@ void qmv(
...
+    array& out,
+    int bits,
+    int group_size,
+    QuantizationMode mode,
+    cu::CommandEncoder& encoder) {
+  const char* tag = "[gather_qmm]";
+  int m = out.shape(-2);
+  int n = out.shape(-1);
+  int k = x.shape(-1);
+  int B = out.size() / (m * n);
```
zcbenz: We use `l` instead of `B` in this file to follow CUTLASS's naming style, which makes the kernel code easier to work with since it uses CUTLASS utils.

mlx/backend/cuda/quantized/qmm/qmv.cu
```diff
@@ -433,4 +525,106 @@ void qmv(
...
+          encoder.set_input_array(rhs_indices);
+          encoder.set_output_array(out);
+
+          constexpr bool has_bias = !cutlass::has_negative_zero_v<Q>;
+          constexpr int rows_per_block = 8;
+          constexpr int elems_per_thread =
+              (cute::sizeof_bits_v<T> <= 16 && cute::sizeof_bits_v<Q> <= 4) ? 16
+                                                                            : 8;
+
+          dim3 num_blocks{
```
zcbenz: Can you move this part of code to `cu::gather_qmv` like what the `cu::qmv` does? It is a bit redundant code but the main purpose is to make the kernel run-able in an independent file which helps a lot when profiling the kernel.

PR #3311, opened 2026-03-24
Fix Metal shader library memory leak in compiled function cache

hnshah:
## Problem

When compiled functions are removed from `CompilerCache` (via `compile_erase()` or `compile_clear_cache()`), the associated Metal shader libraries and pipeline states in `Device::library_kernels_` are never released. They persist until process exit, causing memory accumulation in long-running processes.

**Root cause:** `Device::clear_library()` exists but was never called from the compile cache cleanup path.

## Solution

Added `cleanup_metal_libraries()` helper that:
1. Iterates through cache entries before erasure
2. Extracts library names from `Compiled` primitives in the tape
3. Calls `Device::clear_library()` for each unique library

**Changes:**
- Added Metal device header include (guarded by `#ifdef METAL_AVAILABLE`)
- Added `cleanup_metal_libraries()` private method to `CompilerCache`
- Modified `erase()` to clean up Metal resources before erasing cache entry
- Modified `clear()` to clean up all Metal resources before clearing cache

## Testing

- Tested on Mac Studio M3 Ultra (256GB), macOS 25.3.0
- Build verification: compiles cleanly with no errors
- Code inspection: follows same pattern as `custom_kernel.cpp:371`

## Impact

- Fixes memory leaks in long-running MLX processes that use compiled functions
- No performance impact (cleanup only occurs during explicit cache erasure)
- Particularly benefits serving workloads, training loops, and interactive sessions

Fixes #3297

PR #3322, opened 2026-03-25
Fix Metal library ownership when clearing compiled function cache

### LINKED ISSUE #3297: Metal shader pipeline states not released when compiled functions are erased
When a compiled function is removed from `CompilerCache` (via `compile_erase` or `compile_clear_cache`), the associated Metal shader libraries and pipeline states in `Device::library_kernels_` are not released. They persist until process exit.

`Device::clear_library()` exists but is never called from the compile cache cleanup path.

Found while investigating Metal resource accumulation in a long-running process that creates and discards compiled functions over time.

Aristide021:
Fixes #3297.

Compiled-function cache erasure removed `CompilerCache` entries but did not release the corresponding Metal shader libraries and pipeline states. The missing piece is that `CompilerCache` is thread-local while the Metal device library cache is global, so cleanup needs explicit ownership tracking.

An earlier fix attempt (#3311) was closed because it inferred Metal ownership from generic `Compiled` nodes without accounting for the backend-generic nature of `Compiled` or the fact that `metal::device()` does not dispatch by backend.

This change fixes the ownership model directly:

- capture Metal library names when fused `Compiled` primitives are created in `compile_fuse()`
- retain/release those library names through a process-wide refcount registry
- only clear a Metal library when the last live GPU-backed cache entry owning it is erased
- keep the generic compiler/Metal interaction behind a narrow retain/release boundary
- never call into the Metal device for CPU-backed cache entries

Tests added:

- `compile_erase()` rebuilds a cache entry correctly
- `compile_clear_cache()` rebuilds entries correctly
- shared fused-library ownership: erasing one owner does not break the surviving owner, and erasing the last owner does not break future recompilation

Verification:

- `cmake --build build --target tests -j8`
- `./build/tests/tests`

### DIFF
--- a/mlx/compile.cpp
+++ b/mlx/compile.cpp
@@ -3,6 +3,7 @@
 #include <atomic>
 #include <cstdlib>
 #include <map>
+#include <mutex>
 #include <sstream>
 #include <unordered_map>
 #include <unordered_set>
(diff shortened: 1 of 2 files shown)


zcbenz:
The code uses non-exist macro, we do not use macros for platform-dependent code, the added test does not test what it claims.

Aristide021:
> The code uses non-exist macro, we do not use macros for platform-dependent code, the added test does not test what it claims.

Thanks for the feedback. You're right on both points. The current tests validate compile-cache behavior indirectly, but they do not directly prove Metal library ownership/cleanup semantics, and I understand that's not sufficient. Given the lack of a clean direct test hook for this internal ownership path, I'm okay with closing this PR as-is.

PR #3328, opened 2026-03-28
Add TurboQuant KV cache compression with native Metal SDPA kernel

arozanov:
## Proposed changes

Adds TurboQuant (arXiv 2504.19874) as a native Metal SDPA kernel for KV cache compression.

- New `QuantizationMode::TurboQuant`
- `sdpa_vector_turbo` Metal kernel: reads 3-bit packed K indices with codebook dequant
- Pre-rotated query optimization: no WHT butterfly in attention inner loop
- `TurboQuantSDPA` primitive with full eval_gpu dispatch
- Python API: `mx.fast.turboquant_sdpa()`

Benchmarks (M4 Pro 48GB, 28 query heads, 4 KV heads, D=128):

| Context | Apple SDPA | TurboQuant | Speedup |
|---------|-----------|------------|---------|
| 1K | 0.161ms | 0.108ms | 1.5x |
| 4K | 0.164ms | 0.109ms | 1.5x |
| 8K | 0.209ms | 0.106ms | 2.0x |
| 16K | 0.511ms | 0.104ms | 4.9x |

TurboQuant reads 3-bit packed data (4.6x less memory bandwidth than fp16). Kernel time constant at ~0.1ms regardless of context length.

Related: standalone package at https://github.com/arozanov/turboquant-mlx and mlx-lm PR at https://github.com/ml-explore/mlx-lm/pull/1067

## Checklist

### DIFF
--- a/mlx/backend/metal/kernels/CMakeLists.txt
+++ b/mlx/backend/metal/kernels/CMakeLists.txt
@@ -29,7 +29,7 @@ function(build_kernel_base TARGET SRCFILE DEPS)
                     "-mmacosx-version-min=${CMAKE_OSX_DEPLOYMENT_TARGET}")
   endif()
   add_custom_command(
-    COMMAND xcrun -sdk macosx metal ${METAL_FLAGS} -c ${SRCFILE}
+    COMMAND xcrun metal ${METAL_FLAGS} -c ${SRCFILE}
             -I${PROJECT_SOURCE_DIR} -o ${TARGET}.air
     DEPENDS ${SRCFILE} ${DEPS} ${BASE_HEADERS}
     OUTPUT ${TARGET}.air
@@ -176,7 +176,7 @@ endif()
 
 add_custom_command(
   OUTPUT ${MLX_METAL_PATH}/mlx.metallib
-  COMMAND xcrun -sdk macosx metallib ${KERNEL_AIR} -o
+  COMMAND xcrun metallib ${KERNEL_AIR} -o
           ${MLX_METAL_PATH}/mlx.metallib
   DEPENDS ${KERNEL_AIR}
   COMMENT "Building mlx.metallib"
--- a/mlx/backend/metal/kernels/CMakeLists.txt
+++ b/mlx/backend/metal/kernels/CMakeLists.txt
@@ -29,8 +29,8 @@ function(build_kernel_base TARGET SRCFILE DEPS)
                     "-mmacosx-version-min=${CMAKE_OSX_DEPLOYMENT_TARGET}")
   endif()
   add_custom_command(
-    COMMAND xcrun metal ${METAL_FLAGS} -c ${SRCFILE}
-            -I${PROJECT_SOURCE_DIR} -o ${TARGET}.air
+    COMMAND xcrun metal ${METAL_FLAGS} -c ${SRCFILE} -I${PROJECT_SOURCE_DIR} -o
+            ${TARGET}.air
     DEPENDS ${SRCFILE} ${DEPS} ${BASE_HEADERS}
     OUTPUT ${TARGET}.air
     COMMENT "Building ${TARGET}.air"
@@ -176,8 +176,7 @@ endif()
 
 add_custom_command(
   OUTPUT ${MLX_METAL_PATH}/mlx.metallib
-  COMMAND xcrun metallib ${KERNEL_AIR} -o
-          ${MLX_METAL_PATH}/mlx.metallib
+  COMMAND xcrun metallib ${KERNEL_AIR} -o ${MLX_METAL_PATH}/mlx.metallib
   DEPENDS ${KERNEL_AIR}
   COMMENT "Building mlx.metallib"
   VERBATIM)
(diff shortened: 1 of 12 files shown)


yzamari:
Hey @arozanov , 
 nice to see someone else working on TurboQuant for MLX!

  ▎ I opened a related PR (#3340) that was closed as a duplicate of this one, but the implementations are quite different —
  mine covers the full two-stage algorithm (PolarQuant + QJL residual correction from Algorithm 2 of the paper), which gives
  unbiased attention score estimation. It also supports 2-bit and 4-bit quantization and has a two-pass kernel for long
  sequences.

  ▎ A few questions/observations on this PR:

  ▎ 1. Are there plans to add the QJL stage? Without it, the attention scores have a systematic bias from the quantization
  error that isn't corrected.
  ▎ 2. The codebook centroids appear to be hardcoded from a Gaussian N(0,1) distribution — the paper derives them from the Beta
   distribution on S^{d-1} (Theorem 3.1). This affects the optimality of the quantization boundaries.
  ▎ 3. Have you considered supporting bit widths other than 3-bit?

  ▎ Happy to collaborate if it would help move TurboQuant support forward in MLX. The full algorithm makes a meaningful
  difference for long-context quality.

zcbenz:
Thanks for the PR and I think it is nicely written.

But I think we need to have a generic quantized sdpa kernel first and then add things like turbo quant as part of the API, rather than a specialized turbo quant kernel directly. #3026 is a nice step but as you can see we don't quite have enough maintainers to review.

On turbo quant itself, we don't just add builtin kernels for every new hot thing, users should use [custom extensions](https://ml-explore.github.io/mlx/build/html/dev/extensions.html) or custom kernels instead.

arozanov:
> Thanks for the PR and I think it is nicely written.
> 
> But I think we need to have a generic quantized sdpa kernel first and then add things like turbo quant as part of the API, rather than a specialized turbo quant kernel directly. #3026 is a nice step but as you can see we don't quite have enough maintainers to review.
> 
> On turbo quant itself, we don't just add builtin kernels for every new hot thing, users should use [custom extensions](https://ml-explore.github.io/mlx/build/html/dev/extensions.html) or custom kernels instead.

Makes sense. Happy to help with #3026 if useful - the packed-integer codepath here (3-bit indices + codebook lookup) could serve as a reference for one of the quantization modes a generic API would need.
I'll package this as a custom Metal kernel extension via mx.fast.metal_kernel() in the meantime.

deceptech-packet-ninja:
## Bug: `N` reads wrong dimension of `k_norms`

While testing this kernel locally (built from source), I found that line 450 in `scaled_dot_product_attention.cpp`:

```cpp
int N = k_norms.shape(1); // kv sequence length
```

reads `shape(1)` which is the **head dimension** (H_kv), not the **sequence length** (T). For `k_norms` of shape `(B, H_kv, T)`, this should be:

```cpp
int N = k_norms.shape(2); // kv sequence length
```

With `shape(1)`, the kernel only processes `H_kv` keys (e.g., 8 instead of 500), causing incorrect attention output. After changing to `shape(2)`, the kernel produces **exact match** (1.000 cosine similarity) against the reference `mx.fast.scaled_dot_product_attention`.

Kernel benchmarks after the fix (M-series, Llama-3-8B config: B=1, H_q=32, H_kv=8, D=128):

| Context | Standard SDPA | TurboQuant SDPA | Speedup |
|---------|--------------|-----------------|---------|
| 256 | 0.43ms | 0.44ms | 1.0x |
| 512 | 0.49ms | 0.22ms | **2.2x** |
| 1024 | 0.52ms | 0.20ms | **2.6x** |
| 4096 | 1.91ms | 0.30ms | **6.3x** |

The kernel is impressively fast — essentially constant time regardless of context length.

arozanov:
> ## Bug: `N` reads wrong dimension of `k_norms`
> While testing this kernel locally (built from source), I found that line 450 in `scaled_dot_product_attention.cpp`:
> 
> ```c++
> int N = k_norms.shape(1); // kv sequence length
> ```
> 
> reads `shape(1)` which is the **head dimension** (H_kv), not the **sequence length** (T). For `k_norms` of shape `(B, H_kv, T)`, this should be:
> 
> ```c++
> int N = k_norms.shape(2); // kv sequence length
> ```
> 
> With `shape(1)`, the kernel only processes `H_kv` keys (e.g., 8 instead of 500), causing incorrect attention output. After changing to `shape(2)`, the kernel produces **exact match** (1.000 cosine similarity) against the reference `mx.fast.scaled_dot_product_attention`.
> 
> Kernel benchmarks after the fix (M-series, Llama-3-8B config: B=1, H_q=32, H_kv=8, D=128):
> 
> Context	Standard SDPA	TurboQuant SDPA	Speedup
> 256	0.43ms	0.44ms	1.0x
> 512	0.49ms	0.22ms	**2.2x**
> 1024	0.52ms	0.20ms	**2.6x**
> 4096	1.91ms	0.30ms	**6.3x**
> The kernel is impressively fast — essentially constant time regardless of context length.

Good catch, thanks! Fixed: k_norms.shape(1) → k_norms.shape(2). The shape is (B, H_kv, T) so dimension 2 is the sequence length.

zcbenz:
I'm closing this PR per the comment https://github.com/ml-explore/mlx/pull/3328#issuecomment-4160633267 above.

PR #3342, opened 2026-03-31
Add array.data_ptr() Python binding for low-level interop

Aristide021:
## Proposed changes

Exposes `array.data<void>()` in Python as data_ptr() -> int, returning the address of the first element after evaluation.

Implementation:
- Release GIL before `eval()` to avoid blocking during synchronization
- Return `uintptr_t` cast to Python `int`
- Pointer validity is tied to underlying storage lifetime

Tests:
- Add `test_data_ptr` in `python/tests/test_array.py`
- Cover base contiguous array, positive-offset slice, transpose view, and negative-stride view
- Validate pointers against NumPy `__array_interface__` and verify pointed values via `ctypes`

## Checklist

### DIFF
--- a/python/src/array.cpp
+++ b/python/src/array.cpp
@@ -498,6 +498,27 @@ void init_array(nb::module_& m) {
               new (&arr) mx::array(nd_array_to_mlx(nd, std::nullopt));
             }
           })
+      .def(
+          "data_ptr",
+          [](mx::array& a) {
+            {
+              nb::gil_scoped_release nogil;
+              a.eval();
+            }
+            return reinterpret_cast<uintptr_t>(a.data<void>());
+          },
+          nb::sig("def data_ptr(self) -> int"),
+          R"pbdoc(
+            Return the address of the first element as an integer.
+
+            This returns a pointer to the array's underlying storage and is
+            intended for low-level interop.
+
+            .. warning::
+
+               The pointer is only valid while the array's underlying storage
+               remains alive.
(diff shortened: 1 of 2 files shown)


zcbenz:
Thanks for the PR but conversion of arrays is already supported via buffer protocol and dlpack protocol: https://ml-explore.github.io/mlx/build/html/usage/numpy.html . We should not add another way to do the same thing.

Aristide021:
> Thanks for the PR but conversion of arrays is already supported via buffer protocol and dlpack protocol: https://ml-explore.github.io/mlx/build/html/usage/numpy.html . We should not add another way to do the same thing.

@zcbenz The linked documentation covers the read direction. The use case here was the write direction for zero-copy writes from external kernels into unified memory. Neither DLPack nor the buffer protocol exposes that cleanly.

I also want to note that `array.data<void>()` already exists in the C++ API. So this isn't another way to do the same thing, it's a Python binding of something that's already there. Closing it as "use DLPack" conflates two different things.

zcbenz:
The dlpack protocol does support zero-copy writing, from the linked doc:

```python
a = mx.arange(3)
a_view = np.array(a, copy=False)
print(a_view.flags.owndata) # False
a_view[0] = 1
print(a[0].item()) # 1
```

We don't just expose C++ interface to Python with a 1:1 mapping, for example in C++ we use `begin`/`end` for iterators but `__iter__` in Python.

PR #3343, opened 2026-03-31
Fix use after move

angeloskath:
Well there is 1 very real bug and then the rest are officially bugs but don't really show up.

Basically I came across the `Broadcast` use after move now that I was looking into #3319 and I went to a small quest to find more. Well good news I didn't quite find many.

The main pattern that is questionable is taking a reference to a shape and then moving the array. I think officially this is undefined behavior. The change here is actually incurring no overhead since it copies the shape but then moves it to the array constructor.

### DIFF
--- a/mlx/ops.cpp
+++ b/mlx/ops.cpp
@@ -1658,9 +1658,11 @@ std::vector<array> broadcast_arrays(
       if (in.shape() == out_shape) {
         outputs.push_back(in);
       } else {
-        auto prim = std::make_shared<Broadcast>(to_stream(s), out_shape);
-        outputs.push_back(
-            array(std::move(out_shape), in.dtype(), std::move(prim), {in}));
+        outputs.push_back(array(
+            out_shape,
+            in.dtype(),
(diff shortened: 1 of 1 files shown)


angeloskath:
Oh I see it was already fixed on main sorry for the noise. Well I 'll leave the PR as it does one less copy on that bug and also fixes the other asymptomatic ones.

zcbenz:
With SmallVector there is actually not much meaning to move shapes and strides since they no longer allocate on heap (unless it has very large ndim which is forbidden in CUDA backend), so std::move would just add overhead erasing the old object's memory to 0s.

angeloskath:
Interesting... force of habit I guess. Do you think we should remove the moves of Shape and Strides throughout ?

There is probably more than 100 places.
```
$ grep -nr 'std::move(shape)' mlx | wc -l
      52
```

zcbenz:
I think it does not worth the effort changing all existing code as it is not going to make much difference in performance, we should just avoid introducing new ones.

Issue #3358, opened 2026-04-03
[BUG]  get_shape() has no bounds check allowing read of outside data structure


MillaFleurs:
**Describe the bug**
  1. get_shape() has no bounds check (mlx/io/gguf.cpp:50-57)
  for (int i = tensor.ndim - 1; i >= 0; i--) {
      shape.push_back(tensor.dim[i]);  // no check that ndim <= 8
  }

  2. dim[] is fixed at 8 elements (vendored gguflib.h)
  #define GGUF_TENSOR_MAX_DIM 8
  uint64_t dim[GGUF_TENSOR_MAX_DIM];

  3. The only guard is an assert() (vendored gguflib.c:279)
  tensor->ndim = *num_dim;
  assert(tensor->ndim <= GGUF_TENSOR_MAX_DIM);  // compiled out with NDEBUG

  4. No NDEBUG override in the build (mlx/io/CMakeLists.txt:9-26) — gguflib is built as a plain static library with no flags to
  preserve asserts in release mode.

  5. Two call sites reach get_shape() with untrusted tensor data:
  - mlx/io/gguf.cpp:234 — non-quantized tensor loading
  - mlx/io/gguf_quants.cpp:112 — quantized tensor loading

6.  If you access dim[n] where n >= 8 you read the following data:

```
 typedef struct {
      const char *name;                   //  8 bytes
      size_t namelen;                     //  8 bytes
      uint32_t type;                      //  4 bytes
      uint32_t ndim;                      //  4 bytes
      uint64_t dim[8];                    // 64 bytes  <-- dim[0] through dim[7]
      // --- everything below is what dim[8+] reads ---
      uint64_t offset;                    //  8 bytes  → dim[8]
      uint64_t bsize;                     //  8 bytes  → dim[9]
      uint64_t num_weights;               //  8 bytes  → dim[10]
      uint8_t *weights_data;              //  8 bytes  → dim[11]
      // --- past the struct boundary ---
      // dim[12..31] → stack frame: local variables, saved registers,
      //               return address, caller's frame, etc.
  } gguf_tensor;

```

**To Reproduce**
```c
/*
 * POC: GGUF Tensor ndim Stack Out-of-Bounds Read (ADVISORY-2026-001 / V1-002)
 *
 * Demonstrates two vulnerabilities triggered by a crafted GGUF with ndim > 8:
 *
 *   Part A: Shows that assert() is a no-op in release builds (NDEBUG),
[...]

PR #3359, opened 2026-04-03
Prevent out-of-bounds memory access caused by corrupt tensor.ndim in gguf file

MillaFleurs:
## Proposed changes

Fix for #3358 get_shape() lacks bounds checking.  

The GGUF tensor loader in Apple MLX trusts the `ndim` field from a crafted GGUF file without bounds-checking it against the fixed-size `dim[]` array in the `gguf_tensor` struct. In release builds, the upstream `gguflib.c` `assert()` guard is compiled out (`NDEBUG`), leaving no enforcement. MLX's `get_shape()` function iterates `ndim` times over the stack-allocated `dim[]` array (maximum 8 elements), reading beyond its bounds when `ndim > 8`.

## Checklist

### DIFF
--- a/docs/src/usage/saving_and_loading.rst
+++ b/docs/src/usage/saving_and_loading.rst
@@ -79,3 +79,10 @@ The functions :func:`save_safetensors` and :func:`save_gguf` are similar to
    >>> a = mx.array([1.0])
    >>> b = mx.array([2.0])
    >>> mx.save_safetensors("arrays", {"a": a, "b": b})
+
+.. note::
+
+   When loading GGUF files from untrusted sources, MLX validates tensor
+   dimensions against the maximum supported by the format (8 dimensions).
+   Files with invalid dimension counts will raise an error rather than
+   risk reading out-of-bounds memory.
--- a/docs/src/usage/saving_and_loading.rst
+++ b/docs/src/usage/saving_and_loading.rst
@@ -79,10 +79,3 @@ The functions :func:`save_safetensors` and :func:`save_gguf` are similar to
    >>> a = mx.array([1.0])
    >>> b = mx.array([2.0])
    >>> mx.save_safetensors("arrays", {"a": a, "b": b})
-
-.. note::
-
-   When loading GGUF files from untrusted sources, MLX validates tensor
-   dimensions against the maximum supported by the format (8 dimensions).
-   Files with invalid dimension counts will raise an error rather than
-   risk reading out-of-bounds memory.
--- a/mlx/io/gguf.cpp
+++ b/mlx/io/gguf.cpp
@@ -1,10 +1,10 @@
 // Copyright © 2023-2024 Apple Inc.
 
+#include <fmt/format.h>
 #include <cstdint>
 #include <cstring>
 #include <fstream>
 #include <numeric>
-#include <sstream>
 
 #include "mlx/io/gguf.h"
 #include "mlx/ops.h"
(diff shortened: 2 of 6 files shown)


zcbenz (changes requested):
I don't think this can prevent a bad/malicious model file.

For example the file could provide `ndim: 4` with a tensor of 0 size, which would still crash the reading program.

MillaFleurs:
You're right that this doesn't prevent all malformed files — but it's not intended to. This specifically fixes an out-of-bounds memory read where ndim > 8 causes get_shape() to read past the end of the fixed-size dim[8] array into adjacent stack memory. A zero-dimension tensor with ndim=4 stays within array bounds and doesn't trigger the same class of bug. Additional validation for semantic issues like zero-size tensors could be a good follow-up, but is orthogonal to this memory safety fix.

docs/src/usage/saving_and_loading.rst
```diff
@@ -79,3 +79,10 @@ The functions :func:`save_safetensors` and :func:`save_gguf` are similar to
    >>> a = mx.array([1.0])
    >>> b = mx.array([2.0])
    >>> mx.save_safetensors("arrays", {"a": a, "b": b})
+
+.. note::
+
+   When loading GGUF files from untrusted sources, MLX validates tensor
+   dimensions against the maximum supported by the format (8 dimensions).
+   Files with invalid dimension counts will raise an error rather than
+   risk reading out-of-bounds memory.
```
zcbenz: This note is not really need - it should be for behavior that surprising to users.

mlx/io/gguf.cpp
```diff
@@ -48,6 +49,13 @@ std::optional<Dtype> gguf_type_to_dtype(const uint32_t& gguf_type) {
 }
 
 Shape get_shape(const gguf_tensor& tensor) {
+  if (tensor.ndim > MLX_GGUF_MAX_DIMS) {
+    std::ostringstream msg;
```
zcbenz: Can you use `fmt::format`?

mlx/io/gguf.h
```diff
@@ -10,6 +10,13 @@ extern "C" {
 #include <gguflib.h>
 }
 
+// Maximum number of tensor dimensions supported by the GGUF format.
+// Mirrors GGUF_TENSOR_MAX_DIM from gguflib.h. Override at compile time
+// with -DMLX_GGUF_MAX_DIMS=<value> if the upstream format changes.
+#ifndef MLX_GGUF_MAX_DIMS
+#define MLX_GGUF_MAX_DIMS GGUF_TENSOR_MAX_DIM
```
zcbenz: It is very unlikely we would override this, we should just use `GGUF_TENSOR_MAX_DIM`.

python/tests/test_load.py
```diff
@@ -300,6 +300,43 @@ def test_save_and_load_gguf_metadata_mixed(self):
             else:
                 self.assertEqual(meta_load_dict[k], v)
 
+    @unittest.skipIf(platform.system() == "Windows", "GGUF is disabled on Windows")
+    def test_gguf_rejects_oversized_ndim(self):
```
zcbenz: Only having the C++ test should be enough.

zcbenz (changes requested):
Thanks for the clarification, to be honest I think it should be checked in gguflib, can you send a PR there?

docs/src/usage/saving_and_loading.rst
MillaFleurs: Yes more informational for me but agreed...

MillaFleurs:
I made the requested changes @zcbenz .  Thank you for the feedback it's extremely helpful and insightful.

MillaFleurs:
> Thanks for the clarification, to be honest I think it should be checked in gguflib, can you send a PR there?

I will file a PR on github.com/antirez/gguf-tools as well that's a great idea.  We can do both as well.  By keeping the
  MLX-side check we are not dependent on waiting for their fix.

zcbenz (approved):
Thanks!

zcbenz:
The test does not work since the assertion works under debug build, I think we can `#define NDEBUG` when using gguflib or wait for gguflib to fix it, either way we shouldn't need the fix in this PR?

MillaFleurs:
Updated thank you!

zcbenz (changes requested):
The define does not work somehow, and after thinking more about this I'm worried that we would be disabling actually useful assertions in gguflib.

So I think we have 2 choices here:
* Wait for gguflib to turn the assert into a runtime check.
* Find a way to force enabling all the asserts in gguflib in Release build.

qflen:
Opened #3436 with the "force-enable all the asserts in gguflib" variant you sketched: `target_compile_options(gguflib PRIVATE -UNDEBUG)` plus a defensive `ndim > GGUF_TENSOR_MAX_DIM` check in `get_shape()`. Not trying to step on this PR, just offering an alternative since it has been stalled. Happy to defer if @MillaFleurs wants to take it from here.

zcbenz:
Closing in favor of #3436.

Issue #3363, opened 2026-04-03
[BUG]  SafeTensors data_offsets Not Validated


MillaFleurs:
**Describe the bug**

The SafeTensors loader reads `data_offsets` from JSON metadata but does not validate the entry count, ordering, or consistency with the declared tensor shape and dtype. Either through malice or incompetence, a user can declare a large shape (e.g., 1000×1000 float32 = 4 MB) while specifying `data_offsets` that cover only 4 bytes of actual data. When the tensor is evaluated, the loader reads far beyond the provided data, producing out-of-bounds memory access.

**To Reproduce**

Include code snippet
```python

import mlx.core as mx
arrays, _ = mx.load("crash_v3006.safetensors", return_metadata=True)
mx.eval(arrays["t"])
t = arrays["t"]
print(t.shape)    # [1000, 1000]
print(t.nbytes)   # 4000000
print(t[0, :10])  # whatever was in memory past the file
```

**Expected behavior**
We are fetching 4000 bytes but passed in only 82 bytes.  In MacOS the OS fills the out of bounds read with 0s but I don't think this is in any way guaranteed.  Best practice would be to throw an error if we read outside the array.

**Desktop (please complete the following information):**
 - OS Version: MacOS 26.3.1 (a) on M2 16GB 

[crash_v3006.safetensors.gz](https://github.com/user-attachments/files/26473756/crash_v3006.safetensors.gz)

angeloskath:
Closing as the PR is merged already

PR #3364, opened 2026-04-04
Validate safetensors data offsets

MillaFleurs:
## Proposed changes

#3363 Fix

The SafeTensors loader reads `data_offsets` from JSON metadata but does not validate the entry count, ordering, or consistency with the declared tensor shape and dtype. An attacker can declare a large shape (e.g., 1000×1000 float32 = 4 MB) while specifying `data_offsets` that cover only 4 bytes of actual data. When the tensor is evaluated, the loader reads far beyond the provided data, producing out-of-bounds memory access.

## Checklist

### DIFF
--- a/mlx/io/safetensors.cpp
+++ b/mlx/io/safetensors.cpp
@@ -1,7 +1,8 @@
 // Copyright © 2023 Apple Inc.
-//
+
 #include <json.hpp>
 #include <memory>
+#include <sstream>
 #include <stack>
 
 #include "mlx/backend/cuda/cuda.h"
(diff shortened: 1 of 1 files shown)


MillaFleurs:
Good observation. You're describing a scenario where all three metadata fields (shape, dtype, data_offsets) are internally consistent but the file doesn't actually contain enough data to back them. That's a valid concern, but it's a different class of issue from what this PR addresses.  

I think you're right that this should be a seperate PR!

This PR fixes the case where data_offsets and shape*dtype disagree — an attacker declares a 4-byte data range but a 1000x1000 shape. Without this check, the loader silently constructs a 4MB tensor backed by 4 bytes of data. The consistency check catches this contradiction at load time, which is exactly what the https://docs.rs/safetensors/latest/safetensors/tensor/enum.SafeTensorError.html also enforces (via TensorInvalidInfo).

The scenario you describe is consistent metadata exceeding the actual file size and would require an additional file-size boundary check (which the Rust reference implementation also does via MetadataIncompleteBuffer). 

I'd agree we should add as a follow-up improvement in a new PR.

Importantly, in that scenario the read() call will fail at eval time with an I/O error rather than silently reading garbage, so the impact is lower than the silent OOB this PR prevents.

MillaFleurs:
Rebase done thank you!

Issue #3365, opened 2026-04-04
[BUG] header_len field read from .npy v2 files without any upper-bound validation


MillaFleurs:
**Describe the bug**
The MLX NumPy loader trusts the header_len field read from .npy v2 files without any upper-bound validation. This field is a uint32_t, allowing an attacker-controlled value up to 4,294,967,295 (~4 GiB). The loader passes this value directly to std::vector<char>(header_len + 1), triggering an unchecked heap allocation proportional to the attacker's chosen value — from a file as small as 12 bytes.

On systems where the allocation succeeds (macOS will overcommit virtual memory by default), the subsequent read() call fails because the file contains no actual header data, producing a generic I/O error. On systems with stricter memory limits, the allocation itself throws std::bad_alloc or triggers the OOM killer. In either case, a 12-byte crafted file reliably denies service to any process that loads it.
  
By contrast, the v1 format uses uint16_t for header_len (max 65,535 bytes), which is implicitly safe. The fix caps v2 header_len to the same 65,536-byte limit, which is more than sufficient for any legitimate NumPy header while preventing the allocation amplification attack.

Interestingly, if you check out NumPy's _format_imply.py at https://github.com/numpy/numpy/blob/main/numpy/lib/_format_impl.py we get further validation of the need for upper bounds checking.

Specifically:

```python
# line 193
# Python's literal_eval is not actually safe for large inputs, since parsing
# may become slow or even cause interpreter crashes.
# This is an arbitrary, low limit which should make it safe in practice.
_MAX_HEADER_SIZE = 10000
# further on line 638
    if len(header) > max_header_size:
        raise ValueError(
            f"Header info length ({len(header)}) is large and may not be safe "
            "to load securely.\n"
            "To allow loading, adjust `max_header_size` or fully trust "
            "the `.npy` file using `allow_pickle=True`.\n"
            "For safety against large resource use or crashes, sandboxing "
            "may be necessary.")

```

Note NumPy's approach is still not ideal — _read_bytes(fp, header_length) at line 636 uses Python's fp.read(n) which only returns what's available (won't allocate 4 GiB from a 12-byte file). The validation at line 638 is a second layer of defense.

  MLX has neither protection:
  1. No implicit read safety — std::vector<char>(n) allocates the full size upfront before reading
  2. No post-read max_header_size check

  So NumPy's own code actually validates the need for a fix here.  (even if their 

**To Reproduce**

Include code snippet
```python

  import mlx.core as mx
[...]

PR #3366, opened 2026-04-04
PR3365 Fix for no upper bound check on header_len

MillaFleurs:
## Proposed changes

The fix adds a bounds check on the v2_header_len value immediately after it's read from the .npy file and before it's used to allocate memory. A constexpr uint32_t kMaxNpyHeaderLen = 65536 cap is enforced — if the file declares a header length exceeding 65,536 bytes, a std::runtime_error is thrown with a descriptive message including the offending value and the file path. This prevents the unchecked header_len from flowing into std::vector<char>(header_len + 1) on the next line, which would otherwise attempt an attacker-controlled allocation of up to ~4 GiB. The 65,536-byte cap is well above any legitimate NumPy header (NumPy's own implementation defaults to a 10,000-byte safety limit) while eliminating the allocation amplification vector.

This is the fix for #3365 

## Checklist

- [X ] I have read the [CONTRIBUTING](https://github.com/ml-explore/mlx/blob/main/CONTRIBUTING.md) document
- [ X] I have run `pre-commit run --all-files` to format my code / installed pre-commit prior to committing changes
- [X ] I have updated the necessary documentation (if needed)

### DIFF
--- a/docs/src/usage/saving_and_loading.rst
+++ b/docs/src/usage/saving_and_loading.rst
@@ -79,3 +79,10 @@ The functions :func:`save_safetensors` and :func:`save_gguf` are similar to
    >>> a = mx.array([1.0])
    >>> b = mx.array([2.0])
    >>> mx.save_safetensors("arrays", {"a": a, "b": b})
+
+.. note::
+
+   When loading ``.npy`` files from untrusted sources, MLX validates that the
+   header length does not exceed 65,536 bytes. NumPy v2 format files with
+   excessively large header lengths will raise an error rather than risk
+   denial-of-service through memory exhaustion.
--- a/python/tests/test_load.py
+++ b/python/tests/test_load.py
@@ -127,6 +127,22 @@ class TestLoad(mlx_tests.MLXTestCase):
                             mx.array_equal(load_dict["test"], save_dict["test"])
                         )
 
+    def test_npy_v2_rejects_oversized_header_len(self):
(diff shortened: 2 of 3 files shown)


MillaFleurs:
Excellent catch.

So philosophically this fix follows NumPys implementation which  caps at 10,000 (_MAX_HEADER_SIZE in numpy/lib/_format_impl.py:196). They added this specifically because "parsing may become slow or even cause interpreter crashes" for large headers. 65,536 is already 6x more generous.

I suppose it might make sense to determine at what point slow downs happen.  I'm not wedded to the 65,536 bytes.

But 65,536 is the natural maximum of uint16_t max which is why I chose it

Note that without any cap, a 12-byte file triggers a 4 GiB heap allocation. The allocation happens unconditionally at std::vector<char>(header_len + 1) before any read. Even if the read fails afterward, the process has already committed ~4 GiB of virtual memory. On memory-constrained systems this is an instant OOM kill.

So the question isn't "why reject this value but not that one" — it's "what's the largest header that could ever be legitimate?" NumPy says 10,000.  There may be a better answer, I'm honestly not sure.

Any ideas for how we could test or determine what the limit should be?

zcbenz:
Thanks for the reference on the numpy implementation. I think this check is not going to have practical meanings as a malicious file could still make a OOM by creating millions of small tensors rather than a single large tensor. And OOM is not something we consider guarding against as it happens all the time for users of LLM.

PR #3367, opened 2026-04-04
[CUDA] Thread safety

zcbenz:
Refs https://github.com/ml-explore/mlx/issues/3078.

These are the thread safety changes for CUDA backend, which basically do the same thing with the Metal backend:
* Use lock for JIT compilation.
* Make CommandEncoder thread local.
* Leak the device-level resources on exit.
* Move management of cublasLt/cuDNN handles out of `cu::Device` to avoid using locks for multi-thread environment.

### DIFF
--- a/mlx/backend/cuda/conv.cpp
+++ b/mlx/backend/cuda/conv.cpp
@@ -39,7 +39,7 @@ struct ConvCacheKey {
 };
 
 auto& conv_cache() {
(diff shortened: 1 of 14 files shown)


zcbenz:
Yeah you are right, somehow I remembered it the opposite way. The cudnn/cublas handles are now thread local and destroyed on thread teardown.

Issue #3368, opened 2026-04-04
[BUG] deserializer reads tape_size as a raw uint64_t from the file with no upper bound check


MillaFleurs:
**Describe the bug**

When importing a .mlxfn file, the deserializer reads tape_size as a raw uint64_t from the file and passes it directly to tape.reserve(tape_size) with no upper bound check. Since tape_size is attacker-controlled, this allows a crafted file to trigger an unbounded memory allocation.

The failure mode depends on the chosen value. With tape_size = UINT64_MAX, std::vector::reserve() throws std::length_error because the value exceeds vector::max_size() — this surfaces as ValueError: vector in Python. With a large but representable value (e.g. 10 billion), the allocation attempt succeeds past the vector limit check but exhausts physical memory, causing OOM kill or std::bad_alloc. In either case, a 58-byte crafted file reliably denies service to any process that imports it.

The .mlxfn format has no authentication or integrity checking, and mx.import_function() is intended for loading user-provided files (e.g. exported model functions). The fix caps tape_size at 10,000,000 entries — well above any realistic computation graph while preventing the allocation amplification attack.

**To Reproduce**

Include code snippet
```python
"""
POC: .mlxfn tape_size Memory Bomb 

Demonstrates that a crafted .mlxfn file with an oversized tape_size causes
the MLX importer to attempt an unbounded memory allocation via
tape.reserve(tape_size) with no upper bound check.

The failure mode depends on the tape_size value:
  - UINT64_MAX: std::length_error (exceeds vector::max_size()) -> ValueError
  - Large realistic value (e.g. 10 billion): std::bad_alloc or OOM kill

Both cases confirm the vulnerability: untrusted input directly controls
a memory allocation with no validation.

On unpatched MLX:
  - UINT64_MAX -> ValueError: vector (std::length_error from reserve)
  - 10 billion -> OOM kill or std::bad_alloc (DO NOT test without safeguards)

On patched MLX:
  - RuntimeError at import time before any allocation attempt

Usage:
    python3 poc_mlxfn_tape_bomb.py

Generates: bomb_v5002.mlxfn
"""

import struct
import sys


def build_mlxfn(tape_size):
    """Build a minimal .mlxfn payload with the given tape_size."""
    buf = bytearray()

    # Version string: "0.0.0"
    ver = b"0.0.0"
    buf += struct.pack("<Q", len(ver)) + ver

    # function_count = 1
    buf += struct.pack("<i", 1)

    # shapeless = false
[...]

PR #3370, opened 2026-04-04
Fixing deserializer missing upper bounds check as per PR#3368

MillaFleurs:
## Proposed changes

The fix adds a bounds check on tape_size in import_one() (mlx/export.cpp) immediately after it's deserialized from the .mlxfn file and before it's passed to tape.reserve(). If tape_size exceeds 10,000,000 entries, a std::runtime_error is thrown with a descriptive message. This prevents an attacker-controlled uint64_t from flowing into an unbounded std::vector::reserve() call, which would otherwise attempt to allocate up to exabytes of memory from a 58-byte file. The cap of 10 million is well above any realistic computation graph while eliminating the allocation amplification vector.

Fix for #3368 

## Checklist

### DIFF
--- a/mlx/export.cpp
+++ b/mlx/export.cpp
@@ -1062,6 +1062,13 @@ ImportedFunction::ImportedFunction(const std::string& file)
 
     std::vector<array> tape;
     auto tape_size = deserialize<uint64_t>(is);
+    constexpr uint64_t kMaxTapeSize = 10'000'000;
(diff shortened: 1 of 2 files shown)


zcbenz:
As same with https://github.com/ml-explore/mlx/pull/3366#issuecomment-4186592930 we don't guard against OOM or insanely large graph, it happens a lot with just normal python code.

PR #3374, opened 2026-04-04
Add basic fp8/bfp8 support, future implementations of 8-bit floating …

Geramy:
…point functions next.

## Proposed changes

The proposed changes are to include fp8 and bfp8 support.

## Checklist

### DIFF
--- a/mlx/array.h
+++ b/mlx/array.h
@@ -615,6 +615,12 @@ void array::init(It src) {
     case int64:
       std::copy(src, src + size(), data<int64_t>());
       break;
+    case float8:
+      std::copy(src, src + size(), data<float8_t>());
+      break;
+    case bfloat8:
+      std::copy(src, src + size(), data<bfloat8_t>());
+      break;
     case float16:
       std::copy(src, src + size(), data<float16_t>());
       break;
--- a/mlx/dtype.h
+++ b/mlx/dtype.h
@@ -8,6 +8,7 @@
 #include "mlx/api.h"
 #include "mlx/types/complex.h"
 #include "mlx/types/half_types.h"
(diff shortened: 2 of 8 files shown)


zcbenz:
Rather than `float8`/`bfloat8`, I think it is more common using names like `float8_e4m3`/`float8_e5m2`, and there are also a few other float8 types (e.g. `e8m0`) that we should consider in advance.

JAX's float8 types: https://github.com/jax-ml/ml_dtypes
CUTLASS: https://github.com/NVIDIA/cutlass/blob/main/include/cutlass/float8.h

I think we should have a consensus on the type names with the team first.

Geramy:
> Rather than `float8`/`bfloat8`, I think it is more common using names like `float8_e4m3`/`float8_e5m2`, and there are also a few other float8 types (e.g. `e8m0`) that we should consider in advance.
> 
> JAX's float8 types: https://github.com/jax-ml/ml_dtypes CUTLASS: https://github.com/NVIDIA/cutlass/blob/main/include/cutlass/float8.h
> 
> I think we should have a consensus on the type names with the team first.

Sounds good, I actually think that is a lot better since we have to think about compatibility between all possible backends. It could turn into a big problem later I would assume. Let me know what you guys decide on I 100% agree on changing the names upfront.

zcbenz:
I don't think we have bandwidth for fp8 recently, and how we support fp8 in MLX would mostly depend on how native fp8 would work in Metal: the ideal approach for fp8 would be Metal gets fp8 first => MLX adds support for fp8, rather than adding a half-baked fp8 implementation first and hoping the upcoming Metal fp8 would fit in. So I'm closing this PR.

PR #3382, opened 2026-04-06
Fix: Correct cross-attention query routing in Post-LN TransformerDecoderLayer

suryawanshishantanu6:
## Proposed changes

This PR fixes a critical data flow bug in TransformerDecoderLayer when initialized with norm_first=False.

Previously, the cross-attention module was receiving y (the raw, dropped-out output of the self-attention layer) as its query. This inadvertently bypassed the first LayerNorm (self.ln1) and severed the residual connection (x + y). I updated the query input to x to correctly route the output of the first Add & Norm block into the cross-attention block, aligning with standard Post-LN Transformer architecture.

## Checklist

### DIFF
--- a/python/tests/test_nn.py
+++ b/python/tests/test_nn.py
@@ -2081,59 +2081,6 @@ class TestLayers(mlx_tests.MLXTestCase):
         out = model(src, tgt, src_mask=None, tgt_mask=None, memory_mask=None)
         self.assertEqual(out.shape, tgt.shape)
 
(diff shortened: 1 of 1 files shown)


suryawanshishantanu6:
> Thanks for catching that!
> 
> I removed the last test because it is basically a re-implementation of the layer so it doesn't really help anything. If there is a bug then it is in both and if it is in another component then it will be present in both.

Perfect, thanks for the review and taking care of the test!

PR #3385, opened 2026-04-08
Fix SDPA vmap with GQA/MQA shapes (n_heads != n_kv_heads)

### LINKED ISSUE #3383: mx.fast.scaled_dot_product_attention hangs under vmap when n_heads != n_kv_heads (GQA/MQA)
## Summary

`mx.fast.scaled_dot_product_attention` hangs (or occasionally segfaults) when called under `mx.vmap` with grouped-query attention shapes where `n_heads != n_kv_heads`. The same shapes work correctly without `vmap`, and equal-head (MHA) shapes work correctly with `vmap`.

## Reproducer

```python
import mlx.core as mx

D = 64
mx.random.seed(0)

# Shared projections: GQA with H_q=4, H_kv=2
w_q = mx.random.normal((128, 4 * D))
w_k = mx.random.normal((128, 2 * D))
w_v = mx.random.normal((128, 2 * D))
mx.eval(w_q, w_k, w_v)

def f(xi):
    x = xi[None, :]  # (1, L, dim)
    B, L, _ = x.shape
    q = (x @ w_q).reshape(B, L, 4, D).transpose(0, 2, 1, 3)
    k = (x @ w_k).reshape(B, L, 2, D).transpose(0, 2, 1, 3)
    v = (x @ w_v).reshape(B, L, 2, D).transpose(0, 2, 1, 3)
    return mx.mean(mx.fast.scaled_dot_product_attention(q, k, v, scale=D**-0.5))

xb = mx.random.normal((2, 4, 128))
mx.eval(xb)

# This hangs indefinitely (or crashes with SIGSEGV):
out = mx.vmap(f)(xb)
mx.eval(out)
```

The process hangs at the `mx.vmap(f)(xb)` call and never returns. On some runs it crashes with SIGSEGV instead.

## Diagnostic matrix

All tests use the same structure as above, varying only the head counts and the transform:

| H_q | H_kv | Transform    | Result        |
|-----|------|--------------|---------------|
| 4   | 4    | `vmap(fwd)`  | PASS (0.01s)  |
| 4   | 4    | `vmap(grad)` | PASS (0.01s)  |
| 4   | 2    | `grad`       | PASS (instant)|
| 4   | 2    | `vmap(fwd)`  | **HANG / SIGSEGV** |
| 4   | 2    | `vmap(grad)` | **HANG**      |
| 4   | 1    | `vmap(fwd)`  | **HANG**      |

The boundary is precisely `n_heads != n_kv_heads` under `vmap`. Equal-head configurations pass, and non-vmap'd `grad` passes even with GQA shapes.

## Workaround

Replacing the fused SDPA call with decomposed `matmul → softmax → matmul` (with explicit `mx.repeat` for KV head expansion) works correctly under `vmap` and `vmap(grad)` for all head configurations. This confirms the bug is in the `ScaledDotProductAttention` primitive's vmap rule, not in the attention math.

## Context

This blocks using `mx.vmap(mx.grad(...))` for per-sample gradient computation on any GQA/MQA model (Qwen, Llama, Mistral, etc.) — the standard architecture for all recent LLMs. 

## Environment

- MLX: 0.31.1
- Python: 3.12.4
- macOS: 26.3.1 (Tahoe)
- Chip: Apple M1 Pro (16GB)

Brooooooklyn:
## Proposed changes

`ScaledDotProductAttention` relied on `Custom::vmap` which re-vmapped the
fallback lambda. This always took the decomposed matmul-softmax-matmul
path, bypassing the fused Metal/CUDA kernel even when it was available.
On MLX 0.31.1 this also caused SIGSEGV/hang with GQA shapes due to a
since-fixed transforms infrastructure bug.

Add a dedicated vmap override that merges the vmap axis into the batch
dimension and re-invokes `scaled_dot_product_attention` directly, so the
fused kernel is dispatched under vmap just as it is without vmap.
Falls back to `Custom::vmap` when attention sinks are present.

Close https://github.com/ml-explore/mlx/issues/3383

## Checklist

### DIFF
--- a/mlx/fast.cpp
+++ b/mlx/fast.cpp
@@ -50,6 +50,83 @@ std::pair<std::vector<array>, std::vector<int>> Custom::vmap(
   return {outputs, out_axes};
 }
 
+std::pair<std::vector<array>, std::vector<int>> ScaledDotProductAttention::vmap(
+    const std::vector<array>& inputs,
+    const std::vector<int>& axes) {
+  auto s = stream();
+
+  // Sinks require 1-D input; fall back to generic vmap for that case.
+  if (has_sinks_) {
+    return Custom::vmap(inputs, axes);
+  }
+
+  // Determine vmap size from the first mapped input.
+  int vmap_size = -1;
+  for (int i = 0; i < static_cast<int>(axes.size()); ++i) {
+    if (axes[i] != -1) {
+      vmap_size = inputs[i].shape(axes[i]);
+      break;
+    }
+  }
+
+  auto prepare = [&s, vmap_size](const array& x, int ax) -> array {
+    if (ax == -1) {
+      return repeat(expand_dims(x, 0, s), vmap_size, 0, s);
+    }
+    if (ax != 0) {
+      return moveaxis(x, ax, 0, s);
+    }
+    return x;
+  };
+
+  auto q = prepare(inputs[0], axes[0]);
+  auto k = prepare(inputs[1], axes[1]);
+  auto v = prepare(inputs[2], axes[2]);
+
+  // [V, B, H, L, D] -> [V*B, H, L, D]
+  auto merge_batch = [&s, vmap_size](const array& x) -> array {
+    auto shape = x.shape();
+    Shape new_shape = {vmap_size * shape[1]};
+    new_shape.insert(new_shape.end(), shape.begin() + 2, shape.end());
+    return reshape(x, std::move(new_shape), s);
+  };
+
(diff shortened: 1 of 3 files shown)


zcbenz (commented):
I'm closing this PR as #3383 no longer reproduces. Having a native vmap for fast sdpa is nice, but since users can trivially batch it I think this code path likely would not be used, I don't know why they choose to vmap sdpa in the first place to be honest.

Sorry it took so long to review, the devs who are actually familiar with the code could not find time.

PR #3388, opened 2026-04-09
Avoid joining threads on exit

zcbenz:
On Windows there is a bug that [joining threads on exit would end up with deadlock](https://developercommunity.visualstudio.com/t/1654756), and we can't just leak the threads in `CommandEncoder` because they are thread local and it would be real leaks.

So instead of joining the threads in the destructor, we detach the threads and let them delete themselves once the threads finish working.

### DIFF
--- a/mlx/distributed/nccl/nccl.cpp
+++ b/mlx/distributed/nccl/nccl.cpp
@@ -16,6 +16,7 @@
 #include <mutex>
 #include <stdexcept>
 #include <string>
+#include <thread>
 #include <type_traits>
 
 #include "mlx/backend/cuda/device.h"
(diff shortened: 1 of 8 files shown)


mlx/distributed/nccl/nccl.cpp
```diff
@@ -16,6 +16,7 @@
 #include <mutex>
 #include <stdexcept>
 #include <string>
+#include <thread>
```
angeloskath, line 19: Forgotten?

angeloskath (approved):
Looks great, left a tiny comment.

mlx/distributed/nccl/nccl.cpp
zcbenz: This file implicitly included the `thread` via `device.h` => `worker.h`, which this PR removes from `device.h`'s includes.

PR #3391, opened 2026-04-10
fix: There is no Stream(gpu, {}) in current thread

kyr0:
## Proposed changes

This is for quickly discussing the following idea:

```cpp
CommandEncoder& get_command_encoder(Stream s) {
  auto& encoders = get_command_encoders();
  auto it = encoders.find(s.index);
  if (it == encoders.end()) {
    // Auto-register the stream on this thread.  MLX streams are thread-local,
    // but arrays may carry references to streams created on other threads
    // (e.g. module-level streams created at import time).  Rather than
    // throwing, lazily create a CommandEncoder so the current thread can
    // dispatch work for that stream.
    auto& d = device(s.device);
    it = encoders.try_emplace(s.index, d, s.index, d.residency_set()).first;
  }
  return it->second;
}
```

Instead of:

```cpp
CommandEncoder& get_command_encoder(Stream s) {
  auto& encoders = get_command_encoders();
  auto it = encoders.find(s.index);
  if (it == encoders.end()) {
  throw std::runtime_error(
        fmt::format("There is no Stream(gpu, {}) in current thread.", s.index));
  }
  return it->second;
}
```

When MLX operations in any client application are serialized through a single executor, this is safe. If people use it without serialization... it might break anyways? But for all consumers who build sane, multi-threaded code, this will simply work instead of giving them unnecessary headaches? 

I'm wondering what I'm overlooking.

_(I tested this with oMLX which had issues with newer versions of mlx - anything above 0.31.1 in my custom fork - and it works flawlessly)_ 

https://github.com/ml-explore/mlx/issues/3078

## Checklist

### DIFF
--- a/mlx/backend/metal/device.cpp
+++ b/mlx/backend/metal/device.cpp
@@ -799,8 +799,13 @@ CommandEncoder& get_command_encoder(Stream s) {
   auto& encoders = get_command_encoders();
   auto it = encoders.find(s.index);
   if (it == encoders.end()) {
-    throw std::runtime_error(
-        fmt::format("There is no Stream(gpu, {}) in current thread.", s.index));
+    // Auto-register the stream on this thread.  MLX streams are thread-local,
+    // but arrays may carry references to streams created on other threads
+    // (e.g. module-level streams created at import time).  Rather than
+    // throwing, lazily create a CommandEncoder so the current thread can
+    // dispatch work for that stream.
+    auto& d = device(s.device);
+    it = encoders.try_emplace(s.index, d, s.index, d.residency_set()).first;
   }
   return it->second;
 }


zcbenz:
Can you try the new `ThreadLocalStream` class introduced in https://github.com/ml-explore/mlx/pull/3355, which achieves exactly the same thing with this PR.

We can not accept the change you proposed because it gives user false impression they are committing to the same queue, which we might want to support in future.

zcbenz:
We would like to release it soon but users have been finding new problems with the implementation so we don't have a ETA for now, which is actually a good thing so we have less chance of a bad release.

I'm closing this PR as #3408 provides a simpler solution.

kyr0:
@zcbenz All good but I had a good laugh when I checked the "simpler" solution in #3408 ;)))

PR #3395, opened 2026-04-11
Add clear_streams API for cleanup before exit

zcbenz:
Windows horror story: we should not do anything complicated during thread/process teardown (i.e. in the callbacks registered by system `atexit` explicitly via APIs or implicitly via C++ `static`/`thread_local` keywords), joining threads would cause deadlocks, and calling any CUDA API would fail.

The solution proposed here is to just provide an API to let users invoke before thread/process exit:

```cpp
    std::thread t([]() {
      auto s = new_stream(default_device());
      eval(arange(10, s));
      clear_streams();
    });
    t.join();
```

The API call is optional, if users don't use it the program would still work normally on macOS/Linux that everything gets synchronized and cleared implicitly by us, except for Windows where the program would crash on exit, which is actually fine for most cases.

### DIFF
--- a/mlx/backend/cuda/eval.cpp
+++ b/mlx/backend/cuda/eval.cpp
@@ -80,4 +80,8 @@ void synchronize(Stream s) {
   cu::get_command_encoder(s).synchronize();
 }
 
+void clear_streams() {
+  cu::get_command_encoders().clear();
+}
+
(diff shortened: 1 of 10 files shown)


angeloskath (approved):
👍

Issue #3392, opened 2026-04-10
bump nanobind from 2.10.2 to 2.12.0 ?


jrp2014:
Is there a reason that nano bind is pegged at 2.10.2? The GitHub bot want to update my dependencies to the latter.

zcbenz:
No reason just no one bothered to, feel free sending a pull request.

zcbenz:
Closed by https://github.com/ml-explore/mlx/pull/3396.

PR #3396, opened 2026-04-11
Update nanobind version to v2.12.0

jrp2014:
## Proposed changes

See #3392  Some of the examples pull in nanobind 2.10.2.  Not sure whether they should be changed to lower bound constraints or pinned, as they currently are, like here.

## Checklist

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -348,7 +348,7 @@ if(MLX_BUILD_PYTHON_BINDINGS)
   FetchContent_Declare(
     nanobind
     GIT_REPOSITORY https://github.com/wjakob/nanobind.git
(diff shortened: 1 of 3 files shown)


zcbenz (approved):
Thanks!

Having the examples using the pinned version is usually the safe choice.

jrp2014:
Ok. Ready to merge ... @zcbenz  I don't think that I have permission to merge.

jrp2014:
We now seem to be on nanobind 2.14.  The runtime changes—especially integer conversion semantics and reference-count fixes—mean MLX itself should run its complete build/test suite before changing its CMake pin, but it seems to be a mainly bug fix improvement.

PR #3398, opened 2026-04-12
Add register_stream for cross-thread GPU stream usage

TheTom:
## Summary

After #3348 made `CommandEncoder` thread-local, evaluating arrays on a thread that did not create the stream fails because the encoder does not exist on the new thread. `ThreadLocalStream` (#3355) solves this for Python by creating per-thread streams, but there is no way to share a single stream across threads.

`register_stream(Stream)` registers an existing stream's GPU command encoder on the calling thread, allowing `eval()` to succeed. It is a no-op if already registered, and is safe to call from any thread including the creating thread.

This is needed for Swift concurrency where the runtime may hop tasks between threads, and for Python `threading.Thread` workflows that share a stream.

Related: #3078

## Changes

- `mlx/stream.h` / `mlx/stream.cpp`: add `register_stream(Stream)` 
- `python/src/stream.cpp`: nanobind binding
- `python/tests/test_threads.py`: cross-thread eval test, idempotency test

## Test plan

- [x] `python/tests/test_threads.py` passes (3/3)
- [x] C++ library builds clean
- [ ] CI

### DIFF
--- a/mlx/stream.cpp
+++ b/mlx/stream.cpp
@@ -3,6 +3,7 @@
 #include "mlx/stream.h"
 #include "mlx/backend/cpu/device_info.h"
 #include "mlx/backend/gpu/device_info.h"
+#include "mlx/backend/gpu/eval.h"
 #include "mlx/scheduler.h"
 
 #include <array>
@@ -55,6 +56,16 @@ std::vector<Stream> get_streams() {
   auto& [streams, mtx] = all_streams();
   std::shared_lock lock(mtx);
   return streams;
+}
+
+void register_stream(Stream s) {
+  if (s.device == Device::gpu) {
+    if (!gpu::is_available()) {
+      throw std::invalid_argument(
+          "[register_stream] Cannot register gpu stream without gpu backend.");
+    }
+    gpu::new_stream(s);
+  }
 }
 
 Stream new_stream(Device d) {
--- a/mlx/stream.h
+++ b/mlx/stream.h
@@ -30,6 +30,22 @@ MLX_API void set_default_stream(Stream s);
 /** Make a new stream on the given device. */
 MLX_API Stream new_stream(Device d);
 
+/**
+ * Register an existing stream on the calling thread.
+ *
+ * GPU streams use thread-local command encoders. When a stream is created,
+ * its encoder is registered only on the creating thread. If a different
+ * thread later calls eval() on arrays from that stream, it will fail
+ * because the encoder does not exist on the new thread.
+ *
+ * Call this function on any thread that needs to eval() arrays from a
+ * stream that was created on a different thread.
+ *
+ * Safe to call multiple times or on the creating thread (no-op if already
+ * registered).
+ */
+MLX_API void register_stream(Stream s);
+
 /** Get all available streams. */
 MLX_API std::vector<Stream> get_streams();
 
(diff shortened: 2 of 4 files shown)


zcbenz:
Can you check whether #3405 would work for you? We would like to avoid diversities in different languages when possible.

TheTom:
> Can you check whether #3405 would work for you? We would like to avoid diversities in different languages when possible.

ThreadLocalStream doesn't solve the Swift concurrency case. Swift's async/await runtime can resume a task on a different thread at every await point. A ThreadLocalStream gives each thread its own stream, but when the task hops from thread A to thread B mid-execution, thread B's stream is different from thread A's. Arrays created on A's stream can't be eval'd on B's stream.

register_stream solves this because you register one shared stream on all threads that might run the task. The stream identity is stable across thread hops, which is what Swift concurrency requires.

Python threading doesn't have this problem because you control which thread runs what. Swift's runtime decides, and it can change between any two await points. That's why the "same pattern across languages" framing doesn't apply here. The threading models are fundamentally different.

Happy to add a comment on #3405 explaining this if it would help.

zcbenz:
Thanks for the elaboration, so according to [How async/await works internally in Swift](https://swiftrocks.com/how-async-await-works-internally-in-swift) the async/await in Swift dispatches the task to a thread pool. 😮

In this case I think a more decent solution is to provide an API to make a stream thread local, instead of providing a new abstraction or requiring manual registration in every thread:

```python
s = mx.new_stream(mx.gpu)
mx.make_stream_thread_local(s)
```

It should be fairly easy to implement, I will make a PR.

TheTom:
> Thanks for the elaboration, so according to [How async/await works internally in Swift](https://swiftrocks.com/how-async-await-works-internally-in-swift) the async/await in Swift dispatches the task to a thread pool. 😮
> 
> In this case I think a more decent solution is to provide an API to make a stream thread local, instead of providing a new abstraction or requiring manual registration in every thread:
> 
> ```python
> s = mx.new_stream(mx.gpu)
> mx.make_stream_thread_local(s)
> ```
> 
> It should be fairly easy to implement, I will make a PR.

That would work if the semantics are "one stream, lazily register its encoder on any thread that accesses it." That's essentially auto-register_stream, which is cleaner than manual registration.

The key requirement is that arrays created on this stream on thread A can be eval'd on thread B without thread B needing to do anything explicit. One stream identity, shared across all threads, encoder registered lazily on first use per thread.

If "make_stream_thread_local" means per-thread copies (like the current ThreadLocalStream), that breaks for the same reason: arrays from thread A's copy are on a different stream than thread B's copy, and eval on B can't dispatch to A's stream.

Happy to test whatever you put up against our Swift inference pipeline to verify it handles the task-hopping case.

angeloskath:
I will close this since as discussed in #3408 sharing streams but not sharing their command encoders is not a viable option unfortunately.

I will also comment on #3078 but basically the safe pattern for multithreading with MLX is to evaluate graphs created in a thread always in that thread. At least for now.

We may want to consider a "hack" which would be a graph transform that changes the stream of all unevaluated arrays in the graph. For cases where you know it is safe this could be an option although imho it is simpler to make sure that graphs are evaluated between thread changes or simply have a single background thread per stream.

PR #2667, opened 2025-10-13
Modified sort behavior when running CPU or Metal to match NumPy/JAX

Maalvi14:
#### **CPU Backend (`sort.cpp`)**
1. **New `nan_aware_less` function**: Centralized NaN handling logic
2. **Updated all sort operations**: `sort`, `argsort`, `partition`, `argpartition`
3. **Consistent NaN placement**: NaNs always go to the end regardless of operation type
4. **Type safety**: Only applies NaN logic to floating-point types

#### **Metal Backend (`sort.h`)**
1. **New `NanLastLess` comparator template**: GPU-optimized NaN handling
2. **Specialized implementations**: Separate logic for `float` and `half` types
3. **Bitwise operations**: Efficient GPU implementation using `isnan()` checks
4. **Updated kernel defaults**: All sorting kernels now use `NanLastLess` by default

### **Behavioral Changes**
- **Before**: NaN ordering was undefined/inconsistent
- **After**: NaNs are consistently placed at the end of sorted arrays
- **Compatibility**: Matches NumPy's `np.sort()` behavior exactly
- **Performance**: Minimal overhead with efficient NaN detection

### **Implementation Strategy**
- **CPU**: Uses `std::isnan()` for precise NaN detection
- **Metal**: Uses Metal's `isnan()` function with bitwise logic for GPU efficiency
- **Consistency**: Both backends implement identical NaN placement logic
- **Type coverage**: Handles `float32`, `float64`, `float16`, and `bfloat16` types

### **Example Unit Tests**

```
import mlx.core as mx
import numpy as np
import unittest


class TestNaNAwareSorting(unittest.TestCase):
    """Test that MLX sort places NaNs at the end like NumPy."""
    
    def setUp(self):
        """Store original default device."""
        self.original_device = mx.default_device()
    
    def tearDown(self):
        """Restore original default device."""
        mx.set_default_device(self.original_device)
    
    def _test_sort_matches_numpy(self, data, device):
        """Helper to test that MLX sort matches NumPy behavior."""
        mx.set_default_device(device)
        
        # NumPy reference
        np_data = np.array(data, dtype=np.float32)
        np_sorted = np.sort(np_data)
        
        # MLX result
        mx_data = mx.array(data)
        mx_sorted = mx.sort(mx_data)
        
        # Compare - both NaNs and non-NaNs should be in same positions
        np_result = np.array(mx_sorted)
        
        # Check non-NaN values match
        np_mask = ~np.isnan(np_sorted)
[...]

Issue #2057, opened 2025-04-10
Document the stability and algorithm of sort


zcbenz:
The https://github.com/ml-explore/mlx/pull/844#issuecomment-2005390268 indicates sort in MLX is stable, I think the stability and probably the algorithm should be explicitly documented.

NeuralNoble:
Hi! I'd like to take this on as my first contribution to MLX.

From reading the source, it looks like the sort is stable across all backends (`std::stable_sort` on CPU, merge sort on Metal/CUDA), and NaN values are placed at the end to match NumPy/JAX behavior. I'll update the docstrings for `sort` and `argsort` in the Python bindings to document this.

Please let me know if there's anything else you'd like covered!

Maalvi14:
@zcbenz I believe this issue could be closed, Neural already completed it, and thanks to neural for documenting! :)

PR #3400, opened 2026-04-12
Document sort stability and NaN handling

NeuralNoble:
## Summary

Addresses #2057.

Documents that `sort` and `argsort` are **stable** across all backends and place **NaN values at the end** to match NumPy/JAX.

- CPU: `std::stable_sort` with a NaN-aware comparator (`mlx/backend/cpu/sort.cpp`)
- Metal/CUDA: merge sort with the same NaN handling (`mlx/backend/metal/kernels/sort.h`, `mlx/backend/cuda/sort.cu`)

Both guarantees are true in the current implementation (NaN handling introduced in #2667) but weren't documented, so users couldn't rely on them.

## Changes

- `python/src/ops.cpp` — add note to `sort` and `argsort` docstrings
- `mlx/ops.h` — add note to the four C++ declarations
- `mlx/ops.cpp` — add note to the four C++ definitions

## Test plan

- [x] `pre-commit run --all-files` passes
- [x] C++ build succeeds, `ctest` passes
- [x] `python -m pytest python/tests/test_ops.py -k sort` passes
- [x] `help(mx.sort)` renders the new docstring at runtime

### DIFF
--- a/mlx/ops.cpp
+++ b/mlx/ops.cpp
@@ -2546,13 +2546,15 @@ array blackman(int M, StreamOrDevice s /* = {} */) {
   return add(subtract(alpha, term1, s), term2, s);
 }
 
-/** Returns a sorted copy of the flattened array. */
+/** Returns a sorted copy of the flattened array.
+ * The sort is stable and NaN values are placed at the end. */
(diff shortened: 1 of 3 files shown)


angeloskath (approved):
Thanks.

PR #3405, opened 2026-04-13
ThreadLocalStream in C++

zcbenz:
Refs #3398, #3078.

Add the C++ APIs for `ThreadLocalStream` introduced by https://github.com/ml-explore/mlx/pull/3355, to ease the porting for language bindings. A new `new_thread_local_stream` API is added in python to replace the `ThreadLocalStream` constructor.

### DIFF
--- a/mlx/stream.cpp
+++ b/mlx/stream.cpp
@@ -6,6 +6,7 @@
 #include "mlx/scheduler.h"
 
 #include <array>
+#include <map>
 #include <optional>
 #include <shared_mutex>
 
@@ -26,6 +27,11 @@ auto& default_stream_storage(Device d) {
 
 auto& all_streams() {
   static std::tuple<std::vector<Stream>, std::shared_mutex> streams_and_mtx;
+  return streams_and_mtx;
+}
+
+auto& thread_local_streams() {
+  static std::tuple<std::vector<ThreadLocalStream>, std::mutex> streams_and_mtx;
   return streams_and_mtx;
 }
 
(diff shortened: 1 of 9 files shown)


zcbenz:
Closing in favor of https://github.com/ml-explore/mlx/pull/3408.

mlx/stream.cpp
```diff
@@ -73,4 +75,20 @@ Stream new_stream(Device d) {
   return s;
 }
 
+ThreadLocalStream new_thread_local_stream(Device d) {
+  auto& [streams, mtx] = thread_local_streams();
+  int index = streams.size();
+  std::lock_guard lock(mtx);
```
angeloskath, line 80: The lock guard should probably be above grabbing the index.

python/src/stream.cpp
```diff
@@ -86,29 +61,28 @@ void init_stream(nb::module_& m) {
...
       )pbdoc")
-      .def_ro("device", &PyThreadLocalStream::device)
-      .def(nb::init<mx::Device>())
+      .def_ro("device", &mx::ThreadLocalStream::device)
       .def(
           "__repr__",
-          [](const PyThreadLocalStream& s) {
+          [](const mx::ThreadLocalStream& s) {
             std::ostringstream os;
             os << "ThreadLocalStream(" << s.device << ")";
```
angeloskath: Nitpick but I would probably change this to `os << "ThreadLocalStream(" << s.device << ", " << s.index << ")";`

I know I had it without the index before but that is because it actually had a callable constructor and traditionally one is supposed to have `repr` return valid code. Here I think it makes sense to follow `mx::Stream` and return the full info.

angeloskath (approved):
Nice, yeah having it only in python was a bad choice.

I left two comments one I think is a bug the other is absolutely minor.

PR #3408, opened 2026-04-13
Register existing streams as thread local

zcbenz:
Add a `make_stream_thread_local` API that for each thread using the stream a new `CommandEncoder` would always be created.

This API replaces the `ThreadLocalStream` API introduced by https://github.com/ml-explore/mlx/pull/3355, to address the problem raised in https://github.com/ml-explore/mlx/pull/3398 by @TheTom that, evaluating an array created with `ThreadLocalStream` in another thread would still end up with error.  

Refs https://github.com/ml-explore/mlx/issues/3078.


angeloskath:
I am not quite sure I understand this change.

So having a stream that isn't made thread local results in an error if we try to use it from a different thread. However a thread local stream implicitly creates a new command encoder and then just works. Why wouldn't we make every stream thread local in that case?

I also think that this is both great and bad at the same time. Streams are meant as a sequence of commands. If I perform two async_evals in a specific stream I expect them to be computed one after the other, now this implicitly depends on the thread I am in.

This isn't simply a quirk either. The eval uses this fact to synchronize across streams and across evals.

zcbenz:
Yeah it would indeed be mess when taking account of eval, I didn't think that far. I'm reopening #3405 instead.

So for the case in https://github.com/ml-explore/mlx/pull/3398#issuecomment-4238383335 that do eval in different threads, I wonder if it makes sense providing a `make_stream_thread_unsafe` API that stores the stream globally and requires the user to use locks themselves in a multi-thread environment.

Issue #3409, opened 2026-04-14
SafeTensors loader does not validate data_offsets against file size


matinsaurralde:
## Description

The SafeTensors loader in `mlx/io/safetensors.cpp` reads `data_offsets` from the JSON metadata but does not validate that these offsets fall within the actual file boundaries.

A crafted `.safetensors` file can declare `data_offsets` that are consistent with the tensor's declared `shape` and `dtype`, while the actual file contains far less data. When the tensor is lazily evaluated, this results in an out-of-bounds memory read.

This is a distinct issue from the shape/offset mismatch addressed in #3364. That PR validates internal consistency (shape vs data_offsets), while this issue is about validating data_offsets against the actual file size — matching the `MetadataIncompleteBuffer` check in the [Rust reference implementation](https://github.com/huggingface/safetensors).

## Reproduction

```python
# Given a crafted safetensors file where:
# - JSON header declares shape=[1000,1000], dtype=F32 (needs 4MB)
# - data_offsets=[0, 4000000]
# - Actual file only contains 4 bytes of tensor data

import mlx.core as mx
arrays, _ = mx.load("crafted.safetensors", return_metadata=True)
mx.eval(arrays["tensor"])  # reads beyond file boundary
```

On macOS, the OS may zero-fill out-of-bounds reads from mmap'd files, masking the bug. This is not guaranteed behavior.

## Additional issues found

- `ParallelFileReader::seek()` and `FileWriter::seek()` in `mlx/io/load.h` do not handle `std::ios_base::end`, silently falling through to `SEEK_CUR` behavior. Also uses a hardcoded `0` instead of `SEEK_SET`.

## Affected files

- `mlx/io/safetensors.cpp` — missing file boundary validation
- `mlx/io/load.h` — `seek()` does not support `SEEK_END`

angeloskath:
Fixed in #3410.

PR #3410, opened 2026-04-14
Validate safetensors data offsets against file boundaries

matinsaurralde:
## Proposed changes

Addresses #3409.

The SafeTensors loader reads `data_offsets` from JSON metadata without checking whether they fall within the actual file. A crafted `.safetensors` file can declare offsets that exceed the file size, causing an out-of-bounds memory read when the tensor is evaluated.

This is the file-boundary counterpart to #3364 (which validates internal consistency between `data_offsets` and `shape * dtype`). The Rust reference implementation enforces both checks (`TensorInvalidInfo` and `MetadataIncompleteBuffer`).

### Changes

**`mlx/io/safetensors.cpp`** — After parsing the JSON header, compute the available data size from the actual file size and validate each tensor's `data_offsets`:
- `data_offsets` has at least 2 elements
- `begin <= end`
- `end` does not exceed the data region of the file

**`mlx/io/load.h`** — `ParallelFileReader::seek()` and `FileWriter::seek()` did not handle `std::ios_base::end`, silently falling through to `SEEK_CUR`. Added `SEEK_END` support and replaced the hardcoded `0` with `SEEK_SET`.

**`tests/load_tests.cpp`** — Added tests for:
- File with `data_offsets` exceeding file boundary (must throw)
- File with reversed `data_offsets` (must throw)
- Valid file still loads correctly (regression)

## Checklist
- [x] I have run `pre-commit run --all-files` to format my code


angeloskath (approved):
Thanks for the PR. The original PR did not actually pass the test that it was supposed to be made for, so I would advise compiling and running the tests before PRing.

It is now fixed and the tests unified.

PR #3412, opened 2026-04-14
Jaccl refactor

angeloskath:
This PR does a few things:

1. Refactors JACCL to a standalone C++ library installable with a few lines of CMake
   a. Adds README and examples
   b. Adds a NCCL style benchmark
2. Changes the mesh all reduce to a deterministic one
   a. We pay a small BW penalty and even smaller latency penalty but it does make things more predictable and stable. Eg people don't have to make sure that they always sample the same token or deal with diverging KV caches etc
3. A few more robustness improvements like doing a double `all_gather` in the side-channel barrier to ensure no node can run ahead.
4. Fixes #3389 with a more general solution that checks for IPv4 and also checks all entries.
5. Left the best for last, compiles the reductions for `armv8.6-a` and chooses it at runtime if available, which results in much faster reduction for bf16.
   a. Specifically we now reach 12.2 GB/s algorithm bandwidth for bf16 up from ~9 GB/s for a 256 MB all reduce


zcbenz (approved):
👍

PR #3413, opened 2026-04-15
Fixes for CUDA CI

zcbenz:
* Force using "build<=1.4.2", the 1.4.3 is breaking release and I will look into it later.
* Disable the `test_sdpa_long_masked_sequence` test in CUDA CI which fails with OOM in GPU memory.
* Disable the `test_conv_1d_with_2d` test in CUDA CI which has been a flaky test for a long time, I suspect it is hitting a bad cuDNN kernel but I don't have access to the old hardware used by CI to test.


angeloskath (approved):
Thanks!

PR #3414, opened 2026-04-15
[CUDA] gather_mm

zcbenz:
Implement `gather_mm` by fusing CUTLASS GEMM with a custom prologue.

Note that this implementation is completely unoptimized (e.g. no tensor core, no vectorized read) and only performs 0.3x of cuBLAS, we would need to write our own heuristics to dispatch to proper main loops and alignments, which I'm postponing a bit until we have more CUTLASS kernels that I can make a general design.


angeloskath (approved):
Looks great. Should be fairly easy to extend to at least Ampere tensor cores.

Issue #3335, opened 2026-03-30
Add support for computing matrix determinants


jessegrabowski:
The `mlx.core.linalg` sub-module doesn't have any functions related to computation of determinants. These pop up all over the place, as part of the lop of a multivariate normal in PPL land, but also when compute change-of-variable correction terms in normalizing flow layers. Would be nice to have at least `mx.linalg.det`. A numerically stable `mx.linalg.slogdet` is also important.

abhilashreddys:
@zcbenz I'd like to take this on. I plan to implement both slogdet and det using the existing lu_factor decomposition.

Lucas-Fernandes-Martins:
Hi @abhilashreddys I've been working on the same thing for the past few days. If you want to take a look at my [PR](https://github.com/ml-explore/mlx/pull/3415) maybe we can collaborate on this one. For now, I have only implemented `linalg.det` so there's still slogdet to be done. Also if you have any suggestions on my implementation I'd love to hear your thoughs :)

abhilashreddys:
Hi @zcbenz @Lucas-Fernandes-Martins I was also working on it for the past few days. I have implemented mx.linalg.det and mx.linalg.slogdet as composites on the existing lu_factor(). slogdet is the core implementation, and det is derived from it via sign * exp(logabsdet), so it benefits from the same numerical stability. Also includes small-matrix fast paths (1x1/2x2/3x3) for batched workloads and full test coverage against NumPy.

PR: https://github.com/ml-explore/mlx/pull/3416

Let me know your thoughts!

Lucas-Fernandes-Martins:
@abhilashreddys nice, we can bounce off each other then :) 

I'm testing your implementation and it doesn't seem to work for singular matrices which have 4 or more dimensions. I believe that's because there's some error handling on `[mlx/backend/cpu/luf.cpp](https://github.com/ml-explore/mlx/pull/3415/changes#diff-e4ba8f7bef27f49a7d15aef13ba877301098fdebffbf015af2998e3641dfb1ab)` that throws an error if the matrix is singular. That's why I introduced the `allow_singular` boolean flag on `lu_helper`.

Let me know your opinion on this

abhilashreddys:
[...]

PR #3415, opened 2026-04-15
Adding linalg.det function for determinant calculation 

Lucas-Fernandes-Martins:
## Proposed changes

Hi, tentative implemention of linalg.det for determinant calculation, using LU decomposition through lu_helper. Also added a test file on [`python/tests/test_det.py`](https://github.com/Lucas-Fernandes-Martins/mlx/blob/main/python/tests/test_det.py). Idea was proposed in #3335.

Super happy to iterate on feedback! If maintainers think this is relevant, I'd also be happy to work on implementing mx.linalg.slogdet.

To be able to use `lu_helper`, I decided to add a `allow_singular` parameter to the function so it can also be used for determinant calculation for singular matrices (by default, lu_helper throws an error if the matrix is singular). Happy to hear your thoughts on this!

## Checklist

### DIFF
--- a/mlx/backend/cpu/luf.cpp
+++ b/mlx/backend/cpu/luf.cpp
@@ -16,7 +16,8 @@ void luf_impl(
     array& lu,
     array& pivots,
     array& row_indices,
-    Stream stream) {
+    Stream stream,
+    bool allow_singular /* = false */) {
   int M = a.shape(-2);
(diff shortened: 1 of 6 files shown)


zcbenz:
I'm closing this in favor #3416 per https://github.com/ml-explore/mlx/issues/3335#issuecomment-4323622567.

PR #3416, opened 2026-04-15
Add determinant and sign-log-determinant functions to mlx.core.linalg

abhilashreddys:
## Proposed changes: Add `linalg.det` and `linalg.slogdet`

Related: https://github.com/ml-explore/mlx/issues/3335

Add determinant (`det`) and sign-log-determinant (`slogdet`) functions to `mlx.core.linalg`.

- `mx.linalg.det(a)` - returns the determinant of a square matrix
- `mx.linalg.slogdet(a)` - returns `(sign, logabsdet)` for numerical stability

### Composite on `lu_factor()`, no new primitive

Both functions compose on the existing `lu_factor()` primitive rather than introducing a new one. This mirrors how JAX implements `slogdet` and is consistent with how `solve()` already builds on `lu()` in this codebase. 

### `slogdet` as the core, `det` derived from it

`slogdet` is the numerically stable. For the general case (N > 3), `det` simply calls `slogdet` and reconstructs via `sign * exp(logabsdet)`. This avoids duplicating the LU-based logic. An internal `slogdet_impl` function handles the shared logic so the public `det` and `slogdet` entry points each validate and promote types exactly once.

### Small-matrix fast paths (1x1, 2x2, 3x3)

A shared `det_raw_small()` helper computes determinants directly using closed-form formulas (element extraction via slice/squeeze, then the standard cofactor expansion for 3x3). This avoids LU overhead for the small matrices common in batched ML workloads (e.g., 3x3 covariance matrices in normalizing flows).

`det` calls `det_raw_small` directly for N <= 3, skipping the log/exp round-trip entirely. `slogdet` calls it too, then splits the raw result into `sign(raw)` and `log(abs(raw))`.

### Integer auto-promotion

`det`/`slogdet` promote integer inputs to float32 via `at_least_float()`. This matches NumPy's behavior (`np.linalg.det` accepts integer arrays).

### Singular matrix handling

For the LU path, singularity is detected by checking for exact zeros on the U diagonal (same approach as NumPy). Singular matrices return `(sign=0, logabsdet=-inf)` from `slogdet`, and `0.0` from `det`.

Thanks @Lucas-Fernandes-Martins for pointing out the singular matrix [issue](https://github.com/ml-explore/mlx/issues/3335#issuecomment-4255476950) in LU factorization. 

## Testing

- **Correctness**: Compared against `np.linalg.det` / `np.linalg.slogdet` for 1x1 through 5x5, random matrices, identity
- **Batched**: Shapes `(3, 4, 4)` and `(2, 3, 3, 3)` match NumPy
- **Numerical stability**: `0.1 * I_100` (det underflows in float32, slogdet stays finite)
- **Singular matrices**: `sign=0`, `logabsdet=-inf`
- **Consistency**: `sign * exp(logabsdet)` matches `det` for non-singular inputs
- **Dtypes**: float32, float64, integer auto-promotion
- **Error handling**: non-square and 1D inputs raise `ValueError`
- **Full suite**: All 18 existing linalg tests continue to pass (186 subtests)

## Checklist

### DIFF
--- a/mlx/linalg.cpp
+++ b/mlx/linalg.cpp
@@ -721,8 +721,7 @@ void validate_det(
   }
 
   if (a.shape(-1) != a.shape(-2)) {
-    throw std::invalid_argument(
-        fname + " Only defined for square matrices.");
+    throw std::invalid_argument(fname + " Only defined for square matrices.");
   }
 }
 
@@ -748,16 +747,24 @@ array det_raw_small(const array& a, StreamOrDevice s) {
         multiply(elem(0, 1), elem(1, 0), s),
         s);
   } else {
-    // 3x3: a00*(a11*a22 - a12*a21) - a01*(a10*a22 - a12*a20) + a02*(a10*a21 - a11*a20)
+    // 3x3: a00*(a11*a22 - a12*a21) - a01*(a10*a22 - a12*a20) + a02*(a10*a21 -
+    // a11*a20)
     auto a00 = elem(0, 0), a01 = elem(0, 1), a02 = elem(0, 2);
     auto a10 = elem(1, 0), a11 = elem(1, 1), a12 = elem(1, 2);
     auto a20 = elem(2, 0), a21 = elem(2, 1), a22 = elem(2, 2);
     return add(
         subtract(
-            multiply(a00, subtract(multiply(a11, a22, s), multiply(a12, a21, s), s), s),
-            multiply(a01, subtract(multiply(a10, a22, s), multiply(a12, a20, s), s), s),
+            multiply(
+                a00,
+                subtract(multiply(a11, a22, s), multiply(a12, a21, s), s),
+                s),
+            multiply(
+                a01,
+                subtract(multiply(a10, a22, s), multiply(a12, a20, s), s),
+                s),
             s),
-        multiply(a02, subtract(multiply(a10, a21, s), multiply(a11, a20, s), s), s),
+        multiply(
+            a02, subtract(multiply(a10, a21, s), multiply(a11, a20, s), s), s),
         s);
   }
 }
@@ -785,13 +792,19 @@ std::pair<array, array> slogdet_impl(const array& input, StreamOrDevice s) {
   int k = std::min(input.shape(-2), input.shape(-1));
   auto iota = arange(0, k, uint32, s);
   auto parity = astype(
-      sum(not_equal(pivots, iota, s), /* axis = */ -1, /* keepdims = */ false, s),
+      sum(not_equal(pivots, iota, s),
+          /* axis = */ -1,
+          /* keepdims = */ false,
+          s),
       int32,
       s);
 
   // Count negative diagonal elements
   auto num_neg = astype(
-      sum(less(diag, array(0.0f, dtype), s), /* axis = */ -1, /* keepdims = */ false, s),
+      sum(less(diag, array(0.0f, dtype), s),
+          /* axis = */ -1,
+          /* keepdims = */ false,
+          s),
       int32,
       s);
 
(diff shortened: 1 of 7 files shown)


mlx/linalg.cpp
```diff
@@ -705,4 +705,162 @@ array solve_triangular(
   return matmul(a_inv, b, s);
 }
 
-} // namespace mlx::core::linalg
\ No newline at end of file
+void validate_det(
+    const array& a,
+    const StreamOrDevice& stream,
+    const std::string& fname) {
+  check_cpu_stream(stream, fname);
```
zcbenz, line 712: No need to check cpu stream since `lu_factor` already does it, and there are code paths not requiring `lu_factor`.

zcbenz (approved):
Looks good to me, thanks!

Can you rebase the branch to fix the flaky test?

zcbenz:
Also ping @abeleinin who wrote https://github.com/ml-explore/mlx/pull/1451 for a review if possible.

mlx/linalg.cpp
abhilashreddys: Fixed it and also did the rebase.

abeleinin: @zcbenz Maybe `validate_det` should have a more explicit throw since it supports GPU for smaller sizes (unlike the other linalg ops):
```
[linalg::det] GPU support is currently only available for matrices up to 3x3.
```
Also, this functionality should be noted in the python doc string.

mlx/linalg.cpp
```diff
@@ -705,4 +705,157 @@ array solve_triangular(
   return matmul(a_inv, b, s);
 }
 
-} // namespace mlx::core::linalg
\ No newline at end of file
+void validate_det(const array& a, const std::string& fname) {
```
abeleinin: Add `check_float(...)` inside this validate since `mx.complex` isn't supported in `n > 3` case

python/tests/test_linalg.py
```diff
@@ -616,6 +629,210 @@ def test_solve_triangular(self):
         expected = np.linalg.solve(a, b)
         self.assertTrue(np.allclose(result, expected))
 
+    def test_det(self):
```
abeleinin, line 632: Add a test case for empty matrix `mx.linalg.det(mx.zeros((0, 0)))`

Numpy returns `np.linalg.det(np.zeros((0, 0))) = 1.0` and `np.linalg.slogdet(np.zeros((0, 0))) = (1.0, 0.0)`, which is just the empty product.

Right now `validate_det` will accept this square matrices, but `det_raw_small` falls into the 3x3 branch for `n == 0`. We probably need an explicit `n == 0` case in `det_raw_small`.

abeleinin (commented):
PR looks good! Added some comments about edge cases. @zcbenz Feel free to veto any of my suggestions.

mlx/linalg.cpp
zcbenz: This is a very good point, but I think it would be weird if an op defaults to GPU for some shapes but to CPU otherwise, we should probably just constrain to CPU stream only for simplicity, i.e. bringing the check back. Sorry for bouncing back and forth!

mlx/linalg.cpp
abhilashreddys: Added, using `issubdtype(..., complexfloating)` so int auto-promotion still works. Added regression tests for both the small-matrix and LU paths.

mlx/linalg.cpp
abhilashreddys: Done, moved check_cpu_stream back into `validate_det`. By going CPU-only across all sizes, the explicit GPU error and the docstring note aren't needed, since the behavior is now uniform with the rest of `mlx::linalg`.

python/tests/test_linalg.py
abhilashreddys: added n == 0 short-circuit in `det_raw_small` returning 1.0 (empty product) broadcast to the batch shape. This fixes both `det` and `slogdet` since `slogdet_impl` delegates to `det_raw_small` for n <= 3. 

Added tests for scalar and batched empty inputs.

PR #3421, opened 2026-04-17
Fix flaky TestVmap.test_vmap_masked_scatter

zcbenz:
This test is [flaky in CI](https://github.com/ml-explore/mlx/actions/runs/24481559921/job/71756465584) and luckily I can reproduce it locally:

```
  FAIL: test_vmap_masked_scatter (test_vmap.TestVmap.test_vmap_masked_scatter)
    File "/home/runner/work/mlx/mlx/python/tests/test_vmap.py", line 900, in test_vmap_masked_scatter
      self.assertTrue(mx.array_equal(expected, out))
      ~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  
```

### DIFF
--- a/mlx/backend/cuda/indexing.cpp
+++ b/mlx/backend/cuda/indexing.cpp
@@ -458,7 +458,7 @@ void MaskedScatter::eval_gpu(const std::vector<array>& inputs, array& out) {
   }
 
   array mask_flat = flatten_in_eval(mask, 1, -1, s);
-  if (mask_flat.data<void>() != mask.data<void>()) {
+  if (gpu_ptr<void>(mask_flat) != gpu_ptr<void>(mask)) {
     encoder.add_temporary(mask_flat);
   }
   if (!mask_flat.flags().row_contiguous) {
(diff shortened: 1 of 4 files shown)


mlx/backend/cuda/indexing.cpp
```diff
@@ -458,9 +458,8 @@ void MaskedScatter::eval_gpu(const std::vector<array>& inputs, array& out) {
   }
 
   array mask_flat = flatten_in_eval(mask, 1, -1, s);
-  if (mask_flat.data<void>() != mask.data<void>()) {
-    encoder.add_temporary(mask_flat);
-  }
+  encoder.add_temporary(mask_flat);
```
angeloskath: Hmm the previous code is actually correct in the sense that we should only add it as temporary if a copy was done and the pointer changed.

zcbenz: Yeah I got the cause wrong, it turns out to be the implicit copy in `data()`..

angeloskath (approved):
Nice! That makes a lot of sense and it is a sneaky one hidden in plain sight.

PR #3424, opened 2026-04-18
Generate qmm implementaions with cmake

zcbenz:
Instead of manually iterating all combinations of template args in source files, let cmake generate the files for us. This enables us to split the compilation units into smaller ones for faster compilation.


angeloskath (approved):
👍

PR #3426, opened 2026-04-19
Gate NAX JIT sources by macOS SDK version

ausimian:
## Proposed changes

When `MLX_METAL_JIT=ON`, `backend/metal/CMakeLists.txt` unconditionally invokes `make_jit_source()` on the NAX kernel sources. Those headers transitively `#include <MetalPerformancePrimitives/MetalPerformancePrimitives.h>`, which only ships with the macOS 26.2+ SDK, so JIT builds currently fail on older SDKs (e.g. Xcode 15.x on GitHub-hosted `macos-14` runners):

```
fatal error: 'MetalPerformancePrimitives/MetalPerformancePrimitives.h' file not found
  included from vendor/mlx/mlx/backend/metal/kernels/steel/gemm/nax.h:12
```

The AOT branch in `kernels/CMakeLists.txt` already gates NAX by the same SDK check (`MLX_METAL_VERSION GREATER_EQUAL 400 AND MACOS_SDK_VERSION GREATER_EQUAL 26.2`). This PR mirrors that gate in the JIT branch:

- When the SDK gate passes: no behavioural change — NAX JIT sources are emitted as before.
- When it fails: define `MLX_METAL_NO_NAX` (so `is_nax_available()` returns `false` at runtime) and emit a tiny generated `nax_stubs.cpp` with empty-string stubs for the NAX JIT source providers (`metal::gemm_nax`, `metal::quantized_nax`, …). The stubs are needed only so `jit_kernels.cpp` still links; they are never exercised at runtime because the NAX code paths are guarded by `is_nax_available()`.

On older SDKs, `MLX_METAL_JIT=ON` now produces the same no-NAX binary that `MLX_METAL_JIT=OFF` already does — it just uses the JIT path for the non-NAX kernels.

### Context

Discovered while plumbing `MLX_METAL_JIT=ON` through a downstream consumer's CI (GitHub-hosted `macos-14` / Xcode 15.x). MLX's own CI builds JIT on a self-hosted runner with Xcode 26, so the upstream CI does not currently exercise the JIT path on older SDKs.

### Tested

- Local build on macOS 26.4 / Xcode 26 — SDK gate passes, all NAX JIT sources emitted, behaviour identical to current main.
- Build + run the downstream consumer's full test suite with `MLX_METAL_JIT=ON` on GitHub-hosted `macos-14` — previously broken at the `nax.h` preprocess step, now links and passes after this patch.

## Checklist

- [x] I have run `pre-commit run --all-files` to format my code / installed pre-commit prior to committing changes — CMake-only change; happy to run `cmake-format` if the maintainers prefer a specific output.
- [x] I have added tests that prove my fix is effective or that my feature works — no test changes; this is a build-system guard mirroring an existing one.
- [x] I have updated the necessary documentation (if needed) — no public doc change.

### DIFF
--- a/mlx/backend/metal/CMakeLists.txt
+++ b/mlx/backend/metal/CMakeLists.txt
@@ -83,19 +83,45 @@ if(MLX_METAL_JIT)
 
   make_jit_source(steel/attn/kernels/steel_attention)
 
-  make_jit_source(
-    steel/gemm/gemm_nax kernels/steel/utils.h kernels/steel/gemm/nax.h
-    kernels/steel/gemm/params.h kernels/steel/gemm/transforms.h)
-  make_jit_source(steel/gemm/kernels/steel_gemm_fused_nax)
-  make_jit_source(steel/gemm/kernels/steel_gemm_gather_nax)
-  make_jit_source(steel/gemm/kernels/steel_gemm_splitk_nax)
-  make_jit_source(steel/gemm/kernels/steel_gemm_segmented_nax)
+  # NAX JIT sources transitively #include
+  # <MetalPerformancePrimitives/MetalPerformancePrimitives.h>, which only ships
+  # with the macOS 26.2+ SDK. The AOT branch in kernels/CMakeLists.txt already
+  # gates NAX by the same SDK check; mirror it here so JIT=ON can also be built
+  # on older SDKs. When gated out, MLX_METAL_NO_NAX is defined so
+  # is_nax_available() returns false at runtime, and we emit empty-string stubs
+  # for the NAX JIT source providers so jit_kernels.cpp still links.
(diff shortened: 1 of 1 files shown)


zcbenz (commented):
This only works when you build with old macOS SDK, if you build with the newer SDK and ship the binaries to clients and I think it would still fail with the same error?

I think the correct fix would be making `is_nax_available` also check the SDK version during runtime?

ausimian:
I'm happy to for this to be closed. My project vendors mlx and I actually applied this patch as part of my own build (it seems to work fine in CI on macOS 14).

But I'm just going to require for now that consumers of my library use Xcode 26.2 if they want to enable jit.

PR #3427, opened 2026-04-19
Fix kernel cache collision in Compiled constructor

### LINKED ISSUE #3338: [BUG] mx.compile produces wrong results when chaining unsigned->signed astype with matmul
**Describe the bug**

`mx.compile` produces incorrect results when a function chains `astype` from an unsigned integer type to a signed integer type, followed by `astype` to float and then `matmul` (or `einsum`). Without `compile`, the results are correct.

It appears that `compile` fuses the two `astype` calls (e.g., `uint16→int32→float32`) into a single cast that misinterprets the underlying bytes, similar to the root cause of #2475 but in the compile/matmul path rather than reduce ops.

**To Reproduce**

```python
import mlx.core as mx

def f(x_u16, y_i8):
    a = x_u16.astype(mx.int32)
    b = y_i8.astype(mx.int32)
    af = a.astype(mx.float32)
    bf = b.astype(mx.float32)
    return mx.matmul(bf.reshape(2, 1), af.reshape(1, 2))

x = mx.array([10, 20], dtype=mx.uint16)
y = mx.array([1, 2], dtype=mx.int8)

# Without compile: correct
print("No compile:", f(x, y))
# array([[10, 20],
#        [20, 40]], dtype=float32)

# With compile: wrong
cf = mx.compile(f)
print("Compiled:  ", cf(x, y))
# array([[5130, 10260],
#        [0, 0]], dtype=float32)
```

The value `5130 = 0x140A` is `[0x0A, 0x14]` — the raw little-endian bytes of `uint16` values `10` and `20` packed into one `int32`, confirming the byte-level misinterpretation.

**Affected conversions**

The bug triggers for any unsigned→signed integer chain that feeds into `matmul`/`einsum` under `compile`:

| Source | Intermediate | Compiled result for `[10, 20, 30]` |
|---|---|---|
| `uint8` | `int32` → `float32` | Correct ✓ |
| `uint16` | `int32` → `float32` | Wrong ✗ |
| `uint32` | `int32` → `float32` | Wrong ✗ |

`uint8` works because it's 1-byte aligned; wider unsigned types misread.

Note: The chained `astype` alone (without `matmul`) produces the correct result under `compile`. The bug only manifests when the cast result is consumed by `matmul` or `einsum`.

**Expected behavior**

Compiled and non-compiled results should match: `[[10, 20], [20, 40]]`.

**Desktop**
- macOS 15.5, Apple M2 Max
- MLX 0.31.1

dogukanveziroglu:
The constructor moves `inputs` and `constant_ids` into members but the loop building `kernel_lib_` continued to iterate the moved-from parameters. After the move those parameters are empty, so the kernel name omits the per-input dtype suffix and a kernel compiled for one input type can be incorrectly reused for another. With ops fused around the cast (e.g. chained `astype` feeding `matmul` or `einsum`), this surfaces as byte-level corruption of the output.

Reproducer:

```python
import mlx.core as mx

def f(x_u16, y_i8):
    a = x_u16.astype(mx.int32)
    b = y_i8.astype(mx.int32)
    af = a.astype(mx.float32)
    bf = b.astype(mx.float32)
    return mx.matmul(bf.reshape(2, 1), af.reshape(1, 2))

x = mx.array([10, 20], dtype=mx.uint16)
y = mx.array([1, 2], dtype=mx.int8)
print(f(x, y))                 # [[10, 20], [20, 40]]
print(mx.compile(f)(x, y))     # [[5130, 10260], [0, 0]]  <-- wrong
```

`5130 = 0x140A` is the little-endian byte pattern of the two `uint16` inputs packed into one `int32`, confirming the kernel was specialized for the wrong dtype.

Switching the loop to use the `inputs_` / `constant_ids_` members restores the dtype suffix and the bug goes away.

Verified locally against the issue reproducer, exhaustive dtype matrix, and fresh-process isolation tests (44/44 + 9/9 pass). I can add an inline regression test if you guys would like a specific format / location. 

Fixes #3338.

## Checklist

### DIFF
--- a/mlx/compile.cpp
+++ b/mlx/compile.cpp
@@ -137,8 +137,9 @@ Compiled::Compiled(
     }
   }
   os << "_";
-  for (const auto& x : inputs) {
-    if (constant_ids.find(x.id()) != constant_ids.end()) {
+  // Iterate the moved-into members; the parameters are moved-from above.
+  for (const auto& x : inputs_) {
+    if (constant_ids_.find(x.id()) != constant_ids_.end()) {
       continue;
     }
     os << kindof(x.dtype()) << x.itemsize();


zcbenz (approved):
Great found, thanks for fixing this!

PR #3428, opened 2026-04-19
Throw meaningful error when Metal device is not found

dogukanveziroglu:
## Proposed changes                                                                                                                                                                                                                                                
                                               
  In headless, sandboxed, or virtualized macOS sessions (PyInstaller                                                                                                                                                                                                 
  bundles, CI containers, automation runs) `MTL::CopyAllDevices()` returns
  an empty NSArray. The current code immediately calls `devices->object(0)`,                                                                                                                                                                                         
  which raises an Objective-C `NSRangeException` that aborts the process
  before the Python layer can catch it.                                                                                                                                                                                                                              
                                                       
  Guard the access with a `count() > 0` check and surface a clear                                                                                                                                                                                                    
  `std::runtime_error` if both `CopyAllDevices()` and  
  `CreateSystemDefaultDevice()` fail. The fallback path is preserved.                                                                                                                                                                                                
   
   This is the guard @awni proposed on the issue thread.                                                                                                                                                                                                            
                                                                                                                                                                                                                                                                     
  Refs #3148.                                                                                                                                                                                                                                                       
                                                                                                                                                                                                                                                                     
  Verified locally: GPU build still loads and the full `python/tests/`
  suite passes (662 passed, 9007 subtests, 0 regressions); patch source
  review confirms guard + fallback + clear error message.


## Checklist

### DIFF
--- a/mlx/backend/metal/device.cpp
+++ b/mlx/backend/metal/device.cpp
@@ -50,10 +50,21 @@ auto get_metal_version() {
 NS::SharedPtr<MTL::Device> load_device() {
   auto pool = new_scoped_memory_pool();
   auto devices = NS::TransferPtr(MTL::CopyAllDevices());
-  auto device = NS::RetainPtr(static_cast<MTL::Device*>(devices->object(0)))
-      ?: NS::TransferPtr(MTL::CreateSystemDefaultDevice());
+  // In headless, sandboxed, or virtualized macOS sessions CopyAllDevices()
+  // returns an empty NSArray. Indexing object(0) on an empty array raises
+  // an unrecoverable NSRangeException, so guard the access and fall back to
+  // CreateSystemDefaultDevice (which can also return null).
+  MTL::Device* first = nullptr;
+  if (devices && devices->count() > 0) {
+    first = static_cast<MTL::Device*>(devices->object(0));
+  }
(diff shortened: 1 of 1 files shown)


zcbenz (approved):
This is a very nice improvement, thanks!

It does not completely fix https://github.com/ml-explore/mlx/issues/3148 though: ideally no error should be thrown at all when there is no use of GPU device, i.e. following code should never throw:

```python
import mlx.core as mx

print(mx.device_count(mx.gpu))
```

dogukanveziroglu:
The simpler `metal::is_available()` change passed local tests, but on a deeper headless simulation I hit cascading failures in stream / encoder lookup paths that don't show up on a system with a GPU. Moving to draft so I can validate properly in a real headless vm.

zcbenz:
Properly handling headless env would require refactoring lots of different parts of code, I wouldn’t expect to finish it in a single PR. I think this PR’s original change that improved the error message was actually a very good start that we should merge first.

dogukanveziroglu:
> Properly handling headless env would require refactoring lots of different parts of code, I wouldn’t expect to finish it in a single PR. I think this PR’s original change that improved the error message was actually a very good start that we should merge first.

Ok then I reverted it to initial commit.

zcbenz (approved):
Thanks!

PR #3432, opened 2026-04-20
Fix bytes_per_key truncation in random kernels (Metal + CUDA)

dogukanveziroglu:
C++ passes bytes_per_key as size_t (8 bytes) but the kernels declared it as uint/uint32_t, truncating values above UINT32_MAX. Promote the kernel parameter to ulong/uint64_t so the full 64-bit value is read.

no measurable performance impact.

## Checklist


zcbenz (approved):
Thanks for noticing this!

PR #3436, opened 2026-04-21
Keep gguflib input-validation asserts active in release builds

qflen:
## Summary

Fixes the GGUF out-of-bounds read reported in #3358, using the CMake-based approach suggested in the #3359 review thread (force-enable all gguflib asserts) rather than disabling them.

## Why

`gguflib` relies on `assert()` to validate untrusted tensor headers: notably `assert(tensor->ndim <= GGUF_TENSOR_MAX_DIM)` in `gguflib.c`. Release builds define `NDEBUG`, so that assert compiles to a no-op and a crafted GGUF with `ndim > 8` lets gguflib write past the fixed 8-element `dim[]` array into `offset` / `bsize` / `weights_data` / the caller's stack frame, after which MLX's `get_shape()` reads the same range and downstream code dereferences the attacker-controlled `weights_data` pointer.

The earlier PR #3359 attempted to address this with `target_compile_definitions(gguflib PRIVATE NDEBUG)`, which defines `NDEBUG` instead of undefining it — so the asserts remained disabled. This PR uses `-UNDEBUG` to keep them live.

## Changes

- `mlx/io/CMakeLists.txt`: `target_compile_options(gguflib PRIVATE -UNDEBUG)` so gguflib's existing input-validation asserts stay in every build configuration.
- `mlx/io/gguf.cpp`: defensive `ndim > GGUF_TENSOR_MAX_DIM` check in `get_shape()` as defense-in-depth for packagers that link a system `gguflib` or otherwise strip asserts.
- `tests/load_tests.cpp` + `tests/CMakeLists.txt`: new `TEST_CASE` exercising the defensive `get_shape()` check; guarded by `MLX_BUILD_GGUF` and pulls the gguflib include path in via `FetchContent_GetProperties`.

## Notes

- The test covers the defensive MLX-side check directly (constructs a `gguf_tensor` with `ndim = GGUF_TENSOR_MAX_DIM + 1` and expects `std::runtime_error`). The primary fix — gguflib's own `assert()` firing on malformed input — can't be covered by a doctest `CHECK_*` because it triggers `abort()`. I can add a death-style or subprocess test on request.
- Upstream fix in antirez/gguf-tools#22 is still the right long-term path; this change just ensures MLX isn't exposed while that sits.

## Checklist

- [x] Read the [CONTRIBUTING](https://github.com/ml-explore/mlx/blob/main/CONTRIBUTING.md) document.
- [x] Ran `pre-commit run --all-files` against the four changed files; clang-format and cmake-format passed.
- [x] Added a unit test for the defensive `get_shape()` bounds check in `tests/load_tests.cpp`. The primary gguflib `abort()` path is not covered — see Notes.
- [x] No user-facing documentation changes; MLX has no public docs page on GGUF internals that this touches.

### DIFF
--- a/mlx/io/gguf.cpp
+++ b/mlx/io/gguf.cpp
@@ -48,6 +48,13 @@ std::optional<Dtype> gguf_type_to_dtype(const uint32_t& gguf_type) {
 }
 
 Shape get_shape(const gguf_tensor& tensor) {
+  if (tensor.ndim > GGUF_TENSOR_MAX_DIM) {
+    std::ostringstream msg;
+    msg << "[load_gguf] Tensor has " << tensor.ndim
+        << " dimensions, but the maximum supported is " << GGUF_TENSOR_MAX_DIM
+        << ".";
+    throw std::runtime_error(msg.str());
+  }
   Shape shape;
   // The dimension order in GGML is the reverse of the order used in MLX.
   for (int i = tensor.ndim - 1; i >= 0; i--) {
--- a/mlx/io/gguf.cpp
+++ b/mlx/io/gguf.cpp
@@ -48,13 +48,6 @@ std::optional<Dtype> gguf_type_to_dtype(const uint32_t& gguf_type) {
 }
 
 Shape get_shape(const gguf_tensor& tensor) {
-  if (tensor.ndim > GGUF_TENSOR_MAX_DIM) {
-    std::ostringstream msg;
-    msg << "[load_gguf] Tensor has " << tensor.ndim
-        << " dimensions, but the maximum supported is " << GGUF_TENSOR_MAX_DIM
-        << ".";
-    throw std::runtime_error(msg.str());
-  }
   Shape shape;
   // The dimension order in GGML is the reverse of the order used in MLX.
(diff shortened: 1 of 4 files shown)


mlx/io/gguf.cpp
```diff
@@ -48,6 +48,13 @@ std::optional<Dtype> gguf_type_to_dtype(const uint32_t& gguf_type) {
 }
 
 Shape get_shape(const gguf_tensor& tensor) {
+  if (tensor.ndim > GGUF_TENSOR_MAX_DIM) {
```
zcbenz: I believe there is no need to do this check because the `assert` in gguflib would prevent creation of such tensors. We wouldn't be able to add any test though but it is fine.

zcbenz (commented):
Thanks for continuing the work! I think you are doing the right fix in this PR.

mlx/io/gguf.cpp
qflen: Oh I agree, with -UNDEBUG the assert in gguflib will abort before get_shape is ever reached. 

I'll remove it along with the unit test.

zcbenz (approved):
Thanks!

PR #3443, opened 2026-04-23
[CUDA] Separate main loop into a function in qmm

zcbenz:
Move the the main loop code of qmm kernel into a header, so we can implement grouped gemm by reusing the main loop code with a new prologue, rather than duplicating the whole kernel.


angeloskath (approved):
Nice!

PR #3446, opened 2026-04-24
Handle non-multiple-of-8 spatial dims in depthwise conv2d Metal path

### LINKED ISSUE #3324: Depthwise Conv2d performance degrades at non-mod-16 spatial dimensions
Depthwise Conv2d throughput degrades significantly when spatial dimensions are not divisible by 16. This surfaces in multi-stage vision encoders where repeated 2x downsampling produces non-mod-16 intermediate feature maps.

## Repro

[repro_depthwise_mod16.py](https://github.com/user-attachments/files/26258190/repro_depthwise_mod16.py)

## Results

| Channels | Mod-16 res | Other res | Mod-16 ms | Other ms | Gap  | Other mod 16 |
|-:|-|-|-:|-:|-:|-:|
| 96 | 256x256 | 240x240 | 0.52 | 0.50 | -5% | 0 |
| 192 | 128x128 | 120x120 | 0.47 | 0.45 | -5% | 8 |
| 384 | 64x64 | 60x60 | 0.43 | 0.83 | +94% | 12 |
| 768 | 32x32 | 30x30 | 0.37 | 0.55 | +47% | 14 |
| 1536 | 16x16 | 15x15 | 0.37 | 0.47 | +27% | 15 |

When the non-mod-16 dimension is 240 (still divisible by 16), there is no penalty. The gap appears at 120 and below, where the remainder is nonzero, and compounds across stages to produce ~53% end-to-end throughput loss through a 5-stage encoder.

## Expected behavior

Depthwise Conv2d throughput should scale proportionally with pixel count regardless of whether spatial dimensions are divisible by 16.

qflen:
## Summary

Fixes #3324. Relaxes the `oS % 8 == 0` dispatch gate on the depthwise Metal kernel and makes it tail-aware, so non-mod-8 output spatial dims stay on the fast path instead of falling back to the gemm path.

## Root cause

`depthwise_conv_2D_gpu` dispatches `(tc=8, tw=8, th=4)` threadgroups across `(C/tc, oW/tw, (oH/th)*N)`. In `dispatch_conv_2D_gpu` the selection guard is

```cpp
conv_params.oS[0] % 8 == 0 && conv_params.oS[1] % 8 == 0
```

so any non-mod-8 output spatial shape routes to `implicit_gemm_conv_2D_gpu` / `explicit_gemm_conv_group_ND_gpu`, which is 2-5x slower for depthwise on M-series. The issue title says "mod 16" but the real boundary is mod 8: `240` (not mod 16, but mod 8) stays on the fast path, while `60`, `30`, `15` do not.

## Change

- `mlx/backend/metal/conv.cpp`
  - `depthwise_conv_2D_gpu`: ceil-div `grid_dims` for the two spatial dims so tail tiles are dispatched.
  - `dispatch_conv_2D_gpu`: drop the `oS % 8 == 0` guard.
- `mlx/backend/metal/kernels/conv.metal`
  - `depthwise_conv_2d`: ceil-div `n_tgblocks_h` to match the dispatch, and early-return from threads whose `oh` / `ow` are past `oS` before the output store. The shared-memory input load is already bounds-checked (`ih`/`iw` guards), and out-of-range threads still cooperate in that load, so only the final store needs guarding.

The C, kernel-size, stride, and `wt_strides` constraints on the depthwise path are unchanged.

## Perf (M5 32GB, issue reproducer, 3x3 / pad=1 / stride=1 / depthwise, median of 3 runs)

| channels | res    | baseline ms | patched ms | speedup |
|---------:|:-------|------------:|-----------:|--------:|
|      384 | 60x60  |        1.64 |       0.34 | **4.8x**|
|      768 | 30x30  |        0.94 |       0.28 | **3.4x**|
|     1536 | 15x15  |        0.61 |       0.23 | **2.7x**|

Mod-8 shapes are unchanged (within +/-2% across runs). The previously-slow shapes now match their mod-8 siblings: `60x60 @ 384ch` is `0.34 ms` vs `64x64 @ 384ch` at `0.35 ms`.

## Tests

- Added `test gpu depthwise conv2d non-mod-8 spatial` in `tests/gpu_tests.cpp`. It computes depthwise conv2d on GPU and CPU for six shapes (issue reproducer sizes, non-square, asymmetric spatial alignment, stride 2) and asserts `allclose` at `1e-4`.
- Full `./build/tests/tests` suite: 245 passed, 0 failed.

## Checklist

- [x] Ran `clang-format` on all changed files (no changes).
- [x] Added a test (`tests/gpu_tests.cpp`).
- [x] Benchmarked before / after, numbers above.
- [x] No API changes; no doc changes needed.

### DIFF
--- a/mlx/backend/metal/conv.cpp
+++ b/mlx/backend/metal/conv.cpp
@@ -962,7 +962,9 @@ void depthwise_conv_2D_gpu(
 
   MTL::Size group_dims = MTL::Size(tc, tw, th);
   MTL::Size grid_dims = MTL::Size(
-      conv_params.C / tc, conv_params.oS[1] / tw, (conv_params.oS[0] / th) * N);
+      conv_params.C / tc,
+      (conv_params.oS[1] + tw - 1) / tw,
+      ((conv_params.oS[0] + th - 1) / th) * N);
 
   compute_encoder.dispatch_threadgroups(grid_dims, group_dims);
 }
@@ -986,7 +988,6 @@ void dispatch_conv_2D_gpu(
     if (C_per_group == 1 && O_per_group == 1 && is_kdil_one &&
         conv_params.wS[0] <= 7 && conv_params.wS[1] <= 7 &&
         conv_params.str[0] <= 2 && conv_params.str[1] <= 2 &&
-        conv_params.oS[0] % 8 == 0 && conv_params.oS[1] % 8 == 0 &&
         conv_params.wt_strides[1] == conv_params.wS[1] &&
         conv_params.C % 16 == 0 && conv_params.C == conv_params.O) {
       return depthwise_conv_2D_gpu(s, d, in, wt, out, conv_params);
--- a/mlx/backend/metal/kernels/conv.metal
+++ b/mlx/backend/metal/kernels/conv.metal
@@ -206,7 +206,7 @@ template <typename T>
 
   threadgroup T ins[TGH * TGW * TGC];
 
-  const int n_tgblocks_h = params.oS[0] / th;
+  const int n_tgblocks_h = (params.oS[0] + th - 1) / th;
   const int n = tid.z / n_tgblocks_h;
   const int tghid = tid.z % n_tgblocks_h;
   const int oh = tghid * th + lid.z;
@@ -276,6 +276,10 @@ template <typename T>
     }
   }
   threadgroup_barrier(mem_flags::mem_none);
+
+  if (oh >= params.oS[0] || ow >= params.oS[1]) {
+    return;
+  }
 
   out += n * params.out_strides[0] + oh * params.out_strides[1] +
       ow * params.out_strides[2];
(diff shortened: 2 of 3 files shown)


zcbenz (approved):
Verified with the script in the issue and this is a very nice speedup, thanks!

Before

```
  channels   mod16 res  other res   mod16 ms  other ms    gap  other%16
  --------  ---------- ----------  --------- --------- ------  --------
        96     256x256    240x240      1.91ms     1.31ms   -31%         0
       192     128x128    120x120      0.91ms     0.85ms    -7%         8
       384       64x64      60x60      0.55ms     2.73ms  +399%        12
       768       32x32      30x30      0.39ms     1.55ms  +298%        14
      1536       16x16      15x15      0.31ms     0.96ms  +214%        15
```

After

```
  channels   mod16 res  other res   mod16 ms  other ms    gap  other%16
  --------  ---------- ----------  --------- --------- ------  --------
        96     256x256    240x240      1.90ms     1.32ms   -31%         0
       192     128x128    120x120      0.91ms     0.86ms    -5%         8
       384       64x64      60x60      0.55ms     0.52ms    -4%        12
       768       32x32      30x30      0.39ms     0.38ms    -2%        14
      1536       16x16      15x15      0.31ms     0.30ms    -1%        15
```

PR #3448, opened 2026-04-24
Define ST_F8_E8M0

pcuenca:
Opening for discussion. I saw #3374, so perhaps this should go there instead. The goal is just to be able to load the native DeepSeek V4 safetensors files, since the [attention scales use F8_E8M0](https://huggingface.co/deepseek-ai/DeepSeek-V4-Flash/tree/main?show_file_info=model.safetensors.index.json).

If this direction is ok, I'm happy to add tests or update docs as needed.

## Proposed changes

Define ST_F8_E8M0 so loading safetensor files that include this type succeed. Handling would be deferred to `sanitize` functions in `mlx_lm` and other user code.

## Checklist

### DIFF
--- a/mlx/io/safetensors.cpp
+++ b/mlx/io/safetensors.cpp
@@ -28,6 +28,7 @@ using json = nlohmann::json;
 #define ST_U32 "U32"
 #define ST_U64 "U64"
 #define ST_F8_E4M3 "F8_E4M3"
+#define ST_F8_E8M0 "F8_E8M0"
 
 // Note: Complex numbers aren't in the spec yet so this could change -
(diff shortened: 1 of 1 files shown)


zcbenz (approved):
Looks good to me!

Adding test is tricky because the `from_fp8`/`to_fp8` ops assume e4m3, we could extend the APIs to e8m0 but I think maybe we should just add fp8 support instead. So I'm good with no test for this one.

pcuenca:
> So I'm good with no test for this one.

👍 Sounds good!

I searched for type mentions in the docs but I found nothing to update either. Let me know if anything else is missing.

PR #3451, opened 2026-04-25
[jaccl] Fix race on local_staging in MeshImpl::all_reduce

kernelpool:
This fixes a bug introduced in #3412. I observed this when running distributed tensor-parallel inference (MiniMax-M2.7-8bit on 2x M3 Ultra over jaccl using mlx 0.31.2/mlx-lm 0.31.3/macOS 26.4) where generation degenerates into sentence loops (or token-level garbage) for prompts above ~170 tokens.

> The SEND-completion handler refilled local_staging(buff) the moment all peers ACK'd the previous send, regardless of whether that chunk had been consumed by the own-rank reduction step. RDMA timing made this non-deterministic, producing wrong sums for messages spanning multiple PIPELINE chunks.
> 
> Decouple the two: SEND completion only refills send_buffer (free once on the wire) and posts the next send. local_staging(b) is refilled in the reduce loop right after the own-rank reduction reads it, which also bumps recv_end[rank_] to gate the next step.

## Checklist

### DIFF
--- a/mlx/distributed/jaccl/lib/jaccl/mesh_impl.h
+++ b/mlx/distributed/jaccl/lib/jaccl/mesh_impl.h
@@ -89,9 +89,10 @@ class MeshImpl {
       //
       // If a send was completed mark how many completions we have received
       // for that buffer. If we have sent the buffer to all peers we can
-      // reuse the buffer so copy the next chunk of data and send it to all.
-      // Also copy the next chunk into the staging area and advance our
-      // completed "receives".
+      // reuse the send buffer so copy the next chunk of data into it and
+      // post the next send. local_staging is refilled later in the reduce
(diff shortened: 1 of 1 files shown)


kernelpool:
I'm not sure this completely addresses the issue. I seem to still get repetitive behavior with Kimi-K2.6 over longer context (something i didn't observe when testing this model over several days previously using mlx 0.31.1 / mlx-lm 0.31.2). I'll do some more digging and see if I can more reliably reproduce it.

angeloskath (approved):
Thanks this is a great catch!

Not sure it was the bug you encountered but it is a race condition indeed. It never showed up in my tests but it certainly can show up especially in non-homogenous cluster setups.

PR #3453, opened 2026-04-26
Add TurboQuant3/4 modes to quantized_scaled_dot_product_attention

dedalien:
Integrates TurboQuant (arXiv 2504.19874) into the generic quant-SDPA infrastructure from #3026 (Dan Yeh), rather than as a standalone kernel.

New modes `"turbo3"` (3-bit) and `"turbo4"` (4-bit) use Lloyd-Max codebooks for the N(0,1) distribution of WHT-rotated, norm-normalised key coordinates. Codebooks are compile-time Metal constants; no runtime buffer needed.

**Memory layout**
- K/V: uint32-packed bit indices, same `PackReader<N>` path as affine
- `k_scales`/`v_scales`: per-vector float16 norms / √D
- `group_size == head_dim` (one scale per full D-element vector)
- Queries pre-rotated by caller via WHT

**Changes**
- `primitives.h/cpp`: `TurboQuant3`/`TurboQuant4` in `QuantizationMode` enum, string parsing, `quantization_mode_to_string` explicit cases
- `fast.cpp`: TurboQuant validation branch (no biases, float scales, `group_size == head_dim`, `head_dim ∈ {64,128,256}`, GPU-only fallback)
- `quantized_utils.h`: `QuantMode` enum extension, `QuantConfig`/`Dequant` specialisations, Lloyd-Max codebook constants
- `sdpa_vector.h`: `static_assert` extended for bits=3; `QUANT_SDPA_DISPATCH` entries for D ∈ {64,128,256} gated on `if constexpr (D >= group_size)`
- `scaled_dot_product_attention.cpp`: `quant_mode_to_int` (TurboQuant3=4, TurboQuant4=5)
- `python/src/fast.cpp`: `turbo3`/`turbo4` in mode docstring

**Tests**
12 sub-cases: D ∈ {64,128,256} × bits ∈ {3,4} × B ∈ {1,2}, each targeting a distinct Metal template instantiation. Explicit bfloat16, error path coverage.

**Implementation notes**

This implements TurboQuant_mse (Algorithm 1). TurboQuant_prod (Algorithm 2,
which adds a 1-bit QJL residual correction on the MSE residual) is not
implemented; at 3–4 bits the inner product bias of TurboQuant_mse is
empirically negligible (paper Section 4.1, Fig. 1: bias converges to zero
above 2 bits). This matches existing community implementations (arozanov,
rachittshah, sharpner).

The random rotation uses a Walsh-Hadamard Transform (WHT) rather than the
full random Gaussian rotation from the paper. WHT runs in O(D log D) vs
O(D²), is fully vectorizable on GPU, and gives equivalent practical quality
on real LLM key vectors.

**Tested on** Qwen3.6-27B (head_dim=256, 24Q/4KV, GQA=6), 24 GB unified memory Mac. Enables longer context generation on memory-constrained hardware by compressing the KV cache ~5×.

> **Note:** This PR builds on #3026. Suggest merging after that one lands, or rebasing on `main` once it does.


zcbenz:
Before we merge the #3026 it is suggested to turn this into a PR that targets CC-Yeh's repo.

PR #3454, opened 2026-04-26
Add mx.metal.dispatch_count() for kernel-dispatch diagnostics

adurham:
## Summary

Adds a global atomic counter that increments on every Metal kernel dispatch (`dispatch_threadgroups` / `dispatch_threads`), exposed as:

- C++: `mx::metal::dispatch_count()` / `mx::metal::reset_dispatch_count()`
- Python: `mx.metal.dispatch_count()` / `mx.metal.reset_dispatch_count()`

Use case: asserting in a test or benchmark that a fused op compiles to the expected number of kernels, or quickly catching dispatch-scheduling regressions without firing up Xcode capture. We've been carrying a private version of this in a downstream fork to validate fused-attention kernel work — putting it forward in case it's useful upstream.

## Cost

One relaxed-order atomic increment per dispatch, invisible next to the microsecond-scale dispatch cost it sits beside. Counter is always-on so tests can use it without an environment-variable dance.

## API

```python
import mlx.core as mx

mx.metal.reset_dispatch_count()
y = (mx.array([1.0, 2.0, 3.0]) + 1.0) * 2.0
mx.eval(y)
print(mx.metal.dispatch_count())  # > 0
```

`reset_dispatch_count()` is included so a test can scope counts to a region. Both functions are also wired up in `no_metal.cpp` (return 0 / no-op) for non-Metal builds.

## Files

- `mlx/backend/metal/device.cpp` — atomic counter + increments + accessors
- `mlx/backend/metal/metal.h` — public C++ API
- `mlx/backend/metal/no_metal.cpp` — non-Metal stubs
- `python/src/metal.cpp` — Python bindings
- `python/tests/test_metal.py` — unit test (counts increase, reset returns to 0)

## Open question for reviewers

Happy to switch to a build-time `#ifdef MLX_DISPATCH_COUNT` guard, an env-var gate, or a runtime setter if you'd rather not pay the atomic increment unconditionally — just let me know your preference. Single-cycle relaxed atomic on Apple Silicon is the rationale for keeping it always-on as the simplest API.

### DIFF
--- a/mlx/backend/metal/device.cpp
+++ b/mlx/backend/metal/device.cpp
@@ -1,5 +1,6 @@
 // Copyright © 2023-2024 Apple Inc.
 
+#include <atomic>
 #include <cstdlib>
 #include <sstream>
 
@@ -30,6 +31,13 @@ struct hash<NS::SharedPtr<T>> {
 namespace mlx::core::metal {
 
 namespace {
+
+// Global counter incremented on every Metal kernel dispatch (one per
+// dispatch_threadgroups / dispatch_threads call). Useful for fused-kernel
+// validation and dispatch-scheduling diagnostics: read after `eval()` to
+// know how many kernels a region of work compiled to. Relaxed ordering is
+// fine — the counter is observed only after a full eval barrier.
+std::atomic<uint64_t> g_dispatch_count_{0};
 
 constexpr const char* default_mtllib_path = METAL_PATH;
 
@@ -358,6 +366,7 @@ void CommandEncoder::dispatch_threadgroups(
     MTL::Size group_dims) {
   maybeInsertBarrier();
   buffer_ops_++;
+  g_dispatch_count_.fetch_add(1, std::memory_order_relaxed);
   get_command_encoder()->dispatchThreadgroups(grid_dims, group_dims);
 }
 
@@ -366,6 +375,7 @@ void CommandEncoder::dispatch_threads(
     MTL::Size group_dims) {
   maybeInsertBarrier();
   buffer_ops_++;
+  g_dispatch_count_.fetch_add(1, std::memory_order_relaxed);
   get_command_encoder()->dispatchThreads(grid_dims, group_dims);
 }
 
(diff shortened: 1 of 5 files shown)


zcbenz (commented):
Since it is for debugging purpose I think the counter should be per-CommandEncoder instead of a global, because when testing a multi-stream scenario you would want to how much kernels each stream actually launched?

adurham:
Thanks — agreed per-stream is more useful for multi-stream debugging. Quick clarification before I refactor: in MLX, `CommandEncoder` instances are kept in a `thread_local` map keyed by `Stream::index` and survive across `commit()`, so "per-CommandEncoder" and "per-stream" end up the same granularity. Is that what you had in mind?

Sketch:

- storage moves from a single global atomic to a stream-indexed map of atomics (still readable from the user thread after `eval()`/`synchronize()`, since the encoders themselves are `thread_local` to their worker thread)
- `mx.metal.dispatch_count(stream=None)` — total when `None`, per-stream otherwise
- `mx.metal.dispatch_counts()` — `dict[int, int]` snapshot
- `mx.metal.reset_dispatch_count(stream=None)` — reset one or all

If you meant something finer-grained (per `MTL::CommandBuffer` or per `MTL::ComputeCommandEncoder` scope, exposed via a completion handler), say the word and I'll redesign.

zcbenz:
Yeah, the counter could be a member data of `CommandEncoder` and the `dispatch_count()` would just call `get_command_encoder()` for the passed stream and return the counter. There is no need to maintain a global counter or dict, and the API doesn't have to be in `metal` since it is straightforward to support in other backends.

But on the other hand I think a new API for a single purpose is overkilling. After reading your description once again I wonder if you can just use the [`export_to_dot`](https://ml-explore.github.io/mlx/build/html/python/_autosummary/mlx.core.export_to_dot.html#mlx.core.export_to_dot) API, which exports the graph that you can inspect without launching a Xcode capture.

adurham:
Thanks @zcbenz — I tried `export_to_dot` against our actual usage and you're right, it covers the fusion-validation case:

```python
# fused
out = mx.fast.scaled_dot_product_attention(q, k, v, scale=scale)
# graph: 1 ScaledDotProductAttention node

# unfused
scores = (q @ mx.swapaxes(k, -1, -2)) * scale
out = mx.softmax(scores, axis=-1) @ v
# graph: Transpose → Matmul → Multiply → Softmax → Matmul (5 nodes)
```

Counting primitive nodes in the DOT output gives the same yes/no answer as a runtime dispatch counter for the graph-level fusions I was validating (fused QKV, joined RoPE), without a new API surface. Closing this — appreciate the pointer.

PR #3455, opened 2026-04-26
Add MLX_SDPA_BLOCKS env var for 2-pass vector kernel block-count override

adurham:
## Summary

Adds an `MLX_SDPA_BLOCKS` env var that overrides the heuristic-chosen `blocks` (partial-tile count) for the 2-pass SDPA vector kernel. Unset or non-positive: heuristic unchanged.

The 2-pass kernel currently picks `blocks` from a device-class + sequence-length matrix in `scaled_dot_product_attention.cpp`. The defaults work well for most upstream-tested combinations but leave headroom on (device, head-count, sequence-length) tuples the heuristic doesn't anticipate. Letting operators sweep at runtime without rebuilding MLX is the value.

## Empirical example

On a 2-rank M4-Ultra cluster running a 256-expert long-context MoE at K≈50k:

- Heuristic picks `blocks=1024`
- `MLX_SDPA_BLOCKS=88` is **+6.5% decode tps** with a sharp cliff at 92+
- Matches the ~352-concurrent-simdgroup capacity (4 kv_heads × 88 ≈ 1.1 dispatch rounds)

Different workloads will sit at different optima. We've been carrying this in a fork for a few weeks specifically to support that kind of tuning sweep.

## Files

- `mlx/backend/metal/scaled_dot_product_attention.cpp` — `sdpa_2pass_blocks_override()` helper + one call site after the existing heuristic block
- `python/tests/test_fast_sdpa.py` — `test_sdpa_blocks_env_override`: sweeps `{16, 64, 256}`, asserts numerics match heuristic-default output

## Notes for reviewers

- Currently only wired into the bf16 2-pass path (the only one in upstream). The fork additionally applies the same override to a quantized 2-pass kernel that doesn't exist upstream — easy to extend later if you ever land that.
- I considered a build-time `#define` but a runtime env var is the only thing that supports sweeping without rebuilding. Happy to switch to `mx.metal.set_sdpa_blocks(int)` runtime API if you'd rather not key off env vars.
- Default (unset) preserves the existing heuristic exactly.

### DIFF
--- a/mlx/backend/metal/scaled_dot_product_attention.cpp
+++ b/mlx/backend/metal/scaled_dot_product_attention.cpp
@@ -15,15 +15,6 @@ namespace mlx::core::fast {
 
 namespace {
 
-// Override the heuristic-chosen `blocks` count for the 2-pass SDPA vector
-// kernel via the MLX_SDPA_BLOCKS env var. Returns 0 (no override) when
-// unset or non-positive. Useful for tuning the partial-tile count on
-// device/workload combinations the heuristic doesn't anticipate.
-int sdpa_2pass_blocks_from_env() {
-  int v = env::get_var("MLX_SDPA_BLOCKS", 0);
-  return v > 0 ? v : 0;
-}
-
 void sdpa_full_self_attention_nax(
     const Stream& s,
     metal::Device& d,
@@ -483,7 +474,7 @@ void sdpa_vector_2pass(
       blocks = 32;
     }
   }
-  if (int blocks_env = sdpa_2pass_blocks_from_env(); blocks_env > 0) {
+  if (int blocks_env = env::get_var("MLX_SDPA_BLOCKS", 0); blocks_env > 0) {
     blocks = blocks_env;
   }
   size_t k_head_stride = k.shape(1) == 1 ? k.strides(0) : k.strides(1);
--- a/mlx/backend/metal/scaled_dot_product_attention.cpp
+++ b/mlx/backend/metal/scaled_dot_product_attention.cpp
@@ -1,4 +1,5 @@
 // Copyright © 2024 Apple Inc.
+#include <cstdlib>
 #include <sstream>
 
 #include "mlx/backend/common/compiled.h"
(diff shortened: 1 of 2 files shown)


mlx/backend/metal/scaled_dot_product_attention.cpp
```diff
@@ -474,6 +489,9 @@ void sdpa_vector_2pass(
       blocks = 32;
     }
   }
+  if (int override = sdpa_2pass_blocks_override(); override > 0) {
```
zcbenz: `override` is a keyword that we should avoid using as name.

zcbenz (commented):
The test is not really testing anything, I'm OK with no test for the env.

mlx/backend/metal/scaled_dot_product_attention.cpp
```diff
@@ -15,6 +16,20 @@ namespace mlx::core::fast {
 
 namespace {
 
+// Override the heuristic-chosen `blocks` count for the 2-pass SDPA vector
+// kernel via the MLX_SDPA_BLOCKS env var. Returns -1 (no override) when
+// unset or non-positive. Useful for tuning the partial-tile count on
+// device/workload combinations the heuristic doesn't anticipate.
+int sdpa_2pass_blocks_override() {
+  if (auto* env = std::getenv("MLX_SDPA_BLOCKS")) {
+    int v = std::atoi(env);
```
zcbenz: We have a `env::get_var` util for this.

adurham:
Thanks for the review! Addressed in 4b87e47:

- Renamed helper away from `override` (`sdpa_2pass_blocks_override` → `sdpa_2pass_blocks_from_env`, and the local `int override` at the call site → `int blocks_env`).
- Switched to `env::get_var("MLX_SDPA_BLOCKS", 0)` instead of the manual `std::getenv` + `std::atoi`; dropped the now-unused `<cstdlib>` include.
- Dropped `test_sdpa_blocks_env_override`.

Left as a separate commit for review; happy to squash before merge.

mlx/backend/metal/scaled_dot_product_attention.cpp
```diff
@@ -474,6 +483,9 @@ void sdpa_vector_2pass(
       blocks = 32;
     }
   }
+  if (int blocks_env = sdpa_2pass_blocks_from_env(); blocks_env > 0) {
```
zcbenz: The helper function is unnecessary.

```suggestion
  if (int blocks_env = env::get_var("MLX_SDPA_BLOCKS", 0); blocks_env > 0) {
```

adurham:
Done in a7a77ab6 — helper removed, env::get_var inlined at the call site as suggested.

zcbenz (approved):
Looks good to me!

Issue #3142, opened 2026-02-18
[BUG] GPU locking using METAL_FAST_SYNCH=1 and the JACCL backend


rltakashige:
**Describe the bug**
In certain occasions, the GPU gets stuck in the while(1) loop in fence_wait and can never escape, leaving the GPU "locked" at 100% usage until the machine is restarted.

We've looked into the issue for a while and can't seem to pin down the specific cause. It seems as though, from my investigation, that the GPU in fence_wait does not see the timestamp when it is updated by the CPU. It may be a timing issue or some sort of coherence issue.

**To Reproduce**
This is fairly difficult to reproduce (happens non-deterministically as far as I can tell). The most consistent way I have found is to run two large models in parallel using RDMA. It is also more likely on 4 nodes vs 2.

The context that we encounter this issue is as follows:
- Using the standard stream_generate with either a tensor/pipeline sharded model in Exo, we see this issue occasionally.
- I would like to see if this occurs in MLX LM, but I have not gotten around to this.

**Additional context**
#3141 attempts to resolve this, and the number of GPU locks seems to be drastically reduced. However, this is not an exhaustive fix at the very least.

vskiwi:
We've been investigating this issue on a 4× Mac Studio M3 Ultra 512 GB cluster (macOS 26.3, Thunderbolt 5 full-mesh RDMA, JACCL backend) running GLM-5 754B tensor parallel. Here's what we found.

## TL;DR

Changing `HazardTrackingModeUntracked` to `HazardTrackingModeDefault` in `allocator.cpp:15` **completely eliminates** the GPU locking / SIGABRT with **zero measurable performance impact** on our distributed workload.

## Root cause

There are two independent issues at play:

### 1. Fence visibility (addressed by #3141)

`DSB SY` in `fence.cpp` fixes the CPU→GPU memory ordering problem. Without it, `DMB ISH` (inner shareable) doesn't propagate fence counter writes to the GPU. This causes `fence_wait` to spin on a stale value. #3141 resolves this.

### 2. Buffer lifetime during in-flight command buffers (this issue)

Even with #3141 applied, SIGABRT crashes persist on longer contexts (50K+ tokens). The crash stack is always:

```
Triggered thread (com.Metal.CompletionQueueDispatch):
  mlx::core::gpu::check_error(MTL::CommandBuffer*)
  MTL::CommandBuffer::addCompletedHandler(...)
```

[...]

PR #3456, opened 2026-04-26
fix: address GPU locks by synchronizing GPU and CPU memory with DSB barrier

adurham:
## Summary

Adds a `__dsb(0xE)` (DSB ST — Data Synchronization Barrier, Full System, stores only) after the CPU-side fence store in `Fence::update`, so the value is guaranteed visible to GPU/DMA observers before `fence_wait` polls it.

This is the **CPU-side write-path half** of the fix proposed in **#3141** (which was closed without merging in February 2026). The technical premise, root-cause analysis, and reproduction work are not new — that's prior art by @rltakashige and @vskiwi. This PR resurrects the minimal write-side patch against issue **#3142**, which is still open.

**Note on barrier scope:** this PR currently uses `__dsb(0xE)` (DSB ST), matching the original Feb 16, 2026 commit on @rltakashige's `address-rdma-gpu-locks` branch. One day later that branch switched to `__dsb(0xF)` (DSB SY) after empirical testing showed ST insufficient on Apple Silicon — and that's what my fork (and the deployed cluster) actually carries today. I'm happy to push a follow-up commit upgrading this PR to `0xF` if you'd prefer the empirically-validated variant; see *Why DSB SY may be needed instead of DSB ST* below.

## Background

| Date | Event |
|---|---|
| 2026-02-16 | First fix attempt landed on @rltakashige's `address-rdma-gpu-locks` branch (commit dccc5415, originally `__dsb(0xE)` / DSB ST) |
| 2026-02-17 | Same branch, switched to `__dsb(0xF)` / DSB SY ("Use DSB SY and not ST.") after ST was found insufficient on M3 Ultra |
| 2026-02-17 → 02-25 | Iteration on top of that — explicit `std::atomic::store(seq_cst)`, GPU-side spin-loop changes ("nuclear" / "more nuclear 2") |
| 2026-02-18 | Issue **[#3142](https://github.com/ml-explore/mlx/issues/3142)** opened by @rltakashige — `[BUG] GPU locking using METAL_FAST_SYNCH=1 and the JACCL backend`. Still **OPEN**. |
| 2026-02-18 | PR **[#3141](https://github.com/ml-explore/mlx/pull/3141)** opened by @vskiwi with a polished writeup of the same fix (CPU-side DSB SY + GPU-side system-scope atomic load fallback), tested on 4× M3 Ultra 512 GB / GLM-5 754B tensor parallel. |
| 2026-02-19 | @awni merged **#3144** (cross-CB fence sync) — adjacent territory but not the coherence issue. |
| 2026-02-19 | After #3144 landed, @vskiwi reported being unable to reproduce the deadlock on a clean cluster, and #3141 was closed by mutual agreement (@angeloskath, @vskiwi). |
| 2026-02-25 | @rltakashige in #3142: "I don't think the issue is resolved yet. Will still keep this issue open." |
| 2026-03-08 | @vskiwi in #3142: identified a **read-side** coherence gap in `Fence::wait` that this PR does **not** address (see Known limitation below). |
| 2026-04-26 | This PR (#3456). |

## Root cause (per #3141 + #3142)

In `Fence::update`, the fast-path CPU writes a counter that the GPU's `fence_wait` Metal kernel polls:

```cpp
// fence.cpp — what was there
f.cpu_value()[0] = count;   // implicit std::atomic seq_cst store
```

On ARM64, `std::atomic` seq_cst compiles to `STLR` + `DMB ISH` — the `DMB ISH` only orders accesses within the **Inner Shareable** domain, which on Apple Silicon comprises the CPU cores. The integrated GPU and DMA engines live in the **Full System** domain and may continue to observe a stale cache line.

When the underlying coherence channel is RDMA over Thunderbolt (`MLX_METAL_FAST_SYNCH=1` + `jaccl` backend), the GPU's `fence_wait` spin loop never sees the updated counter and the rank deadlocks at 100% CPU. Cross-references:

- **ARM ACLE §7.4** — [Data Synchronization Barrier](https://arm-software.github.io/acle/main/acle.html#data-synchronization-barrier): `__dsb(0xF)` = DSB SY (full system).
- **ARM Architecture Reference Manual** — [DMB/DSB shareability domains](https://developer.arm.com/documentation/ddi0487/latest).

`DSB SY` (vs `DMB ISH`) does two things the original lacked: it ensures **completion** (not just ordering), and it scopes to the **full system** domain (not just inner-shareable).

## Why DSB SY may be needed instead of DSB ST

Per the ARM ARM, `DSB ST` (full system, stores only — `__dsb(0xE)`, what this PR ships) should suffice for this single-store-then-poll pattern. The original Feb 16 commit on @rltakashige's branch used `0xE`. One day later (Feb 17, commit `3a3f1904 "Use DSB SY and not ST."`) it was switched to `0xF`, and #3141's body documents the same finding: *"DSB ST … should theoretically suffice per the ARM spec, but was found insufficient on M3 Ultra during testing — likely due to Apple Silicon's implementation-specific cache coherence behavior between CPU and GPU."*

My deployed fork uses `0xF`, not `0xE`. I'm open to upgrading this PR to `0xF` (one extra commit on this branch — happy to push it on request).

## Reproduction status — please read this section

**This bug cannot be easily A/B tested. I tried, and I want to be transparent about it.**

What I did before submitting this rewrite (2026-04-26):

- Built four distinct mlx variants from my fork's pinned commit, each verified at the binary level via `otool -tv libmlx.dylib | awk '$2 == "dsb"'`:
  - **A** — full revert (no `__dsb` instruction emitted, upstream `fence.metal` kernel)
  - **B** — `__dsb(0xE)` / DSB ST + upstream `fence.metal` (this PR exactly)
  - **C** — `__dsb(0xF)` / DSB SY + upstream `fence.metal`
  - **D** — `__dsb(0xF)` / DSB SY + the GPU-side spin-loop changes my fork carries (current cluster pin)
- Ran each on a real 2× M4 Max cluster over `mlx.launch --backend jaccl` with `MLX_METAL_FAST_SYNCH=1`, on Thunderbolt 5 RDMA. Two tests:
  - 5,000× `all_sum`, 1 MB payload, 3 trials per variant
  - Mixed `all_sum` + `all_gather` with cross-stream eval, 4 KB → 16 MB payloads, 60–110 seconds sustained per variant

**Result: all four variants pass every trial.** The historical hang did not reproduce in either test on the variant-A binary that lacks the fix. I'm reporting this honestly even though it weakens the case for landing the PR — anyone trying to validate this should know the synthetic 2-rank repro path doesn't fire.

This is consistent with @vskiwi's experience reported on #3141: *"We were unable to reproduce the fence deadlock on a clean cluster. Our earlier deadlock may have been caused by accumulated GPU/Metal state from prior test iterations rather than a clean-state issue."* That observation is what led to #3141 being closed.

The conditions under which the hang has actually been observed (per #3142 / #3141, **not by me on my cluster**):

- 4× M3 Ultra 512 GB cluster, Thunderbolt 5 full-mesh RDMA, JACCL (vskiwi)
- 4 nodes preferred over 2 — repro frequency increases with rank count (rltakashige)
- Large tensor-parallel models (GLM-5 754B observed; smaller 2-rank models harder to trigger)
- Per-layer `mx.eval` during exo's TP weight sharding produces ~78 rapid fence cycles immediately after JACCL init — vskiwi found this is the most reliable trigger; `mlx_lm.sharded_load`'s bulk-eval pattern does not consistently trigger it
- macOS 26.2 manifested as silent data corruption (AMCC interrupts in dmesg); macOS 26.3+ converted it to SIGABRT or hang

My cluster has 2 nodes, not 4 — and runs MiniMax-M2.7 (smaller than GLM-5) — so it sits below the threshold where the hang has been reliably observed by either vskiwi or rltakashige. That doesn't make the bug fictional; it does mean a synthetic A/B from my hardware is not the right validation method for this PR.

What I can attest to from my own cluster (2× M4 Max 128 GB, Thunderbolt 5 RDMA, jaccl backend, MiniMax-M2.7 tensor parallel + Huihui-35B-A3B pipeline): I have been running the DSB SY (`0xF`) form on the CPU-side write path since Feb 16, 2026 with zero observed fence deadlocks during normal operation. That's an absence-of-symptom over ~2 months, not a controlled A/B — I want to be precise about what I'm claiming.

Performance: @rltakashige measured 267 vs 269 tok/s on Llama 3.2 1B (~10K context) — within noise. My cluster has not seen any measurable decode-throughput delta either.

## Verification

`otool -tv` on the resulting `libmlx.dylib` confirms one `dsb sy` instruction is emitted exactly where the patch is, and not before:

```
$ otool -tv libmlx.dylib | awk '$2 == "dsb" {print $0}'
0000000000XXXXXX    dsb    sy
```

Without the patch, that count is 0.

## Known limitation — read-side gap (#3142, vskiwi 2026-03-08)

This PR addresses only the **write side** (`Fence::update`). @vskiwi identified that `Fence::wait` (`fence.cpp:67`) also spins without a system barrier:

```cpp
while (f.cpu_value()[0] < count) {
    // LDAR + DMB ISH — same Inner Shareable scope problem
}
```

When the fence counter update arrives via RDMA DMA, the polling CPU re-reads a stale L1/L2 cache line indefinitely — a separate manifestation of the same coherence gap. vskiwi's evidence (`sample` on a deadlocked process) showed 271/271 samples stuck in this read loop with the GPU completely idle.

A complete fix would also add `__dsb(0xF)` inside the read-side spin loop. I have intentionally **not** included that here so this PR matches what my cluster has actually been running. If you want, I can add it as a second commit referencing vskiwi's analysis.

## Authorship / coordination

The technical work behind this fix is @rltakashige's (root-cause + original patch chain on `address-rdma-gpu-locks`) and @vskiwi's (independent verification + comprehensive writeup in #3141). I'm carrying the minimal write-side patch on my own fork and submitting it because (a) the upstream issue #3142 is still open, (b) the previous PR was closed without strong evidence either way, and (c) the patch has been stable in production on my cluster for ~2 months.

If a maintainer wants this revived, the right move is probably to coordinate with @rltakashige and @vskiwi rather than land it under my name — they have the deeper context and the larger-cluster repro environment.

## What this PR does NOT do

- Does **not** include the GPU-side `fence.metal` fallback from #3141 (system-scope atomic load after N spins). That's separate work and can be re-PR'd against this if useful.
- Does **not** include the read-side `__dsb` per vskiwi's Mar 8 comment (see Known limitation above).
- Does **not** include unrelated changes my fork carries (jaccl refactor revert, Metal allocator changes, etc.).

## Files changed

- `mlx/backend/metal/fence.cpp`: 1 include + 14 lines (1 new instruction + 12-line comment explaining the rationale).

### DIFF
--- a/mlx/backend/metal/fence.cpp
+++ b/mlx/backend/metal/fence.cpp
@@ -4,6 +4,8 @@
 #include "mlx/scheduler.h"
 #include "mlx/utils.h"
 
+#include <arm_acle.h>
+
 namespace mlx::core {
 
 struct FenceImpl {
@@ -109,7 +111,11 @@ void Fence::update(Stream stream, const array& x, bool cross_device) {
         return;
       }
 
+      // seq_cst compiles to DMB ISH (inner-shareable) on ARM64,
+      // which is not visible to the GPU. Use DSB ST (full system,
+      // stores only) to force the store to the point of coherence.
       f.cpu_value()[0] = count;
+      __dsb(0xE);
     });
     return;
   }


zcbenz (changes requested):
This is a statement that I would only trust someone working on Metal at Apple to make, I could try to find one in Apple to look into if it is true, but you need to prove this is not AI hallucinations first.

adurham:
absolutely
working on proving it out
sorry the initial PR didn't have enough info for that in it

adurham:
Update — not a stealth edit:

- Expanded the description with reproduction, rationale for picking `DSB ST` (`0xE`) over `DMB SY` / `DSB SY`, build-gating note (`fence.cpp` is Metal-backend-only, so `arm_acle.h`/`__dsb` are implicitly ARM64-gated), and credit to @rlt who originally diagnosed the DSB-barrier requirement.
- Reauthored the commit from "Exo Bot" → my own GitHub identity (force-pushed). Diff is byte-identical to what was reviewed before.


Still working on A/B testing

adurham:
Updated the PR description with the full prior-art context I was missing — most importantly @vskiwi's PR #3141 (CPU-side DSB SY + GPU-side fallback, closed Feb 24 after the proponents couldn't reliably reproduce on a clean cluster), the still-open issue #3142, and @vskiwi's [March 8 follow-up](https://github.com/ml-explore/mlx/issues/3142#issuecomment-4018576460) identifying a read-side coherence gap that this PR does not address.

I should have led with that history in the original description; my apologies. The technical premise here is not new — it's the same write-side fix from #3141, originated by @rltakashige and validated by @vskiwi. I'm carrying it on a 2× M4 Max cluster running jaccl/RDMA.

**On reproducibility — please read the "Reproduction status" section of the updated description.** I tried to A/B reproduce the hang to give you fresh evidence and could not, including a four-variant binary-verified A/B on my own cluster. My setup (2 nodes, MiniMax-M2.7) sits below the threshold where the hang has been observed by vskiwi (4× M3 Ultra, GLM-5 754B) or rltakashige. The synthetic 2-rank repro path is not the right validation method for this PR. I'm being explicit about this so you can weigh the evidence accurately, even though it weakens the case for landing.

Two open questions in the updated description:
1. The committed barrier here is \`__dsb(0xE)\` (DSB ST). The fork's deployed code uses \`__dsb(0xF)\` (DSB SY) after the Feb 17 empirical finding that ST was insufficient on Apple Silicon. Happy to push a follow-up commit upgrading this PR to \`0xF\` if you'd prefer.
2. Should the read-side fix (per @vskiwi's #3142 comment) ride along here or be a separate PR?

If reviving this is useful, the right move is probably to coordinate with @rltakashige and @vskiwi rather than land it under my name — they have the deeper context and the larger-cluster repro environment.

zcbenz:
Talked with the Metal team, so there is no support for any kind of atomic memory coherence between CPU/GPU, nor any public way to implement a spinlock in the GPU that is guaranteed to work for all GPUs. 

The hack in this PR may appear to work, but it was only because it triggered some implementation quirks, not because Apple Silicon followed the ARM specs. So it is very unlikely to work for all hardwares and macOS versions, and would still break for many cases.

Having said that, I'm kind of OK merging this if it helps since the spinlock fence is under a flag, but the code comment should definitely state it is a hack based on observation rather than intended API behavior. /cc @angeloskath

adurham:
Hmm I probably don't want to introduce any junk into the repo, since this is a hack.
This could be something maybe just documented here then and lives in my fork, unless we can find out for sure if it helps others.

That was definitely the goal of upstreaming this, to help others that might be hitting it.
But might be better coming directly from Apple then?

Also appreciate the review 😄

zcbenz:
Anyway thanks for the detailed investigation on the problem! I'm closing this PR.

PR #3458, opened 2026-04-27
Reuse nightly build's ccache for release

zcbenz:
Since the release build of CUDA backend has become really slow (~5 hours) and often kills the CI server, we should test the release build nightly and reuse its cache for actual releases, so we don't have to rush out CI fixes on the release day.


angeloskath (approved):
👍

PR #3468, opened 2026-04-29
Fix JACCL GID selection on Apple Thunderbolt RDMA (errno 22 RTR fail)

### LINKED ISSUE #3467: [BUG] JACCL "Changing queue pair to RTR failed with errno 22" on Apple Thunderbolt RDMA — GID selection regression in #3412
## Describe the bug

JACCL distributed init fails on Apple Silicon Macs (M3/M4) connected via Thunderbolt 5 RDMA with `[jaccl] Changing queue pair to RTR failed with errno 22` (EINVAL). The runner crashes immediately on `ConnectToGroup`; the QP cannot transition to Ready-To-Receive state.

Affects every mlx version since the JACCL refactor (#3412, merged 2026-04-15, commit `4400504a`) — bug is still present in current `main` HEAD `e8ebdebe`.

## Root cause

In `mlx/distributed/jaccl/lib/jaccl/rdma.cpp`, `Connection::info()` selects the local GID for the QP destination handle. Pre-refactor, the code used `query_gid(ctx, 1, 1, &gid)` (always GID index 1). The refactor replaced this with a scan-and-filter loop:

```cpp
ibv_gid gid;
for (int i = 0; i < port_attr.gid_tbl_len; i++) {
  ibv_gid tmp;
  if (ibv().query_gid(ctx, 1, i, &tmp) == 0) {
    if (*(uint64_t*)&tmp.raw[0] == 0 && *(uint16_t*)&tmp.raw[8] == 0 &&
        *(uint16_t*)&tmp.raw[10] == 0xffff) {
      gid = tmp;
      break;
    }
  }
}
```

The filter accepts only **IPv4-mapped IPv6 GIDs** (the `::ffff:x.x.x.x` format used by RoCE v2). Apple Thunderbolt RDMA exposes only **link-local IPv6 GIDs** (`fe80::...`):

```
$ ibv_devinfo -d rdma_en3 -v
hca_id: rdma_en3
  transport: Thunderbolt (100)
  ...
  GID[0]: fe80::3474:d9ff:fe9d:cc84
  GID[1]: fe80::1042:a3d5:410e:95e9
```

Neither GID matches `raw[0..7]==0 && raw[10..11]==0xffff`. The loop never assigns `gid`, leaving it uninitialized. The garbage value is sent to the peer via the side channel, the peer programs its QP with garbage destination GID, and the kernel rejects the local QP's RTR transition with EINVAL.

## Reproduction

- 2 Mac Studio (M3 Ultra or M4 Max) connected via Thunderbolt 5
- macOS 26.x, RDMA enabled (`rdma_ctl status` → `enabled`), Thunderbolt Bridge disabled, individual `EXO Thunderbolt N` services per port
- Any `mlx.distributed.init(backend="jaccl")` over a 2-node mesh
- Crash: `ValueError: [jaccl] Changing queue pair to RTR failed with errno 22`

## Fix

Initialize `gid` to zero, prefer the IPv4-mapped GID (existing RoCE v2 behavior), and fall back to a non-zero GID — preferring index 1 (the original pre-refactor behavior, which corresponds to the RDMA port's primary link-local GID on Apple TB).

PR follows.

## Environment

- macOS 26.4.1 (build 25E253)
- Apple M4 Max, 2 nodes
- Thunderbolt 5 mesh, `rdma_en3 PORT_ACTIVE`, no Thunderbolt Bridge
- mlx commit `cc3f3e60` (rltakashige fork merged from upstream `e8ebdebe`)

## Related

- exo-explore/exo#1847 (jaccl errno 22/60/2 recurring crashes on M3 Ultra)
- exo-explore/exo#1711 (RDMA stops working — workaround was reboot/cable wiggle)
- exo-explore/exo#1973 (errno 96, similar pattern, maintainer suggested heartbeat)

danielkristofik:
Closes #3467.

## Proposed changes

Fix `[jaccl] Changing queue pair to RTR failed with errno 22` on Apple Thunderbolt RDMA.

`Connection::info()` in `rdma.cpp` selects the local GID by scanning the GID table for an IPv4-mapped IPv6 GID (`::ffff:x.x.x.x` — the RoCE v2 standard). Apple Thunderbolt RDMA exposes only link-local IPv6 GIDs (`fe80::...`), so the filter never matches and `gid` is left uninitialized. The garbage value is propagated to the peer via the side channel, causing the kernel to reject the QP RTR transition with EINVAL.

This regression was introduced by #3412 (Jaccl refactor) when the pre-existing hardcoded `query_gid(ctx, 1, 1, &gid)` was replaced with the filter loop. See full root-cause analysis in the linked issue.

## Changes

`mlx/distributed/jaccl/lib/jaccl/rdma.cpp`:
- Zero-initialize `gid` to avoid undefined behavior when no GID matches.
- After the IPv4-mapped scan, fall back to the first non-zero GID — preferring index 1 (matches pre-refactor behavior; index 0 on Apple TB is typically derived from a non-RDMA interface and routes elsewhere, causing errno 60 ETIMEDOUT instead of errno 22).

The IPv4-mapped path remains the preferred match, so RoCE v2 setups are unaffected.

## Test plan

- 2× Mac Studio M4 Max, Thunderbolt 5 mesh, macOS 26.4.1
- Before patch: `mlx.distributed.init(backend="jaccl")` → `errno 22 RTR fail`
- After patch (preferring index 0 in fallback): `errno 60 ETIMEDOUT` (wrong port — index 0 is non-RDMA on Apple)
- After patch (preferring index 1 in fallback): JACCL init succeeds, `all_sum` benchmark runs, sustained tensor-parallel inference (Qwen3.6-27B-4bit) works end-to-end

## Checklist

Note: no unit tests added — the GID selection is platform-dependent (Apple TB vs RoCE v2 hardware) and the tests in `mlx/distributed/jaccl/lib/examples/` are runtime benchmarks rather than unit tests. Happy to add a mock-based unit test if reviewers prefer.

### DIFF
--- a/mlx/distributed/jaccl/lib/jaccl/rdma.cpp
+++ b/mlx/distributed/jaccl/lib/jaccl/rdma.cpp
@@ -181,13 +181,61 @@ const Destination& Connection::info() {
 
   ibv_port_attr port_attr;
   ibv().query_port(ctx, 1, &port_attr);
-  ibv_gid gid;
-  for (int i = 0; i < port_attr.gid_tbl_len; i++) {
+  ibv_gid gid = {};
+
+  auto try_gid = [&](int i, bool require_ipv4_mapped) -> bool {
+    if (i < 0 || i >= port_attr.gid_tbl_len) {
+      return false;
+    }
     ibv_gid tmp;
-    if (ibv().query_gid(ctx, 1, i, &tmp) == 0) {
-      if (*(uint64_t*)&tmp.raw[0] == 0 && *(uint16_t*)&tmp.raw[8] == 0 &&
-          *(uint16_t*)&tmp.raw[10] == 0xffff) {
-        gid = tmp;
+    if (ibv().query_gid(ctx, 1, i, &tmp) != 0) {
+      return false;
+    }
+    if (require_ipv4_mapped) {
+      if (*(uint64_t*)&tmp.raw[0] != 0 || *(uint16_t*)&tmp.raw[8] != 0 ||
+          *(uint16_t*)&tmp.raw[10] != 0xffff) {
+        return false;
+      }
+    } else {
+      bool is_zero = true;
+      for (int j = 0; j < 16; j++) {
+        if (tmp.raw[j] != 0) {
+          is_zero = false;
+          break;
+        }
+      }
+      if (is_zero) {
+        return false;
+      }
+    }
+    gid = tmp;
+    return true;
+  };
+
+  // 1. Prefer IPv4-mapped IPv6 GID (RoCE v2 standard).
+  bool found = false;
+  for (int i = 0; i < port_attr.gid_tbl_len; i++) {
+    if (try_gid(i, /*require_ipv4_mapped=*/true)) {
+      found = true;
+      break;
+    }
+  }
+
+  // 2. Fallback for Apple Thunderbolt RDMA, which exposes only link-local
+  //    IPv6 (fe80::...) GIDs. Prefer index 1 (the actual rdma_enX port GID;
+  //    index 0 on Apple TB is typically derived from a non-RDMA interface
+  //    and routes elsewhere, surfacing as RTR errno 60 ETIMEDOUT).
+  if (!found) {
+    for (int i : {1, 0}) {
+      if (try_gid(i, /*require_ipv4_mapped=*/false)) {
+        found = true;
+        break;
+      }
+    }
+  }
+  if (!found) {
+    for (int i = 2; i < port_attr.gid_tbl_len; i++) {
+      if (try_gid(i, /*require_ipv4_mapped=*/false)) {
         break;
       }
     }


erwinzhang7:
I had a pair of M4 Pro minis on Thunderbolt 5 free, so I tried this against #3467.

It did not fix it here: the failure moves from `errno 22` to `errno 96 (ENODATA)` at the
same RTR transition. The link-local GID is selected, but the transition still fails.

One detail that may matter for the index-1 preference: without an IPv4 address, these ports publish
a single GID, at index 0.

```
rdma_en3:  gid_tbl_len: 1024
           GID[0] = fe80::347c:94ff:fecd:45c4
```

So there is no index 1 to prefer, and the fallback only works by dropping through to 0.

What did work was giving the interface an IPv4 address, after which an IPv4-mapped GID
appears at index 1 and stock MLX runs unmodified. Details in #3467. I suspect
the IPv6-only path may not be viable, though I have only tested Thunderbolt on
two M4 Pro machines.

zcbenz:
Thanks a lot for verifying the PR!

PR #3469, opened 2026-04-30
[CUDA] Fix half type matmul in cutlass kernels

zcbenz:
Delicate points when using CUTLASS:

* For half type matmul accumulated in float, we should make the main loop write float, and then make the epilogue convert it to half in output.
* Coordinates in CuTe are expected to be signed. (https://github.com/NVIDIA/cutlass/issues/3190)

The remaining tests are mostly blocked by:
* https://github.com/NVIDIA/cutlass/issues/3191.


angeloskath (approved):
👍

PR #3475, opened 2026-05-03
Compute contiguity from the actual occupied data

sofinvalery:
## Proposed changes

Compute `AsStrided` contiguity from the actual occupied data instead of only checking row or column contiguous layouts.

This allows dense non-row/column views, such as dense stride permutations, to be marked contiguous while keeping overlapping, gapped, broadcast-style, and negative-stride views conservative. It also computes a more accurate `data_size` for `as_strided` views.

## Checklist

### DIFF
--- a/mlx/backend/common/common.cpp
+++ b/mlx/backend/common/common.cpp
@@ -19,26 +19,27 @@ void AsStrided::eval(const std::vector<array>& inputs, array& out) {
         "AsStrided must be used with row contiguous arrays only.");
   }
 
-  // Compute the flags given the shape and strides
-  bool row_contiguous = true, col_contiguous = true;
-  size_t r = 1, c = 1;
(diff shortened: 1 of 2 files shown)


zcbenz (approved):
This looks good to me, but I'm not sure about the negative stride part. @angeloskath

angeloskath (approved):
Looks great! Thanks!

angeloskath:
> This looks good to me, but I'm not sure about the negative stride part. @angeloskath

Indeed the contiguous flag could in theory be set even with negative strides, however, the way we use it, it unfortunately cannot be. The contiguous flag would be used mostly for unary ops to select a fast path that doesn't care about indices. Starts at 0 and computes for data_size elements. However if there are negative strides, it should start at low and compute data_size elements.

Anyway, clean improvement in the PR as you also said.

PR #3476, opened 2026-05-04
[Python] Fix mx.array DLPack dispatch

apstenku123:
Fixes the existing mx.array(...) constructor path for DLPack producers/capsules instead of adding a new public workaround API.\n\nmx.array(obj) now detects raw DLPack PyCapsules and objects exposing __dlpack__ before falling through to the generic accessor conversion path, so DLPack producers do not hit to_array_with_accessor / bad_cast. __mlx_array__ remains higher precedence.\n\nThe shared consumer handles kDLCPU by copying into a fresh MLX allocation and kDLMetal by wrapping a shared MTLBuffer without copying. Unsupported devices, non-row-contiguous strides, non-zero byte offsets, invalid shapes/dtypes, and already-consumed capsules are rejected with explicit errors. Rejected capsules are left unconsumed.\n\nTests cover mx.array(...) from a DLPack producer, raw capsules, dtype override/default behavior, __mlx_array__ precedence, used capsule rejection, and preserving rejected capsules.

### DIFF
--- a/python/src/convert.cpp
+++ b/python/src/convert.cpp
@@ -6,6 +6,7 @@
 #include <nanobind/stl/complex.h>
 
 #include "python/src/convert.h"
+#include "python/src/dlpack_consumer.h"
 #include "python/src/utils.h"
 
 #include "mlx/utils.h"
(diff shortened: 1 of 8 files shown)


zcbenz (changes requested):
> mx.array(tvm_metal_tensor) falls into create_array → to_array_with_accessor and fails with bad_cast.

This would be a bug and should be fixed instead of introducing a new API that does the same thing. The constructor of `mx.array` is supposed to take dlpack.

python/src/convert.cpp
```diff
@@ -494,7 +495,13 @@ mx::array create_array(nb::object v, std::optional<mx::Dtype> t) {
   } else if (nb::isinstance<mx::array>(v)) {
     auto arr = nb::cast<mx::array>(v);
     return mx::astype(arr, t.value_or(arr.dtype()));
-  } else if (nb::ndarray_check(v)) {
+  }
+
+  const bool has_mlx_array = nb::hasattr(v, "__mlx_array__");
+  const bool is_dlpack =
+      PyCapsule_CheckExact(v.ptr()) || nb::hasattr(v, "__dlpack__");
```
zcbenz, line 502: The `ndarray` API of nanobind is supposed to handle dlpack:

https://github.com/wjakob/nanobind/blob/acc88f3f91d09889c919ca5ea05cf604c4fce102/src/nb_ndarray.cpp#L489-L492

zcbenz:
I'm closing this PR in favor of #3531.

PR #3485, opened 2026-05-05
Implement output_shapes for GatherMM and GatherQMM

dexwritescode:
## Problem

`mx::compile(shapeless=true)` calls `Primitive::output_shapes()` on every
node when re-tracing a compiled function after input shapes change. `GatherQMM`
was missing this override, causing compiled functions that contain it to throw:

```
[Primitive::output_shapes] GatherQMM cannot infer output shapes
```

## Fix

The output shape is fully inferrable from the stored fields and input shapes:
```
out_shape = lhs_indices.shape() + [x.shape(-2), w_outer_dims]
```
where `w_outer_dims = transpose ? w.shape(-2) : w.shape(-1) * 32 / bits`.

Input layout differs by quantization mode: Affine mode has biases at index 3,
pushing `lhs_indices` to index 4; other modes have `lhs_indices` at index 3.

## Testing

Added `test_shapeless_compile_gather_qmm` to `python/tests/test_compile.py`:
compiles a `gather_qmm` call with `shapeless=True` and verifies it reuses the
graph correctly when the `M` dimension changes between calls.

### DIFF
--- a/mlx/backend/cuda/custom_kernel.cpp
+++ b/mlx/backend/cuda/custom_kernel.cpp
@@ -223,7 +223,7 @@ CustomKernelFunction cuda_kernel(
     }
 
     return array::make_arrays(
-        output_shapes,
+        std::move(output_shapes),
         std::move(output_dtypes),
         std::make_shared<CustomKernel>(
             s,
@@ -236,8 +236,7 @@ CustomKernelFunction cuda_kernel(
             init_value,
             std::vector<ScalarArg>{},
             false,
-            shared_memory,
-            output_shapes),
+            shared_memory),
         std::move(inputs));
   };
 }
@@ -271,8 +270,7 @@ std::vector<array> precompiled_cuda_kernel(
           init_value,
           scalars,
           true,
-          shared_memory,
-          output_shapes),
+          shared_memory),
       inputs);
 }
 
--- a/mlx/backend/cuda/custom_kernel.cpp
+++ b/mlx/backend/cuda/custom_kernel.cpp
@@ -222,6 +222,7 @@ CustomKernelFunction cuda_kernel(
                 << "```" << std::endl;
     }
 
+    auto output_shapes_copy = output_shapes;
     return array::make_arrays(
         std::move(output_shapes),
         std::move(output_dtypes),
@@ -236,7 +237,8 @@ CustomKernelFunction cuda_kernel(
             init_value,
             std::vector<ScalarArg>{},
             false,
-            shared_memory),
+            shared_memory,
+            std::move(output_shapes_copy)),
         std::move(inputs));
   };
 }
@@ -270,7 +272,8 @@ std::vector<array> precompiled_cuda_kernel(
           init_value,
           scalars,
           true,
-          shared_memory),
+          shared_memory,
+          output_shapes),
       inputs);
 }
 
--- a/mlx/backend/cuda/custom_kernel.cpp
+++ b/mlx/backend/cuda/custom_kernel.cpp
@@ -224,7 +224,7 @@ CustomKernelFunction cuda_kernel(
 
     auto output_shapes_copy = output_shapes;
     return array::make_arrays(
-        std::move(output_shapes),
+        output_shapes,
         std::move(output_dtypes),
         std::make_shared<CustomKernel>(
             s,
--- a/mlx/backend/cuda/custom_kernel.cpp
+++ b/mlx/backend/cuda/custom_kernel.cpp
@@ -222,7 +222,6 @@ CustomKernelFunction cuda_kernel(
                 << "```" << std::endl;
     }
 
-    auto output_shapes_copy = output_shapes;
     return array::make_arrays(
         output_shapes,
         std::move(output_dtypes),
@@ -238,7 +237,7 @@ CustomKernelFunction cuda_kernel(
             std::vector<ScalarArg>{},
             false,
             shared_memory,
-            std::move(output_shapes_copy)),
+            output_shapes),
         std::move(inputs));
   };
 }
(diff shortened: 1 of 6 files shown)


mlx/backend/cuda/custom_kernel.cpp
```diff
@@ -236,7 +237,8 @@ CustomKernelFunction cuda_kernel(
             init_value,
             std::vector<ScalarArg>{},
             false,
-            shared_memory),
+            shared_memory,
+            std::move(output_shapes_copy)),
```
zcbenz: `output_shapes` is a const reference so we should just copy, it is surprising previous `std::move(output_shapes)` actually compiles.

mlx/primitives.h
```diff
@@ -1700,6 +1700,21 @@ class GatherQMM : public UnaryPrimitive {
   DEFINE_GRADS()
   DEFINE_NAME(GatherQMM)
   bool is_equivalent(const Primitive& other) const override;
+
+  // inputs layout: Affine → {x, w, scales, biases, lhs_idx, rhs_idx}
+  //                other  → {x, w, scales,          lhs_idx, rhs_idx}
```
zcbenz:
```suggestion
  // inputs layout: Affine → {x, w, scales, biases, lhs_idx, rhs_idx}
  //                other  → {x, w, scales,         lhs_idx, rhs_idx}
```

zcbenz (commented):
The change overall looks good to me, can you add some simple tests?

mlx/backend/cuda/custom_kernel.cpp
dexwritescode: Fixed — removed `std::move` on the const ref in both Metal and CUDA backends. `output_shapes` is now passed directly to `array::make_arrays` (plain copy), while the local `output_shapes_copy` is still moved into `CustomKernel`.

mlx/primitives.h
dexwritescode: Fixed — removed the extra space.

dexwritescode (commented):
Added two tests to `python/tests/test_compile.py`:

- `test_shapeless_compile_custom_kernel` — compiles a `metal_kernel` passthrough with `shapeless=True`, then calls it with a larger shape. Fails before the fix with `CustomKernel cannot infer output shapes`.
- `test_shapeless_compile_gather_qmm` — compiles a `gather_qmm` call with `shapeless=True`, then calls it with a different `M` dimension. Fails before the fix with `GatherQMM cannot infer output shapes`.

Also addressed the two inline comments (removed `std::move` on const ref in both Metal and CUDA backends, fixed comment alignment in `primitives.h`).

mlx/backend/cuda/custom_kernel.cpp
zcbenz: It seems that your later commit reverted the change?

dexwritescode: pushed the corrected change

dexwritescode:
Hey @zcbenz, I've pushed a change to fix a test that failed. Would appreciate if you can trigger the CI run. Thanks!

zcbenz (approved):
Looks good to me, thanks!

angeloskath (changes requested):
Sorry but this is not correct. You can't simply pass the output shapes and then compile in shapeless mode the output shape will most likely change based on the input shape.

Take for instance the following code to see what the problem is

```python
import mlx.core as mx
from functools import partial
# A "double" kernel: output has the same shape as input
double = mx.fast.metal_kernel(
    name="double",
    input_names=["inp"],
    output_names=["out"],
    source="uint i = thread_position_in_grid.x; out[i] = inp[i] * 2;",
)
def fn(x):
    return double(
        inputs=[x],
        grid=(x.size, 1, 1),
        threadgroup=(256, 1, 1),
        output_shapes=[x.shape],  # output shape == input shape
        output_dtypes=[x.dtype],
        stream=mx.gpu,
    )[0]
cfn = mx.compile(fn, shapeless=True)
x = mx.array([1.0, 2.0, 3.0])
print(cfn(x))  # [2, 4, 6] — shape (3,), correct
x = mx.array([1.0, 2.0, 3.0, 4.0, 5.0])
print(cfn(x))  # BUG: output_shapes() returns (3,) from the first trace,
               # but the correct output shape should be (5,)
```

feel free to remove the custom kernel and I will look into the gather qmm one.

dexwritescode:
> Sorry but this is not correct. You can't simply pass the output shapes and then compile in shapeless mode the output shape will most likely change based on the input shape.
> 
> Take for instance the following code to see what the problem is
> 
> ```python
> import mlx.core as mx
> from functools import partial
> # A "double" kernel: output has the same shape as input
> double = mx.fast.metal_kernel(
>     name="double",
>     input_names=["inp"],
>     output_names=["out"],
>     source="uint i = thread_position_in_grid.x; out[i] = inp[i] * 2;",
> )
> def fn(x):
>     return double(
>         inputs=[x],
>         grid=(x.size, 1, 1),
>         threadgroup=(256, 1, 1),
>         output_shapes=[x.shape],  # output shape == input shape
>         output_dtypes=[x.dtype],
>         stream=mx.gpu,
>     )[0]
> cfn = mx.compile(fn, shapeless=True)
> x = mx.array([1.0, 2.0, 3.0])
> print(cfn(x))  # [2, 4, 6] — shape (3,), correct
> x = mx.array([1.0, 2.0, 3.0, 4.0, 5.0])
> print(cfn(x))  # BUG: output_shapes() returns (3,) from the first trace,
>                # but the correct output shape should be (5,)
> ```
> 
> feel free to remove the custom kernel and I will look into the gather qmm one.

Hey @angeloskath, I've removed the custom kernel and kept the qmm.

angeloskath (approved):
Added the GatherMM output_shapes as well. Looks good.

PR #3488, opened 2026-05-06
test: Upcast the random numbers before computing their average

### LINKED ISSUE #3326: "test random uniform" fails for CPU backend
On macOS/Windows the CPP tests would fail with:

```console
$ DEVICE=cpu ./build/tests/tests -tc="test random uniform"
[doctest] doctest version is "2.4.12"
[doctest] run with "--help" for options
===============================================================================
/Users/zcbenz/codes/mlx/tests/random_tests.cpp:244:
TEST CASE:  test random uniform

/Users/zcbenz/codes/mlx/tests/random_tests.cpp:368: ERROR: CHECK( abs(float(mean(out).item<bfloat16_t>()) - 0.5f) < 0.02 ) is NOT correct!
  values: CHECK( 0.244141 <  0.02 )

===============================================================================
[doctest] test cases:  1 |  0 passed | 1 failed | 239 skipped
[doctest] assertions: 33 | 32 passed | 1 failed |
[doctest] Status: FAILURE!
```

which was caused by https://github.com/ml-explore/mlx/pull/2361, but I'm not exactly sure where the bug is. /cc @angeloskath

sofinvalery:
## Proposed changes

Accumulate CPU float16 and bfloat16 sum reductions in float32, while preserving the output dtype. This fixes precision loss in ops that use sum().

Fixes #3326.

## Checklist

### DIFF
--- a/.github/actions/test-windows/action.yml
+++ b/.github/actions/test-windows/action.yml
@@ -16,6 +16,6 @@ runs:
         DEVICE: cpu
       run: |
         echo "::group::CPP tests - CPU"
-        ./build/tests.exe -tce="*gguf*,test random uniform"
+        ./build/tests.exe -tce="*gguf*"
         ./build/test_teardown.exe
         echo "::endgroup::"
--- a/tests/random_tests.cpp
+++ b/tests/random_tests.cpp
@@ -355,7 +355,7 @@ TEST_CASE("test random uniform") {
     CHECK(all(less(out, array(1.0f))).item<bool>());
     CHECK(all(greater_equal(out, array(0.0f))).item<bool>());
     CHECK(!all(equal(out, array(0.0f))).item<bool>());
-    CHECK(abs(float(mean(out).item<float16_t>()) - 0.5f) < 0.02);
+    CHECK(abs(mean(astype(out, float32)).item<float>() - 0.5f) < 0.02);
   }
 
   {
@@ -365,7 +365,7 @@ TEST_CASE("test random uniform") {
     CHECK(all(less(out, array(1.0f))).item<bool>());
     CHECK(all(greater_equal(out, array(0.0f))).item<bool>());
     CHECK(!all(equal(out, array(0.0f))).item<bool>());
-    CHECK(abs(float(mean(out).item<bfloat16_t>()) - 0.5f) < 0.02);
+    CHECK(abs(mean(astype(out, float32)).item<float>() - 0.5f) < 0.02);
   }
 }
 


sofinvalery:
Made it inline. Feels cleaner.

zcbenz (changes requested):
Allocating a new array would have heavy performance penalty, the correct way would be refactoring `strided_reduce`/`contiguous_reduce` to accumulate in float32 rather than the output type.

sofinvalery:
Refactored `strided_reduce`/`contiguous_reduce` to support accumulation in a separate type.
float16/bfloat16 sums now accumulate in float32 without allocating a temp array.
Also added tests for reductions across full, contiguous, and strided axes.

mlx/backend/cpu/reduce.cpp
```diff
@@ -135,27 +146,25 @@ void nd_loop(
...
+    *out_ptr = contiguous_reduce(in_ptr, x.size(), Op{}, init, init);
     return;
   }
 
   if (plan.type == ContiguousReduce && plan.shape.size() == 1) {
     int reduction_size = plan.shape[0];
     for (int i = 0; i < out.size(); i++, out_ptr++, in_ptr += reduction_size) {
-      *out_ptr = init;
-      contiguous_reduce(in_ptr, out_ptr, reduction_size, Op{}, init);
+      *out_ptr = contiguous_reduce(in_ptr, reduction_size, Op{}, init, init);
```
zcbenz: Is it supposed to be:

```suggestion
      *out_ptr = contiguous_reduce(in_ptr, reduction_size, Op{}, *out_ptr, init);
```

mlx/backend/cpu/reduce.cpp
```diff
@@ -96,19 +97,29 @@ void strided_reduce(
...
+  }
+  return accumulator;
+}
+
+template <typename T, typename U, typename Op>
+U contiguous_reduce(const T* x, int size, Op op, U accumulator, U init) {
   constexpr int N = std::min(simd::max_size<T>, simd::max_size<U>);
   simd::Simd<U, N> accumulator_v(init);
   while (size >= N) {
     accumulator_v = op(accumulator_v, simd::Simd<U, N>(simd::load<T, N>(x)));
```
zcbenz, line 103: When building for macOS 14 it seems that we can't simply convert `Simd<float16>` to `Simd<float>`:

```
  /Users/runner/actions-runner/_work/mlx/mlx/mlx/backend/cpu/simd/accelerate_simd.h:63:40: error: no matching function for call to 'convert'
     63 |   Simd<T, N>(Simd<U, N> other) : value(asd::convert<scalar_t>(other.value)) {}
        |                                        ^~~~~~~~~~~~~~~~~~~~~~
  /Users/runner/actions-runner/_work/mlx/mlx/mlx/backend/cpu/reduce.cpp:113:39: note: in instantiation of function template specialization 'mlx::core::simd::Simd<float, 8>::Simd<__fp16>' requested here
    113 |     accumulator_v = op(accumulator_v, simd::Simd<U, N>(simd::load<T, N>(x)));
        |                                       ^
```

mlx/backend/cpu/reduce.cpp
```diff
@@ -404,9 +450,9 @@ void reduce_dispatch_and_or(
     Reduce::ReduceType rtype,
     const std::vector<int>& axes) {
   if (rtype == Reduce::And) {
-    reduction_op<InT, bool, AndReduce>(in, out, axes, true);
+    reduction_op<InT, bool, AndReduce, bool>(in, out, axes, true);
```
zcbenz: I prefer omitting the optional `AccT` so it would be obvious when the code is accumulating in a different type.

angeloskath:
I don't think we should merge this tbh. It's kind of expected for floats that you get the accumulation precision of the float you are in.

I am not against choosing high precision temporaries internally if possible but given the current implementation it is actually faster to cast to float32 and then sum (didn't try it but almost certainly it will be).

For the test by the way, the correct fix is to simply upcast the random numbers before computing their average. We want to test random after all not sum.

sofinvalery:
Updated the test only. Thanks for the feedback! @zcbenz @angeloskath

zcbenz (approved):
Thanks for fixing the test! Can you also update `.github/actions/test-windows/action.yml` to enable the test in CI?

PR #3490, opened 2026-05-06
[CUDA] Fix QMM boundary handling for scales and biases

### LINKED ISSUE #3487: [BUG][CUDA] Illegal memory access in QMM when M/N is not multiple of tile size
The kernel is not checking M/N boundary when reading scales/biases, the fix would be quite straightforward by make predicate arrays and then use `copy_if` for them but it is very tricky to get right due to the layout of scales/biases.

https://github.com/ml-explore/mlx/blob/80bcd1c65867dec1623546d1fdc9d0473ab46657/mlx/backend/cuda/quantized/qmm/qmm_naive.cu#L159-L164

I ignored the problem since it is usually not a big issue but recently it makes the tests a bit flaky.

abhinavgautam01:
Fixes #3487

## Summary

This updates the CUDA naive QMM kernel to predicate scale and bias loads with the same `tBpB` mask already used for quantized weight loads.

Previously, edge tiles where `N` was not a multiple of the tile size could still perform unconditional loads for scales/biases, which could lead to illegal memory access and flaky tests.

## Changes

- Use `copy_if(...)` instead of unconditional `copy(...)` for scale loads in `qmm_naive.cu`
- Use `copy_if(...)` instead of unconditional `copy(...)` for bias loads in `qmm_naive.cu`
- Apply the same fix in the residue preload path
- Add a regression test for a non-multiple `N` case intended to exercise the `qmm` path

## Notes

I could not validate the CUDA path locally on this machine, so CUDA verification still needs to be run in a Linux/NVIDIA environment.

## Checklist

### DIFF
--- a/mlx/backend/cuda/quantized/qmm/qmm_naive.cu
+++ b/mlx/backend/cuda/quantized/qmm/qmm_naive.cu
@@ -159,8 +159,8 @@ __global__ void qmm_naive_kernel(
   auto fetch_gmem = [&](int tile) {
     copy_if(copy_a, tApA, tAgA(_,_,_,tile), tArA);
     copy_if(copy_b, tBpB, tBgB(_,_,_,tile), tBrB);
-    copy(tBgS(_,_,_,tile), tBrS);
-    copy(tBgZ(_,_,_,tile), tBrZ);
+    copy_if(copy_b, tBpB, tBgS(_,_,_,tile), tBrS);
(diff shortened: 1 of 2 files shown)


zcbenz:
> I could not validate the CUDA path locally on this machine, so CUDA verification still needs to be run in a Linux/NVIDIA environment.

🤷

PR #3495, opened 2026-05-08
Improve DLPack-compatible array imports

XXXXRT666:
## Proposed changes

This PR improves `mx.array` import behavior for DLPack-compatible inputs.

Changes include:

- Update the `mx.array` constructor signature to use a `DLPackCompatible` protocol instead of naming `numpy.ndarray` directly.
- Reject non-CPU DLPack inputs consistently, including both `mx.array(...)` and operator argument conversion paths.
- Add test coverage for PyTorch MPS tensors to ensure non-CPU DLPack inputs raise an explicit error.
- Update the PyTorch interoperability docs to use DLPack directly via `torch.tensor(a)` and then move the tensor to CPU before importing it back into MLX.

PyTorch supports the DLPack import path directly: `torch.tensor(data)` checks for `__dlpack__` and routes through `torch.utils.dlpack.from_dlpack` in [`tensor_new.cpp`](https://github.com/pytorch/pytorch/blob/7d9afa6e839cead35485c328d36f91b1eabf317e/torch/csrc/utils/tensor_new.cpp#L347-L363).

This PR intentionally does not add direct Metal buffer import or zero-copy semantics. That path depends on nanobind support that is not upstream yet: ndarray import needs to request the producer's DLPack device with `dl_device` and `copy=False` for non-CPU devices, and MLX needs access to the raw DLPack `data` handle and `byte_offset` to wrap the Metal allocation directly.

## Checklist

### DIFF
--- a/python/src/convert.cpp
+++ b/python/src/convert.cpp
@@ -501,17 +501,15 @@ mx::array create_array(nb::object v, std::optional<mx::Dtype> t) {
     return mx::astype(arr, t.value_or(arr.dtype()));
   } else if (nb::ndarray_check(v)) {
     using ContigArray = nb::ndarray<nb::ro, nb::c_contig>;
-    nb::object src = v;
     ContigArray nd;
     std::optional<nb::dlpack::dtype> nb_dtype;
     // Nanobind does not recognize bfloat16 numpy array:
     // https://github.com/wjakob/nanobind/discussions/560
-    if (nb::hasattr(src, "dtype") &&
-        src.attr("dtype").equal(nb::str("bfloat16"))) {
-      nd = nb::cast<ContigArray>(src.attr("view")("uint16"));
+    if (nb::hasattr(v, "dtype") && v.attr("dtype").equal(nb::str("bfloat16"))) {
+      nd = nb::cast<ContigArray>(v.attr("view")("uint16"));
       nb_dtype = nb::dtype<mx::bfloat16_t>();
     } else {
-      nd = nb::cast<ContigArray>(src);
+      nd = nb::cast<ContigArray>(v);
     }
     return nd_array_to_mlx(nd, t, nb_dtype);
   } else {
--- a/python/src/convert.cpp
+++ b/python/src/convert.cpp
@@ -33,7 +33,7 @@ int check_shape_dim(int64_t dim) {
 
 template <typename T>
 mx::array nd_array_to_mlx_contiguous(
-    nb::ndarray<nb::ro, nb::c_contig, nb::device::cpu> nd_array,
+    nb::ndarray<nb::ro, nb::c_contig> nd_array,
     const mx::Shape& shape,
     mx::Dtype dtype) {
   // Make a copy of the numpy buffer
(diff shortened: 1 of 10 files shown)


zcbenz (commented):
I'm good with most changes in this PR but we should not silently convert GPU tensors to CPU, because it would be a surprising behavior for most users, and would become a breaking change when we support taking GPU tensors in future.

I think a better choice is to simply throw an error when a non-CPU tensor is passed.

XXXXRT666:
> I think a better choice is to simply throw an error when a non-CPU tensor is passed.

That makes sense. I can keep this PR from silently converting GPU tensors to CPU.

> I'm good with most changes in this PR but we should not silently convert GPU tensors to CPU, because it would be a surprising behavior for most users, and would become a breaking change when we support taking GPU tensors in future.

After [this PR](https://github.com/wjakob/nanobind/pull/1338) lands, I can add proper GPU tensor support in a follow up PR. At that point we should be able to support zero-copy where the producer exposes a compatible GPU buffer, instead of falling back to an implicit CPU copy.

python/src/convert.cpp
```diff
@@ -210,6 +210,23 @@ nb::object to_scalar(mx::array& a) {
...
+    return;
+  }
+
+  auto device = nb::cast<nb::tuple>(v.attr("__dlpack_device__")());
+  if (nb::len(device) != 2) {
+    throw std::invalid_argument("Invalid DLPack device.");
+  }
+
+  auto device_type = nb::cast<int>(device[0]);
+  if (device_type != nb::device::cpu::value) {
```
zcbenz: Can this be checked with the `device_type()` member of `nb::ndarray`?

XXXXRT666: Done. It now uses `nb::ndarray::device_type()` after casting to a generic contiguous ndarray.

XXXXRT666:
Updated this PR to use `nb::ndarray::device_type()` after casting once to a generic contiguous ndarray.

The non-CPU guard now lives in `nd_array_to_mlx()`, so all CPU-copy paths reject non-CPU DLPack inputs consistently. That covers both `mx.array(torch_mps_tensor)` and operator argument conversion such as `mx.array([1]) + torch.tensor([2]).to("mps")`. I also removed the float64 import change from the PR since metal doesn't support fp64 computing.

python/src/convert.cpp
```diff
@@ -495,16 +500,18 @@ mx::array create_array(nb::object v, std::optional<mx::Dtype> t) {
     auto arr = nb::cast<mx::array>(v);
     return mx::astype(arr, t.value_or(arr.dtype()));
   } else if (nb::ndarray_check(v)) {
-    using ContigArray = nb::ndarray<nb::ro, nb::c_contig, nb::device::cpu>;
+    using ContigArray = nb::ndarray<nb::ro, nb::c_contig>;
+    nb::object src = v;
```
zcbenz: The copy seems redundant here?

zcbenz (approved):
Nice work, thanks!

python/src/convert.cpp
XXXXRT666: Done

XXXXRT666: Done

Issue #3496, opened 2026-05-07
MLXNN.RoPE produces row-asymmetric output for byte-identical row inputs at [B>1, H, 1, D]


ivan-digital:
## Summary

`MLXNN.RoPE` (which dispatches into `MLXFast.RoPE`) on a `[B, H, 1, D]` input where row 0 == row 1 produces row-asymmetric output. The same op on `[B, H, T, D]` with `T > 1` is row-symmetric. The same op with `B = 1` is deterministic and correct.

This blocks any straightforward `[B,1,H]` batched autoregressive decode in mlx-swift — every decoder layer's attention applies RoPE per step at `T=1`, and the resulting per-row drift compounds through the layers fast enough that argmax tokens diverge within a few decode steps.

## Minimal reproducer (no model, no quantization)

```swift
import MLX
import MLXNN
import MLXRandom

MLXRandom.seed(42)

let rope = MLXNN.RoPE(dimensions: 128, traditional: false, base: 1_000_000)

// Row-symmetric input: row 0 byte-equal to row 1.
let baseRow = MLXRandom.normal([1, 8, 1, 128])  // [1, H, T=1, D]
let input = concatenated([baseRow, baseRow], axis: 0)  // [2, H, T=1, D]

// Sanity: input rows are identical.
let i0 = input[0..<1, 0..., 0..., 0...].asArray(Float.self)
let i1 = input[1..<2, 0..., 0..., 0...].asArray(Float.self)
let inputDiff = zip(i0, i1).map { abs($0 - $1) }.max() ?? 0
assert(inputDiff == 0)

let output = rope(input, offset: 0)

let r0 = output[0..<1, 0..., 0..., 0...].asArray(Float.self)
let r1 = output[1..<2, 0..., 0..., 0...].asArray(Float.self)
let rowDiff = zip(r0, r1).map { abs($0 - $1) }.max() ?? 0

print("RoPE B=2 T=1 offset=0 row-diff: \(rowDiff)")
// Expected: 0.0
// Actual:   5.379... (varies by seed but always non-zero)
```

## Diagnostic matrix

| Shape | Op | Offset | row 0 vs row 1 maxDiff |
|---|---|---|---|
| `[2, 1, 1024]` | `quantizedMatmul` (4-bit, group 64) | — | **0.0** ✓ |
| `[2, 148, 1024]` | `quantizedMatmul` (4-bit, group 64) | — | **0.0** ✓ |
[...]

PR #3498, opened 2026-05-08
Fix rope single token multiple sequences

### LINKED ISSUE #3494: [BUG] mx.fast.rope produces wrong output for batch rows >= 1 when seq_len == 1
## Summary

`mx.fast.rope` returns correct output only for batch row 0 when input has `seq_len == 1` and `batch_size >= 2`, even when all batch rows are bit-identical. Rows beyond row 0 collapse to mostly-zeros (or other garbage; exact failure mode varies with input values). This breaks the incremental decode path (one new token per step against a populated KV cache) for any transformer at `batch_size > 1` — every Llama / Qwen / Mistral-family model in mlx-lm is affected when used with batched generation.

## Minimal reproduction

```python
import mlx.core as mx

mx.random.seed(42)
single = mx.random.normal(shape=(1, 4, 1, 128)).astype(mx.float16)
mx.eval(single)
# Two rows, independently allocated, with bit-identical values
batched = mx.array([single.tolist()[0], single.tolist()[0]], dtype=mx.float16)
assert float(mx.max(mx.abs(batched[0] - batched[1])).item()) == 0.0

out = mx.fast.rope(
    batched, dims=128, traditional=False, base=10000.0, scale=1.0, offset=5,
)
mx.eval(out)

print("row 0:", out[0, 0, 0, :6].tolist())
print("row 1:", out[1, 0, 0, :6].tolist())
print("row diff:", float(mx.max(mx.abs(out[0] - out[1])).item()))
```

Output:
```
row 0: [-0.2481689453125, -1.2119140625, -1.8681640625, -0.1610107421875, -0.31787109375, -0.471923828125]
row 1: [0.0, 0.0, 0.0, 0.0, 0.0, 0.0]
row diff: 3.142578125
```

Per-row separate calls (`mx.fast.rope(batched[0:1], ...)` and `mx.fast.rope(batched[1:2], ...)`) produce identical correct output for both rows, confirming the inputs are equivalent and the bug is in the batched kernel path.

## Expected behavior

Identical input rows -> identical output rows. RoPE is a deterministic per-position transformation; nothing about it should depend on batch index.

## Actual behavior

Only batch row 0 receives correct RoPE rotation. Rows 1+ receive output dominated by zeros.

## Trigger conditions

| Condition | Triggers? |
|---|---|
| `batch >= 2`, `seq_len == 1`, any offset (0 or nonzero), any `traditional` | **Yes** |
| `batch >= 2`, `seq_len >= 2` (prefill-shape) | No |
| `batch == 1`, `seq_len == 1` | No |
| dtype = float16 / float32 / bfloat16 | All triggered |
| Batched tensor built via `mx.repeat(single, B, axis=0)` (aliased rows) | No |
| Independently-allocated batched tensor with bit-identical row values | **Yes** |

The aliased-storage masking is what made this hard to notice. Synthetic tests using `mx.repeat` to build batched-identical inputs don't trigger; only real-model paths that produce non-aliased value-equal rows (e.g., `linear(x).reshape(...).transpose(...)` where `x` has aliased rows) trigger.

## How I encountered this

Implementing multi-sample semantic-entropy generation in mlx-lm (batched prefill at `batch_size = N`, then per-step batched decode at `batch_size = N, seq_len = 1`). Even with bit-identical prefix replicated across rows and identical post-prefill cache state across rows (verified `max-abs-diff = 0.0` on cache.keys for all layers), generated samples diverge from step 1 onwards. Bisected through the model via attention-layer instrumentation:

- Layer 0 input `x` after embedding: rows identical (max-abs-diff = 0.0)
- Layer 0 `v_proj(x)` output: rows identical (V doesn't go through RoPE)
- Layer 0 `k_proj(x)` reshaped+transposed to `(B, n_kv_heads, 1, head_dim)`: rows identical (verified before RoPE call)
- Layer 0 RoPE output on the K input: rows diverge by ~170

Then I confirmed the divergence reproduces with `mx.fast.rope` directly (above) and is independent of any model-specific code path.

## Impact

Breaks any incremental-decode workload at `batch_size >= 2`, which is the standard shape for batched generation. Every transformer architecture in mlx-lm is affected (Llama, Qwen, Mistral, Gemma, etc.) when used in batched-decode mode. Workaround until fixed: serial single-batch decode per sample (loses batched-decode speedup but preserves correctness).

## Environment

- mlx-core: 0.31.2
- mlx-lm: 0.31.2 (manifests through Qwen2.5-7B-Instruct-4bit during testing; not Qwen-specific)
- macOS: 26.4.1, arm64 (Apple Silicon, M1 Pro)
- Python: 3.11

angeloskath:
Fixes #3494, #3496 .

Crazy that this was a bug for more than a year and nobody encountered it (nobody noticed)...

### DIFF
--- a/mlx/backend/metal/rope.cpp
+++ b/mlx/backend/metal/rope.cpp
@@ -133,8 +133,9 @@ void RoPE::eval_gpu(
   if (single) {
     compute_encoder.set_bytes(out_strides, 1, 4);
     uint32_t dim0 = dims_ / 2;
-    group_dims = get_block_dims(dim0, N, 1);
-    grid_dims = MTL::Size(dim0, N, 1);
+    uint32_t dim1 = B * N;
+    group_dims = get_block_dims(dim0, dim1, 1);
+    grid_dims = MTL::Size(dim0, dim1, 1);
   } else {
     compute_encoder.set_bytes(strides, 3, 4);
     compute_encoder.set_bytes(out_strides, 3, 5);
--- a/python/tests/test_fast.py
+++ b/python/tests/test_fast.py
@@ -365,6 +365,28 @@ class TestFast(mlx_tests.MLXTestCase):
         rx = rope_orig(x, dims, traditional, base, scale, offset)
         self.assertLess(mx.abs(rx - rx_fast).max(), 1e-5)
 
+    def test_rope_single_batch(self):
+        base = 10000.0
+        scale = 1.0
+        offset = 5
+
+        for traditional in [True, False]:
+            for B in [2, 4, 8]:
+                for n_head in [1, 4, 7]:
+                    for dims in [64, 128]:
+                        x = mx.random.uniform(shape=(B, n_head, 1, dims))
+                        mx.eval(x)
+                        rx_fast = mx.fast.rope(
+                            x,
+                            dims,
+                            traditional=traditional,
+                            base=base,
+                            scale=scale,
+                            offset=offset,
+                        )
+                        rx = rope_orig(x, dims, traditional, base, scale, offset)
+                        self.assertLess(mx.abs(rx - rx_fast).max(), 1e-5)
+
     def test_rope_with_large_offset(self):
         x = mx.random.normal(shape=(1, 1, 1024, 32))
         rx_fp32 = mx.fast.rope(


zcbenz (approved):
👍

PR #3501, opened 2026-05-09
Pass deployment target when linking metallib

dhiltgen:
MLX compiles each Metal kernel AIR file with -mmacosx-version-min from CMAKE_OSX_DEPLOYMENT_TARGET, but the final metal driver invocation did not receive that deployment target.

With the macOS 26 SDK and CMAKE_OSX_DEPLOYMENT_TARGET=14.0, the metal driver otherwise invokes air-lld with the SDK default platform version and can stamp mlx.metallib as macOS 26 even though the input AIR files target macOS 14.

Repro:
```
cmake -S . -B build-macos14 -DCMAKE_BUILD_TYPE=Release -DCMAKE_OSX_DEPLOYMENT_TARGET=14.0

cmake --build build-macos14 --target mlx-metallib --parallel

strings -a build-macos14/mlx/backend/metal/kernels/mlx.metallib | grep -Eo 'air64[^[:space:]]*macosx[0-9.]+' | sort -u
```
Keep using metal as the linker front-end and pass -mmacosx-version-min through the final link step. Verified with macOS SDK 26.4 that the resulting library reports air64_v26-apple-macosx14.0.0.

## Proposed changes

## Checklist

### DIFF
--- a/mlx/backend/metal/kernels/CMakeLists.txt
+++ b/mlx/backend/metal/kernels/CMakeLists.txt
@@ -177,9 +177,14 @@ if(NOT MLX_METAL_JIT)
 
 endif()
 
+set(METAL_LINK_FLAGS)
+if(NOT CMAKE_OSX_DEPLOYMENT_TARGET STREQUAL "")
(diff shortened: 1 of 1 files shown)


zcbenz (approved):
Looks good to me, thanks!

PR #3503, opened 2026-05-10
[CUDA] Fix gather_mm

zcbenz:
The sm70/sm80 kernels of CUTLASS v3 are in a weird state: no one uses them, no one maintains them, they have bugs and I had a really hard time trying to find a maintainer who actually understand the kernels to look into the bugs.

Anyway it is much easier to just write the kernel in plain CuTe and it should be better for long term maintenance since the code no longer requires domain-specific knowledge of CUTLASS C++ APIs.

This PR removes the last entries in `cuda_skip.py‎` and I'm deleting the file, for future TODO items in CUDA backend we can just use `@unittest.skipIf`.


angeloskath (approved):
Awesome!

Issue #3484, opened 2026-05-05
[BUG] Missing `array-api` methods


ilan-gold:
**Describe the bug**
The current array-api implementation is currently missing a few methods

**To Reproduce**

***`from_dlpack`***:
```python
import mlx.core as mx

mx.array([[0, 1], [2, 3]]).__array_namespace__().from_dlpack
```

***`as_array(..., copy=...)`***:

```python
import mlx.core as mx

mx.array([[0, 1], [2, 3]]).__array_namespace__().asarray(mx.array([[0, 1], [2, 3]]), copy=True)
````

**Expected behavior**
I would expect the above to work.  I assume there are others as well.

**Desktop (please complete the following information):**
 - OS Version: 15.1 (24B83)
 - Version: 0.31.2

**Additional context**

Sahith59:
Hey folks! I was taking a look at the remaining missing methods here and just put together a PR that adds from_dlpack to the array namespace. It looks like asarray(..., copy=True) was already handled recently. Let me know if you want me to tackle any of the other missing array-api methods!

zcbenz:
Note that the array-api defines some ops that have dynamic output shape (for example `nonzero` and the set functions), which are not implementable with MLX at the moment, and we do not plan to include them.

There are also ops defined some optional parameters that would require new kernels to implement efficiently, which we also prefer to omit for now until there are real requests for them.

Having said this it seems that the open pull requests have already covered most of the missing ones, and I will close this issue once they are merged.

rgommers:
That all sounds very reasonable. Awesome to see DLPack support just landed (gh-3531), and `asarray` is present too now. With those, I suspect MLX will be supportable by SciPy, scikit-learn et al. without too much trouble. Dynamic shape-using functions like `nonzero` and `unique_*` are clearly marked in the array API standard, and optional (e.g., `jax.jit` also doesn't support them).

zcbenz:
Closing since we should have all the implementable ops included.

PR #3510, opened 2026-05-10
Add `copy` keyword to `mx.asarray`

eyupcanakman:
## Proposed changes

`mx.asarray` doesn't accept the `copy` keyword from the array-API spec. References #3484.

This adds `copy` as a keyword-only argument. MLX has no in-place operations, so a non-copying conversion isn't meaningful: `copy=False` raises `ValueError`, and `copy=True` or unspecified behave the same as before.

The `from_dlpack` half of #3484 is left out, since PR #3495 reworks dlpack consumption through the `mx.array(...)` constructor and the namespace wiring depends on that landing.

## Checklist

### DIFF
--- a/python/src/ops.cpp
+++ b/python/src/ops.cpp
@@ -1765,23 +1765,35 @@ void init_ops(nb::module_& m) {
       )pbdoc");
   m.def(
       "asarray",
-      [](const nb::object& a, std::optional<mx::Dtype> dtype) {
+      [](const nb::object& a,
+         std::optional<mx::Dtype> dtype,
+         std::optional<bool> copy) {
+        if (copy.has_value() && !*copy) {
+          throw std::invalid_argument("[asarray] copy=False is not supported.");
+        }
         return create_array(a, dtype);
       },
       nb::arg(),
       "dtype"_a = nb::none(),
+      nb::kw_only(),
+      "copy"_a = nb::none(),
       nb::sig(
-          "def asarray(a: Union[scalar, array, Sequence], dtype: "
-          "Optional[Dtype] = None) -> array"),
+          "def asarray(a: Union[scalar, array, Sequence], dtype: Optional[Dtype] = None, *, copy: Optional[bool] = None) -> array"),
       R"pbdoc(
         Convert the input to an array.
 
         Args:
             a: Input data.
             dtype (Dtype, optional): The desired data-type for the array.
+            copy (bool, optional): Must be ``True`` or unspecified. ``False``
+              is not supported, since MLX has no in-place operations and
+              cannot return a non-copying view.
 
         Returns:
             array: An array interpretation of the input.
+
+        Raises:
+            ValueError: If ``copy`` is ``False``.
       )pbdoc");
   m.def(
       "zeros_like",
--- a/python/tests/test_array.py
+++ b/python/tests/test_array.py
@@ -2152,6 +2152,17 @@ class TestArray(mlx_tests.MLXTestCase):
         arr_pass = xp.asarray(existing)
         self.assertEqual(arr_pass.tolist(), [4, 5, 6])
 
+    def test_asarray_copy(self):
+        existing = mx.array([1, 2, 3])
+
+        self.assertEqual(mx.asarray(existing, copy=True).tolist(), [1, 2, 3])
+        self.assertEqual(
+            mx.asarray(existing, dtype=mx.float32, copy=True).dtype, mx.float32
+        )
+
+        with self.assertRaises(ValueError):
+            mx.asarray(existing, copy=False)
+
     def test_asarray(self):
         # List inputs
         self.assertEqual(mx.asarray([1, 2, 3]).tolist(), [1, 2, 3])


zcbenz (changes requested):
The `copy` parameter only makes sense if the framework supports in-place operation and you want to make a deep copy whose content would stay unchanged when the original copied array got an in-place update. But since MLX has no in-place operation and the supposed use case would never happen.

So I think the best way to support `copy` parameter is to simply throw when `copy=False` is passed and return `create_array(a, dtype)` otherwise like before. Because `copy=False` means shallow-copy which makes no sense in MLX, and code using `copy=True|None` would stay semantically correct.

eyupcanakman:
Makes sense, reworked it that way: `copy=False` raises now, `copy=True` and the default go through `create_array` like before.

zcbenz (approved):
Looks good to me, thanks!

Kymi808:
Welcome contribution — closing the `mx.asarray` half of #3484 is the missing piece for cleaner array-API consumption. Two questions before this lands, both about spec alignment:

### `copy=False` raises unconditionally — but the spec only requires raising when a copy *would* be needed

Per the [Array API spec for `asarray`](https://data-apis.org/array-api/latest/API_specification/generated/array_api.asarray.html) (`copy`):

> - if `True`, the function must always copy.
> - if `False`, the function must never copy for input which supports the buffer protocol and **must raise a `ValueError` in case a copy would be made**.
> - if `None`, the function must reuse the existing memory buffer if possible (the function must never copy for input which supports the buffer protocol or other conforming arrays, but may copy in case of incompatible types).

The current diff:

```cpp
if (copy.has_value() && !*copy) {
  throw std::invalid_argument("[asarray] copy=False is not supported.");
}
```

raises whether a copy is necessary or not. The common porting idiom is `xp.asarray(arr, copy=False)` as "convert iff needed, no-op otherwise" — code like that is expected to *succeed* when `arr` is already a same-dtype array of the right namespace, and this PR makes it always fail.

I'd suggest splitting the cases:

```cpp
// Pseudocode
bool same_dtype = !dtype.has_value() || (is_mx_array(a) && a.dtype == *dtype);
bool would_copy = !is_mx_array(a) || !same_dtype;

if (copy.has_value() && !*copy && would_copy) {
    throw std::invalid_argument(
        "[asarray] copy=False but a copy would be required "
        "(input is not an mx.array, or dtype conversion was requested).");
}
// copy=False with no conversion needed → return input as-is
// copy=True or copy=None → existing create_array(a, dtype) path
```

Two possible objections worth pre-empting:
- *"MLX has no in-place ops so non-copying is meaningless."* — The PR description makes this argument, but the spec's copy semantics are about **memory sharing / buffer identity**, not in-place mutation. Even in an immutable-array system, returning the input array (no new buffer allocation) vs. constructing a new one is a real distinction that downstream consumers may rely on for memory accounting.
- *"`create_array` may already short-circuit when the input is already an mx.array of the right dtype, so `copy=None` is fine."* — If true, that just means the `copy=None` path is already spec-compliant for free; the `copy=False` path is the one that needs the conditional raise.

### `copy=None` (the default) isn't actually tested

The new test exercises `copy=True` and `copy=False`, but not `copy=None`. Since `None` is the array-API default and has the most subtle semantics ("reuse buffer if possible; copy if needed"), it's worth a test. Sketch:

```python
def test_asarray_copy_none_passthrough(self):
    # Same dtype: should not raise; should behave as a no-op shape/dtype-wise.
    existing = mx.array([1, 2, 3], dtype=mx.int32)
    out = mx.asarray(existing, copy=None)
    self.assertEqual(out.tolist(), [1, 2, 3])
    self.assertEqual(out.dtype, mx.int32)

def test_asarray_copy_false_same_dtype_is_noop(self):
    # If the spec-compliant path is adopted, this should pass.
    # If you keep the unconditional raise, this test documents that fact.
    existing = mx.array([1, 2, 3], dtype=mx.int32)
    out = mx.asarray(existing, copy=False)  # should NOT raise per spec
    self.assertEqual(out.tolist(), [1, 2, 3])

def test_asarray_copy_false_with_dtype_conversion_raises(self):
    existing = mx.array([1, 2, 3], dtype=mx.int32)
        mx.asarray(existing, copy=False, dtype=mx.float32)
```

Whichever direction the maintainers want to take, having all three states tested makes future refactors safer.

### Docs nit

The docstring currently says:

> ``copy (bool, optional)``: Must be ``True`` or unspecified. ``False`` is not supported, since MLX has no in-place operations and cannot return a non-copying view.

If the spec-compliant path is adopted, this becomes inaccurate. Even if the current "always raise" behavior stays, the explanation is technically misleading — the issue isn't "no in-place ops" but "MLX won't promise buffer identity." Worth tightening either way.

### Tiny

- `nb::kw_only()` placement looks right; just confirming you've checked that downstream callers passing `dtype` positionally still work (the existing single-positional-then-dtype signature).
- The PR description correctly notes that the `from_dlpack` half of #3484 is being intentionally deferred to land alongside #3495 — that's good scope discipline.

---
*Disclosure: this review was prepared with AI assistance; technical claims verified against the codebase and the array-API spec locally.*

Issue #3186, opened 2026-03-01
Kernel panic (IOGPUMemory.cpp:550) on M4 Max with large context prefill (~173K tokens)


kotono-amaha:
**Apple Feedback ID**: FB22091885

This issue has been filed with Apple and is cross-referenced here for the MLX community. A fix may come from either side.

Kernel panic in IOGPUMemory.cpp:550 triggered by large Metal GPU memory allocation during MLX inference on M4 Max.

**PANIC STRING:**
"completeMemory() prepare count underflow" @IOGPUMemory.cpp:550

**SYSTEM:**
- Hardware: Apple M4 Max (36GB unified memory)
- macOS: 26.3 (25D125)
- Kernel: Darwin 25.3.0 xnu-12377.81.4~5/RELEASE_ARM64_T6041

**REPRODUCIBLE:** Yes — confirmed twice with identical call stacks.

**REPRODUCTION STEPS:**
1. Install MLX and mlx-lm via pip on Python 3.14 ARM64
2. Load a large quantized LLM (Qwen3.5-27B Q5_K_M) via mlx_vlm.load()
3. Construct a prompt consisting of 147 concatenated model outputs totalling approximately 173,000 tokens
4. Call mlx_vlm.generate() with this prompt — prefill phase begins processing the full context
5. Kernel panics during prefill, consistently at IOGPUMemory.cpp:550

**ROOT COMPONENT:**
com.apple.iokit.IOGPUFamily (129.3.2)

**NOTES:**
- Panic does not occur with smaller prompts (under ~10,000 tokens)
- Memory capacity is not the issue — system has 36GB and model occupies  ~26GB, leaving sufficient headroom
- Issue appears to be a GPU memory accounting state corruption triggered 
  by a single contiguous Metal allocation for a very large attention  computation, not an out-of-memory condition
- Two panic logs attached with identical backtraces confirming  deterministic reproducibility

**Suggested mitigation for MLX:**
Add a prefill token count guard in mlx_lm before the Metal allocator is called. If the prompt exceeds a safe threshold (empirically somewhere 
below 173K tokens on M4 Max with 36GB), either raise a clear exception with guidance to chunk the prompt, or automatically split the prefill 
into safe-sized segments. This would prevent the IOGPUFamily kernel panic without requiring a macOS fix.

AlchemyHAN:
[...]

PR #1718, opened 2024-12-18
Track resource limit and throw if exceeded

awni:
Track the metal resource limit `iogpu.rsrc_limit` and throw if it is exceeded. `mx.metal.device_info` also returns the resource limit.

Closes #1712

PR #3512, opened 2026-05-10
metal: scale rsrc_limit fallback with hw.memsize

electron-rare:
The current fallback for `iogpu.rsrc_limit` is hardcoded to `499000` in `backend/metal/device_info.cpp`. That value is appropriate for small-RAM Apple Silicon devices, but it truncates Metal command buffers during sustained bf16 fine-tuning of large models on Pro/Max/Ultra-class machines.

This PR replaces the hardcoded fallback with a 4-tier table indexed on `hw.memsize`, which is already queried two lines above:

| `hw.memsize` | rsrc_limit fallback | Multiplier | Target machines |
|---|---|---|---|
| `< 24 GB` | 499 000 | **1× (stock, unchanged)** | MBA M1/M2/M3, Mini base, MBP 16 GB |
| `≥ 24 GB` | 748 500 | 1.5× | M1/M2/M3 Pro 24–36 GB |
| `≥ 64 GB` | 998 000 | 2× | M-series Max/Studio 64–128 GB |
| `≥ 384 GB` | 1 497 000 | 3× | M3 Ultra 512 GB |

The sysctl path is unchanged: when `iogpu.rsrc_limit` is configured explicitly (e.g. `sudo sysctl -w iogpu.rsrc_limit=...`), that value still wins. The tiered fallback only kicks in when the kernel reports `0`, which is the documented "use default" sentinel.

### Why

The 499 K handle ceiling becomes the binding constraint when training or running inference on dense bf16 models that approach unified-memory saturation:

- On a 512 GB M3 Ultra, sustained bf16 training of >100 B-parameter models requires >1 M live `MTLResource` objects at peak.
- On a 32 GB M-series Pro, sustained LoRA fine-tuning of 35 B-A3B-class models hits the same ceiling around ~600 K handles.

Conversely, on 8 / 16 GB machines the handle ceiling is **not** the binding constraint — physical memory saturates first. A typical Q4 7 B model + KV cache reaches ~6 GB resident on a 16 GB MBA before approaching even 30 K active resources, ~1 / 16 of the stock ceiling. Raising the fallback there costs ~10–20 MB of kernel allocator state per tier with zero observable benefit.

Tiering by `hw.memsize` keeps the behavior local to the machines that actually need it, and leaves the defaults untouched for the dominant 8 / 16 GB MacBook installed base.

### Why these thresholds?

The boundaries map 1:1 to shipping Apple Silicon SKUs rather than arbitrary GB cutoffs:

- **24 GB**: catches the M3 Pro 36 GB SKU cleanly while excluding the M3 Pro 18 GB SKU, which behaves like a 16 GB device for practical purposes.
- **64 GB**: separates Pro-class from Max/Studio-class.
- **384 GB**: triggers only on M3 Ultra 512 GB, the only SKU that has demonstrated a need for the 3× uplift in practice.

A continuous formula like `clamp(1 + memsize / 256GB, 1, 3)` was considered but rejected as harder to reason about — discrete tiers are explicit and a user can read the table to know what their machine gets.

### Test plan

Validated on real hardware via a standalone probe replicating the tier logic:

| Machine | `hw.memsize` | Tier | rsrc_limit | ✓ |
|---|---|---|---|---|
| Mac Studio M3 Ultra | 512 GB | 3× | 1 497 000 | ✅ |
| Apple Silicon M5 (16 GB) | 16 GB | stock | 499 000 | ✅ |

Additional coverage by symmetric `if/else` reasoning:

- M-series Pro 32 / 36 GB → 1.5× tier → 748 500
- M-series Max 64 / 96 / 128 GB → 2× tier → 998 000
- Any machine with explicit `sudo sysctl -w iogpu.rsrc_limit=N` → returns `N` (sysctl path wins).

### Coordination with `mx.set_cache_limit()`

The two limits are orthogonal:

- `mx.set_cache_limit(bytes)` controls the GPU memory **allocator cache** (frontend, runtime, byte budget).
- `rsrc_limit` controls the Metal **resource handle count** (backend, build-time, object count).

Both are sometimes needed in different combinations. Happy to add a paragraph in `docs/usage/memory_management.rst` if reviewers think it would help.

---

### Precedent in this codebase

PR #1064 ("Change initial memory limits and add memory size to device info") introduced `memsize` exposure into `device_info.cpp` precisely so downstream limits can scale with hardware. This PR is the natural follow-through for the `rsrc_limit` fallback that was added by PR #1718 ("Track resource limit and throw if exceeded"). The literal `rsrc_limit = 499000;` has been unchanged since that PR landed; this is the first proposed scaling.

### Related user-reported failures

The exact `Resource limit (499000) exceeded` ceiling has been hit and reported across the ecosystem with no upstream fix in flight:

- ml-explore/mlx#3186 — kernel panic on M4 Max with large context (adjacent symptom, IOGPUMemory underflow rather than the rsrc_limit throw, but driven by same large-RAM Apple Silicon usage pattern)
- ml-explore/mlx-lm#831 — "Resource limit (499000) exceeded" with `mlx_lm.server` distributed
- ml-explore/mlx-lm#1015 — `generate()` crashes on Metal OOM instead of recovering gracefully
- ml-explore/mlx-examples#1262 — training crash from rising allocations
- lmstudio-ai/mlx-engine#264 — same `499000` error on M4 Pro 48 GB

Today, every workaround in the wild is either `sudo sysctl -w iogpu.rsrc_limit=…` or `mx.set_wired_limit()` (a different lever — bytes budget, not handle count). This PR offers the in-tree, automatic alternative.

### Validation status

Hardware-validated on real machines:
- Mac Studio M3 Ultra (`hw.memsize = 549755813888`, 512 GB) → 3× tier → `rsrc_limit = 1497000` ✓
- Apple Silicon M5 (`hw.memsize = 17179869184`, 16 GB) → stock tier → `rsrc_limit = 499000` ✓ (no behavior change on small-RAM machines)

### Benchmarks (M3 Ultra 512 GB, macOS 26.4 / Darwin 25.4.0, mlx 0.31.2 + mlx-lm 0.31.3)

Apertus 70B BF16 inference on a single Mac Studio M3 Ultra, with `mx.set_memory_limit(440 * 1024**3)` to leave headroom for the OS. The machine is in its **pre-PR state** (`mx.device_info()["resource_limit"] == 499000`).

| Test | Result |
|---|---|
| Model load time | **7.9 s** |
| Peak Metal memory after load | **131.5 GiB** |
| Short-context gen (41 tok prompt → 512 tok out) | **4.67 tok/s** |
| Long-context gen (2 001 tok prompt → 129 tok out) | **3.25 tok/s** |
| Peak Metal memory under long-context | **132.9 GiB** |
| `resource_limit` reported by `mx.device_info()` | **499 000** (no PR fix yet) |

Reproducer (the full script is checked in at `KIKI-Mac_tunner/lib/mlx_bench_3512.py`):

```python
import mlx.core as mx, sys, time
sys.path.insert(0, "/Users/clems/KIKI-Mac_tunner/lib")
mx.set_memory_limit(440 * 1024**3)
from mlx_lm_fork import load, generate

m, tok = load("…/Apertus-70B-Instruct-2509")
print(mx.get_peak_memory() / 1024**3, "GiB after load")
out = generate(m, tok, prompt=PROMPT, max_tokens=512)
```

#### Handle budget — why 499 000 is the binding ceiling

Apertus 70B is a 64-layer dense transformer (BF16, ~141 GB on disk). A single forward pass keeps live:

- Per-layer weight tensors: q/k/v/o projections + 2× MLP + 2× RMSNorm = **~8 MTLResource handles per layer** × 64 = ~512.
- KV cache, one pair per layer per token: 2 × 64 = **128 handles per cached token**.
- Activation residuals, logits, embedding lookup buffers: ~constant ~5 000 baseline.

At the 2 K-context point measured here we sit at roughly `5 000 + 64*8 + 128*2048 ≈ 267 K` live handles — already more than half of the 499 K ceiling, and the test completed cleanly. Extrapolating linearly: a 4 K-context inference run lands at **~530 K handles**, which exceeds the stock ceiling and is exactly the regime where users report `Resource limit (499000) exceeded` (see the cluster of upstream issues linked above). The 3× tier introduced by this PR (`1 497 000`) gives a ~5.6× headroom over the 2 K-context working point and ~2.8× over the 4 K crossover, matching the empirical budget for 70 B–class bf16 workloads at conversational context lengths.

| Context length | Estimated peak handles | Stock 499 K | This PR (3× = 1 497 K) |
|---|---|---|---|
| 2 K (measured) | ~267 K | ✓ fits | ✓ fits |
| 4 K (extrapolated) | ~530 K | ✗ exceeds | ✓ fits |
| 8 K (extrapolated) | ~1.05 M | ✗ exceeds | ✓ fits |
| 16 K (extrapolated) | ~2.1 M | ✗ exceeds | ✗ exceeds — needs `sudo sysctl -w` |

The PR's 3× ceiling is therefore tight-but-sufficient for the dominant use case (bf16 70 B at 4–8 K context) and intentionally does *not* cover indefinitely-long contexts, for which the existing manual sysctl override remains the correct escape hatch.

Full benchmarks above were collected on `2026-05-11`; the M3 Ultra was reachable over Tailscale and was simultaneously hosting an unrelated Mistral-Medium-3.5-128B Q8 server (~116 GiB resident), so the load-time and peak-memory numbers here are conservative — a freshly booted machine would see slightly better peak headroom.

Happy to also switch to a continuous formula derived from `recommendedMaxWorkingSetSize()` if reviewers prefer that direction over the discrete tier table.

### DIFF
--- a/mlx/backend/metal/device_info.cpp
+++ b/mlx/backend/metal/device_info.cpp
@@ -38,7 +38,20 @@ device_info(int device_index) {
     size_t rsrc_limit = 0;
     sysctlbyname("iogpu.rsrc_limit", &rsrc_limit, &length, NULL, 0);
     if (rsrc_limit == 0) {
-      rsrc_limit = 499000;
+      // Default macOS rsrc_limit (499K) is too tight for large unified
+      // memory training. Scale with hw.memsize so small machines stay
+      // on stock and only large-RAM devices get the uplift.
+      constexpr size_t kBaseRsrcLimit = 499000;
+      constexpr size_t kGB = 1024ULL * 1024ULL * 1024ULL;
+      if (memsize >= 384 * kGB) {
+        rsrc_limit = kBaseRsrcLimit * 3;       // M3 Ultra 512 GB
+      } else if (memsize >= 64 * kGB) {
+        rsrc_limit = kBaseRsrcLimit * 2;       // 64-128 GB Max/Studio
+      } else if (memsize >= 24 * kGB) {
+        rsrc_limit = (kBaseRsrcLimit * 3) / 2; // 24-36 GB Pro (1.5x)
+      } else {
+        rsrc_limit = kBaseRsrcLimit;           // < 24 GB: stock
+      }
     }
 
     return {


zcbenz (commented):
The referenced code is used to __get__ the limit __not set__ it.

To change it you can try calling `sudo sysctl iogpu.rsrc_limit` but I'm not sure if it would work.

PR #3513, opened 2026-05-10
mlx launch clean

nastya236:
- Drop verification from `mlx.launch`. Alternatively we can swap it to `--verify-script` and verify only conditionally ?
- Drop `required=True` for `--over flag` -- this could be not ideal, we might want to require explicit interconnect specification. Then we should `required=True` and drop default ?

### DIFF
--- a/python/mlx/_distributed_utils/config.py
+++ b/python/mlx/_distributed_utils/config.py
@@ -586,7 +586,6 @@ def main():
         choices=["thunderbolt", "ethernet"],
         default="thunderbolt",
         help="What type of connectivity to configure",
-        required=True,
     )
     parser.add_argument(
         "--output-hostfile", help="If provided, save the hostfile to this path"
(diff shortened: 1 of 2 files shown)


angeloskath (approved):
Awesome. You can now do

`mlx.distributed_config --hosts h1,h2,h3,h4 --auto-setup --output hosts.json`

and it will auto select everything.

Do you think we should remove the automatic insertion of `python` if a script is provided? Basically if `foo.py` exists locally and remotely then it will try to add `args.python` in front of it...

I basically think lines 544-548 may cause more harm than good.

angeloskath:
@nastya236 I 'll merge this, let me know if you think we should remove 544-548.

nastya236:
Agree these lines! I had a doubt (removed it first, then committed back :) Lets remove it, I think we should just always give executable full path for simplicity

PR #3318, opened 2026-03-25
fix: defer error propagation from Metal completion handlers

Thump604:
## Summary

`check_error()` throws `std::runtime_error` inside Metal completion handlers (GCD callbacks). C++ exceptions cannot be caught in GCD dispatch contexts, so the throw hits `std::terminate` -> `abort()` -> SIGABRT.

This is a production crash. On M2 Ultra 128GB running Qwen3.5-122B (5-bit, 32K context), it occurs after ~3.5 hours of sustained inference. The crash stack:

```
libc++abi: __cxa_throw -> failed_throw -> std::terminate -> abort
libmlx.dylib: check_error(MTL::CommandBuffer*)
libmlx.dylib: [completion handler block]
Metal: -[_MTLCommandBuffer didCompleteWithStartTime:endTime:error:]
IOGPU: -[IOGPUMetalCommandBuffer fillCommandBufferArgs:]_block_invoke
```

## Fix

Replace `check_error()` with `check_error_deferred()` in all 3 completion handler sites in `eval.cpp`. Errors are stored in a thread-safe slot and re-thrown at the next `eval()` or `synchronize()` call, where C++ exception handling works correctly.

The error is never lost -- it surfaces as the same `std::runtime_error` at the next synchronization point.

## Changes

- `eval.cpp`: Add `check_error_deferred()` with mutex-protected error store
- `eval()`: Check for deferred errors at entry
- `synchronize()`: Check for deferred errors at entry
- All 3 `addCompletedHandler` sites: `check_error` -> `check_error_deferred`

## Testing

- SDPA verification passes
- Production 122B inference verified after rebuild
- Previously crashed after ~3.5h sustained load; fix deployed, monitoring

## Related

- #3317 (issue filed with full crash analysis)
- #3216 (separate Metal thread safety issue, fixed by #3281)

Fixes #3317.

Issue #2670, opened 2025-09-29
when it crashed in the background, can not catch the exception


taocheng-njfu:
<img width="578" height="179" alt="Image" src="https://github.com/user-attachments/assets/e8b05e83-6fb4-4837-b3af-4abc0a6911eb" />
why it can not catch the exception below, how can I catch the exception when run in the background

<img width="570" height="162" alt="Image" src="https://github.com/user-attachments/assets/f2cabc12-3c03-4973-b8d5-5fa2d9dc36e2" />

davidkoski:
This is thrown on the metal worker thread -- it is possible that this is not set up to deliver this back to the calling thread (the one calling eval).

davidkoski:
I think this is the same as ml-explore/mlx-swift#274 -- the suggestion to try and catch the error didn't work because the error wasn't surfaced there.

taocheng-njfu:
> This is thrown on the metal worker thread -- it is possible that this is not set up to deliver this back to the calling thread (the one calling eval).

I think so too. Is there any way to catch this abnormality? I've been blocked for 3 or 4 days, and I don't have any ideas.

taocheng-njfu:
> I think this is the same as [#274](https://github.com/ml-explore/mlx-swift/issues/274) -- the suggestion to try and catch the error didn't work because the error wasn't surfaced there.

<img width="945" height="377" alt="Image" src="https://github.com/user-attachments/assets/d4e0cc86-a7eb-45e3-88df-2ee4eea5385b" />

<img width="922" height="273" alt="Image" src="https://github.com/user-attachments/assets/e250ddcc-a42d-40b2-ad24-478ffc6d8489" />

it can not catch the exception

davidkoski:
It would have to be a change on the mx::core side (mlx project).

davidkoski:
Let me see about transferring this issue.

davidkoski:
Specifically the request is:

- if there is an uncaught exception on the worker thread, surface that or a proxy back in the eval

taocheng-njfu:
Any updates, boss?

davidkoski:
@awni I think this requires a change on the mlx (core) side -- do you agree?  Can you move it to the mlx repo?  Thanks!

awni:
Yea I can move it... but it's not likely we are going to add this feature in the near future. The reason is that we don't have guarantees on the state being in a reasonable condition if there is an exception during eval.

[...]

PR #3519, opened 2026-05-11
fix(metal): defer command-buffer error reporting + poison-on-failure

### LINKED ISSUE #3317: check_error() throws inside Metal completion handler, hits std::terminate
## Summary

`check_error(MTL::CommandBuffer*)` in `mlx/backend/metal/eval.cpp` can throw `std::runtime_error` inside a Metal completion handler callback dispatched by GCD. There is no try/catch in the handler, so the exception hits `std::terminate` -> `abort()` -> SIGABRT.

## Environment

- M2 Ultra 128GB, macOS 26.3.1
- MLX 0.31.2-dev (branch feat/chunked-sdpa-logsumexp, includes PR #3281)
- Qwen3.5-122B-A10B, 5-bit, 32K context inference

## Crash Stack (from .ips report)

```
libc++abi: __cxa_throw -> failed_throw -> std::terminate -> abort
libmlx.dylib: check_error(MTL::CommandBuffer*)
libmlx.dylib: [completion handler block]
Metal: -[_MTLCommandBuffer didCompleteWithStartTime:endTime:error:]
IOGPU: -[IOGPUMetalCommandBuffer fillCommandBufferArgs:]_block_invoke
[GCD dispatch chain]
```

## Root Cause

In `eval.cpp` the completion handler is:

```cpp
    [s, buffers = std::move(buffers)](MTL::CommandBuffer* cbuf) {
        check_error(cbuf);  // throws std::runtime_error if GPU error
    });
```

When the GPU reports a command buffer execution error (OOM, timeout, or corrupted state), `check_error` throws. But this callback runs on a GCD dispatch thread, not in a context where C++ exceptions can be caught. The throw goes straight to `std::terminate`.

## Reproduction

This occurred during sustained 32K-context inference (250 sequential requests, lm_eval RULER benchmark). The crash happens after ~3.5 hours of continuous operation. It's not easily reproducible on demand but occurs reliably under sustained load.

## Suggested Fix

Catch the exception inside the completion handler and propagate the error through a different mechanism (e.g., an error flag checked at the next `eval()` synchronization point):

```cpp
    [s, buffers = std::move(buffers)](MTL::CommandBuffer* cbuf) {
        try {
            check_error(cbuf);
        } catch (const std::exception& e) {
            // Store error for propagation at next sync point
            scheduler::set_task_error(s, e.what());
        }
    });
```

## Related

- #3216 -- Metal driver SIGABRT (different bug: thread safety race). Fixed by PR #3281 (merged).
- This is a separate bug: the completion handler throw path, not thread safety.

andreinknv:
## Summary

C++ exceptions thrown from inside Metal command-buffer completion handlers (the three `addCompletedHandler` callbacks in `mlx/backend/metal/eval.cpp`) hit `std::terminate` → `abort()` because those callbacks run on Metal-managed dispatch threads where C++ exceptions cannot be caught. Production users have hit this in **#3317** (M2 Ultra, Qwen3.5-122B, ~3.5h sustained inference) and again recently on M4 Max running `mlx-lm 0.31.3` 4-server pool with Granite-1b — `kIOGPUCommandBufferCallbackErrorOutOfMemory` from prompt-cache buildup → SIGABRT, killing one pool member mid-run.

Crash signature (matches both reports):

```
  [METAL] Command buffer execution failed: ...
__cxa_throw → _objc_terminate → abort
mlx::core::gpu::check_error(MTL::CommandBuffer*)
[completion handler block]
-[_MTLCommandBuffer didCompleteWithStartTime:endTime:error:]
```

## Background — why prior PR #3318 was rejected

[#3318](https://github.com/ml-explore/mlx/pull/3318) proposed deferring the error to the next user-thread `eval()`. It got production validation (5 days unattended on Qwen3.5-122B by the author, overnight stress on M3 Ultra by hnshah) but was rejected by maintainers:

- **zcbenz on #3317**: *"We can not simply transfer the error: MLX code is not exception-safe so when the exception is thrown the process would be in a stale state that will likely fail later in a much weirder way."*
- **awni on #2670**: *"It's not likely we are going to add this feature in the near future. The reason is that we don't have guarantees on the state being in a reasonable condition if there is an exception during eval."*

The author of #3318 closed it themselves later because they "narrowed active upstream work to vllm-mlx and can't maintain it." The bug remains open on `main`.

## How this PR is different

This PR addresses the state-safety objection head-on by combining the deferred-throw with explicit **stream poisoning**:

1. The async completion handler **never throws**. It captures the error message into a per-`StreamThread` slot and sets `poisoned = true`.
2. The next user-thread `eval()` / `finalize()` / `synchronize()` entry on that stream calls `throw_if_captured()`:
   - **First call after error**: re-throws the original `std::runtime_error` on a thread the language runtime can handle.
   - **Subsequent calls**: refuses with a clear message — *"[METAL] Stream is in error state from a prior failure. Call mx.clear_streams() (or destroy this Stream) before queuing more work."*
3. `mx.clear_streams()` now also calls `scheduler::reset_all_errors()` so the user has a documented path back to a working state.

**This guarantees no further operations execute against a poisoned stream.** Once an async error is captured, NOTHING runs on that stream until the caller explicitly resets — there's no opportunity to operate on partially-initialized encoder state, which was the maintainer's central concern.

Per-stream rather than global: concurrent streams (e.g. mlx-lm's 4-server pool, batched inference) won't cross-pollute errors. A failure on stream 2 only poisons stream 2, not 0/1/3.

## Changes

- **`mlx/scheduler.h`**: add `captured_error` + `poisoned` slots and `capture_error` / `take_error` / `reset_error` methods to `StreamThread`. Expose `Scheduler::capture_error` / `take_error` / `reset_error` / `reset_all_errors` (with `threads_mtx_` shared lock). Add free-function aliases.
- **`mlx/backend/metal/eval.cpp`**: replace `check_error` (still defined for the synchronous `CommandEncoder::synchronize` path) with `capture_async_error` for the three `addCompletedHandler` callsites. Add `throw_if_captured` at `eval()` / `finalize()` / `synchronize()` entry. `clear_streams` now resets all per-stream error slots.

Total diff: **+148 / −6 LOC across 2 files.**

## Test results

- ✅ **709 mlx tests pass, 26 skipped, zero new failures** (`python -m unittest discover` in `python/tests/`)
- ✅ **Stress test**: 8 threads × 1000 iter × (512×512) matmul on per-thread streams → 0.7 s, no errors
- ✅ **Smoke**: addition / matmul / pre-flight allocation check / post-error recovery all behave correctly
- ✅ **Functional**: `mx.clear_streams()` continues to work, ops resume after reset

The deferred error mechanism cannot be unit-tested without injecting a real Metal command-buffer failure (the pre-flight `metal::malloc` size check intercepts most synthetic OOM repros). The crash dumps in #3317 and the local Granite-1b reproduction validate the underlying fix correctness.

## Closes / references

- Closes #3317
- References #2670, #3318, #3216

cc @zcbenz @awni @davidkoski — would appreciate your review on whether stream poisoning addresses the state-safety concerns from your earlier comments.

### DIFF
--- a/mlx/backend/metal/eval.cpp
+++ b/mlx/backend/metal/eval.cpp
@@ -18,18 +18,37 @@ void new_stream(Stream s) {
   encoders.try_emplace(s.index, d, s.index, d.residency_set());
 }
 
-inline void check_error(MTL::CommandBuffer* cbuf) {
+// Capture (don't throw) a Metal command-buffer error from an async
+// completion handler. The handler runs on a Metal-managed dispatch
+// thread; throwing through Objective-C frames hits _objc_terminate ->
+// abort() and crashes the whole process. Storing the message and
+// re-throwing on the next user-thread eval()/finalize() call keeps
+// the error catchable by Python / C++ callers.
+inline void capture_async_error(const Stream& s, MTL::CommandBuffer* cbuf) {
   if (cbuf->status() == MTL::CommandBufferStatusError) {
     std::ostringstream msg;
     msg << "[METAL] Command buffer execution failed: "
         << cbuf->error()->localizedDescription()->utf8String();
-    throw std::runtime_error(msg.str());
+    scheduler::capture_error(s, msg.str());
+  }
+}
+
+// Re-throw any previously-captured async Metal error on the calling
+// thread. Called at the entry of every user-facing eval()/finalize()
+// so failures from the prior step surface before more work is queued.
+inline void throw_if_captured(const Stream& s) {
+  auto msg = scheduler::take_error(s);
+  if (!msg.empty()) {
+    throw std::runtime_error(msg);
   }
 }
 
 void eval(array& arr) {
   auto pool = metal::new_scoped_memory_pool();
   auto s = arr.primitive().stream();
+  // Surface any async Metal error captured by a prior step's completion
+  // handler before queuing more work on this stream.
+  throw_if_captured(s);
   auto& encoder = metal::get_command_encoder(s);
   auto* command_buffer = encoder.get_command_buffer();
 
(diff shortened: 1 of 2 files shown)


zcbenz (commented):
This PR is much better than previous tryings, but the error handling is over complicated, and the timing of error throwing is bad. I'm closing this in favor of https://github.com/ml-explore/mlx/pull/3523.

PR #3523, opened 2026-05-11
Catch error in CommandBuffer and poison the events

### LINKED ISSUE #2670: when it crashed in the background, can not catch the exception

zcbenz:
Close #2670.

When we got error in CommandBuffer's completion handler, save it, and then throw the error later in a few safe points:
* after synchronization;
* before and after command encoding.

In this way the program would be able to catch the error, and could continue running safely after recovering from the error.

The downside is that the timing of error throwing would be delayed much later until the program run into the check points, which is actually quite similar to how CUDA handles errors.

### DIFF
--- a/mlx/backend/metal/device.cpp
+++ b/mlx/backend/metal/device.cpp
@@ -11,6 +11,7 @@
 
 #include "mlx/backend/common/utils.h"
 #include "mlx/backend/metal/device.h"
+#include "mlx/backend/metal/event.h"
 #include "mlx/backend/metal/metal.h"
 #include "mlx/backend/metal/utils.h"
 #include "mlx/utils.h"
@@ -442,12 +443,64 @@ void CommandEncoder::end_encoding() {
   all_inputs_.clear();
 }
 
+void CommandEncoder::signal_event(
+    std::shared_ptr<EventImpl> event,
+    uint64_t value) {
+  end_encoding();
+  buffer_->encodeSignalEvent(event->mtl_event(), value);
+  signal_events_.push_back({std::move(event), value});
+}
+
+void CommandEncoder::wait_event(
+    std::shared_ptr<EventImpl> event,
+    uint64_t value) {
+  end_encoding();
+  buffer_->encodeWait(event->mtl_event(), value);
+  wait_events_.push_back(std::move(event));
+}
+
 bool CommandEncoder::needs_commit() const {
   auto [max_ops, max_mb] = device_.get_max_ops_mb_per_buffer();
   return (buffer_ops_ > max_ops) || ((buffer_sizes_ >> 20) > max_mb);
 }
 
-void CommandEncoder::commit() {
+void CommandEncoder::commit(std::function<void()> completion) {
+  buffer_->addCompletedHandler(
+      [&error_ = error_,
+       wait_events = std::move(wait_events_),
+       signal_events = std::move(signal_events_),
+       completion = std::move(completion)](MTL::CommandBuffer* cbuf) {
+        if (completion) {
+          completion();
+        }
+        // If any of the waited event has error in it, poison the encoder.
+        for (auto& event : wait_events) {
+          if (event->error()) {
+            error_ = event->error();
+            break;
+          }
+        }
+        // Set error only when no error happended before, to preserve the
+        // earliest error.
+        if (!error_ && cbuf->status() == MTL::CommandBufferStatusError) {
+          error_ = std::make_shared<std::string>(fmt::format(
+              "[METAL] Command buffer execution failed: {}.",
+              cbuf->error()->localizedDescription()->utf8String()));
+        }
+        // Poison all the signaled events when error happened.
+        if (error_) {
+          for (auto& [event, value] : signal_events) {
+            event->set_error(error_);
+          }
+        }
+        // Metal won't signal the events for us on error, manually signal them
+        // to avoid infinite waiting.
+        if (cbuf->status() == MTL::CommandBufferStatusError) {
+          for (auto& [event, value] : signal_events) {
+            event->signal(value);
+          }
+        }
+      });
   buffer_->commit();
   buffer_ = NS::RetainPtr(queue_->commandBufferWithUnretainedReferences());
   buffer_ops_ = 0;
@@ -456,17 +509,14 @@ void CommandEncoder::commit() {
 
 void CommandEncoder::synchronize() {
   auto pool = new_scoped_memory_pool();
-  auto cb = NS::RetainPtr(get_command_buffer());
+  auto cbuf = buffer_; // retained
   end_encoding();
   commit();
-  cb->waitUntilCompleted();
-  if (!exiting_) {
-    if (cb->status() == MTL::CommandBufferStatusError) {
-      throw std::runtime_error(
-          fmt::format(
-              "[METAL] Command buffer execution failed: {}.",
-              cb->error()->localizedDescription()->utf8String()));
-    }
+  cbuf->waitUntilCompleted();
+
+  if (error_ && !exiting_) {
+    auto error = std::move(error_);
+    throw std::runtime_error(*error);
   }
 }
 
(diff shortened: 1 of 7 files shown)


mlx/backend/metal/device.cpp
```diff
@@ -456,22 +471,32 @@ void CommandEncoder::commit() {
...
+  // Do not check error during encoding, otherwise it would leave the program in
+  // corrupted state.
+  if (encoder_) {
+    return;
+  }
+  // When exiting with pending GPU commands, errors will happen, ignore them.
+  if (exiting_) {
+    return;
+  }
+  auto error = std::atomic_exchange(&error_, {});
```
angeloskath: A bit extreme but since I went through the C++ reference I might as well add it here.

Apparently `std::atomic_exchange` and `std::atomic_store` are deprecated in C++20 and removed in C++26. From the reference

>These functions are typically implemented using mutexes, stored in a global hash table where the pointer value is used as the key.
The [Concurrency TS](https://en.cppreference.com/cpp/experimental/concurrency) offers atomic smart pointer classes atomic_shared_ptr and atomic_weak_ptr as a replacement for the use of these functions.

Not a particularly big deal but also fairly easy to wrap `std::shared_ptr<std::string>` with `std::atomic<>`.

mlx/backend/metal/event.cpp
```diff
@@ -21,9 +21,13 @@ Event::Event(Stream stream) : stream_(stream) {
...
-  if (!static_cast<MTL::SharedEvent*>(event_.get())
-           ->waitUntilSignaledValue(value(), -1)) {
-    throw std::runtime_error("[Event::wait] Timed out");
+  auto* event = static_cast<MTL::SharedEvent*>(event_.get());
+  // When error happened in command buffer, the event would wait indefinitely
+  // if we don't set a timeout.
+  while (!event->waitUntilSignaledValue(value(), 5 * 1000)) {
+    for (auto& [_, encoder] : metal::get_command_encoders()) {
+      encoder.check_error();
+    }
```
angeloskath: This has a subtle weird use-case.

If an array is created and async_evaled in thread a and then waited upon in thread b then `metal::get_command_encoders()` will not return the command encoder that is actually evaluating the array whose event this is.

This doesn't break the usability, just that the `check_error` is actually a noop in that case and we can't recover from a failure 🤷‍♂️.

mlx/backend/metal/event.cpp
angeloskath: Similar to the case above, just waiting on a GPU event from the CPU means no error recovery since `metal::get_command_encoders()` will be empty.

mlx/backend/metal/fence.cpp
```diff
@@ -56,9 +56,12 @@ void Fence::wait(Stream stream, const array& x) {
...
       auto& f = *static_cast<FenceImpl*>(fence_.get());
       if (!f.use_fast) {
-        if (!static_cast<MTL::SharedEvent*>(f.fence)->waitUntilSignaledValue(
-                count, -1)) {
-          throw std::runtime_error("[Fence::wait] Timed out");
+        // Same with Event::wait
+        auto* event = static_cast<MTL::SharedEvent*>(f.fence);
+        while (!event->waitUntilSignaledValue(count, 5 * 1000)) {
+          for (auto& [_, encoder] : metal::get_command_encoders()) {
+            encoder.check_error();
```
angeloskath: Same as above. Basically these are not bugs but they are not useful since they will never result in anything.

angeloskath (commented):
I left a couple of comments.

I think the main issue is that check_error in the Event/Fence wait is not doing what it's meant to since the calling thread is practically never the thread that scheduled the work.

mlx/backend/metal/device.cpp
zcbenz: Xcode does not compile `std::atomic<std::shared_ptr<std::string>>`, it ships a weird version of libc++ that does not fully support C++20. I added a comment in code so someone will remember to replace it in future.

zcbenz:
I changed the implementation to propagate the error from `CommandBuffer` to `Event`:
- When error happened in `CommandBuffer`, set error in all the signaled `Event`s.
- When an `Event` is waited and then an error was set for it, pass the error to the `CommandEncoder` and all its signaled `Event`s.

So error in one stream would be propagated to all other streams via events. (Except for CPU stream which currently just throws and crashes.)

There is no good way to test it though, I'm mostly just using `mlx.launch --verbose -n 32 python python/tests/ring_test_distributed.py` which reliably times out.

angeloskath:
This honestly looks like an awesome cleanup!

I am curious about any overheads... probably overthinking it. I want to run some generations and small model training to verify no regression and then we can merge.

Love the fence cleanup in particular.

zcbenz:
I did a simple benchmarking by running `mlx_lm.benchmark --model meta-llama/Llama-3.1-8B-Instruct -n 10 -p 64 -g 512` twice for each branch which seemed to be able to reduce variabilities a lot.

For main:

```console
$ mlx_lm.benchmark --model meta-llama/Llama-3.1-8B-Instruct -n 10 -p 64 -g 512
Timing with prompt_tokens=64, generation_tokens=512, batch_size=1.
Trial 1:  prompt_tps=371.523, generation_tps=24.235, peak_memory=16.175, total_time=21.350
Trial 2:  prompt_tps=369.580, generation_tps=24.232, peak_memory=16.175, total_time=21.354
Trial 3:  prompt_tps=354.408, generation_tps=24.237, peak_memory=16.175, total_time=21.357
Trial 4:  prompt_tps=353.003, generation_tps=24.240, peak_memory=16.176, total_time=21.355
Trial 5:  prompt_tps=368.560, generation_tps=24.238, peak_memory=16.176, total_time=21.350
Trial 6:  prompt_tps=370.512, generation_tps=24.241, peak_memory=16.176, total_time=21.345
Trial 7:  prompt_tps=354.048, generation_tps=24.238, peak_memory=16.176, total_time=21.357
Trial 8:  prompt_tps=351.635, generation_tps=24.239, peak_memory=16.177, total_time=21.357
Trial 9:  prompt_tps=371.316, generation_tps=24.239, peak_memory=16.177, total_time=21.347
Trial 10:  prompt_tps=368.632, generation_tps=24.239, peak_memory=16.177, total_time=21.348
Averages: prompt_tps=363.322, generation_tps=24.238, peak_memory=16.176
```

for this branch:

```console
$ mlx_lm.benchmark --model meta-llama/Llama-3.1-8B-Instruct -n 10 -p 64 -g 512
Timing with prompt_tokens=64, generation_tokens=512, batch_size=1.
Trial 1:  prompt_tps=354.694, generation_tps=24.242, peak_memory=16.175, total_time=21.353
Trial 2:  prompt_tps=371.278, generation_tps=24.237, peak_memory=16.175, total_time=21.348
Trial 3:  prompt_tps=370.507, generation_tps=24.240, peak_memory=16.175, total_time=21.347
Trial 4:  prompt_tps=352.574, generation_tps=24.246, peak_memory=16.176, total_time=21.350
Trial 5:  prompt_tps=354.255, generation_tps=24.236, peak_memory=16.176, total_time=21.358
Trial 6:  prompt_tps=355.678, generation_tps=24.242, peak_memory=16.176, total_time=21.353
Trial 7:  prompt_tps=370.067, generation_tps=24.238, peak_memory=16.176, total_time=21.349
Trial 8:  prompt_tps=371.358, generation_tps=24.230, peak_memory=16.177, total_time=21.355
Trial 9:  prompt_tps=353.896, generation_tps=24.246, peak_memory=16.177, total_time=21.350
Trial 10:  prompt_tps=352.977, generation_tps=24.235, peak_memory=16.177, total_time=21.360
Averages: prompt_tps=360.728, generation_tps=24.239, peak_memory=16.176
```

the generation seemed to become slightly faster but it is probably just variability.

Logically I think this branch added overhead when waiting events, but also reduced overhead by reducing the number of completion handlers.

tseylerd:
@zcbenz Thank you very much for this fix!

I ported it to work with mlx v0.31.1 to make it possible to use fixed version with mlx-swift and mlx-swift-lm:
https://github.com/tseylerd/mlx/pull/1

I'd appreciate if someone could take a look and verify it.

angeloskath (approved):
👍

Thanks for running the benchmark and sorry for the delay.

PR #3294, opened 2026-03-22
Fix int32 overflow in Metal conv_general output offset for large tensors

agarwalprakhar2511:
Fixes #3248.

## Summary
- Promote the output write offset calculation in the Metal `implicit_gemm_conv_2d_general` kernel from 32-bit `int` to `size_t`, preventing silent output corruption when `batch * H_out * W_out * C_out` exceeds 2^31
- The already-64-bit `out_strides` in `MLXConvParams` were being multiplied by 32-bit `int` indices and stored back into an `int`, causing the address to wrap negative for large tensors
- Add a Metal regression test that crosses the 2^31 output-offset boundary with varying per-batch input values so both zeroed-output and wrong-batch-write failures are caught

## Test plan
- [x] `python/tests/test_conv.py TestConv.test_conv_general_large_output_offset` — passes
- [x] Full `python/tests/test_conv.py` suite — 19 tests, all pass (9 skipped = no Torch)
- [x] Ad hoc large-shape validation on `(30842, 64, 64, 17)` output — `first_diff = 0.0`, `last_diff = 0.0`

Made with [Cursor](https://cursor.com)

PR #3524, opened 2026-05-11
Detect int32 shape-product overflow at MLX compute-shape boundaries

### LINKED ISSUE #3327: [BUG] Arrays with >= 2^31 elements fail materialization and some indexing on Metal
I ran into this while trying to write an end-to-end regression for a Metal `conv_general` output-offset bug.

### Reproducer

```python
import mlx.core as mx

n = (2**31) // (64 * 64) + 2
x = mx.ones((n, 8, 8, 1), dtype=mx.float16)
w = mx.ones((1, 1, 1, 1), dtype=mx.float16)
y = mx.conv_general(x, w, input_dilation=(9, 9))
mx.eval(y)
```

**Output shape:** `(524290, 64, 64, 1)`, total elements: `2147491840`.

### Expected

Either the array evaluates successfully, or MLX raises a clean size-limit error with the true requested size.

### Actual

`mx.eval(y)` fails with a wrapped allocation size:

```
RuntimeError: [metal::malloc] Attempting to allocate 18446744069414600704 bytes which is greater than the maximum allowed buffer size of 86586540032 bytes.
```

`18446744069414600704` is `2^64 - 4294950912` consistent with signed-to-unsigned overflow in size bookkeeping.

### Related failures from the same `y`

```python
mx.reshape(y, (-1,))          # ValueError: [reshape] Cannot reshape array of size 2147491840 into shape (-2147475456).
mx.take(y, mx.array([0], dtype=mx.uint32))  # ValueError: [gather] Slice sizes must be in [0, a.shape(i)]. Got (1) for array with shape (-2147475456).
```

Lazy slice `y[-1]` reports the correct shape `(64, 64, 1)`, but `mx.eval(y[-1])` triggers the same wrapped `metal::malloc` error.

The inferred shape `-2147475456` is `2147491840` interpreted as signed `int32` confirming overflow is in shape/size accounting, not just the allocator.

### Boundary

Verified on `mlx 0.31.1` and local `0.31.2.dev20260325+84099a14b`:

| output elements              | reshape                  | eval                    | slice materialization   |
|------------------------------|--------------------------|-------------------------|-------------------------|
| 2147479552 (2³¹ - 4096, n=524287) | ✓                        | ✓                       | ✓                       |
| 2147483648 (exactly 2³¹, n=524288) | ✗ negative shape         | ✗ wrapped alloc         | ✗ wrapped alloc         |
| 2147491840 (reproducer, n=524290) | ✗ negative shape         | ✗ wrapped alloc         | ✗ wrapped alloc         |

The overflow triggers at exactly `2^31` elements. The last passing case `2^31 - 4096` falls naturally out of the `(n, 64, 64, 1)` shape grid.

### Root cause

The exact `2^31` boundary is consistent with `ShapeElem = int32_t` in `mlx/array.h`. The wrapped allocation and negative shapes appear to be downstream symptoms of shape/size overflow at the host level before Metal is involved. Additional affected paths likely exist.

### Environment

- **Device:** Apple M3 Max
- **Memory:** 137438953472 bytes
- **Max buffer length:** 86586540032 bytes
- Reproduced three times in independent fresh Python processes with identical results
- Reproduced on both installed wheel (`mlx 0.31.1`) and local main build

qflen:
## Summary

Fixes #3327. 

Shapes whose per-dim values fit in `int32` but whose product exceeds `2^31` silently wrapped at five sites: `reshape(-1)`, `flatten`, `take`, the Metal conv dispatcher's `out.size() / O` cast, and per-thread output offsets inside the Metal conv kernels. All five now either succeed correctly or raise `std::overflow_error` with a clear `[op] Shape dimension ...` message, extending the diagnostic style #3425 established at the Python binding to the internal C++ boundaries.

`ShapeElem` stays `int32_t`. No public API change. The `flatten` case from #2681 is also resolved.

## Root cause

- `Flatten::output_shape` accumulated `flat_size *= ...` as `int32_t` via `auto`.
- `Reshape::output_shape` and `unflatten` truncated `input.size() / size` from `size_t` to `ShapeElem`.
- `indices_or_default` used `std::multiplies<int>`.
- Metal `conv.cpp`: `int implicit_M = out.size() / O` truncated; `int implicit_M = N * oS[0] * oS[1]` wrapped. Wrapped neg dim fed `nbytes()` which sign-extends to ~18 EB.
- Metal `steel_conv{,_3d,_general}.h`: per-thread output offset `c_row * (N * groups) + c_col` wrapped `int32` when `M < 2^31` but `M * O > 2^31`.
- CUDA `conv/{gemm_conv,gemm_grouped_conv}.cu`: same `int mat_M = out.size() / O` truncation as the Metal sites.

## Change

- New `check_shape_dim(int64_t, std::string_view op)` in `mlx/utils.h`.
- `Flatten`, `Reshape`, `unflatten`, `indices_or_default` widen accumulators to `int64_t` and narrow through the helper. Backend-agnostic.
- Four Metal and two CUDA conv dispatchers guard `implicit_M` / `mat_M`. Widen Metal `inp_large` / `out_large` to `int64_t`.
- Three Metal steel conv kernel headers promote per-thread output offset to `size_t`. This is the substance of the prior attempts #3294 and #3320, now exercised by an end-to-end test.

## Tests (`tests/gpu_tests.cpp`)

- `test gpu int32 shape overflow errors` covers `flatten`, `reshape(-1)`, `take`, `eval`. Eval branch guarded by `max_buffer_length`.
- `test gpu conv2d large output offset`: output 2.15 G fp16 elements, varying batch values, allclose vs CPU. Fails on `y[-1]` without the `steel_conv_general.h` patch (verified by stash/restore).

## Validation

Apple M5 32GB, `Release` + `MLX_METAL_JIT=OFF`.

- CPU: `test_ops.py` (139), `test_array.py` (73), `test_conv.py` (18), `test_compile.py` (54), and the reproducer on CPU stream all pass.
- GPU / Metal: 248/248. Metal API validation layer: 0 errors.
- GPU / CUDA: deferred to CI; the change is a two-file mirror of the Metal dispatcher fix with no kernel-side component.

## Performance

One host-side `int64` compare per kernel dispatch. Trimmed-mean of 500 iters, per-call µs:

| Workload                                          | Baseline | Patched |
|---------------------------------------------------|---------:|--------:|
| `conv2d (32,64,64,64) -> (..,128)`, Metal         |    10080 |    9204 |
| `conv2d_general (16,64,64,1) -> (..,17)`, Metal   |     1164 |     717 |
| `take((8192,64), 5 idx)`, Metal                   |      326 |     205 |
| `reshape((4096,1024) -> -1)`, Metal               |       66 |      77 |
| `flatten((4096,1024))`, Metal                     |       83 |      74 |
| `conv2d (4,64,64,32) -> (..,64)`, CPU             |     3623 |    3348 |
| `reshape(-1)`, CPU                                |       27 |      22 |
| `flatten`, CPU                                    |       28 |      23 |

Patched is faster in 7/8 cells; deltas are within sub-ms jitter.

## Scope

I noticed that same-class fixes exist at adjacent sites the reproducer does not reach: 
`mlx/backend/metal/matmul.cpp` gather-MM dispatchers `tile` / `repeat` / `kron` / `concatenate` / `pad`, CPU `mlx/backend/cpu/conv.cpp`, and `dilate_size` / `conv_out_axis_size`. 

Left for follow-up to keep this PR focused, but I can also work on them here.

### DIFF
--- a/mlx/backend/metal/kernels/steel/conv/kernels/steel_conv.h
+++ b/mlx/backend/metal/kernels/steel/conv/kernels/steel_conv.h
@@ -135,7 +135,7 @@ implicit_gemm_conv_2d(
   C += tid.z * N;
 
   B += c_col * K;
-  C += static_cast<size_t>(c_row) * size_t(N * params->groups) + size_t(c_col);
+  C += static_cast<size_t>(c_row) * N * params->groups + c_col;
 
   const int2 offsets_a(0, c_row);
   const int2 offsets_b(0, c_col);
--- a/mlx/backend/metal/kernels/steel/conv/kernels/steel_conv.h
+++ b/mlx/backend/metal/kernels/steel/conv/kernels/steel_conv.h
@@ -135,7 +135,7 @@ implicit_gemm_conv_2d(
   C += tid.z * N;
 
   B += c_col * K;
-  C += c_row * (N * params->groups) + c_col;
+  C += static_cast<size_t>(c_row) * size_t(N * params->groups) + size_t(c_col);
 
   const int2 offsets_a(0, c_row);
   const int2 offsets_b(0, c_col);
--- a/mlx/utils.h
+++ b/mlx/utils.h
@@ -3,6 +3,10 @@
 #pragma once
 
 #include <exception>
+#include <limits>
+#include <sstream>
+#include <stdexcept>
+#include <string_view>
 #include <variant>
 
 #include "mlx/api.h"
@@ -95,6 +99,22 @@ inline Dtype result_type(const array& a, const array& b, const array& c) {
 MLX_API Dtype result_type(const std::vector<array>& arrays);
 
 MLX_API Shape broadcast_shapes(const Shape& s1, const Shape& s2);
+
+inline ShapeElem check_shape_dim(int64_t dim, std::string_view op = "") {
+  constexpr int64_t lo = std::numeric_limits<ShapeElem>::min();
+  constexpr int64_t hi = std::numeric_limits<ShapeElem>::max();
+  if (dim < lo || dim > hi) {
+    std::ostringstream msg;
+    if (!op.empty()) {
+      msg << "[" << op << "] ";
+    }
+    msg << "Shape dimension " << dim << " is outside the supported range ["
+        << lo << ", " << hi
+        << "]. MLX currently uses 32-bit integers for shape dimensions.";
+    throw std::overflow_error(msg.str());
+  }
+  return static_cast<ShapeElem>(dim);
+}
 
 /**
  * Returns the axis normalized to be in the range [0, ndim).
--- a/mlx/utils.h
+++ b/mlx/utils.h
@@ -100,20 +100,22 @@ MLX_API Dtype result_type(const std::vector<array>& arrays);
 
 MLX_API Shape broadcast_shapes(const Shape& s1, const Shape& s2);
 
-inline ShapeElem check_shape_dim(int64_t dim, std::string_view op = "") {
+template <typename T>
+inline ShapeElem safe_cast(T dim, std::string_view op = "") {
   constexpr int64_t lo = std::numeric_limits<ShapeElem>::min();
   constexpr int64_t hi = std::numeric_limits<ShapeElem>::max();
-  if (dim < lo || dim > hi) {
+  auto v = static_cast<int64_t>(dim);
+  if (v < lo || v > hi) {
     std::ostringstream msg;
     if (!op.empty()) {
       msg << "[" << op << "] ";
     }
-    msg << "Shape dimension " << dim << " is outside the supported range ["
-        << lo << ", " << hi
+    msg << "Shape dimension " << v << " is outside the supported range [" << lo
+        << ", " << hi
         << "]. MLX currently uses 32-bit integers for shape dimensions.";
     throw std::overflow_error(msg.str());
   }
-  return static_cast<ShapeElem>(dim);
+  return static_cast<ShapeElem>(v);
 }
 
 /**
--- a/tests/gpu_tests.cpp
+++ b/tests/gpu_tests.cpp
@@ -3,7 +3,6 @@
 #include <array>
 
 #include "doctest/doctest.h"
-#include "mlx/backend/gpu/device_info.h"
 #include "mlx/mlx.h"
 
 using namespace mlx::core;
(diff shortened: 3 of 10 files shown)


mlx/backend/metal/kernels/steel/conv/kernels/steel_conv.h
```diff
@@ -135,7 +135,7 @@ implicit_gemm_conv_2d(
   C += tid.z * N;
 
   B += c_col * K;
-  C += c_row * (N * params->groups) + c_col;
+  C += static_cast<size_t>(c_row) * size_t(N * params->groups) + size_t(c_col);
```
zcbenz: There is no need converting every operand, you can just convert the first one and the rest will promote automatically.


```suggestion
  C += static_cast<size_t>(c_row) * N * params->groups + c_col;
```

mlx/utils.h
```diff
@@ -96,6 +100,22 @@ MLX_API Dtype result_type(const std::vector<array>& arrays);
 
 MLX_API Shape broadcast_shapes(const Shape& s1, const Shape& s2);
 
+inline ShapeElem check_shape_dim(int64_t dim, std::string_view op = "") {
```
zcbenz: We can make this easier to use:

```suggestion
template <typename T>
inline ShapeElem safe_cast(T dim, std::string_view op = "") {
```

so you don't have to force converting the argument to int64_t:

```c++
    shape[infer_idx] = safe_cast(a.shape(ax) / size);
```

Also a helper named `check_` is usually supposed to return void or bool, using `_cast` as name would reduce cognitive burden.

qflen: Thanks for the reviews. Added changes.

tests/gpu_tests.cpp
```diff
@@ -477,6 +478,101 @@ TEST_CASE("test gpu validation") {
...
+  auto idx = array({0u}, uint32);
+  CHECK_THROWS_AS(take(y, idx), std::overflow_error);
+
+  // The conv dispatcher refuses to compute a >2^31-element output. eval
+  // allocates the ~4 GB float16 output before the dispatcher check fires,
+  // so skip on small-GPU devices.
+  size_t needed = size_t(n) * 64 * 64 * sizeof(float16_t);
+  auto max_buf = std::get<size_t>(gpu::device_info().at("max_buffer_length"));
+  if (max_buf >= needed) {
+    CHECK_THROWS_AS(eval(y), std::overflow_error);
```
zcbenz: Allocating such a large array would drag down the test speed a lot, I prefer just removing this test since it is not a serious issue.

zcbenz (approved):
Nice fix, thanks!

zcbenz:
Sorry I meant deleting the tests that actually allocates the 4GB array (including the "test gpu conv2d large output offset"), the tests that check overflow are nice to keep.

qflen:
Oh my bad, I'll rewrite.

dmlsrc:
Standalone `mx.conv3d` reproducer for the `steel_conv_3d.h` output-offset overflow your PR fixes.

Confirmed on MLX 0.31.2, macOS Tahoe 26.4.1, Apple M1 Max 64 GB.

```python
import time
import mlx.core as mx

mx.set_default_device(mx.gpu)
mx.set_cache_limit(1024 * 1024 * 1024)
print("MLX:", mx.__version__)

# Output shape: (1, 457, 72, 128, 512)
# M = 1 * 457 * 72 * 128 = 4,211,712
# N = 512
# M * N = 2,156,396,544  >  INT32_MAX (2,147,483,647)
#
# C=64 satisfies C_per_group % 16 == 0 (implicit path),
# and keeps input/weight memory reasonable.

x = mx.ones((1, 457, 72, 128, 64), dtype=mx.float16)
w = mx.ones((512, 3, 3, 3, 64), dtype=mx.float16)
t0 = time.perf_counter()
out = mx.conv3d(x, w, padding=1)
mx.eval(out)
print(f"conv eval: {time.perf_counter() - t0:.2f}s")

for label, d, expected in [
    ("before boundary", 454, 1728),
    ("after boundary",  455, 1728),
    ("last depth",      456, 1152),
]:
    vals = out[0, d, 36, 64, :8]
    mx.eval(vals)
    print(f"{label:16s} expected={expected:4d} actual={vals}")

print(f"peak GB: {mx.get_peak_memory() / 1e9:.2f}")
```

On 0.31.2 (before your fix):

```
before boundary  expected=1728 actual=array([1728, 1728, 1728, ..., 1728, 1728, 1728], dtype=float16)
after boundary   expected=1728 actual=array([0, 0, 0, ..., 0, 0, 0], dtype=float16)
last depth       expected=1152 actual=array([0, 0, 0, ..., 0, 0, 0], dtype=float16)
```

The overflow boundary is at `2^31 / (72 * 128 * 512) = depth index 455.11`, which is exactly where the output goes to zero. The affected region is unwritten (no garbage values, just zeros), consistent with the stores targeting an incorrect address and missing the intended output region entirely.

This may be useful as a regression test for the `steel_conv_3d.h` path specifically, since it exercises `mx.conv3d` directly rather than `conv_general`.

qflen:
Thanks for the repro and pinpointing the 2^31 / (72 * 128 * 512) boundary.

The `size_t` cast on `steel_conv_3d.h:97` covers your case since per-thread offsets `store_result_safe` adds on top are tile-bounded, so they don't re-introduce overflow.

With @zcbenz's earlier feedback, the >=4 GB-allocating tests were dropped, and this one would also need ~4.3 GB to trigger (M * N > INT32_MAX). The host-side overflow checks in `gpu_tests.cpp` stay.

Issue #3514, opened 2026-05-10
Add full `array-api` tests


ilan-gold:
From our discussion here https://github.com/ml-explore/mlx/issues/48#issuecomment-4378140759 I agree with @mtsokol that the sensible thing to do would be not to one-off #3484 but instead to fix things wholesale by adding a full test suite from  https://github.com/data-apis/array-api-tests

mtsokol:
I think we can mimic JAX and NumPy when it comes to array-api-tests suite setup: https://github.com/ml-explore/mlx/pull/3526

ev-br:
Now that https://github.com/ml-explore/mlx/pull/4334 has landed, the array-api test suite does not need patching to run. 
Local testing showed that about 3/4 of the test suite passes, and the rest are mostly small-ish issues, plus a couple of larger items which probably are just fundamental differences between MLX and the Array API spec (and the test suite has a mechanism to mark them as such).

https://github.com/data-apis/array-api-compat/issues/450 tracks an overall status, and https://github.com/data-apis/array-api-compat/issues/450#issuecomment-5345122998 has a specific suggestion.
Overall, compatibility has been improving recently, mostly thanks to tireless work of @aaishwarymishra and @prady0t and reviewers (thanks!).

Would the maintainers be open to reviewing a PR setting up the CI run similar to gh-3526, with xfails for the currently known failures? Or would y'all prefer that we track the status externally and keep sending individual PRs for small issues found in local testing?

Either way, I'm happy to help reviewing from the Array API perspective if that'd be helpful.

zcbenz:
I'm good running array-api tests in our CI, with the known failures masked out.

PR #3526, opened 2026-05-11
Introduce Array API Tests suite

mtsokol:
Addresses https://github.com/ml-explore/mlx/issues/3514

Hi @ilan-gold,
Here's an array-api-tests CI setup based on:
- JAX: https://github.com/jax-ml/jax/blob/main/.github/workflows/jax-array-api.yml
- NumPy: https://github.com/numpy/numpy/blob/3bea241a0fbf5610f876dc54d4caf7380f75471e/.github/workflows/linux.yml#L296

I anticipate that the library isn't 100% compatible (there's always some edge case) so we still need to prepare:
`--skips-file "${GITHUB_WORKSPACE}/tests/array_api_skips.txt"` file.

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -86,6 +86,34 @@ jobs:
       - uses: ./.github/actions/build-windows
       - uses: ./.github/actions/test-windows
 
+  array_api_tests:
+    name: Array API Tests
+    if: github.repository == 'ml-explore/mlx'
+    needs: check_lint
+    runs-on: ubuntu-22.04
+    steps:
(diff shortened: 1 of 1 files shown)


mtsokol:
Ok, we first need:
https://data-apis.org/array-api/latest/API_specification/generated/array_api.info.__array_namespace_info__.html
to run the test suite.

zcbenz:
I'm closing this PR since it seems that we are very far from making the test suite work with MLX.

Zethson:
@zcbenz instead of closing it, I'd instead regularly merge into this PR to retrigger the CI and understand the status.

Issue #2848, opened 2025-11-28
[Feature] Construct `mx.array` from `mps` and `cuda` arrays from other frameworks 


nbardy:
**Describe the bug**
dlpack doesn't work on mps tensors.

**To Reproduce**

Include code snippet
```python
>>> import torch
>>> import mlx.core as mx
>>> mx.array(torch.tensor([1,2,3]).to("mps"))
Exception ignored in PyObject_HasAttrString(); consider using PyObject_HasAttrStringWithError(), PyObject_GetOptionalAttrString() or PyObject_GetAttrString():
Traceback (most recent call last):
  File "<python-input-7>", line 1, in <module>
RuntimeError: imag is not implemented for tensors with non-complex dtypes.
Exception ignored in PyObject_HasAttrString(); consider using PyObject_HasAttrStringWithError(), PyObject_GetOptionalAttrString() or PyObject_GetAttrString():
Traceback (most recent call last):
  File "<python-input-7>", line 1, in <module>
RuntimeError: imag is not implemented for tensors with non-complex dtypes.
Traceback (most recent call last):
  File "<python-input-7>", line 1, in <module>
    mx.array(torch.tensor([1,2,3]).to("mps"))
    ~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
ValueError: Invalid type torch.Tensor received in array initialization.
>>> mx.array(torch.tensor([1,2,3]).to("cpu"))
array([1, 2, 3], dtype=int64)
>>>
```

**Expected behavior**

**Desktop (please complete the following information):**
 - MacOS 15.5 (24F74)
```
>>> mx.__version__
'0.30.0'
```

**Additional context**

awni:
Yes right now you can only construct MLX arrays from CPU tensors in other frameworks. It may be possible to change that to accept MPS / CUDA arrays.

awni:
Regarding `mps` from PyTorch it seems like PyTorch does not support dlpack with MPS which is too bad. But until they do so we can't really accept MPS PyTorch tensors via that protocol.


```
import torch
t = torch.tensor([1,2,3]).to("mps")
t.__dlpack__()
```
```
RuntimeError: Cannot pack tensors on mps:0
```

See related discussion: https://github.com/wjakob/nanobind/discussions/1233#discussioncomment-15129537

PR #3531, opened 2026-05-11
Add Metal DLPack zero-copy sharing

XXXXRT666:
## Proposed changes

This draft adds zero-copy Metal DLPack sharing for MLX arrays and PyTorch MPS tensors.

This PR builds on the merged [DLPack import PR #3495](https://github.com/ml-explore/mlx/pull/3495) and requires [nanobind support](https://github.com/wjakob/nanobind/pull/1338).

The main changes are:

- Import Metal DLPack arrays by wrapping the underlying Metal buffer instead of copying through CPU.
- Export MLX arrays to Metal DLPack using the MLX Metal buffer and DLPack `byte_offset`.
- Add `mx.from_dlpack(..., copy=...)` controls for Metal DLPack inputs.
- Keep `mx.array(...)` zero-copy for Metal DLPack inputs unless an explicit different dtype is requested.
- Document the explicit synchronization requirements between PyTorch MPS and MLX.

The shared lifetime is tied to the exported or imported buffer. Synchronization remains explicit: PyTorch writes require `torch.mps.synchronize()` before MLX reads, and MLX writes require `mx.eval(...)` before PyTorch reads.

For MLX arrays exported to PyTorch, later MLX updates may rebind the MLX array to a new buffer while the PyTorch tensor continues to reference the exported buffer.

## Checklist

### DIFF
--- a/docs/src/usage/numpy.rst
+++ b/docs/src/usage/numpy.rst
@@ -76,6 +76,7 @@ PyTorch
 -------
 
 PyTorch supports DLPack inputs and can import MLX arrays directly.
+MLX can also import PyTorch tensors through DLPack with ``mx.array``.
 
 .. code-block:: python
 
@@ -84,7 +85,53 @@ PyTorch supports DLPack inputs and can import MLX arrays directly.
 
   a = mx.arange(3)
   b = torch.tensor(a)
-  c = mx.array(b.cpu())
+  c = mx.array(b)
+
+Creating an MLX array from a CPU tensor copies the data into MLX-owned storage.
+The arrays do not share memory:
+
+.. code-block:: python
+
+  b = torch.arange(3)
+  c = mx.array(b)
+
+  b += 10
+  print(c.tolist()) # [0, 1, 2]
+
+Metal DLPack inputs are different. If a PyTorch MPS tensor is passed to
+``mx.array``, MLX imports the underlying Metal buffer without copying it. The
+PyTorch tensor and the MLX array then share the same storage.
+
+Since the buffer is shared across frameworks, synchronization has to be managed
+explicitly. After PyTorch writes to an MPS tensor, call
+``torch.mps.synchronize()`` before reading the shared data from MLX. After MLX
+writes to the shared array, call ``mx.eval`` on the MLX result before reading
+the shared data from PyTorch. Without these synchronization points, the other
+framework may read the shared buffer before the producer has finished writing,
+so it can observe stale data.
+
+.. code-block:: python
+
+  b = torch.arange(3, device="mps", dtype=torch.float32)
+  torch.mps.synchronize()
+  c = mx.array(b) # zero-copy Metal DLPack import
+
+  b.add_(10)
+  torch.mps.synchronize()
+  print(c.tolist()) # [10.0, 11.0, 12.0]
+
+Updates made by MLX can also be observed from PyTorch after the MLX computation
+has been evaluated:
+
+.. code-block:: python
+
+  b = torch.arange(3, device="mps", dtype=torch.float32)
+  torch.mps.synchronize()
+  c = mx.array(b)
+
+  c += 10
+  mx.eval(c)
+  print(b.cpu()) # tensor([10., 11., 12.])
 
 JAX
 ---
--- a/docs/src/usage/numpy.rst
+++ b/docs/src/usage/numpy.rst
@@ -102,7 +102,8 @@ The arrays do not share memory:
 Metal DLPack inputs are different. If a PyTorch MPS tensor is passed to
 ``mx.array`` or to ``mx.from_dlpack`` with ``copy=None`` or ``copy=False``, MLX
 imports the underlying Metal buffer without copying it. The PyTorch tensor and
-the MLX array then share the same storage.
+the MLX array then share the same storage. MLX arrays exported to PyTorch with
+DLPack are also shared without a copy.
 
 Since the buffer is shared across frameworks, synchronization has to be managed
 explicitly. After PyTorch writes to an MPS tensor, call
@@ -134,6 +135,10 @@ has been evaluated:
   c += 10
   mx.eval(c)
   print(b.cpu()) # tensor([10., 11., 12.])
+
+For MLX arrays exported to PyTorch, the share is tied to the exported buffer.
+MLX updates after export may rebind the MLX array to a new buffer, while the
+PyTorch tensor continues to reference the exported buffer.
 
 Use ``mx.from_dlpack`` when you need to control the copy behavior. Specifying
 ``copy=True`` asks MLX to create a new array instead of sharing the Metal
--- a/docs/src/usage/numpy.rst
+++ b/docs/src/usage/numpy.rst
@@ -76,7 +76,8 @@ PyTorch
 -------
 
 PyTorch supports DLPack inputs and can import MLX arrays directly.
-MLX can also import PyTorch tensors through DLPack with ``mx.array``.
+MLX can also import PyTorch tensors through DLPack with ``mx.array`` or
+``mx.from_dlpack``.
 
 .. code-block:: python
 
@@ -99,8 +100,9 @@ The arrays do not share memory:
   print(c.tolist()) # [0, 1, 2]
 
 Metal DLPack inputs are different. If a PyTorch MPS tensor is passed to
-``mx.array``, MLX imports the underlying Metal buffer without copying it. The
-PyTorch tensor and the MLX array then share the same storage.
+``mx.array`` or to ``mx.from_dlpack`` with ``copy=None`` or ``copy=False``, MLX
+imports the underlying Metal buffer without copying it. The PyTorch tensor and
+the MLX array then share the same storage.
 
 Since the buffer is shared across frameworks, synchronization has to be managed
 explicitly. After PyTorch writes to an MPS tensor, call
@@ -132,6 +134,20 @@ has been evaluated:
   c += 10
   mx.eval(c)
   print(b.cpu()) # tensor([10., 11., 12.])
+
+Use ``mx.from_dlpack`` when you need to control the copy behavior. Specifying
+``copy=True`` asks MLX to create a new array instead of sharing the Metal
+buffer:
+
+.. code-block:: python
+
+  b = torch.arange(3, device="mps", dtype=torch.float32)
+  torch.mps.synchronize()
+  c = mx.from_dlpack(b, copy=True)
+
+  b.add_(10)
+  torch.mps.synchronize()
+  print(c.tolist()) # [0.0, 1.0, 2.0]
 
 JAX
 ---
--- a/docs/src/usage/numpy.rst
+++ b/docs/src/usage/numpy.rst
@@ -77,90 +77,58 @@ PyTorch
 
 PyTorch supports DLPack inputs and can import MLX arrays directly.
 MLX can also import PyTorch tensors through DLPack with ``mx.array`` or
-``mx.from_dlpack``.
-
-Creating an MLX array from a CPU tensor copies the data into MLX-owned storage.
-The arrays do not share memory:
+``mx.from_dlpack``. Use ``torch.as_tensor`` to import an MLX array with
+DLPack; ``torch.tensor`` copies the data instead:
 
 .. code-block:: python
 
   import mlx.core as mx
   import torch
 
-  b = torch.arange(3).cpu()
+  a = mx.arange(3, dtype=mx.float32)
+  mx.eval(a)
+
+  shared = torch.as_tensor(a)
+  copied = torch.tensor(a)
+
+Creating an MLX array from a CPU tensor copies the data into MLX-owned storage.
+The arrays do not share memory:
+
+.. code-block:: python
+
+  b = torch.arange(3)
   c = mx.array(b)
 
   b += 10
   print(c.tolist()) # [0, 1, 2]
 
 Metal DLPack inputs are different. If a PyTorch MPS tensor is passed to
-``mx.array`` or to ``mx.from_dlpack`` with ``copy=None``, MLX shares the
-underlying Metal buffer when it is not private. Private Metal buffers are copied
-into MLX-managed storage instead. Passing ``copy=False`` requires sharing and
-raises an error if a copy would be needed. MLX arrays exported to PyTorch with
-DLPack are also shared without a copy. In particular, PyTorch 2.12 and later use
-shared storage for ordinary MPS tensors on Apple silicon, while older PyTorch
-versions may use private storage and require a copy on import.
+``mx.array`` or to ``mx.from_dlpack`` with ``copy=None``, MLX imports it
+without a copy when the underlying Metal buffer is not private. Private Metal
+buffers are copied into MLX-managed storage instead. Passing ``copy=False``
+requires zero-copy import and raises an error if a copy would be needed.
+Passing ``copy=True`` asks MLX to create a new array instead of reusing the
+Metal buffer. MLX arrays exported to PyTorch with DLPack are also exported
+without a copy on Metal.
 
-Since the buffer is shared across frameworks, synchronization has to be managed
-explicitly. After PyTorch writes to an MPS tensor, call
-``torch.mps.synchronize()`` before reading the shared data from MLX. After MLX
-writes to the shared array, call ``mx.eval`` on the MLX result before reading
-the shared data from PyTorch. Without these synchronization points, the other
-framework may read the shared buffer before the producer has finished writing,
-so it can observe stale data.
-
-Do not rely on regular MLX update expressions to preserve sharing. MLX may
-rebind an array to a new buffer when evaluating an expression, while PyTorch
-continues to reference the original shared buffer. Use
-:meth:`mlx.core.array.copy` to write into an array's existing storage when you
-need the update to be visible to another framework sharing the same buffer:
+In particular, PyTorch 2.12 and later use shared storage for ordinary MPS
+tensors on Apple silicon, while older PyTorch versions may use private storage
+and require a copy on import. DLPack conversion does not synchronize pending
+Metal work; synchronize or evaluate the producing framework before reading the
+converted array.
 
 .. code-block:: python
 
   b = torch.arange(3, device="mps", dtype=torch.float32)
   torch.mps.synchronize()
   c = mx.array(b) # zero-copy if the Metal buffer is not private
-
-  b.add_(10)
-  torch.mps.synchronize()
-  print(c.tolist()) # [10.0, 11.0, 12.0]
-
-  c.copy(mx.array([4, 5, 6], dtype=mx.float32))
-  mx.eval(c)
-  print(b.cpu()) # tensor([4., 5., 6.])
-
-The same synchronization rules apply when PyTorch imports an MLX array through
-DLPack. Use ``torch.as_tensor`` to share the buffer with PyTorch; ``torch.tensor``
-copies the data instead:
+  d = mx.from_dlpack(b, copy=True) # explicit copy
 
 .. code-block:: python
 
   a = mx.arange(3, dtype=mx.float32)
   mx.eval(a)
-  b = torch.as_tensor(a)
-
-  b.add_(10)
-  torch.mps.synchronize()
-  print(a.tolist()) # [10.0, 11.0, 12.0]
-
-  a.copy(mx.array([4, 5, 6], dtype=mx.float32))
-  mx.eval(a)
-  print(b.cpu()) # tensor([4., 5., 6.])
-
-Use ``mx.from_dlpack`` when you need to control the copy behavior. Specifying
-``copy=True`` asks MLX to create a new array instead of sharing the Metal
-buffer:
-
-.. code-block:: python
-
-  b = torch.arange(3, device="mps", dtype=torch.float32)
-  torch.mps.synchronize()
-  c = mx.from_dlpack(b, copy=True)
-
-  b.add_(10)
-  torch.mps.synchronize()
-  print(c.tolist()) # [0.0, 1.0, 2.0]
+  b = torch.as_tensor(a) # zero-copy DLPack import on Metal
 
 JAX
 ---
--- a/docs/src/usage/numpy.rst
+++ b/docs/src/usage/numpy.rst
@@ -76,9 +76,10 @@ PyTorch
 -------
 
 PyTorch supports DLPack inputs and can import MLX arrays directly.
-MLX can also import PyTorch tensors through DLPack with ``mx.array`` or
+MLX can also import PyTorch tensors through DLPack with ``mx.asarray`` or
 ``mx.from_dlpack``. Use ``torch.as_tensor`` to import an MLX array with
-DLPack; ``torch.tensor`` copies the data instead:
+DLPack; ``torch.tensor`` copies the data instead. Similarly, ``mx.asarray``
+can share DLPack inputs when possible, while ``mx.array`` copies:
 
 .. code-block:: python
 
@@ -103,13 +104,14 @@ The arrays do not share memory:
   print(c.tolist()) # [0, 1, 2]
 
 Metal DLPack inputs are different. If a PyTorch MPS tensor is passed to
-``mx.array`` or to ``mx.from_dlpack`` with ``copy=None``, MLX imports it
+``mx.asarray`` or to ``mx.from_dlpack`` with ``copy=None``, MLX imports it
 without a copy when the underlying Metal buffer is not private. Private Metal
 buffers are copied into MLX-managed storage instead. Passing ``copy=False``
 requires zero-copy import and raises an error if a copy would be needed.
 Passing ``copy=True`` asks MLX to create a new array instead of reusing the
-Metal buffer. MLX arrays exported to PyTorch with DLPack are also exported
-without a copy on Metal.
+Metal buffer. ``mx.array`` also creates a new array instead of reusing the
+Metal buffer. MLX arrays exported to PyTorch with DLPack are exported without a
+copy on Metal.
 
 In particular, PyTorch 2.12 and later use shared storage for ordinary MPS
 tensors on Apple silicon, while older PyTorch versions may use private storage
@@ -121,7 +123,7 @@ converted array.
 
   b = torch.arange(3, device="mps", dtype=torch.float32)
   torch.mps.synchronize()
-  c = mx.array(b) # zero-copy if the Metal buffer is not private
+  c = mx.asarray(b) # zero-copy if the Metal buffer can be reused
   d = mx.from_dlpack(b, copy=True) # explicit copy
 
 .. code-block:: python
--- a/docs/src/usage/numpy.rst
+++ b/docs/src/usage/numpy.rst
@@ -94,10 +94,13 @@ The arrays do not share memory:
   print(c.tolist()) # [0, 1, 2]
 
 Metal DLPack inputs are different. If a PyTorch MPS tensor is passed to
-``mx.array`` or to ``mx.from_dlpack`` with ``copy=None`` or ``copy=False``, MLX
-imports the underlying Metal buffer without copying it. The PyTorch tensor and
-the MLX array then share the same storage. MLX arrays exported to PyTorch with
-DLPack are also shared without a copy.
+``mx.array`` or to ``mx.from_dlpack`` with ``copy=None``, MLX shares the
+underlying Metal buffer when it is host-accessible. Private Metal buffers are
+copied into MLX-managed storage instead. Passing ``copy=False`` requires
+sharing and raises an error if a copy would be needed. MLX arrays exported to
+PyTorch with DLPack are also shared without a copy. In particular, PyTorch 2.12
+and later use shared storage for ordinary MPS tensors on Apple silicon, while
+older PyTorch versions may use private storage and require a copy on import.
 
 Since the buffer is shared across frameworks, synchronization has to be managed
 explicitly. After PyTorch writes to an MPS tensor, call
@@ -117,7 +120,7 @@ need the update to be visible to another framework sharing the same buffer:
 
   b = torch.arange(3, device="mps", dtype=torch.float32)
   torch.mps.synchronize()
-  c = mx.array(b) # zero-copy Metal DLPack import
+  c = mx.array(b) # zero-copy if the Metal buffer is host-accessible
 
   b.add_(10)
   torch.mps.synchronize()
--- a/docs/src/usage/numpy.rst
+++ b/docs/src/usage/numpy.rst
@@ -79,21 +79,15 @@ PyTorch supports DLPack inputs and can import MLX arrays directly.
 MLX can also import PyTorch tensors through DLPack with ``mx.array`` or
 ``mx.from_dlpack``.
 
-.. code-block:: python
-
-  import mlx.core as mx
-  import torch
-
-  a = mx.arange(3)
-  b = torch.tensor(a)
-  c = mx.array(b)
-
 Creating an MLX array from a CPU tensor copies the data into MLX-owned storage.
 The arrays do not share memory:
 
 .. code-block:: python
 
-  b = torch.arange(3)
+  import mlx.core as mx
+  import torch
+
+  b = torch.arange(3).cpu()
   c = mx.array(b)
 
   b += 10
@@ -113,6 +107,12 @@ the shared data from PyTorch. Without these synchronization points, the other
 framework may read the shared buffer before the producer has finished writing,
 so it can observe stale data.
 
+Do not rely on regular MLX update expressions to preserve sharing. MLX may
+rebind an array to a new buffer when evaluating an expression, while PyTorch
+continues to reference the original shared buffer. Use
+:meth:`mlx.core.array.copy` to write into an array's existing storage when you
+need the update to be visible to another framework sharing the same buffer:
+
 .. code-block:: python
 
   b = torch.arange(3, device="mps", dtype=torch.float32)
@@ -123,22 +123,27 @@ so it can observe stale data.
   torch.mps.synchronize()
   print(c.tolist()) # [10.0, 11.0, 12.0]
 
-Updates made by MLX can also be observed from PyTorch after the MLX computation
-has been evaluated:
+  c.copy(mx.array([4, 5, 6], dtype=mx.float32))
+  mx.eval(c)
+  print(b.cpu()) # tensor([4., 5., 6.])
+
+The same synchronization rules apply when PyTorch imports an MLX array through
+DLPack. Use ``torch.as_tensor`` to share the buffer with PyTorch; ``torch.tensor``
+copies the data instead:
 
 .. code-block:: python
 
-  b = torch.arange(3, device="mps", dtype=torch.float32)
+  a = mx.arange(3, dtype=mx.float32)
+  mx.eval(a)
+  b = torch.as_tensor(a)
+
+  b.add_(10)
   torch.mps.synchronize()
-  c = mx.array(b)
+  print(a.tolist()) # [10.0, 11.0, 12.0]
 
-  c += 10
-  mx.eval(c)
-  print(b.cpu()) # tensor([10., 11., 12.])
-
-For MLX arrays exported to PyTorch, the share is tied to the exported buffer.
-MLX updates after export may rebind the MLX array to a new buffer, while the
-PyTorch tensor continues to reference the exported buffer.
+  a.copy(mx.array([4, 5, 6], dtype=mx.float32))
+  mx.eval(a)
+  print(b.cpu()) # tensor([4., 5., 6.])
 
 Use ``mx.from_dlpack`` when you need to control the copy behavior. Specifying
 ``copy=True`` asks MLX to create a new array instead of sharing the Metal
--- a/docs/src/usage/numpy.rst
+++ b/docs/src/usage/numpy.rst
@@ -95,12 +95,12 @@ The arrays do not share memory:
 
 Metal DLPack inputs are different. If a PyTorch MPS tensor is passed to
 ``mx.array`` or to ``mx.from_dlpack`` with ``copy=None``, MLX shares the
-underlying Metal buffer when it is host-accessible. Private Metal buffers are
-copied into MLX-managed storage instead. Passing ``copy=False`` requires
-sharing and raises an error if a copy would be needed. MLX arrays exported to
-PyTorch with DLPack are also shared without a copy. In particular, PyTorch 2.12
-and later use shared storage for ordinary MPS tensors on Apple silicon, while
-older PyTorch versions may use private storage and require a copy on import.
+underlying Metal buffer when it is not private. Private Metal buffers are copied
+into MLX-managed storage instead. Passing ``copy=False`` requires sharing and
+raises an error if a copy would be needed. MLX arrays exported to PyTorch with
+DLPack are also shared without a copy. In particular, PyTorch 2.12 and later use
+shared storage for ordinary MPS tensors on Apple silicon, while older PyTorch
+versions may use private storage and require a copy on import.
 
 Since the buffer is shared across frameworks, synchronization has to be managed
 explicitly. After PyTorch writes to an MPS tensor, call
@@ -120,7 +120,7 @@ need the update to be visible to another framework sharing the same buffer:
 
   b = torch.arange(3, device="mps", dtype=torch.float32)
   torch.mps.synchronize()
-  c = mx.array(b) # zero-copy if the Metal buffer is host-accessible
+  c = mx.array(b) # zero-copy if the Metal buffer is not private
 
   b.add_(10)
   torch.mps.synchronize()
--- a/docs/src/usage/numpy.rst
+++ b/docs/src/usage/numpy.rst
@@ -109,8 +109,7 @@ without a copy when the underlying Metal buffer is not private. Private Metal
 buffers are copied into MLX-managed storage instead. Passing ``copy=False``
 requires zero-copy import and raises an error if a copy would be needed.
 Passing ``copy=True`` asks MLX to create a new array instead of reusing the
-Metal buffer. Copied DLPack inputs are materialized as row-contiguous MLX
-arrays; zero-copy imports preserve the DLPack strides. ``mx.array`` also
+Metal buffer. Zero-copy imports preserve the DLPack strides. ``mx.array`` also
 creates a new array instead of reusing the Metal buffer. MLX arrays exported to
 PyTorch with DLPack are exported without a copy on Metal.
 
--- a/docs/src/usage/numpy.rst
+++ b/docs/src/usage/numpy.rst
@@ -109,9 +109,10 @@ without a copy when the underlying Metal buffer is not private. Private Metal
 buffers are copied into MLX-managed storage instead. Passing ``copy=False``
 requires zero-copy import and raises an error if a copy would be needed.
 Passing ``copy=True`` asks MLX to create a new array instead of reusing the
-Metal buffer. ``mx.array`` also creates a new array instead of reusing the
-Metal buffer. MLX arrays exported to PyTorch with DLPack are exported without a
-copy on Metal.
+Metal buffer. Copied DLPack inputs are materialized as row-contiguous MLX
+arrays; zero-copy imports preserve the DLPack strides. ``mx.array`` also
+creates a new array instead of reusing the Metal buffer. MLX arrays exported to
+PyTorch with DLPack are exported without a copy on Metal.
 
 In particular, PyTorch 2.12 and later use shared storage for ordinary MPS
 tensors on Apple silicon, while older PyTorch versions may use private storage
--- a/mlx/array.cpp
+++ b/mlx/array.cpp
@@ -64,6 +64,7 @@ array array::unsafe_weak_copy(const array& other) {
       other.data_size(),
       other.strides(),
       other.flags(),
+      0,
       [](auto) {});
   cpy.array_desc_->offset = other.array_desc_->offset;
   return cpy;
@@ -180,9 +181,10 @@ void array::set_data(
     size_t data_size,
     Strides strides,
     Flags flags,
+    int64_t offset,
     Deleter d) {
   array_desc_->data = std::make_shared<Data>(buffer, d);
-  array_desc_->offset = 0;
+  array_desc_->offset = itemsize() * offset;
   array_desc_->data_size = data_size;
   array_desc_->strides = std::move(strides);
   array_desc_->flags = flags;
--- a/mlx/array.cpp
+++ b/mlx/array.cpp
@@ -184,7 +184,7 @@ void array::set_data(
     int64_t offset,
     Deleter d) {
   array_desc_->data = std::make_shared<Data>(buffer, d);
-  array_desc_->offset = itemsize() * offset;
+  array_desc_->offset = offset;
   array_desc_->data_size = data_size;
   array_desc_->strides = std::move(strides);
   array_desc_->flags = flags;
--- a/mlx/backend/cuda/allocator.cpp
+++ b/mlx/backend/cuda/allocator.cpp
@@ -398,6 +398,10 @@ void* Buffer::raw_ptr() {
   return cbuf.data;
 }
 
+bool Buffer::is_host_accessible() const {
+  return true;
+}
+
 } // namespace allocator
 
 size_t get_active_memory() {
--- a/mlx/backend/cuda/allocator.cpp
+++ b/mlx/backend/cuda/allocator.cpp
@@ -399,11 +399,7 @@ void* Buffer::raw_ptr() {
 }
 
 bool can_reuse_alien_buffer(void* ptr) {
-  if (!ptr) {
-    return true;
-  }
-  auto& cbuf = *static_cast<cu::CudaBuffer*>(ptr);
-  return cbuf.device == -1;
+  return true;
 }
 
 } // namespace allocator
--- a/mlx/backend/cuda/allocator.cpp
+++ b/mlx/backend/cuda/allocator.cpp
@@ -399,7 +399,11 @@ void* Buffer::raw_ptr() {
 }
 
 bool Buffer::is_host_accessible() const {
-  return true;
+  if (!ptr_) {
+    return true;
+  }
+  auto& cbuf = *static_cast<cu::CudaBuffer*>(ptr_);
+  return cbuf.device == -1;
 }
 
 } // namespace allocator
--- a/mlx/backend/cuda/allocator.cpp
+++ b/mlx/backend/cuda/allocator.cpp
@@ -398,14 +398,6 @@ void* Buffer::raw_ptr() {
   return cbuf.data;
 }
 
-bool Buffer::is_host_accessible() const {
-  if (!ptr_) {
-    return true;
-  }
-  auto& cbuf = *static_cast<cu::CudaBuffer*>(ptr_);
-  return cbuf.device == -1;
-}
-
 } // namespace allocator
 
 size_t get_active_memory() {
--- a/mlx/backend/cuda/allocator.cpp
+++ b/mlx/backend/cuda/allocator.cpp
@@ -398,6 +398,14 @@ void* Buffer::raw_ptr() {
   return cbuf.data;
 }
 
+bool can_reuse_alien_buffer(void* ptr) {
+  if (!ptr) {
+    return true;
+  }
+  auto& cbuf = *static_cast<cu::CudaBuffer*>(ptr);
+  return cbuf.device == -1;
+}
+
 } // namespace allocator
 
 size_t get_active_memory() {
--- a/mlx/backend/metal/allocator.cpp
+++ b/mlx/backend/metal/allocator.cpp
@@ -24,7 +24,23 @@ void* Buffer::raw_ptr() {
   if (!ptr_) {
     return nullptr;
   }
-  return static_cast<MTL::Buffer*>(ptr_)->contents();
+  auto* buf = static_cast<MTL::Buffer*>(ptr_);
+  auto* contents = buf->contents();
+  if (!contents && buf->length() > 0) {
+    throw std::runtime_error(
+        "[metal::Buffer::raw_ptr] Cannot access Metal buffer contents on the "
+        "host. The buffer is not CPU-addressable, for example because it uses "
+        "private storage.");
+  }
+  return contents;
+}
+
+bool Buffer::is_host_accessible() const {
+  if (!ptr_) {
+    return true;
+  }
+  auto* buf = static_cast<MTL::Buffer*>(ptr_);
+  return buf->storageMode() != MTL::StorageModePrivate;
 }
 
 } // namespace allocator
--- a/mlx/backend/metal/allocator.cpp
+++ b/mlx/backend/metal/allocator.cpp
@@ -24,15 +24,13 @@ void* Buffer::raw_ptr() {
   if (!ptr_) {
     return nullptr;
   }
-  auto* buf = static_cast<MTL::Buffer*>(ptr_);
-  auto* contents = buf->contents();
-  if (!contents && buf->length() > 0) {
+  if (!is_host_accessible()) {
     throw std::runtime_error(
         "[metal::Buffer::raw_ptr] Cannot access Metal buffer contents on the "
         "host. The buffer is not CPU-addressable, for example because it uses "
         "private storage.");
   }
-  return contents;
+  return static_cast<MTL::Buffer*>(ptr_)->contents();
 }
 
 bool Buffer::is_host_accessible() const {
--- a/mlx/backend/metal/allocator.cpp
+++ b/mlx/backend/metal/allocator.cpp
@@ -24,21 +24,14 @@ void* Buffer::raw_ptr() {
   if (!ptr_) {
     return nullptr;
   }
-  if (!is_host_accessible()) {
+  auto* buf = static_cast<MTL::Buffer*>(ptr_);
+  if (buf->storageMode() == MTL::StorageModePrivate) {
     throw std::runtime_error(
         "[metal::Buffer::raw_ptr] Cannot access Metal buffer contents on the "
         "host. The buffer is not CPU-addressable, for example because it uses "
         "private storage.");
   }
-  return static_cast<MTL::Buffer*>(ptr_)->contents();
-}
-
-bool Buffer::is_host_accessible() const {
-  if (!ptr_) {
-    return true;
-  }
-  auto* buf = static_cast<MTL::Buffer*>(ptr_);
-  return buf->storageMode() != MTL::StorageModePrivate;
+  return buf->contents();
 }
 
 } // namespace allocator
--- a/mlx/backend/metal/allocator.cpp
+++ b/mlx/backend/metal/allocator.cpp
@@ -7,6 +7,7 @@
 
 #include <mach/vm_page_size.h>
 #include <unistd.h>
+#include <cassert>
 #include <cstdlib>
 
 namespace mlx::core {
@@ -25,12 +26,7 @@ void* Buffer::raw_ptr() {
     return nullptr;
   }
   auto* buf = static_cast<MTL::Buffer*>(ptr_);
-  if (buf->storageMode() == MTL::StorageModePrivate) {
-    throw std::runtime_error(
-        "[metal::Buffer::raw_ptr] Cannot access Metal buffer contents on the "
-        "host. The buffer is not CPU-addressable, for example because it uses "
-        "private storage.");
-  }
+  assert(buf->storageMode() != MTL::StorageModePrivate);
   return buf->contents();
 }
 
--- a/mlx/backend/metal/allocator.cpp
+++ b/mlx/backend/metal/allocator.cpp
@@ -34,6 +34,14 @@ void* Buffer::raw_ptr() {
   return buf->contents();
 }
 
+bool can_reuse_alien_buffer(void* ptr) {
+  if (!ptr) {
+    return true;
+  }
+  auto* buf = static_cast<MTL::Buffer*>(ptr);
+  return buf->storageMode() != MTL::StorageModePrivate;
+}
+
 } // namespace allocator
 
 namespace metal {
--- a/mlx/ops.cpp
+++ b/mlx/ops.cpp
@@ -300,6 +300,16 @@ array copy(array a, StreamOrDevice s /* = {} */) {
       {std::move(a)});
 }
 
+array copy_to_new_buffer(array a, StreamOrDevice s /* = {} */) {
+  auto copied_shape = a.shape(); // |a| will be moved
+  auto dtype = a.dtype();
+  return array(
+      std::move(copied_shape),
+      dtype,
+      std::make_shared<AsType>(to_stream(s), dtype),
+      {std::move(a)});
+}
+
 array full_impl(array vals, Dtype dtype, StreamOrDevice s /* = {} */) {
   return array(
       vals.shape(),
--- a/mlx/ops.cpp
+++ b/mlx/ops.cpp
@@ -261,13 +261,16 @@ array linspace(
       s);
 }
 
-array astype(array a, Dtype dtype, StreamOrDevice s /* = {} */) {
-  if (dtype == a.dtype()) {
+array astype(
+    array a,
+    Dtype dtype,
+    std::optional<bool> copy,
+    StreamOrDevice s /* = {} */) {
+  if (dtype == a.dtype() && !copy.value_or(false)) {
     return a;
   }
-  auto copied_shape = a.shape(); // |a| will be moved
   return array(
-      std::move(copied_shape),
+      a.shape(),
       dtype,
       std::make_shared<AsType>(to_stream(s), dtype),
       {std::move(a)});
@@ -297,16 +300,6 @@ array copy(array a, StreamOrDevice s /* = {} */) {
       std::move(copied_shape),
       dtype,
       std::make_shared<Copy>(to_stream(s)),
-      {std::move(a)});
-}
-
-array copy_to_new_buffer(array a, StreamOrDevice s /* = {} */) {
-  auto copied_shape = a.shape(); // |a| will be moved
-  auto dtype = a.dtype();
-  return array(
-      std::move(copied_shape),
-      dtype,
-      std::make_shared<AsType>(to_stream(s), dtype),
       {std::move(a)});
 }
 
--- a/mlx/ops.cpp
+++ b/mlx/ops.cpp
@@ -300,20 +300,6 @@ array copy(array a, StreamOrDevice s /* = {} */) {
       {std::move(a)});
 }
 
-array copy_into(
-    const array& dst,
-    const array& src,
-    StreamOrDevice s /* = {} */) {
-  auto stream = to_stream(s);
-  auto update =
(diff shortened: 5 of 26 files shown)


megacpp:
Hi @XXXXRT666 — read through this PR after @awni redirected us here from #3548. The `nb::ndarray<nb::ro, nb::c_contig>` approach over the in-flight nanobind PR (#1338) is materially cleaner than the manual capsule parsing we had in our downstream PoC, and lifting `is_host_accessible()` into `mlx/allocator.h` is the right level of abstraction. Closed the RFC; happy with this being the path forward for #2848.

Wanted to offer some testing help that complements the PyTorch MPS bring-up you have:

We maintain a downstream TileLang fork (https://github.com/DatasunriseOU/tilelang) whose TVM-FFI bridge exports `kDLMetal` DLPack capsules for tensors backed by `id<MTLBuffer>`. That gives a **non-PyTorch** Metal DLPack producer that exercises the same import path you're adding here. Specifically it covers:

- `kDLMetal` producers that do **not** require the PyTorch-MPS workaround for `__dlpack__` — exercises the import path directly.
- Round-trip `mx.array → DLPack → TVM-FFI Metal kernel → DLPack → mx.array` zero-copy.
- Custom Metal kernels (via `mlx.fast.metal_kernel`) consuming an imported `mx.array` whose underlying `MTLBuffer` was allocated outside MLX.
- `storageMode` matrix (we hit `Shared` and `Managed`; `Private` is the obvious edge case the spec needs to nail down — your `is_host_accessible()` decision likely answers this implicitly but worth a sanity check from the producer side).
- `byte_offset != 0` cases that we'd previously rejected outright in our PoC — your PR seems to handle these via `byte_offset`-aware import; happy to write a TileLang-side test for it.

If useful, once the PR converges I can:

1. Pull this branch into our TileLang test matrix and report back on any rough edges (CI on macOS Metal hardware).
2. Send a minimal standalone repro (no TileLang dependency) for any of the above scenarios if you'd like them as additions to `python/tests/test_array.py`.
3. Beta-test the `mx.from_dlpack(..., copy=...)` semantics against the dtype-mismatch-shares case (`002360faa`) once the API stabilizes.

Tag me here when you'd like input — no rush, just don't want this to slip past once it's review-ready.

(For the orthogonal `mx.empty()` piece that was also in our PoC, opened it as a separate issue per @awni's guidance.)

McPatate:
This would be super cool if it landed for end to end "0-copy" support in safetensors! I'm working (https://github.com/safetensors/safetensors/pull/767) on adding reading bytes from disk in raw `MTLBuffer`s, which can then be handed to the framework via dlpack with 0-copy. Works well with torch, would be happy to see that land in mlx!

Also, support for `byte_offset !=0` would be nice (already in the PR but commenting to notify it's useful) since we can go one step further: currently the mps path is `pread` -> `MTLBuffer`, but that goes through kernel pages before hitting userspace buffer. Having `byte_offset` non zero support would enable mmap-ing the file and creating `MTLBuffer`s that reference specific slices of the mmap, which would demand-fault pages from disk into the page cache on first access and give userspace access directly, leaving only the disk -> kernel-side copy.

Quick question on the `dl_tensor.data` convention, torch's mps treats it as `id<MTLBuffer>`, passing the contents segfaults. Curious to know which direction MLX will be taking, as it impacts us downstream!

XXXXRT666:
> Quick question on the `dl_tensor.data` convention, torch's mps treats it as `id<MTLBuffer>`, passing the contents segfaults. Curious to know which direction MLX will be taking, as it impacts us downstream!

https://dmlc.github.io/dlpack/latest/c_api.html#c.DLTensor.data

The data pointer points to the allocated data. This will be CUDA device pointer, `cl_mem handle` in OpenCL, or `id<MTLBuffer>` for Metal.

docs/src/usage/numpy.rst
```diff
@@ -84,7 +86,73 @@ PyTorch supports DLPack inputs and can import MLX arrays directly.
...
+
+.. code-block:: python
+
+  b = torch.arange(3, device="mps", dtype=torch.float32)
+  torch.mps.synchronize()
+  c = mx.array(b)
+
+  c += 10
+  mx.eval(c)
+  print(b.cpu()) # tensor([10., 11., 12.])
```
zcbenz: Are you sure about this? I think in MLX `c += 10` would replace the underlying buffer of `c` rather than doing an inplace update, so the buffer of `b` should not be changed.

mlx/backend/cuda/allocator.cpp
```diff
@@ -416,6 +416,10 @@ void* Buffer::raw_ptr() {
   return cbuf.data;
 }
 
+bool Buffer::is_host_accessible() const {
+  return true;
```
zcbenz, line 402:
```suggestion
  if (!ptr_) {
    return true;
  }
  auto& cbuf = *static_cast<cu::CudaBuffer*>(ptr_);
  return cbuf.device == -1;
```

mlx/backend/metal/allocator.cpp
```diff
@@ -24,9 +24,23 @@ void* Buffer::raw_ptr() {
...
+  }
   return static_cast<MTL::Buffer*>(ptr_)->contents();
 }
 
+bool Buffer::is_host_accessible() const {
+  if (!ptr_) {
+    return true;
+  }
+  auto* buf = static_cast<MTL::Buffer*>(ptr_);
+  return buf->storageMode() != MTL::StorageModePrivate;
```
zcbenz, line 38: I don't have experience with the mps backend of PyTorch, but it seems that they do not really create buffers under private mode?
https://github.com/pytorch/pytorch/blob/e0fb301e042ad4bb791a6bae24a7a773a6b9de2b/aten/src/ATen/mps/MPSAllocator.mm#L41-L54

Unless it is quite common for PyTorch tensors to use private storage I think we should just throw when users try to use one.

CMakeLists.txt
```diff
@@ -357,8 +357,7 @@ if(MLX_BUILD_PYTHON_BINDINGS)
   FetchContent_Declare(
     nanobind
     GIT_REPOSITORY https://github.com/wjakob/nanobind.git
-    GIT_TAG v2.12.0
-    GIT_SHALLOW TRUE
```
zcbenz, line 381: `GIT_SHALLOW TRUE` should be kept.

python/src/convert.cpp
```diff
@@ -42,88 +57,170 @@ mx::array nd_array_to_mlx_contiguous(
...
-    shape.push_back(check_shape_dim(nd_array.shape(i)));
-  }
-  auto type = nb_dtype.value_or(nd_array.dtype());
-
-  // Copy data and make array
+template <typename F>
+auto dispatch_dlpack_dtype(
+    nb::dlpack::dtype type,
+    F&& f,
+    const char* error_message) {
```
zcbenz, line 98: Not a big issue but we always put `f` at last:

```suggestion
    const char* error_message,
    F&& f) {
```

docs/src/usage/numpy.rst
XXXXRT666: `__iadd__` does replace the MLX descriptor, but the binary op can donate/reuse the input buffer before that replacement happens. For this imported Metal DLPack case, `c` is donatable, so `set_binary_op_output_data` uses `copy_shared_buffer(c)` and the Metal kernel writes into the shared `MTLBuffer`. I verified that PyTorch observes the update after `mx.eval(c)`.

Image

mlx/backend/metal/allocator.cpp
XXXXRT666: I kept the check because PyTorch MPS tensors were not always shared historically.

Before pytorch/pytorch#175818, the default `GetMPSAllocator()` path used `useSharedAllocator=false`, so ordinary MPS tensors could come from the private allocator. PyTorch 2.12 changed this by removing the private pools and making `getIMPSAllocator()` return the shared allocator, and the release notes now say all MPS tensors are allocated in unified memory.

So this check keeps MLX correct for older PyTorch-produced `kDLMetal` tensors: importing the `id<MTLBuffer>` is fine, but host access through `raw_ptr()` should throw if that PyTorch buffer is private.

XXXXRT666: Image

Image

docs/src/usage/numpy.rst
zcbenz: Thanks for testing this, it makes sense. I think this behavior would be surprising to users because MLX does not do in-place updates, i.e. if `b` is a MLX array you wouldn't expect `b` to change after `c += 10`. This is only happening because `b` essentially referenced `c`'s buffer but did not add its reference count, which makes `c` donate-able but it really shouldn't. And this is also bad because the behavior would depend on the donate-able state of `c`, if you add `d = c` then the PyTorch tensor `b` would not be updated.

The best fix would be storing a reference to MLX array in the PyTorch tensor, but I don't know if that's possible. Or we can probably add a flag to prevent the MLX array from being donate-able.

mlx/backend/metal/allocator.cpp
zcbenz: Thanks for checking PyTorch. We actually do something similar in CUDA backend that, buffers are per-device by default and we would automatically move them to unified memory when they are accessed by CPU, to implement the same auto-migration in Metal backend we would need some large refactoring in Metal code and the implementation in this PR could easily break when the arrays were passed to CPU kernels.

But the problem is it would really complicate the code a lot, and the tech debt would only benefit users of old PyTorch versions. What do you think if we copy the buffers instead when they use private storage, so we can keep the code simple while making things work with old versions of PyTorch?

docs/src/usage/numpy.rst
XXXXRT666: Agreed. I removed the example that relied on regular MLX update expressions.

The docs now explicitly say not to rely on that behavior for sharing/synchronization, because MLX may rebind the array to a new buffer. The example uses `array.copy(...)` instead when the update needs to be visible through the shared buffer.

If you think we should also prevent donation for these imported buffers with a flag, I can add that too.

CMakeLists.txt
XXXXRT666: I removed `GIT_SHALLOW TRUE` because nanobind is pinned to a specific commit. In a clean build without a local nanobind override, the shallow FetchContent clone failed to checkout that commit with `fatal: unable to read tree (...)`. Removing the shallow clone makes the pinned commit available and the build succeeds.

mlx/backend/metal/allocator.cpp
XXXXRT666: I updated the Metal DLPack import path to copy private Metal buffers instead of wrapping them directly.

So shared Metal buffers still use zero-copy import, while private buffers are copied into MLX-managed storage. `copy=False` now raises if sharing would require such a copy.

I also updated the docs to mention that PyTorch 2.12+ uses shared storage for ordinary MPS tensors on Apple silicon, while older PyTorch versions may use private storage and therefore be copied on import.

docs/src/usage/numpy.rst
```diff
@@ -76,15 +76,91 @@ PyTorch
...
+
+  a = mx.arange(3, dtype=mx.float32)
+  mx.eval(a)
+  b = torch.as_tensor(a)
+
+  b.add_(10)
+  torch.mps.synchronize()
+  print(a.tolist()) # [10.0, 11.0, 12.0]
+
+  a.copy(mx.array([4, 5, 6], dtype=mx.float32))
```
zcbenz: I'm not sure about adding a `array.copy` op, I think the mental model would be simple if users just assume MLX array being immutable, is there a reason for this?

python/src/convert.cpp
```diff
@@ -42,72 +57,163 @@ mx::array nd_array_to_mlx_contiguous(
...
+  }
+
+  auto byte_offset = owner->byte_offset();
+  if (byte_offset % itemsize != 0) {
+    throw std::invalid_argument(
+        "Metal DLPack byte offset is not aligned to dtype size.");
   }
+
+  auto out = mx::array(
+      mx::allocator::Buffer(owner->data_handle()),
```
zcbenz: We should check whether the alien buffer uses private storage, and if so create a new buffer and do the copy here, this way we can simply ensure that all Metal buffers in MLX are using unified memory.

It is likely going to require a lot more work, but we can then avoid introducing the whole `is_host_accessible` concept in Metal backend and isolate the dirty work in one function.

python/src/convert.cpp
```diff
@@ -42,72 +57,163 @@ mx::array nd_array_to_mlx_contiguous(
...
+    case nb::device::cuda_managed::value:
+      throw std::invalid_argument("CUDA DLPack import is not supported.");
+    default:
+      throw std::invalid_argument("Unsupported DLPack device.");
+  }
+}
+
+template <typename T>
+mx::array metal_dlpack_to_mlx_contiguous(
+    std::shared_ptr<nb::ndarray<nb::ro, nb::c_contig>> owner,
```
zcbenz: I'm not sure if we need to store it in a shared_ptr, copying by value should be able to reserve the reference.

python/src/convert.cpp
```diff
@@ -168,8 +275,91 @@ nb::ndarray<nb::numpy> mlx_to_np_array(const mx::array& a) {
...
-nb::ndarray<> mlx_to_dlpack(const mx::array& a) {
-  return mlx_to_nd_array<>(a);
+template <typename T>
+nb::ndarray<> mlx_to_dlpack_impl(mx::array a, int dl_device_type) {
+  void* data = nullptr;
+  uint64_t byte_offset = 0;
+  {
+    nb::gil_scoped_release nogil;
+    a.eval();
+    a.wait();
```
zcbenz: `array.eval()` should have already implied `wait()`.

python/src/convert.cpp
```diff
@@ -168,8 +275,91 @@ nb::ndarray<nb::numpy> mlx_to_np_array(const mx::array& a) {
...
+    if (dl_device_type == nb::device::cpu::value) {
+      data = a.data<T>();
```
zcbenz: It should be fine to use same code for cpu/gpu devices? i.e. just `data = a.buffer().ptr()`.

python/src/convert.cpp
```diff
@@ -168,8 +275,91 @@ nb::ndarray<nb::numpy> mlx_to_np_array(const mx::array& a) {
...
+
+  std::vector<size_t> shape(a.shape().begin(), a.shape().end());
+  auto owner = nb::cast(a);
+  return nb::ndarray<>(
+      data,
+      a.ndim(),
+      shape.data(),
+      /* owner= */ owner,
+      a.strides().data(),
+      nb::dtype<T>(),
```
zcbenz: It would be simpler writing a `mlx_dtype_to_dl_dtype` utility and then pass a `dlpack::dtype`, to avoid the whole type dispatch.

python/src/convert.cpp
```diff
@@ -42,72 +57,163 @@ mx::array nd_array_to_mlx_contiguous(
...
+  }
+}
+
+mx::array from_dlpack(nb::object v, std::optional<bool> copy) {
+  using ContigArray = nb::ndarray<nb::ro, nb::c_contig>;
+  auto nd = nb::cast<ContigArray>(v);
+
+  switch (nd.device_type()) {
+    case nb::device::cpu::value:
+      if (copy == false) {
```
zcbenz: This is probably just checking if `copy` has a value in it.
```suggestion
      if (copy.has_value() && copy.value() == false) {
```

python/src/convert.cpp
```diff
@@ -42,72 +57,163 @@ mx::array nd_array_to_mlx_contiguous(
...
+          });
+    }
+    case nb::device::metal::value:
+      return metal_dlpack_to_mlx(std::move(nd_array), dtype, std::nullopt);
+    default:
+      throw std::invalid_argument("Unsupported DLPack device.");
+  }
+}
+
+mx::array from_dlpack(nb::object v, std::optional<bool> copy) {
```
zcbenz: `nd_array` and `dlpack` basically mean the same thing and having both `nd_array_to_mlx` and `from_dlpack` is confusing. Maybe moving the `copy` parameter to `nd_array_to_mlx` and have the python `from_dlpack` op just call it?

python/src/convert.cpp
```diff
@@ -42,72 +57,163 @@ mx::array nd_array_to_mlx_contiguous(
...
+
+mx::array metal_dlpack_to_mlx(
+    nb::ndarray<nb::ro, nb::c_contig> nd_array,
+    std::optional<mx::Dtype> dtype,
+    std::optional<bool> copy);
+
+mx::array nd_array_to_mlx(
+    nb::ndarray<nb::ro, nb::c_contig> nd_array,
+    std::optional<mx::Dtype> dtype,
+    std::optional<nb::dlpack::dtype> nb_dtype) {
```
zcbenz: Might be a good change giving them better names.

```suggestion
    std::optional<mx::Dtype> dst_dtype,
    std::optional<nb::dlpack::dtype> src_dtype) {
```

python/src/convert.cpp
XXXXRT666: For CPU DLPack export, `a.buffer().ptr()` is the allocator-owned buffer handle, not the host data pointer. The CPU allocator stores its size header before the actual data, while `a.data<T>()` goes through `raw_ptr()` and returns the correct host data address

For Metal, `a.buffer().ptr()` is the expected `id<MTLBuffer>` handle. So the CPU and Metal branches intentionally use different pointers here

docs/src/usage/numpy.rst
XXXXRT666: I removed `array.copy` and the `CopyInto` primitive. The docs now keep this PR focused on zero-copy DLPack import/export and avoid introducing a new mutating array API.

python/src/convert.cpp
XXXXRT666: Private Metal DLPack buffers are now handled in the import path

python/src/convert.cpp
XXXXRT666: The nanobind ndarray owner is now kept by value

python/src/convert.cpp
zcbenz: That makes sense, thanks for clarification. In this case I think we can use `data<void>()` to avoid the type dispatching.

python/src/convert.cpp
XXXXRT666: The redundant wait is removed now

python/src/convert.cpp
XXXXRT666: DLPack export now maps MLX dtype with `mlx_dtype_to_dl_dtype` and passes the resulting `dlpack::dtype` directly.

python/src/convert.cpp
XXXXRT666: `from_dlpack` now just casts once to the nanobind ndarray and calls `nd_array_to_mlx(copy=...)`; the device and copy handling lives in one path.

XXXXRT666:
One API question: should `mx.array(...)` always copy DLPack inputs, and should zero-copy / copy control live in `mx.asarray(..., copy=...)` instead?

That would match the mental model used by NumPy/PyTorch more closely: `array` creates a new array, while `asarray` may avoid a copy depending on `copy`. In that design, `mx.from_dlpack(..., copy=...)` could remain the explicit DLPack entry point, while `mx.array(torch_mps_tensor)` would not unexpectedly share the underlying Metal buffer by default.

python/src/array.cpp
```diff
@@ -496,7 +496,32 @@ void init_array(nb::module_& m) {
...
             }
           })
-      .def("__dlpack__", [](const mx::array& a) { return mlx_to_dlpack(a); })
+      .def(
+          "__dlpack__",
+          [](const mx::array& a,
+             nb::object,
+             nb::object,
+             nb::object dl_device,
+             nb::object copy) {
```
zcbenz: nanobind should be able to auto-convert the parameters:
```suggestion
             std::optional<std::tuple<int, int>> dl_device,
             std::optional<bool> copy) {
```

python/src/buffer.h
```diff
@@ -91,8 +91,8 @@ extern "C" inline int getbuffer(PyObject* obj, Py_buffer* view, int flags) {
   {
     nb::gil_scoped_release nogil;
     a.eval();
+    a.wait();
```
zcbenz: `wait()` is not needed.

python/src/convert.h
```diff
@@ -64,16 +63,20 @@ struct ArrayLike {
...
     std::optional<mx::Dtype> mx_dtype,
-    std::optional<nb::dlpack::dtype> nb_dtype = std::nullopt);
+    std::optional<nb::dlpack::dtype> nb_dtype = std::nullopt,
+    std::optional<bool> copy = std::nullopt);
 
 nb::ndarray<nb::numpy> mlx_to_np_array(const mx::array& a);
-nb::ndarray<> mlx_to_dlpack(const mx::array& a);
+nb::ndarray<> mlx_to_dlpack(
+    const mx::array& a,
+    std::optional<int> dl_device_type = std::nullopt);
```
zcbenz: When implementing this for CUDA in future we will need the device ID.
```suggestion
    std::optional<std::tuple<int, int>> dl_device_type = std::nullopt);
```

python/src/ops.cpp
```diff
@@ -1783,6 +1783,29 @@ void init_ops(nb::module_& m) {
         Returns:
             array: An array interpretation of the input.
       )pbdoc");
+  m.def(
+      "from_dlpack",
+      [](const nb::object& x, std::optional<bool> copy) {
+        return from_dlpack(x, copy);
+      },
```
zcbenz: It should be possible to just do this in nanobind:

```suggestion
      [](nb::ndarray<nb::ro, nb::c_contig> x, std::optional<bool> copy) {
        return nd_array_to_mlx(x, std::nullopt, std::nullopt, copy);
      },
```

`std::move` is also not needed when copying the object does not involve heap allocations.

python/src/convert.cpp
```diff
@@ -168,8 +320,59 @@ nb::ndarray<nb::numpy> mlx_to_np_array(const mx::array& a) {
   return mlx_to_nd_array<nb::numpy>(a);
 }
 
-nb::ndarray<> mlx_to_dlpack(const mx::array& a) {
-  return mlx_to_nd_array<>(a);
+nb::ndarray<>
+mlx_to_dlpack_impl(mx::array a, int dl_device_type, nb::dlpack::dtype dtype) {
```
zcbenz: The `mlx_to_dlpack_impl` is not used anywhere else, I think it should be just merged in `mlx_to_dlpack`?

python/src/convert.cpp
```diff
@@ -215,6 +418,25 @@ nb::object to_scalar(mx::array& a) {
...
 }
 
+mx::array metal_dlpack_to_mlx(
+    nb::ndarray<nb::ro, nb::c_contig> nd_array,
+    std::optional<mx::Dtype> dtype,
+    std::optional<bool> copy) {
+  if (!mx::metal::is_available()) {
+    throw std::invalid_argument("Metal DLPack import is not available.");
+  }
+  auto owner = std::move(nd_array);
```
zcbenz: This is not needed.

python/src/convert.cpp
```diff
@@ -511,7 +733,13 @@ mx::array create_array(nb::object v, std::optional<mx::Dtype> t) {
     } else {
       nd = nb::cast<ContigArray>(v);
     }
-    return nd_array_to_mlx(nd, t, nb_dtype);
+    auto type = nb_dtype.value_or(nd.dtype());
+    std::optional<mx::Dtype> copy_dtype;
+    if (t &&
+        *t != mlx_dtype_from_dlpack(type, "Cannot convert array to mlx.")) {
+      copy_dtype = t;
```
zcbenz: Does this mean to force the types of dlpack and mlx arrays to be the same? I think the implementation means to support type conversions?

python/src/convert.cpp
```diff
@@ -42,74 +75,192 @@ mx::array nd_array_to_mlx_contiguous(
...
+      return nb::dtype<std::complex<float>>();
+    default:
+      throw nb::buffer_error("Cannot export mlx array with unsupported dtype.");
+  }
+}
+
+mx::array metal_dlpack_to_mlx(
+    nb::ndarray<nb::ro, nb::c_contig> nd_array,
+    std::optional<mx::Dtype> dtype,
+    std::optional<bool> copy);
```
zcbenz: This forward declaration seems unnecessary since we can just move the implementation here?

python/src/convert.cpp
```diff
@@ -17,6 +25,21 @@ enum PyScalarT {
   pycomplex = 3,
 };
 
+namespace {
+
+bool metal_buffer_is_private(void* ptr) {
```
zcbenz: We don't put platform-specific code here. It is usually done by putting something like `bool can_reuse_alien_buffer(void* ptr)` in `mlx/allocator.h` and then implement it in `mlx/backend/metal/allocator.cpp`.

mlx/ops.cpp
```diff
@@ -294,6 +294,16 @@ array copy(array a, StreamOrDevice s /* = {} */) {
       {std::move(a)});
 }
 
+array copy_to_new_buffer(array a, StreamOrDevice s /* = {} */) {
```
zcbenz: Instead of adding a new op, I think we should just add a new parameter to `astype`.

```c++
MLX_API array astype(array a, Dtype dtype, std::optional<bool> copy, StreamOrDevice s = {});
inline array astype(array a, Dtype dtype,StreamOrDevice s = {}) {
  return as type(a, dtype, std::nullopt, s);
}
```

mlx/ops.cpp
```diff
@@ -294,6 +294,16 @@ array copy(array a, StreamOrDevice s /* = {} */) {
...
+  auto copied_shape = a.shape(); // |a| will be moved
+  auto dtype = a.dtype();
+  return array(
+      std::move(copied_shape),
```
zcbenz: Also there is no need to use `std::move` since shapes are no longer heap allocated, we don't migrate the old code but try to use the new pattern in new code.

python/src/mlx_func.cpp
```diff
@@ -2,6 +2,8 @@
 
 #include "python/src/mlx_func.h"
 
+#include <structmember.h>
```
zcbenz, line 5: Is this needed?

python/src/convert.cpp
```diff
@@ -42,74 +75,192 @@ mx::array nd_array_to_mlx_contiguous(
...
+  out.set_data(
+      out.buffer(),
+      out.data_size(),
+      out.strides(),
+      flags,
+      [owner = std::move(owner)](mx::allocator::Buffer) {});
+
+  auto offset = static_cast<int64_t>(byte_offset / itemsize);
+  if (offset != 0) {
+    out.copy_shared_buffer(out, out.strides(), flags, out.data_size(), offset);
```
zcbenz: I think we should just add `offset` parameter to `array::set_data`.

zcbenz:
> One API question: should `mx.array(...)` always copy DLPack inputs, and should zero-copy / copy control live in `mx.asarray(..., copy=...)` instead?
> 
> That would match the mental model used by NumPy/PyTorch more closely: `array` creates a new array, while `asarray` may avoid a copy depending on `copy`. In that design, `mx.from_dlpack(..., copy=...)` could remain the explicit DLPack entry point, while `mx.array(torch_mps_tensor)` would not unexpectedly share the underlying Metal buffer by default.

I think this is very good design.

python/src/mlx_func.cpp
XXXXRT666: image

I think `gc_func_members` needs it

XXXXRT666:
#### mlx 0.31.2

```text
mlx_to_torch  (median µs / call, lower is better)
  shape               current → mps   dlpack → mps   dlpack stay-on-cpu
  16K  f32                      378            191                   16
  1M   f32                      404            402                   17
  16M  f32 (64MB)              2717           2501                   16
  1M  bf16                      276 n/a (TypeError)      n/a (TypeError)
  16M bf16 (32MB)              1302 n/a (TypeError)      n/a (TypeError)

torch_to_mlx  (median µs / call, lower is better)
  shape                current (.cpu+numpy)      dlpack (from MPS)
  16K  f32                              175                    172
  1M   f32                              691                    666
  16M  f32 (64MB)                      8216                   8470
  1M  bf16                              522                    422
  16M bf16 (32MB)                      2023                   2037
```

#### mlx 0.32.0.dev20260523+4e8decde9

```text
mlx_to_torch  (median µs / call, lower is better)
  shape               current → mps   dlpack → mps   dlpack stay-on-cpu
  16K  f32                      237             18                   15
  1M   f32                      371             16                   17
  16M  f32 (64MB)              2688             15                   15
  1M  bf16                      279             19                   16
  16M bf16 (32MB)              1340             16                   14

torch_to_mlx  (median µs / call, lower is better)
  shape                current (.cpu+numpy)      dlpack (from MPS)
  16K  f32                              189                     16
  1M   f32                              676                     16
  16M  f32 (64MB)                      8806                     17
  1M  bf16                              484                     16
  16M bf16 (32MB)                      2106                     16
```


<details>

<summary>benchmark function</summary>

```python
# --- mlx -> torch candidates ---------------------------------------------------


def mlx_to_torch_current(arr: mx.array, device: torch.device) -> torch.Tensor:
    arr = mx.contiguous(arr)
    mx.eval(arr)
    buf = memoryview(arr)
    dtype_map = {
        mx.float32: torch.float32,
        mx.float16: torch.float16,
        mx.bfloat16: torch.bfloat16,
    }
    t = torch.frombuffer(buf, dtype=dtype_map[arr.dtype]).reshape(arr.shape)
    if device.type == "mps":
        t = t.to(device)
    return t


def mlx_to_torch_dlpack_mps(arr: mx.array, device: torch.device) -> torch.Tensor:
    mx.eval(arr)
    t = torch.from_dlpack(arr)
    if device.type == "mps":
        t = t.to(device)
    return t


def mlx_to_torch_dlpack_cpu(arr: mx.array, device: torch.device) -> torch.Tensor:
    """Force a CPU-typed capsule via `dl_device=(kDLCPU, 0)` (Phase 2+).
    Falls back to the no-kwarg form for builds that don't accept it."""
    mx.eval(arr)
    try:
        cap = arr.__dlpack__(dl_device=(1, 0))
    except TypeError:
        # Older builds: zero-arg lambda. Capsule is already kDLCPU there.
        cap = arr.__dlpack__()
    return torch.from_dlpack(cap)


# --- torch -> mlx candidates ---------------------------------------------------


def torch_to_mlx_current(t: torch.Tensor) -> mx.array:
    if t.device.type != "cpu":
        t = t.cpu()
    t = t.detach()
    if t.dtype == torch.bfloat16:
        return mx.array(t)
    return mx.array(t.numpy())


def torch_to_mlx_dlpack(t: torch.Tensor) -> mx.array:
    """Use mx.from_dlpack when the API exists; old MLX falls back to CPU copy."""
    if hasattr(mx, "from_dlpack"):
        return mx.from_dlpack(t)
    return mx.array(t.detach().cpu())
```
</details>

python/src/buffer.h
XXXXRT666: Done

python/src/array.cpp
XXXXRT666: Done

python/src/convert.h
XXXXRT666: Done

python/src/ops.cpp
XXXXRT666: `std::move` is removed now

python/src/convert.cpp
XXXXRT666: Removed

mlx/ops.cpp
XXXXRT666: Removed now

python/src/convert.cpp
XXXXRT666: Merged into one function now

python/src/convert.cpp
XXXXRT666: I changed the `copy_dtype ` to `dst_dtype`, I think it would be a better name

python/src/convert.cpp
XXXXRT666: Done, the forward declaration is no longer needed after moving the implementation earlier

python/src/convert.cpp
XXXXRT666: Done, the check is now behind `allocator::can_reuse_alien_buffer`

mlx/ops.cpp
XXXXRT666: Done, the copy control is moved to `astype` instead.

python/src/convert.cpp
XXXXRT666: Done, `array::set_data` now takes an offset

XXXXRT666:
> I think this is very good design.

I updated the PR to follow this design: `mx.array(...)` now copies DLPack inputs, while `mx.asarray(..., copy=...)` and `mx.from_dlpack(..., copy=...)` provide the copy-control paths.

mlx/backend/metal/allocator.cpp
```diff
@@ -24,7 +24,22 @@ void* Buffer::raw_ptr() {
   if (!ptr_) {
     return nullptr;
   }
-  return static_cast<MTL::Buffer*>(ptr_)->contents();
+  auto* buf = static_cast<MTL::Buffer*>(ptr_);
+  if (buf->storageMode() == MTL::StorageModePrivate) {
```
zcbenz: I think we should just do `assert(buf->storageMode() != MTL::StorageModePrivate)` since it is a path that should never happen.

mlx/utils.cpp
```diff
@@ -7,6 +7,7 @@
 #include <vector>
 
 #include "mlx/dtype_utils.h"
+#include "mlx/ops.h"
```
zcbenz: Not needed?

python/src/buffer.h
```diff
@@ -92,7 +92,6 @@ extern "C" inline int getbuffer(PyObject* obj, Py_buffer* view, int flags) {
     nb::gil_scoped_release nogil;
     a.eval();
   }
-
```
zcbenz, line 95: Not needed?

mlx/backend/cuda/allocator.cpp
```diff
@@ -416,6 +416,14 @@ void* Buffer::raw_ptr() {
   return cbuf.data;
 }
 
+bool can_reuse_alien_buffer(void* ptr) {
+  if (!ptr) {
+    return true;
+  }
+  auto& cbuf = *static_cast<cu::CudaBuffer*>(ptr);
```
zcbenz: In CUDA we probably need to check with `cudaPointerGetAttributes`, for now I think should just return true.

mlx/array.cpp
```diff
@@ -180,9 +181,10 @@ void array::set_data(
     size_t data_size,
     Strides strides,
     Flags flags,
+    int64_t offset,
     Deleter d) {
   array_desc_->data = std::make_shared<Data>(buffer, d);
-  array_desc_->offset = 0;
+  array_desc_->offset = itemsize() * offset;
```
zcbenz: I think we should just make `offset` being byte offset so you can just pass `owner.byte_offset()` to it without checking alignment.

python/src/convert.cpp
```diff
@@ -42,71 +58,195 @@ mx::array nd_array_to_mlx_contiguous(
...
+      [](mx::allocator::Buffer) {});
+  auto flags = out.flags();
+  auto offset = static_cast<int64_t>(byte_offset / itemsize);
+  out.set_data(
+      out.buffer(),
+      out.data_size(),
+      out.strides(),
+      flags,
+      offset,
+      [owner = std::move(owner)](mx::allocator::Buffer) {});
```
zcbenz:
```suggestion
  array out(shape, type, nullptr, {});
  out.set_data(
      mx::allocator::Buffer(owner.data_handle()),
      out.data_size(),
      out.strides(),
      out.flags(),
      offset,
      [owner = std::move(owner)](mx::allocator::Buffer) {});
```

python/src/convert.cpp
```diff
@@ -42,71 +58,195 @@ mx::array nd_array_to_mlx_contiguous(
...
+      throw nb::buffer_error("Cannot export mlx array with unsupported dtype.");
+  }
+}
+
+template <typename T>
+mx::array metal_dlpack_to_mlx_contiguous(
+    nb::ndarray<nb::ro, nb::c_contig> owner,
+    const mx::Shape& shape,
+    mx::Dtype type,
+    std::optional<mx::Dtype> dtype,
```
zcbenz:
```suggestion
    mx::Dtype src_dtype,
    std::optional<mx::Dtype> dst_dtype,
```

python/src/convert.cpp
```diff
@@ -42,71 +58,195 @@ mx::array nd_array_to_mlx_contiguous(
...
+  if (copy.has_value() && copy.value() == false && !can_reuse_buffer) {
+    throw std::invalid_argument(
+        "Cannot import a private Metal DLPack buffer without a copy.");
+  }
+
+  if (dtype || !can_reuse_buffer) {
+    auto result_dtype = dtype.value_or(out.dtype());
+    auto result = mx::astype(out, result_dtype, true, mx::Device::gpu);
+    result.eval();
+    result.detach();
```
zcbenz: `detach` should not be needed. 

And I'm hesitated whether we should call `eval` here, I think it would not harm at all if we don't?

python/src/convert.cpp
```diff
@@ -42,71 +58,195 @@ mx::array nd_array_to_mlx_contiguous(
...
+    case mx::float64:
+      return nb::dtype<double>();
+    case mx::complex64:
+      return nb::dtype<std::complex<float>>();
+    default:
+      throw nb::buffer_error("Cannot export mlx array with unsupported dtype.");
+  }
+}
+
+template <typename T>
```
zcbenz: The type parameter is no longer used? Can merge this function to `metal_dlpack_to_mlx`.

python/src/convert.cpp
```diff
@@ -168,8 +309,55 @@ nb::ndarray<nb::numpy> mlx_to_np_array(const mx::array& a) {
...
+  }
+
+  auto arr = a;
+  void* data = nullptr;
+  uint64_t byte_offset = 0;
+  {
+    nb::gil_scoped_release nogil;
+    arr.eval();
+    if (device_type == nb::device::cpu::value) {
+      data = static_cast<char*>(arr.buffer().raw_ptr()) + arr.offset();
```
zcbenz: Is it possible to just pass offset?
```suggestion
      data = arr.buffer().raw_ptr();
      byte_offset = arr.offset();
```

python/src/convert.cpp
```diff
@@ -511,9 +720,16 @@ mx::array create_array(nb::object v, std::optional<mx::Dtype> t) {
     } else {
       nd = nb::cast<ContigArray>(v);
     }
-    return nd_array_to_mlx(nd, t, nb_dtype);
+    auto type = nb_dtype.value_or(nd.dtype());
+    std::optional<mx::Dtype> dst_dtype;
+    if (t &&
+        *t != mlx_dtype_from_dlpack(type, "Cannot convert array to mlx.")) {
+      dst_dtype = t;
```
zcbenz: This should be done inside `nd_array_to_mlx`.

python/src/convert.cpp
```diff
@@ -471,34 +659,55 @@ mx::array array_from_list(nb::tuple pl, std::optional<mx::Dtype> dtype) {
...
+    check_copy_false(copy);
     return array_from_list(nb::cast<nb::list>(v), t);
   } else if (nb::isinstance<nb::tuple>(v)) {
+    check_copy_false(copy);
     return array_from_list(nb::cast<nb::tuple>(v), t);
   } else if (nb::isinstance<mx::array>(v)) {
     auto arr = nb::cast<mx::array>(v);
-    return mx::astype(arr, t.value_or(arr.dtype()));
+    auto dtype = t.value_or(arr.dtype());
+    if (copy.has_value() && copy.value() == false && dtype != arr.dtype()) {
```
zcbenz: When user uses `a = asarray(b, copy=False)` they expect `a` to be a semantically shallow-copy of `b` that, `a` would also be changed when you run `b += 1`. Which however is not true in MLX and we should just always throw when `b` is a MLX array.

zcbenz (commented):
This basically looks good to me, thanks for the nice work!

python/src/convert.cpp
XXXXRT666: I tried this construction locally, but it does not preserve the same invariants as the leaf buffer constructor.

`array(shape, dtype, nullptr, {})` uses the lazy-output constructor, so the descriptor starts as `unscheduled` and `data_size == 0`. We would need to manually set the status to `available`, and also avoid passing `out.data_size()` into `set_data()` because that preserves the zero `data_size`. In my local test, this hybrid state could get through `mx.eval(y + 1)`, but reading the result back with `tolist()` segfaulted.

The current constructor path goes through the leaf data-array constructor, which initializes the descriptor consistently before attaching the external buffer, so I kept that path.

zcbenz: That makes sense, it is not hard to calculate the data_size and flags though and we would be able to make it work for strided arrays, you can find an example at https://github.com/ml-explore/mlx/blob/2165dc08d7b33258260aa849d39f087d50e62962/mlx/backend/common/slicing.cpp#L20-L36.

XXXXRT666:
```cpp
mx::array out(shape, src_dtype, nullptr, {});
auto flags = out.flags();
flags.contiguous = true;
flags.row_contiguous = true;
flags.col_contiguous = out.size() <= 1 ||
    out.size() == *std::max_element(out.shape().begin(), out.shape().end());

out.set_data(
    mx::allocator::Buffer(data_handle),
    out.size(),
    out.strides(),
    flags,
    nd_array.byte_offset(),
    [owner = std::move(nd_array)](mx::allocator::Buffer) {});
out.set_status(mx::array::Status::available);
```

Based on my testing, this way might work, but I think it’s a bit more confusing than the previous one.

python/src/convert.cpp
XXXXRT666: Kept the `eval()` because this branch is the explicit copy path. Without evaluating here, the returned array would still be a lazy copy from the external DLPack buffer, so mutations to the producer before the first MLX eval could affect the result. Keeping `eval()` makes `copy=True` or private buffer fallback behave like a snapshot at import time.

mlx/utils.cpp
XXXXRT666: Removed.

python/src/buffer.h
XXXXRT666: Removed.

mlx/backend/cuda/allocator.cpp
XXXXRT666: The CUDA implementation always returns `true` for now.

mlx/backend/metal/allocator.cpp
XXXXRT666: Done, changed `raw_ptr()` to assert that the Metal buffer is not private.

mlx/array.cpp
XXXXRT666: `set_data` now treats the offset as a byte offset.

python/src/convert.cpp
XXXXRT666: Done, cleaned up the dtype naming and made `dst_dtype` the final output dtype.

python/src/convert.cpp
XXXXRT666: The Metal contiguous helper is merged into `metal_dlpack_to_mlx` now

python/src/convert.cpp
XXXXRT666: Yes, CPU DLPack export now passes the base data pointer and uses `byte_offset`

python/src/convert.cpp
XXXXRT666: Done, `copy=False` now raises for MLX array inputs.

python/src/convert.cpp
XXXXRT666: The source dtype handling now lives in `nd_array_to_mlx`

python/src/convert.cpp
XXXXRT666: I kept the current constructor since it preserves the descriptor invariants directly. If you prefer the `array(shape, dtype, nullptr, {}) + set_data(...)` approach, I’m happy to switch.

zcbenz: The constructor of array assumes the data to be contiguous, the data_size and flags would be wrong when the tensor is not. You can do something like:

```c++
   auto [data_size, is_row_contiguous, is_col_contiguous] = 
       check_contiguity(shape, strides); 
  
   Array::Flags flags;
   flags.row_contiguous = is_row_contiguous; 
   flags.col_contiguous = is_col_contiguous; 
   flags.contiguous = (data_size * itemsize == nd_array.nbytes()); 

   out.set_data(..., data_size, flags);
   out.wait();
```

XXXXRT666:
I just realized that `nanobind` does not handle `c_contig` arrays very well. For `torch.Tensor`, it calls `contiguous()`, but it does not perform `torch.accelerator.synchronization` afterwards.

python/src/convert.cpp
XXXXRT666: Yeah, I realized that, `nanobind` try to make it `c_contig`, but doesn't sync after calling, so with `copy=True
` we will get the wrong result

python/src/convert.cpp
```diff
@@ -31,82 +38,257 @@ int check_shape_dim(int64_t dim) {
...
+  }
+  auto strides = get_strides(nd_array);
+  auto data_size = strided_data_size(shape, strides);
+  auto [no_bsx_size, is_row_contiguous, is_col_contiguous] = shape.empty()
+      ? std::make_tuple(data_size, true, true)
+      : mx::check_contiguity(shape, strides);
+  auto data_handle = nd_array.data_handle();
+  mx::array out(shape, src_dtype, nullptr, {});
+  auto flags = out.flags();
+  flags.contiguous = no_bsx_size == data_size;
```
zcbenz: `nd_array` has a size property, can it be used instead?

```suggestion
  flags.contiguous = no_bsx_size == nd_array.size();
```

python/src/convert.cpp
```diff
@@ -31,82 +38,257 @@ int check_shape_dim(int64_t dim) {
...
+  if (!mx::metal::is_available()) {
+    throw std::invalid_argument("Metal DLPack import is not available.");
+  }
+  auto shape = get_shape(nd_array);
+  if (nd_array.itemsize() != mx::size_of(src_dtype)) {
+    throw std::invalid_argument(
+        "Cannot convert Metal DLPack dtype to mlx dtype.");
+  }
+  auto strides = get_strides(nd_array);
+  auto data_size = strided_data_size(shape, strides);
```
zcbenz: `data_size` is usually used for the actual buffer size, not the strided/broadcasted array size.

python/src/convert.cpp
```diff
@@ -31,82 +38,257 @@ int check_shape_dim(int64_t dim) {
...
+      mlx_dtype_to_dl_dtype(dst_dtype),
+      [&]<typename DstT>(mx::Dtype) {
+        auto out = mx::array(shape, dst_dtype, nullptr, {});
+        out.set_data(mx::allocator::malloc(out.nbytes()));
+        auto src = static_cast<const SrcT*>(nd_array.data());
+        auto dst = out.data<DstT>();
+        auto strides = get_strides(nd_array);
+        strided_data_size(shape, strides);
+        size_t dst_index = 0;
+        copy_strided_nd_array(src, dst, shape, strides, 0, 0, dst_index);
```
zcbenz:
```c++
   out.set_data(
      mx::allocator::malloc(data_size * itemsize),
      data_size,
      strides,
      flags);
   std::copy(src, src + data_size, out.data<DstT>());
```

python/src/convert.cpp
XXXXRT666: Well, no. `nd_array.size` is just the product of the shape. For example, `x = torch.tensor([2, 7]).expand(3, 2)` has strides `(0, 1)` and shape `[3, 2]`, so `nd_array.size` is 6, while the actual storage span is only 2. I do agree that the name `data_size` is confusing, though, so I’ll rename it to something more precise.

XXXXRT666: Another example is a sliced view such as `x[2:10:2]`. Its logical size is 4, since it contains the elements at indices 2, 4, 6, and 8. However, the storage span is 7: the view starts at offset 2 and ends at offset 8, so the span is `8 - 2 + 1 = 7`, not 4. In this case, both `no_bsx_size` and `nd_array.size()` return 4, which means they do not capture the actual storage span.

python/src/convert.cpp
XXXXRT666: I renamed this to `storage_size` to make the meaning clearer, it is the element span of the underlying storage covered by the strided view, not the logical array size.

python/src/convert.cpp
XXXXRT666: Done. I changed the CPU import path to copy the storage span with `std::copy`

python/src/convert.cpp
zcbenz: Sorry I have been misleading, so the condition of contiguous is:
https://github.com/ml-explore/mlx/blob/2165dc08d7b33258260aa849d39f087d50e62962/mlx/array.h#L249-L254

i.e. `contiguous = (no_bsx_size == data_size)`

The `data_size` is the number of elements in the underlying buffer, which dlpack does not actually give us.

I think the way you calculate `storage_size` is exactly how `check_contiguity` calculates `no_bsx_size`, i.e. `prod(shape[i] for i in range(ndim) if strides[i] > 0)`, which would only be correct when the tensor is contiguous.

XXXXRT666:
```cpp
    size_t size() const {
        size_t ret = is_valid();
        for (size_t i = 0; i < ndim(); ++i)
            ret *= shape(i);
        return ret;
    }

    size_t itemsize() const { return ((size_t) dtype().bits + 7) / 8; }
    size_t nbytes() const { return ((size_t) dtype().bits * size() + 7) / 8; }
```

`nd_array.nbytes() / itemsize` represents the logical element count, similar to `ndarray.size`. For x[200:1000:200], the logical size is 4. If we want to preserve the original storage span when copying, it would require a storage size of 601 elements, covering offsets 200 through 800 inclusive. If we only copy the actual selected data and create a new compact array, the resulting storage size would be 4. Which approach to implement depends on the trade-off

XXXXRT666:
```cpp
// Get the location of the end based on the inp strides and out.shape()
int64_t low_idx = 0;
int64_t high_idx = 0;
for (int i = 0; i < inp_strides.size(); ++i) {
  auto delta = inp_strides[i] * (out.shape()[i] - 1);
  if (inp_strides[i] > 0) {
    high_idx += delta;
  } else {
    low_idx += delta;
  }
}
int64_t data_size = (high_idx - low_idx) + 1;
```

after `slice` the data size is calculated like this in `mlx/backend/common/slicing.cpp`

XXXXRT666:
```python
>>> x = np.arange(1200, dtype=np.int32)[200:1000:200]
>>> x
array([200, 400, 600, 800], dtype=int32)
>>> a = mx.from_dlpack(x)
>>> a
array([200, 400, 600, 800], dtype=int32)
>>> a._debug_layout()
{'shape': (4,), 'strides': (200,), 'size': 4, 'data_size': 601, 'no_bsx_size': 4, 'flags': (False, False, False), 'computed_flags': (False, False, False)}
>>> b = mx._debug_from_dlpack_logical_size(x)
>>> b._debug_layout()
{'shape': (4,), 'strides': (200,), 'size': 4, 'data_size': 4, 'no_bsx_size': 4, 'flags': (True, False, False), 'computed_flags': (True, False, False)}
>>> b
array([200, 0, 0, 0], dtype=int32)
```

zcbenz: Thanks for looking into this, so it seems that for non-copy Metal buffer, we should do:

```
data_size = (MTL::Buffer*)(data_handle)->size() / itemsize;
contiguous = (no_bsx_size == data_size);
```

which preserves original layout and correctly set the flags in MLX.

And for buffers to copy into, we should do:

```
data_size = no_bsx_size;
contiguous = true;
```

which just creates a contiguous buffer.

Do you think whether this is correct?

XXXXRT666: >  Which approach to implement depends on the trade-off
> 
I think we don't have to duplicate everything like strides from the old tensor/ndarray in copy path, so it‘s better to construct a new `c_contig` array

But for zero-copy path, we have to calculate the `data_size` using strides

zcbenz: Isn't the result of `strided_storage_size()` is always the same with `no_bsx_size` since they are calculated in the same way? So `contiguous` would just always be true?

XXXXRT666: > Isn't the result of `strided_storage_size()` is always the same with `no_bsx_size` since they are calculated in the same way? So `contiguous` would just always be true?

```cpp
if (strides[i] > 0) {
      no_broadcast_data_size *= shape[i];
}
```

No, `no_bsx_size` is the `no_broadcast_data_size` which is almost the same as product of shape exclude some zero-stride dimension

XXXXRT666: `no_bsx_size` is the number of logical elements after excluding broadcast dimensions, i.e. dimensions with zero stride.

`strided_storage_size()` is the storage span covered by the strided view, i.e. `1 + sum((shape[i] - 1) * strides[i])`.

So they are equal only when the non-broadcast view is compact. For example, `shape=(3, 4), strides=(8, 1)` gives `no_bsx_size=12` but `strided_storage_size=20`, so `contiguous` is false. It's an array with gaps, produced by indexing like `[:,0:4]`. 

Another example, `shape=(3, 4), strides=(0, 1)` has `prod(shape)=12`, but `no_bsx_size=4` and `strided_storage_size()=4`, so it is compact after excluding the broadcast dimension and `contiguous=true`. It's an array  produced by `expand`.

Also `contiguous` here means compact storage. It must have no gaps

zcbenz: Thanks for clarification. My understanding is that `strided_storage_size()` is an effort to calculate the `data_size`, which does not necessarily match the real `data_size` from PyTorch buffer, so it would not be reliable to set `contiguous = (no_bsx_size == storage_size)`? Should we just set `contiguous = is_row_contiguous || is_col_contiguous`, so the kernels would only use the contiguous path when it is absolutely sure?

My other question is, can we use the size of MTLBuffer divided by itemsize as `data_size`?

zcbenz: To answer my own question, the calculated `data_size` is guaranteed to be equal or smaller than the actual size of buffer so it does not matter if we are just working on a slice. I think this implementation is good to go, thanks for your patience!

python/src/convert.cpp
```diff
@@ -31,82 +39,256 @@ int check_shape_dim(int64_t dim) {
...
+      [&]<typename DstT>(mx::Dtype) {
+        auto out = mx::array(shape, dst_dtype, nullptr, {});
+        auto strides = get_strides(nd_array);
+        strided_storage_size(shape, strides);
+        out.set_data(mx::allocator::malloc(out.nbytes()));
+        if (out.size() > 0) {
+          auto src = static_cast<const SrcT*>(nd_array.data());
+          auto dst = out.data<DstT>();
+          for (size_t i = 0; i < out.size(); ++i) {
+            dst[i] = static_cast<DstT>(src[strided_offset(i, shape, strides)]);
```
zcbenz: This is going to be really slow, would it be practical to preserve the original strides and and just do memcpy?

```c++
out.set_data(
    mx::allocator::malloc(data_size * itemsize),
    data_size,
    strides);
std::copy(src, src + data_size, dst);
```

python/src/convert.cpp
```diff
@@ -31,82 +39,256 @@ int check_shape_dim(int64_t dim) {
...
+  mx::array out(shape, src_dtype, nullptr, {});
+  auto flags = out.flags();
+  flags.contiguous = no_bsx_size == storage_size;
+  flags.row_contiguous = is_row_contiguous;
+  flags.col_contiguous = is_col_contiguous;
+  auto import_flags = flags;
+  if (copy && !import_flags.row_contiguous) {
+    // Force the copy primitive to materialize the virtual strided input into a
+    // row-contiguous output instead of preserving a dense non-row layout.
+    import_flags.contiguous = false;
```
zcbenz: Can you collaborate on this? If the input is truly contiguous we shouldn't need this for copying, otherwise the flag was set for non-contiguous input.

mlx/ops.cpp
XXXXRT666: I think this should be handled explicitly rather than relying on argument evaluation order. Since `a.shape()` and `{std::move(a)}` are evaluated as separate function arguments, C++ may evaluate `{std::move(a)}` first. In that case, `a.shape()` would be called after a has been moved from.

Linux and Windows CPU CI failed because of this

python/src/convert.cpp
XXXXRT666: This makes the `astype` copy force the array to be `row_contiguous`. However, I think this is no longer necessary if we preserve the strides and do not require imported arrays to always be `row_contiguous`

python/src/convert.cpp
XXXXRT666: Yes, I checked that PyTorch preserves the strides provided by DLPack on import, including for tensors that are not `non_overlapping_and_dense`. Strides only change if PyTorch performs a copy afterwards, for example when `torch.as_tensor` needs to change `device` or `dtype`, or when using `torch.tensor`, which always copies. In those cases, tensors that are not `non_overlapping_and_dense` are materialized into a compact layout.

I think we do not need to do that in MLX, and can simply preserve the original strides.

XXXXRT666: A `non_overlapping_and_dense` tensor refers to a tensor that is **not** like `x[::2]` or `torch.tensor([1]).expand(3)`

XXXXRT666:
CPU imports now copy the underlying storage span and preserve the `DLPack` strides.

For Metal:
- `copy=False` preserves the `DLPack` strides by wrapping the original buffer.
- `copy=True` preserves strides for `non_overlapping_and_dense` layouts, since those stay on the vector copy path.
- Layouts that are not `non_overlapping_and_dense` may still go through the general copy path and be materialized to a compact layout.

XXXXRT666:
The earlier CI failures came from three independent issues. 

First, the new `array::set_data(..., offset, deleter)` signature accidentally broke existing CUDA call sites that still used the old argument order, causing CUDA compilation failures. 

Second, `astype` had a use-after-move bug: the array was moved into the input list while its shape was also read in the same expression, and the C++ argument evaluation order made this crash on some platforms. 

Third, the macOS Metal validation failure was from a test mutating the original PyTorch MPS tensor after exporting/importing it through DLPack; I think it's a problem with [PyTorch](https://github.com/pytorch/pytorch/pull/181993)

```bash
METAL_DEVICE_WRAPPER_TYPE=1 METAL_DEBUG_ERROR_MODE=0 python -c 'import torch; x=torch.tensor([1.0,2.0,3.0], device="mps"); x+=1; torch.mps.synchronize(); print(x.cpu())'
```

will cause the same error without `MLX`

python/src/convert.cpp
```diff
@@ -471,36 +642,54 @@ mx::array array_from_list(nb::tuple pl, std::optional<mx::Dtype> dtype) {
...
+        "Unable to avoid copy while creating an array as requested.");
+  }
+}
+
+mx::array create_array(
+    nb::object v,
+    std::optional<mx::Dtype> t,
+    std::optional<bool> copy) {
   if (nb::isinstance<nb::bool_>(v)) {
+    check_copy_false(copy);
```
zcbenz: It is error-prone to have to call `check_copy_false` for every branch except one, it might worth doing early return here:

```c++
if (nb::ndarray_check(v)) {
  ...
  return;
}

if (copy.has_value() && copy.value() == false) {
  throw;
}

if (nb::isinstance<nb::bool_>(v)) {
  ...
} else if (...) {
  ...
}
```

python/src/convert.cpp
```diff
@@ -31,145 +39,308 @@ int check_shape_dim(int64_t dim) {
...
+}
+
+template <typename SrcT>
+mx::array cpu_nd_array_to_mlx(
+    nb::ndarray<nb::ro> nd_array,
+    const mx::Shape& shape,
+    mx::Dtype dst_dtype) {
+  return dispatch_dlpack_dtype(
+      mlx_dtype_to_dl_dtype(dst_dtype),
+      [&]<typename DstT>(mx::Dtype) {
```
zcbenz: We have a `dispatch_all_types` utility for this purpose, it was written in C++17 so the usage is a bit awkward though:

```c++
auto out = mx::array(shape, dst_dtype, nullptr, {});
...
if (storage_size > 0) {
  dispatch_all_types(dst_dtype, [&](auto type_tag) {
    using DstT = MLX_GET_TYPE(type_tag);
    auto* src = static_cast<const SrcT*>(nd_array.data());
    auto* dst = out.data<DstT>();
    std::copy(src, src + storage_size, dst);
  });
}
return out;

File: python/src/convert.cpp [hunk 47]
```diff
@@ -31,145 +39,308 @@ int check_shape_dim(int64_t dim) {
...
+    const char* error_message) {
+  return dispatch_dlpack_dtype(
+      type, []<typename T>(mx::Dtype dtype) { return dtype; }, error_message);
 }
 
-template <typename... NDParams>
-nb::ndarray<NDParams...> mlx_to_nd_array(const mx::array& a) {
-  switch (a.dtype()) {
+nb::dlpack::dtype mlx_dtype_to_dl_dtype(mx::Dtype dtype) {
+  switch (dtype) {
```
zcbenz: Similarly here:

```c++
nb::dlpack::dtype result;
dispatch_all_types(dtype, [&](auto type_tag) {
  using T = MLX_GET_TYPE(type_tag);
  result = nb::dtype<T>();
};
return result;
```

(You might need to add some `dtype_traits` for `complex_t` in `python/src/convert.h` though.)

python/src/array.cpp
```diff
@@ -496,7 +500,23 @@ void init_array(nb::module_& m) {
...
+          "__dlpack__",
+          [](const mx::array& a,
+             nb::object,
+             nb::object,
+             std::optional<std::tuple<int, int>> dl_device,
+             std::optional<bool> copy) {
+            if (copy.value_or(false)) {
+              return mlx_to_dlpack(mx::astype(a, a.dtype(), true), dl_device);
+            }
+            return mlx_to_dlpack(a, dl_device);
```
zcbenz, line 523: Should this be:
```suggestion
            if (copy.value_or(false)) {
              return mlx_to_dlpack(a, dl_device);
            }
            return mlx_to_dlpack(mx::astype(a, a.dtype(), true), dl_device);
```

i.e. only copy when `copy=True`?

zcbenz (approved):
Very nice work, thanks!

python/src/array.cpp
XXXXRT666: I think the current code already does this: `copy=True` goes through astype(..., copy=true), while `copy=False/None` exports the original array. `copy.value_or(false)` will be true only if `copy=true`

zcbenz: Yeah you are right, not sure what I was thinking 🤦

python/src/convert.cpp
XXXXRT666: Done. The `ndarray` path now returns early, and the `copy=False` error for the remaining non-DLPack branches is handled once before other branches.

python/src/convert.cpp
XXXXRT666: Done. `cpu_nd_array_to_mlx` now dispatches on `dst_dtype` with `dispatch_all_types`

python/src/convert.cpp
XXXXRT666: Done. `mlx_dtype_to_dl_dtype` now uses `dispatch_all_types`. No extra complex dtype trait was needed I think

XXXXRT666:
Sorry, I forgot to update the nanobind version used by the example extension. It was still building against the PyPI 2.12.0 release while MLX itself uses the pinned nanobind commit. I updated the example extension requirements to use the same commit, so the macOS CI failure should be fixed now

python/src/convert.cpp
```diff
@@ -31,145 +53,282 @@ int check_shape_dim(int64_t dim) {
...
+    std::optional<nb::dlpack::dtype> src_dlpack_dtype_override,
+    std::optional<bool> copy) {
+  auto src_dlpack_dtype = src_dlpack_dtype_override.value_or(nd_array.dtype());
+  auto src_mlx_dtype =
+      mlx_dtype_from_dlpack(src_dlpack_dtype, "Cannot convert array to mlx.");
+  auto dst_dtype = requested_dtype.value_or(src_mlx_dtype);
+  auto device_type = nd_array.device_type();
+  bool can_reuse_buffer = device_type == nb::device::cpu::value ||
+      mx::allocator::can_reuse_alien_buffer(nd_array.data_handle());
+  // call this for cpu array will raise error, use or to aviod it
```
zcbenz: This comment is no longer relevant?

python/src/convert.cpp
```diff
@@ -31,145 +53,282 @@ int check_shape_dim(int64_t dim) {
...
-        nd_array, shape, dtype.value_or(mx::float32));
+    return f.template operator()<float>(mx::float32);
   } else if (type == nb::dtype<double>()) {
-    return nd_array_to_mlx_contiguous<double>(
-        nd_array, shape, dtype.value_or(mx::float32));
+    return f.template operator()<double>(mx::float32);
+  } else if (
+      type ==
+      nb::dlpack::dtype{uint8_t(nb::dlpack::dtype_code::Complex), 32, 1}) {
+    return f.template operator()<DLPackComplex32>(mx::complex64);
```
zcbenz: This works for CPU but would fail for GPU backends, I think we should just let the conversion fail.

python/src/convert.cpp
```diff
@@ -31,145 +53,282 @@ int check_shape_dim(int64_t dim) {
...
   } else if (type == nb::dtype<std::complex<float>>()) {
-    return nd_array_to_mlx_contiguous<mx::complex64_t>(
-        nd_array, shape, dtype.value_or(mx::complex64));
+    return f.template operator()<mx::complex64_t>(mx::complex64);
   } else if (type == nb::dtype<std::complex<double>>()) {
-    return nd_array_to_mlx_contiguous<mx::complex128_t>(
-        nd_array, shape, dtype.value_or(mx::complex64));
+    return f.template operator()<mx::complex128_t>(mx::complex64);
```
zcbenz, line 128: Same here.

zcbenz (commented):
This PR is good to merge, but I think we should wait for a new release in nanobind, I will ask for one.

python/src/convert.cpp
XXXXRT666: The old comment was a bit stale and unclear. We should not call `can_reuse_alien_buffer()` on CPU `ndarray.data_handle()`, since that is a host pointer rather than a GPU buffer handle. I updated the code to make this explicit.

python/src/convert.cpp
XXXXRT666: Removed now

python/src/convert.cpp
XXXXRT666: Done

XXXXRT666: I restored the `complex128 -> complex64` mapping because `NumPy` creates complex arrays as `complex128` by default, and the existing NumPy import path used to accept that and cast it to `MLX` `complex64`

For Metal DLPack, I think keeping this mapping should still be OK in practice, since most frameworks do not support creating `complex128` Metal arrays. If such an array is passed in, the `itemsize` check will still reject layouts that do not match the MLX dtype size

zcbenz: That makes sense, thanks!

PR #3537, opened 2026-05-12
Apply the same thread-local approach to the CPU

angeloskath:
As the title says, apply the same logic to protect the CPU command encoders.

Specifically, the temporaries and the num_ops which could be edited by different threads simultaneously.

Also fixed the test since it was silently failing for the CPU.


zcbenz (approved):
Nice!

PR #3538, opened 2026-05-13
ci: Show stack trace on crash

zcbenz:
When test fails, check if there is coredump generated, and use gdb to print symbolicated stack trace when there is there is one.

A test run:
https://github.com/ml-explore/mlx/actions/runs/25771334892/job/75695094341

Screenshot 2026-05-13 at 10 05 40


angeloskath (approved):
🙏

PR #3551, opened 2026-05-14
[Metal] Reject tensor-scale nvfp4 in qqmm

### LINKED ISSUE #3550: Metal: `QQMatmul::eval_gpu` gemv path silently drops `global_scale_x`/`global_scale_w` for nvfp4
## Summary

`QQMatmul::eval_gpu` on Metal silently drops the `global_scale_x` / `global_scale_w` inputs when running `qqmm` in `nvfp4` mode along the gemv special case (`x.shape(-2) == 1` with pre-quantized `w`). The general case correctly throws `"[QQMatmul] NYI for the general case"`, but the gemv branch runs to completion using only per-group scales — producing numerically incorrect results when tensor-scale nvfp4 weights are actually in use.

The sibling APIs `quantize()` and `dequantize()` already guard this on Metal:

```cpp
// mlx/ops.cpp:4940-4945  (quantize)
if (to_stream(s).device == Device::gpu && metal::is_available() &&
    global_scale.has_value()) {
  throw std::invalid_argument("[quantize] Global scale is not supported on the Metal backend.");
}
// mlx/ops.cpp:5205-5210 has the same guard for dequantize().
```

…but `qqmm()` has no equivalent guard, so the global scales get packed into `inputs` (`ops.cpp:4666-4669`) and propagate to Metal where the gemv path ignores them.

## Where the inputs get dropped

`mlx/backend/metal/quantized.cpp:1611-1655`:

```cpp
void QQMatmul::eval_gpu(const std::vector<array>& inputs, array& out) {
  // ...
  bool w_quantized = (inputs[1].dtype() == uint32);
  if (w_quantized && inputs[0].shape(-2) == 1) {
    // ...
    quantize_dequantize(x, xhat, mode, group_size_, bits_, d, s); // <-- no global_scale_x
    // ...
    dispatch_qmv(xhat, w, scales, std::nullopt, out, group_size_, bits_, M, N, K, d, s, mode);
    //                                ^^^^^^^^^^                                        ^^^^
    //                            biases slot                              no global_scale_w
    return;
  } else {
    throw std::runtime_error("[QQMatmul] NYI for the general case");
  }
}
```

`inputs[3]` (`global_scale_x`) and `inputs[4]` (`global_scale_w`) — packed by `ops.cpp` when both are provided for nvfp4 — are never read on this branch.

For comparison, the CUDA backend reads them in `mlx/backend/cuda/quantized/qqmm.cpp:93,134-135`:

```cpp
// CUDA reads from the tail of inputs[]
global_scale = inputs[inputs.size() - 2];                   // qqmm.cpp:93
global_scale_x = inputs[inputs.size() - 2];                 // qqmm.cpp:134
global_scale_w = inputs[inputs.size() - 1];                 // qqmm.cpp:135
```

## Impact

Tensor-scale nvfp4 (introduced in #3022) is intentionally CPU/CUDA-only — That decision was carried out for `quantize` / `dequantize` via the explicit Metal guards above, but `qqmm` was missed: the gemv path on Metal does not throw, it just silently produces wrong outputs.

## Suggested fix

Mirror the general case's behavior — throw NYI from the gemv path when global scales are present — until proper tensor-scale-aware Metal kernels land. Something like:

```cpp
void QQMatmul::eval_gpu(const std::vector<array>& inputs, array& out) {
  // Reject tensor-scale nvfp4 on Metal until the kernels support it.
  if (mode_ == QuantizationMode::Nvfp4 && inputs.size() > 3) {
    throw std::runtime_error(
        "[QQMatmul] nvfp4 with tensor-scale (global_scale_x/w) is not supported on the Metal backend.");
  }
  // ...existing body...
}
```

Alternatively, add the API-level guard in `qqmm()` in `ops.cpp` next to the existing `quantize` / `dequantize` ones so the error is raised at graph-construction time rather than during evaluation.

## Context

Found while bringing up tensor-scale nvfp4 in a downstream project. We work around it by sticking to per-group nvfp4 (no `global_scale`) for Metal-only inference paths, which is unaffected. Filing this so the silent-incorrect behavior is either gated with an explicit error or eventually implemented in Metal.

Brooooooklyn:
## Summary

`QQMatmul::eval_gpu` on Metal silently dropped `global_scale_x` / `global_scale_w` in its gemv special case (pre-quantized `w`, `x.shape(-2) == 1`), producing numerically incorrect results when tensor-scale nvfp4 weights were in use. The general case already throws `[QQMatmul] NYI for the general case`. `qqmm()` was missed when tensor-scale nvfp4 landed in #3022.

This change adds a backend-level guard at the top of `QQMatmul::eval_gpu` in `mlx/backend/metal/quantized.cpp`: when `mode_ == Nvfp4` and extra inputs beyond `(x, w[, scales_w])` are present (i.e. global scales were packed by `ops.cpp`), it throws `std::runtime_error` (Python `RuntimeError`) — matching the local throw style used by the existing NYI path and keeping the device-specific check out of the backend-agnostic `ops.cpp`. CUDA is unaffected.

Per-group nvfp4 (no `global_scale`) on Metal continues to work — that path is exercised by the existing `test_qqmv` and is unchanged.

Fixes #3550.

## Test plan

- [x] Added `test_qqmm_metal_global_scale_rejected` in `python/tests/test_quantized.py`, asserts `RuntimeError` when `mx.qqmm` is evaluated on Metal with both global scales set. Verified the test fails on `main` (silently runs to completion) and passes with this fix.
- [x] Full `python/tests/test_quantized.py` (29 tests) still passes locally on Apple Silicon, including `test_qqmv` which exercises the gemv branch being guarded.

### DIFF
--- a/mlx/backend/metal/quantized.cpp
+++ b/mlx/backend/metal/quantized.cpp
@@ -1614,6 +1614,17 @@ void QQMatmul::eval_gpu(const std::vector<array>& inputs, array& out) {
 
   auto mode = quantization_mode_to_string(mode_);
   bool w_quantized = (inputs[1].dtype() == uint32);
+  // Tensor-scale nvfp4 (global_scale_x / global_scale_w) is packed into
+  // inputs by ops.cpp but no Metal qqmm kernel currently consumes the
+  // global scales. Reject the request rather than silently dropping them
+  // in the gemv path below.
+  int base_size = w_quantized ? 3 : 2;
+  if (mode_ == QuantizationMode::Nvfp4 &&
+      static_cast<int>(inputs.size()) > base_size) {
+    throw std::runtime_error(
+        "[QQMatmul] Global scale (tensor-scale nvfp4) is not supported "
+        "on the Metal backend.");
+  }
   if (w_quantized && inputs[0].shape(-2) == 1) {
     out.set_data(allocator::malloc(out.nbytes()));
 
--- a/python/tests/test_quantized.py
+++ b/python/tests/test_quantized.py
@@ -205,6 +205,32 @@ class TestQuantized(mlx_tests.MLXTestCase):
                     self.assertEqual(y_q.shape, y_hat.shape)
                     self.assertLess((y_q - y_hat).abs().max(), 1e-3)
 
+    def test_qqmm_metal_global_scale_rejected(self):
+        # Tensor-scale nvfp4 (global_scale_x / global_scale_w) is not
+        # implemented in the Metal qqmm kernels. mx.qqmm must reject the
+        # request on Metal rather than silently dropping the global scales
+        # in the gemv path and producing incorrect results.
+        if not mx.metal.is_available():
+            return
+
+        w = mx.random.normal(shape=(64, 64))
+        w_q, scales = mx.quantize(w, mode="nvfp4")
+        x = mx.random.normal(shape=(1, 64))
+        gx = mx.array(1.0, dtype=mx.float32)
+        gw = mx.array(1.0, dtype=mx.float32)
+
+        with self.assertRaises(RuntimeError):
+            y = mx.qqmm(
+                x,
+                w_q,
+                scales,
+                mode="nvfp4",
+                global_scale_x=gx,
+                global_scale_w=gw,
+                stream=mx.gpu,
+            )
+            mx.eval(y)
+
     def test_qmm(self):
         key = mx.random.key(0)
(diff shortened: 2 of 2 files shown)


zcbenz (commented):
This check should be done in `QQMatmul::eval_gpu`, generally we should not check backend types in `ops.cpp`.

Brooooooklyn:
Updated

zcbenz (approved):
Thanks!

PR #3554, opened 2026-05-16
Add buffer caching to no_gpu CPU allocator

dhiltgen:
## Proposed changes

Split out from #3019

Integrate BufferCache into the CPU allocator to enable memory reuse for CPU-only builds. Previously the no_gpu allocator called malloc/free on every allocation with no caching, while the Metal and CUDA backends had buffer caching for better performance.

Track cached buffers by their physical capacity when they are reused so get_cache_memory(), active memory, and cache limit enforcement continue to reflect retained memory. Add a regression test for reusing a larger cached block for a smaller request.

Changes:
- Add CpuCachedBuffer struct with intrusive freelist for object pooling
- Use BufferCache to recycle freed buffers with a 32MB default cache limit
- Preserve cached block capacity across reuse and avoid caching zero-size allocations
- Implement get_cache_memory(), set_cache_limit(), clear_cache() (were no-ops)
- Cache-first allocation path with fallback to OS malloc on cache miss

## Checklist

### DIFF
--- a/mlx/backend/no_gpu/allocator.cpp
+++ b/mlx/backend/no_gpu/allocator.cpp
@@ -1,9 +1,10 @@
-// Copyright © 2023 Apple Inc.
+// Copyright © 2023-2026 Apple Inc.
 
 #include <algorithm>
 #include <mutex>
 
 #include "mlx/allocator.h"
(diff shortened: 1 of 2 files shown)


zcbenz (approved):
Looks good to me, thanks!

PR #3555, opened 2026-05-16
Fix singleton lifetime issues at process exit

dhiltgen:
## Proposed changes

Split out from #3019

Leak the IO ThreadPool singletons and CPU CompilerCache using the same process-lifetime pattern already used by the Scheduler singleton.

The CompilerCache owns dlopen handles for JIT shared libraries. Destroying it during static teardown can dlclose generated code while stream worker threads may still be winding down. The IO loader thread pools have the same shutdown-order risk on Windows CRT teardown. These objects are process-lifetime infrastructure, and the OS reclaims their resources at exit.

Changes:
- Leak CompilerCache so JIT libraries remain mapped through process exit
- Leak IO ThreadPool singletons to avoid teardown-order races
- Clarify the Scheduler singleton comment that documents this pattern

## Checklist

### DIFF
--- a/mlx/backend/cpu/compiled.cpp
+++ b/mlx/backend/cpu/compiled.cpp
@@ -41,8 +41,11 @@ struct CompilerCache {
 };
 
 static CompilerCache& cache() {
-  static CompilerCache cache_;
-  return cache_;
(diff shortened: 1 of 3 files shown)


zcbenz (approved):
👍

PR #3463, opened 2026-04-28
Do not use prebuilt cpu compile preamble when headers are installed

zcbenz:
Check whether the headers are shipped with the binary, if so use the preamble from the headers, otherwise use the prebuilt one embeded in binary.

We can't just always use the prebuilt one because it may not work with all the compilers, closes #3430, closes #3080.

PR #3556, opened 2026-05-16
Add JIT compiler support for Windows

dhiltgen:
## Proposed changes

Split out from #3019

Enable CPU mx.compile() on Windows by detecting and using clang-cl bundled with Visual Studio, or MSVC cl.exe, for JIT compilation.

Keep GPU compile availability independent from the CPU compiler probe so CPU+GPU builds do not disable GPU mx.compile() when a host C++ compiler is unavailable.

Changes:
- Add clang-cl detection via vswhere and prefer a compiler matching the build toolchain
- Add JitCompiler::available() to probe CPU JIT availability
- Emit and load .dll JIT libraries on Windows
- Support both MSVC and GCC/Clang preamble generation scripts, including optional SIMD flags
- Use WIN32 shell detection and pass preamble SIMD flags through CMake
- Define NOMINMAX/WIN32_LEAN_AND_MEAN on all WIN32 compilers

## Checklist


mlx/backend/cpu/jit_compiler.cpp
```diff
@@ -70,12 +77,34 @@ struct VisualStudioInfo {
...
+        msvc_cl = fmt::format("{0}\\bin\\Host{1}\\{1}\\cl.exe", value, arch);
+      }
+    }
+
+    // Check for clang-cl bundled with Visual Studio.
+    std::string clang_cl_path = fmt::format(
+        "{0}\\VC\\Tools\\Llvm\\{1}\\bin\\clang-cl.exe", vs_path, arch);
+    {
+      std::ifstream f(clang_cl_path);
+      if (f.good()) {
```
zcbenz: We usually use `std::filesystem::exists`.

mlx/backend/cpu/jit_compiler.cpp
```diff
@@ -110,36 +139,92 @@ const std::tuple<bool, std::string, std::string>& JitCompiler::get_preamble() {
...
+#else
+      result = !info.msvc_cl.empty() ? 1 : 0;
+#endif
+    } catch (...) {
+      result = 0;
+    }
+  }
+  return result == 1;
+#else
+  static int result = -1;
```
zcbenz:
```suggestion
  static int result = std::system("...");
  return result == 0;
```

mlx/backend/cpu/jit_compiler.cpp
```diff
@@ -70,12 +77,34 @@ struct VisualStudioInfo {
...
+      std::ifstream f(clang_cl_path);
+      if (f.good()) {
+        clang_cl = clang_cl_path;
       }
     }
+
+    // Select the JIT compiler. The preamble was preprocessed at build time
+    // by whichever compiler built the library -- it contains compiler-specific
+    // builtins (e.g. __builtin_fpclassify for Clang, __is_same for MSVC) that
+    // are only valid for the same compiler family. Prefer the matching one.
```
zcbenz: You should ship the preamble headers along with your binary, which we do in python and can root out this problem completely (https://github.com/ml-explore/mlx/pull/3463).

mlx/backend/cpu/jit_compiler.cpp
```diff
@@ -110,36 +139,92 @@ const std::tuple<bool, std::string, std::string>& JitCompiler::get_preamble() {
...
+    libpaths += fmt::format(" /libpath:\"{}\"", lib);
+  }
+  // clang-cl accepts the same flags as cl.exe (/LD, /EHsc, etc.)
+  // but we add -Wno-everything to suppress warnings from the preprocessed
+  // preamble, which may contain pragmas or builtins from a different compiler
+  // (e.g. MSVC pragmas when compiling with clang-cl, or vice versa).
+  if (!info.clang_cl.empty() && info.cl_exe == info.clang_cl) {
+    compiler_flags += " -Wno-everything";
   }
+#ifdef __AVX2__
```
zcbenz, line 243: AVX2 should be checked at run time, you can't ensure binary built with AVX2 enabled would always run on machines with it.

mlx/backend/cpu/jit_compiler.cpp
```diff
@@ -110,36 +139,92 @@ const std::tuple<bool, std::string, std::string>& JitCompiler::get_preamble() {
...
+  compiler_flags += " /arch:AVX2";
+#endif
   return fmt::format(
       "\""
-      "cd /D \"{}\" && "
-      "\"{}\" /LD /EHsc /MD /Ox /nologo /std:c++17 {} \"{}\" "
-      "/link /out:\"{}\" 2>&1"
+      "cd /D \"{0}\" && "
+      "\"{1}\" /LD /EHsc /MD /Ox /nologo /std:c++17{5} \"{2}\" "
```
zcbenz: I prefer listing the args in increasing order otherwise I have to count the index of the args to know where they would be inserted.

zcbenz (approved):
Thanks!

PR #3559, opened 2026-05-16
Remove reference to groups in Conv3D extra_repr

GriffinMB:
## Proposed changes

Conv3D does not have a `groups` property. This removes the reference to the invalid prop to resolve errors when printing the object. 

## Checklist


zcbenz (approved):
Thanks!

PR #3560, opened 2026-05-17
Fix steel GEMM safe load offset

sofinvalery:
Fixed typo:)


jagrit06 (approved):
Thanks for catching that!

PR #3561, opened 2026-05-18
Fix doubled-word typos in docstrings

LeSingh1:
## Summary

Five trivial doubled-word typos in user-visible docstrings:

| File | Line | Fix |
|---|---|---|
| `python/mlx/nn/losses.py` | 275 | `gaussian_nll_loss` docstring: "are the the predicted variances" → "are the predicted variances" |
| `python/mlx/nn/layers/distributed.py` | 202, 286 | `AllToShardedLinear` / `ShardedToAllLinear` bias arg: "If set to ``False`` the the layer will not use a bias" → "the layer will not use a bias" |
| `python/src/ops.cpp` | 840 | `logaddexp` docstring: "The computation is is a numerically stable version" → "The computation is a numerically stable version" |
| `python/src/transforms.cpp` | 1520 | `compile()` return-value docstring: "returns the the same output(s)" → "returns the same output(s)" |

## Test plan

- [x] `python3 -c "import ast; ast.parse(open('python/mlx/nn/losses.py').read()); ast.parse(open('python/mlx/nn/layers/distributed.py').read())"` — both files still parse.
- [x] `grep -rEn '\b(the the|of of|is is|to to|and and|that that|in in|a a)\b' --include='*.py' --include='*.cpp' --include='*.h' python/` — no remaining hits in user-visible code.

## Scope

Docstring text only. No code, behavior, or API affected.

### DIFF
--- a/python/mlx/nn/layers/distributed.py
+++ b/python/mlx/nn/layers/distributed.py
@@ -199,7 +199,7 @@ class AllToShardedLinear(Module):
     Args:
         input_dims (int): The dimensionality of the input features
         output_dims (int): The dimensionality of the output features
-        bias (bool, optional): If set to ``False`` the the layer will not use a
+        bias (bool, optional): If set to ``False`` the layer will not use a
             bias. Default is ``True``.
         group (mx.distributed.Group, optional): The sharding will happen across
             this group. If not set then the global group is used. Default is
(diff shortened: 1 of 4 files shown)


angeloskath (approved):
Thanks!

PR #3562, opened 2026-05-19
Find SwiftPM metallib bundle near test binary

RNT56:
## Summary

Fixes SwiftPM package/test layouts where `default.metallib` is copied to the generated `SWIFTPM_BUNDLE.bundle` next to the test binary, but MLX only checks nested bundle locations.

This keeps the existing lookup order and adds:

- direct lookup for `SWIFTPM_BUNDLE.bundle` next to the current binary, walking a few parent directories
- direct lookup when `NSBundle.allBundles()` already contains the SwiftPM resource bundle

## Validation

- `pre-commit run --all-files` passed
- validated through the `mlx-swift` SwiftPM packaging branch:
  - `swift test --filter StreamTests/testDeviceType` passes after applying this MLX loader patch and generating `default.metallib`
  - the same test failed before this patch with `Failed to load the default metallib`

### DIFF
--- a/mlx/backend/metal/device.cpp
+++ b/mlx/backend/metal/device.cpp
@@ -80,12 +80,10 @@ std::pair<MTL::Library*, NS::Error*> load_library_from_path(
 }
 
 #ifdef SWIFTPM_BUNDLE
-MTL::Library* try_load_bundle(
+MTL::Library* try_load_bundle_path(
     MTL::Device* device,
-    NS::URL* url,
+    const std::string& bundle_path,
     const std::string& lib_name) {
-  std::string bundle_path = std::string(url->fileSystemRepresentation()) + "/" +
-      SWIFTPM_BUNDLE + ".bundle";
   auto bundle = NS::Bundle::alloc()->init(
       NS::String::string(bundle_path.c_str(), NS::UTF8StringEncoding));
   if (bundle != nullptr) {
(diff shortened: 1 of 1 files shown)


zcbenz:
Ping @davidkoski for review.

zcbenz:
At the moment I prefer adding an API to set the path to metallib directly, i.e. https://github.com/ml-explore/mlx/pull/3597, and I plan to close this PR, please let me know if the API would not be able to solve your use case.

davidkoski:
I think API to set a path would be fine, but I don't know how early it needs to be called.  IIRC the worker threads start via static initializers and there is no way to guarantee being able to get in before that (perhaps that has changed?).  Anyway I think the automatic scheme we have is ok and having an override would be useful.

davidkoski:
Ah, it looks like it _is_ changed -- it happens when you first ask for `Stream default_stream(Device d)` / `scheduler::scheduler().get_default_stream(d)`

zcbenz:
Yeah the metal initialization has been delayed until a stream is actually created, it was to solve issues like https://github.com/ml-explore/mlx/issues/3148. So the API only needs to be called before the stream was explicitly or implicitly created.

I think we should keep current metallib discovery behavior unchanged, and ask users to use the API if they have some unusual bundle layout, do you think whether it is a good idea?

davidkoski:
That seems like a good plan -- then if somebody has custom lookup needs they can implement it in application code.

PR #3563, opened 2026-05-19
Fix CPU GatherMM one-row NaNs

### LINKED ISSUE #3200: [BUG] test_gather_qmm_sorted consistently fails in some aarch64-darwin environments
**Describe the bug**

This library is packaged for nixpkgs (for both darwin and linux). Its test suite is executed as part of package build. When it's executed on Apple Silicon machines I have in my possession, the test suite passes fine. When it's executed on Nix build system (Hydra) machines, it often [fails](https://cache.nixos.org/log/qvx2kwkjqj9rdi90h259jh6ck09cdgk0-python3.13-mlx-0.30.5.drv) as follows:

```
pytest flags: -m pytest python/tests/ -k not\ \(test_siblings_without_eval\)
============================= test session starts ==============================
platform darwin -- Python 3.13.11, pytest-8.4.2, pluggy-1.6.0
rootdir: /nix/var/nix/builds/nix-13401-1802369029/source
configfile: pyproject.toml
collected 672 items / 1 deselected / 671 selected                              

python/tests/test_array.py .......................................s..... [  6%]
.......................                                                  [ 10%]
python/tests/test_autograd.py ...............................            [ 14%]
python/tests/test_bf16.py ...s..                                         [ 15%]
python/tests/test_blas.py ........................                       [ 19%]
python/tests/test_compile.py ........................................... [ 25%]
.........                                                                [ 26%]
python/tests/test_constants.py ...                                       [ 27%]
python/tests/test_conv.py .s........ssssssss                             [ 30%]
python/tests/test_conv_transpose.py sssssssss                            [ 31%]
python/tests/test_device.py ..s......s                                   [ 32%]
python/tests/test_double.py ..........                                   [ 34%]
python/tests/test_einsum.py ..........                                   [ 35%]
python/tests/test_eval.py ...........s.                                  [ 37%]
python/tests/test_export_import.py ....s...........                      [ 40%]
python/tests/test_fast.py ssssss..............                           [ 43%]
python/tests/test_fast_sdpa.py .....s..........                          [ 45%]
python/tests/test_fft.py ...ss........                                   [ 47%]
python/tests/test_graph.py .                                             [ 47%]
python/tests/test_init.py ..........                                     [ 49%]
python/tests/test_linalg.py ................                             [ 51%]
python/tests/test_load.py ............                                   [ 53%]
python/tests/test_losses.py ..............                               [ 55%]
python/tests/test_memory.py ..s                                          [ 55%]
python/tests/test_nn.py ................................................ [ 63%]
.............                                                            [ 64%]
python/tests/test_ops.py ............................................... [ 71%]
........................................................................ [ 82%]
.............                                                            [ 84%]
python/tests/test_optimizers.py ....s...................                 [ 88%]
python/tests/test_quantized.py ......F...................                [ 92%]
python/tests/test_random.py ..............                               [ 94%]
python/tests/test_reduce.py ..........                                   [ 95%]
python/tests/test_tree.py ....                                           [ 96%]
python/tests/test_upsample.py s                                          [ 96%]
python/tests/test_vmap.py ........................                       [100%]

=================================== FAILURES ===================================
_____________________ TestQuantized.test_gather_qmm_sorted _____________________

self = <test_quantized.TestQuantized testMethod=test_gather_qmm_sorted>

    def test_gather_qmm_sorted(self):
        def quantize(w, transpose=True, group_size=None, mode="affine"):
            if mode == "affine":
                qw, s, b = mx.quantize(w, group_size=group_size, mode=mode)
            else:
                qw, s = mx.quantize(w, mode=mode)
                b = None
    
            w_hat = mx.dequantize(qw, s, b, group_size=group_size, mode=mode)
            if transpose:
                w_hat = w_hat.swapaxes(-1, -2)
            return w_hat, qw, s, b
    
        def gather_sort(x, indices):
            N, M = indices.shape
            indices = indices.flatten()
            order = mx.argsort(indices)
            inv_order = mx.argsort(order)
            return x.flatten(0, -3)[order // M], indices[order], inv_order
    
        def scatter_unsort(x, inv_order, shape=None):
            x = x[inv_order]
            if shape is not None:
                x = mx.unflatten(x, 0, shape)
            return x
    
        parameters = [
            # L, K, D, E, I, transpose
            (32, 512, 512, 4, 2, True, "affine"),
            (32, 512, 544, 4, 2, True, "mxfp4"),
            (32, 512, 544, 4, 2, True, "nvfp4"),
            (32, 512, 544, 4, 2, True, "mxfp8"),
            (133, 512, 512, 4, 2, True, "affine"),
            (133, 512, 555, 4, 2, True, "affine"),
            (133, 512, 512, 4, 2, True, "affine"),
            (64, 512, 512, 4, 2, False, "affine"),
            (64, 512, 544, 4, 2, False, "mxfp4"),
            (64, 512, 544, 4, 2, False, "nvfp4"),
            (64, 512, 544, 4, 2, False, "mxfp8"),
            (133, 512, 512, 4, 2, False, "affine"),
            (133, 512, 544, 4, 2, False, "affine"),
            (133, 512, 555, 4, 2, False, "affine"),
            (64, 512, 512, 4, 2, False, "affine"),
        ]
    
        key = mx.random.key(0)
        k1, k2, k3 = mx.random.split(key, 3)
        dtype = mx.float16 if (mx.default_device() == mx.gpu) else mx.float32
    
        for L, K, D, E, I, transpose, mode in parameters:
            with self.subTest(L=L, K=K, D=D, E=E, I=I, transpose=transpose, mode=mode):
                if mode != "affine":
                    group_size = None
                    dtype = (
                        mx.bfloat16 if (mx.default_device() == mx.gpu) else mx.float32
                    )
                else:
                    group_size = 64
                    dtype = (
                        mx.float16 if (mx.default_device() == mx.gpu) else mx.float32
                    )
    
                K, D = (K, D) if transpose else (D, K)
                ishape = (L, I)
                xshape = (L, 1, 1, K)
                wshape = (E, D, K) if transpose else (E, K, D)
    
                indices = (mx.random.uniform(shape=ishape, key=k1) * E).astype(
                    mx.uint32
                )
                x = mx.random.normal(xshape, key=k2) / K**0.5
                w = mx.random.normal(wshape, key=k3) / K**0.5
    
                x = x.astype(dtype)
                w = w.astype(dtype)
    
                w, *wq = quantize(
                    w, group_size=group_size, mode=mode, transpose=transpose
                )
    
                y1 = mx.gather_mm(x, w, rhs_indices=indices)
                y2 = mx.gather_qmm(
                    x,
                    *wq,
                    group_size=group_size,
                    mode=mode,
                    transpose=transpose,
                    rhs_indices=indices
                )
                xs, idx, inv_order = gather_sort(x, indices)
                y3 = mx.gather_mm(xs, w, rhs_indices=idx, sorted_indices=True)
    
                y4 = mx.gather_qmm(
                    xs,
                    *wq,
                    group_size=group_size,
                    mode=mode,
                    rhs_indices=idx,
                    transpose=transpose,
                    sorted_indices=True
                )
                y3 = scatter_unsort(y3, inv_order, indices.shape)
                y4 = scatter_unsort(y4, inv_order, indices.shape)
    
                tol = 1.5e-5 if (dtype == mx.float32) else 2.5e-4
    
                self.assertLess((y1 - y2).abs().max(), tol)
>               self.assertLess((y1 - y3).abs().max(), tol)
E               AssertionError: array(nan, dtype=float32) not less than 1.5e-05

python/tests/test_quantized.py:980: AssertionError
=========================== short test summary info ============================
FAILED python/tests/test_quantized.py::TestQuantized::test_gather_qmm_sorted - AssertionError: array(nan, dtype=float32) not less than 1.5e-05
===== 1 failed, 634 passed, 36 skipped, 1 deselected in 164.14s (0:02:44) ======
```

**The test doesn't fail every time the suite runs, but most of the times.**

The execution environment may be somewhat different, but regardless of the machine, the package is always built [without](https://github.com/NixOS/nixpkgs/blob/82f83ba131f02dd10cd0c9d6b38d876485835589/pkgs/development/python-modules/mlx/default.nix#L89) Metal support.

All machines - both Hydra and my private machines - have Apple Silicon. Hydra machines may have older M-series CPUs, but I also tried to reproduce the failure on a M1 Air laptop I have, with no luck.

Sadly, we do not have direct access to these machines to run a debugging session, so it makes it hard to debug.

**To Reproduce**

As I said, sadly we are only able to reproduce it in build environment that is not conducive to debugging or reproduction attempts.

**Expected behavior**
Tests don't fail.

 - OS Version: I believe all MacOS Hydra machines were upgraded to 26.x.
 - Version: 0.30.5, but it was happening since forever.

**Additional context**

Maybe worth noting is that other tests in the test class always pass. Also maybe worth noting is that we [skipped](https://github.com/NixOS/nixpkgs/blob/82f83ba131f02dd10cd0c9d6b38d876485835589/pkgs/development/python-modules/mlx/default.nix#L131) another test on aarch64-darwin because of a memory allocator related failure, for which I have a [patch](https://github.com/ml-explore/mlx/pull/3088) that adjusts the test case. I wonder if this slight difference in allocator behavior could point to a memory leak and be relevant here.

We have means to disable tests in nixpkgs package builds, but I am hesitant to sweep it under the rug and wonder if mlx core team members could give an idea on how to fix or debug it. If not a fix, then maybe the test case or suite could be improved to allow to collect more information about this failure, so that the next time we hit it we could have a better idea of what happens.

TyceHerrman:
# Fix CPU GatherMM one-row NaNs

Fixes #3200.

This fixes an intermittent CPU-only `GatherMM` failure that can produce NaNs after quantized gather warmup on no-Metal aarch64-darwin builds.

I reproduced the intermittent NaN locally with a CPU-only build and the quantized test loop now passes 30 consecutive runs.

The CPU `GatherMM` path now handles `M == 1` gathered matmuls with a small local one-row helper instead of routing that shape through BLAS. The regular `cblas_sgemm` path remains unchanged for larger `M`.

Changes:
- Add `gather_mm_one_row(...)` for one-row CPU gathered matmuls.
- Use that helper only when `M == 1` in `GatherMM::eval_cpu`.
- Add a focused CPU regression covering the stateful quantized warmup sequence that reproduced the NaN before this fix.

Testing:

```sh
CMAKE_ARGS='-DMLX_BUILD_METAL=OFF' PIP_CACHE_DIR=/private/tmp/codex-pip-cache python -m pip install -e . --no-build-isolation -q
PYTHONPATH=python/tests DEVICE=cpu python -m unittest test_blas.TestBlas.test_gather_mm_cpu_m1_quantized_warmup
for i in $(seq 1 30); do PYTHONPATH=python/tests DEVICE=cpu python -m unittest test_quantized.TestQuantized.test_gather_qmm_sorted || exit 1; done
PYTHONPATH=python/tests DEVICE=cpu python -m unittest test_blas test_quantized
git diff --check
```

AI Assistance Disclosure: Codex assisted with debugging, implementation, and PR draft preparation. The changes were reviewed and tested locally by me before submission.

@booxter

### DIFF
--- a/mlx/backend/cpu/masked_mm.cpp
+++ b/mlx/backend/cpu/masked_mm.cpp
@@ -105,6 +105,27 @@ inline void segmented_mm(
   }
 }
 
+inline void gather_mm_one_row(
+    const float* a,
+    const float* b,
+    float* out,
+    size_t N,
+    size_t K,
+    size_t lda,
+    size_t ldb,
+    bool a_transposed,
+    bool b_transposed) {
+  for (size_t n = 0; n < N; n++) {
+    float sum = 0.0f;
+    for (size_t k = 0; k < K; k++) {
+      auto a_val = a_transposed ? a[k * lda] : a[k];
+      auto b_val = b_transposed ? b[n * ldb + k] : b[k * ldb + n];
+      sum += a_val * b_val;
+    }
+    out[n] = sum;
+  }
+}
+
 } // namespace
 
 void BlockMaskedMM::eval_cpu(const std::vector<array>& inputs, array& out) {
@@ -468,21 +489,31 @@ void GatherMM::eval_cpu(const std::vector<array>& inputs, array& out) {
       uint32_t indx_B = rhs_indices_ptr[elem_to_loc(
           i, rhs_indices_shape, rhs_indices_strides)];
 
-      cblas_sgemm(
-          CblasRowMajor,
-          a_transposed ? CblasTrans : CblasNoTrans, // transA
-          b_transposed ? CblasTrans : CblasNoTrans, // transB
-          M,
-          N,
-          K,
-          1.0f, // alpha
-          a_ptr + elem_to_loc(indx_A, batch_shape_A, batch_strides_A),
-          lda,
-          b_ptr + elem_to_loc(indx_B, batch_shape_B, batch_strides_B),
-          ldb,
-          0.0f, // beta
-          out_ptr + matrix_stride_out * i,
-          ldc);
+      auto ai = a_ptr + elem_to_loc(indx_A, batch_shape_A, batch_strides_A);
+      auto bi = b_ptr + elem_to_loc(indx_B, batch_shape_B, batch_strides_B);
+      auto ci = out_ptr + matrix_stride_out * i;
+
+      if (M == 1) {
+        // Avoid intermittent BLAS NaNs for one-row gathered CPU matmuls.
+        gather_mm_one_row(
+            ai, bi, ci, N, K, lda, ldb, a_transposed, b_transposed);
+      } else {
+        cblas_sgemm(
+            CblasRowMajor,
+            a_transposed ? CblasTrans : CblasNoTrans, // transA
+            b_transposed ? CblasTrans : CblasNoTrans, // transB
+            M,
+            N,
+            K,
+            1.0f, // alpha
+            ai,
+            lda,
+            bi,
+            ldb,
+            0.0f, // beta
+            ci,
+            ldc);
+      }
     }
   });
   encoder.add_temporaries(std::move(temps));
--- a/python/tests/test_blas.py
+++ b/python/tests/test_blas.py
@@ -1264,6 +1264,77 @@ class TestBlas(mlx_tests.MLXTestCase):
                     c2 = gather_mm_test(a, b, rhs)
                     self.assertTrue(mx.allclose(c1, c2, rtol=tol, atol=tol))
 
+    def test_gather_mm_cpu_m1_quantized_warmup(self):
+        if mx.default_device() != mx.cpu:
+            self.skipTest("CPU only")
+
+        def make_case(L, K, D, E, I, transpose, mode):
+            group_size = None if mode != "affine" else 64
+            K, D = (K, D) if transpose else (D, K)
+            key = mx.random.key(0)
+            k1, k2, k3 = mx.random.split(key, 3)
+            indices = (mx.random.uniform(shape=(L, I), key=k1) * E).astype(
+                mx.uint32
+            )
+            x = (mx.random.normal((L, 1, 1, K), key=k2) / K**0.5).astype(
+                mx.float32
+            )
+            w = (
+                mx.random.normal(
+                    (E, D, K) if transpose else (E, K, D), key=k3
+                )
+                / K**0.5
+            ).astype(mx.float32)
+            if mode == "affine":
+                qw, s, b = mx.quantize(w, group_size=group_size, mode=mode)
+            else:
+                qw, s = mx.quantize(w, mode=mode)
+                b = None
+            w = mx.dequantize(qw, s, b, group_size=group_size, mode=mode)
+            if transpose:
+                w = w.swapaxes(-1, -2)
+            return x, w, indices, qw, s, b, group_size
+
+        def warmup_gather_qmm(L, K, D, E, I, transpose, mode):
+            x, w, indices, qw, s, b, group_size = make_case(
+                L, K, D, E, I, transpose, mode
+            )
+            mx.eval(
+                mx.gather_mm(x, w, rhs_indices=indices),
+                mx.gather_qmm(
+                    x,
+                    qw,
+                    s,
+                    b,
+                    group_size=group_size,
+                    mode=mode,
+                    transpose=transpose,
+                    rhs_indices=indices,
+                ),
+            )
+
+        # Keep the quantized warmup sequence: the CPU NaN repro is stateful.
+        warmups = [
+            (32, 512, 512, 4, 2, True, "affine"),
+            (32, 512, 544, 4, 2, True, "mxfp4"),
+            (32, 512, 544, 4, 2, True, "nvfp4"),
+            (32, 512, 544, 4, 2, True, "mxfp8"),
+            (133, 512, 512, 4, 2, True, "affine"),
+            (133, 512, 555, 4, 2, True, "affine"),
+            (133, 512, 512, 4, 2, True, "affine"),
+            (64, 512, 512, 4, 2, False, "affine"),
+            (64, 512, 544, 4, 2, False, "mxfp4"),
+            (64, 512, 544, 4, 2, False, "nvfp4"),
+        ]
+        for params in warmups:
+            warmup_gather_qmm(*params)
+
+        x, w, indices, *_ = make_case(64, 512, 544, 4, 2, False, "mxfp8")
+        expected = x @ w[indices]
+        actual = mx.gather_mm(x, w, rhs_indices=indices)
+        self.assertTrue(mx.isfinite(actual).all().item())
+        self.assertTrue(mx.allclose(expected, actual, rtol=1e-5, atol=1e-5).item())
+
     def test_gather_mm_sorted_vjp(self):
         def gather_mm_ref(a, b, rhs):
             b = b[rhs]


angeloskath:
I can't reproduce the error. Could you provide a reproduction. I ran the test in the diff of course and it passes without the PR.

TyceHerrman:
I reproduced the failure locally with the full nixpkgs package build/test path rather than by running the added regression test in isolation.

I tested nixpkgs `python313Packages.mlx` on `aarch64-darwin` with this MLX checkout as `src`, so it uses the nixpkgs package recipe, including CPU-only MLX (`MLX_BUILD_METAL=false`) and the package pytest suite.

Base `upstream/main` at `2414e5df6a8cbc0d03f978d7212229f626f1f23b` failed. That SHA is the unpatched MLX main commit I used as the base comparison:

```text
SUBFAILED(L=64, K=512, D=544, E=4, I=2, transpose=False, mode='nvfp4') python/tests/test_quantized.py::TestQuantized::test_gather_qmm_sorted
AssertionError: array(nan, dtype=float32) not less than 1.5e-05
===== 1 failed, 671 passed, 40 skipped, 1 deselected in 136.77s (0:02:16) =====
```

This PR at `a01171bd3537b01b475c01d29affce324fd36093` passed the same nixpkgs package build/test:

```text
===== 672 passed, 40 skipped, 1 deselected in 129.63s (0:02:09) =====
```

```sh
nix build --impure -L --keep-failed \
  --argstr rev "$REV" \
  --out-link /private/tmp/mlx-result \
  -f /private/tmp/mlx-local-src-test.nix
```

with this local expression:

```nix
{ rev }:

let
  pkgs = import <nixpkgs> {};
  src = builtins.fetchGit {
    url = "<local-path>";
    inherit rev;
    allRefs = true;
  };
in
  pkgs.python313Packages.mlx.overridePythonAttrs (old: {
    inherit src;
  })
```

I think added regression test in isolation is not enough - running only that test can pass on unpatched `main`. The failure appears to be state/order/environment sensitive. The full nixpkgs package test run reproduces the existing `test_gather_qmm_sorted` NaN on base, and the same package build/test passes with this PR.

angeloskath:
I understand that but this doesn't mean that the error is in gather mm unfortunately. It could be an out of bounds write in random number generation or something equivalent. I think unless you can reproduce with static inputs we shouldn't merge this PR.

Sorry for that, it's just probably not where the error is.

TyceHerrman:
The fundamental issue appears to be an Accelerate `cblas_sgemm` `M == 1` edge case

In the failing `GatherMM` call, BLAS operands:

- `A`: selected left-hand row from `x`
- `B`: selected right-hand matrix from `w`
- `C`: output row buffer

MLX trace shows that `A` and `B` are finite, `A`/`B`/`C` do not overlap in memory, and `B` is unchanged by the call.

With temporary local instrumentation in `GatherMM`, I also ran a preflight CBLAS call immediately before the real one. That preflight used the same GatherMM call site, same selected x row (A), same selected w matrix (B), and same CBLAS arguments, but wrote into a fresh clean scratch output buffer instead of the real MLX output (C). That scratch call passed; the real production call failed only when the selected (C) output lane already contained `NaN`.

That matches the standalone Accelerate repro. Pure `cblas_sgemm` with `M=1,N=512,K=544`, `beta=0`, misaligned `B` (`mod64 != 0`), and `C` prefilled with `NaN` returns `NaN` in tail output columns. With `beta=0`, BLAS semantics say previous `C` contents should not affect the result.

So the fix could be to avoid this Accelerate `M == 1` SGEMM path in MLX. Could use dedicated one-row dot/GEMV-style path, rather than depending on `beta=0` to ignore stale output contents.

if that seems like the correct route to take, happy to work on that here or in new pr

booxter:
@TyceHerrman thanks for working on this. I have a question: is the included regression test consistently hitting the failure scenario for you? On any machine or just a specific one? 

So far we saw it in nixpkgs Hydra CI only and I could not consistently reproduce it locally on any machine I had **direct** access to. Which complicates debugging. If this regression test can indeed trigger the problem somewhat consistently, it's a big step forward even if the fix may not be complete. 

Thanks in advance for clarification.

TyceHerrman:
the committed regression on this PR passes reliably for me locally. I ran the unit test 100 times in fresh processes on a01171bd and it passed every time.

I ran the same test body as a temporary script against base `2414e5df` built CPU-only. That failed 99/100 fresh-process runs.

I also checked the underlying cause separately with an accelerate reproduction. `cblas_sgemm` with `M=1,N=512,K=544`, `beta=0`, misaligned `B`, and `C` prefilled with `NaN` failed 100/100 fresh-process runs, always at tail column `510`. each control test passed 100/100 individually: finite dirty `C`, zero `C`, aligned `B`, and `M=2`.

the failure needs the exact combo of M=1 + misaligned B + stale NaN in output C.

So seems the Python regression is highly reliable in local CPU-only setup but still state/layout-sensitive rather than perfectly deterministic. The accelerate case does make the lower-level failure condition deterministic in isolation.

lmk if there's other testing you'd like to see

zcbenz:
> That scratch call passed; the real production call failed only when the selected (C) output lane already contained `NaN`.

According to [the doc of cblas_sgemm ](https://developer.apple.com/documentation/accelerate/cblas_sgemm(_:_:_:_:_:_:_:_:_:_:_:_:_:_:)?language=objc) it computes `C←αAB + βC` so can you try if zeroing out `out` helps?

TyceHerrman:
Zeroing `out` before the `cblas_sgemm` call does avoid the failure for me locally

zcbenz:
A plausible explanation is that the Accelerate implementation (assuming that is the backend you are using for CPU) of `cblas_sgemm` does not ignore `C` when `β` is 0, so an uninitialized `C` that contains NaN would lead to wrong results, which I think is a bug. Would you mind making an independent code snippet that reproduces the behavior, which I can use to talk to whoever is responsible to check what would be the right fix?

TyceHerrman:
the CPU path I tested is using Accelerate on Darwin CPU-only build. The rebuilt `python/mlx/lib/libmlx.dylib` links Accelerate, and `nm` shows `cblas_sgemm` resolving from Accelerate.

snippet attached:

```sh
clang -O2 -Wall -Wextra -DACCELERATE_NEW_LAPACK repro.c -framework Accelerate -o repro
./repro
```

It reproduces consistently:

```text
M=1 + 64-byte-misaligned B + NaN C + beta=0 -> BAD, got NaN at col 511
```

Controls pass:

```text
C=zero: OK
C=finite dirty: OK
B aligned: OK
M=2: OK
100/100 runs: same result
```

this points to the Accelerate `cblas_sgemm` path observing or propagating `C` even though `beta=0`
[repro.c](https://github.com/user-attachments/files/29142934/repro.c)

zcbenz:
@TyceHerrman Thanks for the repro! I tested it on a few machines but could not reproduce, I wonder if this is related to macOS/Xcode version, can you paste the environment you are using?

TyceHerrman:
sure. don't know if all this is needed but:

- MacBook Air M1, arm64
- macOS 26.5.1 (25F80), Darwin 25.5.0
- Active developer directory: `/Library/Developer/CommandLineTools`
- `xcodebuild` reports CLT-only
- macOS SDK 26.5
- Apple clang 21.0.0 (`clang-2100.1.1.101`), target `arm64-apple-darwin25.5.0`
- CLT package `com.apple.pkg.CLTools_Executables` version `26.5.0.0.1777544298`
- Python 3.14.6 via mise
- CMake 4.3.3
- Running local MLX extension via `PYTHONPATH=python`: `python/mlx/core.cpython-314-darwin.so`

used CPU/no-Metal path with `CMAKE_ARGS=-DMLX_BUILD_METAL=OFF`

zcbenz:
Thanks your environment is fairly new so this seems to be a hardware specific issue, I will see if I can find a M1 to reproduce.

zcbenz:
We have verified that this is a M1-only bug of Accelerate and a fix is on the way, so this will be automatically fixed in the next macOS update. @TyceHerrman Are you good waiting for the fix in macOS, or do you prefer patching it in MLX (by memset out to 0 for M1)?

TyceHerrman:
Fix in next macos release is fine by me! Thanks for digging in to this

booxter:
@zcbenz Any idea which MacOS builds should carry the fix?

zcbenz:
Unfortunately completely no idea :(

PR #3566, opened 2026-05-19
Synchronize no-GPU cache eviction with CPU streams

dhiltgen:
## Proposed changes

Follow up to #3554 - while testing variations on my AVX2 branch I found this bug on that change.

The no-GPU CPU allocator cache could release cached buffers while asynchronous CPU work was still queued on a non-default stream. mlx-lm generation can hit this by queuing the next token on a CPU stream and calling clear_cache between yielded tokens, which can leave queued work referencing freed cached memory.

Synchronize CPU streams before evicting cached buffers from clear_cache or set_cache_limit, and protect get_cache_memory with the allocator mutex. Add a regression test that verifies clear_cache waits for queued CPU stream work before returning.

## Checklist

### DIFF
--- a/mlx/backend/no_gpu/allocator.cpp
+++ b/mlx/backend/no_gpu/allocator.cpp
@@ -6,6 +6,7 @@
 #include "mlx/allocator.h"
 #include "mlx/backend/common/buffer_cache.h"
 #include "mlx/memory.h"
+#include "mlx/stream.h"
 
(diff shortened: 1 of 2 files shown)


angeloskath (approved):
Looks great and makes total sense, thanks!

zcbenz (approved):
Nice fix!

PR #3572, opened 2026-05-21
Validate segmented_mm inner dimensions

### LINKED ISSUE #3571: segmented_mm should validate inner matrix dimensions
## Summary
`segmented_mm` accepts two 2D inputs documented as `MxK` and `KxN`, but it does not currently validate that the inner dimensions match before constructing the primitive.

Neighboring matmul-like APIs such as `matmul`, `addmm`, `block_masked_mm`, and `gather_mm` reject incompatible matrix dimensions at the API boundary.

## Example
```python
a = mx.ones((2, 3))
b = mx.ones((4, 5))
segments = mx.array([[0, 3]], dtype=mx.uint32)
mx.segmented_mm(a, b, segments)
```

This should raise a `ValueError` because `3 != 4`.

## Impact
Rejecting the mismatch early avoids constructing a segmented matmul with incompatible input shapes and makes `segmented_mm` consistent with the rest of the matmul API surface.

fallintoplace:
## Summary
- Add an inner-dimension check to `segmented_mm` before creating the primitive.
- Add a regression test for mismatched `MxK` and `KxN` inputs.

## Why
`segmented_mm` expects `a` to have shape `MxK` and `b` to have shape `KxN`, but it did not reject incompatible inner dimensions. This now matches the validation style used by neighboring matmul APIs.

## Validation
- `git diff --check HEAD~1..HEAD`
- `python3 -m py_compile python/tests/test_blas.py`

I could not run the full MLX Python test locally because this machine does not have the Metal compiler available through `xcrun -sdk macosx metal`.

Fixes #3571

### DIFF
--- a/mlx/ops.cpp
+++ b/mlx/ops.cpp
@@ -5895,6 +5895,14 @@ array segmented_mm(
     throw std::invalid_argument("[segmented_mm] Batched matmul not supported");
   }
 
+  if (a.shape(-1) != b.shape(-2)) {
+    std::ostringstream msg;
+    msg << "[segmented_mm] Last dimension of first input with shape "
+        << a.shape() << " must match first dimension of second input with "
+        << "shape " << b.shape() << ".";
+    throw std::invalid_argument(msg.str());
+  }
+
   if (segments.ndim() < 1 || segments.shape().back() != 2) {
     std::ostringstream msg;
     msg << "[segmented_mm] The segments should have shape (..., 2) but "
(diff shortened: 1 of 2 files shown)


zcbenz:
This is not how `segmented_mm` is supposed to work, see reference implementation for example:
https://github.com/ml-explore/mlx/blob/2d2d59e7720981f4f7c1ff845403b19b2593f7fc/python/tests/test_blas.py#L1295-L1300

Issue #3567, opened 2026-05-20
JIT compilation for CUTLASS/CuTe kernels


zcbenz:
Our [requests for increasing file limit on PyPi](https://github.com/pypi/support/issues/10359) was rejected so I think the only reasonable moving forward is to use JIT compilation for the CUTLASS/CuTe kernels.

PR #3576, opened 2026-05-21
[CUDA] JIT-compile qmm_naive

zcbenz:
Bundle the headers of CUTLASS and JIT-compile the qmm_naive kernels, which reduces the binary size (https://github.com/ml-explore/mlx/issues/3567) and is required for meeting the size limit of PyPI. Most of the changes are moving code to `backend/cuda/device` and reducing uses of advanced C++ features to make NVRTC happy.

An unfortunate side effect is `test_quantized.py` now takes half an hour to run, I will try if I can make some sub-tests run in parallel or do some proper caching in CI.


angeloskath (approved):
🙏

PR #3577, opened 2026-05-22
Add a minimal data-parallel distributed example (#2930)

adityasingh2400:
## Summary

Closes (partially) #2930 — awni asked for a few short, standalone `mlx.distributed` examples back in December 2025 and nothing landed in the months since. This PR starts small with a single data-parallel example so we can lock in the structure before adding more.

- `examples/distributed/data_parallel/main.py`: ~80-line MLP trained on a synthetic classification task. Each rank gets a different data shard; gradients are averaged with `nn.average_gradients` (which batches `mx.distributed.all_sum` calls).
- `examples/distributed/README.md`: how to run with `mlx.launch -n 2`, plus a pointer to the existing distributed docs.

The script also runs unchanged under plain `python` because `mx.distributed` ops are no-ops at world size 1, so it doubles as a smoke test.

If this structure looks good, I'm happy to send a follow-up PR with a tensor-parallel inference example (and any others you have in mind).

## Test plan
- [x] `python3 -m py_compile examples/distributed/data_parallel/main.py`
- [ ] `mlx.launch -n 2 examples/distributed/data_parallel/main.py` (maintainer-side, requires a built MLX with a distributed backend available)
- [ ] `python examples/distributed/data_parallel/main.py` (single process, should run with any MLX install)

### DIFF
--- a/examples/distributed/README.md
+++ b/examples/distributed/README.md
@@ -1,0 +1,28 @@
+# Distributed MLX examples
+
+Small, runnable examples of `mlx.distributed`. Each example is self-contained
+and uses synthetic data so there is nothing to download.
+
+## Data parallel training
+
+`data_parallel/main.py` trains a small MLP on a synthetic classification task.
+Each rank processes a different shard of the data and gradients are averaged
+across ranks after every step using `mlx.nn.average_gradients`, which batches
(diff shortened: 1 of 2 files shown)


zcbenz:
Closing since this is not even wrong.

PR #3578, opened 2026-05-22
Add new_thread_unsafe_stream API

zcbenz:
Add an API to let users create streams that can be freely passed to and evaluated on any thread, but as the API name implies MLX is not applying any synchronization around the streams and it is users' responsibility to ensure there is no race conditions.

This provides a way for language bindings (like Swift) to evaluate graph asynchronously via threads, on the condition that they add locks on themselves. Hopefully after making streams fully thread-safe in future and we would be able to remove this API, but it is going to be very tricky and might not come any time soon.

### DIFF
--- a/mlx/backend/cpu/encoder.cpp
+++ b/mlx/backend/cpu/encoder.cpp
@@ -6,20 +6,29 @@
 
 namespace mlx::core::cpu {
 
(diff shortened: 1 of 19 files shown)


angeloskath (approved):
Looks great!

Issue #3533, opened 2026-05-12
Slow 1D dot product `mx.inner` compared to custom kernel


SuriyaaMM:
Hi, the dot product (`mx.inner` is at least 3 times slower for 1d compared to this kernel implemented using `metal_kernel`).

```python
import mlx
import mlx.core as mx
import numpy as np
import time

with open("kernels/dot.msl") as f:
    source = f.read()

if source == None or source == "":
    raise ValueError("nothing to read in dot.msl")

with open("kernels/reduce_add.msl") as f:
    source2 = f.read()

if source2 == None or source2 == "":
    raise ValueError("nothing to read in reduce_add.msl")

dot_kernel = mx.fast.metal_kernel(
    name="dot",
    input_names=["a", "b"],
    output_names=["output"],
    source=source
)
reduce_add_kernel = mx.fast.metal_kernel(
    name="reduce_add",
    input_names=["a"],
    output_names=["output"],
    source=source2
)

def __dot(a: mx.array, b: mx.array) -> float:
    if a.shape != b.shape:
        raise ValueError("a and b must have same dimension")
    if len(a.shape) != 1:
        raise ValueError("dot is supported only for 1d array")
    
    n = a.size
    tg_size = 512
    ceil = (n + tg_size - 1) // tg_size
    
    output = dot_kernel(
        inputs=[a, b],
        template=[("F32", mx.float32), ("n", n)],
        grid=(n, 1, 1),
        threadgroup=(tg_size, 1, 1),
        output_shapes=[(ceil,)],
        output_dtypes=[mx.float32],
        init_value=0
    )

    while ceil != 1:
        partial = output[0]  
        n = partial.size
        ceil = (n + tg_size - 1) // tg_size

        output = reduce_add_kernel(
            inputs=[partial], 
            template=[("F32", mx.float32), ("n", n)],
            grid=(n, 1, 1),
            threadgroup=(tg_size, 1, 1),
            output_shapes=[(ceil,)],
            output_dtypes=[mx.float32],
            init_value=0
        )

    return output[0][0]  

def bench(fn, rounds=20, label=""):
    for _ in range(3):
        r = fn()
        mx.eval(r)

    times = []
    for _ in range(rounds):
        mx.eval()  
        t0 = time.perf_counter()
        r = fn()
        mx.eval(r) 
        times.append(time.perf_counter() - t0)

    times.sort()
    median = times[len(times) // 2]
    best = times[0]
    worst = times[-1]
[...]

PR #3580, opened 2026-05-22
Implement batched matmul for large 1D dot products

Ved235:
## Proposed changes

Addresses issue #3533. Adds routing logic in `mlx/ops.cpp` so that it divides the large 1D dot product into chunks so gemv parallelizes. 

### Benchmark
```python
import mlx.core as mx
import numpy as np
import time

def bench(fn, rounds=100, label=""):
    for _ in range(3):
        r = fn()
        mx.eval(r)

    times = []
    for _ in range(rounds):
        mx.eval()  
        t0 = time.perf_counter()
        r = fn()
        mx.eval(r) 
        times.append(time.perf_counter() - t0)

    times.sort()
    median = times[len(times) // 2]
    best = times[0]
    worst = times[-1]
    print(f"{label}")
    print(f"median={median*1000:.3f}ms | min={best*1000:.3f}ms | max={worst*1000:.3f}ms")
    return r

a = mx.random.normal(shape=(50_000_000,), dtype=mx.float32)
b = mx.random.normal(shape=(50_000_000,), dtype=mx.float32)

a_np = np.array(a, copy=False)
b_np = np.array(b, copy=False)

ccc = bench(lambda: mx.inner(a, b), label="MLX native")

print(f"mx.inner : {float(ccc)}")
```
Using this benchmarking script the performance changes are:
```
median=15.393ms | min=15.323ms | max=15.769ms
```
to
```
median=1.741ms | min=1.682ms | max=1.835ms
```
## Checklist


Ved235:
@zcbenz would it possible for you to review this?

mlx/ops.cpp
```diff
@@ -5406,6 +5428,18 @@ array tensordot(
     }
   }
 
+  auto stream = to_stream(s);
+  auto device = stream.device;
+  if (a.has_primitive()) {
+    device = a.primitive().stream().device;
+  } else if (b.has_primitive()) {
+    device = b.primitive().stream().device;
+  }
```
zcbenz: I think we don't need to check operands' device because we only need to ensure `inner_1d_gpu_chunked` run on gpu?

zcbenz (approved):
This basically looks good to me, thanks!

mlx/ops.cpp
Ved235: I have replaced this with
```cpp
  auto stream = to_stream(s);
  if (a.ndim() == 1 && b.ndim() == 1 && axes_a.size() == 1 &&
      axes_b.size() == 1 && stream.device == Device::gpu &&
      csize >= 32 * 4096) {
    return inner_1d_gpu_chunked(a, b, stream);
  }
```

Ved235:
@zcbenz I have made the changes could you review and merge.

zcbenz (approved):
I would like another review from maintainers before merging.

Ved235:
> I would like another review from maintainers before merging.

Ok sure, is there someone specific I should tag to request a review?

Ved235:
@angeloskath @jagrit06 it would be great if you could review these changes

Ved235:
@zcbenz Its been several weeks since the changes and I believe that the changes are not very large in terms of number of lines, so would it be possible for this to be merged?

zcbenz:
Sorry WWDC had largely disrupted our schedule, there is no wrong with this PR it is just I need another view on this since I lack the background knowledge. WWDC is over now and there is a large backlog so please give us more time.

Ved235:
No worries and thanks for the update.

angeloskath (dismissed):
@Ved235 sorry for the super late reply, especially since it will be a negative response.

This is something we need to fix indeed but unfortunately this is not the way to fix it. Basically the ops should not really change based on the shape but the implementation should. The same way that we route to split-k kernel when the matrix K dimension is large.

I am gonna mark this as requested changes so we don't merge it by accident and make a new PR with a specialized kernel for this particular case. I am not sure how important it is but I think the kernel will end up being simple enough.

Ved235:
Should I make a PR for this specialised kernel? I have a solution ready

Ved235:
@angeloskath I have implemented a specialised kernel. The performance benchmarks show the same improvement. Is this what you were expecting?

mlx/backend/metal/kernels/dot.metal
```diff
@@ -0,0 +1,123 @@
...
+
+  float c0 = 0.0f;
+  float c1 = 0.0f;
+  float c2 = 0.0f;
+  float c3 = 0.0f;
+
+  // Fast path: no per-element branches.
+  if (start + ITEMS_PER_THREAD <= n) {
+    MLX_MTL_PRAGMA_UNROLL
+    for (int i = 0; i < ITEMS_PER_THREAD; i += 4) {
```
angeloskath: It's better to do coalesced reads of size 16 / sizeof(T) rather the one below. So now thread 0 reads 0-31, thread 1 reads 32 - 63 and so on. Coalesced reads you can still process 32 items per thread (or however many) but you process 0-3, 128-131, ... etc.

mlx/backend/metal/kernels/dot.metal
```diff
@@ -0,0 +1,123 @@
...
+[[kernel]] void dot_reduce(
+    const device float* input [[buffer(0)]],
+    device T* output [[buffer(1)]],
+    const constant int& n [[buffer(2)]],
+    uint gid [[thread_position_in_grid]],
+    uint tid [[thread_position_in_threadgroup]],
+    uint lane [[thread_index_in_simdgroup]],
+    uint simd_id [[simdgroup_index_in_threadgroup]],
+    uint tg_id [[threadgroup_position_in_grid]]) {
+  float c = gid < uint(n) ? float(input[gid]) : 0.0f;
```
angeloskath: You can loop obviously no need to loop by launching many kernels and allocating many intermediates.

mlx/backend/metal/matmul.cpp
```diff
@@ -1039,6 +1039,68 @@ void steel_matmul_axpby(
...
+  compute_encoder.dispatch_threads(
+      MTL::Size(size_t(blocks) * thread_group_size, 1, 1),
+      MTL::Size(thread_group_size, 1, 1));
+
+  array current = partials;
+  kname = "dot_reduce_" + type_to_name(out);
+  auto kernel_final = d.get_kernel(kname);
+  auto kernel_intermediate = d.get_kernel("dot_reduce_float32");
+
+  while (blocks > 1) {
```
angeloskath: As mentioned above you don't need to launch many kernels, the kernel should loop.

mlx/backend/metal/matmul.cpp
```diff
@@ -1278,6 +1340,18 @@ void Matmul::eval_gpu(const std::vector<array>& inputs, array& out) {
   /////////////////////////////////////////////////////////////////////////////
   // Gemv specialization
 
+  if (M == 1 && N == 1 && K > 16384 && batch_size_out == 1 && !a_transposed &&
+      !b_transposed && a.flags().row_contiguous && b.flags().row_contiguous) {
```
angeloskath: So here we should quite significantly relax this.

For the case `M = N = 1` we care about row or col contiguous as it really means it 's a contiguous vector.

In addition, it should probably route to this kernel even for the case `K <= 16384` and the kernel shouldn't necessarily require a split-k approach ie a single thread group is almost certainly the best possible choice for vectors <= 4k and it could even be for 16k ones and so on.

Finally, I think this is still likely the best approach for small matrices ie `batch_size_out * M * N < 64`. This needs to be benchmarked of course.

We don't have to do all of the above from the get go but at least relaxing the `!a_transposed && !b_transposed && ...` checks so all dot products route to this would be the minimum imho.

angeloskath (commented):
It's nearly there.

I think the kernel needs a bit of restructuring but it is quite close to what is needed. I left comments inline.

mlx/backend/metal/matmul.cpp
Ved235: I am now launching one threadgroup of 512 threads (the partials array is significantly smaller as compared to the input arrays so 512 threads should be sufficient for a really large input)

mlx/backend/metal/kernels/dot.metal
Ved235: I have coalesced them according to the each SIMD group. I believe the compiler will automatically replace the scalar fetches with a vector wide instruction.

jeonghoon-ad:
Thanks for pushing this forward. I took a pass over the current specialized
kernel direction and ran a small baseline sweep on the PyPI MLX wheel.

I could not benchmark this PR branch directly yet because my local source build
fails after the Metal library is built (`dot.air` / `mlx.metallib` are produced,
then the build fails later in the C++/SDK setup). So please treat the comments
below as review notes rather than measured PR results.

Two things look worth checking before merge:

1. Dtype routing

The new `Matmul::eval_gpu` route does not appear to guard dtype, but
`dot.metal` only instantiates float32, float16, and bfloat16 kernels. Existing
Metal GEMV has complex64 support. If a contiguous complex64 `1 x K @ K x 1`
case reaches this path, it looks like it may try to load a missing
`dot_product_complex64` kernel. A dtype guard that falls back to GEMV for
complex64, or a complex64 kernel plus test, would make this safer.

2. Small-K regression risk

The current route sends all contiguous `M == 1 && N == 1 && batch_size_out == 1`
cases to the dot path. The implementation always launches `dot_product` and
then `dot_reduce`, even when `blocks == 1`, and also allocates a partials array.
That looks great for huge vectors, but it could regress small vectors where the
old GEMV path is already dominated by launch overhead.

On the current PyPI wheel, I see this baseline on GPU:

```
K=1          median 210 us
K=16         median 207 us
K=256        median 283 us
K=4,096      median 311 us
K=16,384     median 316 us
K=131,072    median 349 us
K=1,000,000  median 927 us
K=50,000,000 median 19.36 ms
```

So the large-vector problem is very real, but the small-K path probably needs
either a direct one-kernel `blocks == 1` case or a measured threshold.

A benchmark table over K and dtype would make the PR much easier to review:

- K: 1, 16, 256, 4096, 16384, 131072, 1M, 50M
- dtype: float32, float16, bfloat16, plus complex64 fallback/guard coverage
- shape cases: `mx.inner(1D, 1D)` and explicit `(1, K) @ (K, 1)`

The main thing to rule out is that broad routing trades the big-vector win for
small-vector or complex64 regressions.

Ved235:
I have added a dtype checking for complex64. Apart from that here is a benchmark table for this implementation:

Size | MLX native median (ms) | NumPy median (ms)
-- | -- | --
1 | 0.163 | 0.000
16 | 0.146 | 0.000
256 | 0.143 | 0.000
256 | 0.153 | 0.000
4096 | 0.151 | 0.001
16384 | 0.156 | 0.001
131072 | 0.144 | 0.003
1000000 | 0.188 | 0.011
50000000 | 1.755 | 2.611

I tried comparing this implementation with the old GEMV path and according to me the results are highly variable. If required I think a condition around `>= 500000 `would be good enough to separate small vectors and big enough vectors.

Ved235:
@angeloskath would it possible for you to review this?

Ved235:
@zcbenz A quick follow up on this, its been a month since the PR addressed the comments left by @angeloskath. There is nothing outstanding from my side. Is there any timeline for this to be implemented?

mlx/backend/metal/kernels/dot.metal
```diff
@@ -0,0 +1,125 @@
...
+[[kernel]] void dot_product(
+    const device T* a [[buffer(0)]],
+    const device T* b [[buffer(1)]],
+    device float* output [[buffer(2)]],
+    const constant int& n [[buffer(3)]],
+    uint tid [[thread_position_in_threadgroup]],
+    uint lane [[thread_index_in_simdgroup]],
+    uint simd_id [[simdgroup_index_in_threadgroup]],
+    uint tg_id [[threadgroup_position_in_grid]]) {
+  constexpr int ITEMS_PER_THREAD = 32;
```
zcbenz: static parameters like this are usually declared as template args and encoded in kernel names, see the `gemv_t` in `mlx/backend/metal/kernels/gemv.metal` and `mlx/backend/metal/kernels/gemv.h` for example.

mlx/backend/metal/matmul.cpp
```diff
@@ -1455,6 +1499,16 @@ void Matmul::eval_gpu(const std::vector<array>& inputs, array& out) {
...
+  if (M == 1 && N == 1 && batch_size_out == 1 && a.flags().row_contiguous &&
+      b.flags().row_contiguous && a.dtype() != complex64) {
+    return dot_product(
+        /* const Stream& s = */ s,
+        /* metal::Device& d = */ d,
+        /* const array& a = */ a,
+        /* const array& b = */ b,
+        /* array& out = */ out,
+        /* int K = */ K,
+        /* std::vector<array>& copies = */ copies);
```
zcbenz, line 1516: Seems like a `}` is missing?

mlx/backend/metal/kernels/dot.metal
```diff
@@ -0,0 +1,125 @@
...
+    c = smem[tid];
+    c = simd_sum(c);
+    if (tid == 0) {
+      output[tg_id] = c;
+    }
+  }
+}
+
+template <typename T>
+[[kernel]] void dot_reduce(
```
zcbenz: It seems that this is just a all reduce, can the existing `all_reduce_dispatch` be used?

mlx/backend/metal/kernels/dot.metal
```diff
@@ -0,0 +1,125 @@
...
+  float c = c0 + c1 + c2 + c3;
+  c = simd_sum(c);
+
+  if (lane == 0) {
+    smem[simd_id] = c;
+  }
+
+  threadgroup_barrier(mem_flags::mem_threadgroup);
+
+  if (tid < 16) {
```
zcbenz: Should avoid using hardcoded numbers like this.

mlx/backend/metal/kernels/dot.metal
```diff
@@ -0,0 +1,125 @@
...
+      }
+    } else {
+      MLX_MTL_PRAGMA_UNROLL
+      for (int j = 0; j < VEC; ++j) {
+        int nidx = idx + j;
+        if (nidx < n) {
+          float v = float(a[nidx]) * float(b[nidx]);
+          switch (j & 3) {
+            case 0:
+              c0 += v;
```
zcbenz: If you use `float4 c` then you can just do `c[j & 3] += v`.

mlx/backend/metal/kernels/dot.metal
```diff
@@ -0,0 +1,125 @@
...
+  MLX_MTL_PRAGMA_UNROLL
+  for (int i = 0; i < ITEMS_PER_THREAD; i += VEC) {
+    int idx = start + i * ITEMS_PER_THREAD;
+    if (idx + VEC <= n) {
+      MLX_MTL_PRAGMA_UNROLL
+      for (int j = 0; j < VEC; j += 4) {
+        c0 += float(a[idx + j + 0]) * float(b[idx + j + 0]);
+        c1 += float(a[idx + j + 1]) * float(b[idx + j + 1]);
+        c2 += float(a[idx + j + 2]) * float(b[idx + j + 2]);
+        c3 += float(a[idx + j + 3]) * float(b[idx + j + 3]);
```
zcbenz: It might worth reading a/b with something like `float4(reinterpret_cast<metal::vec<T, 4>*>(a)[j])`, not necessarily faster I'm just not sure if compiler would do vectorized read in current form.

zcbenz (commented):
Can you check whether existing tests actually covers the dot product? If not we would need new tests.

mlx/backend/metal/kernels/dot.metal
Ved235: Moved the dot kernel implementation into `dot.metal` and `dot.h`

mlx/backend/metal/kernels/dot.metal
Ved235: Removed dot_reduce kernel and instead am using:
```  array tempResult({1}, float32, nullptr, {});
  tempResult.set_data(allocator::malloc(tempResult.nbytes()));
  copies.push_back(tempResult);
  all_reduce_dispatch(partials, tempResult, "sum", compute_encoder, d, s);
  copy_gpu(tempResult, out, CopyType::Scalar, s);
```
 A temporary has to be used to cast the result to correct output type

mlx/backend/metal/kernels/dot.metal
Ved235: I have made these vectorized changes. There hasn't been a performance change at least from my benchmarks. The timings are hovering near the same results. Although I don't know if using vectorized operations can cause errors for unaligned addresses.

Ved235:
> Can you check whether existing tests actually covers the dot product? If not we would need new tests.

I have added few tests specifically for dot product under `python/tests/test_blas.py`

bench.py
zcbenz, line 1: Can you move this file to `benchmarks/python/dot_bench.py`?

Also for benchmarking, it would be more useful comparing with PyTorch, and printing the memory bandwidth of the kernel.

zcbenz (commented):
Thanks for updating the code, the kernel implementation looks good to me, but I still want to have another view from @angeloskath when he got time. It is a really busy year for us and there are some tasks having much higher priority than reviewing code.

And please also forgive us if we ask you to experiment different things and then revert back, it is just how kernel development works.

PR #3581, opened 2026-05-22
Fix typos in backend code comments

adityasingh2400:
Three small comment typos in the backend that I noticed while reading the code. None affect behavior.

In `mlx/backend/metal/kernels/fft.h` the Rader's algorithm comment said "decomposable into to prime factors", which should be "decomposable into prime factors". In `mlx/backend/cuda/eval.cpp` and `mlx/backend/cuda/allocator.cpp` the words "initalization" and "aquired" are misspellings of "initialization" and "acquired" respectively; both spellings are used correctly elsewhere in the repo.

Comment text only, no code or behavior changes.

### DIFF
--- a/mlx/backend/cuda/allocator.cpp
+++ b/mlx/backend/cuda/allocator.cpp
@@ -318,7 +318,7 @@ void CudaAllocator::move_to_unified_memory(
   buf.device = -1;
 }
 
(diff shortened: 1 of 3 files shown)


zcbenz (approved):
Thanks!

PR #3582, opened 2026-05-22
[metal] disable heap allocation for small buffers

Sonareal:
Metal heaps with ResourceHazardTrackingModeUntracked reuse memory as soon as a heap-allocated buffer is "freed" via reference count drop, even while the GPU may still be executing commands that reference it. This shows up as intermittent garbage in tensor results or hard crashes from MTLBuffer access-after-free.

The cheap fix: always go through device_->newBuffer() which gives a fresh allocation tracked by the resource pool. Performance impact on inference is negligible (allocator is already locked critical section); the win is correctness.

Carry-on patch for elejometa/speech-server. Upstream PR not (yet) filed — this branch is the source of truth for our mlx-swift fork's submodule.

## Proposed changes

## Checklist

### DIFF
--- a/mlx/backend/metal/allocator.cpp
+++ b/mlx/backend/metal/allocator.cpp
@@ -144,12 +144,10 @@ Buffer MetalAllocator::malloc(size_t size) {
       throw std::runtime_error(msg.str());
     }
     lk.unlock();
-    if (size < small_size_ && heap_) {
-      buf = heap_->newBuffer(size, resource_options);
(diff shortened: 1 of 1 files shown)


zcbenz:
The buffers are kept until GPU command is finished, whatever crash you saw were very unlikely caused by this. Please reference https://github.com/ml-explore/mlx/pull/3462 on how you could find out the root cause on your side.

PR #3587, opened 2026-05-24
Cache JIT-compiled CUDA kernels

zcbenz:
This reduces test run time to 7m, it is a must when we start JIT-compiling CUTLASS/CuTe kernels which would spend over an hour on running tests in CI.

The cache key is the hash of all files under `mlx/backend/cuda`, which is not 100% bullet-proof but should be robust enough to dodge most cache invalidation problems.


angeloskath (approved):
That's awesome!

PR #3307, opened 2026-03-24
Chunked full-attention SDPA for long key sequences

Thump604:
## Summary

Split key dimension into chunks when `kL >= 65536` to avoid GPU compute timeout on Apple Silicon. Each chunk runs the existing fused `steel_attention` kernel with `output_logsumexp=true`, then a reduction kernel combines chunks using logsumexp-weighted averaging.

- New `sdpa_full_self_attention_chunked()` dispatch function
- New `sdpa_chunked_reduce` Metal kernel for logsumexp-weighted chunk combination
- Threshold routing before NAX dispatch (chunked path uses non-NAX kernel)
- Configurable via `MLX_SDPA_CHUNK_THRESHOLD` (default 65536) and `MLX_SDPA_CHUNK_SIZE` (default 32768)

**Depends on:** #3306 (logsumexp output)

## Problem

The fused `steel_attention` kernel dispatches ALL key tokens in a single Metal compute dispatch. At ~65K+ key tokens, the dispatch exceeds the macOS GPU watchdog timeout and is killed by the `AGXMetal` driver.

**Evidence:**
- 65K keys: completes (55s on 4B model, 186s on 122B model)
- 128K keys: GPU watchdog kill on both 4B and 122B models
- Confirmed NOT out-of-memory: 4B model uses 3GB weights on 128GB machine — kill happens with 125GB free RAM

**Impact:** Models with 128K+ native context (Qwen3.5, Llama 3.1, etc.) cannot use their full context window on Apple Silicon with fused SDPA.

## How it works

```
kL < 65536  →  existing single-pass (unchanged)

kL >= 65536 →  chunked path:
  For each chunk (32K keys):
    steel_attention(Q, K_chunk, V_chunk, output_logsumexp=true)
    → normalized attention output + per-row logsumexp
  
  sdpa_chunked_reduce:
    max_lse = max(lse_1, ..., lse_N)
    out = Σ(exp(lse_c - max_lse) · out_c) / Σ(exp(lse_c - max_lse))
```

This is the standard FlashAttention-2 chunk reduction formula. Causal masking is correctly adjusted per chunk: `qL_off = (global_kL - qL) - k_start`. Sinks applied only to chunk 0.

## Results

**128K context on 122B model — was impossible, now works:**

| Context | Depth 10% | Depth 50% | Depth 90% | Accuracy |
|---------|-----------|-----------|-----------|----------|
| 128K | PASS (485s) | PASS (479s) | PASS (449s) | 100% |

[...]

PR #3306, opened 2026-03-24
Add logsumexp output to fused SDPA kernel

Thump604:
## Summary

Add `output_logsumexp` support to the fused `steel_attention` Metal kernel via function constant (304). When enabled, the kernel writes per-row logsumexp (`float32`) to buffer 8 alongside the normal normalized attention output. Zero overhead when disabled — the function constant eliminates the code path at compile time.

- Add `output_logsumexp` function constant (304) and `lse_out` buffer (8) to `steel_attention.h`
- Update pipeline cache hash to include logsumexp state
- Bind logsumexp output buffer in dispatch when requested
- Remove forced fallback to unfused SDPA when logsumexp is requested

**Motivation:** The fused SDPA kernel currently forces fallback to unfused SDPA when logsumexp output is needed (e.g., for training VJP). This achieves parity with the CUDA/cuDNN backend which already supports `set_generate_stats`. It is also a prerequisite for chunked SDPA dispatch (#3302) which uses logsumexp output to combine per-chunk results.

**Note:** This PR does NOT remove the `is_training` fallback — it only adds the kernel capability. Enabling fused SDPA for training VJP requires additional work (the VJP implementation itself).

## How it works

The `steel_attention` kernel already maintains per-row `max_score` and `sum_score` in registers for its online softmax. Logsumexp is derived from these existing values:

```
lse = max_score * M_LN2_F + log(sum_score)
```

The `M_LN2_F` conversion is needed because the kernel uses `exp2` internally (scores are pre-scaled by `M_LOG2E_F`). Uses `metal::precise::log` for numerical accuracy.

## Test plan

- [x] Logsumexp correctness tests across all dtypes (float16, bfloat16, float32)
- [x] All head dimensions (64, 80, 128, 256)
- [x] Causal attention, cross-attention (qL ≠ kL)
- [x] GQA (num_kv_heads ≠ num_heads)
- [x] Long context (8K)
- [x] Batched inputs
- [x] No regression on existing SDPA tests (`test_fast_sdpa.py`)
- [x] 53 tests pass, 664 subtests, 0 regressions

Refs: #3302

PR #3594, opened 2026-05-27
feat: enable logsumexp output in fused SDPA dispatch

adurham:
**Repo:** ml-explore/mlx
**Branch:** adurham:pr/sdpa-logsumexp-output
**Commits:** 3

---

## Acknowledgement

This PR revives [@Thump604](https://github.com/Thump604)'s work from [#3306](https://github.com/ml-explore/mlx/pull/3306), which was self-closed on 2026-04-04 ("Closing this because I'm narrowing active upstream work to vllm-mlx and I'm no longer in a position to maintain/rebase this PR properly. […] If I need this again, I'll reintroduce it from a fresh branch with current benchmarks/tests.").

It carries [@hnshah](https://github.com/hnshah)'s positive validation report from M3 Ultra forward, and adds production-cluster validation on M4 Max RDMA. Authorship and tests preserved from the original commits where possible.

---

## Summary

Add support for returning the log-sum-exp ("LSE") output from the fused full-attention path in `mx.fast.scaled_dot_product_attention`. The kernel template already produces it; this PR exposes it through:

1. A new `output_logsumexp` function constant on `steel_attention` (slot 304)
2. Removed forced fallback for `output_logsumexp` in `use_fallback()` so the fused path is taken when LSE is requested
3. Pipeline cache hash includes the new function constant
4. LSE output buffer bound on the host side
5. NAX path skipped when LSE is requested (NAX kernel doesn't support it)

This is a precondition for chunked SDPA dispatch (separate PR coming) where K/V is split along the sequence dimension and per-chunk outputs need to be merged using their LSE values to preserve numerical equivalence with the un-chunked computation.

## Motivation

For long-context inference (>65K key tokens), the fused full-attention kernel materializes a tile that risks exceeding Metal's threadgroup memory limits. The standard fix is to chunk K/V along the sequence axis and merge partial outputs. But merging is only mathematically correct if each chunk's LSE is preserved.

Without this PR, requesting LSE from the fused path silently routed to the (slower) fallback. This change makes the fused path return LSE directly, with no perf regression for the existing no-LSE case.

## Changes

- `mlx/backend/metal/kernels/steel/attn/kernels/steel_attention.h`: add `output_logsumexp` function constant (id 304), gate LSE store on it
- `mlx/backend/metal/scaled_dot_product_attention.cpp`:
  - Remove the `output_logsumexp` forced fallback
  - Include `output_logsumexp` in the pipeline cache hash
  - Allocate + bind LSE output buffer (buffer 8) when requested
  - Skip NAX path when LSE is needed
- `python/tests/test_sdpa_logsumexp.py`: 12 new tests covering vector vs fused LSE correctness across head_dims, dtypes, mask configurations

## Test Plan

- [x] `python -m pytest python/tests/test_sdpa_logsumexp.py` — 12/12 pass on M4 Max
- [x] `python -m pytest python/tests/test_fast_sdpa.py` — 15 pass, 1 skip (pre-existing)
- [x] Build clean on Apple Silicon M4 Max via `pip install -e .`
- [x] Originally validated by @hnshah on M3 Ultra 256GB (full LGTM report at https://github.com/ml-explore/mlx/pull/3306#issuecomment-2737293186)
- [x] Production end-to-end: 2× M4 Max RDMA cluster, DeepSeek-V4-Flash-8bit at 100K context — runs clean (this is the prerequisite for the chunked dispatch path that fixes our GPU-watchdog issue)

No perf regression measured for existing no-LSE callers (function constant defaults to false; the LSE store is hoisted under the new constant).

## Notes

This PR is a pure capability addition: existing callers see no behavior change. The LSE output is opt-in via the `output_logsumexp=True` parameter (existing API surface, previously routed to fallback).

A follow-up PR will use this to add **chunked SDPA dispatch** for sequences >65K keys (revival of @Thump604's #3307).

### DIFF
--- a/mlx/backend/metal/kernels/steel/attn/kernels/steel_attention.h
+++ b/mlx/backend/metal/kernels/steel/attn/kernels/steel_attention.h
@@ -14,6 +14,7 @@ constant bool align_K [[function_constant(201)]];
 constant bool has_mask [[function_constant(300)]];
 constant bool do_causal [[function_constant(301)]];
 constant bool has_sinks [[function_constant(302)]];
+constant bool output_logsumexp [[function_constant(304)]];
 
 struct MaxOp {
   template <typename T>
@@ -76,6 +77,7 @@ template <
     const constant AttnMaskParams* mask_params [[buffer(5), function_constant(has_mask)]],
     const device MaskType* mask [[buffer(6), function_constant(has_mask)]],
     const device T* sinks [[buffer(7), function_constant(has_sinks)]],
+    device float* lse_out [[buffer(8), function_constant(output_logsumexp)]],
     uint simd_lane_id [[thread_index_in_simdgroup]],
     uint simd_group_id [[simdgroup_index_in_threadgroup]],
     uint3 tid [[threadgroup_position_in_grid]],
@@ -472,5 +474,19 @@ template <
     Otile.template store_safe<T, 1, 1>(O, params->O_strides[2], dst_tile_dims);
   } else {
     Otile.template store<T, 1, 1>(O, params->O_strides[2]);
+  }
+
+  // Write per-row logsumexp if requested
+  if (output_logsumexp) {
+    STEEL_PRAGMA_UNROLL
+    for (short i = 0; i < kRowsPT; ++i) {
+      int row = int(tid.x) * BQ + tm + sm + (i * kFragSize);
+      if (row < params->qL) {
+        int64_t idx = int64_t(tid.z) * params->H * params->qL
+                    + int64_t(tid.y) * params->qL + row;
+        lse_out[idx] = float(max_score[i]) * M_LN2_F
+                     + metal::precise::log(float(sum_score[i]));
+      }
+    }
   }
 }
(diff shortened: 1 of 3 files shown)


zcbenz:
> ## Motivation
> For long-context inference (>65K key tokens), the fused full-attention kernel materializes a tile that risks exceeding Metal's threadgroup memory limits. The standard fix is to chunk K/V along the sequence axis and merge partial outputs. But merging is only mathematically correct if each chunk's LSE is preserved.
> 
> Without this PR, requesting LSE from the fused path silently routed to the (slower) fallback. This change makes the fused path return LSE directly, with no perf regression for the existing no-LSE case.

Outputting logsumexp is mostly used by backward in training, so the motivation does not makes sense for me. For VJP there is https://github.com/ml-explore/mlx/pull/3241 which includes the logsumexp output implementation.

adurham:
Makes sense — thanks for the pointer to #3241. I came at this from the forward-pass chunking angle (splitting K/V along the sequence axis for >65K-token decode and merging partial outputs via their LSE), but you're right that #3241 already lands the logsumexp output as part of the fused backward, which is the more general motivation. Once #3241 merges, the chunked-forward path I had stacked on this can just consume that LSE output instead of adding a separate code path. Closing-acknowledged; I'll revisit the chunking piece on top of #3241 if it's still needed after that lands.

PR #3595, opened 2026-05-27
perf(scheduler): MLX_STREAM_QOS env var to pin StreamThread QoS class (Apple-only)

adurham:
**Repo:** ml-explore/mlx
**Branch:** adurham:pr/mlx-stream-qos
**Commits:** 1 (`1540ecbc`)

---

# Title

`perf(scheduler): MLX_STREAM_QOS env var to pin StreamThread QoS class (Apple-only)`

# Body

## Summary

Add `MLX_STREAM_QOS` env var to optionally pin every `StreamThread` worker thread to a specified Apple QoS class via `pthread_set_qos_class_self_np()`.

Off by default (safe: misconfiguration could affect system scheduling). Apple-only — no-op on other platforms.

## Motivation

On a 2-node M4 Max RDMA cluster with concurrent foreground workloads, JACCL communication-stream worker threads sometimes hit asymmetric busy-poll stalls during all-reduce / all-gather. Diagnosed via `JACCL_TRACE_PROGRESS=1`:

- Rank 0's comm-stream worker: 1M-17M `poll_iters/call` (descheduled by macOS)
- Rank 1's comm-stream worker on the same call: 2 poll_iters

The C++ poll loop never re-enters Python, so the GIL isn't involved. macOS deprioritizes the worker thread because it inherits the parent's QoS class, which can drift during high system load.

Pinning to `USER_INTERACTIVE` keeps the thread on a P-core under contention. Empirically eliminates the stall asymmetry.

## Usage

```bash
MLX_STREAM_QOS=user_interactive  # max boost (recommended for latency-sensitive)
MLX_STREAM_QOS=user_initiated    # mild boost
MLX_STREAM_QOS=default           # explicit baseline
MLX_STREAM_QOS=utility           # demote (background work)
# (unset)                        # macOS default (off)
```

## Changes

- `mlx/scheduler.h`:
  - Add include for `<pthread/qos.h>` (Apple), `<cstdio>` `<cstdlib>` `<cstring>` for env parsing
  - Apply QoS pin to `StreamThread`'s `worker` thread on construction
  - Single function, ~35 lines

## Safety

- Gated on `__APPLE__` — no behavior change on Linux/Windows
- Gated on env var being set — no behavior change in default operation
- Unknown values are silently ignored (defaults preserved)
- The QoS pin is per-thread, applied only to MLX's StreamThread workers

## Test Plan

- [x] Build clean on Apple Silicon M4 Max
- [x] `MLX_STREAM_QOS=user_interactive python -c "import mlx.core as mx; mx.eval(mx.random.normal((4096,4096)) @ mx.random.normal((4096,4096)))"` — runs clean
- [x] Default (env unset): identical behavior to pre-PR
- [x] Production load: 2-node M4 Max cluster, 100K-context DeepSeek-V4 decode, `MLX_STREAM_QOS=user_interactive` over multi-hour benches — no regressions, eliminates the asymmetric stall pattern

## Notes

Linux equivalent (`sched_setattr(SCHED_DEADLINE)` or `pthread_setschedparam`) is out of scope for this PR — only `__APPLE__` path is implemented. Happy to add Linux support in a follow-up.

### DIFF
--- a/mlx/scheduler.h
+++ b/mlx/scheduler.h
@@ -3,11 +3,18 @@
 #pragma once
 
 #include <atomic>
+#include <cstdio>
+#include <cstdlib>
+#include <cstring>
 #include <future>
 #include <queue>
 #include <shared_mutex>
 #include <thread>
 #include <unordered_map>
+
+#if defined(__APPLE__)
+#include <pthread/qos.h>
+#endif
 
 #include "mlx/api.h"
 #include "mlx/backend/gpu/eval.h"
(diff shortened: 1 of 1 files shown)


mlx/scheduler.h
```diff
@@ -35,6 +42,34 @@ struct StreamThread {
   }
 
   void thread_fn() {
+#if defined(__APPLE__)
+    // exo-mlx-tune: env-gated QoS pin for stream worker threads.
+    // Diagnosed via JACCL_TRACE_PROGRESS=1: rank 0 (the API/master
+    // host) sees asymmetric busy-poll stalls (17M+ poll_iters)
```
zcbenz, line 48: The thread doing busy 17M+ iterations means it is not being descheduled?

adurham:
You're right to push on this, and on reflection I'm going to close it.

The 17M+ poll-iteration observation means the thread *was* getting scheduled — it wasn't starved, so "descheduled under QoS pressure" was the wrong diagnosis. That conclusion came out of a debugging session for an unrelated cross-rank bistability issue on our cluster, and I carried the QoS theory over here without independently confirming the mechanism. The `MLX_STREAM_QOS` pin "helped" in our setup, but I can't actually prove *why* with the evidence I have, and shipping an env var into MLX on the strength of an unverified mechanism isn't the right call.

If I can later isolate a reproducible scheduling effect with a clean before/after (e.g. via `os_signpost` / Instruments showing the worker actually losing the core to a competing thread), I'll reopen with real data. Closing for now — thanks for the sanity check.

PR #3597, opened 2026-05-28
Add metal::set_metallib_path() to override the metallib search path

olilarkin:
## What

Adds `mlx::core::metal::set_metallib_path(const std::string&)`, allowing a caller to specify an explicit path to `mlx.metallib` before the Metal device initializes. When set, `load_default_library()` tries this path first and falls back to the existing search logic if it fails to load.

## Why

When MLX is embedded inside a plugin bundle (e.g. Audio Unit / VST3), the *host application* — not MLX — owns the main bundle, so the default colocated / SwiftPM / bundle search paths don't resolve `mlx.metallib` and Metal init fails. There is currently no public way to point MLX at the library.

## Notes

- The override is only used if the library loads successfully; otherwise behavior is unchanged.
- Must be called before any MLX operation (before the device is created).
- No change to default behavior when the override is unset.

## Companion PRs

This is the root of a three-PR chain (C++ → C → Swift):

1. **ml-explore/mlx#3597** — this PR (C++ `set_metallib_path`)
2. ml-explore/mlx-c#117 — C API wrapper
3. ml-explore/mlx-swift#416 — Swift `GPU.setMetallibPath()`

Tracking issue: ml-explore/mlx-swift#415.

### DIFF
--- a/mlx/backend/metal/device.cpp
+++ b/mlx/backend/metal/device.cpp
@@ -30,12 +30,6 @@ struct hash<NS::SharedPtr<T>> {
 
 namespace mlx::core::metal {
 
-static std::string g_metallib_override_path;
-
-void set_metallib_path(const std::string& path) {
-  g_metallib_override_path = path;
-}
-
 namespace {
 
 constexpr const char* default_mtllib_path = METAL_PATH;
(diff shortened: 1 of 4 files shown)


mlx/backend/metal/device.cpp
```diff
@@ -15,19 +13,13 @@
...
-    return std::hash<T*>{}(p.get());
-  }
-};
+namespace mlx::core::metal {
 
-} // namespace std
+static std::string g_metallib_override_path;
 
-namespace mlx::core::metal {
+void set_metallib_path(const std::string& path) {
```
zcbenz: Can you move this to `mlx/backend/metal/metal.cpp`, and add a stub in `mlx/backend/metal/no_metal.cpp`? Probably also needs to add a `get_metallib_path`.

mlx/backend/metal/device.cpp
```diff
@@ -15,19 +13,13 @@
...
-template <class T>
-struct hash<NS::SharedPtr<T>> {
-  size_t operator()(const NS::SharedPtr<T>& p) const {
-    return std::hash<T*>{}(p.get());
-  }
-};
+namespace mlx::core::metal {
 
-} // namespace std
+static std::string g_metallib_override_path;
```
zcbenz: The globals are usually put in anonymous namespace rather than using `static`.

zcbenz (commented):
I think this is a good API to have, better than adding numerous fallbacks for searching.

zcbenz (approved):

PR #3600, opened 2026-05-28
block_fp8: 2D-block FP8 quantized matmul + MoE kernels (DeepSeek-V3 / MiMo)

yohann-bearzi:
# block_fp8: 2D-block FP8 quantized matmul + MoE kernels for DeepSeek-V3 / MiMo

## Summary

Adds first-class support for the **block_fp8** quantization format used by
DeepSeek-V3 / MiMo (and other recent MoE LLMs): 2D 128×128 weight blocks of
unpacked E4M3 (`F8_E4M3`) codes with a per-block `float32` scale. Implements
the kernel set needed to run inference end-to-end (decode + prefill, dense +
MoE gather), along with `quantize` / `dequantize` ops so the format is
producible and consumable in-library.

Validated on the MiMo-V2.5 base (310 B, 48 layers, 256 experts top-8, hybrid
5:1 SWA/GA, head_dim 192/128) on M3 Ultra: **30.3 tok/s decode**, with the
canonical greedy continuation reproduced bit-for-bit (top-1 = 151667 on the
Einstein verification prompt).

### Kernels (block_fp8 quantization mode)

* `block_fp8_qmv_fast` — M = 1 decode vector kernel.
* `block_fp8_qmm_t` — tiled prefill matmul using a new
  `BlockFp8QuantizedLoader` (handles unpacked E4M3 + 2D `F32` block scales).
* `block_fp8_qmm_t_splitk` — short-M splitk variant for shapes where the
  M·N tile count can't fill the GPU (covers the verify-style 4-token
  forwards used by speculative decoding).
* `block_fp8_gather_qmv_fast` — MoE decode vector kernel (top-K routing,
  single token).
* `block_fp8_gather_qmm_t` / `block_fp8_gather_qmm_rhs` — MoE prefill
  tiled kernels (the `rhs` variant matches the SwitchGLU global-sort
  contract used by `mlx_lm`).

### Dispatcher / plumbing

* `block_fp8` mode threaded through `quantized_matmul`, `gather_qmm`,
  `dequantize`, and the kernel-name selection (`quantized.cpp`).
* 2D-block-aware shape validator (`validate_quantized_input`) that accepts
  the 2D `F32` block-scale layout and the fused-QKV padding convention
  (scale tensor may include trailing rows beyond `ceil(w / 128)`, which the
  kernel never reads).
* `_skip_init` plumbing for `QuantizedLinear` / `QuantizedSwitchLinear` so
  pre-quantized weights can be installed without re-allocating placeholder
  buffers.

### SDPA

* `sdpa`: the fused vector kernel now handles asymmetric Q/V `head_dim`
  (192/128) used by MiMo's SWA blocks. Previously gated by an equality
  check that excluded this shape.

### `quantize` / `dequantize` ops (this commit's diff)

Prior to this PR, `mx.quantize` / `mx.dequantize` with `mode="block_fp8"`
inherited the 1D-group / E8M0-scale fallback from `mxfp8`, which does not
match the 2D 128×128 / `F32`-scale format the kernels consume; `dequantize`
additionally rejected `uint8` weights via a hardcoded `uint32` guard
unrelated to the mode validator.

Three minimal edits in `mlx/ops.cpp`:

* `fp_quantize`: adds a `block_fp8` branch performing per-block round-to-
  nearest quantization (`scale = max(|W_block|) / 448`, codes via
  `to_fp8(W/scale)`). Returns `uint8` codes `[..., N, K]` plus `float32`
  scales `[..., N/gs, K/gs]`, identical to the on-disk format the kernels
  already consume.
* `fp_dequantize`: adds a `block_fp8` branch that reconstructs the weight
  as `from_fp8(codes) * scale` with the per-block scale broadcast over the
  block grid.
* `dequantize`: the post-validator weight-dtype guard is made mode-aware
  (`uint8` for block_fp8, `uint32` otherwise). The behavior of every
  non-block_fp8 mode is byte-identical.

Forward-quantize uses only `to_fp8` and shape ops already present in
`ops.cpp` — no new Metal kernel is required (the format is producible from
the existing primitives).

### Kernel coverage test (new)

`python/tests/test_quantized.py::TestQuantized::test_block_fp8_kernel_coverage`
is a single test with one subtest per kernel; a regression in any kernel
fails the corresponding subTest:

| subTest | exercises |
| --- | --- |
| `block_fp8_qmv_fast` | M = 1 decode vector |
| `block_fp8_qmm_t` | M = 64 tiled prefill |
| `block_fp8_qmm_t_splitk` | M = 4, K = 4096 short-M splitk path |
| `block_fp8_qmm_t_large` | M = 256 large tiled |
| `block_fp8_gather_qmv_fast` | T = 1, KP = 4 MoE decode |
| `block_fp8_gather_qmm_rhs` | T = 32, KP = 4 MoE prefill, global-sort contract |
| `sdpa_fused_asymmetric_192_128` | fused vector SDPA at MiMo head dims |
| `padded_scale_validator` | scales tensor with trailing rows beyond `ceil(w/128)` |

All subtests compare against `mx.dequantize`-then-reference-matmul or the
unfused SDPA reference; threshold `cos > 0.99`.

### Round-trip / idempotency

* `test_block_fp8_quantize_dequantize` — quantize/dequantize round-trip
  across several `(N, K)` shapes.
* `test_block_fp8_quantize_idempotent` — re-quantizing a dequantized tensor
  reproduces identical codes (with scales matching to float-rounding
  tolerance). This is the stability property iterative quantizers (e.g.
  GPTQ) require to converge.
* `test_block_fp8_outlier_block_scaling` — a single outlier block must not
  corrupt the precision of unrelated blocks (the rationale for per-block
  vs per-tensor scaling).
* `test_block_fp8_invalid_shapes` — non-128-multiple dims rejected by
  `quantize`; `uint32` weights rejected by `dequantize(block_fp8)`.

### Numerics on real weights

`quantized_matmul(codes, scales)` agrees with
`x @ dequantize(codes, scales).T` at `cos > 0.99999` on real MiMo
projections, across the decode (M = 1), prefill (M = 256), and MoE gather
paths. The numpy-RTN E4M3 reference (computed offline) matches the
in-library `quantize` codes 100 % on the same inputs.

### End-to-end model

On the MiMo-V2.5 (310 B, block_fp8) checkpoint, M3 Ultra 512 GB, wired
memory limit 300 GB:

* **decode 30.3 tok/s** (single-stream, batch = 1, no speculative decoding)
* **prefill** ≈ 525 tok/s at L = 2048
* greedy continuation reproduces the bf16-reference continuation token-for-
  token on the Einstein verification prompt (top-1 = 151667 at position 29,
  with the expected continuation `[785, 785, 66622, 54052, 320, 23, 22,
  24, 4142, …]`).

### No regressions in existing modes

* Existing `affine` / `mxfp4` / `mxfp8` / `nvfp4` quantize/dequantize paths
  are byte-identical (the `block_fp8` branches are entered first via mode
  comparison and `return` before touching the existing code).
* Pre-existing test failures in `test_quantized.py` (e.g. `test_qvm` /
  `test_qmm` at `bits=6`) are independent of this PR; they reproduce on
  the merge-base without these changes, on M3 Ultra (`arch_gen=15`).
  Likely M4+/NAX-gated paths that need an arch decorator in a separate
  commit.

## Hardware notes (M3 Ultra)

NAX is `applegpu_g17+` (M4+ on macOS 26.2+); on M3 Ultra `is_nax_available()`
returns false, and this PR's dispatcher explicitly skips NAX routing for
block_fp8. The MLX_TRACE_KERNELS env var (added earlier in the branch) was
useful during development for confirming kernel selection at runtime.

### DIFF
--- a/mlx/backend/metal/kernels/fp_quantized.h
+++ b/mlx/backend/metal/kernels/fp_quantized.h
@@ -1409,6 +1409,98 @@ template <
       w, scales, x, y, Xs, Ws, K, N, M, tid, lid, simd_gid, simd_lid);
 }
 
+// --------------------------------------------------------------------------
+// block_fp8 gather_qmv (MoE decode path).
+// One threadgroup per (token, expert). Per-token rhs_indices selects the
+// expert slice of w/scales for this token's top-k experts. Calls the same
+// block_fp8_qmv_fast_impl after offset-adjusting the pointers.
+// --------------------------------------------------------------------------
+
+template <typename T>
+METAL_FUNC void block_fp8_adjust_matrix_offsets(
+    const device T*& x,
+    const device uint8_t*& w,
+    const device float*& scales,
+    const device uint32_t* lhs_indices,
+    const device uint32_t* rhs_indices,
+    device T*& y,
+    int output_stride,
+    const constant int& batch_ndims,
+    const constant int* batch_shape,
+    const constant int64_t* lhs_strides,
+    const constant int64_t* rhs_strides,
+    const constant int& x_batch_ndims,
+    const constant int* x_shape,
+    const constant int64_t* x_strides,
+    const constant int& w_batch_ndims,
+    const constant int* w_shape,
+    const constant int64_t* w_strides,
+    const constant int64_t* s_strides,
+    uint3 tid) {
+  uint32_t x_idx;
+  uint32_t w_idx;
+  if (batch_ndims == 1) {
+    x_idx = lhs_indices[tid.z * lhs_strides[0]];
+    w_idx = rhs_indices[tid.z * rhs_strides[0]];
+  } else {
+    ulong2 idx = elem_to_loc_broadcast(
+        tid.z, batch_shape, lhs_strides, rhs_strides, batch_ndims);
+    x_idx = lhs_indices[idx.x];
+    w_idx = rhs_indices[idx.y];
+  }
+  if (x_batch_ndims == 1) {
+    x += x_idx * x_strides[0];
+  } else {
+    x += elem_to_loc(x_idx, x_shape, x_strides, x_batch_ndims);
+  }
+  if (w_batch_ndims == 1) {
+    w += w_idx * w_strides[0];
+    scales += w_idx * s_strides[0];
+  } else {
+    ulong2 idx = elem_to_loc_broadcast(
+        w_idx, w_shape, w_strides, s_strides, w_batch_ndims);
+    w += idx.x;
+    scales += idx.y;
+  }
+  y += tid.z * output_stride;
+}
+
+template <typename T, int group_size, int bits>
+[[kernel]] void block_fp8_gather_qmv_fast(
+    const device uint8_t* w,
+    const device float* scales,
+    const device T* x,
+    const device uint32_t* lhs_indices,
+    const device uint32_t* rhs_indices,
+    device T* y,
+    const constant int& in_vec_size,
+    const constant int& out_vec_size,
+    const constant int& x_batch_ndims,
+    const constant int* x_shape,
+    const constant int64_t* x_strides,
+    const constant int& w_batch_ndims,
+    const constant int* w_shape,
+    const constant int64_t* w_strides,
+    const constant int64_t* s_strides,
+    const constant int& batch_ndims,
+    const constant int* batch_shape,
+    const constant int64_t* lhs_strides,
+    const constant int64_t* rhs_strides,
+    uint3 tid [[threadgroup_position_in_grid]],
+    uint simd_gid [[simdgroup_index_in_threadgroup]],
+    uint simd_lid [[thread_index_in_simdgroup]]) {
+  int M = x_shape[x_batch_ndims];
+  block_fp8_adjust_matrix_offsets(
+      x, w, scales, lhs_indices, rhs_indices, y,
+      out_vec_size * M,
+      batch_ndims, batch_shape, lhs_strides, rhs_strides,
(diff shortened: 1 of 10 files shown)


yohann-bearzi:
For anyone who wants to try this end-to-end on a real model, I've published a ready-to-run MiMo-V2.5 checkpoint:

**https://huggingface.co/bearzi/MiMo-V2.5-MLX**

It's the text-only path of [XiaomiMiMo/MiMo-V2.5](https://huggingface.co/XiaomiMiMo/MiMo-V2.5) (310B MoE, 256 experts top-8, hybrid 5:1 SWA/GA, block_fp8) repacked for MLX. Weight values are bit-identical to upstream — the only transform is concatenating the 32 source shards and stacking the per-expert MoE tensors along a new leading axis (required by MLX MoE loaders). The repo also ships the MLX model class (`mimo_v2_block_fp8.py`) and the converter script, so anyone with Xiaomi's release can reproduce the file in ~10 minutes.

Note that running it end-to-end also requires the unmerged [mlx-lm#1219](https://github.com/ml-explore/mlx-lm/pull/1219) (base `mimo_v2` model class) — the README in the HF repo documents the full prerequisite chain.

zcbenz:
Thanks for the pull request! Unfortunately due to maintenance burden we do not plan to add built-in GPU kernel for every possible op, for cases like this we suggest using [custom extensions](https://ml-explore.github.io/mlx/build/html/dev/extensions.html) and [custom kernels](https://ml-explore.github.io/mlx/build/html/dev/custom_metal_kernels.html) instead.

PR #3603, opened 2026-05-28
Fix CUDA all-reduce planning for large inputs

### LINKED ISSUE #3602: [BUG]  All_reduce over >INT_MAX elements hits RuntimeError
**Describe the bug**
CUDA global reductions fail when the input has more than `INT_MAX` elements.

**To Reproduce**
```python
import mlx.core as mx
a = mx.ones((32768, 32768, 2), dtype=mx.float16)
mx.eval(mx.sum(a))
```
This fails with:
`RuntimeError: cudaGraphAddKernelNode(&node, graph_, NULL, 0, &params) failed: invalid argument`

**Expected behavior**
This reduction should complete and return a scalar.

OS Version: Ubuntu 22.04.5 LTS
Version: mlx 0.31.1, mlx-cuda-13 0.31.1
GPU: NVIDIA RTX 6000 Ada Generation, driver 580.126.09

**Additional context**
32768 * 32768 * 2 == 2**31, which is one element past INT_MAX. Other reductions such as prod, min, max, all, and any hit the same issue.

sofinvalery:
Fixes #3602.
Not sure if I should add a test.

### DIFF
--- a/mlx/backend/cuda/reduce/all_reduce.cu
+++ b/mlx/backend/cuda/reduce/all_reduce.cu
@@ -68,12 +68,12 @@ void all_reduce(
 
   out.set_data(cu::malloc_async(out.nbytes(), encoder));
 
-  auto get_args = [](int size, int N) {
-    int threads = std::min(512, (size + N - 1) / N);
+  auto get_args = [](size_t size, size_t N) {
+    int threads =
+        static_cast<int>(std::min(size_t{512}, cuda::ceil_div(size, N)));
     threads = ((threads + WARP_SIZE - 1) / WARP_SIZE) * WARP_SIZE;
-    int reductions_per_step = threads * N;
-    size_t steps_needed =
-        (size + reductions_per_step - 1) / reductions_per_step;
+    size_t reductions_per_step = threads * N;
+    size_t steps_needed = cuda::ceil_div(size, reductions_per_step);
 
     int blocks;
     if (steps_needed < 32) {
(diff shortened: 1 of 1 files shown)


mlx/backend/cuda/reduce/all_reduce.cu
```diff
@@ -68,12 +68,13 @@ void all_reduce(
...
+  auto get_args = [](size_t size, size_t N) {
+    auto ceil_div = [](size_t x, size_t y) { return x / y + (x % y != 0); };
+
+    int threads = static_cast<int>(std::min(size_t{512}, ceil_div(size, N)));
     threads = ((threads + WARP_SIZE - 1) / WARP_SIZE) * WARP_SIZE;
-    int reductions_per_step = threads * N;
-    size_t steps_needed =
-        (size + reductions_per_step - 1) / reductions_per_step;
+    size_t reductions_per_step = threads * N;
+    size_t steps_needed = ceil_div(size, reductions_per_step);
```
zcbenz: Can `cuda::ceil_div` be used here?

zcbenz (approved):
Thanks for fixing this! I'm good without a test since it is going to require allocating a very large buffer.

mlx/backend/cuda/reduce/all_reduce.cu
sofinvalery: Oh, of course. Fixed it.

PR #3605, opened 2026-05-29
Fix complex VJPs for log and exp

CameronChurchwell:
## Fix complex VJPs for log and exp

This is the same as #3433, just with the `exp` and `log` functions now. Specifically, the VJP implementations for these primitives were lacking a conjugation which is necessary for correctness under the Wirtinger convention used by MLX. The change to `Log::vjp` looks more extensive since the previous implementation re-uses `Log::jvp`, but I have effectively copied over and modified the body of `Log::jvp`, so the actual change is just the addition of the conjugation

I have added tests to verify correctness, and I ran pre-commit already. No documentation changes should be necessary, as this is just a correctness fix.

## Checklist

### DIFF
--- a/mlx/primitives.cpp
+++ b/mlx/primitives.cpp
@@ -1992,7 +1992,7 @@ std::vector<array> Exp::vjp(
     const std::vector<array>& cotangents,
     const std::vector<int>& argnums,
     const std::vector<array>& outputs) {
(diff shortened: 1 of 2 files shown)


zcbenz (approved):
Looks good to me, thanks!

PR #2862, opened 2025-12-03
[Feature] Additional padding modes

sidtuladhar:
## Proposed changes

Hi team, this PR adds `symmetric` and `reflect` padding modes. It addresses issue #2155 

## Checklist

Put an `x` in the boxes that apply.

- [x] I have read the [CONTRIBUTING](https://github.com/ml-explore/mlx/blob/main/CONTRIBUTING.md) document
- [x] I have run `pre-commit run --all-files` to format my code / installed pre-commit prior to committing changes
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have updated the necessary documentation (if needed)

PR #3608, opened 2026-05-30
Add reflect and symmetric padding modes to mx.pad

katlun-lgtm:
## Summary

Adds `numpy.pad`-compatible **`\"reflect\"`** and **`\"symmetric\"`** modes to `mx.pad`. These join the existing `\"constant\"` and `\"edge\"` modes.

Both modes match `numpy.pad` semantics for **arbitrary pad sizes** — when the pad width exceeds the axis length the reflection repeats, exactly as NumPy does. (Earlier attempts at these modes were limited to `pad < dim`; this implementation removes that restriction.)

- **`reflect`** — mirror padding that does *not* repeat the edge value (period `2(n-1)`).
- **`symmetric`** — mirror padding that *does* repeat the edge value (period `2n`).

## Motivation

This was developed as part of porting [**mobileportrait-mlx**](https://github.com/katlun-lgtm/mobileportrait-mlx) (a TPS-based face-animation model) to pure MLX for training and inference on Apple Silicon. The full pipeline — CelebV-HQ frame extraction, keypoint cache, TPS warping, perceptual loss — trains end-to-end in MLX on an M3 Max. After training on 3,000 clips the model runs inference at **22.6 fps** (256×256, dense-motion + inpainting stage, M3 Max) and reaches **render L1 = 0.0566** on held-out clips.

## Implementation

`reflect_pad` (in `mlx/ops.cpp`) builds a per-axis index map with a triangle-wave reflection function and gathers with `take` — one `take` per padded axis. A degenerate axis of length 1 maps every coordinate to 0. No new primitive or kernel is introduced; it composes existing ops, so it works on every backend and is differentiable for free.

## Files changed

- `mlx/ops.cpp` — `reflect_pad` helper + `reflect` / `symmetric` dispatch branches in `pad`.
- `python/src/ops.cpp` — extend the `mode` `Literal` and docstring.
- `python/tests/test_ops.py` — `test_pad_reflect_symmetric`.
- `tests/ops_tests.cpp` — reflect/symmetric `CHECK` cases incl. multi-reflect.

## Testing

All run locally on an M3 Max:

- **Python** `test_pad_reflect_symmetric` — 13 shape/pad-width cases × 2 modes compared element-for-element against `numpy.pad` (in-bounds, multi-reflect where pad ≫ axis, asymmetric per-axis, zero-width sides, and degenerate axes `n==1`, `n==2`). Exact match.
- **C++** `tests/ops_tests.cpp` \"test pad\" — 9 assertions pass.
- **Full C++ suite** — `251 cases / 251 passed`, `3442 assertions / 0 failed`.

```python
>>> import mlx.core as mx
>>> a = mx.array([1, 2, 3])
>>> mx.pad(a, 2, mode=\"reflect\")
array([3, 2, 1, 2, 3, 2, 1], dtype=int32)
>>> mx.pad(a, 2, mode=\"symmetric\")
array([2, 1, 1, 2, 3, 3, 2], dtype=int32)
```

### DIFF
--- a/ACKNOWLEDGMENTS.md
+++ b/ACKNOWLEDGMENTS.md
@@ -20,6 +20,7 @@ MLX was developed with contributions from the following individuals:
 - Paul Paczuski: Improved stability of BCE loss calculation
 - Max-Heinrich Laves: Added `conv_transpose1d`, `conv_transpose2d`, and `conv_transpose3d` ops.
 - Gökdeniz Gülmez: Added the `Muon (MomentUm Orthogonalized by Newton-schulz)` optimizer, and the `ReLU²` activation function.
+- katlun-lgtm: Added `reflect` and `symmetric` padding modes.
 
 <a href="https://github.com/ml-explore/mlx/graphs/contributors">
   <img class="dark-light" src="https://contrib.rocks/image?repo=ml-explore/mlx&anon=0&columns=20&max=100&r=true" />
--- a/mlx/ops.cpp
+++ b/mlx/ops.cpp
@@ -1468,9 +1468,9 @@ array reflect_pad(
     bool include_edge,
     StreamOrDevice s /* = {} */) {
   // Reflect (include_edge=false) or symmetric (include_edge=true) padding.
-  // Matches numpy.pad for arbitrary pad sizes (the reflection repeats as needed).
-  // For an out-of-range coordinate r (relative to the original axis [0, n)),
-  // map it back into [0, n) by reflection:
+  // Matches numpy.pad for arbitrary pad sizes (the reflection repeats as
+  // needed). For an out-of-range coordinate r (relative to the original axis
+  // [0, n)), map it back into [0, n) by reflection:
   //   reflect   -> period 2(n-1), edge NOT repeated
   //   symmetric -> period 2n,     edge repeated
   auto reflect_coord = [](int r, int n, bool include_edge) -> int {
--- a/mlx/ops.cpp
+++ b/mlx/ops.cpp
@@ -1435,9 +1435,9 @@ array reflect_pad(
     bool include_edge,
     StreamOrDevice s /* = {} */) {
   // Reflect (include_edge=false) or symmetric (include_edge=true) padding.
-  // Matches numpy.pad for arbitrary pad sizes (the reflection repeats as needed).
-  // For an out-of-range coordinate r (relative to the original axis [0, n)),
-  // map it back into [0, n) by reflection:
+  // Matches numpy.pad for arbitrary pad sizes (the reflection repeats as
+  // needed). For an out-of-range coordinate r (relative to the original axis
+  // [0, n)), map it back into [0, n) by reflection:
   //   reflect   -> period 2(n-1), edge NOT repeated
   //   symmetric -> period 2n,     edge repeated
   auto reflect_coord = [](int r, int n, bool include_edge) -> int {
--- a/mlx/ops.cpp
+++ b/mlx/ops.cpp
@@ -1460,6 +1460,53 @@ array tile(
   return reshape(x, std::move(final_shape), s);
 }
 
+array reflect_pad(
+    const array& a,
+    const std::vector<int>& axes,
+    const Shape& low_pad_size,
+    const Shape& high_pad_size,
+    bool include_edge,
+    StreamOrDevice s /* = {} */) {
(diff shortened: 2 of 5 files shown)


zcbenz:
Thanks for the PR, there was actually https://github.com/ml-explore/mlx/pull/2862 which is still waiting for a thorough review to merge. I think your PR is more elegant and I'm more in favor of this one, just in case do you think we are missing anything from https://github.com/ml-explore/mlx/pull/2862?

katlun-lgtm:
Thanks @zcbenz! I went through #2862 carefully — functionally the two cover the same ground (`reflect` + `symmetric`, matching `numpy.pad` including the edge-excluded vs edge-included distinction), so nothing user-facing is missing.

Differences I found:

- **Implementation:** #2862 builds the padding with a slice/concatenate/tile/slice_update chain (a separate function per mode). This PR instead computes a single triangle-wave index map per padded axis and does one `take` (gather), with both modes sharing one helper via an `include_edge` flag — fewer lines and a lighter op graph.
- **Multi-reflect (pad wider than the axis):** handled in both; here it falls out of the modular index map rather than needing explicit tiling/`reps` logic.
- **Tests:** the suite here is a superset — it adds degenerate axes (`n == 1`, `n == 2`), multi-reflect on both sides simultaneously, asymmetric per-axis widths, and 3D cases, all checked elementwise against `numpy.pad`.
- The one thing #2862 had that this PR didn't — an `ACKNOWLEDGMENTS.md` entry — I've just added.

Happy to adjust naming or docs to whatever you prefer.

zcbenz (approved):
Thanks!

katlun-lgtm:
Following up now that #2862 has been closed — that leaves this PR as the remaining path for `reflect` / `symmetric` padding modes. It's approved with all CI green; happy to rebase onto latest main if you'd like it current, otherwise it's ready to merge whenever convenient. Thanks!

katlun-lgtm:
Rebased onto current main — this had drifted into a conflict, so it wasn't mergeable regardless of review. It's clean now at `f38f4522`.

The only conflict was documentation: main renamed the documented argument `constant_value` to `constant_values` to match the actual keyword, while this branch still had the old spelling. I resolved in main's favour and kept the two new mode lines above it. The implementation, the C++ tests and the Python tests all applied unchanged, and nothing in `mlx/ops.cpp` has touched `pad` since this branch was cut.

For the record on the earlier hold: #2862 was closed unmerged on July 1, so this is the remaining path for `reflect` / `symmetric`. The June 13 approval carried across the rebase.

The new head has no checks on it yet — fork runs need an approval to start, if someone can grant it.

zcbenz (changes requested):
Sorry for the late response, I had to find someone to verify the correctness of the implementation before merging, since no one did I have to spend some time myself.

So this is not a performant implementation: it builds the indices with the same length of output in the host, which would be rather slow for large inputs.

A performant implementation should follow `edge_pad`: use `slice_update` to write paddings and unpadded array to output, with host calculating only the start/stop indices. The `reflect_pad` only needs to change how the paddings are generated, by slicing and then flipping, you can find some idea in JAX's implementation https://github.com/jax-ml/jax/blob/8d1be7d7d1c4bd19a0de12c917709dd20f2de34d/jax/_src/numpy/lax_numpy.py#L3988.

katlun-lgtm:
Rewrote it to follow `edge_pad`'s pattern, pushed as `beec433e`.

Same structure as `edge_pad`: place the input into a zero-filled output via `slice_update`, then extend each axis outward with `slice` + `flip` on already-written data. Host only computes start/stop indices, no per-element index array. For a pad width larger than the axis, it loops in tiles, re-slicing the data it just wrote to continue the periodic pattern — same idea as your JAX reference, just landing in a fixed `slice_update` buffer instead of growing via `concatenate`.

All 154 `test_ops.py` cases pass, including the existing multi-reflect coverage (pad > axis) that exercises the tiling loop. Benchmarked on M3 Max: 5M-element 1D and 2000x2000 2D arrays with small in-bounds padding are sub-millisecond now (0.7-1.0ms); a deliberately pathological multi-reflect case (5M elements, pad 2M each side, 3 tiles) is still ~1.1ms.

katlun-lgtm:
Fixed the clang-format failure from the last run — my two `reflect_pad(...)` call sites in `pad()` were too long, wrapped them to match what the hook wanted. Pushed as `0f87956b`. Also picked up the main-merge that landed on this branch in the meantime.

zcbenz (approved):

PR #3612, opened 2026-06-01
Return zero VJPs for scatter indices

### LINKED ISSUE #1439: [BUG] ValueError: [scatter] Cannot calculate VJP with respect to indices.
**Describe the bug**
I understand that HungarianMatching algorithm requires linear_sum_assignment from scipy, which needs cost matrix to be evaluated. Hence, I cannot compile my train step function. However, if I use SimpleMatching algorithm and then compile my train step, I get the following error:

```
    (loss_value, loss_dict), grads = train_step_fn(samples, targets, need_tgt_for_training, return_outputs=False)
  File "site-packages/mlx/nn/utils.py", line 35, in wrapped_value_grad_fn
    value, grad = value_grad_fn(model.trainable_parameters(), *args, **kwargs)
ValueError: [scatter] Cannot calculate VJP with respect to indices.
```
Code for matching  is as follows:

``` python
        C = self.cost_bbox * cost_bbox + self.cost_class * cost_class + self.cost_giou * cost_giou
        C = C.reshape(bs, num_queries, -1)

        sizes = [v["num_objects"] for v in targets]
        indices = []
        
        for i, c in enumerate(C):
            # print(c.shape)
            if i == 0:
                start_index = 0
                end_index = start_index + sizes[i]
            else:
                start_index = sizes[i-1]
                end_index = start_index + sizes[i]
            cost_matrix = c[:, start_index:end_index]
            size_ = cost_matrix.shape[1]
            idx_i = cost_matrix.argmin(0)
            idx_j = mx.arange(size_)
            indices.append((idx_i, idx_j))
```

Also, is there a work around to get mx.compile working for Hungarian Matching algorithm?

Will greatly appreciate your help to solve this.

**Additional context**
Using MLX 0.17.3

devin-lai:
## Summary
- Return zero VJPs for scatter and scatter_axis index inputs.
- Match gather behavior and fix gradients when scatter indices come from differentiable values.
- Add autograd coverage for direct and derived scatter indices.

Closes #1439.

## Tests
- /opt/homebrew/bin/black --check python/tests/test_autograd.py
- clang-format --dry-run --Werror mlx/primitives.cpp
- git diff --check
- PYTHONPATH=/Users/ldy/Desktop/project/mt/opencode/mlx/python:/Users/ldy/Desktop/project/mt/opencode/mlx/python/tests python3.11 -m unittest python.tests.test_autograd.TestAutograd.test_scatter_index_vjp
- PYTHONPATH=/Users/ldy/Desktop/project/mt/opencode/mlx/python:/Users/ldy/Desktop/project/mt/opencode/mlx/python/tests python3.11 -m unittest python.tests.test_autograd

### DIFF
--- a/mlx/primitives.cpp
+++ b/mlx/primitives.cpp
@@ -4391,8 +4391,9 @@ std::vector<array> Scatter::vjp(
         }
       }
     } else {
-      throw std::invalid_argument(
-          "[scatter] Cannot calculate VJP with respect to indices.");
+      // Grads w.r.t. indices are zero
+      vjps.push_back(
+          zeros(primals[num].shape(), primals[num].dtype(), stream()));
     }
   }
   return vjps;
@@ -4503,8 +4504,9 @@ std::vector<array> ScatterAxis::vjp(
     } else if (num == 2) {
       vjps.push_back(take_along_axis(cotangents[0], indices, axis_, stream()));
     } else {
-      throw std::invalid_argument(
-          "[scatter_axis] Cannot calculate VJP with respect to indices.");
+      // Grads w.r.t. indices are zero
+      vjps.push_back(
+          zeros(primals[num].shape(), primals[num].dtype(), stream()));
     }
   }
   return vjps;
--- a/python/tests/test_autograd.py
+++ b/python/tests/test_autograd.py
@@ -367,6 +367,46 @@ class TestAutograd(mlx_tests.MLXTestCase):
         self.assertTrue(mx.allclose(vjps[0], mx.array([4.0, 0.0, 6.0, 0.0])))
         self.assertTrue(mx.allclose(vjps[1], mx.array([5.0, 7.0])))
 
+    def test_scatter_index_vjp(self):
+        # The gradient with respect to scatter indices is zero, consistent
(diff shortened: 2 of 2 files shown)


zcbenz (approved):
Looks good to me, thanks!

angeloskath:
I am a bit torn about this. The current mlx approach about this is wrap the indices with `mx.stop_gradient()`. This is for instance what we do in `gather_mm`.

I think the current approach is less of a footgun in that it won't let you try to calculate those gradients so you don't ever do unnecessary computations with 0. On the other hand it is more of a footgun because it will crash with this error if you try to calculate that gradient.

Regarding this PR, I think we should normalize to either return 0s (like in gather) or throw (like in scatter before this PR). Wdyt @nastya236 and @zcbenz ?

nastya236:
I'd lean toward making it explicit, through an error and force`mx.stop_gradient` for the indices (and we should do the same for `Gather`). It avoids hidden cost for zero VJPs and catches cases where someone unintentionally passes a differentiable value as indices.

zcbenz:
I'm closing this PR as we all agree on the above solution.

PR #3613, opened 2026-06-02
Correct p-norm computation in triplet_loss

pchintar:
## Description

`nn.losses.triplet_loss` exposes a configurable norm degree through the `p` argument and documents the distance term as a p-norm:

```python
||A - P||_p
```

However, the current distance computation is:

```python
mx.sqrt(mx.power(anchors - positives, p).sum(axis) + eps)
```

which only matches the documented p-norm when `p=2`. For other values of `p`, it can produce incorrect results and even `NaN` values.

This change updates the distance computation to use the standard p-norm:

```python
(sum(abs(x - y) ** p) + eps) ** (1 / p)
```

which matches both the function documentation and MLX's `linalg.norm` behavior.

On a sidenote, I also noticed that the LayerNorm, GroupNorm, and BatchNorm docstrings in `normalization.py` had `epsilon` outside the square root (`\sqrt{Var[x]} + \epsilon`) which doesn't match the standard normalization formula, so I changed it to (`\sqrt{Var[x] + \epsilon}`) instead.

## Changes

* Corrected the p-norm computation in `nn.losses.triplet_loss`
* Added test coverage for non-default `p` values

## Testing

```bash
python -m unittest python.tests.test_losses.TestLosses.test_triplet_loss
```

cc @zcbenz

### DIFF
--- a/python/mlx/nn/layers/normalization.py
+++ b/python/mlx/nn/layers/normalization.py
@@ -73,7 +73,7 @@ class LayerNorm(Module):
 
     .. math::
 
-        y = \frac{x - E[x]}{\sqrt{Var[x]} + \epsilon} \gamma + \beta,
+        y = \frac{x - E[x]}{\sqrt{Var[x] + \epsilon}} \gamma + \beta,
 
     where :math:`\gamma` and :math:`\beta` are learned per feature dimension
     parameters initialized at 1 and 0 respectively.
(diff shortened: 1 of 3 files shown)


zcbenz (approved):
Looks good to me, thanks for fixing this!

pchintar:
@zcbenz thnx again

PR #3615, opened 2026-06-02
Handle invalid dimensions in SinusoidalPositionalEncoding

pchintar:
## Description

`nn.SinusoidalPositionalEncoding` currently assumes that `dims` is positive and even, but this requirement is not enforced.

The implementation constructs sinusoidal embeddings from sine/cosine frequency pairs using:

```python
one_zero = 1 - mx.arange(0, dims // 2) / (dims // 2 - 1)
```

This works for the existing tested case (`dims=16`), but some edge cases are not handled correctly.

For example, if a user constructs:

```python
nn.SinusoidalPositionalEncoding(3)
```

the implementation produces only 2 embedding dimensions instead of the requested 3. Similarly, `dims=2` results in a division by zero while computing the values used to generate the sinusoidal frequencies, and non-positive values are not rejected explicitly.

This change validates that `dims` is a positive even integer and handles the `dims=2` case separately to avoid the division by zero.

## Tests

Added coverage for:

* `dims=2`
* Invalid dimensions: `0`, `1`, and `3`

cc @zcbenz

### DIFF
--- a/python/mlx/nn/layers/positional_encoding.py
+++ b/python/mlx/nn/layers/positional_encoding.py
@@ -85,7 +85,19 @@ class SinusoidalPositionalEncoding(Module):
     ):
         super().__init__()
 
-        one_zero = 1 - mx.arange(0, dims // 2) / (dims // 2 - 1)
+        # Sinusoidal embeddings are constructed from sin/cos pairs, so the
+        # embedding dimension must be positive and even.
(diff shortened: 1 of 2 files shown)


zcbenz (approved):
Thanks!

pchintar:
> Thanks!

thank you @zcbenz

PR #3617, opened 2026-06-02
Enable the Metal backend by default on iOS

### LINKED ISSUE #3493: Why is `MLX_BUILD_METAL` disabled on iOS
When targeting iOS, the `CMAKE_SYSTEM_NAME` fails the `Darwin` equality check, resulting in `MLX_BUILD_METAL` being disabled. Does MLX Metal not work on iOS?

eyupcanakman:
Fixes #3493.

iOS reports `iOS` for `CMAKE_SYSTEM_NAME`, so it failed the `Darwin` check and `MLX_BUILD_METAL` was forced off. The fix enables Metal by default on iOS and runs the `xcrun -sdk macosx` probe only on macOS, so the iOS configure no longer hard-errors. As discussed in the issue, the SDK is left to the caller and version detection is deferred, so iOS uses the baseline feature set for now. macOS is unchanged.

There's no configure-level test in the suite, so I didn't add one.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -55,7 +55,7 @@ message(
     "Building MLX for ${CMAKE_SYSTEM_PROCESSOR} processor on ${CMAKE_SYSTEM_NAME}"
 )
 
-if(${CMAKE_SYSTEM_NAME} MATCHES "Darwin")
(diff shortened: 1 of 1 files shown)


zcbenz (approved):
Thanks!

olokobayusuf:
Thanks @eyupcanakman !

PR #3621, opened 2026-06-04
Fix flaky test_siblings_without_eval

zcbenz:
Instead of measuring the RAM used by the process, which is very unreliable, change the test to directly verify we are not leaking `ArrayDesc` references.


angeloskath (approved):
🙏

PR #3632, opened 2026-06-05
Fix gather_qmm NAX kernel name mismatch

scyyh11:
## Proposed changes

`gather_qmm_nax` builds the NAX kernel name with `bk = 32`, but the gather qmm NAX kernels are only instantiated with `_bk64_` names. On NAX devices this makes the matrix-size `gather_qmm` path (taken for half-precision inputs, or float32 with tf32 enabled) fail during kernel lookup with an `Unable to load kernel ..._bk32...` error.

This changes the `gather_qmm_nax` kernel-name construction to use `bk = 64`, matching the existing instantiations. In this path `bk` is only part of the kernel name; the dispatch geometry is unchanged.

The PR also adds a regression test for matrix-size half-precision `gather_qmm` covering both `affine` and `mxfp4`, checked against a float32 reference.

Tested with:

```bash
DEVICE=gpu .venv/bin/pytest -q python/tests/test_quantized.py::TestQuantized::test_gather_qmm_matrix_path
```

## Checklist

### DIFF
--- a/mlx/backend/metal/quantized.cpp
+++ b/mlx/backend/metal/quantized.cpp
@@ -596,7 +596,9 @@ void gather_qmm_nax(
   int wn = 2;
   int bm = 64;
   int bn = 64;
-  int bk = 32;
+  // The gather qmm NAX kernels are instantiated with BK = 64 only; any
+  // other value here makes the kernel name lookup fail.
+  int bk = 64;
   MTL::Size group_dims(32, wn, wm);
   MTL::Size grid_dims((N + bn - 1) / bn, (M + bm - 1) / bm, B);
 
--- a/python/tests/test_quantized.py
+++ b/python/tests/test_quantized.py
@@ -992,6 +992,43 @@ class TestQuantized(mlx_tests.MLXTestCase):
         g2 = mx.grad(f_test)(x, qw, s, b, lhs_indices, rhs_indices)
         self.assertTrue(mx.allclose(g1, g2, atol=1e-4))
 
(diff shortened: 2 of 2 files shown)


zcbenz (approved):
Nice fix, thanks!

programVeins:
Confirming this on an **M5 Pro (macOS 27 beta, Metal NAX active)** - we hit the
exact same bug independently and this patch fixes it.

**Symptom:** any transposed `gather_qmm` that takes the matrix-size path
(`M >= vector_limit`) on a NAX device fails at kernel load:

```
[metal::Device] Unable to load kernel
affine_gather_qmm_t_nax_bfloat16_t_gs_64_b_4_bm64_bn64_bk32_...
```

because the dispatcher builds the name with `bk=32` while only the `bk64`
kernels are instantiated in `quantized_nax.metal`.

**Where it bites:** quantized MoE / switch layers. Single-token AR decode
(`M=1`) routes through `gather_qmv` and is unaffected - which is probably why
it slipped through - but anything that pushes many tokens through experts at
once trips it. For us it was a 4-bit MoE (DiffusionGemma 26B-A4B, 128 experts
top-8): a 256-token diffusion canvas hits `gather_qmm` at matrix size on every
denoising step, so every step died until this fix.

Rebuilt MLX from source with `bk=32 → bk=64` applied → the kernel loads, the
MoE path runs, and 4-bit expert generation works end to end. The change is
name-only here (tile geometry is unchanged), so it's purely a lookup fix.

+1 to merging, thanks!

Issue #3629, opened 2026-06-04
[BUG] JVP non-deterministic crashes  when called in a loop.


kyrollosyanny:
**Describe the bug**
I have some minimal ray tracing code that fails non-deterministically. Same code, same inputs, fails at different points. I share an example below. VJP on the other hand works correctly but is vey inefficient for this problem type. 

**To Reproduce**

Include code snippet
```
import numpy as np
import mlx.core as mx

DTYPE = mx.float32


def trace_module_mimic(theta, P0, q0, n_total, scalar_a):
    """Minimum 4-surface mimic of a ray-trace through optical elements."""
    xdec = mx.array([theta, 0.0, 0.0, 0.0], dtype=DTYPE)
    xdec = mx.repeat(xdec[None, :], n_total, axis=0)
    ydec = mx.repeat(mx.array([0.0, -1.5, -1.4, -1.3], dtype=DTYPE)[None, :], n_total, axis=0)
    t    = mx.repeat(mx.array([0.0,  5.0,  6.0,  7.0], dtype=DTYPE)[None, :], n_total, axis=0)
    c    = mx.array([0.0, 0.01, -0.008, 0.0], dtype=DTYPE)

    for surf in range(4):
        xs = xdec[:, surf:surf+1]; ys = ydec[:, surf:surf+1]; zs = t[:, surf:surf+1]
        P0 = mx.concatenate([P0[:, 0:1] - xs, P0[:, 1:2] - ys, P0[:, 2:3] - zs], axis=1)
        k = q0[:, 0:1]; l = q0[:, 1:2]; m = q0[:, 2:3]
        s0 = -P0[:, 2:3] / m
        X1 = P0[:, 0:1] + k*s0
        Y1 = P0[:, 1:2] + l*s0
        c_surf = c[surf]
        sj = 0.0
        for _ in range(5):
            Xj = X1 + k*sj; Yj = Y1 + l*sj; Zj = m*sj
            rho2 = Xj*Xj + Yj*Yj
            sqrt_val = mx.sqrt(1.0 - c_surf*c_surf*rho2 + 1e-12)
            F = Zj - c_surf*rho2/(1.0+sqrt_val)
            E = c_surf/sqrt_val
            sj = sj - F/(-Xj*E*k + -Yj*E*l + m)
        Xj = X1 + k*sj; Yj = Y1 + l*sj; Zj = m*sj
        # mx.where(cond, traced, const) — known-safe ordering
        TIR = (sj > -1e8)
[...]

PR #3633, opened 2026-06-05
Fix JVPs of select, arctan2, masked_scatter, and bitwise ops

### LINKED ISSUE #3627: [BUG] JVP silently returns a zero tangent when differentiating through mx.where
**Describe the bug**
JVP silently returns a zero tangent when differentiating through mx.where(cond, constant, traced_value) specifically when the constant is in the TRUE branch and the traced value is in the FALSE branch. The bug fires regardless of whether cond is ever True at runtime. 

**To Reproduce**

Include code snippet
```
"""Minimum reproducer: MLX `mx.jvp` silently returns 0 through
`mx.where(cond, <constant>, <traced_value>)`.

# TLDR

`mx.jvp(f, ...)` returns a *zero tangent* when `f` contains
`mx.where(cond, constant, traced_value)` — i.e. when the constant is in
the **TRUE branch** and the traced (tangent-bearing) value is in the
**FALSE branch**. `mx.vjp` and central finite differences both return
the correct gradient.

The bug fires regardless of:
  - whether `cond` is ever True at runtime (it is False everywhere here),
  - what the constant is (NaN, ±inf, 0.0, 999.0, mx.array(0.0), ...),
  - whether the `traced_value` was produced by simple arithmetic.

Swapping the branches — putting the traced value in the TRUE branch and
the constant in the FALSE branch — gives the correct JVP. So does using
two traced values in both branches.


# Expected output (MLX 0.31.2, Apple Silicon)

    TRUE=traced FALSE=const  : VJP=+2.000e+00  JVP=+2.000e+00  ratio=1.00e+00
    TRUE=const  FALSE=traced : VJP=+2.000e+00  JVP=+0.000e+00  ratio=2.00e+30   <-- BUG
    TRUE=traced FALSE=traced : VJP=+2.000e+00  JVP=+2.000e+00  ratio=1.00e+00
    FD ground truth          : +1.999736e+00



# Fix for now

Swap the branches: replace `mx.where(cond, const, traced)` with
`mx.where(mx.logical_not(cond), traced, const)`.

# Tested

MLX 0.31.2, Apple Silicon .
"""
import mlx.core as mx


def _bench(f, label):
    theta = mx.array(0.0, dtype=mx.float32)
    v = mx.array(1.0, dtype=mx.float32)
    cot = mx.array(1.0, dtype=mx.float32)

    primal = float(f(theta))
    _, vj = mx.vjp(f, [theta], [cot]); mx.eval(vj[0]); vjp = float(vj[0])
    _, jv = mx.jvp(f, [theta], [v]);  mx.eval(jv[0]); jvp = float(jv[0])
    ratio = abs(vjp) / max(abs(jvp), 1e-30)
    flag = " <-- BUG" if ratio > 1e6 else ""
    print(f"  {label:25s}: primal={primal:+.4f}  VJP={vjp:+.4e}  "
          f"JVP={jvp:+.4e}  ratio={ratio:.2e}{flag}")
    return ratio > 1e6


# Variant A: traced value in TRUE branch, constant in FALSE branch — works
def fA(theta):
    y = (theta + 1.0) ** 2
    return mx.where(y > -1.0, y, 999.0)        # cond always True


# Variant B: constant in TRUE branch, traced value in FALSE branch — BUG
def fB(theta):
    y = (theta + 1.0) ** 2
    return mx.where(y < -1.0, 999.0, y)        # cond always False


# Variant C: both branches traced — works
def fC(theta):
    y = (theta + 1.0) ** 2
    z = (theta + 2.0)
    return mx.where(y > -1.0, y, z)


print("=" * 78)
print("MLX `mx.jvp` silent-zero bug through `mx.where(cond, const, traced)`")
print("=" * 78)
_bench(fA, "TRUE=traced  FALSE=const")
buggy = _bench(fB, "TRUE=const   FALSE=traced")
_bench(fC, "TRUE=traced  FALSE=traced")

# Finite-difference ground truth for fB (same primal as fA/fC)
theta = mx.array(0.0, dtype=mx.float32); h = 1e-4
rp = float(fB(theta + mx.array(h, dtype=mx.float32)))
rm = float(fB(theta - mx.array(h, dtype=mx.float32)))
print(f"\n  FD ground truth on fB    : {(rp - rm) / (2.0 * h):+.6e}")

print()
if buggy:
    print("==> BUG CONFIRMED: `mx.jvp` through `mx.where(cond, const, traced)` "
          "returns a zero tangent, even when `cond` is always False at "
          "runtime. `mx.vjp` and finite-differences both return the "
          "correct gradient (+2.0).")
    print()
    print("    Workaround in user code: swap the branches —")
    print("      mx.where(mx.logical_not(cond), traced, const)")
else:
    print("==> Bug not triggered in this run.")

```

**Expected behavior**
I expect both the VJP and JVP answers to match. This does not happen here. 

 - Version MLX 0.31.2

qflen:
Fixes #3627 AND #3629

The jvp transform passes tangents packed: `tangents[i]` is the tangent of input `argnums[i]`, and a jvp returns one tangent per output. Four jvps broke this contract and read out of bounds when only some of their inputs were traced.

| jvp | bug | fix |
|---|---|---|
| `Select` | helper called with `argnums[i]` instead of `i`, so it read `argnums[argnums[i]]` and `tangents[1]`/`tangents[2]` | index by position, use `tangents[i]`; condition tangent gets the output dtype |
| `ArcTan2` | always read `tangents[1]`, OOB for `atan2(traced, constant)` | accumulate one term per traced input |
| `MaskedScatter` | read `tangents[1]` for the update even when it was the only traced input | use `tangents[i]` |
| `BitwiseBinary` | returned one tangent per traced input (vjp semantics), so the transform indexed `outputs[1]` OOB and corrupted the tangent map | return one zero tangent per output; vjp written out instead of delegating to jvp |

**Before**

```python
_, (dout,) = mx.jvp(lambda t: mx.where(t < -1.0, 999.0, t * t), [mx.array(3.0)], [mx.array(1.0)])
# dout = 0.0, silently wrong (vjp and finite differences give 6.0)
```

The repro in #3629 crashed 10/10 runs with segfaults and non-deterministic shape errors.

**After**

`dout = 6.0`, jvp matches vjp and finite differences, and the repro passes 10/10 runs.

Added a regression test per case and a debug assert in the jvp transform that a primitive returns at most one tangent per output. Also fixed the stale `argnums.size() == 2` assert in `ArcTan2::vjp` (one-sided grads aborted debug builds) and added a `where` test with a constant condition and both branches traced.

## Checklist

### DIFF
--- a/mlx/primitives.cpp
+++ b/mlx/primitives.cpp
@@ -551,14 +551,18 @@ std::vector<array> ArcTan2::jvp(
   const array& x1 = primals[0];
   const array& x2 = primals[1];
 
-  auto jvp_fun = [&](int i) {
-    return argnums[i] == 0 ? multiply(x2, tangents[i], s)
-                           : negative(multiply(x1, tangents[i], s), s);
-  };
-  auto numerator = jvp_fun(0);
-  if (argnums.size() > 1) {
-    numerator = add(numerator, jvp_fun(1), s);
-  }
+  auto numerator = [&](){
+    if (argnums.size() == 2) {
+      return subtract(
+          multiply(x2, tangents[0], s),
+          multiply(x1, tangents[1], s),
+          s);
+    } else if (argnums[0] == 0) {
+      return multiply(x2, tangents[0], s);
+    } else {
+      return negative(multiply(x1, tangents[1], s), s);
+    }
+  }();
   return {divide(numerator, add(square(x1, s), square(x2, s), s), s)};
 }
 
--- a/mlx/primitives.cpp
+++ b/mlx/primitives.cpp
@@ -518,7 +518,7 @@ std::vector<array> ArcTan2::vjp(
     const std::vector<int>& argnums,
     const std::vector<array>&) {
   assert(primals.size() == 2);
-  assert(argnums.size() == 2);
+  assert(cotangents.size() == 1);
 
   const auto& s = stream();
   const array& x1 = primals[0];
--- a/mlx/primitives.cpp
+++ b/mlx/primitives.cpp
@@ -545,18 +545,21 @@ std::vector<array> ArcTan2::jvp(
     const std::vector<array>& tangents,
     const std::vector<int>& argnums) {
   assert(primals.size() == 2);
-  assert(argnums.size() == 2);
+  assert(tangents.size() == argnums.size());
 
   const auto& s = stream();
   const array& x1 = primals[0];
   const array& x2 = primals[1];
-  const array& dx1 = tangents[0];
-  const array& dx2 = tangents[1];
 
-  return {divide(
-      subtract(multiply(x2, dx1, s), multiply(x1, dx2, s), s),
-      add(square(x1, s), square(x2, s), s),
-      s)};
+  auto jvp_fun = [&](int i) {
+    return argnums[i] == 0 ? multiply(x2, tangents[i], s)
+                           : negative(multiply(x1, tangents[i], s), s);
+  };
+  auto numerator = jvp_fun(0);
+  if (argnums.size() > 1) {
+    numerator = add(numerator, jvp_fun(1), s);
+  }
+  return {divide(numerator, add(square(x1, s), square(x2, s), s), s)};
 }
 
 std::pair<std::vector<array>, std::vector<int>> ArcTan2::vmap(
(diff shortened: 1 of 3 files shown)


zcbenz (approved):
Nice fix, thanks!

angeloskath (approved):

PR #3636, opened 2026-06-06
Fix JVPs of power, divmod, and slice_update with partly traced inputs

### LINKED ISSUE #3634: [BUG] JVPs of power, divmod, and slice_update mishandle partly traced inputs
**Describe the bug**

Three more `Primitive::jvp` implementations mishandle partly traced inputs — the same class as #3627/#3629 (fixed by #3633), found while auditing the remaining jvp implementations against the packed-tangents convention (`tangents[i]` is the tangent of input `argnums[i]`; a jvp returns one tangent per output):

1. `Power::jvp` delegates to `vjp`, which scales every partial by `cotangents[0]`. With both inputs traced, both partials are multiplied by the first input's tangent and the second tangent is ignored — silently wrong tangents.

2. `DivMod::jvp` returns one tangent for a two-output primitive, so the second output never enters the tangent map and its consumer is invoked with empty `argnums`/`tangents` — segfault in release builds.

3. `DynamicSliceUpdate::jvp` ignores `argnums` and reads `tangents[0]`/`tangents[1]` unconditionally — segfault when only one input is traced, and mispaired tangents for `argnums = {0, 2}`.

**To Reproduce**

```python
import mlx.core as mx

# 1) power: silently wrong with both inputs traced
_, (j,) = mx.jvp(lambda a, b: a**b, [mx.array(2.0), mx.array(3.0)],
                 [mx.array(0.0), mx.array(1.0)])
print(j)  # 0.0       — expected d/db 2^b = ln(2)*8 ~ 5.5452
_, (j,) = mx.jvp(lambda a, b: a**b, [mx.array(2.0), mx.array(3.0)],
                 [mx.array(1.0), mx.array(0.0)])
print(j)  # 17.5452   — expected d/da a^3 = 12.0 (got 12 + 5.5452)

# 2) divmod: segfault when the second output is consumed
mx.jvp(lambda x: -(mx.divmod(x, mx.array(3.0))[1]), [mx.array(7.0)], [mx.array(1.0)])

# 3) slice_update with array start indices, only the update traced: segfault
x = mx.zeros((4,))
mx.jvp(lambda u: mx.slice_update(x, u, mx.array([1]), axes=[0]),
       [mx.ones(2)], [mx.ones(2)])
```

**Expected behavior**

1. `power` uses each input's own tangent: 5.5452 and 12.0 respectively.
2. `divmod` returns a tangent per output (zeros, matching its vjp semantics); no crash.
3. `slice_update` treats untraced inputs as zero tangents; no crash.

In a debug build, 2) aborts at `Negative::jvp`'s `assert(argnums.size() == 1)`, confirming the downstream primitive is invoked with no tangents. `mx.vjp`/`mx.grad` are unaffected for all three.

PR coming soon.

**Desktop**
- macOS, Apple Silicon
- main @ 8f0e8b14

qflen:
Fixes #3634

Same class as #3627/#3629 (fixed by #3633): 

`Primitive::jvp` implementations that break the packed-tangents convention (`tangents[i]` is the tangent of input `argnums[i]`; a jvp returns one tangent per output).

| jvp | bug | fix |
|---|---|---|
| `Power` | delegated to `vjp`, which scales every partial by `cotangents[0]`, so with both inputs traced both partials used the first input's tangent and the second tangent was ignored | one `vjp` call per traced input with its own tangent as the cotangent, summed |
| `DivMod` | returned one tangent for a two-output primitive, so the second output never entered the tangent map and its consumer was invoked with empty `argnums`/`tangents` — segfault | return one zero tangent per output (divmod stays non-differentiable, matching its vjp) |
| `DynamicSliceUpdate` | ignored `argnums` and read `tangents[0]`/`tangents[1]` unconditionally — OOB segfault with one traced input, mispaired tangents for `argnums = {0, 2}` | untraced inputs get zero tangents; traced start indices throw, matching the vjp |

**Before**

```python
_, (j,) = mx.jvp(lambda a, b: a**b, [mx.array(2.0), mx.array(3.0)],
                 [mx.array(0.0), mx.array(1.0)])
# j = 0.0, silently wrong (vjp and finite differences give ln(2)·8 ≈ 5.5452)
```

The divmod and slice_update repros in #3634 segfaulted.

**After**

`j = 5.5452`, jvp matches vjp and finite differences, and all repros pass.

Added a regression test covering tangent pairing for power, per-output tangents for divmod (including a consumer of the second output), and dynamic slice_update with each subset of traced inputs. Each case fails on unfixed main (wrong value for power, segfault for the other two).

## Checklist

### DIFF
--- a/mlx/primitives.cpp
+++ b/mlx/primitives.cpp
@@ -1806,7 +1806,9 @@ std::vector<array> DivMod::jvp(
     const std::vector<array>& primals,
     const std::vector<array>& tangents,
     const std::vector<int>& argnums) {
-  return {zeros_like(primals[0], stream())};
+  // One tangent per output; inputs are pre-broadcast and pre-promoted
+  auto zero = zeros_like(primals[0], stream());
+  return {zero, zero};
 }
 
 std::pair<std::vector<array>, std::vector<int>> DivMod::vmap(
@@ -3409,12 +3411,12 @@ std::vector<array> Power::jvp(
     const std::vector<array>& tangents,
     const std::vector<int>& argnums) {
   auto output = power(primals[0], primals[1], stream());
-  auto grads = vjp(primals, tangents, argnums, {output});
-  if (argnums.size() > 1) {
-    return {add(grads[0], grads[1], stream())};
-  } else {
-    return grads;
+  auto jvp = vjp(primals, {tangents[0]}, {argnums[0]}, {output})[0];
+  for (int i = 1; i < argnums.size(); ++i) {
+    jvp = add(
+        jvp, vjp(primals, {tangents[i]}, {argnums[i]}, {output})[0], stream());
   }
+  return {jvp};
 }
 
 std::pair<std::vector<array>, std::vector<int>> Power::vmap(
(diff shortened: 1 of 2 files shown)


zcbenz (approved):
Thanks!

angeloskath (approved):
Awesome, thanks!

PR #3639, opened 2026-06-08
Add stft and istft to mlx.core.fft

eyupcanakman:
## Proposed changes

Adds `stft` and `istft` to `mlx.core.fft`, closing #1004. Both are written in C++ so they are available from every language binding.

`stft` frames the last axis with `as_strided`, applies the window, and runs `rfft` (or `fft` for two-sided output). `istft` inverts that with an overlap-add normalized by the squared-window envelope. The signature follows `torch.stft`, except `norm` takes the same `"backward"`/`"ortho"`/`"forward"` string the rest of `mlx.fft` uses. Batched input works and the output has shape `(..., n_freq, n_frames)`.

Tested against `torch.stft` and `torch.istft` over reflect, constant, and edge padding, one-sided and two-sided transforms, the `ortho` norm, batched input, and `n_fft` that is not a multiple of `hop_length`. The C++ tests cover the shapes, an overlap-add round-trip, and the error paths.

## Checklist

### DIFF
--- a/mlx/fft.cpp
+++ b/mlx/fft.cpp
@@ -1,4 +1,5 @@
 // Copyright © 2023 Apple Inc.
+#include <algorithm>
 #include <cmath>
 #include <functional>
 #include <numeric>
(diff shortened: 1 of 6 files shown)


mlx/fft.cpp
```diff
@@ -240,6 +241,270 @@ array irfftn(
   return fft_impl(a, true, true, norm, s);
 }
 
+namespace {
+
+// Pad the last axis; mx::pad has no reflect mode, so build it from take.
```
zcbenz, line 341: Would #3608 help?

eyupcanakman: Yeah, its reflect mode matches what this helper does and also handles pad >= n, so I can drop the helper and call pad directly once it lands. Want me to rebase on it after it merges, or keep the local version for now?

zcbenz: Thanks for verifying, let's just keep the local version for now.

mlx/fft.h
```diff
@@ -212,6 +213,33 @@ inline array irfft2(
...
+    const std::optional<int>& win_length = std::nullopt,
+    const std::optional<array>& window = std::nullopt,
+    bool center = true,
+    const std::string& pad_mode = "reflect",
+    FFTNorm norm = FFTNorm::Backward,
+    const std::optional<bool>& onesided = std::nullopt,
+    StreamOrDevice s = {});
+
+/** Compute the inverse Short-Time Fourier Transform. */
+MLX_API array istft(
```
zcbenz, line 253: Can you move the function declarations and implementations to the end of the files, so the functions are listed in the same order?

zcbenz (approved):
I must admit I'm not familiar with short-time fft at all, but the implementation looks good, and the tests are well-written, so this PR looks good to me.

mlx/fft.h
eyupcanakman: Moved `stft` and `istft` to the end of `fft.h` and `fft.cpp` so they line up.

eyupcanakman:
Approved and green since mid-June. Not sure if this is waiting on something or just got buried.

PR #3644, opened 2026-06-08
Propagate NaN through sign to match NumPy

devYRPauli:
## Summary

`mx.sign` returns `0` for NaN inputs instead of propagating NaN.

- CPU (`backend/cpu/unary_ops.h`): the `Sign` functor selects the `0` branch for any value that is neither `< 0` nor `> 0`, which includes NaN.
- Metal (`backend/metal/kernels/unary_ops.h`): `(x > T(0)) - (x < T(0))` evaluates to `0 - 0 = 0` for NaN.

NumPy's `np.sign` returns NaN for NaN, and the rest of the library already propagates NaN (`max`, `maximum`, `clip`, `log`, `sqrt`), so `sign` was inconsistent both with NumPy and with itself.

```
mx.sign(mx.array([nan, -nan, 1.5, -1.5, 0.0]))  ->  [0, 0, 1, -1, 0]   (before)
np.sign(np.array([nan, -nan, 1.5, -1.5, 0.0]))  ->  [nan, nan, 1, -1, 0]
```

## Fix

Return the input itself when it is neither `<`, `>`, nor `== 0`, so NaN flows through. Integer types are unaffected: every integer is `<`, `>`, or `== 0`, so the new fall-through is never taken for them. Zero (including `-0.0`) and the signed finite cases are unchanged.

## Tests

Extended `test_sign` in `python/tests/test_ops.py` with a NaN vector across `float32`, `float16`, and `bfloat16`, comparing to `np.sign` with `equal_nan=True`.

Verified locally with a CPU build (`-DMLX_BUILD_METAL=OFF`): NaN now propagates for all three float dtypes and integer `sign` is unchanged (`[-3, 0, 7] -> [-1, 0, 1]`). I don't have the Metal toolchain on this machine, so the Metal kernel change wasn't compiled locally; it applies the same `x != x ? x : ...` guard as the CPU path and relies on CI for the GPU build.

### DIFF
--- a/mlx/backend/cpu/unary_ops.h
+++ b/mlx/backend/cpu/unary_ops.h
@@ -94,7 +94,11 @@ struct Sign {
     } else if constexpr (std::is_same_v<T, complex64_t>) {
       return simd::select(x == z, x, Simd<T, N>(x / simd::abs(x)));
     } else {
-      return simd::select(x < z, m, simd::select(x > z, o, z));
+      // For floating types a NaN is neither <, >, nor == 0, so the innermost
+      // branch returns x to propagate NaN (matching NumPy). For integer types
+      // every value is <, >, or == 0, so the x fall-through is never taken.
+      return simd::select(
+          x < z, m, simd::select(x > z, o, simd::select(x == z, z, x)));
     }
   }
   SINGLE()
--- a/mlx/backend/metal/kernels/unary_ops.h
+++ b/mlx/backend/metal/kernels/unary_ops.h
@@ -316,7 +316,9 @@ struct Sigmoid {
 struct Sign {
   template <typename T>
   T operator()(T x) {
-    return (x > T(0)) - (x < T(0));
+    // x != x is true only for NaN (floating types), which propagates to match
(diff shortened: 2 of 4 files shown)


zcbenz (commented):
PyTorch does not propagate NaN for this case so I think we should not change the behavior.

```console
print(torch.sign(torch.tensor([nan, -nan, 1.5, -1.5, 0.0])))
> tensor([ 0.,  0.,  1., -1.,  0.])
```

From performance's perspective correctly handling NaN requires adding more instructions so it is usually provided as an extra option when needed like the `equal_nan=True` in `array_equal`.

devYRPauli:
That's fair — you're right that `torch.sign` returns 0 for NaN, and since propagating it costs extra instructions on the hot path, keeping it as default-off (the way `array_equal` exposes `equal_nan`) makes more sense than changing `sign` unconditionally. I'd been anchoring on NumPy's behavior, but matching PyTorch and avoiding the per-element cost is the better default. Closing this; happy to revisit as an opt-in if it ever comes up. Thanks for the look.

PR #3648, opened 2026-06-09
Emit valid kernel source for non-finite float constants

tillahoffmann:
## Summary

Compiling a function that bakes a **NaN** (or infinity) scalar constant into a fused kernel fails to compile. `print_float_constant` streams the value with `operator<<`, producing a bare token like `nan`, which is interpolated verbatim into the generated kernel source and is not a valid identifier:

```
ternary_ops.h:23:35: error: use of undeclared identifier 'nan'
  auto tmp_D = static_cast<float>(nan);
```

Eager evaluation is fine; only the compiled path is affected.

## Repro

```python
import mlx.core as mx
f = mx.compile(lambda x: mx.where(x > 0, x, float("nan")))
f(mx.array([1.0, -1.0]))   # before: Metal compile error; after: array([1, nan])
```

## Fix

Emit the `NAN` / `INFINITY` / `-INFINITY` macros for non-finite values. Both kernel toolchains provide them — Metal via `<metal_stdlib>` (`#define NAN __builtin_nanf("")`, `#define INFINITY __builtin_inff()`) and the CPU C++ codegen via `<cmath>` — so the same emitted text is valid for both backends. The finite check widens to `double` first so it is exact for every float type.

## Tests

- `tests/compile_tests.cpp`: `test compile with non-finite constants`.
- `python/tests/test_compile.py`: `test_compile_nonfinite_constants` (float32/float16/bfloat16, NaN and -inf).

Verified red→green on both: the unfixed build reproduces the `nan` error, the fixed build computes the expected values. Full C++ compile suite and `test_compile.py` pass.

### DIFF
--- a/mlx/backend/common/compiled.h
+++ b/mlx/backend/common/compiled.h
@@ -1,6 +1,7 @@
 // Copyright © 2023-2024 Apple Inc.
 #pragma once
 
+#include <cmath>
 #include <functional>
 #include <iomanip>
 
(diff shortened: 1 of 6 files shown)


zcbenz (approved):
This would be a nice improvement, can you also remove the `#define inf`s in https://github.com/search?q=repo%3Aml-explore%2Fmlx%20%22define%20inf%22&type=code ?

PR #3649, opened 2026-06-09
fix(shapeless): infer dynamic dim in Unflatten::output_shapes (#2607)

### LINKED ISSUE #2607: [BUG] shapeless matmul isn't
**Describe the bug**

In some cases, a shapeless export of matmul isn't truly shapeless.

**To Reproduce**

Add this to python/tests/test_export_import.py

```python
    def test_export_matmul_shapeless_mid_dim(self):
        path = os.path.join(self.test_dir, "matmul_shapeless.mlxfn")

        E, H = 64, 17
        arr = mx.arange(E * H, dtype=mx.float32).reshape((E, H)) * (1.0 / (E * H))

        def fn(x):
            return mx.matmul(x, arr)

        sample = mx.zeros((1, 40, E), dtype=mx.float32)
        mx.export_function(path, fn, sample, shapeless=True)
        imported = mx.import_function(path)

        for seq_len in (40, 248, 623):
            with self.subTest(seq_len=seq_len):
                x = mx.arange(seq_len * E, dtype=mx.float32).reshape((1, seq_len, E))
                expected = fn(x)
                (y,) = imported(x)
                self.assertEqual(y.shape, (1, seq_len, H))
                self.assertTrue(mx.allclose(y, expected))
```

It shapelessly exports a matmul, then re-imports it and confirms the behavior is identical to the original function across different input shapes.

**Expected behavior**

Want: middle dimension (40) adjusts shapelessly.

Got:

```
self.assertEqual(y.shape, (1, seq_len, H))
AssertionError: Tuples differ: (1, 40, 17) != (1, 248, 17)

self.assertEqual(y.shape, (1, seq_len, H))
AssertionError: Tuples differ: (1, 40, 17) != (1, 623, 17)
```

 - OS Version: macOS 15.6.1 (24G90)
 - Version: main as of 9/16/25 (3f730e77aa3d14e3d52688b8bd6a24bace500166)

**Additional context**

I maybe have a tentative fix, but my confidence in it is low. Happy to send it if the reviewer is feeling patient. :)

nac7:
## What

`mx.matmul(x, W)` where `x` has `ndim > 2` and `W` has `ndim ≤ 2`
is implemented in `ops.cpp` as:

```
Flatten(x, 0, -2)  →  Matmul  →  Unflatten(out, 0, orig_shape)
```

`orig_shape` is `x.shape()[:-1]` at trace time (e.g. `{1, 40}`), which
is stored verbatim as state inside the `Unflatten` primitive.

In shapeless mode `compile_replace` calls `output_shapes` on every
primitive to recompute output shapes for the actual inputs.
`Unflatten::output_shapes` returned the stored shape unchanged, so a
call with `seq_len=248` would produce `(1, 40, H)` instead of
`(1, 248, H)` — silently wrong.

## Fix

`mlx/primitives.cpp` — `Unflatten::output_shapes`:

When `inputs[0].shape()[axis_] ≠ product(shape_)`, scan `shape_`
right-to-left for the unique dimension whose removal leaves a divisor
of the actual flat size, then scale that dimension to absorb the new
size.  If no such dimension exists, raise `std::invalid_argument`.

This matches the common transformer pattern where leading batch
dimensions are static and `seq_len` is dynamic.

## Tests

Two new tests in `python/tests/test_export_import.py`:

- **`test_export_matmul_shapeless_mid_dim`** — the exact reproduction
  case from #2607 (`B=1`, varying `seq_len` ∈ {40, 248, 623}); checks
  both shape and numerical agreement with the reference function.
- **`test_export_matmul_shapeless_batch_and_mid_dim`** — higher-rank
  variant (`B=2`, varying `seq_len`); ensures the fix generalises
  beyond the single-batch case.

Fixes #2607

### DIFF
--- a/mlx/primitives.cpp
+++ b/mlx/primitives.cpp
@@ -2200,7 +2200,36 @@ Shape Unflatten::output_shape(
 }
 
 std::vector<Shape> Unflatten::output_shapes(const std::vector<array>& inputs) {
-  return {Unflatten::output_shape(inputs[0], axis_, shape_)};
+  auto flat_actual = inputs[0].shape()[axis_];
+
+  // Compute the stored product so we can detect a shape mismatch in
+  // shapeless mode (e.g. when the contracted matmul batch dim changes).
+  int flat_stored = 1;
+  for (auto s : shape_) {
+    flat_stored *= s;
+  }
+
+  if (flat_actual == flat_stored) {
+    return {Unflatten::output_shape(inputs[0], axis_, shape_)};
+  }
+
+  // Shapeless mode: the flat dimension changed.  Find the rightmost dimension
+  // in shape_ whose removal leaves a divisor of flat_actual and scale it to
+  // absorb the new size.  This handles the common transformer pattern where
+  // batch dims are static but seq_len is dynamic.
+  for (int i = (int)shape_.size() - 1; i >= 0; --i) {
+    int other = flat_stored / shape_[i];
+    if (other > 0 && flat_actual % other == 0) {
+      Shape dyn_shape = shape_;
+      dyn_shape[i] = flat_actual / other;
+      return {Unflatten::output_shape(inputs[0], axis_, dyn_shape)};
+    }
+  }
+
+  throw std::invalid_argument(
+      "[Unflatten::output_shapes] Cannot infer output shape in shapeless mode: "
+      "flat size " + std::to_string(flat_actual) +
+      " is not divisible by any sub-product of the stored expansion.");
 }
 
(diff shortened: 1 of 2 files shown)


zcbenz:
Can you fix the lint error?

nac7:
Hi @zcbenz , fixed the lint error. Thanks for your review!

zcbenz (changes requested):
It is impossible to fix the issue by modifying `Unflatten::output_shapes`, the hack here may work for the very specific case but won't work generally. You have to get rid of the `unflatten` op that takes dynamic shape as input.

PR #3650, opened 2026-06-10
[fix] allow exporting custom Metal kernels without a Metal backend

### LINKED ISSUE #3240: [Feature Request] Fallback for custom kernels
We have a specific use-case whereby we need to export MLX callables from Linux. Currently, this prevents us from using any operations that require custom kernels (e.g. grid sample) because otherwise, MLX complains that no Metal GPU was found.

Ideally, we would want to keep the fast kernel so that we can export without a Metal GPU, but execute on a Metal GPU.

xthomaswang:
## Proposed changes

Fixes #3240

`mx.fast.metal_kernel` throws at construction on non-Metal builds, so functions
using custom Metal kernels can't be exported from Linux.

- Move `metal_kernel` graph construction to backend/common so it builds on all platforms
- During export tracing without Metal, record the kernel on a placeholder GPU stream which the importing process remaps to a real stream
- Outside of export tracing, calling the kernel without Metal still raises an error

The same mechanism could support exporting `mx.fast.cuda_kernel` from non-CUDA
platforms; happy to do that as a follow-up.

## Checklist


xthomaswang:
This now conflicts with #3638, which made the tracing state thread-local. My plan is to rebase on main and make the new `InExportTracing` counter thread-local as well to match the rest of the tracing state, let me know if sounds good, or do you suggest any other approach.

zcbenz:
This basically looks good to me, a rebase and changing the counter to threadlcoal would be enough.

xthomaswang:
> This basically looks good to me, a rebase and changing the counter to threadlcoal would be enough.

Done, rebase on main and change counter to threadlocal.

mlx/backend/common/metal_kernel.cpp
```diff
@@ -0,0 +1,344 @@
...
+  // Exporting without Metal: the kernel cannot run here but it can still be
+  // recorded in the graph on a placeholder GPU stream. The importing process
+  // remaps it to one of its own streams.
+  auto* device = std::get_if<Device>(&s);
+  auto* stream = std::get_if<Stream>(&s);
+  auto* tl_stream = std::get_if<ThreadLocalStream>(&s);
+  if ((device && *device != Device::gpu) ||
+      (stream && stream->device != Device::gpu) ||
+      (tl_stream && tl_stream->device != Device::gpu)) {
+    throw std::invalid_argument("[metal_kernel] Only supports the GPU.");
```
zcbenz, line 46: Is it necessary to require a gpu stream here? Can we just return `to_stream(s)` for exporting?

xthomaswang: Since `CustomKernel` has no CPU reval path, if we return to_stream(s) with cpu, then the exporting customkernel will not be able to run in metal.

I didn't see any import-time API to override the stream placement for this, so I use placeholder of gpu here.

xthomaswang:
Just fixed explicitly pass `stream=mx.gpu` in `test_export_import.py`.

zcbenz (approved):
Thanks!

zcbenz:
Would you mind rebasing the PR to fix merge conflict with #3662?

PR #3652, opened 2026-06-10
Fix Adafactor factored update for params with more than 2 dims

ishtihoss:
## Proposed changes

Adafactor crashes on any parameter with more than 2 dimensions, for example a conv weight or a stacked weight:

```python
import mlx.core as mx

p = {"w": mx.ones((2, 3, 4))}
g = {"w": mx.full((2, 3, 4), 0.1)}
optim.Adafactor().apply_gradients(g, p)
# ValueError: [matmul] Last dimension of first input with shape (2,3,1) must match
# second to last dimension of second input with shape (1,2,4).
```

`init_single` creates factored state for anything with ndim >= 2, but `_approximate_exp_moving_avg` recombines the factors with a matmul of `expand_dims(r, -1)` and `expand_dims(c, 0)`, which only works for 2d. The reference implementation (huggingface Adafactor `_approx_sq_grad`) uses `unsqueeze(-1)` and `unsqueeze(-2)` with an elementwise multiply, which is the same thing for 2d and also handles leading batch dims. This switches to that form.

Checked the 2d path against the huggingface implementation before and after (max diff 1.9e-9, unchanged) and the 3d path now matches it too (max diff 3.7e-9). Added a 3d case to the existing test.

## Checklist

### DIFF
--- a/python/mlx/optimizers/optimizers.py
+++ b/python/mlx/optimizers/optimizers.py
@@ -797,9 +797,7 @@ class Adafactor(Optimizer):
             exp_avg_sq_row / mx.mean(exp_avg_sq_row, axis=-1, keepdims=True)
         )
         c_factor = mx.rsqrt(exp_avg_sq_col)
-        return mx.matmul(
-            mx.expand_dims(r_factor, axis=-1), mx.expand_dims(c_factor, axis=0)
-        )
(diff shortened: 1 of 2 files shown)


zcbenz (approved):
Thanks!

PR #3653, opened 2026-06-10
Validate inputs in GroupNorm and InstanceNorm

ishtihoss:
## Proposed changes

Two cases where the norm layers accept bad inputs and return meaningless results instead of raising:

1. `InstanceNorm` on an input with no spatial dimensions returns all zeros. For a `(N, C)` or `(C,)` input the reduction axes tuple is empty, so `mx.mean` returns the input itself and `mx.var` returns 0, and the normalized output is exactly zero everywhere:

```python
import mlx.core as mx
import mlx.nn as nn

nn.InstanceNorm(dims=8)(mx.random.normal((4, 8)))
# all zeros, no error
```

2. `GroupNorm` does not check that `num_groups` divides `dims`. With say `num_groups=4, dims=6` the reshape can still succeed depending on the spatial size, and the groups end up mixing channel and spatial positions, so the statistics are garbage. The `pytorch_compatible=True` path floor-divides the group size and silently drops the intended grouping too. PyTorch raises for this at construction.

Both now raise a `ValueError`, same idea as #3615 did for `SinusoidalPositionalEncoding`. Added tests to the existing group norm and instance norm tests.

## Checklist

### DIFF
--- a/python/mlx/nn/layers/normalization.py
+++ b/python/mlx/nn/layers/normalization.py
@@ -56,6 +56,11 @@ class InstanceNorm(Module):
         return f"{self.dims}, eps={self.eps}, affine={'weight' in self}"
 
     def __call__(self, x: mx.array) -> mx.array:
+        if x.ndim < 3:
+            raise ValueError(
+                f"InstanceNorm expects inputs with at least 3 dimensions"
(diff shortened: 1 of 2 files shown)


zcbenz (approved):
👍

PR #3655, opened 2026-06-10
Add missing break in bool_ switch case that would result in falling through to the uint8_t case

psolanki:
## Proposed changes

Add a missing break to the switch case statement. Uncovered by a clang warning when compiling

```
binary.cpp:57:7: warning: unannotated fall-through between switch labels [-Wimplicit-fallthrough]
   57 |       case uint8:
      |       ^
```

## Checklist


zcbenz (approved):
Nice catch, thanks!

Issue #3658, opened 2026-06-11
Allow force using fused kernel in scaled_dot_product_attention


hojin12312:
## Summary

PR #3293 added a fused full-attention path for head_dim=192/256 (`steel_attention` instantiation + `use_fallback` routing at kL>16384). The core change was reviewed favorably and validated by third parties on M2 Ultra and M3 Ultra, but the PR was closed when the author stepped away from this line of work. The last open question from @angeloskath was:

> Simply put, a real world scenario would be even better. Something like `mlx_lm.generate ....` results in OOM ...

We have been running head_dim=256 models in production on Apple Silicon for months, and have exactly that evidence.

## Setup

- Mac Studio (Apple Silicon, 36 GB unified memory), MLX 0.31.x
- Runtime: oMLX (OpenAI-compatible server on MLX) with chunked prefill, prefix KV cache, and a Metal memory guard
- Model: Gemma 4 26B-A4B MoE (3-bit), head_dim=256 — 30 layers, of which **only 5 attend over the full context** (25 are sliding-window 1024)

## Chunked prefill does not solve it

Chunked prefill avoids the single qL x kL self-attention, as noted in the PR thread. But on the unfused path each chunk still materializes fp32 scores against the whole cache, per full-attention layer:

```
transient ≈ n_q_heads × chunk_len × kv_len × 4 bytes
```

This grows linearly with kv_len, so a chunk size that is safe at 16K is not safe at 130K. In practice on our setup:

| Scenario (Gemma 4 26B, 36 GB machine) | Result |
|---|---|
| Cold prefill of a ~130K-token prompt, 1024-token chunks | 226 s — completes, transient-bound |
| Default 2048-token chunks at long context | Metal OOM abort (`kIOGPUCommandBufferCallbackErrorOutOfMemory`) — we had to patch the chunk size down and add memory guards |
| Multi-turn conversation growing past ~133K, memory ceiling active | guard shrinks chunks to 32–512 tokens to stay alive → **a single turn's prefill takes 26+ minutes** |

At kv_len=130K, even a 32-token chunk allocates ~665 MB for scores alone. Decode is unaffected (`sdpa_vector` already covers head_dim=256) — the gap is exactly the full-attention path #3293 addressed.

Note this is with only 5/30 layers attending over the full context; dense head_dim=256 models hit the same wall proportionally harder.

## Control group (same hardware, same runtime)

[...]

PR #3660, opened 2026-06-11
Fused SDPA full-attention for head_dim=192/256 (revival of #3293)

hojin12312:
## Proposed changes

Revival of #3293: add head_dim=192/256 instantiations to the fused
`steel_attention` full-attention kernel and route prefill SDPA to them when
`kL > 16384`. Below that dispatch threshold the unfused path remains selected
for lower latency while its score-shaped transient is still considered
acceptable.

Decode's `sdpa_vector` path already supported head_dim=256. This PR also adds
the missing head_dim=192 vector and vector-aggregation instantiations, while
closing the full-attention/prefill dispatch gap for both 192 and 256.

The three original commits by @Thump604 are preserved with their authorship.
Additional commits port the original work to the code structure at submission
time and add focused regression coverage:

- add the missing `bd=192` steel instantiation already targeted by routing
- exclude head_dim >= 192 from the NAX branch, whose kernel family has no
  192/256 instantiations
- exercise the 192/256 full-attention route above the dispatch threshold and
  the newly instantiated head_dim=192 vector route

## Scope correction

The original version of this PR description incorrectly used Gemma 4 26B as
end-to-end evidence for this change. That attribution was wrong.

Gemma 4 26B/31B use `global_head_dim=512` in the full-attention layers that
see the long context. Their head_dim=256 layers are sliding-window layers with
bounded `kL`, so the new 192/256 large-`kL` route in this PR is not selected
for their long-context attention. A stock `mlx-lm` A/B subsequently showed
identical Gemma peak memory and prompt throughput between vanilla MLX and this
branch.

The previously reported Gemma 156K -> 184K admission increase came from a
serving-side estimator change in oMLX, not from this kernel. The fallback score
tensor was also previously described as fp32; for the measured bf16 Gemma
workload the materialized array is bf16, with fp32 softmax accumulation.

Therefore this PR should be evaluated for models whose actual long-context
full-attention head dimension is 192 or 256. Gemma 4 26B/31B are not evidence
for or against that kernel.

Detailed correction and the stock `mlx-lm` measurements:
https://github.com/ml-explore/mlx/issues/3658#issuecomment-4698010439

## Kernel microbenchmarks

Synthetic GQA shape, not labeled as a Gemma model shape:

`n_q=16, n_kv=8, head_dim=256, fp16`, M4 Max 36 GB. Peak memory includes
inputs.

| qL | kL | inputs | fused peak | unfused peak | max abs err |
| 1024 | 32768 | 264M | **272M** | 1,824M | 0.0 |
| 2048 | 65536 | 528M | **544M** | 5,696M | 0.0 |
| 2048 | 131072 | 1,040M | **1,056M** | 11,328M | 0.0 |

For this shape, fused overhead above the input allocation stays approximately
8-16 MB, while the unfused score-shaped transient grows linearly with `kL`.
The fused path is slower per kernel above the current dispatch threshold:
approximately 0.74-0.79x the unfused throughput on this machine. The threshold
preserves the lower-latency unfused route for shorter contexts. Error values
in the table are shown at the benchmark's printed precision.

## Stock mlx-lm validation

After a maintainer request, the comparison was rerun through
`python -m mlx_lm.benchmark`, without oMLX or a serving scheduler.

### Gemma 4 26B

Gemma 4 26B UD-MLX 3-bit and 4-bit were tested from 32K through 131K. Peak
memory was identical between the two builds to the benchmark's 0.001 GB
precision, and prompt throughput matched within 0.1%. Static dispatch analysis
confirms both builds use the same unfused fallback for the long-context
head_dim=512 global layers.

### Qwen3.6 35B

An independent stock `mlx-lm` smoke by @Thump604 showed the expected
memory/speed tradeoff on a genuine head_dim=256 workload:

| context | main prompt_tps | PR prompt_tps | main peak GB | PR peak GB |
|---:|---:|---:|---:|---:|
| 32K | 1005.743 | 871.489 | 41.316 | **39.882** |
| 64K | 802.929 | 624.513 | 44.201 | **40.068** |

The 131K PR run aborted with an interactivity-impacting command-buffer error,
so that row is unresolved and is not presented as a valid A/B result.

These stock benchmark arms were not a strict single-commit comparison because
the tested builds were separated by other main-branch commits. The static
dispatch change and focused route tests establish what this PR changes; the
stock benchmarks are supporting workload evidence.

Independent validation:
https://github.com/ml-explore/mlx/issues/3658#issuecomment-4697602644

## Validation

- focused route tests added in `python/tests/test_fast_sdpa.py`:
  - head_dim 192/256 full attention with `qL=9`, `kL=16385`
  - head_dim 192 vector attention
- current `python/tests/test_fast_sdpa.py`: 18 tests run, 1 skipped
- `python/tests/test_fast.py`: 23 passed before the focused tests were added;
  the kernel source is unchanged by the test-only follow-up commit
- head_dim=256 fp16 fused/unfused microbenchmark errors were reported as
  `0.0000` at the benchmark's printed precision
- pre-commit hooks passed on all files changed by this PR
- original #3293 work was also tested on M2 Ultra and M3 Ultra
- stock `mlx-lm` correction run: Gemma 4 26B 3-bit/4-bit, 32K-131K

## Tradeoff

This is intentionally not a blanket performance optimization. For the tested
head_dim=256 prefill shapes, the fused kernel trades prompt throughput for a
bounded-memory attention path. The dispatch threshold keeps the unfused path
where it is faster and reserves the fused path for long-`kL` workloads where
the growing score transient is the limiting factor.

Credit: the kernel work and original routing are @Thump604's (#3293). This PR
preserves those commits and authorship, ports the work to the submission-time
code structure, and adds the small completeness fixes and regression tests
described above.

## Checklist

  - pre-commit hooks were run successfully on every file changed by this PR
- [x] I have updated the necessary documentation (not applicable; no public API or user documentation changed)

### DIFF
--- a/mlx/backend/metal/kernels/scaled_dot_product_attention.metal
+++ b/mlx/backend/metal/kernels/scaled_dot_product_attention.metal
@@ -32,10 +32,12 @@ using namespace metal;
   instantiate_sdpa_vector(type, 64, 64)          \
   instantiate_sdpa_vector(type, 96, 96)          \
   instantiate_sdpa_vector(type, 128, 128)        \
+  instantiate_sdpa_vector(type, 192, 192)        \
   instantiate_sdpa_vector(type, 256, 256)        \
   instantiate_sdpa_vector_aggregation(type, 64)  \
   instantiate_sdpa_vector_aggregation(type, 96)  \
   instantiate_sdpa_vector_aggregation(type, 128) \
+  instantiate_sdpa_vector_aggregation(type, 192) \
   instantiate_sdpa_vector_aggregation(type, 256)
 
 instantiate_sdpa_vector_heads(float)
--- a/mlx/backend/metal/kernels/steel/attn/kernels/steel_attention.metal
+++ b/mlx/backend/metal/kernels/steel/attn/kernels/steel_attention.metal
@@ -12,6 +12,7 @@
   attention, dtype, bq, bk, bd, wm, wn, mtype, float)
 
 #define instantiate_attn_shapes_helper(iname, itype, mname, mtype)  \
+    instantiate_attn(iname, itype, 32, 16, 256, 4, 1, mname, mtype) \
     instantiate_attn(iname, itype, 32, 16, 128, 4, 1, mname, mtype) \
     instantiate_attn(iname, itype, 32, 32,  80, 4, 1, mname, mtype) \
     instantiate_attn(iname, itype, 32, 32,  64, 4, 1, mname, mtype)
--- a/mlx/backend/metal/kernels/steel/attn/kernels/steel_attention.metal
+++ b/mlx/backend/metal/kernels/steel/attn/kernels/steel_attention.metal
@@ -13,6 +13,7 @@
 
 #define instantiate_attn_shapes_helper(iname, itype, mname, mtype)  \
     instantiate_attn(iname, itype, 32, 16, 256, 4, 1, mname, mtype) \
+    instantiate_attn(iname, itype, 32, 16, 192, 4, 1, mname, mtype) \
     instantiate_attn(iname, itype, 32, 16, 128, 4, 1, mname, mtype) \
     instantiate_attn(iname, itype, 32, 32,  80, 4, 1, mname, mtype) \
     instantiate_attn(iname, itype, 32, 32,  64, 4, 1, mname, mtype)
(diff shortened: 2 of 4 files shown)


hojin12312:
I updated the PR description to correct the Gemma attribution and make the validation scope explicit.

The earlier description incorrectly treated Gemma 4's sliding head_dim=256 shape as its long-context full-attention shape. Its long-context global layers use head_dim=512, so this PR's 192/256 full-attention route is not selected there. I removed the Gemma serving results as evidence for this PR, corrected the bf16/fp32 description, and linked the stock mlx-lm A/B and detailed correction in #3658.

I also added focused regression tests in c9f8a154 for the head_dim=192/256 full-attention route above kL=16384 and the newly instantiated head_dim=192 vector route. Both focused tests and pre-commit on the changed files pass.

zcbenz:
I'm closing this PR since we decided to add a force_fused option to the API to let users make the decision, instead of providing builtin heuristics.

PR #3662, opened 2026-06-11
Fix use-after-free when a custom Metal kernel is called with different dtypes in one graph

### LINKED ISSUE #3347: [BUG] mx.fast.metal_kernel: use-after-free when multiple custom kernels compose in lazy graph

discobot:
## Proposed changes

Fixes #3347.

Calling the same `mx.fast.metal_kernel` with different input dtypes inside one lazy graph regenerates different source under the same kernel name, and the cache eviction (`clear_library`) releases a pipeline state that an uncommitted command buffer still references (mechanism detailed in #3347). This implements the approach @zcbenz suggested on #3434: put the input dtypes in the generated kernel name, so a given name always maps to exactly one source and dtype variants coexist in the cache instead of evicting each other.

Dtypes turn out not to be the only per-call fact `write_signature` bakes into the source: the address space choice (`constant` for inputs with `size() < 8`) and pass-by-reference for 0-dim inputs vary per call too. The generated name now encodes all of it — each input's dtype with a `c`/`s` suffix for those two cases, plus the output dtypes — and `max_constant_array_size` is hoisted to file scope so the naming and `write_signature` can't drift. The `clear_library` path is kept for genuinely different user sources reusing one name (covered by the existing `test_custom_kernel_caching`).

Added a regression test (`test_custom_kernel_mixed_dtypes`) that composes float16 and float32 invocations of one kernel in a single graph; with `METAL_DEVICE_WRAPPER_TYPE=1` as CI sets, the unfixed code aborts on it with the `command buffer references deallocated object` assertion. Also updated the generated-name example in the custom Metal kernels docs (names now look like `custom_kernel_myexp_float_float16_float16`).

With the fix, `test_fast.py` (24 tests) and `test_export_import.py` (16 tests) pass under Metal API validation.

## Checklist

### DIFF
--- a/docs/src/dev/custom_metal_kernels.rst
+++ b/docs/src/dev/custom_metal_kernels.rst
@@ -62,7 +62,7 @@ The full function signature will be generated using:
     so we add ``device float16_t* out``.
 * Template parameters passed using ``template``
     In the above, ``template=[("T", mx.float32)]`` adds a template of ``template <typename T>`` to the function
-    and instantiates the template with ``custom_kernel_myexp_float<float>``.
+    and instantiates the template with ``custom_kernel_myexp_float_float16_float16<float>``.
     Template parameters can be ``mx.core.Dtype``, ``int`` or ``bool``.
 * Metal attributes used in ``source`` such as ``[[thread_position_in_grid]]``
     These will be added as function arguments.
@@ -73,7 +73,7 @@ Putting this all together, the generated function signature for ``myexp`` is as 
 .. code-block:: cpp
 
   template <typename T>
-  [[kernel]] void custom_kernel_myexp_float(
+  [[kernel]] void custom_kernel_myexp_float_float16_float16(
     const device float16_t* inp [[buffer(0)]],
     device float16_t* out [[buffer(1)]],
     uint3 thread_position_in_grid [[thread_position_in_grid]]) {
@@ -84,7 +84,7 @@ Putting this all together, the generated function signature for ``myexp`` is as 
 
   }
 
-  template [[host_name("custom_kernel_myexp_float")]] [[kernel]] decltype(custom_kernel_myexp_float<float>) custom_kernel_myexp_float<float>;
+  template [[host_name("custom_kernel_myexp_float_float16_float16")]] [[kernel]] decltype(custom_kernel_myexp_float_float16_float16<float>) custom_kernel_myexp_float_float16_float16<float>;
 
 Note: ``grid`` and ``threadgroup`` are parameters to the Metal `dispatchThreads
 <https://developer.apple.com/documentation/metal/mtlcomputecommandencoder/2866532-dispatchthreads>`_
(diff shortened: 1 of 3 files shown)


zcbenz (approved):
Looks good to me, thanks!

PR #3664, opened 2026-06-12
Fix Q4_1 GGUF loading

ricky-chaoju:
Problem

`mx.load()` on a GGUF file containing **Q4_1** tensors returns silently corrupted weights. Scales and biases load correctly and every shape checks out, so nothing fails loudly — the dequantized values are just wrong.

`unpack_32_4()` in `mlx/io/gguf_quants.cpp` hard-codes a 2-byte block-header skip (`data[j + 2]`, "to skip scale bytes"). That matches Q4_0's 18-byte block (`|f16 d|16B quants|`), but Q4_1 blocks are 20 bytes with a 4-byte header (`|f16 d|f16 m|16B quants|`). `extract_q4_1_data()` passes the block start, so the two bias bytes get decoded as the first four nibbles and the last two quant bytes of every block are dropped.

## Fix

`unpack_32_4()` now takes a pointer to the first quant byte and each caller skips its own header explicitly (`data + 2` for Q4_0, `data + 4` for Q4_1). No behavior change for Q4_0/Q8_0.

## Validation

Against the `gguf` python package's reference `dequantize` (N(0,1) weights, shape `(8, 64)`):

| qtype | max abs err before | after |
|---|---|---|
| Q8_0 | 0.0020 | 0.0020 |
| Q4_0 | 0.0020 | 0.0020 |
| Q4_1 | **5.08** | 0.0020 |

Found while building GGUF support for the vLLM Metal plugin (vllm-project/vllm-metal#415).

### DIFF
--- a/mlx/io/gguf_quants.cpp
+++ b/mlx/io/gguf_quants.cpp
@@ -8,10 +8,12 @@
 
 namespace mlx::core {
 
-void unpack_32_4(uint8_t* data, int8_t* dst) {
+// Unpacks a block of 32 4-bit weights. `data` must point at the first
+// quant byte, past the block header.
+void unpack_32_4(const uint8_t* data, int8_t* dst) {
   std::fill_n(dst, 16, 0);
   for (int j = 0; j < 16; ++j) {
-    uint8_t x = (data[j + 2] & 0x0F); // j+2 to skip scale bytes.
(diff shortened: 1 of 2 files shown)


zcbenz (approved):
Looks good to me, thanks!

Issue #3659, opened 2026-06-11
[BUG] [CUDA] rope cudaGraphAddKernelNode


nastya236:
Rope kernel bug:
```python
import mlx.core as mx
import mlx.nn as nn

batch_size = 32
seq_len = 8192
heads = 32
head_dim = 128
x = mx.random.normal(shape=(batch_size, seq_len, heads * head_dim), dtype=mx.bfloat16)
rope = nn.RoPE(head_dim)  
x = x.reshape(batch_size, seq_len, heads, head_dim).transpose(0, 2, 1, 3)
rotated_x = rope(x)
mx.eval(rotated_x)
```

returns:
Traceback (most recent call last):
```python
RuntimeError: cudaGraphAddKernelNode(&node, graph_, NULL, 0, &params) failed: invalid argument

```

PR #3666, opened 2026-06-12
fix grid for large uncontiguous input

nastya236:
CUDA kernels for non-contiguous tensor operations used a 2D launch grid where y covered all rest dimensions. CUDA limits gridDim.y <= 65,535. This results in errors for large noncontiguous arrays for binary, copy and some unary ops. 
For example, for a tensor of shape (16, 8192, 32, 128) -- queries of batch size 16 for qwen 4B: 

```python
import mlx.core as mx

shape = (32, 8192, 32, 128)  
x = mx.random.normal(shape=shape)
y = mx.random.normal(shape=shape)

x = x.transpose(0, 2, 1, 3) 
y = y.transpose(0, 2, 1, 3)

z = x + y
mx.eval(z)
```

-> `RuntimeError: cudaGraphAddKernelNode(&node, graph_, NULL, 0, &params) failed: invalid argument`. Same will be for `mx.exp(x)` and https://github.com/ml-explore/mlx/issues/3659. 

I split the overflow to `gridDim.z`.

### DIFF
--- a/mlx/backend/cuda/binary/binary.cuh
+++ b/mlx/backend/cuda/binary/binary.cuh
@@ -135,7 +135,9 @@ __global__ void binary_g_nd(
   auto block = cg::this_thread_block();
   auto grid = cg::this_grid();
   IdxT index_rest =
-      grid.block_index().y * block.dim_threads().y + block.thread_index().y;
+      (grid.block_index().z * grid.dim_blocks().y + grid.block_index().y) *
+          block.dim_threads().y +
+      block.thread_index().y;
   if (index_rest >= size_rest) {
     return;
   }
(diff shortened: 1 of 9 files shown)


zcbenz (approved):
Nice fix!

I think it is worth adding a new utility function to compute the dims, like the `get_launch_args` but for noncontiguous ones.

Also the test is allocating a large buffer that crashes CI, and likely would also be too slow to run locally. I remember there is a slow test set that run separately, we should move it there, not sure if it is still a thing though @angeloskath

chrismicah:
I opened #3697 as a small follow-up against this branch to make the large-uncontiguous ROPE regression CI-friendly instead of CI-skipped. It keeps the CUDA-only launch coverage but uses zero float16 input and no reference allocation, so it should exercise the grid-z path without the memory pressure from the original random/reference test.

zcbenz:
The updated test only takes 270MB memory and 0.11s to run locally, I think we can just merge this as is.

nastya236:
Yea, I changed it to minimal size that will create the overflow, thanks @zcbenz for the review!

PR #3671, opened 2026-06-13
Add from_dlpack to the array API namespace

Ritesh102000:
Fixes part of #3484.

Adds `from_dlpack` to the MLX array API namespace (`mlx.core`), so `x.__array_namespace__().from_dlpack(...)` is available, consistent with the [Python array API standard](https://data-apis.org/array-api/latest/). It reuses the existing `create_array` DLPack / ndarray import path that `mx.array` and `asarray` already use, plus a focused test (`test_array_namespace_from_dlpack`).

Note: `asarray(..., copy=...)` (the other item in #3484) already landed in #3510, so no change was needed there.

### DIFF
--- a/python/src/ops.cpp
+++ b/python/src/ops.cpp
@@ -1796,6 +1796,20 @@ void init_ops(nb::module_& m) {
             ValueError: If ``copy`` is ``False``.
       )pbdoc");
   m.def(
+      "from_dlpack",
+      [](const nb::object& x) { return create_array(x, std::nullopt); },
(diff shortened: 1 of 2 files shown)


zcbenz:
Thanks for the PR! This is actually covered by https://github.com/ml-explore/mlx/pull/3531 which we are waiting for a nanobind release to merge, sorry I can't merge this one first because there are some git improvements that would cause merge conflicts, but still thanks!

PR #3672, opened 2026-06-13
Fix shapeless compile eliding reductions over size-1 dimensions

### LINKED ISSUE #3201: [BUG] compile(shapeless=True): Reduce returns stale values on dynamically-shaped inputs from take/gather

discobot:
## Proposed changes

Fixes #3201.

This picks up where #3202 left off: reductions over dimensions that are size 1
at trace time are elided from the graph, so shapeless replays with larger
dynamic shapes return stale values.

The reason #3202 failed CI was not a numerical regression. Once the reduction
is kept in the graph it can be a runtime identity (`out.size() == in.size()`,
e.g. a replay with size 1), and `Reduce::eval_gpu` rejects that case with
debug-only asserts in both the Metal and CUDA backends — which is why the
CPU-only jobs passed while the macOS and CUDA jobs aborted on the PR's own new
test. This PR replaces those asserts with an explicit identity path that
cast-copies the input to the output (the CPU backend already computes the
identity case correctly).

The frontend change is also restricted to non-empty reduction axes. The
unconditional version in #3202 could create a `Reduce` with no axes (0-dim
input, or an empty axes list for `all`/`any`/`max`), tripping
`assert(!axes_.empty())` in the same functions. No-axis reductions are shape
independent so eliding them stays correct.

The regression test covers all eight reductions plus `mean` on the output of
`mx.take`, in both directions: trace at size 1 and replay larger (the original
bug), and trace at size 2 and replay at size 1 (the runtime-identity path that
broke #3202). Verified against a Debug (assertions enabled) CPU build;
`test_compile`, `test_reduce`, and `test_ops` pass.

## Checklist

Notes on the pre-commit box: `clang-format` (the three C++ files) and
`black`/`isort --profile=black` (the test file) were run and report clean, but
with `clang-format` v22.1.5 rather than the v21.1.8 the repo pins, and the full
`pre-commit run --all-files` was not run, so I have left that box unchecked for
the maintainer to confirm.

This change was developed with the help of Claude Code.

### DIFF
--- a/mlx/backend/cuda/reduce.cu
+++ b/mlx/backend/cuda/reduce.cu
@@ -15,12 +15,20 @@ void Reduce::eval_gpu(const std::vector<array>& inputs, array& out) {
   assert(inputs.size() == 1);
   array in = inputs[0];
 
-  // Make sure no identity reductions trickle down here.
   assert(!axes_.empty());
-  assert(out.size() != in.size());
 
   auto& s = stream();
   auto& encoder = cu::get_command_encoder(s);
+
+  // When all the reduced axes have size 1 at runtime, which can happen with
+  // shapeless compilation, the reduction is the identity so just cast-copy
+  // the input to the output.
+  if (out.size() == in.size()) {
+    CopyType ctype =
+        in.flags().contiguous ? CopyType::Vector : CopyType::General;
+    copy_gpu(in, out, ctype, s);
+    return;
+  }
 
   if (in.size() == 0) {
     init_reduce(encoder, in, out, reduce_type_);
--- a/mlx/backend/metal/reduce.cpp
+++ b/mlx/backend/metal/reduce.cpp
@@ -952,9 +952,17 @@ void Reduce::eval_gpu(const std::vector<array>& inputs, array& out) {
   assert(inputs.size() == 1);
   array in = inputs[0];
 
-  // Make sure no identity reductions trickle down here
   assert(!axes_.empty());
-  assert(out.size() != in.size());
+
+  // When all the reduced axes have size 1 at runtime, which can happen with
+  // shapeless compilation, the reduction is the identity so just cast-copy
+  // the input to the output.
+  if (out.size() == in.size()) {
+    CopyType ctype =
+        in.flags().contiguous ? CopyType::Vector : CopyType::General;
+    copy_gpu(in, out, ctype, stream());
+    return;
+  }
 
   // Continue with reduction operation
   // Minimum of 4 bytes since we use size 4 structs for all reduce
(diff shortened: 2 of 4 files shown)


zcbenz (approved):
Nice fix, thanks!

PR #3678, opened 2026-06-14
Add array API standard aliases for existing ops

katlun-lgtm:
## Proposed changes

`array.__array_namespace__()` returns the `mlx.core` module, so array API conformance depends on `mlx.core` exposing the [array API standard](https://data-apis.org/array-api/latest/) names. Several elementwise ops are currently only available under their numpy-style names. This adds spec-conformant aliases that forward to the existing operations:

| array API name | forwards to |
| --- | --- |
| `acos`, `acosh`, `asin`, `asinh`, `atan`, `atanh` | `arccos`, `arccosh`, `arcsin`, `arcsinh`, `arctan`, `arctanh` |
| `atan2` | `arctan2` |
| `pow` | `power` |
| `bitwise_left_shift`, `bitwise_right_shift` | `left_shift`, `right_shift` |

Each alias is a thin forwarder following the existing `concat` -> `concatenate` precedent, is added to the ops documentation, and is covered by a new test (`test_array_api_aliases`) that checks output equality with the originals and presence on the `__array_namespace__()` object.

This is a focused first step toward #3484. (`asarray` / `asarray(copy=)` from that issue already landed on `main`; `from_dlpack` and other gaps remain for follow-ups.)

## Checklist

- [x] Added a test (`python/tests/test_ops.py::TestOps::test_array_api_aliases`)
- [x] Built from source and ran `python -m unittest test_ops` locally — all 140 tests pass; the existing `__array_namespace__` tests in `test_array.py` also pass

### DIFF
--- a/docs/src/python/ops.rst
+++ b/docs/src/python/ops.rst
@@ -9,8 +9,6 @@ Operations
   :toctree: _autosummary
 
    abs
-   acos
-   acosh
    add
    addmm
    all
@@ -30,21 +28,14 @@ Operations
    argsort
    array_equal
    asarray
-   asin
-   asinh
    as_strided
-   atan
-   atan2
-   atanh
    atleast_1d
    atleast_2d
    atleast_3d
    bartlett
    bitwise_and
    bitwise_invert
-   bitwise_left_shift
    bitwise_or
-   bitwise_right_shift
    bitwise_xor
    blackman
    block_masked_mm
@@ -148,7 +139,6 @@ Operations
    partition
    pad
    permute_dims
-   pow
    power
    prod
    put_along_axis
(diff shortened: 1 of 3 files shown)


zcbenz (commented):
I think we should not expose them as public API in `docs/src/python/ops.rst`, having 2 APIs doing the same thing would be very confusing.

For implementation, I wonder if it is possible to do something like:

```c++
m["acos"] = m["arccos"];
```

katlun-lgtm:
Good call on both. Pushed b9c646c9 — switched to module-level aliases via m.attr instead of re-wrapping each op, and removed them from docs/src/python/ops.rst so each op keeps a single documented name. The aliases sit at the end of init_ops after the targets register, and the tests are behavior-only so they still cover the array-API names. Thanks for the review!

zcbenz (approved):
Thanks!

PR #3679, opened 2026-06-14
Add bits and smallest_normal to finfo

katlun-lgtm:
## Proposed changes

The array API [`finfo` object](https://data-apis.org/array-api/latest/API_specification/generated/array_api.finfo.html) requires `bits` (number of bits of the type) and `smallest_normal` (smallest positive normal number), in addition to the `min`/`max`/`eps`/`dtype` that `mlx.core.finfo` already exposes. This adds both.

It is also a practical prerequisite for running the conformance suite against MLX: `hypothesis.extra.array_api` reads `finfo(dtype).bits` and `finfo(dtype).smallest_normal` when generating float arrays, and currently raises `AttributeError: 'mlx.core.finfo' object has no attribute 'bits'`, which aborts data generation for every float test.

Changes:
- Add `bits` and `smallest_normal` to `mlx::finfo` (`mlx/utils.h`, `mlx/utils.cpp`) and to the Python binding (`python/src/array.cpp`).
- Add `min()` (smallest positive normal) to the `float16_t` / `bfloat16_t` `numeric_limits` specializations in `mlx/types/limits.h`, which previously defined only `lowest()`/`max()`/`epsilon()`. `float`/`double` already inherit it from `std::numeric_limits`.
- Extend `test_finfo` to cover `bits` and `smallest_normal` (vs NumPy for float32/float16, vs known IEEE values for bfloat16, and the real-component remap for complex64).

Resulting values:

| dtype | bits | smallest_normal |
| --- | --- | --- |
| float32 | 32 | 1.1754943508222875e-38 |
| float16 | 16 | 6.103515625e-05 |
| bfloat16 | 16 | 1.1754943508222875e-38 (2^-126) |
| float64 | 64 | 2.2250738585072014e-308 |
| complex64 | 32 | 1.1754943508222875e-38 (reports the real component) |

## Checklist

- [x] Added tests (`test_finfo`)
- [x] Built from source and ran the tests locally: `test_array.TestDtypes` passes. Verified against `data-apis/array-api-tests` (`ARRAY_API_TESTS_MODULE=mlx.core`) that the previous `finfo` `AttributeError` no longer occurs and real-float elementwise tests (e.g. `test_ceil`, `test_floor`) now run and pass where they previously aborted during data generation

### DIFF
--- a/mlx/types/limits.h
+++ b/mlx/types/limits.h
@@ -30,6 +30,9 @@ struct numeric_limits<float16_t> {
   constexpr static float16_t lowest() {
     return bits_to_half(0xFBFF);
   }
+  static constexpr float16_t min() {
+    return bits_to_half(0x0400);
+  }
   static constexpr float16_t max() {
     return bits_to_half(0x7BFF);
   }
@@ -55,6 +58,9 @@ struct numeric_limits<bfloat16_t> {
  public:
   constexpr static bfloat16_t lowest() {
     return bits_to_bfloat(0xFF7F);
+  }
+  static constexpr bfloat16_t min() {
+    return bits_to_bfloat(0x0080);
   }
   static constexpr bfloat16_t max() {
     return bits_to_bfloat(0x7F7F);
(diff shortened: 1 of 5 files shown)


zcbenz (approved):
Thanks!

PR #3680, opened 2026-06-14
Add from_dlpack

katlun-lgtm:
## Proposed changes

Adds the array API [`from_dlpack`](https://data-apis.org/array-api/latest/API_specification/generated/array_api.from_dlpack.html) creation function:

```python
from_dlpack(x, /, *, device=None, copy=None) -> array
```

MLX already imports objects that implement the DLPack or Python buffer protocol through `mx.array(...)` (via `create_array` in `python/src/convert.cpp`); this exposes that capability under the array API name. The existing import path always copies the incoming (CPU) buffer, so the option handling mirrors `asarray`:

- `copy=False` raises `ValueError` (MLX cannot return a non-copying view).
- a non-default `device` raises `ValueError` (MLX arrays are not pinned to a device).
- `copy=True` / `None` and `device=None` import normally.

This is the first of the two methods named directly in #3484 (`asarray` / `asarray(copy=)` already landed on `main`).

## Checklist

- [x] Added a test (`test_array.TestArray.test_from_dlpack`)
- [x] Built from source and ran the tests locally: `test_from_dlpack`, the existing `test_dlpack`, and `test_array_namespace` pass. Verified manually that `from_dlpack` imports a NumPy array, round-trips an MLX array, is present on the `__array_namespace__()` object, and raises for `copy=False` / non-default `device`

### DIFF
--- a/docs/src/python/ops.rst
+++ b/docs/src/python/ops.rst
@@ -83,6 +83,7 @@ Operations
    floor
    floor_divide
    full
+   from_dlpack
    from_fp8
    gather_mm
    gather_qmm
--- a/python/src/ops.cpp
+++ b/python/src/ops.cpp
@@ -1796,6 +1796,49 @@ void init_ops(nb::module_& m) {
             ValueError: If ``copy`` is ``False``.
       )pbdoc");
   m.def(
+      "from_dlpack",
+      [](const nb::object& x,
+         const nb::object& device,
+         std::optional<bool> copy) {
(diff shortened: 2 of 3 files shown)


zcbenz:
Thanks for the PR! This is actually covered by https://github.com/ml-explore/mlx/pull/3531 which we are waiting for a nanobind release to merge, sorry I can't merge this one first because there are some git improvements that would cause merge conflicts, but still thanks!

PR #3681, opened 2026-06-14
Add isdtype, result_type, and can_cast

katlun-lgtm:
## Proposed changes

Adds the array API [data-type functions](https://data-apis.org/array-api/latest/API_specification/data_type_functions.html) to `mlx.core`:

- **`result_type(*arrays_and_dtypes)`** — the dtype that results from applying type promotion to the inputs. Folds the existing `promote_types` over each input's dtype; accepts arrays and/or dtypes and any number of them.
- **`can_cast(from_, to)`** — `True` when `promote_types(from_, to) == to`. Accepts an array or a dtype for `from_`.
- **`isdtype(dtype, kind)`** — tests a dtype against a dtype, a kind string (`"bool"`, `"signed integer"`, `"unsigned integer"`, `"integral"`, `"real floating"`, `"complex floating"`, `"numeric"`), or a tuple of those, built on the existing `issubdtype` categories.

These build entirely on existing primitives (`promote_types`, `issubdtype`, the dtype categories), so there are no core changes. Added to the ops docs and tested in `TestDtypes`. Part of #3484.

## Checklist

- [x] Added tests (`test_array.TestDtypes::test_result_type`, `test_can_cast`, `test_isdtype`)
- [x] Built from source and ran the tests locally: `test_array.TestDtypes` passes (7/7)

### DIFF
--- a/docs/src/python/ops.rst
+++ b/docs/src/python/ops.rst
@@ -42,6 +42,7 @@ Operations
    broadcast_arrays
    broadcast_shapes
    broadcast_to
+   can_cast
    ceil
    clip
    concat
@@ -94,6 +95,7 @@ Operations
    identity
    imag
    inner
+   isdtype
    isfinite
    isclose
    isinf
@@ -148,6 +150,7 @@ Operations
    remainder
    repeat
    reshape
+   result_type
    right_shift
    roll
    round
--- a/python/src/ops.cpp
+++ b/python/src/ops.cpp
@@ -5039,6 +5039,133 @@ void init_ops(nb::module_& m) {
           False
       )pbdoc");
   m.def(
+      "result_type",
+      [](const nb::args& arrays_and_dtypes) {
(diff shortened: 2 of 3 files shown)


katlun-lgtm:
Following the direction on #3683 (logic in `mlx/ops.{h,cpp}` rather than the binding layer): these three are currently composed in `python/src/ops.cpp` on top of `promote_types`/`issubdtype`. Happy to move them into the C++ core (with C++ tests) if you'd prefer — just let me know.

A couple of things worth your call first, since `result_type` and `isdtype` are a bit Python-API-shaped:
- `result_type` takes a variadic mix of arrays and dtypes.
- `isdtype`'s `kind` is a dtype / kind-string / tuple-of-those.

Are all three in scope, and do you want them in core or is the binding-level composition fine for these?

zcbenz (approved):
I think it makes more sense for these ops to stay as is, thanks!

PR #3682, opened 2026-06-14
Add metadata to exported functions

### LINKED ISSUE #2410: [Feature requires] Support saving meta data to mlxfn files and reading them back
# Motivation

mlx supports exporting a model with its parameters into an `mlxfn` file and later loads it into other programming languages, e.g., Python, C++, without the need for the source code of the model, which is very similar to torchscript.

That is a very nice feature.

However, mlx does not support saving meta data into the mlxfn file and later reading it from the code.

Model meta data is a very useful feature. For instance, there are many parakeet models in the mlx-community for ASR. Some of the model uses feature dim 128, while others use 80. Some of the model has 2 LSTM layers in the decoder, while others have only 1 layer.

The feature dim info is required to initialize the feature extractor, while the LSTM info is required to initialize the state array that are later sent to the decoder model.

If we can save such information in the `mlxfn` file and later read them from the code, it can greatly simplify how users use the model. They are not forced to provide the information about the model files.

onnx supports saving meta data to `*.onnx` files. See
https://onnxruntime.ai/docs/api/python/auto_examples/plot_metadata.html

There are APIs in onnxruntime to read the meta data.

---

To give you a concrete example, we have converted some parakeet models to `mxlfn` format 
https://github.com/k2-fsa/sherpa-mlx/releases
which can be used in sherpa-mlx.

The problem is that different parakeet models need to be handled differently in the code, e.g., feature dim, decoder LSTM state, etc. But such information is not available from the `mlxfn` files.

Ritesh102000:
## Proposed changes

This adds support for saving metadata along with an exported function, as requested in #2410.

- `mx.export_function(..., metadata={...})` and `mx.exporter(..., metadata={...})` store a `dict[str, str]` in the `.mlxfn` header.
- `mx.import_function(file, return_metadata=True)` returns the imported function together with its metadata.
- The metadata is also accessible from C++ through `ImportedFunction::metadata()`.

I kept the metadata as a string-to-string dictionary to match the existing `save_safetensors` / `load` metadata API. The issue mentioned possibly allowing `Union[str, int, float]` values — happy to widen it if you'd prefer, but I went with string keys/values for consistency. Passing `metadata` together with a callback raises, since it is only written when exporting to a file.

Closes #2410

### DIFF
--- a/mlx/export.cpp
+++ b/mlx/export.cpp
@@ -520,10 +520,15 @@ struct PrimitiveFactory {
   };
 };
 
-void write_header(Writer& os, int count, bool shapeless) {
+void write_header(
+    Writer& os,
+    int count,
+    bool shapeless,
+    const std::unordered_map<std::string, std::string>& metadata) {
   serialize(os, std::string(version()));
   serialize(os, count);
   serialize(os, shapeless);
+  serialize(os, metadata);
 }
 
 // A struct to hold and retrieve the graphs that are exported / imported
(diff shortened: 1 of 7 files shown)


mlx/export.cpp
```diff
@@ -1047,6 +1062,10 @@ ImportedFunction::ImportedFunction(const std::string& file)
   auto mlx_version = deserialize<std::string>(is);
   auto function_count = deserialize<int>(is);
   ftable->shapeless = deserialize<bool>(is);
+  auto metadata_pairs =
+      deserialize<std::vector<std::pair<std::string, std::string>>>(is);
```
zcbenz, line 1066: What would happen when importing functions exported from previous versions without metadata support? (There is no need to be compatible but I'm curious about the behavior)

zcbenz:

PR #3683, opened 2026-06-14
Add flip and unstack

katlun-lgtm:
## Proposed changes

Adds two array API [manipulation functions](https://data-apis.org/array-api/latest/API_specification/manipulation_functions.html) to `mlx.core`:

- **`flip(a, /, axis=None)`** — reverse the order of elements along the given axis, or all axes when `axis` is `None`. Implemented with a negative-stride `slice` (the same mechanism that backs `a[::-1]`), so no new primitive is needed.
- **`unstack(x, /, *, axis=0)`** — split an array into a sequence of arrays along an axis, with that axis removed (the inverse of `stack`). Built on `split` + `squeeze`. Returns a list, consistent with `split`.

Added to the ops docs and tested in `test_ops.py`.  Part of #3484.

## Checklist

- [x] Added tests (`test_ops.py::TestOps::test_flip`, `test_unstack`)
- [x] Built from source and ran the tests locally: `test_flip`, `test_unstack`, and the existing `test_split` pass

### DIFF
--- a/mlx/ops.cpp
+++ b/mlx/ops.cpp
@@ -645,6 +645,33 @@ array expand_dims(
   return expand_dims_impl(a, std::move(sorted_axes), s);
 }
 
+array flip(
+    const array& a,
+    const std::vector<int>& axes,
+    StreamOrDevice s /* = {} */) {
+  auto ndim = static_cast<int>(a.ndim());
+  Shape start(ndim, 0);
+  Shape stop = a.shape();
+  Shape strides(ndim, 1);
+  for (auto ax : axes) {
+    int axis = normalize_axis_index(ax, ndim, "[flip] ");
+    start[axis] = a.shape(axis) - 1;
+    stop[axis] = -a.shape(axis) - 1;
+    strides[axis] = -1;
+  }
+  return slice(a, std::move(start), std::move(stop), std::move(strides), s);
+}
+
+array flip(const array& a, int axis, StreamOrDevice s /* = {} */) {
+  return flip(a, std::vector<int>{axis}, s);
+}
+
+array flip(const array& a, StreamOrDevice s /* = {} */) {
+  std::vector<int> axes(a.ndim());
+  std::iota(axes.begin(), axes.end(), 0);
+  return flip(a, axes, s);
+}
+
 // Slice helper
 namespace {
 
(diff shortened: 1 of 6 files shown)


zcbenz (commented):
Can you add C++ versions in `mlx/ops.cpp` and then export them in `mlx/python/src/ops.cpp`?

katlun-lgtm:
Done — moved both to `mlx/ops.{h,cpp}` (`flip`/`unstack` as core C++ ops) and made the Python bindings thin wrappers. Added C++ tests in `tests/ops_tests.cpp` (alongside the existing Python tests). Thanks for the review!

python/tests/test_ops.py
```diff
@@ -1374,6 +1374,43 @@ def test_split(self):
...
+        self.assertEqual(mx.flip(a).tolist(), [4, 3, 2, 1])
+
+        a = mx.arange(6).reshape(2, 3)
+        self.assertEqual(mx.flip(a, axis=0).tolist(), [[3, 4, 5], [0, 1, 2]])
+        self.assertEqual(mx.flip(a, axis=1).tolist(), [[2, 1, 0], [5, 4, 3]])
+        self.assertEqual(mx.flip(a, axis=-1).tolist(), [[2, 1, 0], [5, 4, 3]])
+        # axis=None flips every axis.
+        self.assertEqual(mx.flip(a).tolist(), [[5, 4, 3], [2, 1, 0]])
+        # A tuple of axes.
+        self.assertEqual(mx.flip(a, axis=(0, 1)).tolist(), [[5, 4, 3], [2, 1, 0]])
```
zcbenz: For the python tests maybe test against the results of numpy or PyTorch?

mlx/ops.cpp
```diff
@@ -1560,6 +1586,33 @@ array swapaxes(
   return transpose(a, std::move(reorder), s);
 }
 
+array flip(
```
zcbenz: Can you move the implementations so their order in `ops.cpp` matches the order in `ops.h`?

katlun-lgtm:
Addressed both:
- Reordered `ops.cpp` so `flip`/`unstack` appear in the same order as in `ops.h` (`flip` now right after `expand_dims`, `unstack` after `split`).
- Reworked the Python tests to assert against NumPy (`np.flip` and per-axis `np.take`).

python/tests/test_ops.py
```diff
@@ -1374,6 +1374,42 @@ def test_split(self):
...
+
+        with self.assertRaises(ValueError):
+            mx.flip(a, axis=2)
+
+    def test_unstack(self):
+        a_np = np.arange(6).reshape(3, 2)
+        a = mx.array(a_np)
+        for axis in [0, 1, -1]:
+            parts = mx.unstack(a, axis=axis)
+            expected = [np.take(a_np, i, axis=axis) for i in range(a_np.shape[axis])]
```
zcbenz: There is a `numpy.unstack`?
https://numpy.org/doc/stable/reference/generated/numpy.unstack.html

katlun-lgtm:
Good call — switched the test to compare against `numpy.unstack` directly (consistent with the `np.flip` check).

zcbenz (approved):
Looks good to me, thanks!

PR #3684, opened 2026-06-14
Add more array API functions

katlun-lgtm:
## Proposed changes

Adds a batch of array API functions to `mlx.core`, all built on existing primitives (no core/Metal changes):

**Elementwise / utility**
- `positive(a)`, `logical_xor(a, b)`, `trunc(a)`
- `count_nonzero(a, /, *, axis=None, keepdims=False)`
- `diff(a, /, n=1, axis=-1, *, prepend=None, append=None)`

**Creation**
- `full_like(a, vals, dtype=None)`
- `empty(shape, dtype=...)`, `empty_like(a, dtype=None)` — these return zeros, since MLX does not expose uninitialized memory

**Free-function wrappers**
- `astype(a, dtype)` and `matrix_transpose(a)` mirror the existing array method/property
- `cumulative_sum` / `cumulative_prod` wrap `cumsum` / `cumprod` with the array API `axis` (flatten when `None`), `dtype`, and `include_initial` semantics

**Inspection**
- `__array_namespace_info__()` returning an object with `capabilities()`, `default_device()`, `default_dtypes()`, `devices()`, and `dtypes(kind=...)`

All added to the ops docs and tested in `test_ops.py` / `test_array.py`. Part of #3484.

## Checklist

- [x] Added tests (`test_ops.py`: `test_array_api_elementwise`, `test_diff`, `test_array_api_creation`, `test_astype_and_matrix_transpose`, `test_cumulative_sum_prod`; `test_array.py`: `test_array_namespace_info`)
- [x] Built from source and ran the tests locally: the new tests pass, and the full `test_ops.py` (144) and `test_array.py` (75) suites pass with no regressions

### DIFF
--- a/docs/src/python/ops.rst
+++ b/docs/src/python/ops.rst
@@ -29,6 +29,7 @@ Operations
    array_equal
    asarray
    as_strided
+   astype
    atleast_1d
    atleast_2d
    atleast_3d
@@ -59,19 +60,25 @@ Operations
    conv_general
    cos
    cosh
+   count_nonzero
    cummax
    cummin
    cumprod
    cumsum
+   cumulative_prod
+   cumulative_sum
    degrees
    depends
    dequantize
    diag
    diagonal
+   diff
    divide
    divmod
    einsum
    einsum_path
+   empty
+   empty_like
    equal
    erf
    erfinv
@@ -83,6 +90,7 @@ Operations
    floor
    floor_divide
    full
+   full_like
    from_fp8
    gather_mm
    gather_qmm
@@ -116,8 +124,10 @@ Operations
    logical_not
    logical_and
    logical_or
+   logical_xor
    logsumexp
    matmul
+   matrix_transpose
    max
    maximum
    mean
@@ -136,6 +146,7 @@ Operations
    partition
    pad
    permute_dims
+   positive
    power
    prod
    put_along_axis
@@ -189,6 +200,7 @@ Operations
    tri
    tril
    triu
+   trunc
    unflatten
    var
    view
--- a/python/src/array.cpp
+++ b/python/src/array.cpp
@@ -97,6 +97,53 @@ class ArrayPythonIterator {
   std::vector<mx::array> splits_;
 };
 
+// Returned by ``__array_namespace_info__()``; exposes array API inspection
+// information about the namespace.
+struct ArrayNamespaceInfo {};
+
+namespace {
+
+// Whether ``dtype`` matches an array API dtype "kind" (a dtype, a kind string,
+// or a tuple of those). A null/None kind matches everything.
+bool dtype_matches_kind(const mx::Dtype& dtype, const nb::handle& kind) {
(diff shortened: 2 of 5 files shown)


katlun-lgtm:
Same question as #3683: these are currently composed in `python/src/ops.cpp` (all pure compositions of existing ops, no new kernels). I'm happy to move them into `mlx/ops.{h,cpp}` as core ops with C++ tests, the same way as #3683 — but it's a fair number of functions, so I wanted to check scope before reworking.

A few design calls I'd like your steer on first:
- `empty` / `empty_like`: MLX has no uninitialized buffer, so these currently return zeros — happy to drop them if you'd rather not have them.
- `astype`: a free-function form of the existing `array.astype` method.
- `positive`: identity (array API completeness).
- `cumulative_sum` / `cumulative_prod`: wrap `cumsum`/`cumprod` with the array-API `axis=None` flatten + `include_initial` semantics.

Want me to move the whole set to core, or would you prefer to split/trim it first?

zcbenz:
This PR is adding too much different things so it is hard to review. Generally speaking for pure aliases it should be done like https://github.com/ml-explore/mlx/pull/3678, otherwise the ops should have C++ versions and then exposed in python. I'm not sure about `__array_namespace_info__` and I need to read about the docs later.

> * `empty` / `empty_like`: MLX has no uninitialized buffer, so these currently return zeros — happy to drop them if you'd rather not have them.

I'm good just making them alias of `zeros`/`zeros_like`.

> * `astype`: a free-function form of the existing `array.astype` method.

👍 

> * `positive`: identity (array API completeness).

It can just be a shallow copy like how `__copy__` is implemented.

> * `cumulative_sum` / `cumulative_prod`: wrap `cumsum`/`cumprod` with the array-API `axis=None` flatten + `include_initial` semantics.

I didn't check how `cumulative_sum` differences from `cumsum` yet but in principle we should not have 2 APIs having similar names but do different things.

katlun-lgtm:
Per @zcbenz's feedback I've split this PR into smaller, focused pieces. All three new PRs are open against main:

- **#3729** — `empty`, `empty_like` (pure aliases via `m.attr`), `astype`, `matrix_transpose`
  - `empty`/`empty_like` now use `m.attr("empty") = m.attr("zeros")` exactly like #3678 did for `acos` etc.

- **#3730** — `positive`, `logical_xor`, `trunc`, `count_nonzero`, `diff`, `full_like`
  - New ops with substantive implementations

- **#3731** — `cumulative_sum` and `cumulative_prod`
  - PR description explains why these can't just be aliases of `cumsum`/`cumprod`: `axis=None` semantics, `include_initial`, and `dtype` param are all incompatible additions

**Held pending review:**
Branch `katlun-lgtm:array-api-namespace-info` contains `__array_namespace_info__` + `dtype_matches_kind`. Not opening as a PR until there is clarity on the approach. Closing this original PR in favour of the above.

PR #3686, opened 2026-06-14
Add NF4 quantization mode

lyonsno:
### Summary

Adds NormalFloat4 (NF4) as a quantization mode for MLX quantized primitives and
quantized `nn` layers.

This extends the existing quantization surface with:

- `mode="nf4"` parsing and defaults
- NF4 quantize/dequantize support with float32 absmax scales
- Metal NF4 quantized matmul kernels
- `nn.QuantizedLinear(..., mode="nf4")`
- explicit errors for unsupported NF4 routes such as indexed `gather_qmm` and non-Metal NF4 `quantized_matmul`

NF4 uses 4-bit indices into the standard 16-value NormalFloat4 lookup table from QLoRA, with one float32 absmax scale per quantization group. The Metal kernels are separate from the existing FP4/FP8 kernels because NF4 dequantization is lookup-table based and uses float32 absmax scales rather than encoded byte scales.

### Details

The new mode defaults to `bits=4` and `group_size=64`, matching the common NF4
layout. The core NF4 paths also accept group sizes 32 and 128.

`mx.quantize(..., mode="nf4")` uses the graph fallback path and returns packed
`uint32` weights plus float32 scales. `mx.dequantize(..., mode="nf4")` and
`mx.quantized_matmul(..., mode="nf4")` route through the existing quantized
primitive structure, with Metal kernels for the performance-sensitive matmul
path.

The initial implementation does not support indexed NF4 `gather_qmm`; those calls now fail explicitly instead of falling through to missing Metal kernels.

NF4 `quantized_matmul` is currently Metal-only and fails explicitly on non-Metal streams. NF4 `dequantize` remains available through the graph fallback path.

### Tests

```sh
PYTHONPATH=python/tests uv run python -m unittest \
  python.tests.test_quantized \
  python.tests.test_nn.TestBase.test_quantize \
  python.tests.test_nn.TestBase.test_quantize_freeze \
  python.tests.test_nn.TestBase.test_nf4_quantized_linear \
  python.tests.test_nn.TestBase.test_nf4_quantized_embedding -v
```

```text
Ran 35 tests in 0.514s
OK
```

```sh
cmake --build build --target tests test_teardown --parallel 10
./build/tests/tests
./build/tests/test_teardown
```

```text
test cases: 256 | 256 passed | 0 failed | 0 skipped
assertions: 3453 | 3453 passed | 0 failed
test_teardown exited successfully
```

### DIFF
--- a/mlx/backend/metal/CMakeLists.txt
+++ b/mlx/backend/metal/CMakeLists.txt
@@ -79,6 +79,7 @@ if(MLX_METAL_JIT)
   make_jit_source(quantized kernels/quantized_utils.h)
   make_jit_source(fp_quantized kernels/quantized_utils.h kernels/fp8.h
                   kernels/fp4.h)
+  make_jit_source(nf4_quantized kernels/quantized_utils.h kernels/nf4.h)
   make_jit_source(gemv_masked)
 
   make_jit_source(steel/attn/kernels/steel_attention)
--- a/mlx/backend/metal/jit/includes.h
+++ b/mlx/backend/metal/jit/includes.h
@@ -26,6 +26,7 @@ const char* logsumexp();
 const char* quantized_utils();
 const char* quantized();
 const char* fp_quantized();
+const char* nf4_quantized();
 const char* ternary();
 const char* scan();
 const char* scatter_axis();
--- a/mlx/backend/metal/jit_kernels.cpp
+++ b/mlx/backend/metal/jit_kernels.cpp
@@ -1,4 +1,6 @@
 // Copyright © 2024 Apple Inc.
+#include <stdexcept>
+
 #include "mlx/backend/common/compiled.h"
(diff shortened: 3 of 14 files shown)


zcbenz:
Thanks for the pull request! Unfortunately due to maintenance burden we do not plan to add built-in GPU kernel for every possible op, especially for the new quantization modes that are not commonly accepted, for cases like this we suggest using [custom extensions](https://ml-explore.github.io/mlx/build/html/dev/extensions.html) and [custom kernels](https://ml-explore.github.io/mlx/build/html/dev/custom_metal_kernels.html) instead.

lyonsno:
Thanks for taking a look. That makes sense; I’ll close this and pursue the NF4 path as a custom extension/custom kernel package instead.

PR #3674, opened 2026-06-13
Fix tree_map_with_path for namedtuples

chrismicah:
## Summary

- Preserve namedtuple reconstruction in `tree_map_with_path`, matching `tree_map` behavior.
- Add coverage for namedtuple `tree_map_with_path` paths, return type, and rest-argument mapping.
- Add a pure-Python regression that can run without initializing MLX/Metal.

## Tests

- `python -m unittest discover python/tests -p 'test_utils_pyonly.py' -v`
- `python -m pytest python/tests/test_utils_pyonly.py -q`
- `git diff --check upstream/main...HEAD`

## Not run locally

- `python -m pytest python/tests/test_tree.py -q` blocked before collection in this headless session: `RuntimeError: [metal::load_device] No Metal device available`

PR #3687, opened 2026-06-15
Fix CUDA grid launch for large uncontiguous tensors

chrismicah:
## Summary

- split oversized CUDA grid-y launches across grid-z for general unary, binary, ternary, and copy kernels
- centralize the split calculation in `get_2d_grid_dims_split_y`
- add a CUDA-only ROPE regression for a large non-contiguous tensor shape that exercises the split-grid path without allocating a separate random/reference tensor

## Guardrail notes

This is a CUDA/Linux-scoped correctness fix, not an Apple Metal validation task. I kept it separate from Metal work and from PR #3674.

The regression is CUDA-only and uses a zero `float16` input to keep memory bounded while still requiring the launch geometry that previously exceeded the CUDA grid-y limit.

## Tests

```bash
/Users/chris.s/.pyenv/versions/3.11.7/bin/python -m black --check python/tests/test_fast.py
# 1 file would be left unchanged

git diff --check
# passed

/Users/chris.s/.pyenv/versions/3.11.7/bin/python -m pytest python/tests/test_fast.py -q
# 17 passed, 7 skipped in 0.43s
```

Note: the new regression is skipped in this local environment because CUDA is unavailable. It is intended to run in the upstream CUDA CI lanes.

### DIFF
--- a/mlx/backend/cuda/binary/binary.cuh
+++ b/mlx/backend/cuda/binary/binary.cuh
@@ -135,7 +135,9 @@ __global__ void binary_g_nd(
   auto block = cg::this_thread_block();
   auto grid = cg::this_grid();
   IdxT index_rest =
-      grid.block_index().y * block.dim_threads().y + block.thread_index().y;
+      (grid.block_index().z * grid.dim_blocks().y + grid.block_index().y) *
+          block.dim_threads().y +
+      block.thread_index().y;
   if (index_rest >= size_rest) {
     return;
   }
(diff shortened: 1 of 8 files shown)


zcbenz:
This commit comes from an open PR.

chrismicah:
Thanks, understood. I should have checked the open PR history before opening this. I do not have a distinct improvement on top of the existing open work here, so I will leave this closed and avoid reopening or duplicating it.

PR #3688, opened 2026-06-15
Fix use-after-free: buffer-cache trim can free a buffer used by an in-flight command buffer

### LINKED ISSUE #3689: Use-after-free under memory pressure: buffer-cache trim frees an MTLBuffer still used by an in-flight command buffer (kIOGPUCommandBufferCallbackErrorInvalidResource)
### Summary

Under sustained allocation + buffer-cache churn, the Metal allocator can free an `MTLBuffer` that an **in-flight command buffer still references**, crashing the process with `kIOGPUCommandBufferCallbackErrorInvalidResource`. Command buffers are created with `commandBufferWithUnretainedReferences()`, so Metal does not keep referenced buffers alive — and the buffer cache's trim releases them anyway.

Fix proposed in #3688.

### Environment

- **MLX**: consumed via `mlx-swift` 0.31.4 → mlx core `v0.31.1` (`ce45c525`). Also reproduced/inspected against current `main` (which has refactored to a `CommandEncoder` with two unretained command-buffer creation sites).
- **Device**: MacBook Pro (`Mac15,9`), **Apple M3 Max, 48 GB** unified memory
- **OS**: macOS 26.5.1 (25F80)
- **Workload**: an on-device, OpenAI-compatible inference server ([Tesseract](https://github.com/spokvulcan/tesseract)) running a 27B 4-bit **vision** model behind a tiered RAM+SSD prefix cache that restores KV state and re-runs prefill — i.e. steady allocation and cache eviction/trim while prior command buffers are still executing.

### Symptom

```
Exception Type:  EXC_CRASH (SIGABRT)
Triggered by Thread: com.Metal.CompletionQueueDispatch
  mlx::core::gpu::check_error(...)
Termination: abort() called
```

With the Metal validation layer (`MTL_DEBUG_LAYER=1 MTL_DEBUG_LAYER_ERROR_MODE=assert`):

```
-[MTLDebugCommandBuffer preCommit]: ... failed assertion
'command buffer references deallocated object which previously existed at address 0x...'
```

The IOGPU error code is `kIOGPUCommandBufferCallbackErrorInvalidResource`.

### Root cause

1. `MetalAllocator::free()` recycles a buffer into the reuse cache (`recycle_to_cache`) as soon as its refcount hits 0 — even if a command buffer that used it is still in flight (e.g. after `async_eval`).
2. When `MetalAllocator::malloc()` finds the cache over `max_pool_size_` (or memory pressure past `gc_limit_`), it calls `buffer_cache_.release_cached_buffers(...)`, whose free callback runs the real `buf->release()` (and `residency_set_.erase(buf)`).
3. Buffers use `MTL::ResourceHazardTrackingModeUntracked`, and command buffers are created with `commandBufferWithUnretainedReferences()`, so nothing keeps the `MTLBuffer` alive — `release()` deallocates the allocation out from under the GPU.
4. The in-flight command buffer then fails completion with `kIOGPUCommandBufferCallbackErrorInvalidResource`.

(All in `mlx/backend/metal/allocator.cpp` and `mlx/backend/metal/device.cpp`.)

### Reproduction

The bug is a **timing-sensitive race**: it needs malloc-driven cache trimming to release a buffer in the window between a command buffer being committed and completing. Anything that slows that window down hides it — running under a debugger, or raising the cache limit so the trim never fires (both make it disappear, which is itself diagnostic).

Reliable trigger in our app:

1. Send a large multi-image request to a 27B vision model through the prefix-cache server (cold request — populates + persists a KV snapshot).
2. Re-send the same request so the **warm** path restores the snapshot from SSD and re-runs prefill (sustained allocation + cache churn).
3. Crashes within ~16 s on the warm request. Concurrency makes it crash on the first round.

Repro harness (replays recorded requests against the live server and watches for the abort): [`scripts/repro-image-cache-crash.sh`](https://github.com/spokvulcan/tesseract/blob/main/scripts/repro-image-cache-crash.sh) in the linked repo.

I'm happy to put together a **minimal standalone MLX repro** (allocate under memory pressure + `async_eval` + cache trim while command buffers are in flight) if that would help — just let me know.

### Fix

#3688 — use retained references (`commandBuffer()`) at the command-buffer creation site(s) so Metal holds referenced buffers until completion; the cache trim's `release()` then only drops MLX's own reference. With the fix the previously-crashing workload runs to completion, and warm prefix-cache restores are *faster* than cold runs (confirming this is a correctness fix, not a slowdown that masks the race).

If a lower-overhead approach is preferred over global retained references (e.g. deferring release of cached buffers still referenced by in-flight command buffers), I'm glad to rework the PR along those lines.

spokvulcan:
Resolves #3689

## Summary

Command buffers are created with `commandBufferWithUnretainedReferences()`, so Metal does not keep referenced `MTLBuffer`s alive — MLX owns their lifetime. Under memory pressure the allocator's buffer-cache trim can free a buffer that an in-flight command buffer still references, causing a GPU-side use-after-free.

## Root cause

1. `MetalAllocator::free()` recycles a buffer into the reuse cache as soon as its refcount hits 0 — even if a command buffer that uses it is still in flight (e.g. after `async_eval`).
2. When `MetalAllocator::malloc()` finds the cache over `max_pool_size_` (or under memory pressure past `gc_limit_`), it calls `buffer_cache_.release_cached_buffers(...)`, whose free callback runs the real `buf->release()` (and `residency_set_.erase(buf)`).
3. Buffers use `MTL::ResourceHazardTrackingModeUntracked`, and command buffers are unretained, so nothing keeps the `MTLBuffer` alive. Releasing it deallocates the allocation out from under the GPU.
4. The in-flight command buffer then fails completion with `kIOGPUCommandBufferCallbackErrorInvalidResource`, surfacing as a `SIGABRT` on `com.Metal.CompletionQueueDispatch` via `mlx::core::gpu::check_error`.

## Reproduction

Reproduces reliably on a 27B vision model behind a tiered prefix cache that restores KV state and re-runs prefill — steady allocation + cache churn while prior command buffers are still in flight. It is timing-sensitive: running under a debugger or with a much larger cache limit hides it (the larger limit just suppresses the trim), which is the signature of this race.

## Fix

Use retained references at both command-buffer creation sites in `CommandEncoder` (`commandBuffer()` instead of `commandBufferWithUnretainedReferences()`), so Metal holds referenced buffers until the command buffer completes. The cache trim's `release()` then only drops MLX's own reference; Metal keeps the allocation alive until the GPU is done with it.

This trades a small amount of per-command-buffer retain/release overhead for correctness under memory pressure. If a lower-overhead approach is preferred, an alternative would be to defer releasing cached buffers that are still referenced by in-flight command buffers — happy to rework along those lines.

## Testing

With this change, the previously-crashing workload runs to completion (and warm prefix-cache restores remain *faster* than cold runs, confirming the fix is correctness, not a slowdown that masks the race).

### DIFF
--- a/mlx/backend/metal/device.cpp
+++ b/mlx/backend/metal/device.cpp
@@ -276,7 +276,18 @@ CommandEncoder::CommandEncoder(
     queue_->addResidencySet(residency_set.mtl_residency_set());
   }
   debug_set_stream_queue_label(queue_.get(), index);
-  buffer_ = NS::RetainPtr(queue_->commandBufferWithUnretainedReferences());
+  // Use retained references so Metal keeps every referenced MTLBuffer alive
+  // until the command buffer completes. With unretained references the
+  // allocator's buffer cache can deallocate a buffer that an in-flight command
+  // buffer still references: when malloc() finds the cache over its limit it
+  // calls buffer_cache_.release_cached_buffers(), whose callback runs
+  // buf->release() on a cached buffer whose command buffer has not yet
+  // completed. With ResourceHazardTrackingModeUntracked buffers that frees the
+  // underlying allocation out from under the GPU, which fails completion with
+  // kIOGPUCommandBufferCallbackErrorInvalidResource. Retained references let
+  // Metal hold the buffer until the command buffer finishes, so release() from
+  // the cache trim only drops MLX's own ref.
+  buffer_ = NS::RetainPtr(queue_->commandBuffer());
 }
 
 CommandEncoder::~CommandEncoder() {
@@ -510,7 +521,10 @@ void CommandEncoder::commit(std::function<void()> completion) {
         }
       });
   buffer_->commit();
-  buffer_ = NS::RetainPtr(queue_->commandBufferWithUnretainedReferences());
+  // Retained references (see CommandEncoder ctor): keep referenced buffers
+  // alive until the command buffer completes so the allocator's cache trim
+  // cannot free a buffer still in use by the GPU.
+  buffer_ = NS::RetainPtr(queue_->commandBuffer());
   buffer_ops_ = 0;
   buffer_sizes_ = 0;
 }


zcbenz:
> Under memory pressure the allocator's buffer-cache trim can free a buffer that an in-flight command buffer still references, causing a GPU-side use-after-free.

That is very unlikely the root cause of crash you saw, and we can't accept this PR because it would introduce heavy performance penalty. Usually this kind of crash was caused by custom extensions, and and please reference https://github.com/ml-explore/mlx/pull/3462 on how you could find out the root cause on your side.

It is also possible that this could be a bug of MLX, and in that case a reproducible script with mlx and mlx-lm would be appreciated.

PR #3692, opened 2026-06-15
fix: avoid NaN gradients for masked sqrt at zero

### LINKED ISSUE #3668: [BUG] VJP mx.maximum(rho, eps) is NaN at rho=0 even though the forward value is finite
**Describe the bug**
VJP does not match the output of JVP and finite difference. 0 * NaN = NaN trap: gradient through mx.maximum(rho, eps) is NaN at rho=0 even though the forward value is finite. 

**To Reproduce**

Include code snippet
```
"""Minimal MLX reproducer: grad of `mx.maximum(sqrt(x^2 + y^2), const)` is NaN
at the origin (x=y=0), even though the forward value is finite.

Run:
    python repro_max_sqrt_nan.py

Expected output: the grad at the (0,0) sample is NaN.

Tested on MLX 0.31.2 / Apple Silicon GPU.

The forward clamp via `mx.maximum` works fine, but
the gradient still flows through the `sqrt(0)` branch where the local
derivative is mathematically `0/0 = NaN`. The `0 * NaN = NaN` trap then
poisons the gradient at every input that produces rho=0.


"""
import mlx.core as mx

mx.set_default_device(mx.gpu)


def f_maximum_outside_sqrt(xy):
    """Clamp via mx.maximum AFTER sqrt — gradient is NaN at (0,0)."""
    x, y = xy[..., 0], xy[..., 1]
    rho = mx.sqrt(x * x + y * y)
    rho_safe = mx.maximum(rho, mx.array(1e-10, dtype=rho.dtype))
    return mx.sum(rho_safe)


def f_where_outside_sqrt(xy):
    """Clamp via mx.where AFTER sqrt — also NaN at (0,0)."""
    x, y = xy[..., 0], xy[..., 1]
    rho = mx.sqrt(x * x + y * y)
    rho_safe = mx.where(rho > 1e-10, rho, mx.array(1e-10, dtype=rho.dtype))
    return mx.sum(rho_safe)


def f_eps_inside_sqrt(xy):
    """Add small epsilon INSIDE sqrt — bounds rho away from zero, gradient is fine."""
    x, y = xy[..., 0], xy[..., 1]
    rho = mx.sqrt(x * x + y * y + 1e-20)
    return mx.sum(rho)


# Three sample points: chief ray at origin (0,0), and two off-axis rays.
xy = mx.array([
    [0.0, 0.0],
    [0.5, 0.3],
    [-0.7, 0.1],
], dtype=mx.float32)

print(f"input xy:\n{xy}\n")

for name, f in [
    ("mx.maximum(sqrt, const) outside",  f_maximum_outside_sqrt),
    ("mx.where(>eps, sqrt, const)",       f_where_outside_sqrt),
    ("sqrt(x*x + y*y + 1e-20)  [fix]",    f_eps_inside_sqrt),
]:
    val = f(xy)
    g = mx.grad(f)(xy)
    print(f"{name}:")
    print(f"  forward = {float(val):.6f}    (always finite)")
    print(f"  grad    = {g}")
    print()

```

**Expected behavior**
Match forward gradient.

Lyxot:
Fixes #3668

## Proposed changes

Fixes NaN gradients when `sqrt` receives a zero cotangent at singular inputs.

For expressions like:

```python
rho = mx.sqrt(x * x + y * y)
rho_safe = mx.maximum(rho, 1e-10)
```

the forward value is finite at `x = y = 0`, but the VJP previously evaluated the inactive `sqrt(0)` path as `0 / 0`, producing NaN gradients.

This updates `Sqrt::vjp` to substitute a safe denominator when the incoming cotangent is zero, so masked-out paths contribute zero instead of NaN. The change is local to `Sqrt::vjp` and preserves the singular gradient when the cotangent is nonzero.

## Tests

- `DEVICE=cpu python -m pytest -q python/tests/test_autograd.py::TestAutograd::test_masked_sqrt_grad_is_finite`
- `DEVICE=gpu python -m pytest -q python/tests/test_autograd.py::TestAutograd::test_masked_sqrt_grad_is_finite`
- `python -m pytest -q python/tests/test_autograd.py python/tests/test_ops.py::TestOps::test_where python/tests/test_ops.py::TestOps::test_maximum python/tests/test_ops.py::TestOps::test_minimum`

## Checklist

### DIFF
--- a/mlx/primitives.cpp
+++ b/mlx/primitives.cpp
@@ -5374,14 +5374,13 @@ std::vector<array> Sqrt::vjp(
   assert(cotangents.size() == 1);
   auto dtype = primals[0].dtype();
   if (recip_) {
-    auto scaled_cotan =
-        multiply(array(-0.5, dtype), cotangents[0], stream());
+    auto scaled_cotan = multiply(array(-0.5, dtype), cotangents[0], stream());
     auto mask =
         not_equal(cotangents[0], array(0, cotangents[0].dtype()), stream());
     auto output = where(mask, outputs[0], array(1, dtype), stream());
     auto input = where(mask, primals[0], array(1, dtype), stream());
-    auto out = divide(
-        multiply(scaled_cotan, output, stream()), input, stream());
+    auto out =
+        divide(multiply(scaled_cotan, output, stream()), input, stream());
     return {out};
   } else {
     auto scaled_cotan = multiply(array(0.5, dtype), cotangents[0], stream());
--- a/mlx/primitives.cpp
+++ b/mlx/primitives.cpp
@@ -5374,16 +5374,21 @@ std::vector<array> Sqrt::vjp(
   assert(cotangents.size() == 1);
   auto dtype = primals[0].dtype();
   if (recip_) {
-    auto one_over_x_root_x = divide(outputs[0], primals[0], stream());
-    return {multiply(
-        multiply(array(-0.5, dtype), cotangents[0], stream()),
-        one_over_x_root_x,
-        stream())};
+    auto scaled_cotan =
+        multiply(array(-0.5, dtype), cotangents[0], stream());
+    auto mask =
+        not_equal(cotangents[0], array(0, cotangents[0].dtype()), stream());
+    auto output = where(mask, outputs[0], array(1, dtype), stream());
+    auto input = where(mask, primals[0], array(1, dtype), stream());
+    auto out = divide(
+        multiply(scaled_cotan, output, stream()), input, stream());
+    return {out};
   } else {
-    return {divide(
-        multiply(array(0.5, dtype), cotangents[0], stream()),
-        outputs[0],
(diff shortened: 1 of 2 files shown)


zcbenz (approved):
Looks good to me, thanks!

Lyxot:
Pushed a formatting commit

angeloskath:
I am sorry to say I will close this.

First, all frameworks behave the same way. Secondly, I do think that the behavior is correct. The derivative of `sqrt(x)` is not really defined at 0 so if you are using `sqrt(x)` and `x` can be 0 you can't really get normal gradients.

That being said all you have to do is put the mask before the `sqrt` and you 're done. ie change the code to

```python
rho_squared = x * x + y * y
rho_squared_safe = mx.maximum(rho_squared, 1e-20)
rho_safe = mx.sqrt(rho_squared_safe)
```

PR #3693, opened 2026-06-15
Fix dimension count in orthogonal initializer error message

Pablosinyores:
The error raised by `nn.init.orthogonal` for a non-2D input never reported the actual number of dimensions. The message was split across two string literals where only the first carried the `f` prefix:

```python
raise ValueError(
    f"Orthogonal initialization requires a 2D array but got"
    " a {a.ndim}D array."
)
```

The `{a.ndim}` lives on the plain (non-f) string, so it was printed verbatim:

```
Orthogonal initialization requires a 2D array but got a {a.ndim}D array.
```

Moving the `f` prefix onto the segment that contains the placeholder fixes the interpolation:

```
Orthogonal initialization requires a 2D array but got a 3D array.
```

Added an assertion to `test_orthogonal` that a 3D input reports `3D array` in the message.

### DIFF
--- a/python/mlx/nn/init.py
+++ b/python/mlx/nn/init.py
@@ -415,8 +415,8 @@ def orthogonal(
     def initializer(a: mx.array) -> mx.array:
         if a.ndim != 2:
             raise ValueError(
-                f"Orthogonal initialization requires a 2D array but got"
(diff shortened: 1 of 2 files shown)


zcbenz (approved):
Thanks!

PR #3694, opened 2026-06-15
Make step activation inclusive at the threshold

Pablosinyores:
The `step` activation documents its boundary as inclusive:

```
step(x) = 0  if x < threshold
          1  if x >= threshold
```

Both `nn.step` and the `nn.Step` module render this `x >= threshold` form. The implementation, however, used a strict comparison:

```python
return mx.where(x > threshold, 1, 0)
```

so an input exactly equal to the threshold returned `0` instead of `1`. With the default `threshold=0` this means `step(0.0)` evaluates to `0`, which contradicts the documented behavior.

Switched to `x >= threshold` so the value at the boundary is included, matching the docstring. Added a `test_step` case covering the boundary for both the functional and module forms.

### DIFF
--- a/python/mlx/nn/layers/activations.py
+++ b/python/mlx/nn/layers/activations.py
@@ -237,7 +237,7 @@ def step(x: mx.array, threshold: float = 0.0):
         threshold: The value to threshold at.
     """
 
-    return mx.where(x > threshold, 1, 0)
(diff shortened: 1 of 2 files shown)


angeloskath (approved):
Thanks!

PR #3697, opened 2026-06-15
Make CUDA grid regression CI-friendly

chrismicah:
## Summary

This is a follow-up to #3666, not a replacement for it.

- keep the large-uncontiguous ROPE regression on the CUDA path
- remove the CI skip by making the test CUDA-only instead of CI-skipped
- use a zero float16 input and assert zero output shape/value, avoiding the large random input and reference ROPE allocation

## Tests

ssssss.............s....                                                 [100%]
17 passed, 7 skipped in 0.37s

Note: the updated regression is skipped locally because CUDA is unavailable here. It is intended to run in the CUDA CI lanes for #3666.

### DIFF
--- a/python/tests/test_fast.py
+++ b/python/tests/test_fast.py
@@ -1,7 +1,6 @@
 # Copyright © 2023-2024 Apple Inc.
 
 import math
-import os
 import unittest
 
 import mlx.core as mx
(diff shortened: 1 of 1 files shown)


nastya236:
Unfortunately the test does not make sense because zero input will give zero output even if:
  - The kernel never runs
  - The strides are computed completely wrong 
  - The rotation matrix is wrong
I think skipping this test in CI is enough and in the future if we realise it is too slow, we will create separate slow pull.

chrismicah:
You're right — the zero-input version was a bad test. It could pass without proving the copy/ROPE path actually touched the right elements or applied the rotation correctly.

I updated the branch with a sentinel-based version instead:

- keeps the same large transposed shape so the non-contiguous copy path still crosses the CUDA grid-y limit
- writes nonzero sentinel pairs, including one at the far end of the rest dimensions
- samples the resulting ROPE output against the analytic expected values instead of materializing the full reference tensor

Local verification here:

ssssss.............s....                                                 [100%]
17 passed, 7 skipped in 0.51s

I am leaving this PR closed since you said the CI skip is enough for #3666. Just wanted to correct the branch so it no longer contains the nonsensical zero assertion.

Issue #3645, opened 2026-06-09
[BUG][CUDA] Segmentation Fault with CUDA on Jetson/Tegra


gmondada:
**Description**

MLX on Jetson/Tegra (on Linux, with CUDA backend) sometimes causes Segmentation Faults:

```
signal SIGSEGV: invalid permissions for mapped object
```

This happens on memory allocated by MLX through `cudaMallocManaged()`. The memory is mapped (visible in /proc/xxx/maps) but has no read or write access.

Note that this CUDA platform has `concurrentManagedAccess = 0`, a.k.a. "Limited unified memory support" (see [NVIDIA doc](https://docs.nvidia.com/cuda/cuda-programming-guide/04-special-topics/unified-memory.html#um-legacy-devices)). The CUDA runtime is known to remove RW access from the CPU when such an access could violate the memory model.

Honestly, I'm not sure if this is a bug or if MLX does not support this memory model.

**Additional notes**

* I often see this crash when MLX tries to set MLXArray data right after allocation. Adding `cudaDeviceSynchronize()` just after `cudaMallocManaged()` often "solves" the problem.
* I also see the code crash while trying to read memory from an MLXArray.
* Once the problem appears, it is reproducible every time.
* Nothing changes whether I use CUDA 12.6 or CUDA 13.2.
* I see the same problem whether I use Swift or Python.

**To Reproduce**

I saw the problem by running mlx-lm and was not able to shrink my example down to just a few MLX calls. Example:

```python
import mlx.core as mx
from mlx_lm import load, generate

with mx.stream(mx.gpu):
    model, tokenizer = load("mlx-community/Llama-3.2-3B-Instruct-4bit")
    prompt = "What's MLX?"
    print(f"Prompt: {prompt}")
    tokens = tokenizer.encode(prompt)
    print(f"Prompt tokens: {tokens}")
    response = generate(model, tokenizer, prompt=tokens, max_tokens=512, verbose=False)
    print(f"Response tokens: {response}")

```

Even if it's not a classic use case, the easiest way to reproduce the problem with a few instructions, is to run in CPU mode:

```python
import mlx.core as mx

with mx.stream(mx.cpu):
    arr = mx.array([2, 4, 6, 8], dtype=mx.int32).reshape(2, 2)
    print(arr)
```

**Platform**
- Jetson Orin Nano
- NVIDIA Jetpack 7.2
- Linux Ubuntu 5.15.185-tegra
- aarch64

[...]

PR #3701, opened 2026-06-16
[CUDA] Disable managed memory on Tegra

gmondada:
## Proposed changes

Currently, on Windows and WSL, managed memory support is disabled when concurrentManagedAccess = 0.
This PR wants to disable it on all platforms where concurrentManagedAccess = 0, not just on Windows.
This addresses segmentation fault problems seen on NVIDIA Tegra. See https://github.com/ml-explore/mlx/issues/3645.
In fact, Windows, WSL, and Tegra share the same unified memory paradigm called "Limited unified memory support", which is identified by concurrentManagedAccess = 0.
See
- https://docs.nvidia.com/cuda/cuda-programming-guide/02-basics/understanding-memory.html#table-unified-memory-levels
- https://docs.nvidia.com/cuda/cuda-programming-guide/04-special-topics/unified-memory.html#um-legacy-devices

## Checklist

### DIFF
--- a/mlx/backend/cuda/allocator.cpp
+++ b/mlx/backend/cuda/allocator.cpp
@@ -33,9 +33,6 @@ bool supports_managed_memory() {
     int device_count = gpu::device_count();
     for (int i = 0; i < device_count; ++i) {
       auto& d = cu::device(i);
-      if (!d.managed_memory()) {
-        return false;
-      }
       // Empirically on Windows (and WSL) if there is no concurrentManagedAccess
       // the managed memory also does not work.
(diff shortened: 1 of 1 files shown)


gmondada:
@Joannis 👆

mlx/backend/cuda/allocator.cpp
zcbenz: This check can also be removed.

gmondada: Good point, removed

zcbenz (approved):
Thanks!

PR #3705, opened 2026-06-16
Make gemv JIT compilable

angeloskath:
Big diff but basically moves the gemv to be JIT compilable in metal. Part of a quest started from #3580 to address the lack of split-k on gemv (more PRs will be coming).


angeloskath:
The metallib was already small when JIT is on but still we go from 3.3MB to 1.1MB after moving gemv to JIT.

zcbenz (approved):
👍

PR #3706, opened 2026-06-17
[CUDA] JIT-compile qmm_sm80/qmm_sm90/gather_gemm

### LINKED ISSUE #3567: JIT compilation for CUTLASS/CuTe kernels

zcbenz:
Use JIT compilation for the remaining cutlass/cute kernels, close https://github.com/ml-explore/mlx/issues/3567.

The cute kernels are simple to convert just like https://github.com/ml-explore/mlx/pull/3576, but the cutlass-based `qmm_sm90` kernel is annoying because we still have to instantiating the `cutlass::gemm::kernel::GemmUniversal` class in host for all supported type parameters to get the smem size and launch grid dims, which slows down compilation a lot.


angeloskath (approved):
👍

PR #3709, opened 2026-06-17
Optimize low-channel 3D convolutions via 2D slices

chrismicah:
## Summary
- Optimize eligible low-channel 3D convolutions on GPU by decomposing the depth dimension into 2D convolution slices.
- Keep the optimization narrowly gated to low input/output channels and enough depth work, where local Metal benchmarks show speedups.
- Add coverage comparing the decomposed result with an explicit sum of 2D convolutions.

## Local validation
- PYTHONPATH=python:python/tests python -m unittest test_conv.TestConv.test_conv3d_depth_decomposition
- python setup.py build_ext --inplace

## Local benchmark notes
Built both upstream main and this branch from source on Apple Silicon / Metal. Representative best-of-run timings:

shape: B=1 D=8 H=32 W=32 C=4 O=8 k=3x3x3; upstream 0.3408 ms; branch 0.2688 ms; speedup 1.27x
shape: B=1 D=16 H=64 W=64 C=4 O=8 k=3x3x3; upstream 0.5058 ms; branch 0.3059 ms; speedup 1.65x
shape: B=1 D=32 H=96 W=96 C=4 O=8 k=3x3x3; upstream 1.8684 ms; branch 0.8960 ms; speedup 2.09x
shape: B=2 D=8 H=64 W=64 C=4 O=8 k=3x3x3; upstream 0.4602 ms; branch 0.2968 ms; speedup 1.55x
shape: B=2 D=16 H=96 W=96 C=4 O=8 k=3x3x3; upstream 1.7632 ms; branch 0.9387 ms; speedup 1.88x
shape: B=2 D=32 H=96 W=96 C=4 O=8 k=3x3x3; upstream 3.7208 ms; branch 1.7842 ms; speedup 2.09x

Higher-channel cases are intentionally excluded because the generic 3D convolution path was faster there in local tests.

### DIFF
--- a/mlx/ops.cpp
+++ b/mlx/ops.cpp
@@ -4358,12 +4358,74 @@ array conv_general(
       padding_hi,
       kernel_dilation,
       input_dilation);
+  auto stream = to_stream(s);
+
+  auto all_equal = [](const std::vector<int>& values, int target) {
+    return std::all_of(values.begin(), values.end(), [target](int value) {
+      return value == target;
+    });
+  };
+
+  const bool low_channel_3d =
+      spatial_dims == 3 && in.shape(4) <= 4 && wt.shape(0) <= 8;
+  const bool enough_depth_work =
+      in.shape(0) > 1 || out_shape[1] >= 14 || in.shape(2) <= 32;
+  if (stream.device == Device::gpu && low_channel_3d && groups == 1 && !flip &&
+      stride[0] == 1 && all_equal(padding_lo, 0) && all_equal(padding_hi, 0) &&
+      all_equal(kernel_dilation, 1) && all_equal(input_dilation, 1) &&
+      wt.shape(1) > 1 && enough_depth_work) {
+    const int out_depth = out_shape[1];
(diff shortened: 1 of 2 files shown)


zcbenz (changes requested):
Thanks for the PR, but generally we should not do optimization like this in `ops.cpp`, it is something that should be done in `Convolution::eval_gpu`. And while the 2x speedup is awesome, I'm not sure if the tested shapes are commonly used so we might not want to complicate the code for that, improvements for shapes used by real life model would be needed for justification.

chrismicah:
Thanks for the review. I agree with both points: this should not live in generic ops.cpp wiring, and the benchmark shapes here are not enough justification without real model/workload evidence. I am going to close this version rather than keep it open in the wrong layer. If I come back to it, I will rework from Convolution::eval_gpu with representative model-profile evidence first.

chrismicah:
Closing this version because the optimization is in the wrong layer and lacks real-workload justification. I will not push more changes here without a backend-layer design and profile evidence.

PR #3712, opened 2026-06-17
Fix optimizer epsilon validation message and AdaDelta eps doc

Pablosinyores:
### Epsilon validation message

`RMSprop`, `Adagrad` and `AdaDelta` validate epsilon with `if self.eps < 0.0`, so `eps == 0` is accepted — the same contract as `Adamax` (`if not 0.0 <= eps`). But their error message claimed epsilon "should be >0":

```python
if self.eps < 0.0:
    raise ValueError(
        f"RMSprop epsilon should be >0, {self.eps} was provided instead"
    )
```

The message contradicts the guard (and `Adamax`, which already says ">=0"). This reword changes the three messages to ">=0" so they match the actual check. No behavior change.

### AdaDelta default epsilon docstring

`AdaDelta`'s docstring documented the default epsilon as `1e-8`, but the constructor default is `1e-6`:

```python
eps: float = 1e-6,
```

Corrected the docstring to `1e-6`.

### Test

Added `test_epsilon_validation` asserting the negative-eps message matches `>=0` and that `eps == 0` is permitted for `RMSprop`, `Adagrad` and `AdaDelta`. The full `TestOptimizers` suite passes.

### DIFF
--- a/python/mlx/optimizers/optimizers.py
+++ b/python/mlx/optimizers/optimizers.py
@@ -328,9 +328,9 @@ class RMSprop(Optimizer):
             raise ValueError(
                 f"RMSprop alpha should be >=0, {self.alpha} was provided instead"
             )
-        if self.eps < 0.0:
+        if self.eps <= 0.0:
             raise ValueError(
-                f"RMSprop epsilon should be >=0, {self.eps} was provided instead"
+                f"RMSprop epsilon should be >0, {self.eps} was provided instead"
             )
 
     def init_single(self, parameter: mx.array, state: dict):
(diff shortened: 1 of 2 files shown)


aeiwz (commented):
> eps == 0 is unsafe for these optimizers. Their accumulators start at zero, so a zero gradient makes RMSprop/Adagrad compute 0 / 0, while AdaDelta computes sqrt(0) / sqrt(0) * 0; all
  > produce NaN parameters. I reproduced this for all three classes. Please reject eps <= 0, retain the >0 message, and change this test to assert that zero raises.

  The AdaDelta default documentation correction is valid.

Pablosinyores:
You're right, thanks for the repro. The accumulators start at zero, so eps == 0 gives 0 / 0 (or sqrt(0) / sqrt(0)) -> NaN; the `>0` message was the correct intent and the guard (`eps < 0.0`) was the actual bug, not the other way round. Pushed 6def5db: the guard now rejects `eps <= 0` for all three, the message stays `>0`, and the test asserts `eps == 0` raises. The AdaDelta default-eps doc correction (1e-8 -> 1e-6) stays.

zcbenz (approved):
Thanks!

PR #3714, opened 2026-06-18
fix(metal): end encoding before addCompletedHandler for macOS 26 compatibility

vernonstinebaker:
## Proposed changes

On macOS 26.4.1, calling `-[MTLCommandBuffer addCompletedHandler:]` while a
`MTLComputeCommandEncoder` is still active on the same buffer triggers
`MTLReportFailure` → `abort()`. Apple's Metal framework now asserts that
completion handlers must not be added while an encoder is active.

This affects two code paths:

1. **eval.cpp else branch**: `addCompletedHandler` called without first ending
   the encoder (the if branch already called `end_encoding` correctly).

2. **fence.cpp fast paths** (`Fence::wait`, `Fence::update`): `addCompletedHandler`
   called after dispatching fence kernels without ending the encoder first.

### Fix

Adds `end_encoding()` calls before `addCompletedHandler` in all three locations.

### Testing

This is an OS-specific Metal framework assertion added in macOS 26.4.1. A C++
unit test cannot detect this regression on older OS versions. The existing
test suite passes on macOS 26.4.1 with this fix; the crash is reliably
reproducible under model inference (eval/fence code paths are hit on
every MLX operation) and is eliminated by these changes.

Two new GPU tests have been added to `test_eval.py` that exercise the
affected code paths:
- `test_eval_small_gpu_ops_else_branch`: small GPU operations that keep the
  command buffer under the commit threshold, hitting the eval else-branch
- `test_fence_paths_via_multistream_eval`: multi-stream GPU operations that
  trigger internal fence synchronization

## Checklist

### DIFF
--- a/mlx/backend/metal/eval.cpp
+++ b/mlx/backend/metal/eval.cpp
@@ -63,6 +63,7 @@ void eval(array& arr) {
       scheduler::notify_task_completion(s);
     });
   } else {
+    encoder.end_encoding();
     command_buffer->addCompletedHandler(
         [buffers = std::move(buffers)](MTL::CommandBuffer* cbuf) {});
   }
--- a/mlx/backend/metal/fence.cpp
+++ b/mlx/backend/metal/fence.cpp
@@ -78,6 +78,7 @@ void Fence::wait(Stream stream, const array& x) {
   compute_encoder.set_bytes(f.count, 1);
   compute_encoder.dispatch_threads(kernel_dims, kernel_dims);
 
+  compute_encoder.end_encoding();
   compute_encoder.get_command_buffer()->addCompletedHandler(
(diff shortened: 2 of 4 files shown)


zcbenz:
> On macOS 26.4.1, calling `-[MTLCommandBuffer addCompletedHandler:]` while a
> `MTLComputeCommandEncoder` is still active on the same buffer triggers
> `MTLReportFailure` → `abort()`. Apple's Metal framework now asserts that
> completion handlers must not be added while an encoder is active.

I don't think we have observed or received any report on this, if you could make a independent repro with Metal I would be happy to report to the team.

vernonstinebaker:
After upgrading to macOS 27 beta I'm unable to reproduce the results. I'm going to close this Issue since I'm not able to reproduce the Issue (and don't want to roll back to try to reverify the original results).

PR #3715, opened 2026-06-18
Fix gaussian_nll_loss documented default reduction (none -> mean)

Pablosinyores:
`nn.losses.gaussian_nll_loss` defaults to `reduction="mean"` in its signature — consistent with the PyTorch API and the other mean-reducing losses (`l1_loss`, `mse_loss`, `binary_cross_entropy`) — but its docstring stated the default was `'none'`.

Calling it without a reduction returns a scalar (the mean), so the documented `'none'` misled callers expecting a per-element array:

```python
>>> nn.losses.gaussian_nll_loss(inputs, targets, vars).ndim
0  # scalar (mean), not the per-element array the docs implied
```

This corrects the docstring to `'mean'` and adds a regression test pinning the default reduction. No behavior change.

### DIFF
--- a/python/mlx/nn/losses.py
+++ b/python/mlx/nn/losses.py
@@ -286,7 +286,7 @@ def gaussian_nll_loss(
         eps (float, optional): Small positive constant for numerical stability.
             Default: ``1e-6``.
         reduction (str, optional): Specifies the reduction to apply to the output:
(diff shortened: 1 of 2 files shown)


zcbenz (approved):
Thanks!

PR #3717, opened 2026-06-18
Fix compiled ops on negative-strided slices

### LINKED ISSUE #3716: mx.compile: assigning an elementwise expression to a negative-strided slice writes only one element
### Describe the bug

### To Reproduce

```python
import mlx.core as mx

def f(x):
    base = mx.zeros_like(x)
    return base

```

```python
def g(x):
    base = mx.zeros_like(x)
    base[::-1] = 2.0 * x[::-1]
    return base

```

|---|---|---|

### Expected behavior

- MLX version: 0.31.2

chrismicah:
## Summary
- materialize negative-strided slice inputs before they feed generated `Compiled` elementwise kernels
- add regression coverage for compiled expressions over `x[::-1]` and slice updates using those expressions

Fixes #3716.

## Tests
- `python setup.py build_ext --inplace -j8`
- `PYTHONPATH=python:python/tests python -m unittest python.tests.test_compile.TestCompile.test_compile_negative_strided_slice_update_expr`
- `python -m black --check python/tests/test_compile.py`
- `git diff --check`

Additional local check:
- `PYTHONPATH=python:python/tests python -m pytest python/tests/test_compile.py -q` -> 58 passed, 1 unrelated local failure in `test_compile_donates_input_buffer` because local NumPy raises `TypeError: asarray() got an unexpected keyword argument copy`.

### DIFF
--- a/mlx/compile.cpp
+++ b/mlx/compile.cpp
@@ -1037,9 +1037,8 @@ std::vector<array> compile_replace(
 
   auto is_load = [](const Primitive& p) { return typeid(p) == typeid(Load); };
   auto has_negative_strides = [](const array& a) {
-    return std::any_of(a.strides().begin(), a.strides().end(), [](auto s) {
-      return s < 0;
-    });
+    return std::any_of(
+        a.strides().begin(), a.strides().end(), [](auto s) { return s < 0; });
   };
   auto is_negative_strided_slice = [](const array& a) {
     if (!a.has_primitive()) {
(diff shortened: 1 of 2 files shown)


zcbenz:
Thanks for the PR but I agree with #3720 that we can actually support the negative stride in compiled kernels.

chrismicah:
Thanks, that makes sense. I agree the better direction is to support negative strides in the compiled kernels rather than adding this narrower workaround, so I will leave this closed and avoid pushing more changes on this approach.

PR #3720, opened 2026-06-19
Fix compiled kernel correctness for negative-strided inputs

### LINKED ISSUE #3716: mx.compile: assigning an elementwise expression to a negative-strided slice writes only one element

lyonsno:
Fixes #3716.

This is an alternative to #3717 that fixes the compiled-kernel path in place instead of materializing negative-strided inputs before the generated kernel runs.

`mx.compile` produces incorrect results for negative-strided inputs such as `x[::-1]`, while eager execution is correct.

This patch fixes two compiled-kernel path issues:

- single-input contiguity detection now requires row- or column-contiguous layout instead of the broader contiguous flag, so negative-strided views route to the strided compiled path;
- Metal/CUDA compiled strided kernels use signed 64-bit indexing when an input has negative strides. The ndim=1 `_large` Metal variant is now always generated (matching ndim > 1 behavior) since negative strides force int64_t indexing even for 1D arrays.

Added regression coverage for 1D reverse, slice update, 2D reverse, mixed positive/negative strides, and a 4D negative-stride case.

Tests:

```sh
python -m pytest python/tests/test_compile.py -q
# 59 passed
```

### DIFF
--- a/mlx/backend/common/compiled.cpp
+++ b/mlx/backend/common/compiled.cpp
@@ -102,7 +102,7 @@ bool compiled_check_contiguity(
   }
   if (non_scalar_inputs > 1 && !all_row_contig && !all_col_contig) {
     contiguous = false;
-  } else if (non_scalar_inputs == 1 && !all_contig) {
+  } else if (non_scalar_inputs == 1 && !(all_row_contig || all_col_contig)) {
     contiguous = false;
   } else if (non_scalar_inputs == 0 && !shape.empty()) {
     contiguous = false;
@@ -224,7 +224,8 @@ std::tuple<bool, Shape, std::vector<Strides>> compiled_collapse_contiguous_dims(
 bool compiled_use_large_index(
     const std::vector<array>& inputs,
     const std::vector<array>& outputs,
-    bool contiguous) {
+    bool contiguous,
+    const std::vector<Strides>& strides) {
   if (contiguous) {
     size_t max_size = 0;
     for (const auto& in : inputs) {
@@ -236,7 +237,18 @@ bool compiled_use_large_index(
     for (const auto& o : outputs) {
       max_size = std::max(max_size, o.size());
     }
-    return max_size > UINT32_MAX;
+    if (max_size > UINT32_MAX) {
+      return true;
+    }
+    // Check for negative strides in inputs (strides[0] is the output).
+    for (size_t i = 1; i < strides.size(); ++i) {
+      for (auto v : strides[i]) {
+        if (v < 0) {
+          return true;
+        }
+      }
+    }
+    return false;
   }
 }
 
--- a/mlx/backend/common/compiled.cpp
+++ b/mlx/backend/common/compiled.cpp
@@ -82,7 +82,16 @@ std::string get_type_string(Dtype d) {
   }
 }
 
-bool compiled_check_contiguity(
+bool has_negative_strides(const array& x) {
+  for (auto s : x.strides()) {
+    if (s < 0) {
+      return true;
+    }
+  }
+  return false;
+}
+
+std::pair<bool, bool> compiled_check_contiguity(
     const std::vector<array>& inputs,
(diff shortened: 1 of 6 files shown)


mlx/backend/cuda/compiled.cpp
```diff
@@ -288,8 +288,23 @@ void Compiled::eval_gpu(
   auto [contiguous, shape, strides_vec] =
       compiled_collapse_contiguous_dims(inputs, outputs[0], is_constant_);
 
-  // Whether to use large index.
+  // Whether to use large index. Also force int64_t indices when any input has
+  // negative strides, since elem_to_loc with unsigned IdxT wraps negative
+  // stride values. Note: this also changes work_per_thread for ndim > 3.
   bool large = compiled_use_large_index(inputs, outputs, contiguous);
+  if (!contiguous && !large) {
```
zcbenz: Should this be part of the `compiled_use_large_index` function?

lyonsno: Good call — moved the negative-stride check into `compiled_use_large_index` so it's centralized instead of duplicated at both call sites.

mlx/backend/common/compiled.cpp
```diff
@@ -236,7 +237,18 @@ bool compiled_use_large_index(
     for (const auto& o : outputs) {
       max_size = std::max(max_size, o.size());
     }
-    return max_size > UINT32_MAX;
+    if (max_size > UINT32_MAX) {
+      return true;
+    }
+    // Check for negative strides in inputs (strides[0] is the output).
```
zcbenz: I think it makes sense skipping strides[0], but can you elaborate what "strides[0] is the output" means?

lyonsno: In `compiled_collapse_contiguous_dims`, the returned strides vector is ordered `[output_strides, input_0_strides, input_1_strides, ...]`. The output is freshly allocated with non-negative strides, so only the input entries (index 1+) can have negative values. I'll add a clearer comment.

zcbenz (approved):
Thanks!

mlx/backend/common/compiled.cpp
```diff
@@ -102,7 +102,7 @@ bool compiled_check_contiguity(
   }
   if (non_scalar_inputs > 1 && !all_row_contig && !all_col_contig) {
     contiguous = false;
-  } else if (non_scalar_inputs == 1 && !all_contig) {
+  } else if (non_scalar_inputs == 1 && !(all_row_contig || all_col_contig)) {
```
angeloskath: Why the change here? Note that `all_contig != (all_row_contig || all_col_contig)`

lyonsno: A negative-strided array (e.g. `x[::-1]`) has `flags().contiguous == true` (densely packed, no gaps) but `flags().row_contiguous == false`. The old check let these through to the contiguous compiled kernel, which iterates linearly from the data pointer — but for a reversed view the data pointer is at the last element, so the linear iteration reads the wrong values. This aligns the single-input path with the multi-input check on line 103 (which already uses `row_contiguous` / `col_contiguous`).

angeloskath: Yeah I know that but for a single contiguous input with positive strides we don't need to use the strides at all which is why that check was like that.

Basically now doing gelu on a transposed array will be slower than before for no reason.

angeloskath (approved):
Thanks for the fix!

Did a little change, feel free to take a look, but basically `gelu(x.transpose(0, 2, 1))` will now route to the appropriate contiguous kernel.

PR #3721, opened 2026-06-19
Fix logsumexp JVP axis reduction

chrismicah:
## Summary
- reduce `LogSumExp::jvp` over the primitive reduction axis so the tangent matches the logsumexp output shape
- add Python regression coverage for `mx.jvp(mx.logsumexp(..., axis=-1))` with and without `keepdims`

Fixes #3707.

## Tests
- `python -m black --check python/tests/test_autograd.py`
- `git diff --check`
- `PYTHONPATH=/private/tmp/mlx-upstream-pr-work/python python -m pytest python/tests/test_autograd.py::TestAutograd::test_reduce_jvp -q` currently exercises the installed pre-fix extension in this checkout and reproduces the old failure (`[squeeze] Cannot squeeze axis 1 with size 3`); the source fix requires rebuilding MLX to exercise the updated C++ primitive.

Note: `clang-format` is not installed in this local environment.

### DIFF
--- a/mlx/primitives.cpp
+++ b/mlx/primitives.cpp
@@ -2949,10 +2949,11 @@ std::vector<array> LogSumExp::jvp(
     const std::vector<int>& argnums) {
   assert(primals.size() == 1);
   assert(tangents.size() == 1);
-  return {multiply(
+  auto weighted_tangent = multiply(
       tangents[0],
       softmax(primals[0], std::vector<int>{-1}, true, stream()),
-      stream())};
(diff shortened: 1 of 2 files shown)


zcbenz:
Closing as duplicate of #3708.

chrismicah:
Thanks, agreed. I missed #3708 here, so I am leaving this closed as duplicate and will check the active PR/issue history more carefully before opening similar fixes.

PR #3722, opened 2026-06-19
Update nanobind to 2.13.0

XXXXRT666:
## Proposed changes

Update nanobind to 2.13.0 and align the example extension dependency pins.

This also enables `nb::pooled()` for the short-lived `array` and `ArrayAt` wrappers, following the upstream nanobind change in wjakob/nanobind#1382. That PR made `nb::pooled()` available for GC-tracked types such as classes using `nb::is_weak_referenceable()` and specifically suggested this optimization for MLX symbolic tracing. A small local wrapper creation benchmark showed `ArrayAt` creation improving from about 35 ns to 29 ns.

Two small compatibility fixes are included:

- Qualify `::mlx::core::SmallVector` inside the nanobind namespace now that nanobind defines `nb::mlx`.
- Include `<structmember.h>` directly for `PyMemberDef`, `T_PYSSIZET`, and `READONLY`.

## Checklist

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -377,7 +377,7 @@ if(MLX_BUILD_PYTHON_BINDINGS)
   FetchContent_Declare(
     nanobind
     GIT_REPOSITORY https://github.com/wjakob/nanobind.git
-    GIT_TAG v2.12.0
+    GIT_TAG v2.13.0
     GIT_SHALLOW TRUE
     EXCLUDE_FROM_ALL)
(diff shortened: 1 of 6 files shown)


jrp2014:
Any reason not to bump further to 2.15, which is designed to have performance improvements?

zcbenz (approved):
Thanks! Did not observe any performance regression with `nb::pooled` when doing some simple inference.

wjakob:
@jrp2014 Latest is 2.13.. :)

PR #3723, opened 2026-06-19
[CUDA] Make qmv support global scale

zcbenz:
`qqmm` reroutes to `qmv` when M=1, while the latter did not support global scales. This PR fixes it for CUDA backend, and it should be trivial to implement in Metal and CPU later if this is the right approach.

Another thing we can do is to add global scale support to `qmm` generally, not necessarily as a public API but as a fallback for `qqmm` when there is no native nvfp4 kernels, for example for Metal backend and for sm80/sm90 in CUDA backend.

### DIFF
--- a/mlx/backend/cuda/device/utils.cuh
+++ b/mlx/backend/cuda/device/utils.cuh
@@ -167,6 +167,9 @@ inline __device__ void store_vector(
 // Type limits utils
 ///////////////////////////////////////////////////////////////////////////////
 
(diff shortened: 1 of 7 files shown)


angeloskath (approved):
I think this is great!

I also think it is a good API change to add `global_scale_x/w` to `qmm` as well even though it might end up adding a lot of kernels (using function constants in Metal should handle this nicely).

PR #3724, opened 2026-06-19
Fix ConvTranspose kernel size repr

ishtihoss:
## Summary
- include all spatial dimensions in ConvTranspose2d and ConvTranspose3d kernel_size repr
- add a regression test with non-square and non-cubic kernels

## Tests
- python -m unittest test_nn.TestLayers.test_conv1d test_nn.TestLayers.test_conv2d test_nn.TestLayers.test_conv_transpose_extra_repr
- python -m unittest test_nn


zcbenz (approved):
Thanks!

PR #3728, opened 2026-06-19
Add math mode option for custom Metal kernels

### LINKED ISSUE #3592: `mx.fast.metal_kernel`: Add support for compiler options (`-fmetal-math-mode`, integer template parameters, Metal 4 Tensor types
`mx.fast.metal_kernel` has no way to pass compiler flags. The kernel source is compiled with MLX's default math mode. Bare `exp()` calls in the kernel source resolve to `metal::fast::exp`, which may not handle `exp(-INFINITY) = 0.0` correctly under relaxed IEEE assumptions.

Any kernel implementing softmax with masked attention (causal, sliding-window) relies on exp(-∞) = 0. In fast math mode this is not guaranteed. The only current workaround is to explicitly call metal::precise::exp(...) in the kernel source — which works but requires users to know this undocumented constraint.

Proposed API addition:

```
    name="my_kernel",
    input_names=["x"],
    output_names=["y"],
    source=src,
    compile_options=["-fmetal-math-mode=relaxed"],   # NEW
)
```
Alternatively, a math_mode enum parameter ("fast" / "relaxed" / "safe").

shubhxho:
# Summary

Closes #3592.

This PR introduces an explicit `math_mode` option to `mx.fast.metal_kernel`:

```python
mx.fast.metal_kernel(
    name="my_kernel",
    input_names=["x"],
    output_names=["y"],
    source=source,
    math_mode="safe",  # "safe", "relaxed", or "fast"
)
```

Custom Metal kernels now default to `math_mode="safe"` to preserve IEEE-compliant behavior. This is particularly important for operations involving special floating-point values, such as ensuring:

```python
exp(-inf) == 0
```

Preserving this behavior is critical for masked softmax implementations used in causal and sliding-window attention, where masked logits are commonly represented as `-inf`.

## Checklist

### DIFF
--- a/mlx/backend/metal/device.h
+++ b/mlx/backend/metal/device.h
@@ -20,6 +20,10 @@ namespace mlx::core::metal {
 using MTLFCList =
     std::vector<std::tuple<const void*, MTL::DataType, NS::UInteger>>;
 
+struct CompileOptions {
+  std::optional<MTL::MathMode> math_mode = std::nullopt;
+};
+
 class Device;
 class EventImpl;
 
@@ -174,7 +178,7 @@ class MLX_API Device {
   MTL::Library* get_library(
       const std::string& name,
       const std::function<std::string(void)>& builder,
-      std::optional<MTL::MathMode> math_mode);
+      CompileOptions compile_options);
 
   void clear_library(const std::string& name);
 
@@ -198,7 +202,7 @@ class MLX_API Device {
  private:
   NS::SharedPtr<MTL::Library> build_library_(
       const std::string& source_string,
-      std::optional<MTL::MathMode> math_mode = std::nullopt);
+      CompileOptions compile_options = {});
 
   NS::SharedPtr<MTL::Function> get_function_(
       const std::string& name,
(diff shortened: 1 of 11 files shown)


mlx/backend/metal/device.h
```diff
@@ -170,6 +171,11 @@ class MLX_API Device {
       const std::string& name,
       const std::function<std::string(void)>& builder);
 
+  MTL::Library* get_library(
+      const std::string& name,
+      const std::function<std::string(void)>& builder,
+      std::optional<MTL::MathMode> math_mode);
```
zcbenz: Can you refactor code so the interface accepts a `CompileOptions` struct that has math mode as its member, which would make future extension easier?

Similarly also make the python API take a `compile_options` dict.

zcbenz (commented):
It seems that the test is not working, if I change `{"math_mode": "safe"}` to `{"math_mode": "fast"}` it still passes, I'm not sure if there is a good way to detect fast math in Metal though.

zcbenz (approved):
Thanks!

PR #3730, opened 2026-06-20
array API: add positive, logical_xor, trunc, count_nonzero, diff, full_like

katlun-lgtm:
## Summary

Focused split from #3684. Adds six new ops required by the Python array API standard that have substantive implementations (not pure aliases).

| Op | Implementation |
|----|---------------|
| `positive` | `mx::astype(a, a.dtype())` — copy with same dtype (shallow copy semantics per @zcbenz's comment on #3684) |
| `logical_xor` | `not_equal(bool(a), bool(b))` |
| `trunc` | `where(a < 0, ceil(a), floor(a))` — truncate toward zero |
| `count_nonzero` | `sum(not_equal(a, 0).astype(int32))` with axis/keepdims support |
| `diff` | n-th discrete difference via `slice` subtraction; supports `prepend`/`append` |
| `full_like` | `mx::full(a.shape(), vals)` with optional dtype override |

## Files changed

- `python/src/ops.cpp` — C++ lambda registrations in `init_ops()`
- `docs/src/python/ops.rst` — alphabetical entries
- `python/tests/test_ops.py` — `test_array_api_elementwise`, `test_diff`, `test_array_api_creation` (full_like part)

## Related

Split from #3684 per reviewer feedback from @zcbenz.

### DIFF
--- a/docs/src/python/ops.rst
+++ b/docs/src/python/ops.rst
@@ -70,6 +70,7 @@ Operations
    dequantize
    diag
    diagonal
+   diff
    divide
    divmod
    einsum
@@ -87,6 +88,7 @@ Operations
    floor_divide
    full
    from_dlpack
+   full_like
    from_fp8
    gather_mm
    gather_qmm
@@ -121,6 +123,7 @@ Operations
    logical_not
    logical_and
    logical_or
+   logical_xor
    logsumexp
    matmul
    max
@@ -141,6 +144,7 @@ Operations
    partition
    pad
    permute_dims
+   positive
    power
    prod
    put_along_axis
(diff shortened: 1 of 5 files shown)


zcbenz (changes requested):
ops that are not pure aliases should have C++ versions and then exposed in python

zcbenz:
Most of this PR looks good to me, but same with https://github.com/ml-explore/mlx/pull/3731#issuecomment-4808834264 would you mind removing the `append`/`prepend` args of `diff`?

katlun-lgtm:
Done — removed the `prepend`/`append` arguments and the second `diff` overload. `diff` is now just `diff(a, n, axis, stream)`. Thanks for the review!

zcbenz (approved):
Thanks!

PR #3731, opened 2026-06-20
array API: add cumulative_sum and cumulative_prod

katlun-lgtm:
## Summary

Focused split from #3684. Adds `cumulative_sum` and `cumulative_prod` as the Array API standard equivalents of `cumsum`/`cumprod`.

### Why separate names? (addressing @zcbenz's concern in #3684)

These are intentionally separate from `cumsum`/`cumprod` because they have different semantics on three axes:

| Feature | `cumsum` / `cumprod` | `cumulative_sum` / `cumulative_prod` |
|---------|---------------------|--------------------------------------|
| `axis=None` | **error** — axis required | flattens the array first |
| Initial element | not supported | `include_initial=True` prepends identity (0 or 1), output length = input length + 1 |
| `dtype` param | not supported | casts input before accumulating |

`include_initial` is specified by the [Array API standard](https://data-apis.org/array-api/latest/API_specification/generated/array_api.cumulative_sum.html) and enables patterns like prefix-sum where you need the neutral element at position 0. It cannot be retrofitted onto `cumsum`/`cumprod` without a breaking change.

## Files changed

- `python/src/ops.cpp` — C++ lambda registrations
- `docs/src/python/ops.rst` — alphabetical entries
- `python/tests/test_ops.py` — `test_cumulative_sum_prod`

## Related

Split from #3684 per reviewer feedback from @zcbenz.

### DIFF
--- a/docs/src/python/ops.rst
+++ b/docs/src/python/ops.rst
@@ -65,6 +65,8 @@ Operations
    cummin
    cumprod
    cumsum
+   cumulative_prod
+   cumulative_sum
    degrees
    depends
    dequantize
--- a/docs/src/python/ops.rst
+++ b/docs/src/python/ops.rst
@@ -65,8 +65,6 @@ Operations
    cummin
    cumprod
    cumsum
-   cumulative_prod
-   cumulative_sum
    degrees
    depends
    dequantize
--- a/mlx/ops.cpp
+++ b/mlx/ops.cpp
@@ -3947,6 +3947,8 @@ array cumsum(
     int axis,
     bool reverse /* = false*/,
     bool inclusive /* = true*/,
+    std::optional<Dtype> dtype /* = std::nullopt*/,
+    bool include_initial /* = false*/,
     StreamOrDevice s /* = {}*/) {
   int ndim = a.ndim();
   if (axis >= ndim || axis < -ndim) {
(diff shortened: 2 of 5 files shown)


zcbenz (changes requested):
It should be done by making `cumsum` support the args of `cumulative_sum` and then making `cumulative_sum` a pure alias.

zcbenz:
Sorry if I was not clear but the correct way to fix this is to add the `include_initial` arg to existing `cumsum` op, and then make the `cumulative_sum` an alias of `cumsum`.

zcbenz (commented):
The implementation of new parameter should live in the C++ version of the ops not the python wrapper.

angeloskath:
I agree with @zcbenz 's comments and I want to add that this implementation is kind of inefficient and in a hidden way. 

Can we collect a bit of information regarding the use of this particular API specifically the `include_initial` part? Is it for a specific project or just to be as compatible as possible with the array namespace?

Just to be clear I foresee code being written like
```python
mx.cumulative_sum(x, axis=-1, include_initial=True)[..., :-1]
```
which is a very inefficient way of writing

```python
mx.cumsum(x, axis=-1, inclusive=False)
```

katlun-lgtm:
Thanks for the feedback @angeloskath.

To answer your question: this is purely for compatibility with the [Python array API spec](https://data-apis.org/array-api/latest/API_specification/generated/array_api.cumulative_sum.html), which requires `include_initial` as a parameter of `cumulative_sum`. The goal of this PR is to make MLX a conformant implementation of that spec.

You're right that the current approach (cumsum + zeros + concatenate) is three ops and can lead to the inefficient usage pattern you showed. To address that, we could implement `include_initial` natively inside the `Scan` primitive so it produces the N+1 output directly in a single pass — eliminating the extra allocation and concat. Would that be the preferred direction, or would you rather we keep the simpler multi-op approach given that `include_initial=True` is not expected to be a hot path?

zcbenz:
We discussed on this and we prefer omitting the parameter if it requires extra efforts to be efficient, because we don't aim for 100% compatibility with array API which is impossible with MLX anyway.

So would you mind removing the `include_initial` arg?

katlun-lgtm:
Done — removed `include_initial`. `cumsum`/`cumprod` now take only the `dtype` arg (no extra concatenate, so the inefficiency @angeloskath flagged is gone), and `cumulative_sum`/`cumulative_prod` stay pure aliases of them. Thanks both!

zcbenz (approved):
Thanks!

PR #3734, opened 2026-06-21
Propagate CPU load task exceptions

aleroot:
## Proposed changes

Propagate exceptions from CPU scheduler tasks, including lazy-load read failures.

CPU lazy loading dispatches file reads through an IO future, but the scheduler task previously used `future::wait()`, which does not rethrow exceptions from the read task. This could silently lose read failures during lazy evaluation. This change uses `future::get()` and adds per-stream scheduler exception propagation so CPU task failures are rethrown on `synchronize(stream)`.

## Checklist

### DIFF
--- a/mlx/backend/common/load.cpp
+++ b/mlx/backend/common/load.cpp
@@ -3,8 +3,8 @@
 #include <algorithm>
 #include <utility>
 
+#include "mlx/backend/cpu/encoder.h"
 #include "mlx/primitives.h"
-#include "mlx/scheduler.h"
 
 namespace {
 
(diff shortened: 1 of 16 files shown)


zcbenz (changes requested):
Propagating CPU exceptions is hard to implement and this change is too wrong and I probably need to spend an hour explaining it and I don't think it would worth the time as AI is quite bad at handling this.

In short the cpu command encoder should propagate the error to the event, like what https://github.com/ml-explore/mlx/pull/3523 does.

aleroot:
> Propagating CPU exceptions is hard to implement and this change is too wrong and I probably need to spend an hour explaining it and I don't think it would worth the time as AI is quite bad at handling this.
> 
> In short the cpu command encoder should propagate the error to the event, like what #3523 does.

@zcbenz thank you for the insulting comment, at the end this is the real thrill of trying to contribute to some open source projects. 

See if the last changes are less wrong. 
In any case feel free to scrap this PR and implement this yourself, the fact that remains is that not propagating those exceptions is simply simplistic and too wrong as well.

zcbenz:
I'm closing this in favor of https://github.com/ml-explore/mlx/pull/3742, please let me know if there are cases that it fails to handle.

PR #3735, opened 2026-06-21
Fix compiled negative-strided slice updates

### LINKED ISSUE #3716: mx.compile: assigning an elementwise expression to a negative-strided slice writes only one element

chrismicah:
## Summary

Fix compiled slice updates when a negative-strided input is used by an
elementwise expression assigned into a negative-strided output slice.

## Details

Issue #3716 reports that `mx.compile` can write only one element for patterns
like:

```python
out[::-1] += 2.0 * x[::-1]
out[::-1] = 2.0 * x[::-1]
```

The compiled Metal kernel used unsigned index math for 1D strided kernels. When
the collapsed stride is negative, the input location wraps instead of walking
backward. This change selects the large-index strided kernel path whenever a
compiled kernel has negative strides, and lets the 1D strided index helper use
the selected signed index type.

## Tests

```bash
CMAKE_BUILD_PARALLEL_LEVEL=8 python setup.py build_ext --inplace
PYTHONPATH=python python -m pytest python/tests/test_compile.py -q -k negative_strided_slice_update_expr --tb=short
PYTHONPATH=python python -m pytest python/tests/test_compile.py -q -k 'negative_strided_slice_update_expr or compile_nonfinite_constants or tuple_output_in_thread' --tb=short
git diff --check
```

I also ran a direct CPU/GPU eager-vs-compiled repro for both `+=` and `=`
negative-strided update forms.

Fixes #3716

### DIFF
--- a/mlx/backend/metal/compiled.cpp
+++ b/mlx/backend/metal/compiled.cpp
@@ -12,6 +12,20 @@
 #include "mlx/utils.h"
 
 namespace mlx::core {
+namespace {
+
+bool has_negative_strides(const std::vector<Strides>& strides) {
+  for (const auto& stride : strides) {
+    for (auto value : stride) {
+      if (value < 0) {
+        return true;
+      }
+    }
+  }
+  return false;
+}
+
+} // namespace
 
 inline void build_kernel(
     std::string& os,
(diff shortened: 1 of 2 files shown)


chrismicah:
Closing this as a duplicate of #3720. I missed that open PR in my initial search; #3720 already fixes #3716 with broader Metal/CUDA coverage and passing CI.

Issue #3568, opened 2026-05-20
mx.random.normal: bit-exact divergence on Apple M1 Max vs M3 Ultra / M5


electron-rare:
## Description

`mx.random.normal(shape=..., key=mx.random.key(seed))` produces **different float32 outputs** on Apple M1 Max vs Apple M3 Ultra and Apple M5, given identical MLX version, identical Python, and identical inputs. `mx.random.uniform` and `mx.random.split` are unaffected — both produce bit-identical output across all three machines.

This breaks bit-exact reproducibility for any code that relies on `mx.random.normal` with a directly-constructed key, across Apple Silicon hardware generations.

## Reproducer

```python
import mlx.core as mx
import numpy as np
import hashlib

key = mx.random.key(0)
a = mx.random.normal(shape=(8,), key=key)
mx.eval(a)
print(hashlib.sha256(np.asarray(a, dtype=np.float32).tobytes()).hexdigest())
```

## Observed output (MLX 0.31.1, Python 3.14.3-4, macOS 26)

| Chip | `sha256` of `normal(shape=(8,), key=mx.random.key(0))` |
|------|--------------------------------------------------------|
| Apple M5         | `a662607828c52cafea1f79d98f09bd81855784116acd23364a6130eb037d6a5e` |
| Apple M3 Ultra   | `a662607828c52cafea1f79d98f09bd81855784116acd23364a6130eb037d6a5e` |
| **Apple M1 Max** | `8a8c5a6737feea9eeee3c7056f4699250e6b8440db8930fd10495ada9117c05a` |

Same pattern for `shape=(8,) key=mx.random.key(42)` and `shape=(16, 4) key=mx.random.key(0)` — M1 Max always differs from {M5, M3 Ultra}.

## Adjacent observations — narrowing the cause

The following calls produce **bit-identical** output across all three machines:

| Probe | M5 | M3 Ultra | M1 Max | All match ? |
|-------|----|----------|--------|-------------|
| `mx.random.key(0)` materialised | ✓ | ✓ | ✓ | **yes** |
| `mx.random.split(mx.random.key(7))` (both halves) | ✓ | ✓ | ✓ | **yes** |
| `mx.random.uniform(low=-1, high=1, shape=(8,), key=mx.random.key(0))` | ✓ | ✓ | ✓ | **yes** |
| `mx.random.normal(shape=(4,), key=<split-child-of-key(7)>)` | ✓ | ✓ | ✓ | **yes** |
[...]

PR #3736, opened 2026-06-21
Document random.normal reproducibility boundary

chrismicah:
## Summary

- Document that `mx.random.normal` seeded outputs are reproducible for a fixed MLX/backend/hardware target
- Clarify that bit-exact outputs are not guaranteed across Apple GPU generations

## Rationale

Issue #3568 reports cross-generation Apple GPU differences for `mx.random.normal`
while `mx.random.uniform` and key generation remain bit-exact. The issue's
follow-up narrows the likely source to hardware-dependent Metal math in the
normal sampling path. Until MLX exposes a portable deterministic normal path,
the API docs should make the reproducibility boundary explicit.

## Tests

- `git diff --check`
- `python -m py_compile python/mlx/core/random.pyi`

### DIFF
--- a/mlx/random.h
+++ b/mlx/random.h
@@ -93,7 +93,12 @@ inline array uniform(
   return uniform(shape, float32, key, s);
 }
 
-/** Generate samples from the standard normal distribution. */
+/** Generate samples from the standard normal distribution.
+ *
+ * Seeded samples are reproducible for a fixed MLX version, backend, and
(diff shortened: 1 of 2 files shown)


zcbenz (commented):
I tested a few frameworks and none of them provide bit-reproducibility across different hardwares and none mentioned the behavior in the docs neither, so I think this is just a common behavior that not needed to be documented.

PR #3739, opened 2026-06-21
Fix CPU dynamic slice copy bound for collapsed shapes

Lyxot:
## Proposed changes

Fixes a CPU copy bug where `copy_general_general` iterated over `src.size()` even when callers provided a smaller `data_shape`.

`DynamicSlice::eval_cpu` passes the output shape into `copy_cpu_inplace`, so CPU dynamic slices with higher-rank collapsed shapes could copy past the destination bounds. In practice this caused crashes and unnecessary extra work. The fix computes the copy size from `data_shape`, matching the requested copy region.

## Reproduce

Before this patch, the CPU path crashes while the GPU path returns the expected slice.

```python
import mlx.core as mx

x = mx.arange(5 * 6 * 7 * 8).reshape(5, 6, 7, 8)
starts = mx.array([1, 2, 3, 4])
axes = (0, 1, 2, 3)
sizes = (2, 2, 2, 2)

with mx.stream(mx.gpu):
    expected = x[1:3, 2:4, 3:5, 4:6]
    gpu_out = mx.slice(x, starts, axes, sizes)
    print("gpu ok:", mx.array_equal(expected, gpu_out).item())

with mx.stream(mx.cpu):
    expected = x[1:3, 2:4, 3:5, 4:6]
    cpu_out = mx.slice(x, starts, axes, sizes)
    print("cpu ok:", mx.array_equal(expected, cpu_out).item())
```

Behavior:
- GPU: prints `gpu ok: True`
- CPU: segfaults on the dynamic slice

## Tests
```bash
python -m pytest python/tests/test_ops.py -k dynamic_slicing -q
DEVICE=cpu ./build/tests/tests --test-case='test dynamic slice'
```

## Checklist

### DIFF
--- a/mlx/backend/cpu/copy.cpp
+++ b/mlx/backend/cpu/copy.cpp
@@ -70,7 +70,6 @@ void copy_general_general(
       dynamic_i_offset ? dynamic_i_offset->data<int64_t>() : nullptr;
   auto o_offset_ptr =
       dynamic_o_offset ? dynamic_o_offset->data<int64_t>() : nullptr;
-  auto size = src.size();
   if (data_shape.empty()) {
     auto val = static_cast<DstT>(*src_ptr);
     *dst_ptr = val;
@@ -107,6 +106,8 @@ void copy_general_general(
     dst_ptr += o_offset_ptr[0];
   }
 
+  auto size = std::accumulate(
+      shape.begin(), shape.end(), int64_t{1}, std::multiplies<int64_t>());
   ContiguousIterator in(shape, strides[0], ndim - 3);
   ContiguousIterator out(shape, strides[1], ndim - 3);
   auto stride = std::accumulate(
--- a/python/tests/test_ops.py
+++ b/python/tests/test_ops.py
@@ -3209,6 +3209,11 @@ class TestOps(mlx_tests.MLXTestCase):
         out = mx.slice(x, mx.array([1, 2, 3]), (0, 1, 2), (3, 2, 1))
         self.assertTrue(mx.array_equal(expected, out))
 
+        x = mx.arange(5 * 6 * 7 * 8).reshape(5, 6, 7, 8)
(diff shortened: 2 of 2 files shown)


mlx/backend/cpu/copy.cpp
```diff
@@ -70,12 +70,16 @@ void copy_general_general(
...
   auto o_offset_ptr =
       dynamic_o_offset ? dynamic_o_offset->data<int64_t>() : nullptr;
-  auto size = src.size();
   if (data_shape.empty()) {
     auto val = static_cast<DstT>(*src_ptr);
     *dst_ptr = val;
     return;
   }
+  auto size = std::accumulate(
+      data_shape.begin(),
```
zcbenz: Maybe using the collapsed `shape` instead?

tests/ops_tests.cpp
```diff
@@ -482,6 +482,29 @@ TEST_CASE("test dynamic slice") {
...
+       812,
+       813,
+       820,
+       821,
+       868,
+       869,
+       876,
+       877},
+      {2, 2, 2, 2});
+  CHECK(array_equal(out, expected).item<bool>());
```
zcbenz: This is testing the exactly the same thing with python test and I think it is not needed?

zcbenz (approved):
Nice fix, thanks!

python/tests/test_ops.py
```diff
@@ -3209,6 +3209,12 @@ def test_dynamic_slicing(self):
         out = mx.slice(x, mx.array([1, 2, 3]), (0, 1, 2), (3, 2, 1))
         self.assertTrue(mx.array_equal(expected, out))
 
+        with mx.stream(mx.cpu):
```
angeloskath: I don't think we need the `with mx.stream(mx.cpu)` here. The test should run both on CPU and on GPU

Issue #3548, opened 2026-05-13
RFC: DLPack consumer support for MLX arrays


megacpp:
We have a working downstream prototype that lets MLX consume DLPack-exporting
objects:

- CPU DLPack capsules / producers.
- MLX self round-trips through mx.array(x.__dlpack__()).
- Metal-resident kDLMetal capsules where data is a foreign
  MTL::Buffer*.
- TileLang / TVM-FFI Metal tensors exported as DLPack and re-wrapped as
  mx.array without a host copy.

This is directly related to ml-explore/mlx issue #2848, where the current
behavior is that mx.array(...) accepts CPU tensors from other frameworks but
not device tensors. The maintainer comment said that accepting MPS / CUDA
arrays may be possible, while PyTorch MPS is blocked until PyTorch exports MPS
DLPack.

Issue link:

- https://github.com/ml-explore/mlx/issues/2848

Proof-of-concept branch:

- https://github.com/DatasunriseOU/mlx

Scope note: this PoC does **not** implement generic CUDA DLPack import yet. It
supports CPU and Metal. kDLCUDA is explicitly rejected today, so CUDA should
be treated as a follow-up if the MLX team wants parity with the new MLX CUDA
backend.

---

## Verified State

I re-checked the repos on 2026-05-13.

Upstream MLX:

text
ml-explore/mlx main
8f4099d8243dfbc814b07ca6b4606196f0c490ce
2026-05-12 [CUDA] Guard qmm_naive scale and bias loads at tile boundaries (#3509)

Downstream PoC:

text
DatasunriseOU/mlx main
3a6039d776fb2d331d8d8575f9f2af533a3b1680
2026-05-13 Add Python array bridge dylib

Merge-base:

text
b08ec310f1506487a32d8fdeb71cc8465c242eae
2026-05-11 Fix scatter_prod GPU hang on NaN with contention (#3492)

Ahead/behind after fetching both remotes:

text
upstream/main...origin/main = 5 behind / 4 ahead

Net downstream delta from the merge-base:

text
25 files changed, 1282 insertions(+), 24 deletions(-)

That is the measured diff for the current downstream branch; the count should
be refreshed before opening PRs because ml-explore/mlx@main is moving.

---

## Why This Matters

MLX already exports DLPack from mx.array:

- mx.array.__dlpack__
- mx.array.__dlpack_device__

But upstream MLX does not currently expose the corresponding consumer side:

- no public mx.from_dlpack.
[...]

PR #3747, opened 2026-06-22
Add from_dlpack to array namespace to fix #3484

### LINKED ISSUE #3484: [BUG] Missing `array-api` methods

Sahith59:
Hey! While checking out the missing array-api methods mentioned in #3484, I noticed from_dlpack wasn't exposed in the core namespace yet.

Since create_array via nanobind already handles DLPack payloads natively under the hood, I just wrapped it into a standard from_dlpack binding in ops.cpp so it strictly complies with the Array API spec. I also threw in a couple of tests in test_array.py to make sure we can round-trip NumPy arrays flawlessly.

Let me know if there's anything else needed here or if I should make any adjustments!

### DIFF
--- a/python/src/ops.cpp
+++ b/python/src/ops.cpp
@@ -1850,6 +1850,20 @@ void init_ops(nb::module_& m) {
             ValueError: If ``copy`` is ``False``.
       )pbdoc");
   m.def(
+      "from_dlpack",
+      [](const nb::object& a) { return create_array(a, std::nullopt); },
+      nb::arg(),
+      nb::sig("def from_dlpack(x: object, /) -> array"),
+      R"pbdoc(
+        Convert a DLPack compatible object to an array.
+
+        Args:
+            x: A DLPack compatible object.
+
+        Returns:
+            array: An array view of the DLPack object.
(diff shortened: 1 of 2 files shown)


zcbenz:
Please do a search of from_dlpack in PRs.

Sahith59:
Ah, I see that this is already being tackled in a larger overhaul over in #3548! Thanks for pointing that out. I'll close out my branch and look for another open bug to help out with. Cheers..!

PR #3749, opened 2026-06-23
Add matrix_norm to array API namespace

wgu9:
## Summary
- add `mx.matrix_norm` for Array API compatibility
- reuse the existing `mx::linalg::norm` implementation over the last two axes instead of adding a second norm formula
- document the API and cover stacked matrices, keepdims, complex inputs, SVD-backed orders, namespace access, and error cases

Refs #3484

## Tests
- `python setup.py build_ext --inplace`
- `PYTHONPATH=python/tests python -m unittest python.tests.test_linalg python.tests.test_array`
- `PYTHONPATH=python/tests python -m unittest python.tests.test_linalg.TestLinalg.test_matrix_norm python.tests.test_array.TestDtypes.test_isdtype`
- `python -m black --check python/tests/test_array.py python/tests/test_linalg.py`
- `/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/clang-format --dry-run --Werror python/src/ops.cpp`
- `git diff --check`

### DIFF
--- a/docs/src/python/ops.rst
+++ b/docs/src/python/ops.rst
@@ -123,6 +123,7 @@ Operations
    logical_or
    logsumexp
    matmul
+   matrix_norm
    max
    maximum
    mean
--- a/docs/src/python/ops.rst
+++ b/docs/src/python/ops.rst
@@ -123,7 +123,6 @@ Operations
    logical_or
    logsumexp
    matmul
-   matrix_norm
    max
    maximum
    mean
--- a/python/src/linalg.cpp
+++ b/python/src/linalg.cpp
@@ -170,7 +170,6 @@ void init_linalg(nb::module_& parent_module) {
           >>> la.norm(m[0, :, :]), LA.norm(m[1, :, :])
           (array(3.74166, dtype=float32), array(11.225, dtype=float32))
       )pbdoc");
-  parent_module.attr("matrix_norm") = m.attr("norm");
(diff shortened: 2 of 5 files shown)


zcbenz (commented):
It seems that we can simply make this an alias of `linalg.norm` like what we do with `bitwise_left_shift`/`matrix_transpose`? They are not the exact same thing but should be enough for API compatibility.

wgu9:
Thanks, updated this to make `mx.matrix_norm` a direct alias of `mx.linalg.norm` instead of adding a separate wrapper. I also narrowed the tests to cover the alias and Array API namespace exposure.

Validation:
- `python setup.py build_ext --inplace`
- `PYTHONPATH=python:python/tests python -m unittest python.tests.test_linalg.TestLinalg.test_matrix_norm python.tests.test_array.TestDtypes.test_isdtype`
- `ASDF_PYTHON_VERSION=3.11.14 black --check python/tests/test_array.py python/tests/test_linalg.py`
- `/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/clang-format --dry-run --Werror python/src/ops.cpp python/src/linalg.cpp`
- `git diff --check`

zcbenz (approved):
Thanks!

PR #3725, opened 2026-06-19
Fix sparse initializer column sparsity

ishtihoss:
## Summary
- make nn.init.sparse zero the documented fraction of entries in each column
- align the zero-mask orientation with PyTorch's sparse initializer
- add a deterministic column-count regression test

## Tests
- python -m unittest test_init

PR #3752, opened 2026-06-23
Document row-wise sparsity in nn.init.sparse

Pablosinyores:
nn.init.sparse zeros ceil(sparsity * cols) entries in each row, but the docstring described the fraction "in each column" and its example zeroed a full row of a 2x2 (which row-wise sparsification cannot produce).

Following @angeloskath's note that the row-wise behavior is intentional - with weights applied as x @ w.T it bounds the input features used by each output feature (Martens, 2010) - this keeps the implementation unchanged and fixes the documentation instead:

- Describe sparsity as applied along each row, with the feature-dimension rationale.
- Correct the sparsity argument wording (row, not column).
- Fix the docstring example so each row carries one zero.
- Replace the per-column test with test_sparse_zeros_per_row, pinning the per-row zero count.

No behavior change.

### DIFF
--- a/python/mlx/nn/init.py
+++ b/python/mlx/nn/init.py
@@ -358,9 +358,15 @@ def sparse(
 ) -> Callable[[mx.array], mx.array]:
     r"""An initializer that returns a sparse matrix.
 
+    Sparsity is applied along each row: every row has the same fraction of its
+    entries set to zero. With a weight matrix applied as ``x @ w.T``, this
+    limits each output feature to at most a ``1 - sparsity`` fraction of the
+    input features, following the sparse initialization of Martens, J. (2010),
+    "Deep learning via Hessian-free optimization".
+
     Args:
-        sparsity (float): The fraction of elements in each column to be set to
-        zero.
+        sparsity (float): The fraction of elements in each row to be set to
+          zero.
         mean (float, optional): Mean of the normal distribution. Default:
           ``0.0``.
         std (float, optional): Standard deviation of the normal distribution.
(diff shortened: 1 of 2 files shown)


zcbenz:
This is a duplicate of #3725 and it is running into the exact same test failure with that PR.

angeloskath:
Well this PR does change it so we sparsify columns instead of rows indeed. The previous test needs fixing since we did change the semantics of the initializer, however, I would argue that the documentation is wrong and we should actually keep sparsifying rows (just mention it in the docs).

Let me make the case for keeping the current way of operating.

One of the first mentions of this type of sparsity in the context of neural network weight initializers is Martens, J. (2010) "Deep learning via Hessian-free optimization". It's definitely not the first and/or seminal work for sparsity (the point of the paper is something else) but in general the idea for sparse initialization is that we limit the number of **input features** that are used for each output feature. As a result the sparsity should be **in the feature dimension**.

In MLX we use the matrices as `x @ w.T`. So the feature dimension is the row dimension.

Using the initializer as is now

```python
initializer = sparse(0.5)
w = initializer(w)
y = x @ w.T
```

makes `y` a linear function of `x` where each output feature uses at most half the features of `x`; which is the intended purpose of the initializer.

So @Pablosinyores feel free to change the PR to a documentation change.

Pablosinyores:
Thanks @angeloskath, that makes sense - the sparsity belongs in the feature dimension, and with x @ w.T that is the row of w. I have reverted the code change and turned this into a documentation fix: the docstring now describes the row-wise behavior (with the Martens reference and the x @ w.T reasoning), the example is corrected to put one zero per row, and the test pins ceil(sparsity * cols) zeros per row instead of per column.

The earlier macOS float16 failure came from the per-column put_along_axis path; since this keeps the original row-wise indexing, that path is no longer touched.

zcbenz (approved):
Thanks!

PR #3754, opened 2026-06-23
metal: cache x in registers in rms_single_row to avoid redundant global read

will-march:
## Problem

`rms_single_row` reads `x[i]` twice from device (global) memory:

1. First pass — accumulates sum-of-squares: `float xi = x[i]; acc += xi * xi;`
2. Second pass — writes scaled output: `out[i] = w[...] * (x[i] * local_inv_mean[0]);`

The backward kernel (`vjp_rms_single_row`) already avoids this by storing x in a thread-local register array and reusing it for both passes. The forward kernel didn't.

## Fix

Introduce `float thread_x[N_READS]` in the forward kernel, fill it in the squaring pass, then read from registers in the output pass. This is exactly what the VJP kernel does.

## Impact

For `axis_size=4096` (e.g. Llama 8B hidden dim), `N_READS=4`, float16, 1024 threads:

- Saves `1024 × 4 × 2 = 8 KB` of global memory traffic per norm call
- For 64 norm calls per decode token (32 layers × 2): 512 KB saved per token

The redundant read was also simplified in the partial-block branch: instead of a guarded read that skips some elements, we now write `0.0f` for out-of-bounds lanes (which contribute 0 to the accumulator), keeping the loop unconditional and the bounds check only in the write-back path.

## Testing

Verified numerically against the reference implementation. Max error 0.001, within float16 round-trip tolerance.

### DIFF
--- a/mlx/backend/metal/kernels/rms_norm.metal
+++ b/mlx/backend/metal/kernels/rms_norm.metal
@@ -27,19 +27,18 @@ template <typename T, int N_READS = RMS_N_READS>
   threadgroup float local_sums[SIMD_SIZE];
 
   float acc = 0;
+  float thread_x[N_READS];
   x += gid * size_t(axis_size) + lid * N_READS;
   w += w_stride * lid * N_READS;
   if (lid * N_READS + N_READS <= axis_size) {
     for (int i = 0; i < N_READS; i++) {
-      float xi = x[i];
-      acc += xi * xi;
+      thread_x[i] = x[i];
+      acc += thread_x[i] * thread_x[i];
     }
   } else {
     for (int i = 0; i < N_READS; i++) {
-      if ((lid * N_READS + i) < axis_size) {
-        float xi = x[i];
-        acc += xi * xi;
-      }
+      thread_x[i] = (lid * N_READS + i < axis_size) ? (float)x[i] : 0;
+      acc += thread_x[i] * thread_x[i];
     }
   }
   acc = simd_sum(acc);
@@ -64,16 +63,16 @@ template <typename T, int N_READS = RMS_N_READS>
   }
   threadgroup_barrier(mem_flags::mem_threadgroup);
 
-  // Write the outputs
+  // Write the outputs using cached x values
   out += gid * size_t(axis_size) + lid * N_READS;
   if (lid * N_READS + N_READS <= axis_size) {
     for (int i = 0; i < N_READS; i++) {
-      out[i] = w[w_stride * i] * static_cast<T>(x[i] * local_inv_mean[0]);
+      out[i] = w[w_stride * i] * static_cast<T>(thread_x[i] * local_inv_mean[0]);
     }
   } else {
     for (int i = 0; i < N_READS; i++) {
       if ((lid * N_READS + i) < axis_size) {
-        out[i] = w[w_stride * i] * static_cast<T>(x[i] * local_inv_mean[0]);
+        out[i] = w[w_stride * i] * static_cast<T>(thread_x[i] * local_inv_mean[0]);
       }
     }
   }
--- a/mlx/backend/metal/kernels/rms_norm.metal
+++ b/mlx/backend/metal/kernels/rms_norm.metal
@@ -67,12 +67,14 @@ template <typename T, int N_READS = RMS_N_READS>
   out += gid * size_t(axis_size) + lid * N_READS;
   if (lid * N_READS + N_READS <= axis_size) {
     for (int i = 0; i < N_READS; i++) {
-      out[i] = w[w_stride * i] * static_cast<T>(thread_x[i] * local_inv_mean[0]);
+      out[i] =
+          w[w_stride * i] * static_cast<T>(thread_x[i] * local_inv_mean[0]);
     }
   } else {
     for (int i = 0; i < N_READS; i++) {
       if ((lid * N_READS + i) < axis_size) {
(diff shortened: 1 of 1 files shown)


will-march:
## AIR bytecode proof — the compiler does not eliminate the redundant load

The concern is that the Metal shader compiler might already cache `x[i]` in registers and eliminate the second read. It does not. The `threadgroup_barrier` between the two passes is an optimization barrier even for `const device` pointers; LLVM conservatively re-reads from global memory after it.

### Method

Both versions were compiled to AIR (Metal's LLVM IR) with `-O2`:

```bash
xcrun --sdk macosx metal -O2 -S -emit-llvm rms_orig.metal  -o rms_orig.ll
xcrun --sdk macosx metal -O2 -S -emit-llvm rms_fixed.metal -o rms_fixed.ll
```

Device memory reads appear as `load half, half addrspace(1)*` in the IR.  
Register reads appear as `load float, float*` (private stack alloca).

---

### Original — output pass (after all three barriers)

```llvm
; --- output loop, unrolled 2× ---
; iteration 0
%83 = load half, half addrspace(1)* %82,  !alias.scope !55  ; w[0]  device read
%86 = load half, half addrspace(1)* %85,  !alias.scope !46  ; x[0]  device read  ← REDUNDANT
store half %90, half addrspace(1)* %91,   !alias.scope !57  ; out[0] write

; iteration 1
%102 = load half, half addrspace(1)* %101, !alias.scope !55 ; w[1]  device read
%105 = load half, half addrspace(1)* %104, !alias.scope !46 ; x[1]  device read  ← REDUNDANT
store half %109, half addrspace(1)* %110,  !alias.scope !57 ; out[1] write
```

`!alias.scope !46` = `x` (input array). It is loaded from `addrspace(1)` (device memory) in **both** the squaring pass and the output pass.

Total `half addrspace(1)` loads: **6** (2 x-reads pass 1 + 2 x-reads pass 2 + 2 w-reads pass 2).

---

### Fixed — output pass (after all three barriers)

```llvm
; --- first pass stored x to thread_x[] (private alloca) ---
%31 = load half, half addrspace(1)* %30,  !alias.scope !46  ; x[0]  device read (pass 1)
store float %32, float* %33                                   ; thread_x[0] = x[0]  (register store)
%46 = load half, half addrspace(1)* %45,  !alias.scope !46  ; x[1]  device read (pass 1)
store float %49, float* %50                                   ; thread_x[1] = x[1]  (register store)

; --- output loop, unrolled 2× ---
; iteration 0
%87  = load half,  half addrspace(1)* %86, !alias.scope !55 ; w[0]  device read
%90  = load float, float* %89                                 ; x[0]  register read  ← no device load
store half %93, half addrspace(1)* %94,   !alias.scope !57  ; out[0] write

; iteration 1
%105 = load half,  half addrspace(1)* %104, !alias.scope !55 ; w[1]  device read
%108 = load float, float* %107                                 ; x[1]  register read  ← no device load
store half %111, half addrspace(1)* %112,  !alias.scope !57  ; out[1] write
```

Total `half addrspace(1)` loads: **4** (2 x-reads pass 1 only + 2 w-reads pass 2).  
The x values are reused from `float*` (private register file) in the output pass.

---

### Summary

| | `half addrspace(1)` loads | x loaded from device in pass 2? |
|---|---|---|
| Original | 6 | yes — twice (once per unroll) |
| Fixed | 4 | no — reads private alloca |

The compiler emits two fewer global device loads per thread. For `N_READS=4` fully unrolled this is 4 fewer loads per thread. Across 1024 threads per norm call that is **8 KB of avoidable device traffic per call**.

will-march:
## Runtime impact clarification

After running a head-to-head benchmark (interleaved, 60 runs, 1k–20k rows), the measured wall-clock speedup is ~0%:

```
rows=  1000  orig=0.349ms  fixed=0.345ms  Δ=+1.3%
rows=  5000  orig=0.850ms  fixed=0.856ms  Δ=-0.7%
rows= 10000  orig=1.463ms  fixed=1.463ms  Δ= 0.0%
rows= 20000  orig=2.780ms  fixed=2.732ms  Δ=+1.7%
```

The reason: 1024 threads in a single threadgroup collectively read the 8 KB x-vector during the squaring pass, which populates the L1 cache. The second read in the output pass hits that cache rather than global memory, so it costs nearly nothing at runtime — even though the compiler emits the load (as shown in the AIR above).

**What this PR actually is:** a code correctness/clarity fix. The backward kernel (`vjp_rms_single_row`) already uses `thread_x[N_READS]` to cache values between passes; the forward kernel didn't, creating an asymmetry with no good reason. This brings them in line with each other.

If the asymmetry is considered intentional or too minor to merge, happy to close.

angeloskath (changes requested):
Even though it will likely not have any change in performance cause x will be in the cache anyway, it is indeed more correct to save it in registers, that's why the single_row vs looped versions exist anyway.

Can you please remove the unrelated changes and then we can merge this?

will-march:
Rebased onto current main — the PR now contains only the `rms_single_row` register cache change (1 file, 8 additions, 9 deletions). The unrelated commits that were previously included have been removed.

will-march:
Done, rebased onto main so it's just the one change now. Should be good to go!

angeloskath (approved):

angeloskath:
@will-march thanks for the change! Lmk if you can fix the lint and then we can run the tests and merge.

will-march:
@angeloskath all sorted, thanks!

PR #3757, opened 2026-06-24
Add gather_qqmm

zcbenz:
Add a simple `gather_qqmm` implementation, by supporting global scale in `qmm` kernels and use them as fallback.

The implementation is bare minimum, for CUDA it is rerouted to `qmm_naive` and for Metal it is `qmv`. We will add global scale support to more `qmm` kernels in followup PRs, and will have specialized fast kernels in future. This PR mostly serves as a way for users to use `qqmm` before we have the native kernels ready.

Support for CPU backend is lacked at the moment..


nastya236:
Looks great, thank you!

nastya236 (approved):

PR #3758, opened 2026-06-24
Fix docstring nits

ishtihoss:
## Summary
- fix the margin_ranking_loss example output for the default reduction='none'
- fix the Bilinear docstring formula and weight shape order

## Tests
- /Users/porkr/projects/contributor/.venv/bin/black --check python/mlx/nn/losses.py python/mlx/nn/layers/linear.py
- PYTHONPATH=python:python/tests /Users/porkr/projects/contributor/.venv/bin/python -m unittest python.tests.test_losses.TestLosses.test_margin_ranking_loss python.tests.test_nn.TestLayers.test_bilinear -v

### DIFF
--- a/python/mlx/nn/layers/linear.py
+++ b/python/mlx/nn/layers/linear.py
@@ -115,10 +115,10 @@ class Bilinear(Module):
 
     .. math::
 
-        y_i = x_1^\top W_i x_2 + b_i
+        y_i = x_2^\top W_i x_1 + b_i
 
     where:
-    :math:`W` has shape ``[output_dims, input1_dims, input2_dims]``, :math:`b` has shape ``[output_dims ]``,
+    :math:`W` has shape ``[output_dims, input2_dims, input1_dims]``, :math:`b` has shape ``[output_dims]``,
     and :math:`i` indexes the output dimension.
 
     The values are initialized from the uniform distribution :math:`\mathcal{U}(-{k}, {k})`,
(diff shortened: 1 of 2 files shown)


python/mlx/nn/layers/linear.py
```diff
@@ -115,10 +115,10 @@ class Bilinear(Module):
 
     .. math::
 
-        y_i = x_1^\top W_i x_2 + b_i
+        y_i = x_2^\top W_i x_1 + b_i
 
     where:
-    :math:`W` has shape ``[output_dims, input1_dims, input2_dims]``, :math:`b` has shape ``[output_dims ]``,
+    :math:`W` has shape ``[output_dims, input2_dims, input1_dims]``, :math:`b` has shape ``[output_dims]``,
```
zcbenz, line 121: The implementation transposes the `self.weight` in `__call__` so I think the math formula is correct here?

ishtihoss: I think the updated wording is describing the stored `W_i` slice. The transpose in `__call__` is on the reshaped `(out * input2_dims, input1_dims)` storage so that `x1 @ w.T` computes the intermediate `W_i x1`.

Tracing one output:

```text
w = weight.reshape(out * in2, in1)
x1 @ w.T -> sum_k x1_k * weight[i, j, k]
x2 @ ... -> sum_j,k x2_j * weight[i, j, k] * x1_k
```

So with the stored weight shape `[output_dims, input2_dims, input1_dims]`, the operation is `y_i = x_2^T W_i x_1 + b_i`. The old formula is equivalent only if `W_i` is defined as the transpose of the stored slice, but then the documented shape was also transposed from the actual storage.

zcbenz (approved):
Thanks!

PR #3763, opened 2026-06-24
Fix fp_qmv_impl small-output-dim branch using raw fp8 scale byte

### LINKED ISSUE #3762: Metal fp_qmv_impl: out_vec_size < 8 branch uses raw scale byte instead of dequantize_scale → wrong mxfp4 matvec for output dim < 8
### Summary

In the Metal `fp_qmv_impl` kernel, the `out_vec_size < 8` (small output-dim) branch's **full-block loop** loads the fp8 scale as a **raw byte** and feeds it straight to `qdot` without decoding it through `dequantize_scale`. Every other fp scale-load site in the same file decodes it. As a result, fp-quantized (`mxfp4` / fp8) **mat-vec** with output dimension `N < 8` multiplies by the raw `0–255` e8m0/e4m3 byte pattern instead of the actual fp8 scale, producing grossly wrong output.

`mlx/backend/metal/kernels/fp_quantized.h` (current `main`, blob `f4bf438df2679161917414ff02c9870589c6eb26`):

```cpp
// fp_qmv_impl, out_vec_size < 8 branch, full-block loop
441    const device auto* sl = scales + row * in_vec_size_g;
442
443    uint8_t s = sl[0];                                                   // <-- raw scale byte
444    result[row] += qdot<U, values_per_thread, bits>(wl, x_thread, s);
```

vs the sibling sites (incl. the remainder loop of this *same* branch at :464):

```
306:      U s = dequantize_scale<U, group_size>(sl[0]);
368:      U s = dequantize_scale<U, group_size>(sl[0]);
464:        U s = dequantize_scale<U, group_size>(sl[0]);   // remainder loop of the same <8 branch
495:        U s = dequantize_scale<U, group_size>(sl[0]);
514:        U s = dequantize_scale<U, group_size>(sl[0]);
```

`qdot` (`fp_quantized.h:82`) takes the scale as a parameter and applies it directly (`scale * accum`) — it does **not** decode the scale internally (the reproduction below confirms this). `dequantize_scale<U, group_size>` reinterprets the byte as `fp8_e4m3` (group_size == 16) or `fp8_e8m0` and converts to float. Skipping it means `qdot` multiplies by the literal `0–255` byte value instead of the decoded fp8 scale.

### Reproduction (mlx 0.31.2, Apple M5 Max, macOS 26.5.1)

```python
import mlx.core as mx

def check(mode, bits, gs, K, Ns, seed=0):
    print(f"# mode={mode} bits={bits} group_size={gs} K={K}")
    for N in Ns:
        x = mx.random.normal((1, K), key=mx.random.key(seed))
        w = mx.random.normal((N, K), key=mx.random.key(seed + 100 + N))
        q = mx.quantize(w, bits=bits, group_size=gs, mode=mode)   # (wq, scales[, biases])
        yq   = mx.quantized_matmul(x, *q, transpose=True, group_size=gs, bits=bits, mode=mode)
        wdeq = mx.dequantize(*q, group_size=gs, bits=bits, mode=mode)
        yref = x @ wdeq.T
        mx.eval(yq, yref)
        rel = (mx.max(mx.abs(yq - yref)) / (mx.max(mx.abs(yref)) + 1e-9)).item()
        branch = "fast(N%8==0)" if (N % 8 == 0 and K % 512 == 0) else ("non-fast <8" if N < 8 else "non-fast >=8")
        print(f"  N={N:>2} {branch:<14} rel_err={rel:.3e}  {'WRONG' if rel > 0.05 else 'ok'}")

check("mxfp4",  4, 32, 512, [5, 7, 12, 8, 16])
check("affine", 4, 32, 512, [5, 7, 12, 8, 16])   # integer path = control, clean everywhere
```

Output:

```
# mode=mxfp4 bits=4 group_size=32 K=512
  N= 5 non-fast <8    rel_err=2.180e+02  WRONG
  N= 7 non-fast <8    rel_err=1.833e+02  WRONG
  N=12 non-fast >=8   rel_err=7.835e-08  ok
  N= 8 fast(N%8==0)   rel_err=2.033e-07  ok
  N=16 fast(N%8==0)   rel_err=6.738e-08  ok

# mode=affine bits=4 group_size=32 K=512
  N= 5 non-fast <8    rel_err=2.518e-07  ok
  N= 7 non-fast <8    rel_err=3.587e-07  ok
  N=12 non-fast >=8   rel_err=2.422e-07  ok
  N= 8 fast(N%8==0)   rel_err=4.346e-07  ok
  N=16 fast(N%8==0)   rel_err=1.992e-07  ok
```

The error fires **only** for `mxfp4` with `N < 8` (218× / 183× relative error; `max|yq|` ≈ 6900 vs reference ≈ 32 — gross, consistent with using the raw exponent byte). The `N=12` non-fast case (which takes the `>=8` branch, correctly decoding the scale) is clean, the fast path is clean, and the integer `affine` path is clean at every `N`. This isolates the defect to the `out_vec_size < 8` full-block loop's scale load.

### Why this is a bug, not an intentional fp-vs-int difference

The integer `quantized.h` `qmv_impl` legitimately does `U s = sl[0];` because there `scales` is `const device T*` (a real float array). In `fp_quantized.h`, `fp_qmv_impl`'s `scales` is `const device uint8_t*` (packed e8m0/e4m3), so it must be decoded — which is why every other fp site here calls `dequantize_scale`. Line 443 looks like a copy-paste from the integer path.

### Reachability

Host dispatch `qmv()` in `mlx/backend/metal/quantized.cpp` selects the non-fast kernel when `N % 8 == 0 && K % 512 == 0` is false. `N % 8 != 0` → `fp_qmv` → `fp_qmv_impl`; `out_vec_size < 8` enters the affected branch; `in_vec_size > block_size` exercises the full-block loop (line 443). So it bites fp-quantized mat-vecs (M=1) whose output dim is `< 8`.

### Proposed fix (1 line)

```cpp
// fp_quantized.h:443
-        uint8_t s = sl[0];
+        U s = dequantize_scale<U, group_size>(sl[0]);
```

This makes the full-block loop identical to the remainder loop of the same branch (:464) and the `>=8` branch loops (:495/:514). Happy to open a PR with a regression test (a `quantized_matmul` vs `dequantize`+matmul check parametrized over `N ∈ {1..7, 8, 12, 16}` for the fp modes) if that's useful — just let me know if you'd prefer to take the kernel change directly.

Introduced by the original fp-qmv kernel (the integer path's `U s = sl[0]` pattern carried over). Verified present in current `main` (blob `f4bf438`) and reproduced on the released `mlx==0.31.2`.

jax-0n-git:
Fixes #3762.

### Problem

In `fp_qmv_impl` (`mlx/backend/metal/kernels/fp_quantized.h`), the `out_vec_size < 8` branch's **full-block loop** loads the fp8 scale as a **raw byte** and passes it straight to `qdot`, skipping `dequantize_scale`:

```cpp
uint8_t s = sl[0];                                                   // raw e8m0/e4m3 byte
result[row] += qdot<U, values_per_thread, bits>(wl, x_thread, s);
```

Every other fp scale-load site in the file decodes the byte first — including the **remainder loop of this same branch** (:464) and the `>= 8` branch loops (:306/:368/:495/:514):

```cpp
U s = dequantize_scale<U, group_size>(sl[0]);
```

`qdot` applies the scale directly (`scale * accum`); it does not decode internally. So fp-quantized (`mxfp4` / `mxfp8` / `nvfp4`) **mat-vec** with output dim `< 8` multiplies by the literal `0–255` exponent/mantissa byte instead of the real fp8 scale, producing grossly wrong output.

The line looks copy-pasted from the integer `quantized.h` `qmv_impl`, where `U s = sl[0];` is correct because there `scales` is a real `const device T*` float array. In `fp_quantized.h`, `scales` is `const device uint8_t*` (packed e8m0/e4m3) and must be decoded.

### When it bites

Both conditions are required:
1. **output dim `N < 8`** → enters the `out_vec_size < (num_simdgroups * results_per_simdgroup) = 8` branch, **and**
2. **`K > block_size`** (`values_per_thread * SIMD_SIZE`; 256 for 4-bit) → the full-block loop at the buggy line actually runs. (For `K ≤ block_size` only the correct remainder loop runs.)

This is why it slipped past CI: `test_fp_qmv` uses output dim ≥ 8 (wrong branch), and `test_qmv_small_non_multiples` uses `K = 32` (below `block_size`, so only the correct remainder loop executes) and doesn't cover `mxfp4`.

Repro on `0.31.2` (M5 Max) — relative error vs `dequantize`+matmul, `K=512`:

| mode | N=5 | N=7 | N=12 | N=8/16 (fast) | affine N<8 (int control) |
|------|-----|-----|------|---------------|--------------------------|
| mxfp4 | 2.2e2 | 1.8e2 | 7.8e-8 | clean | clean |
| mxfp8 | 2.2e4 | 1.8e4 | clean | clean | — |
| nvfp4 | 7.9e1 | 7.8e1 | clean | clean | — |

The integer `affine` path is clean at every `N`, isolating the defect to the fp scale-decode.

### Fix

Decode the scale, matching the four sibling loops:

```cpp
-        uint8_t s = sl[0];
+        U s = dequantize_scale<U, group_size>(sl[0]);
```

### Test

Adds `test_fp_qmv_small_non_multiples` in `python/tests/test_quantized.py`: fp modes {mxfp4, mxfp8, nvfp4} × M∈{1,2} × N∈{1,2,3,5,7} at `K=512` (forcing at least one full block), inputs normalized by `1/sqrt(K)` like the other large-dim tests, asserting `(y_q - y_hat).abs().max() < 1e-3`.

Verified locally on an M5 Max (Metal) by building MLX from source at the current `main`:
- On the current kernel the new test fails every subtest by a wide margin (relative error ~1e1–1e4).
- With the fix it passes, and the small-output-dim path's error matches the always-correct `>= 8` path (~5e-4 relative).
- The full `test_quantized.py` suite (32 tests) passes with the fix — no regressions.
- `pre-commit` (clang-format, black, isort) is clean on both files.

### DIFF
--- a/mlx/backend/metal/kernels/fp_quantized.h
+++ b/mlx/backend/metal/kernels/fp_quantized.h
@@ -440,7 +440,7 @@ METAL_FUNC void fp_qmv_impl(
         auto wl = (const device uint8_t*)(ws + row * in_vec_size_w);
         const device auto* sl = scales + row * in_vec_size_g;
 
-        uint8_t s = sl[0];
+        U s = dequantize_scale<U, group_size>(sl[0]);
         result[row] += qdot<U, values_per_thread, bits>(wl, x_thread, s);
       }
 
--- a/python/tests/test_quantized.py
+++ b/python/tests/test_quantized.py
@@ -482,6 +482,47 @@ class TestQuantized(mlx_tests.MLXTestCase):
             self.assertEqual(y_q.shape, y_hat.shape)
             self.assertLess((y_q - y_hat).abs().max(), 1e-3)
 
+    def test_fp_qmv_small_non_multiples(self):
+        # Regression test for the fp_qmv_impl out_vec_size < 8 branch, whose
+        # full-block loop loaded the fp8 scale as a raw byte instead of decoding
+        # it through dequantize_scale (see fp_quantized.h). The bug only bites
+        # fp-quantized matvec when the output dim is < 8 AND K is large enough to
+        # run at least one full block (K > block_size), so it slipped past both
+        # test_fp_qmv (output dim >= 8) and test_qmv_small_non_multiples (K = 32,
+        # below block_size, exercising only the correct remainder loop).
+        # K = 512 forces full blocks; N in {1..7} hits the < 8 branch.
+        key = mx.random.key(0)
+        k1, k2 = mx.random.split(key)
+        K = 512
+        for mode, group_size, bits in [
+            ("mxfp4", 32, 4),
+            ("mxfp8", 32, 8),
+            ("nvfp4", 16, 4),
+        ]:
+            for M in [1, 2]:
+                for N in [1, 2, 3, 5, 7]:
+                    with self.subTest(M=M, K=K, N=N, mode=mode):
+                        x = mx.random.normal(shape=(M, K), key=k1) / K**0.5
+                        w = mx.random.normal(shape=(N, K), key=k2) / K**0.5
+                        w_q, scales = mx.quantize(
+                            w, group_size=group_size, bits=bits, mode=mode
+                        )
+                        w_hat = mx.dequantize(
+                            w_q, scales, group_size=group_size, bits=bits, mode=mode
+                        )
+                        y_q = mx.quantized_matmul(
+                            x,
+                            w_q,
+                            scales,
+                            transpose=True,
+                            group_size=group_size,
+                            bits=bits,
+                            mode=mode,
+                        )
+                        y_hat = x @ mx.swapaxes(w_hat, -1, -2)
+                        self.assertEqual(y_q.shape, y_hat.shape)
+                        self.assertLess((y_q - y_hat).abs().max(), 1e-3)
+
     def test_qvm(self):
         key = mx.random.key(0)
         k1, k2 = mx.random.split(key)


zcbenz (commented):
Nice catch! And I think you get the right fix.

But the test is not really doing its job, if I run in it in the main branch it would still pass.

zcbenz:
I'm closing this PR in favor of https://github.com/ml-explore/mlx/pull/3804.

PR #3767, opened 2026-06-25
Find SwiftPM metallib bundle near test binary

RNT56:
## Summary

- search the test binary directory and its parents for the SwiftPM resource bundle
- handle the case where `Bundle.allBundles` already contains the SwiftPM resource bundle itself
- keep the existing main-bundle, bundle-resource, framework, colocated, and `METAL_PATH` fallback behavior

## Validation

- `git diff --check upstream/main...upstream-pr/swiftpm-metallib-bundle-lookup`
- downstream `mlx-swift` SwiftPM resource branch builds `MLX` and produces platform-specific `default.metallib` resources; runtime loading of those resources depends on this lookup change

### DIFF
--- a/mlx/backend/metal/device.cpp
+++ b/mlx/backend/metal/device.cpp
@@ -80,12 +80,10 @@ std::pair<MTL::Library*, NS::Error*> load_library_from_path(
 }
 
 #ifdef SWIFTPM_BUNDLE
-MTL::Library* try_load_bundle(
+MTL::Library* try_load_bundle_path(
     MTL::Device* device,
-    NS::URL* url,
+    const std::string& bundle_path,
     const std::string& lib_name) {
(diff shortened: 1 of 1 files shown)


RNT56:
This is the small MLX-side lookup change needed for the SwiftPM `default.metallib` resource work in `mlx-swift`.

It keeps the existing lookup behavior and adds the SwiftPM resource bundle case. That should let the `mlx-swift` packaging PR use an upstream MLX commit once this lands.

Local validation:

- `git diff --check upstream/main...upstream-pr/swiftpm-metallib-bundle-lookup`
- downstream `mlx-swift` packaging branch builds `MLX` and produces platform-specific `default.metallib` resources

RNT56:
I realized this is effectively the same automatic lookup approach as #3562, which was already superseded by the explicit metallib path API in #3597. I’ll close this and move the SwiftPM packaging work to that API route instead.

PR #3768, opened 2026-06-25
[WIP] [CUDA] fsdp

nastya236:
Fully sharded data parallel.

```python
dtype = mx.bfloat16
for i, layer in enumerate(model.layers):
    model.layers[i] = fully_shard(layer, group=groups["fsdp"], compute_dtype=dtype)
model.embed_tokens = fully_shard(model.embed_tokens, group=groups["fsdp"], compute_dtype =dtype)
```

It is incomplete, because bf16 weights that are gathered during forward are not resharded and used in backward. So we can cut another ~6GB by resharding and gathering back during backward. 
image

Peak memory difference between sharded optimizer and this PR is 14 GB [4B * 4 = 16GB -- master copy weights, therefore each rank holds only 16GB / 8 =2 GB 2 instead of 16GB] as expected.


angeloskath (approved):
This looks really good!

I am curious about how you are thinking to approach the re-gathering the weights in the backward. I see a few ways none of which I like.

Either way we should definitely merge this as is. Really nice!

nastya236:
Thank you! General high level thought was to checkpoint gather so the forward weights don't stay in the graph, but you probably though about it more than me..

PR #3769, opened 2026-06-26
Fix Upsample align_corners singleton output

ishtihoss:
## Summary

Fix `nn.Upsample` with `align_corners=True` when an output spatial dimension is 1.

The aligned-corners scale used `(N - 1) / (M - 1)`, so `M == 1` raised a `ZeroDivisionError`. Using `max(M - 1, 1)` keeps existing behavior for `M > 1` and maps singleton outputs to the first input coordinate.

## Tests

- `black --check python/mlx/nn/layers/upsample.py python/tests/test_nn.py`
- `python -m unittest test_nn.TestLayers.test_upsample_align_corners_one_dim -v`
- `python -m unittest test_nn.TestLayers.test_upsample -v`


zcbenz (approved):
Thanks!

PR #3770, opened 2026-06-26
Fix Bilinear weight shape in docstring

Pablosinyores:
### Summary

`nn.Bilinear` stores its weight as `[output_dims, input2_dims, input1_dims]` — both `__call__` and `extra_repr` unpack it as `out, in2, in1 = self.weight.shape`, and the forward computes `W[o, in2, in1]`. The docstring, however, documented `[output_dims, input1_dims, input2_dims]`, swapping the two input dimensions.

This corrects the docstring to the actual stored order and pins the weight shape in `test_bilinear` so the documented order stays in sync with the implementation.

Behavior is unchanged (docs + test only).

### Checklist

- [x] I have added tests that prove my fix is effective (pins the weight shape)
- [x] I have run `pre-commit`/`black` on my changes

### DIFF
--- a/python/mlx/nn/layers/linear.py
+++ b/python/mlx/nn/layers/linear.py
@@ -118,7 +118,7 @@ class Bilinear(Module):
         y_i = x_1^\top W_i x_2 + b_i
 
     where:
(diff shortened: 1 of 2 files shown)


zcbenz:
Thanks for the PR, this had been fixed by https://github.com/ml-explore/mlx/pull/3758, please let us know if it is missing anything.

Issue #1441, opened 2024-09-27
[FEATURE] MLX equivalent for jax.lax.scan and jax.lax.while_loop 


sachinraja13:
Would it be possible to have MLX equivalents for jax.lax.scan and jax.lax.while_loop?

Matching algorithms like Hungarian matching for object detection (DETR) need boolean arrays to be evaluated to correctly compute optimal matchings. As far as my understanding goes, this would make it improbable to write an MLX specific implementation which can be compiled to optimise the end-to-end training process.

What I'm not sure about is how much performance benefit can be obtained if the loss computation (that depends on matching) itself can be compiled. From what I see on the documentation, compiling can yield significant performance improvements.

PR #3772, opened 2026-06-26
Add mx.while_loop (scan like)

cetagostini:
## Summary

Adds a pure-Python `mx.while_loop` that solves the **per-iteration GPU-sync pathology** of while loops on accelerators. A `@mx.compile`d "chunk" applies up to `chunk_size` body iterations lazily (with a running mask so post-termination iterations are no-ops) before a *single* host sync to read the condition, reducing host GPU syncs from **O(N) to O(N / chunk_size)**.

```python
mx.while_loop(cond_fun, body_fun, init_val, *, max_iterations=None, chunk_size=16) -> Any
# semantics: while bool(mx.all(cond_fun(c))): c = body_fun(c)
```

## Issues solved

- **tillahoffmann/jax-mps#83** — *While loop performance: ~0.7ms per iteration due to per-iteration GPU sync.* While loops (`stablehlo.while` / `lax.fori_loop`) are ~1750x slower than CPU because each iteration syncs to read the condition on the host. This PR provides the "MLX while primitive" (approach #2 in that issue) and the chunked-speculative/batch-eval approaches (#1/#3) at the MLX level, dropping per-iteration syncs to per-chunk syncs.
- **#1441** — *[FEATURE] MLX equivalent for jax.lax.scan and jax.lax.while_loop.* Provides the `while_loop` half of that request as a pure-Python transform (no risky C++ control-flow primitive; the `scan` half remains future work).

## Changes

| File | Change |
|---|---|
| `python/mlx/_while_loop.py` (new) | The implementation: chunked-speculative driver, pre-flight validation (args / pytree structure / shape / dtype / all-`mx.array` leaves), multi-layer transform guard, bounded `chunk_size`, `max_iterations` cap. |
| `python/mlx/__init__.py` (new) | Registers `mx.while_loop` after `mlx.core` is fully initialized. Required: registering from `_reprlib_fix` (runs mid-`NB_MODULE`) fatally re-enters `NB_MODULE`. Uses `# isort: off` to preserve the load order. |
| `python/src/array.cpp` (+5 lines) | One-line binding `mx.array.is_tracer()` → `mx::array::is_tracer()` (`is_tracer_flag && in_tracing() || retain_graph()`). Used for robust transform-context detection. |
| `python/mlx/_stub_patterns.txt` (+21) | Type stub for `while_loop` in the `mlx.core.__suffix__` block (`is_tracer` is auto-generated by stubgen). |
| `python/tests/test_while_loop.py` (new) | 53 tests. |
| `benchmarks/python/while_loop_bench.py` (new) | Naive per-iter-sync vs chunked, + `chunk_size` sweep. |

Only **one 5-line C++ addition** (the `is_tracer` binding); the rest is pure Python.

## Key design points / properties

- **Not differentiable and not traceable** into `mx.grad` / `mx.compile` / `mx.vmap` — raises a clear `RuntimeError` and **never silently produces wrong gradients**. The pre-flight calls `body_fun(init_val)` / `cond_fun(init_val)` *uncompiled* and checks `is_tracer()` on the outputs, which catches the constant-init + closure-over-tracer `grad` case that `@mx.compile` would otherwise mask (the v4 key fix, empirically verified).
- **`cond_final` is computed in the driver, outside `@mx.compile`** — avoids a backend `IndexError` dedup bug for constant conditions and fixes an off-by-one cap bug (trip-count == `max_iterations` returns instead of raising).
- **A fresh `@mx.compile`d chunk is built per call** (no cross-call cache) — `@mx.compile` bakes closure-captured Python values as constants, so a cross-call cache would reuse a stale chunk when a closed-over variable changes (hang). `mx.compile`'s own cache still amortizes compilation across the many chunks within a single call.
- `body_fun` is evaluated eagerly/speculatively (keep it **pure and total**; NaN on an *active* carry contaminates, NaN on a *stopped* carry is masked). All carry leaves must be `mx.array`. `max_iterations` caps the loop (raises if exceeded); `chunk_size` trades syncs vs memory (default 16; 128–256 typically optimal for long loops).
- **Documented limitation:** `body_fun`/`cond_fun` must not mutate closure-captured Python state that the other reads (the chunk bakes closure constants; `max_iterations` will not bound such a hang — pass loop-dependent values through the carry).

## Test cases (53 in `test_while_loop.py`)

- **Correctness vs Python reference** (scalar, tuple, list, dict, nested pytree, vectorized/multi-element carry) across `chunk_size` sweeps; random/property tests (seeded) for a linear recurrence and a random pytree carry.
- **Termination semantics**: cond false at init → returns init with **zero body calls** (matches JAX); trip-count == `max_iterations` returns (no spurious raise); `max_iterations` cap raises exactly when exceeded; `max_iterations=0` semantics; partial chunks (`max` not a multiple of `chunk_size`).
- **NaN safety**: stopped-lane NaN masked (result finite); active-lane NaN contaminates (documented).
- **Validation**: `chunk_size` ∈ [1, 4096] (rejects 0/-1/non-int/bool/>4096; accepts `np.int64`); `max_iterations` ∈ [0, 2^63-1] or `None` (rejects bool/float/negative/overflow); `init_val` ≥1 `mx.array` leaf (rejects empty/`str`/`None`/numpy/Python-scalar leaves); body structure / shape / dtype invariant (tuple-vs-list, extra/missing key, shape change, dtype change → `ValueError`; same-keys-different-order accepted).
- **Not differentiable / not traceable**: `mx.grad` (direct tracer init + closure-over-tracer), `mx.vmap`, `mx.compile`, and recursive `while_loop` in a body all raise `RuntimeError` (skip-guarded on `is_tracer`).
- **Cond coercion**: Python `bool`, `np.bool_`, 0-d / 1-element / multi-element arrays, oscillating `T→F→T` (stops at first false), constant `True`/`False`.
- **Performance (sync-count, not wall-time)**: asserts `mx.eval` count ≤ `ceil(N / chunk_size) + 2`; skipped on CPU / `MLX_SKIP_PERF_TESTS`.
- **Concurrency**: 4 threads × 5 calls (each sets `mx.set_default_device`).
- **Misc**: determinism (5×), `@mx.compile`d body inside `while_loop`, kw-only args, `numpy` integer args, constant-cond-no-crash.

## Benchmark

Naive per-iter-sync Python loop vs chunked `mx.while_loop` (prints; no assertions):
- N=1000: naive ~361 ms vs chunked (cs=64) ~19 ms → **~18× speedup**.
- chunked N=10000 (cs=64): ~168 ms (~157 syncs vs 10000 for naive).
- `chunk_size` sweep shows a U-shape (optimal ~128–256; worsens at 4096 due to compile/memory).

## Tested version / environment

- **MLX 0.32.0.dev** (`0.32.0.dev20260626+0386ad49e`, built from `main` @ `0386ad49e`, 2026-06-26).
- **macOS 26.5.1**, **Apple Silicon (arm64)**, **Metal toolchain 17F109**.
- Conda env `mlx-dev` rebuilt from source (with the one-line `is_tracer` C++ binding).
- **53 `test_while_loop` tests pass**; **125 pass** with `test_compile` + `test_eval` (no regression). The 4 `test_array` torch/dlpack failures are **pre-existing** (unrelated; confirmed on baseline without these changes).

## Notes

- The implementation went through a Construction Beehive workflow: exploration → planning (4-iteration plan-review loop, converged) → implementation → test gate → implementation review (3 cycles, all 4 reviewers converged) → fact-check (12/12 claims verified) → supervisor (CONDITIONAL PASS). See the audit trail for details.
- Non-differentiability and the closure-mutation limitation are inherent to a host-driven, `@mx.compile`-chunked `while_loop` (matching JAX's `lax.while_loop` non-differentiability). A native C++ `while` primitive with on-device condition evaluation (no host sync) is future work for fully traceable loops.

### DIFF
--- a/benchmarks/python/while_loop_bench.py
+++ b/benchmarks/python/while_loop_bench.py
@@ -1,0 +1,109 @@
+# Copyright © 2026 Apple Inc.
+"""Benchmark: chunked mx.while_loop vs naive per-iter-sync Python loop.
+
+Reproduces the per-iteration GPU-sync pathology of
+https://github.com/tillahoffmann/jax-mps/issues/83 and shows the chunked
+amortization. Prints results; does NOT assert a speedup.
+"""
+
+import math
+import time
+
+import mlx.core as mx
+
+
+def naive_while(cond, body, init, max_iterations):
+    """Python while loop that syncs every iteration (the pathology)."""
+    carry = init
+    n = 0
+    while bool(mx.all(cond(carry))):
+        carry = body(carry)
+        mx.eval(carry)  # per-iteration host sync
+        n += 1
+        if n >= max_iterations:
+            break
+    return carry
+
+
+def time_it(fn, *args, warmup=2, iters=5, **kwargs):
+    """Return median ms per call (syncs included)."""
+    for _ in range(warmup):
+        mx.eval(fn(*args, **kwargs))
+    times = []
+    for _ in range(iters):
+        t0 = time.perf_counter()
+        mx.eval(fn(*args, **kwargs))
+        t1 = time.perf_counter()
+        times.append((t1 - t0) * 1000)
+    times.sort()
+    return times[len(times) // 2]
+
+
+def main():
+    print(f"MLX version: {mx.__version__}  Metal: {mx.metal.is_available()}")
+
+    def make_cond(n):
+        # Capture n by value (default arg) to avoid late-binding surprises if
+        # n is reassigned later in this function.
+        def cond(c, _n=n):
+            return c < _n
+
+        return cond
+
+    def body(c):
+        return c + 1
+
+    # Warm up compile caches.
+    mx.while_loop(
+        make_cond(1000), body, mx.array(0), max_iterations=2000, chunk_size=64
+    )
+    naive_while(make_cond(100), body, mx.array(0), 100)
+
+    # Naive is O(N) syncs; only measure at N=1000 (the pathology reproducer).
+    print("\n=== Naive (per-iter sync) vs chunked (chunk_size=64), N=1000 ===")
+    N = 1000
+    cond = make_cond(N)
+    init = mx.array(0)
+    t_naive = time_it(naive_while, cond, body, init, N)
+    t_chunk = time_it(
+        lambda init, cap, cs: mx.while_loop(
+            cond, body, init, max_iterations=cap, chunk_size=cs
+        ),
+        init,
+        N * 2,
+        64,
+    )
+    print(
+        f"  N={N}: naive={t_naive:8.1f}ms  chunked={t_chunk:7.1f}ms  "
+        f"speedup={t_naive / t_chunk:5.1f}x  syncs_chunked~={math.ceil(N / 64)}"
+    )
+
+    print("\n=== chunked only at N=10000 (naive too slow to time) ===")
+    N = 10000
+    cond = make_cond(N)
+    init = mx.array(0)
+    t_chunk10 = time_it(
+        lambda init, cap, cs: mx.while_loop(
+            cond, body, init, max_iterations=cap, chunk_size=cs
+        ),
+        init,
+        N * 2,
+        64,
+    )
+    print(f"  N={N} chunked(cs=64): {t_chunk10:7.1f}ms  syncs~={math.ceil(N / 64)}")
+
+    print("\n=== chunk_size sweep at N=10000 (skipping cs<4, too slow) ===")
+    for cs in [4, 8, 16, 32, 64, 128, 256, 512, 1024, 4096]:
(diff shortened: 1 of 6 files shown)


angeloskath:
Hi @cetagostini thanks for the PR but I don't think we should merge this.

I think it is a complicated hidden way to write the same thing that can be written with a loop that doesn't do unnecessary evaluations. In a purely compiled framework it makes sense because potentially the whole while loop can be compiled to a single kernel for CPU or GPU, this is not something that we do here and can be confusing for someone coming from Jax for instance.

Finally, if this was ever to be added it should be added in the C++ api and not just in python.

I am closing this but feel free to discuss it if you think that this is a mistake and we can reopen or revisit this feature in the future.

PR #3783, opened 2026-06-30
fix: Quote hostname in mlx.launch ssh commands

### LINKED ISSUE #3776: mlx.launch should properly quote the hostname
Not really a problem since apparently it is user's responsibility to pass a trusted hostname here but some AI scanners just keep reporting it as security issue.

https://github.com/ml-explore/mlx/blob/c9ccaba99743bc5ea33ef9f74e7f1969dffc9d92/python/mlx/_distributed_utils/launch.py#L47-L66

varshneydevansh:
## Proposed changes

- Quote the hostname when constructing `ssh` shell commands in `mlx.launch`.
- Applies the quoting to the launch command and the remote termination command.
- Fixes #3776.

Tests: not added; this is a minimal shell-quoting fix.

## Checklist


zcbenz (approved):
Thanks!

PR #3785, opened 2026-07-01
Decompose small kernel-depth 3D convs into 2D convs

### LINKED ISSUE #3625: [Performance] conv_general 5D (3D convolution) is 2-5x slower than equivalent per-frame 2D conv
## Summary

`mx.conv_general` with 5D inputs (3D convolution) is significantly slower than decomposing the same operation into per-frame 2D convolutions with a Python loop. This affects video generation workloads (VAE decoders) where 3D convolutions over temporal/spatial dimensions are common.

## Environment

- Hardware: Apple M4 Ultra (T6050), 128GB
- macOS: 26.5 (Tahoe)
- MLX: latest (via pip, mlx-video dependency)
- Python: 3.12

## Reproduction

```python
import os
os.environ["MACGEN_ALLOW_PARENT_MLX"] = "1"
import time
import mlx.core as mx

# Test: 3x3x3 conv on [1, 41, 120, 208, 512] input
# Equivalent to a CausalConv3d in video VAE decoder

B, T, H, W, Cin, Cout = 1, 41, 120, 208, 512, 512
kd, kh, kw = 3, 3, 3

# Random input and weight
x = mx.random.normal((B, T + 2, H + 2, W + 2, Cin), dtype=mx.bfloat16)  # pre-padded
w = mx.random.normal((Cout, kd, kh, kw, Cin), dtype=mx.bfloat16)
bias = mx.zeros((Cout,), dtype=mx.bfloat16)
mx.eval(x, w, bias)

# Method 1: Native 3D conv (5D conv_general)
mx.synchronize()
t0 = time.perf_counter()
y_3d = mx.conv_general(x, w, stride=(1, 1, 1)) + bias
mx.eval(y_3d)
mx.synchronize()
native_time = time.perf_counter() - t0

# Method 2: Per-frame 2D conv (Python loop, 3 depth positions)
mx.synchronize()
t0 = time.perf_counter()
outputs_d = []
for d in range(kd):
    frames = x[:, d:d+T].reshape(B * T, H + 2, W + 2, Cin)
    w2d = w[:, d, :, :, :]  # [Cout, kh, kw, Cin]
    conv_out = mx.conv_general(frames, w2d)
    outputs_d.append(conv_out.reshape(B, T, conv_out.shape[1], conv_out.shape[2], Cout))
y_2d = outputs_d[0] + outputs_d[1] + outputs_d[2] + bias
mx.eval(y_2d)
mx.synchronize()
loop_time = time.perf_counter() - t0

print(f"Native 3D conv: {native_time*1000:.0f}ms")
print(f"Per-frame 2D loop: {loop_time*1000:.0f}ms")
print(f"Slowdown: {native_time/loop_time:.1f}x")

# Verify correctness
diff = mx.max(mx.abs(y_3d[:, :T] - y_2d)).item()
print(f"Max diff: {diff:.6f}")
```

## Results

Across multiple spatial resolutions (relevant to video VAE decoder stages):

```
Input shape                      Native 3D    Per-frame 2D    Ratio
──────────────────────────────────────────────────────────────────────
[1, 21, 60, 104, 1024]           488ms        231ms           2.1x slower
[1, 41, 120, 208, 512]           966ms        240ms           4.0x slower
[1, 41, 240, 416, 512]          3888ms        815ms           4.8x slower
```

The gap grows with spatial resolution, suggesting the 3D conv Metal kernel has suboptimal memory access patterns or lacks tiling optimizations that the 2D conv kernel has.

## Additional Findings

A "depth-chunked" approach (batching temporal frames into groups of 2-8, then using 2D conv per depth position) achieves 1.03-1.72x speedup over the naive per-frame loop:

```
Input shape                      Chunk size   Optimized    Speedup vs loop
────────────────────────────────────────────────────────────────────────────
[1, 11, 30, 52, 1024]              cs=8        63ms         1.40x
[1, 21, 60, 104, 1024]             cs=4       199ms         1.72x
[1, 41, 120, 208, 1024→512]        cs=3       731ms         1.13x
[1, 41, 240, 416, 512→256]         cs=2      1132ms         1.03x
```

This suggests the 2D conv Metal kernel benefits significantly from batched inputs, and a well-optimized 3D conv kernel could match or exceed this.

## Impact

This affects all video generation models using 3D convolutions in MLX:
- Wan2.2 VAE decoder: 3D conv takes ~70% of VAE decode time
- LTX-2.3 VAE: similar 3D conv usage
- Any video model with CausalConv3d or 3D convolution layers

An optimized 3D conv kernel would reduce VAE decode time by 15-40%, translating to meaningful end-to-end pipeline improvements for video generation on Apple Silicon.

## Related

- #1409 (general conv_general performance)
- #3611 (conv shape overflow for large inputs)

katlun-lgtm:
Closes #3625.

## Problem

`mx.conv_general` with 5D inputs (3D convolution) is 2–5× slower than decomposing the
same op into per-frame 2D convolutions with a Python loop. This hits video-generation
workloads (VAE decoders with `CausalConv3d`).

The reason isn't the explicit-gemm fallback — for the common case (mod16 channels,
`idil==1`, `groups==1`) `dispatch_conv_3D_gpu` already routes to
`implicit_gemm_conv_3D_gpu`. The real cause is that the **2D dispatch has a Winograd
kernel for 3×3 stride-1 convs** (`dispatch_conv_2D_gpu` → `winograd_conv_2D_gpu`), and the
**3D path has no Winograd or 3×3-specialized kernel** — so a 3×3×3 conv misses Winograd
entirely. Decomposing a small kernel-depth 3D conv into `KD` 2D convs lets each 2D conv
hit the tuned 2D path.

## Change

`small_kd_conv_3D_gpu`: for small kernel depth, for each depth tap `kd` we build a
zero-copy strided view of the input frames (`[OD, H, W, C]` at depth offset `kd`) and the
weight depth-slice (`[O, KH, KW, C]`), run `conv_2D_gpu`, and accumulate. The accumulator
buffer is repointed into `out` via `copy_shared_buffer` (the `pad_and_slice` pattern).

A guard in `dispatch_conv_3D_gpu` takes this path **only when it is valid and faster**:

- `idil == 1` (all dims), `groups == 1`, `N == 1`;
- `KD <= 7` (the `KD-1` accumulate adds erode the win for large `KD`);
- depth `stride == 1` and `kernel_dilation == 1`, and `pad[0] == 0`;
- `mod16` channels (so the per-frame 2D convs hit the fast path).

Everything else falls through to the existing implicit gemm unchanged.

(Skeleton follows @Ved235's sketch in the issue; the depth-tap decomposition idea is
theirs.)

## Results (M3 Max, mlx built from this branch)

Correctness — the fast path vs the CPU reference (fp32, exact):

| KD | max\|gpu − cpu\| | max\|cpu\| |
|----|-----------------|-----------|
| 1 | 3e-5 | 61.0 |
| 2 | 1e-4 | 91.6 |
| 3 | 1e-4 | 106.8 |
| 5 | 2e-4 | 142.2 |

Speed (bf16, 3×3×3), native 3D vs the per-frame 2D decomposition:

| case (T×H×W, C→O) | before | after | per-frame 2D |
|---|---|---|---|
| 41×60×104, 256→256 | 71 ms | **28 ms** | 27 ms |
| 41×120×208, 512→512 | 1128 ms | **335 ms** | 336 ms |
| 17×64×64, 256→256 | 20 ms | **9 ms** | 8 ms |

i.e. 2.4–3.4× → **~1.0× (parity)** with the per-frame 2D path, and numerically identical
to it.

## Tests

`python/tests/test_conv.py`:
- `test_conv_3D_small_kd_decomposition` — the fast path vs CPU across 5 shapes
  (`Cout != Cin`, `KD ∈ {1,2,3,5}`, 1×1 spatial), fp32 exact.
- `test_conv_3D_small_kd_fallback_cases` — depth stride > 1, depth padding, and non-mod16
  channels stay correct (must take the fall-back path).

Full `test_conv.py` suite passes with no regressions.

## Open questions for maintainers

1. **Accumulation** — currently the first tap owns the buffer and later taps do an in-place
   `binary_op_gpu_inplace(..., "Add")`. Prefer that, or a `beta`-accumulate added to
   `conv_2D_gpu`?
2. **`KD` threshold** — fixed at 7, or a cost heuristic vs the 3D implicit gemm?
3. **Scope** — is the decomposition the accepted fix, or do you want a native 3D Winograd
   later? Also: should the fast path be extended to `N > 1` and depth padding, or is the
   `N == 1` / `pad[0] == 0` guard fine for a first PR?

### DIFF
--- a/mlx/backend/metal/conv.cpp
+++ b/mlx/backend/metal/conv.cpp
@@ -777,12 +777,11 @@ void small_kd_conv_3D_gpu(
       binary_op_gpu_inplace({acc, conv_out}, acc, "Add", s);
       copies.push_back(conv_out);
     }
-    copies.push_back(in_2d);
-    copies.push_back(wt_2d);
   }
 
   // Repoint the [1, OD, OH, OW, O] output at the contiguous [OD, OH, OW, O]
-  // accumulator buffer (same element count).
+  // accumulator buffer (same element count). No temporary needed for `acc`
+  // since `out` shares its buffer.
   out.copy_shared_buffer(
       acc,
       {static_cast<int64_t>(OD) * OH * OW * O,
@@ -793,7 +792,6 @@ void small_kd_conv_3D_gpu(
       {true, true, false},
       static_cast<size_t>(OD) * OH * OW * O,
       0);
-  copies.push_back(acc);
 }
 
 void dispatch_conv_3D_gpu(
--- a/mlx/backend/metal/conv.cpp
+++ b/mlx/backend/metal/conv.cpp
@@ -5,6 +5,7 @@
 
 #include "mlx/backend/gpu/copy.h"
 #include "mlx/backend/gpu/slicing.h"
+#include "mlx/backend/metal/binary.h"
 #include "mlx/backend/metal/device.h"
 #include "mlx/backend/metal/kernels.h"
 #include "mlx/backend/metal/kernels/defines.h"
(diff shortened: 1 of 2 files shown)


mlx/backend/metal/conv.cpp
```diff
@@ -676,6 +677,125 @@ void pad_and_slice_conv_3D_gpu(
...
+    if (kd == 0) {
+      // First tap owns the accumulator buffer.
+      acc = conv_out;
+    } else {
+      // Elementwise in-place accumulate (safe: add is per-element).
+      binary_op_gpu_inplace({acc, conv_out}, acc, "Add", s);
+      copies.push_back(conv_out);
+    }
+    copies.push_back(in_2d);
+    copies.push_back(wt_2d);
```
zcbenz: I might be missing something but I think there is no need to add `in_2d` and `wt_wd` to temporaries as they are just views of `in` and `wt`.

katlun-lgtm: You are right — they are just views sharing the (already retained) input buffers, so the extra retention is redundant. Removed both in eca101d9, and dropped the `copies.push_back(acc)` at the end for the same reason: `out` shares the accumulator buffer via `copy_shared_buffer`, same as the pad-and-slice path notes above. Rebuilt and re-ran the conv suite plus the #3625 repro on M3 Max — still element-exact vs the per-frame reference and speed parity holds.

zcbenz (approved):
Looks good to me, thanks!

katlun-lgtm:
Gentle ping on this one — it's approved with all CI green. Happy to rebase onto the latest main if you'd like it current; otherwise it's ready to merge whenever you're batching. Thanks!

zcbenz:
@katlun-lgtm For changes that I'm not 100% sure I keep them opened for a while and ask other maintainers for a second view.

katlun-lgtm:
Checking in on this one. It's been approved since July 3, CI went green on `237ac5f` (21/21), and it's still clean against current main — no conflicts, nothing outstanding on my side.

As I understand it the only thing holding it is the second opinion @zcbenz mentioned on July 7:

> For changes that I'm not 100% sure I keep them opened for a while and ask other maintainers for a second view.

That's a reasonable ask for this change — it decomposes small kernel-depth 3D convs into 2D convs, so the decomposition boundaries are the part worth a second pair of eyes. It just hasn't come up since, and I'd rather it not quietly age out.

If another maintainer can give that second view, I'm glad to answer questions, add cases to the test, or rerun the benchmarks on whatever shapes you'd want to see.

PR #3788, opened 2026-07-02
Fix kl_div_loss docstring formula and two doc nits

Pablosinyores:
### Summary

Three documentation fixes in the pure-Python layer, one of which corrects a formula a user would copy and get wrong results from.

**`kl_div_loss` documented formula (main fix).** The `reduction == 'none'` snippet reads `mx.exp(targets) * (targets - inputs).sum(axis)`. In Python `.sum(axis)` binds tighter than `*`, so that expression reduces the difference first and then broadcast-multiplies by `exp(targets)` — a different value *and* shape than the implementation, which sums the elementwise product over `axis`:

```python
loss = mx.sum(mx.exp(targets) * (targets - inputs), axis)
```

Corrected the snippet to `(mx.exp(targets) * (targets - inputs)).sum(axis)`. Verified against the existing `test_kl_div_loss` values: the code returns `[0.0, 0.831777]` (shape `[2]`), the old snippet returns shape `[2, 2]`, the parenthesized snippet matches the code. Added an assertion pinning the documented formula to the implementation.

**`Dropout2d` channel-order acronym.** Docstring said `NWHC`/`WHC`; MLX is channels-last, so it should be `NHWC`/`HWC` (`Dropout3d` already documents `NDHWC`). No behavior change — the layer masks whole channels.

**`Optimizer.init` argument name.** Docstring listed the argument as `model`; the parameter is `parameters`.

### Testing

`python -m unittest test_losses.TestLosses.test_kl_div_loss` passes; `black` clean. (No local Metal build — relying on CI for the full suite.)

### DIFF
--- a/python/mlx/nn/layers/dropout.py
+++ b/python/mlx/nn/layers/dropout.py
@@ -40,7 +40,7 @@ class Dropout2d(Module):
 
     Randomly zero out entire channels independently with probability :math:`p`.
     This layer expects the channels to be last, i.e. the input shape should be
-    ``NWHC`` or ``WHC`` where:``N`` is the batch dimension,``H`` is the input
+    ``NHWC`` or ``HWC`` where:``N`` is the batch dimension,``H`` is the input
     image height,``W`` is the input image width, and``C`` is the number of
     input channels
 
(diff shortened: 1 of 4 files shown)


zcbenz (approved):
Thanks!

PR #3790, opened 2026-07-02
Fix zero VJPs for scatter indices

### LINKED ISSUE #3787: [BUG] ScatterAxis::vjp throws "Cannot calculate VJP with respect to indices" for computed indices (blocks autograd through MoE routing)
## Describe the bug

`ScatterAxis::vjp` (and `Scatter::vjp`) throw `Cannot calculate VJP with respect to indices.` whenever the **index** argument of a scatter/gather-by-index op lies on the autodiff path — i.e. when the indices are *computed from* a tensor we are differentiating with respect to. Instead of contributing a **zero** cotangent for the (non-differentiable, integer) index argument, the primitive raises.

This makes `mx.grad`/`value_and_grad` fail for any graph where indices are derived from a trainable tensor. In particular it **blocks LoRA fine-tuning of quantized Mixture-of-Experts models** (the expert routing scatters/gathers by top-k indices that depend on the trainable router), e.g. GLM-5.2 (`glm_moe_dsa`) via `mlx-lm` — training crashes deterministically in the backward pass. Downstream report: ml-explore/mlx-lm (companion issue).

## Minimal reproduction

```python
import mlx.core as mx  # 0.31.2

def f(w):
    idx = mx.argsort(w, axis=1)[:, :1]                       # indices derived from w
    out = mx.put_along_axis(mx.zeros((3, 3)), idx,           # ScatterAxis
                            w.sum(axis=1, keepdims=True), axis=1)
    return out.sum()

mx.grad(f)(mx.ones((3, 3)))
# ValueError: [scatter_axis] Cannot calculate VJP with respect to indices.
```

Notes:
- `mx.argmax` + `mx.take_along_axis` does **not** reproduce, because `argmax` yields a stop-gradient output, so the index never enters `argnums`. `argsort` has no VJP registered (its output is treated as differentiable and flows into the scatter's index arg), so `ScatterAxis::vjp` is called with `num == 1` and throws.

## Root cause (current `main`)

`mlx/primitives.cpp`:
- `ScatterAxis::vjp` (~L4570) handles `num == 0` (input) and `num == 2` (updates) but `throw`s for the index arg:
  ```cpp
  throw std::invalid_argument(
      "[scatter_axis] Cannot calculate VJP with respect to indices.");   // ~L4598
  ```
- `Scatter::vjp` (~L4400) does the same (`~L4486`), and `GatherMM::vjp` (`~L6013`) similarly.

The mathematically-correct behavior is a **zero** cotangent for the integer index argument (indices are non-differentiable), rather than raising — the same treatment `argsort`/`argmax` VJPs should return zero.

## Related prior work (both closed, unmerged)

- #3612 "Return zero VJPs for scatter indices" — proposed exactly this for `Scatter`; closed without merge. Does not cover `ScatterAxis`.
- #2345 "Add zero for argsort vjp" — closed without merge; would also resolve this repro at the `argsort` end.

## Proposed fix

Return a zero-valued cotangent for the index `argnum` in `Scatter::vjp`, `ScatterAxis::vjp`, and `GatherMM::vjp` (and/or register a zero VJP for `argsort`), instead of throwing. Happy to open a PR along the lines of #3612 extended to `ScatterAxis` if that direction is acceptable.

## Environment

- mlx **0.31.2**
- Python 3.12.13, macOS 26.2, Apple Silicon (M3 Ultra, 512 GB)
- Confirmed `throw` sites still present on `main` as of filing.

wgu9:
Fixes #3787.

Summary:
- Return zero VJPs for non-differentiable index arguments in Scatter, ScatterAxis, and GatherMM instead of throwing.
- Add regression coverage for computed indices flowing through argsort into scatter, put_along_axis, and gather_mm.

Validation:
- git diff --check
- python -m py_compile python/tests/test_autograd.py python/tests/test_blas.py

I also tried `python setup.py build_ext --inplace` locally, but this machine is missing the Metal toolchain:

`cannot execute tool 'metal' due to missing Metal Toolchain; use: xcodebuild -downloadComponent MetalToolchain`

So I could not run the new MLX runtime tests locally in this environment.

### DIFF
--- a/mlx/primitives.cpp
+++ b/mlx/primitives.cpp
@@ -4482,8 +4482,7 @@ std::vector<array> Scatter::vjp(
         }
       }
     } else {
-      throw std::invalid_argument(
-          "[scatter] Cannot calculate VJP with respect to indices.");
+      vjps.push_back(zeros_like(primals[num], stream()));
     }
   }
   return vjps;
@@ -4594,8 +4593,7 @@ std::vector<array> ScatterAxis::vjp(
     } else if (num == 2) {
       vjps.push_back(take_along_axis(cotangents[0], indices, axis_, stream()));
     } else {
-      throw std::invalid_argument(
-          "[scatter_axis] Cannot calculate VJP with respect to indices.");
+      vjps.push_back(zeros_like(primals[num], stream()));
     }
   }
   return vjps;
@@ -6009,8 +6007,7 @@ std::vector<array> GatherMM::vjp(
           -2,
           stream()));
     } else {
-      throw std::invalid_argument(
-          "[GatherMM] Cannot calculate VJP with respect to indices.");
+      vjps.push_back(zeros_like(primals[arg], stream()));
     }
   }
   return vjps;
--- a/python/tests/test_autograd.py
+++ b/python/tests/test_autograd.py
@@ -444,6 +444,25 @@ class TestAutograd(mlx_tests.MLXTestCase):
         self.assertTrue(mx.array_equal(dfdx, mx.array([1.0, 1.0])))
         self.assertEqual(dfdx.dtype, mx.float32)
 
+    def test_scatter_index_vjp_is_zero(self):
+        def scatter_fun(w):
+            idx = mx.argsort(w)[:2]
+            out = mx.zeros((4,))
+            out[idx] = w[:2]
+            return out.sum()
+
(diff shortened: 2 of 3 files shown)


zcbenz:
Please check https://github.com/ml-explore/mlx/pull/3612#issuecomment-4736395632.

PR #3792, opened 2026-07-03
Fix CUDA RMSNorm small-row dispatch

dhiltgen:
## Proposed changes

CUDA RMSNorm used a launch shape that packed multiple rows into one block even when each row required more than one reduction group. BlockBroadcastReduce is block-wide, so those rows mixed partial reductions and produced incorrect results for small dimensions such as fp32 D=256 and bf16/fp16 D=512.

Fix the affected dispatch case by launching one row per block when two reduction groups are needed. Add CUDA regression coverage for forward and VJP fast RMSNorm on the affected small multi-row shapes.

## Checklist


zcbenz (approved):
Nice catch, thanks!

PR #3793, opened 2026-07-03
Fix Conv3d docstring argument order (padding before dilation)

Pablosinyores:
The `Conv3d` docstring lists the constructor arguments as `stride`, `dilation`, `padding`, but the actual signature order is `stride`, `padding`, `dilation` — matching `Conv1d` and `Conv2d`.

A caller who follows the docstring order with positional arguments (e.g. `nn.Conv3d(3, 16, 3, 1, 2)` intending `dilation=2`) would instead set `padding=2`. `Conv3d` was the lone outlier here; `Conv1d` and `Conv2d` both document `stride -> padding -> dilation`.

Reorders the `Conv3d` docstring so `padding` precedes `dilation`, matching the signature and the sibling conv layers. Documentation-only, no behavior change.

### DIFF
--- a/python/mlx/nn/layers/convolution.py
+++ b/python/mlx/nn/layers/convolution.py
@@ -179,9 +179,9 @@ class Conv3d(Module):
         kernel_size (int or tuple): The size of the convolution filters.
         stride (int or tuple, optional): The size of the stride when
             applying the filter. Default: ``1``.
-        dilation (int or tuple, optional): The dilation of the convolution.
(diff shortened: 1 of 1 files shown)


Pablosinyores:
Heads-up: the red macOS (15.0) check is the flaky distributed test (`mpirun -np 8 mpi_test_distributed.py` segfaults, exit 139) — all 773 unit tests pass. This PR is docstring-only, so the failure is unrelated to the change.

zcbenz (approved):
Thanks!

Issue #3777, opened 2026-06-29
[BUG] JACCL segfaults in ibv_reg_mr (null PD) when RDMA absent; distributed_config says "RDMA not enabled" despite rdma_ctl status=enabled, survives cold boot (macOS 26.5.1 / 25F80)


andrewkriley:
## Environment
2× **M4 Mac mini (Mac16,10), 16 GB**, **macOS 26.5.1 (25F80)**, direct Thunderbolt cable (**40 Gb/s** link confirmed), **MLX 0.31.2**. JACCL driven in practice via **EXO 1.0.71** (bundles `libjaccl`/`libmlx`/`libibverbs`).

## (1) `distributed_config` reports "RDMA not enabled" despite `rdma_ctl status: enabled`
```
mlx.distributed_config --over thunderbolt --backend jaccl --hosts localhost,mac-mini-1
  ...
  [ERROR] <host> does not seem to have RDMA enabled
```
on **both** hosts, even though:
- `rdma_ctl status` → `enabled`
- `nvram rdma-enable` → `1`
- `ibv_devices` → empty (**No IB devices found**); `/dev/infiniband` is **absent**

We exhausted the local enablement path: ran `rdma_ctl enable` (twice) + reboot, then a **full coordinated cold boot** of both nodes (powered fully off with the TB cable attached, cold-powered both back up). `ibv_devices` **still enumerates zero devices on macOS 25F80**. The Thunderbolt link itself is healthy (`system_profiler SPThunderboltDataType`: peer connected, 40 Gb/s).

**Question:** what is actually required for `ibv_devices` to enumerate `rdma_enX` after enabling on macOS 26.5 (25F80)? Did a point OS update change the enablement procedure (e.g. must `rdma_ctl enable` be re-run in Recovery, or is there a newer host requirement)? The docs say enablement can't be done over SSH, but a Recovery re-enable + cold boot here has not produced any verbs device.

## (2) JACCL **segfaults** in this state instead of erroring cleanly
When the JACCL backend is initialised while no RDMA device is enumerated, it crashes rather than reporting the missing device:
```
EXC_BAD_ACCESS (SIGSEGV), KERN_INVALID_ADDRESS at 0x0
  libibverbs.dylib  ibv_reg_mr_iova2
  libjaccl.dylib    jaccl::SharedBuffer::register_to_protection_domain(ibv_pd*)
  libjaccl.dylib    jaccl::MeshGroup::allocate_buffers()
[...]

PR #3802, opened 2026-07-05
Fix JACCL error handling for missing RDMA devices

furk4neg3:
## Summary

This PR improves JACCL error handling when requested RDMA devices are missing or unavailable.

Instead of allowing initialization to continue far enough that memory registration can be attempted with a null or invalid protection domain, JACCL now fails earlier with clearer runtime errors.

Partially addresses #3777 by fixing the crash/error-handling path.

## Context

Issue #3777 appears to involve two separate concerns:

1. `mlx.distributed_config` reports RDMA as unavailable because `ibv_devices` does not enumerate any `rdma_*` devices.
2. JACCL can later crash during initialization instead of reporting the missing RDMA device/protection domain cleanly.

This PR focuses on the second part.

If the OS/hardware does not expose RDMA verbs devices, JACCL cannot initialize successfully. This PR does not change RDMA enablement behavior, macOS setup requirements, Thunderbolt hardware support detection, or entitlement behavior.

## Root Cause

When a requested RDMA device was missing or unavailable, JACCL could continue into later initialization with missing/null RDMA state. In particular, a null or invalid protection domain could reach memory registration and crash inside `ibv_reg_mr` instead of producing an actionable JACCL error.

## Fix

* Fail fast when a requested non-empty RDMA device name is not found.
* Preserve empty-device placeholders for intentional self entries.
* Add clearer diagnostics for missing RDMA devices, including available-device context where possible.
* Guard null protection domains before memory registration.
* Add defensive checks before mesh/ring buffer registration uses peer protection domains.
* Add a conditional doctest for null protection-domain registration when the `jaccl` target is available.

## Testing

* Ran `pre-commit run --all-files`.
* Ran `git diff --check`.
* Ran `cmake -S . -B build`.
* Ran `cmake --build build -j`.
* Ran `ctest --test-dir build --output-on-failure` — 261/261 tests passed.

Note: the JACCL-specific doctest is gated on the `jaccl` CMake target. In my local environment, CMake configured MLX with the `no_jaccl` path because the installed macOS SDK does not provide the required JACCL/RDMA headers, so full JACCL/RDMA initialization was not exercised locally.

### DIFF
--- a/mlx/distributed/jaccl/lib/jaccl/mesh.cpp
+++ b/mlx/distributed/jaccl/lib/jaccl/mesh.cpp
@@ -81,6 +81,25 @@ void MeshGroup::initialize() {
 }
 
 void MeshGroup::allocate_buffers() {
+  if (connections_.size() != static_cast<size_t>(size_)) {
+    std::ostringstream msg;
+    msg << "[jaccl] Expected " << size_ << " mesh connections but found "
+        << connections_.size() << ".";
+    throw std::runtime_error(msg.str());
+  }
+  for (int peer = 0; peer < size_; peer++) {
+    if (peer == rank_) {
+      continue;
+    }
+    if (connections_[peer].protection_domain == nullptr) {
+      std::ostringstream msg;
+      msg << "[jaccl] Missing RDMA protection domain for mesh peer " << peer
+          << ". Check that ibv_devices lists rdma_* devices and that "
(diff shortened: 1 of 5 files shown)


zcbenz (commented):
> so full JACCL/RDMA initialization was not exercised locally.

Thanks for the PR but you have to at least test that the error actually throws in the case of JACCL without RMDA devices.

Regarding the implementation it adds a lot of null checks which I think most of them are not necessary so I'm closing this PR.

PR #3814, opened 2026-07-07
Add matmul epilogue fusion (AddMM)

UndercoverMathGuy:
## What

- Adds a general-purpose matmul epilogue compiler that greedily merges `Compiled` objects (fused elementwise chains) after matmuls into a new `CompiledMatmul` object with a new compiler pass running after `compile_fuse` named `compile_fuse_matmul`

- Currently, as there is no backend codegen for arbitrary epilogues, `AddMM` kernels enable this kernel optimization for operations that can be represented as:

$\alpha\cdot(a@b)+\beta\cdot c \text{ where }\alpha \text{, } \beta \text{ are scalar constants and c is a runtime input}$

- A fallback path is present while backend codegen catches up: non-AddMMs run the matmul into a scratch accumulator and then run the compiled elementwise epilogue separately.
 
## Why

This is a first step toward general matmul epilogue fusion to further accelerate memory bandwidth-bound workloads such as linear layers in an MLP (xW+b)

This introduces the compiler-side representation and rewrite machinery independently of full arbitrary epilogue backend codegen. Existing `AddMM` kernels provide the first optimized path for epilogues matching `alpha * (a @ b) + beta * c`, while the fallback path keeps more general epilogues correct until backend support is added for more arbitrary elementwise chains. 

## Details

- Adds a new Primitive named `CompiledMatmul` that contains an anchor `Matmul` object and a epilogue `Compiled` object
- Updated fusion process with new `compile_fuse_matmul` pass to greedily absorb epilogue `Compiled`s into `CompiledMatmul` objects
- Temporary detection function `CompiledMatmul::matches_addmm`: recognizes and gates AddMM kernel fusion
- Adds support for `CompiledMatmul` evaluation for CUDA, CPU and Metal backends with the fallback and AddMM pattern detection and dispatch
- Guards against matmul epilogue fusion for: multi-parent producers, duplicate producer references, global outputs, cross-stream consumers, broadcast consumers, non-fusable consumers, and multi-output epilogues.
- Handles broadcasts before dispatching `c` for AddMM 

## Tests

- Adds compile tests for
    - Matmul + bias fusion
    -  $\alpha$, $\beta$ detection and AddMM dispatch conditions
    -  Fallback epilogue detection and de-fusion
    -  Guarded do-not-fuse cases


zcbenz (commented):
Thanks for the PR, fusing matmul with epilogue is really a good thing to have, but I think we should implement necessary facilities first and then add the compilation support, rather than the reverse order which usually end up with rewriting the whole thing.

I think a good way to begin with is to check what kinds of epilogue fusion are provided in cuBLAS, and then refactor our Metal and CPU kernels to be capable of doing the same level of epilogue fusion, and then extend our matmul APIs to support custom epilogue fusion. Only after we have done all the preparation work should we start implementing automatic matmul fusion in compile.

Implementing fusion for AddMM alone is not very useful in my opinion, the layers that can use them should have already been using it.

UndercoverMathGuy:
Hi @zcbenz, thanks for the feedback - that really makes sense - I'll start from the kernels going upwards. 

Did some investigation and cuBLAS seems to run with a small, pre-defined set of custom built epilogues (ReLU, bias, GELU etc.) while CUTLASS has a more ambitious Epilogue Visitor Tree that enables near-arbitrary epilogue fusion by building a tree of elementwise ops.

The current Metal system seems to already be somewhat templated on basic Epilogues (None, Axpby).

Which direction would you like me to take this forward between the cuBLAS and CUTLASS approaches?

If we're going cuBLAS, should I copy the epilogue set or are there any changes/extensions we should make?

In the meantime, I'm parking this PR for reference unless you'd rather I close it.

zcbenz:
I think the answer would be clear if you read more about MLX's matmul implementations. To be honest I don't think this is a good task for beginners, you have to fully understand the matmul implementations of all backends to be able to make a good design, and AI is really bad at this and we had to close PRs that tried to make too ambitious changes in most cases.

I don't think the approach taken in this PR would be used in the actual implementation, so I'm closing this PR.

PR #3817, opened 2026-07-08
Fix BatchNorm running variance estimator

ishtihoss:
## Summary
- update BatchNorm running_var with the unbiased variance estimate while keeping training normalization on the existing biased batch variance
- raise for the one-value-per-channel training case, matching PyTorch because unbiased variance is undefined
- cover 2D, 3D, and 4D BatchNorm training outputs, running stats, and eval outputs against PyTorch, plus the one-value error case

## Tests
- black --check python/mlx/nn/layers/normalization.py python/tests/test_nn.py
- python3 -m py_compile python/mlx/nn/layers/normalization.py python/tests/test_nn.py
- git diff --check upstream/main..HEAD
- PYTHONPATH=python:python/tests /Users/porkr/projects/contributor/.venv/bin/python -m unittest python.tests.test_nn.TestLayers.test_batch_norm python.tests.test_nn.TestLayers.test_batch_norm_matches_torch python.tests.test_nn.TestLayers.test_batch_norm_stats -v
  - test_batch_norm_matches_torch skips locally when PyTorch is unavailable

### DIFF
--- a/python/mlx/nn/layers/normalization.py
+++ b/python/mlx/nn/layers/normalization.py
@@ -336,13 +336,14 @@ class BatchNorm(Module):
             f"track_running_stats={self.track_running_stats}"
         )
 
-    def _calc_stats(self, x: mx.array) -> Tuple[mx.array, mx.array]:
+    def _calc_stats(self, x: mx.array, ddof: int = 0) -> Tuple[mx.array, mx.array]:
         """
         Calculate the mean and variance of the input tensor across the batch
         and spatial dimensions.
 
         Args:
             x (array): Input tensor.
+            ddof (int): Delta degrees of freedom for variance.
 
         Returns:
             tuple: Tuple containing mean and variance.
@@ -350,7 +351,7 @@ class BatchNorm(Module):
         reduction_axes = tuple(range(0, x.ndim - 1))
 
         mean = mx.mean(x, axis=reduction_axes)
-        var = mx.var(x, axis=reduction_axes)
+        var = mx.var(x, axis=reduction_axes, ddof=ddof)
 
         return mean, var
 
@@ -369,21 +370,21 @@ class BatchNorm(Module):
                 f"Expected input tensor to have 2, 3 or 4 dimensions, but got {x.ndim}"
             )
 
+        if self.training:
+            stats_size = 1
+            for size in x.shape[:-1]:
+                stats_size *= size
+            if stats_size == 1:
+                raise ValueError(
+                    "BatchNorm training requires more than one value per channel."
+                )
+
         # Calculate the mean and variance used to normalize the input x. If we
         # are in training mode update the running stats if needed.
         mean, var = self._calc_stats(x)
         if self.training and self.track_running_stats:
             mu = self.momentum
-            stats_size = 1
-            for size in x.shape[:-1]:
-                stats_size *= size
-            # Normalize with biased batch variance, but store the unbiased
-            # estimate for inference. Keep stats_size == 1 finite.
-            running_var = (
-                mx.var(x, axis=tuple(range(0, x.ndim - 1)), ddof=1)
-                if stats_size > 1
-                else var
-            )
+            _, running_var = self._calc_stats(x, ddof=1)
             self.running_mean = (1 - mu) * self.running_mean + mu * mean
             self.running_var = (1 - mu) * self.running_var + mu * running_var
         elif self.track_running_stats:
--- a/python/mlx/nn/layers/normalization.py
+++ b/python/mlx/nn/layers/normalization.py
@@ -374,8 +374,14 @@ class BatchNorm(Module):
         mean, var = self._calc_stats(x)
         if self.training and self.track_running_stats:
             mu = self.momentum
+            stats_size = 1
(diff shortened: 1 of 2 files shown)


zcbenz (changes requested):
Reading the [original paper](https://arxiv.org/abs/1502.03167) I think the code should use unbiased variance for inference but this PR seems to use it for training instead?

Image

My other question is that our implementation calculates mean/var over the whole batch which is correct for training but I think for inference we only want to normalize per input?

For tests we should verify the results against PyTorch's BatchNorm, things seem tricky to get correct.

Pablosinyores:
I looked into this since I've been in these normalization layers recently, and I think the PR is actually consistent with the paper — the unbiased estimate is applied only to `running_var`, not to the training-time normalization.

Walking `__call__`:
- `mean, var = self._calc_stats(x)` gives the biased batch variance (ddof=0).
- In the `self.training` branch the PR derives `running_var = var * n/(n-1)` (unbiased) and folds only that into `self.running_var`.
- The normalization a few lines down, `x = (x - mean) * mx.rsqrt(var + self.eps)`, still uses the biased `var` — that line is untouched.

So training normalization stays on the biased batch variance and only the inference statistic becomes unbiased, which matches Ioffe & Szegedy (inference `Var[x] = m/(m-1)·E[σ²_B]`) and PyTorch, whose docs state the batch is normalized with the biased estimator while `running_var` stores the unbiased one.

Quick sanity check (n=5): `var(x, ddof=1)/var(x, ddof=0)` is exactly `1.25 = n/(n-1)`, so the `stats_size/(stats_size-1)` factor is the standard Bessel correction, and the `stats_size == 1` guard keeps `running_var` finite where the unbiased estimate is undefined.

Might be worth a one-line code comment on why `running_var` uses the corrected estimate while the batch is normalized with the biased one — that asymmetry is exactly the part that's easy to misread.

zcbenz:
Thanks a lot for looking into this! After checking PyTorch's implementation I agree this PR does the right fix.

On the implementation I think we can use `mx.var(ddof=1)` to get unbiased variance directly. For the tests I do prefer verification against PyTorch's implementation rather than reimplementing the algorithm in the test code.

ishtihoss:
Thanks, updated in 9a4c1be1:\n\n- switched the running_var update to guarded `mx.var(..., ddof=1)` while keeping training normalization on the biased batch variance\n- added a PyTorch parity test for 2D, 3D, and 4D BatchNorm covering training outputs, running stats, and eval outputs\n- kept the one-value-per-channel fallback as an MLX-only test since PyTorch raises for that case

python/mlx/nn/layers/normalization.py
```diff
@@ -371,8 +371,18 @@ def __call__(self, x: mx.array) -> mx.array:
         mean, var = self._calc_stats(x)
         if self.training and self.track_running_stats:
             mu = self.momentum
+            stats_size = 1
+            for size in x.shape[:-1]:
+                stats_size *= size
+            # Normalize with biased batch variance, but store the unbiased
+            # estimate for inference. Keep stats_size == 1 finite.
+            running_var = (
+                mx.var(x, axis=tuple(range(0, x.ndim - 1)), ddof=1)
```
zcbenz: Can you update `_calc_stats` to return unbiased variance for `self.training` instead?

For `stats_size == 1` I'm not sure if returning biased variance would be correct, what does PyTorch do and should we throw error instead?

python/tests/test_nn.py
```diff
@@ -780,6 +787,70 @@ def test_batch_norm(self):
...
+            torch_bn = torch_module(features, eps=eps, momentum=momentum)
+            with torch.no_grad():
+                torch_bn.weight.copy_(torch.from_numpy(weight_np))
+                torch_bn.bias.copy_(torch.from_numpy(bias_np))
+            x_torch_np = x_np.transpose(to_torch) if to_torch else x_np
+            torch_y = torch_bn(torch.from_numpy(x_torch_np)).detach().numpy()
+            if from_torch:
+                torch_y = torch_y.transpose(from_torch)
+
+            np.testing.assert_allclose(np.array(mlx_y), torch_y, rtol=1e-4, atol=1e-4)
```
zcbenz: Should use `self.assertTrue(mx.allclose(mlx_y, mx.array(torch_y)))`

ishtihoss: Updated in 41d888ab: the PyTorch parity checks now use `self.assertTrue(mx.allclose(..., mx.array(...), ...))` with explicit tolerances.

python/mlx/nn/layers/normalization.py
ishtihoss: Updated in 41d888ab: `_calc_stats` now takes `ddof`, the running variance path uses `_calc_stats(x, ddof=1)`, and training now raises when there is only one value per channel to match PyTorch instead of falling back to the biased variance.

zcbenz (approved):
Thanks!

ishtihoss:
Hi @zcbenz, thanks again for the approval. I synced the branch with main, and the new CI run is waiting for maintainer approval. When convenient, could you approve the run and squash-merge once it's green?

PR #3819, opened 2026-07-08
Document default values in normalization layer docstrings

Pablosinyores:
The `LayerNorm`, `RMSNorm` and `GroupNorm` docstrings list `eps`, `affine` and `bias` without their default values, unlike `BatchNorm` and `InstanceNorm` in the same module which do.

This adds the missing `Default:` annotations so the docstrings match the constructor signatures and the surrounding convention:

- `LayerNorm`: `eps=1e-5`, `affine=True`, `bias=True`
- `RMSNorm`: `eps=1e-5`
- `GroupNorm`: `eps=1e-5`, `affine=True`, `pytorch_compatible=False`

Docs only, no behavior change.

### DIFF
--- a/python/mlx/nn/layers/normalization.py
+++ b/python/mlx/nn/layers/normalization.py
@@ -87,12 +87,13 @@ class LayerNorm(Module):
 
     Args:
         dims (int): The feature dimension of the input to normalize over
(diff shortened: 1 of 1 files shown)


zcbenz (approved):
Thanks!

PR #3820, opened 2026-07-08
Make index autodiff errors explicitly recommend stop_gradient

dogukanveziroglu:
Based on the discussion in #3612 and the closing note in #1439, this makes gather match the other index-based ops, if gradients reach index arguments, users now get an explicit error asking them to use `mx.stop_gradient` instead of getting a silent zero VJP. 

I also updated the related error messages and added small regression tests for both the error case and the intended workaround. (Tests might be unnecessary so I can trim them if you want)

This actually happens on mask VJP error messages too. I didn't add them on this PR but I can add a similar `stop_gradient` on there too.


mlx/primitives.cpp
```diff
@@ -2481,9 +2481,9 @@ std::vector<array> Gather::vjp(
   std::vector<array> vjps;
   for (int argnum : argnums) {
     if (argnum > 0) {
-      // Grads w.r.t. indices are zero
-      vjps.push_back(
-          zeros(primals[argnum].shape(), primals[argnum].dtype(), stream()));
+      throw std::invalid_argument(
+          "[gather] Cannot calculate VJP with respect to indices. "
+          "Use mx.stop_gradient on indices to make this explicit.");
```
zcbenz: This is a little confusing, maybe saying something like  "Use stop_gradient on indices to stop gradients from being computed.". Also we don't add "mx." prefix when mentioning ops.

zcbenz (commented):
Looks good to me, thanks! I think it makes sense also adding the error message to masked ops.

mlx/primitives.cpp
dogukanveziroglu: Sure I will change it that way.

zcbenz (approved):
👍

dogukanveziroglu:
Force pushed for fixing my author signature. Plus fixing the CI issue.

PR #3823, opened 2026-07-09
Fix int64 type cast error when loading GGUF metadata arrays

danlee2002:
## Proposed changes
Small nit but basically fixes an incorrect `reinterpret_cast` from `uint64_t*` to `int64_t*` when loading 64-bit integer metadata arrays. Doesn't have impact on actual behavior of the program as `array::init()` copies the data into `int64_t` storage. 
### Testing 
- Ran existing tests
- Added tests for negative metadata for 64-bit integer array cases 
## Checklist

### DIFF
--- a/mlx/io/gguf.cpp
+++ b/mlx/io/gguf.cpp
@@ -161,7 +161,7 @@ void set_mx_value_from_gguf(
           value = array(reinterpret_cast<uint64_t*>(data), {size}, uint64);
           break;
         case GGUF_VALUE_TYPE_INT64:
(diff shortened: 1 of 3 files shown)


zcbenz (commented):
Thanks for the fix! The python and cpp tests are testing the exact same thing so I think the cpp test along is enough.

danlee2002:
> Thanks for the fix! The python and cpp tests are testing the exact same thing so I think the cpp test along is enough.

Done.

zcbenz (approved):
👍

PR #3622, opened 2026-06-04
NAX requires setting MACOSX_DEPLOYMENT_TARGET=26.2

zcbenz:
Close https://github.com/ml-explore/mlx/issues/3586.

The neural accelerator requires macOS 26.2 and with Xcode 26.5 we have to explicitly pass `-mmacosx-version-min=26.2` to metal toolchain to compile NAX kernels correctly.

This PR does 2 things:
* Target 26.2 instead of 26.0 in CI.
* Target the host macOS version when `MACOSX_DEPLOYMENT_TARGET` is not set.

Note that strictly we should provide 2 distributions separately for 26.0 and 26.2 since the binaries would be different, but setuptools [hard-codes plat name as`macosx_26_0`](https://github.com/pypa/setuptools/blob/84ed5913724df5a12dc804e1d5efe12508e706d2/setuptools/_vendor/wheel/macosx_libfile.py#L425) for wheels built for 26.x and I'm afraid the package managers also do the same, so users would be downloading binaries built for 26.2 on 26.0.

PR #3824, opened 2026-07-09
Warn at configure time when NAX kernels are disabled

### LINKED ISSUE #3821: Source builds silently drop the NAX kernels when MACOSX_DEPLOYMENT_TARGET < 26.2 — no configure-time warning
Since #3622, `mlx/backend/metal/kernels/CMakeLists.txt` only builds the `*_nax` kernels when **`CMAKE_OSX_DEPLOYMENT_TARGET >= 26.2`** in addition to `MACOS_SDK_VERSION >= 26.2`; otherwise it defines `MLX_METAL_NO_NAX`, which makes `is_nax_available()` a hard `false`:

```cmake
if(MLX_METAL_VERSION GREATER_EQUAL 400
   AND MACOS_SDK_VERSION VERSION_GREATER_EQUAL 26.2
   AND CMAKE_OSX_DEPLOYMENT_TARGET VERSION_GREATER_EQUAL 26.2)
  ... build *_nax kernels ...
else()
  target_compile_definitions(mlx PRIVATE MLX_METAL_NO_NAX)
endif()
```

The problem: a plain `pip install .` (or `uv pip install .`) on macOS inherits `MACOSX_DEPLOYMENT_TARGET` from CPython's build config (typically **11.0** for python.org/uv interpreters), so on M5-class hardware with a current SDK the resulting build **silently drops every NAX kernel**. There is no warning — the only trace is a `message(STATUS ...)` line swallowed by pip's captured output.

The perf cliff is large, and only on GEMM-shaped work, which makes it easy to misdiagnose as a kernel regression (we spent a morning bisecting exactly that):

| kernel (M5 Max, macOS 26.5, main @ e9463bb) | target ≥ 26.2 | target < 26.2 (NO_NAX) |
|---|---|---|
| bf16 GEMM 4096×4096×4096 | 2.5–3.8 ms | 9.2 ms (~2.7×) |
| affine 4-bit qmm, M=32768, 2880→11520 | 38 ms | 158 ms (~4×) |
| SDPA causal L=32k, 32h/8kv, d64 | 70 ms | 355 ms (~5×) |
| end-to-end 32k one-shot MoE prefill (gpt-oss-20b via mlx-lm) | 13.7 s | 45.7 s |
| M=1 decode-shape kernels (qmv etc.) | unchanged | unchanged |

Two aggravating details:

1. **`26.0` also trips it** (verified with `MACOSX_DEPLOYMENT_TARGET=26.0 pip install .`). On v0.31.2 the gate was SDK-only, so the same build recipe that produced a NAX-enabled 0.31.2 source build now produces a NAX-less build of main — the "regression" appears exactly when upgrading the source checkout, which strengthens the wrong conclusion.
2. Decode-shape ops are untouched, so casual `mlx_lm.generate` smoke tests look fine; only prefill/batch workloads show it.

### Suggestion

Any (or all) of:

- `message(WARNING "NAX kernels disabled: CMAKE_OSX_DEPLOYMENT_TARGET=${CMAKE_OSX_DEPLOYMENT_TARGET} < 26.2 (SDK ${MACOS_SDK_VERSION} supports them). Set MACOSX_DEPLOYMENT_TARGET=26.2 to enable.")` in the `else()` branch when the SDK/Metal-version conditions would otherwise qualify;
- defaulting the kernel deployment target to the host macOS version when the env var comes from the Python toolchain rather than the user (the top-level `CMakeLists.txt` already falls back to `MACOS_VERSION` when the variable is empty — it just never is under pip);
- a line in the build docs next to the source-install instructions.

Happy to send a PR for the warning if that's the preferred shape.

### Environment

- M5 Max, macOS 26.5, Xcode 26.6 (Metal toolchain component installed)
- mlx main @ e9463bb, Python 3.12.13
- verified both directions: `MACOSX_DEPLOYMENT_TARGET=26.0` → `MLX_METAL_NO_NAX` present in `flags.make`, GEMM 9.18 ms; `=26.2` → flag absent, GEMM 2.5 ms, same source, same machine

pierre427:
Fixes #3821.

## What

Add a `message(WARNING)` in the `else()` branch of the NAX-kernel gate in
`mlx/backend/metal/kernels/CMakeLists.txt`, so a build that silently drops the
M-series Neural-Accelerator (NAX) kernels says so — and prints the two inputs
that decided it:

```cmake
  else()
    message(
      WARNING "NAX kernels require Metal 4, macOS SDK >= 26.2, and "
              "MACOSX_DEPLOYMENT_TARGET >= 26.2 (SDK ${MACOS_SDK_VERSION}, "
              "CMAKE_OSX_DEPLOYMENT_TARGET=${CMAKE_OSX_DEPLOYMENT_TARGET}). "
              "Building without NAX kernels.")
    target_compile_definitions(mlx PRIVATE MLX_METAL_NO_NAX)
  endif()
```

## Why

The NAX gate (added in #3622) requires Metal 4, macOS SDK ≥ 26.2, **and**
`CMAKE_OSX_DEPLOYMENT_TARGET ≥ 26.2`. CMake takes the deployment target from the
`MACOSX_DEPLOYMENT_TARGET` environment variable when the Python build toolchain
exports one (with it unset, the top-level CMakeLists falls back to the host OS
version). Any target in **[14.0, 26.2)** configures fine but silently compiles
without NAX — that is exactly how we hit it: a build with target 26.0 lost the
NAX kernels with no indication, costing ≈3–4× on prefill-dominated workloads
(NAX covers GEMM / quantized-matmul / attention) and a debugging session that
initially looked like an upstream performance regression. A configure-time
warning turns that silent cliff into a one-line diagnosis; printing the actual
SDK and deployment-target values makes it self-explanatory.

## Notes on scope

- The warning fires on the **condition** (the gate evaluated false → NAX is
  being dropped), not on any inference about *why* the target is set the way it
  is. zcbenz noted on the issue he wasn't sure there is a reliable way to detect
  whether the target came from the Python toolchain vs. elsewhere — this change
  sidesteps that question entirely: it states the requirement, prints the
  observed values, and makes no claim about intent. Anyone who set the target
  deliberately can read it and move on.
- It also fires for hosts that *cannot* build NAX (older SDK / Xcode), where it
  is informational rather than actionable. If that is too chatty, the warning
  could be limited to the actionable case (`SDK ≥ 26.2` but target < 26.2) with
  a `message(STATUS)` otherwise — either shape works; this PR keeps the simple
  form.
- JIT builds are unaffected (kernels are runtime-compiled and NAX availability
  is a runtime check there); non-macOS builds never enter this file.

+5 lines, `cmake-format` (v0.6.13, the pre-commit pin) clean, no behavior change
to the build output itself.

### DIFF
--- a/mlx/backend/metal/kernels/CMakeLists.txt
+++ b/mlx/backend/metal/kernels/CMakeLists.txt
@@ -173,6 +173,11 @@ if(NOT MLX_METAL_JIT)
                  ${STEEL_NAX_ATTN_HEADERS})
 
   else()
+    message(
+      WARNING "NAX kernels require Metal 4, macOS SDK >= 26.2, and "
+              "MACOSX_DEPLOYMENT_TARGET >= 26.2 (SDK ${MACOS_SDK_VERSION}, "
+              "CMAKE_OSX_DEPLOYMENT_TARGET=${CMAKE_OSX_DEPLOYMENT_TARGET}). "
+              "Building without NAX kernels.")
     target_compile_definitions(mlx PRIVATE MLX_METAL_NO_NAX)
   endif()
 


zcbenz (approved):
Thanks!

PR #3827, opened 2026-07-09
Fixes an axis mismatch bug in matrix norm for case -1 and 1 

danlee2002:
## Proposed changes

The following PR aims to fix a  matrix norm calculation bug caused by similar reason #3756 for when we are consider the $\pm1$ norm. Since the matrix norm is calculated as a series of reduction. When `keepdims` is false and negatives axes are passed in, the method fails to adjust the `col_axis` accordingly to account for the loss in dimension. This results in either an index out of bound error or the wrong index being reduced along. The following PR normalizes the axes accordingly to account for this reduction. 
### Reproduction 
To reproduce the bug 
```Python
import mlx.core as mx
import numpy as np
import mlx.core.linalg as la

a_mx = mx.arange(9).reshape(3,3)
a_np = np.arange(9).reshape(3,3)
print(np.linalg.norm(a_np, ord = -1, axis = (-1, -2))) #5.0
print(la.norm(a_mx, ord = -1, axis = [-1, -2])) # throws exception

a_mx = mx.arange(18).reshape(2,3,3)
a_np = np.arange(18).reshape(2,3,3)
print(np.linalg.norm(a_np, ord = -1, axis = (-1, -2))) # [21., 48.]
print(la.norm(a_mx, ord = -1, axis = [-1, -2])) # array([3, 12, 21], dtype=float32)
```
## Testing 
- Existing tests were ran
- Existing test suite for negative axis reduction was extended to account for $\pm 1$ norm to account for all possible negative axes combination for `(3, 3), (2, 3, 3), (2, 3, 3, 3)` matrices 
## Checklist

### DIFF
--- a/mlx/linalg.cpp
+++ b/mlx/linalg.cpp
@@ -84,43 +84,33 @@ inline array matrix_norm(
     bool keepdims,
     StreamOrDevice s) {
   auto dtype = at_least_float(a.dtype());
-  auto row_axis = axis[0];
-  auto col_axis = axis[1];
+  int row_axis = (axis[0] < 0) ? axis[0] + a.ndim() : axis[0];
+  int col_axis = (axis[1] < 0) ? axis[1] + a.ndim() : axis[1];
   if (ord == -1.0) {
-    row_axis = (axis[0] < 0) ? axis[0] + a.ndim() : axis[0];
-    col_axis = (axis[1] < 0) ? axis[1] + a.ndim() : axis[1];
     col_axis -= (!keepdims && col_axis > row_axis && col_axis > 0);
     return astype(
         min(sum(abs(a, s), row_axis, keepdims, s), col_axis, keepdims, s),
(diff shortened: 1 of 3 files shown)


mlx/linalg.cpp
```diff
@@ -87,12 +87,16 @@ inline array matrix_norm(
   auto row_axis = axis[0];
   auto col_axis = axis[1];
   if (ord == -1.0) {
+    row_axis = (axis[0] < 0) ? axis[0] + a.ndim() : axis[0];
+    col_axis = (axis[1] < 0) ? axis[1] + a.ndim() : axis[1];
```
zcbenz: It seems that every branch is doing the same should this be moved out of `if`?

danlee2002: Done.

zcbenz (approved):
Thanks!

PR #3828, opened 2026-07-10
Fix captured random state in compile

### LINKED ISSUE #1439: temperature > 0 is silently deterministic on the second call onward — categorical_sampling's @mx.compile(outputs=mx.random.state) doesn't advance the PRNG

angeloskath:
This is an interesting bug. In our attempts at thread-safety, compile now captures the random state of the thread that created the function and not the thread that called the function which can (and often is) different.

There isn't too clean of a solution that I could think of so please be as strict as possible in your judgment. The current approach is twofold

1. Make the random state a singleton object that fetches the thread local state. This makes it ok to capture on any thread. Accessing its one element will get you the correct array based on which thread you are on.
2. The ugly part of this fix is that `tree_visit` and `tree_visit_update` now have a special case where if they encounter a random state they instead operate on the underlying array. This is ugly because it is not necessarily consistent with `tree_map` or multi-tree `tree_visit`.

This closes https://github.com/ml-explore/mlx-lm/pull/1444 and fixes https://github.com/ml-explore/mlx-lm/issues/1439 .

### DIFF
--- a/python/src/random.cpp
+++ b/python/src/random.cpp
@@ -70,16 +70,13 @@ PyKeySequence& default_key() {
 class RandomState {};
 
 nb::object random_state_sentinel() {
-  static nb::object sentinel;
-  if (sentinel.ptr() == nullptr) {
-    sentinel = nb::cast(RandomState{});
+  static nb::object sentinel = []() {
+    auto sentinel = nb::cast(RandomState{});
     sentinel.inc_ref();
-  }
-  return sentinel;
-}
+    return sentinel;
+  }();
 
-bool is_random_state(nb::handle obj) {
-  return obj.ptr() == random_state_sentinel().ptr();
+  return sentinel;
 }
 
 mx::array random_state_key() {
--- a/python/src/random.cpp
+++ b/python/src/random.cpp
@@ -9,6 +9,7 @@
 
 #include "mlx/ops.h"
 #include "mlx/random.h"
+#include "python/src/random.h"
 #include "python/src/small_vector.h"
 #include "python/src/utils.h"
 
@@ -63,15 +64,57 @@ PyKeySequence& default_key() {
   return ks;
 }
 
+// A process-global sentinel for `mx.random.state`. Since it is the same object
+// on every thread, capturing it (e.g. with `mx.compile`) is thread-independent;
+// the pytree traversal in trees.cpp resolves it to the calling thread's key.
+class RandomState {};
+
+nb::object random_state_sentinel() {
+  static nb::object sentinel;
+  if (sentinel.ptr() == nullptr) {
+    sentinel = nb::cast(RandomState{});
+    sentinel.inc_ref();
+  }
+  return sentinel;
+}
+
+bool is_random_state(nb::handle obj) {
+  return obj.ptr() == random_state_sentinel().ptr();
+}
+
+mx::array random_state_key() {
+  return nb::cast<mx::array>(default_key().state()[0]);
+}
+
+void set_random_state_key(const mx::array& key) {
+  default_key().state()[0] = nb::cast(key);
+}
+
 void init_random(nb::module_& parent_module) {
   auto m = parent_module.def_submodule(
       "random",
       "mlx.core.random: functionality related to random number generation");
 
+  nb::class_<RandomState>(m, "_RandomState")
+      .def("__len__", [](const RandomState&) { return 1; })
+      .def(
+          "__getitem__",
+          [](const RandomState&, int i) -> nb::object {
+            if (i != 0 && i != -1) {
+              throw nb::index_error("random state index out of range");
+            }
+            return default_key().state()[0];
+          },
+          "index"_a)
+      .def("__iter__", [](const RandomState&) {
+        return nb::iter(default_key().state());
+      });
+
   m.def("__getattr__", [&](nb::handle key) -> nb::object {
     // Create random.state lazily to avoid initializing device during import.
     if (nb::isinstance<nb::str>(key) && nb::cast<std::string>(key) == "state") {
-      return default_key().state();
+      default_key().state();
+      return random_state_sentinel();
     }
     return nb::steal(PyErr_Format(
         PyExc_AttributeError,
--- a/python/src/trees.cpp
+++ b/python/src/trees.cpp
@@ -1,6 +1,7 @@
 // Copyright © 2023-2024 Apple Inc.
 
 #include "python/src/trees.h"
+#include "python/src/random.h"
 
 template <typename T, typename U, typename V>
 void validate_subtrees(const std::vector<nb::object>& subtrees) {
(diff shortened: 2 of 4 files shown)


python/src/random.cpp
```diff
@@ -63,15 +64,57 @@ PyKeySequence& default_key() {
...
 }
 
+// A process-global sentinel for `mx.random.state`. Since it is the same object
+// on every thread, capturing it (e.g. with `mx.compile`) is thread-independent;
+// the pytree traversal in trees.cpp resolves it to the calling thread's key.
+class RandomState {};
+
+nb::object random_state_sentinel() {
+  static nb::object sentinel;
+  if (sentinel.ptr() == nullptr) {
```
zcbenz: This is not thread safe.

python/src/trees.cpp
```diff
@@ -153,7 +154,10 @@ void tree_visit(
 void tree_visit(nb::handle tree, std::function<void(nb::handle)> visitor) {
   std::function<void(nb::handle)> recurse;
   recurse = [&](nb::handle subtree) {
-    if (nb::isinstance<nb::list>(subtree) ||
+    if (is_random_state(subtree)) {
+      visitor(nb::cast(random_state_key()));
```
zcbenz, line 159: I wonder if we can do this in a less intrusive way by replacing

https://github.com/ml-explore/mlx/blob/4367c73b60541ddd5a266ce4644fd93d20223b6e/python/src/transforms.cpp#L560-L563

with

```cpp
if (!captured_inputs.is_none()) { 
  trace_captures.push_back(random_state_sentinel());
  flat_in_captures.push_back(random_state_key());
  tree_replace(captured_inputs, trace_captures, flat_in_captures);
}
```

python/src/random.cpp
angeloskath: Yeah true let me change it to a lambda call. I just kind of took it from the structure sentinel in trees.cpp, will change both.

python/src/trees.cpp
angeloskath: The problem is the random state can be in any place in the captured inputs. As a result we have two options

1. Walk the captured inputs and replace the random state with the thread local one before flattening
2. Do it when the random state is encountered while flattening and while writing back

Option 1 has the overhead of 2 extra tree walks which for say 100k nodes (like a big transformer or sth) can be non-negligible.

Option 2 is intrusive and ugly.

zcbenz: That makes sense, we can probably add a callback to `tree_flatten` to make it cheap to do replacement but it feels over-engineered and still does not fix the hook in `tree_visit_update`, I'm good with current approach and I can't come with a better one.

zcbenz (approved):

PR #3831, opened 2026-07-10
Fix Glorot/He uniform init docstrings to label the uniform bound, not sigma

vineethsaivs:
## Proposed changes

The `glorot_uniform` and `he_uniform` docstrings in `python/mlx/nn/init.py`
write the sampling bound with the symbol `\sigma`, but both initializers draw
from a symmetric uniform distribution, `uniform(-a, a)`, so the value shown is
the bound `a`, not a standard deviation. `\sigma` conventionally denotes the
standard deviation, and for `uniform(-a, a)` the standard deviation is
`a / sqrt(3)`:

- `glorot_uniform`: bound `a = gain * sqrt(6 / (fan_in + fan_out))`, so the
  actual standard deviation is `gain * sqrt(2 / (fan_in + fan_out))`, which is
  exactly the `\sigma` documented for `glorot_normal`.
- `he_uniform`: bound `a = gain * sqrt(3 / fan)`, so the actual standard
  deviation is `gain / sqrt(fan)`, which is exactly the `\sigma` documented for
  `he_normal`.

So the documented `\sigma` for the uniform initializers is off by a factor of
`sqrt(3)` from the real standard deviation. It is also inconsistent with the
surrounding prose, which already describes the quantity as a "range", and with
the normal initializers in the same file, where `\sigma` correctly denotes the
standard deviation. For comparison, PyTorch's `kaiming_uniform_` docstring
labels the same quantity `bound` and samples `U(-bound, bound)`.

This relabels the quantity as the bound `a` and shows the sampling interval
`[-a, a]`. It is a documentation-only change with no effect on behavior; the
sampled values are unchanged.

## Checklist

- [ ] I have run `pre-commit run --all-files` to format my code / installed pre-commit prior to committing changes (docstring text only; verified it compiles and does not change `black` formatting)
- [ ] I have added tests that prove my fix is effective or that my feature works (documentation-only change; no code path is affected)

### DIFF
--- a/python/mlx/nn/init.py
+++ b/python/mlx/nn/init.py
@@ -194,12 +194,13 @@ def glorot_uniform(
 ) -> Callable[[mx.array, float], mx.array]:
     r"""A Glorot uniform initializer.
 
-    This initializer samples from a uniform distribution with a range
-    computed from the number of input (``fan_in``) and output (``fan_out``)
+    This initializer samples from a uniform distribution on the interval
+    :math:`[-\text{limit}, \text{limit}]`, where the bound :math:`\text{limit}`
+    is computed from the number of input (``fan_in``) and output (``fan_out``)
     units according to:
 
     .. math::
-        \sigma = \gamma \sqrt{\frac{6.0}{\text{fan\_in} + \text{fan\_out}}}
+        \text{limit} = \gamma \sqrt{\frac{6.0}{\text{fan\_in} + \text{fan\_out}}}
 
     For more details see the original reference: `Understanding the difficulty
     of training deep feedforward neural networks
@@ -295,13 +296,14 @@ def he_uniform(
 ) -> Callable[[mx.array, Literal["fan_in", "fan_out"], float], mx.array]:
     r"""A He uniform (Kaiming uniform) initializer.
 
-    This initializer samples from a uniform distribution with a range
-    computed from the number of input (``fan_in``) or output (``fan_out``)
+    This initializer samples from a uniform distribution on the interval
+    :math:`[-\text{limit}, \text{limit}]`, where the bound :math:`\text{limit}`
+    is computed from the number of input (``fan_in``) or output (``fan_out``)
     units according to:
 
     .. math::
 
-        \sigma = \gamma \sqrt{\frac{3.0}{\text{fan}}}
+        \text{limit} = \gamma \sqrt{\frac{3.0}{\text{fan}}}
 
     where :math:`\text{fan}` is either the number of input units when the
     ``mode`` is ``"fan_in"`` or output units when the ``mode`` is


Pablosinyores (commented):
Confirmed correct against the implementation. Both `glorot_uniform` and `he_uniform` compute `limit = gain * math.sqrt(6.0 / (fan_in + fan_out))` (resp. `sqrt(3.0 / fan)`) and then `mx.random.uniform(-limit, limit, ...)` — so that quantity is the half-width of the uniform interval, not a standard deviation. Labeling it `\sigma` overstates the spread by a factor of √3 (for `U[-a, a]` the actual std is `a/√3`), so the relabel is right. The explicit `[-a, a]` interval line you added makes the meaning unambiguous, which is the better framing.

Worth noting the normal siblings (`glorot_normal` / `he_normal`) correctly keep `\sigma` alongside `mx.random.normal(scale=std)` — so this reads like the `\sigma` label was carried over from the normal variants into the uniform ones by mistake.

One small thought: the implementation names this variable `limit`, so using `\text{limit}` as the symbol (as your #3835 does) ties the formula straight to the source. `a` with the explicit interval is also perfectly clear and arguably more conventional — either works.

Heads up that this overlaps with your own #3835 — same two functions, same lines — so they'll conflict and one probably wants closing. This is the more complete of the two (the interval prose is the useful part), so if you keep one, I'd keep this.

vineethsaivs:
Thanks for the careful review. Agreed the `\sigma` label looks carried over from the normal siblings. I have consolidated the two PRs into this one: it keeps the explicit `[-limit, limit]` interval wording and now uses `\text{limit}` as the symbol so the formula matches the code's variable name, as you suggested. Closing #3835 in favor of this PR.

Pablosinyores (approved):
Accurate fix. Both initializers compute `limit = gain * sqrt(...)` then `mx.random.uniform(-limit, limit, ...)`, so labeling that bound `\sigma` (a std-dev symbol) was wrong for a uniform range. Renaming to `\text{limit}` and stating the interval [-limit, limit] matches the variable that's literally named `limit`, and the sqrt(6/(fan_in+fan_out)) / sqrt(3/fan) factors are the variance-match to the normal variants. LGTM.

vineethsaivs:
Gentle ping @angeloskath. This is a two-line docstring fix, CI is green, and it picked up a community review confirming the diagnosis.

`glorot_uniform` and `he_uniform` compute `limit = gain * sqrt(6 / (fan_in + fan_out))` (resp. `sqrt(3 / fan)`) and then sample `mx.random.uniform(-limit, limit, ...)`. The docstrings label that quantity `\sigma`, a standard-deviation symbol, but it is the half-width of the uniform interval, not a standard deviation (for a uniform on `[-l, l]` the standard deviation is `l / sqrt(3)`). The fix renames it to `\text{limit}` and states the interval, matching the variable that is literally named `limit` in the implementation. The normal variants are untouched, since `\sigma` is correct there.

Happy to rebase if it has gone stale.

zcbenz (approved):
Thanks!

PR #3833, opened 2026-07-11
Fix custom metal kernel cache collision for same name, different source

### LINKED ISSUE #3832: [BUG] `fast.metal_kernel`: same-name kernels with different sources silently run stale code within one eval batch
**Describe the bug**
Two `mx.fast.metal_kernel` instances sharing `name` but with different `source`, both dispatched before a single `mx.eval`, silently execute the **first** kernel's compiled code for the second call. No error or warning. Return wrong results.

**To Reproduce** (affected: MLX 0.31.1, 0.31.2, 0.32.0 from PyPI, and git main `0.32.1.dev20260710+4367c73b` built from source; M4 Pro, macOS 26):
```python
import mlx.core as mx
n = 64
a = mx.arange(n, dtype=mx.float32)
mk = lambda src: mx.fast.metal_kernel(
    name="probe", input_names=["a"], output_names=["out"], source=src)
body = "uint i = thread_position_in_grid.x; if (i < %du) out[i] = a[i] %s;"
kA = mk(body % (n, "* 2.0f"))
kB = mk(body % (n, "+ 100.0f"))
run = lambda k, x: k(inputs=[x], output_shapes=[(n,)], output_dtypes=[mx.float32],
                     grid=(n, 1, 1), threadgroup=(min(n, 256), 1, 1))[0]
yA, yB = run(kA, a), run(kB, a)
mx.eval(yA, yB)                # ONE eval batch
print(yB[:3])                  # observed: [0, 2, 4]   (kernel A's code!)
                               # expected: [100, 101, 102]
# With mx.eval(yA) before building yB, the result is correct.
```

**Expected behavior**
Correct dispatch (cache keyed by name + source identity), or a loud error on name collision. Shouldn't perform silent execution of a different kernel's code.

 - OS Version: macOS 26 (Darwin 25.5.0)
 - M4 Pro, MLX 0.31.1 (PyPI), Python 3.13

**Additional context**
The custom-kernel library cache is keyed by kernel name (`Device::get_library(name_, …)` in `backend/metal/custom_kernel.cpp`); the stale-source invalidation works across eval boundaries but not within one batch. Frameworks that execute a whole program as one batch (e.g. PJRT plugins lowering to `fast::metal_kernel`) hit this deterministically for any name reuse. 

Potential workaround: suffix the name with a content hash of the source.

katlun-lgtm:
## Summary

Fixes #3832. Two `fast.metal_kernel` instances that share a `name` but have different `source`, dispatched before a single `mx.eval`, silently execute the **first** kernel's compiled code for the second — wrong results, no error.

## Cause

The custom-kernel library cache is keyed by `name_` alone (`backend/metal/custom_kernel.cpp`), with a hand-rolled "if the source changed, `clear_library(name_)` and rebuild" invalidation. That is safe *across* `eval` boundaries (the command buffer has completed before the clear) but not *within* one batch: the second kernel clears and rebuilds the shared `name_` slot while the first kernel's dispatch is still pending in the same command buffer, and the second ends up bound to the first's code.

## Fix

Key the library by `name_ + hash(source_) + compile_options_`, so kernels that differ in any of those map to distinct libraries — and therefore distinct pipelines (the pipeline cache is already keyed by `(library, function)`). They now coexist for the whole batch instead of thrashing one slot, and the fragile clear-and-rebuild (with its `CustomKernelCache` bookkeeping) is removed. `compile_options_` is still passed to `get_library`, so a kernel's math mode is honored.

The source is hashed with `std::hash<std::string>` rather than embedded whole, so the key stays short — the same treatment `backend/cpu/compiled.cpp` gives over-long kernel names. A wrong cache hit would need two kernels to share a name, share compile options, *and* collide on the 64-bit hash.

## Test

Adds `test_custom_kernel_same_name_different_source_one_eval`: two same-name / different-source kernels evaluated in a single batch must each return their own result. It fails before this change and passes after.

## Validation

Built against `main`; `python/tests/test_fast.py` passes (27/27, including the new test and the compile-options / caching tests), and the issue's repro now returns the correct `[100, 101, 102]` instead of `[0, 2, 4]`.

Thanks to @Jinge-Wang for the precise report and minimal repro.

### DIFF
--- a/mlx/backend/metal/custom_kernel.cpp
+++ b/mlx/backend/metal/custom_kernel.cpp
@@ -8,16 +8,6 @@
 
 namespace mlx::core::fast {
 
-struct CustomKernelCache {
-  std::unordered_map<std::string, std::pair<std::string, CompileOptions::Data>>
-      libraries;
-};
-
-static CustomKernelCache& cache() {
-  static CustomKernelCache cache_;
-  return cache_;
-};
-
 void CustomKernel::eval_gpu(
     const std::vector<array>& inputs,
     std::vector<array>& outputs) {
@@ -55,25 +45,15 @@ void CustomKernel::eval_gpu(
 
   auto& d = metal::device(s.device);
 
-  {
-    // Clear kernels from the device library cache if needed
-    auto& kernel_cache = cache();
-    if (auto it = kernel_cache.libraries.find(name_);
-        it != kernel_cache.libraries.end()) {
-      if (it->second.first != source_ ||
-          it->second.second != compile_options_) {
-        auto& d = metal::device(s.device);
-        d.clear_library(name_);
-        it->second = std::make_tuple(source_, compile_options_);
-      }
-    } else {
-      kernel_cache.libraries.emplace(
-          name_, std::make_tuple(source_, compile_options_));
-    }
-  }
-
+  // Key the library by (name, source, compile options) so kernels that differ
+  // in any of those coexist instead of thrashing a single cache slot within one
+  // eval batch (fixes #3832). The map key is never used as a Metal identifier,
+  // so the full source is a collision-free key (a content hash could alias
+  // distinct sources and silently reintroduce the bug).
+  std::string lib_name =
+      name_ + "\n" + source_ + "\n" + std::to_string(compile_options_);
   auto lib = d.get_library(
-      name_, compile_options_, [this] { return metal::utils() + source_; });
+      lib_name, compile_options_, [this] { return metal::utils() + source_; });
   auto kernel = d.get_kernel(name_, lib);
   auto& compute_encoder = metal::get_command_encoder(s);
   compute_encoder.set_compute_pipeline_state(kernel);
--- a/mlx/backend/metal/custom_kernel.cpp
+++ b/mlx/backend/metal/custom_kernel.cpp
@@ -45,13 +45,13 @@ void CustomKernel::eval_gpu(
 
   auto& d = metal::device(s.device);
 
-  // Key the library by (name, source, compile options) so kernels that differ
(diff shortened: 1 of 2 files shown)


mlx/backend/metal/custom_kernel.cpp
```diff
@@ -55,25 +45,15 @@ void CustomKernel::eval_gpu(
...
-    }
-  }
-
+  // Key the library by (name, source, compile options) so kernels that differ
+  // in any of those coexist instead of thrashing a single cache slot within one
+  // eval batch (fixes #3832). The map key is never used as a Metal identifier,
+  // so the full source is a collision-free key (a content hash could alias
+  // distinct sources and silently reintroduce the bug).
+  std::string lib_name =
+      name_ + "\n" + source_ + "\n" + std::to_string(compile_options_);
```
zcbenz: Can you hash the source to avoid long filename issue.

katlun-lgtm: Done in `3eb338b` — the key is now `name + std::hash<std::string>(source) + compile options`, the same shape as the file-name shortening in `backend/cpu/compiled.cpp`.

The description had argued for the raw source on collision grounds; with the hash a wrong hit needs two kernels to share a name, share compile options, *and* collide on a 64-bit hash — not worth multi-MB map keys. Description updated.

Rebuilt on an M3 Max: `python/tests/test_fast.py` 27/27, including the same-name/different-source test this PR adds. CI still hasn't run on this branch — the fork workflow run is sitting in `action_required`, so an approval click would put checks on it.

zcbenz (approved):
Thanks for the PR, the original cache design looked odd to me but I was worried about breaking things, let's just make the change and it should be safe.

PR #3834, opened 2026-07-11
Fix eigh UPLO and zero-size eigh/svd on the CPU

kapellirohith:
## Proposed changes

`eigh`/`eigvalsh` read the opposite triangle from what `UPLO` asks for.
The row-major buffer is handed to LAPACK, which reads it as its transpose,
so `UPLO="L"` actually used the upper triangle. Only observable when the
two triangles differ:

```python
>>> a = mx.array([[1.0, 999.0], [2.0, 3.0]])
>>> mx.linalg.eigvalsh(a, UPLO="L", stream=mx.cpu)
array([-997.001, 1001], dtype=float32)     # before (upper triangle)
array([-0.236068, 4.23607], dtype=float32) # after, matches numpy
```

Also fixes crashes for zero-size inputs to `eigh` and `svd`: LAPACK
rejects `lda = 0` and the batch-count division was by zero.

No behavior change for inputs with consistent triangles (symmetric or
Hermitian), which is why the existing tests didn't catch it.

## Checklist

### DIFF
--- a/mlx/backend/cpu/eigh.cpp
+++ b/mlx/backend/cpu/eigh.cpp
@@ -155,10 +155,12 @@ void eigh_impl(
   auto& encoder = cpu::get_command_encoder(stream);
   encoder.set_output_array(vectors);
   encoder.set_output_array(values);
+  // LAPACK reads the row-major input as its (conjugate) transpose, so the
+  // requested triangle is the opposite one in LAPACK's view.
   encoder.dispatch([vec_ptr,
                     eig_ptr,
                     jobz,
-                    uplo = uplo[0],
+                    uplo = uplo[0] == 'L' ? 'U' : 'L',
                     N = vectors.shape(-1),
                     size = vectors.size()]() mutable {
     // Work query
@@ -201,6 +203,15 @@ void Eigh::eval_cpu(
       vectors,
       a.flags().row_contiguous ? CopyType::Vector : CopyType::General,
       stream());
+
+  // Nothing to decompose for n = 0; LAPACK rejects lda = 0.
+  if (a.shape(-1) == 0) {
+    if (!compute_eigenvectors_) {
+      auto& encoder = cpu::get_command_encoder(stream());
+      encoder.add_temporary(vectors);
+    }
+    return;
+  }
 
   if (compute_eigenvectors_) {
     // Set the strides and flags so the eigenvectors
--- a/mlx/backend/cpu/eigh.cpp
+++ b/mlx/backend/cpu/eigh.cpp
@@ -192,28 +192,26 @@ void Eigh::eval_cpu(
   const auto& a = inputs[0];
   auto& values = outputs[0];
 
+  values.set_data(allocator::malloc(values.nbytes()));
+
+  // Nothing to decompose for n = 0; LAPACK rejects lda = 0. The input is
+  // square, so both outputs are empty and there is nothing to write.
+  if (a.shape(-1) == 0) {
+    if (compute_eigenvectors_) {
+      outputs[1].set_data(allocator::malloc(outputs[1].nbytes()));
+    }
+    return;
+  }
+
   auto vectors = compute_eigenvectors_
       ? outputs[1]
       : array(a.shape(), a.dtype(), nullptr, {});
-
-  values.set_data(allocator::malloc(values.nbytes()));
 
   copy_cpu(
       a,
       vectors,
       a.flags().row_contiguous ? CopyType::Vector : CopyType::General,
       stream());
-
-  // Nothing to decompose for n = 0; LAPACK rejects lda = 0. The input is
-  // square, so both outputs are empty here and the copy above has already
-  // initialized the eigenvectors.
-  if (a.shape(-1) == 0) {
-    if (!compute_eigenvectors_) {
-      auto& encoder = cpu::get_command_encoder(stream());
-      encoder.add_temporary(vectors);
-    }
-    return;
-  }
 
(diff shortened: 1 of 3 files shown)


mlx/backend/cpu/eigh.cpp
```diff
@@ -202,6 +204,15 @@ void Eigh::eval_cpu(
       a.flags().row_contiguous ? CopyType::Vector : CopyType::General,
       stream());
 
+  // Nothing to decompose for n = 0; LAPACK rejects lda = 0.
+  if (a.shape(-1) == 0) {
+    if (!compute_eigenvectors_) {
+      auto& encoder = cpu::get_command_encoder(stream());
+      encoder.add_temporary(vectors);
```
zcbenz: For early return the output should be set to 0, you can find an example at https://github.com/ml-explore/mlx/blob/43c9e03912ec19e5b6f28b24a6a49d1fa2f923fe/mlx/backend/cpu/matmul.cpp#L124-L129

mlx/backend/cpu/svd.cpp
```diff
@@ -206,6 +206,14 @@ void svd_impl(
 
   using R = typename SVDWork<T>::R;
 
+  // Nothing to decompose for empty matrices; LAPACK rejects lda = 0.
+  if (M == 0 || N == 0) {
+    for (auto& o : outputs) {
+      o.set_data(allocator::malloc(o.nbytes()));
```
zcbenz, line 218: Same here.

zcbenz (commented):
Nice catch!

kapellirohith:
Thanks for catching that, it was a real one. For a shape like `(5, 0)` the
factors are not empty (`U` is 5x5), so the early return was handing back
whatever the allocator happened to have in that buffer. My tests only covered
shapes where every output is empty, which is why they missed it.

I followed the matmul.cpp structure but wrote the identity rather than literal
zeros, since that is what LAPACK and numpy return for the orthonormal factors:

```python
>>> np.linalg.svd(np.zeros((5, 0)))[0]   # 5x5 identity
```

Zeros would leave `U` non-orthonormal. Happy to switch to plain zeros if you
prefer that.

eigh does not need a fill. The input is square, so `n = 0` makes both outputs
empty and there is nothing to write, and the copy above already sets the
eigenvectors. I put a comment there instead.

Also added tests for the rectangular zero-size shapes, checked against numpy
for float32, float64 and complex64, with and without `compute_uv`.

mlx/backend/cpu/eigh.cpp
```diff
@@ -202,6 +204,17 @@ void Eigh::eval_cpu(
       a.flags().row_contiguous ? CopyType::Vector : CopyType::General,
       stream());
 
+  // Nothing to decompose for n = 0; LAPACK rejects lda = 0. The input is
+  // square, so both outputs are empty here and the copy above has already
+  // initialized the eigenvectors.
+  if (a.shape(-1) == 0) {
+    if (!compute_eigenvectors_) {
+      auto& encoder = cpu::get_command_encoder(stream());
+      encoder.add_temporary(vectors);
```
zcbenz: I think you are right there is no need to fill 0 for this case, but for `a.shape(-1) == 0 && !compute_eigenvectors_` the code is copying `a` to a temporary array and then throw it?

mlx/backend/cpu/svd.cpp
```diff
@@ -206,6 +208,30 @@ void svd_impl(
...
+  // does, which keeps them orthonormal.
+  if (M == 0 || N == 0) {
+    auto& encoder = cpu::get_command_encoder(stream);
+    for (auto& o : outputs) {
+      o.set_data(allocator::malloc(o.nbytes()));
+      if (o.size() == 0) {
+        continue;
+      }
+      encoder.set_output_array(o);
+      encoder.dispatch([ptr = o.data<T>(), size = o.size(), n = o.shape(-1)]() {
```
zcbenz: The `o` should be captured here, see `ArgReduce::eval_cpu` for example.

kapellirohith:
Both done.

eigh: you are right, that copy was pointless. Moved the early return above it,
so for `n = 0` nothing is copied and no temporary is created at all, for either
value of `compute_eigenvectors_`.

svd: switched to capturing the array with `array::unsafe_weak_copy` like
`ArgReduce::eval_cpu` does.

Rebased on main (8056817) and checked rather than assumed:

- Full python suite, `python -m unittest discover python/tests`, run on a clean
  checkout of main and on this branch: `Ran 804 tests ... OK (skipped=3)` in
  both, and diffing the per test results of the two runs shows no test changing
  status.
- C++ tests: 261/261 on the metal build, 261/261 with `DEVICE=cpu`, and 245/245
  for a `MLX_BUILD_METAL=OFF` build.
- Zero size eigh and svd against numpy for float32, float64 and complex64, with
  and without `compute_uv`, including the batched shapes. Each call is made with
  the allocator pool dirtied first, so anything left uninitialized shows up as
  the dirty value rather than passing by luck.

One unrelated thing that turned up while testing, on a clean main with no
linalg involved:

```python
>>> np.array(mx.zeros((0, 0)) @ mx.zeros((0, 0)))   # segfault
```

`Matmul::eval_cpu` allocates the zero size output and returns, and the buffer
protocol then reads from it. `mx.zeros((0, 0))` on its own is fine, and
`mx.zeros((2, 0)) @ mx.zeros((0, 2))` is fine since the output is not empty.
Happy to open a separate issue or PR for that one.

zcbenz (approved):
Thanks!

PR #3835, opened 2026-07-11
Fix uniform-initializer docstrings to use the range limit, not sigma

vineethsaivs:
## Proposed changes

`nn.init.glorot_uniform` and `nn.init.he_uniform` sample from a uniform distribution on `[-limit, limit]`: the code computes `limit = gain * math.sqrt(...)` and calls `mx.random.uniform(-limit, limit, ...)`. Their docstring math, however, labels that bound as `\sigma`, the standard-deviation symbol:

```
\sigma = \gamma \sqrt{\frac{6.0}{\text{fan\_in} + \text{fan\_out}}}   # glorot_uniform
\sigma = \gamma \sqrt{\frac{3.0}{\text{fan}}}                        # he_uniform
```

The standard deviation of `U(-limit, limit)` is `limit / sqrt(3)`, not `limit`, so using `\sigma` for the bound is misleading. This renames the symbol to `\text{limit}` in both docstrings so the documented quantity matches the code variable and its true meaning.

`glorot_normal` and `he_normal` are left unchanged: they sample from a normal distribution, where `\sigma` is the correct symbol for the standard deviation.

Documentation-only change; no code paths are affected.

## Checklist

- [x] The change is docstring text only (inside a `.. math::` block), so `black` formatting is unchanged
- [ ] I have added tests (not applicable: documentation-only change, no behavior changed)
- [x] I have updated the necessary documentation

### DIFF
--- a/python/mlx/nn/init.py
+++ b/python/mlx/nn/init.py
@@ -199,7 +199,7 @@ def glorot_uniform(
     units according to:
 
     .. math::
-        \sigma = \gamma \sqrt{\frac{6.0}{\text{fan\_in} + \text{fan\_out}}}
+        \text{limit} = \gamma \sqrt{\frac{6.0}{\text{fan\_in} + \text{fan\_out}}}
 
     For more details see the original reference: `Understanding the difficulty
     of training deep feedforward neural networks
@@ -301,7 +301,7 @@ def he_uniform(
 
     .. math::
 
-        \sigma = \gamma \sqrt{\frac{3.0}{\text{fan}}}
+        \text{limit} = \gamma \sqrt{\frac{3.0}{\text{fan}}}
 
     where :math:`\text{fan}` is either the number of input units when the
     ``mode`` is ``"fan_in"`` or output units when the ``mode`` is


Pablosinyores (commented):
The relabel is correct. `glorot_uniform` / `he_uniform` sample `mx.random.uniform(-limit, limit, ...)` where `limit = gain * math.sqrt(6.0 / (fan_in + fan_out))` (resp. `sqrt(3.0 / fan)`), so that value is the uniform bound, not a standard deviation — `\sigma` was wrong here (for `U[-a, a]`, σ is `a/√3`). Using `\text{limit}` is a nice touch since it matches the code's actual variable name.

This overlaps with your own #3831, which makes the same `\sigma`-is-wrong fix and additionally rewrites the prose to state the `[-a, a]` interval explicitly. Since both edit the same glorot/he docstring lines they'll conflict, so it's worth consolidating into one — I'd lean toward #3831 for the extra interval clarity, ideally keeping this PR's `\text{limit}` symbol. Flagging so one doesn't end up rotting behind the other.

vineethsaivs:
Closing in favor of #3831, which makes the same relabel with clearer wording: it states the `[-limit, limit]` interval explicitly. I have folded this PR's `\text{limit}` symbol into #3831 so nothing here is lost. Thanks for flagging the overlap.

PR #3843, opened 2026-07-13
Use unroll_count(4) for the NAX attention Q@K.T loop

wyanzhao:
## Proposed changes

Change the NAX attention `Q @ K.T` head-dimension loop from full unrolling
to `#pragma clang loop unroll_count(4)`.

> [!NOTE]
> This is a scheduling-only change. The accumulation order and numerical
> results are unchanged.

## Motivation

With full unrolling, the compiler hoists all `TD` K-tile loads to the start
of the block, ahead of the running MMA chain. Unrolling by four instead lets
the compiler interleave K-tile loads with matrix accumulation.

| Configuration | Full unroll | `unroll_count(4)` |
|---|---|---|
| `head_dim == 128` (`TD == 8`) | All K-tile loads are hoisted | Loads can interleave with the MMA chain |
| `head_dim == 64` (`TD == 4`) | Four iterations unrolled | Still fully unrolled; effectively unchanged |

## Performance

M5 Max, 16 Q / 2 KV heads, bf16, causal, `8192 x 8192`; median of 12 runs:

| Head dimension | Before | After | Change |
|---|---:|---:|---:|
| `128` | `5.70 ms` / `48.2 TF` | `5.08 ms` / `54.1 TF` | `+12%` throughput |
| `64` | `2.45 ms` | `2.44 ms` | No material change |

## Scope and compatibility

- Changes one pragma in the existing `wn=1` NAX QK loop.
- Does not change the `head_dim == 256`, `wn=2` D-split path.
- Preserves the exact accumulation order.
- Composes cleanly with the `head_dim == 256` PR; the two changes touch
  different regions of `steel_attention_nax.h` and can merge in either order.

## Validation

- Outputs remain bitwise-identical because the loop-carried accumulation
  order is unchanged.
- Full local `test_fast_sdpa.py` result: **16 tests, 1 platform-dependent
  skip, otherwise passing**.
- A three-way merge with the `head_dim == 256` PR completes without conflict.

## Checklist

- [x] I have read the [CONTRIBUTING](https://github.com/ml-explore/mlx/blob/main/CONTRIBUTING.md) document.
- [x] I have run `pre-commit run --all-files`.
- [x] Existing fast-SDPA tests pass with no numerical changes.
- [x] No documentation changes are required for this compiler-scheduling change.

### DIFF
--- a/mlx/backend/metal/kernels/steel/attn/kernels/steel_attention_nax.h
+++ b/mlx/backend/metal/kernels/steel/attn/kernels/steel_attention_nax.h
@@ -207,7 +207,12 @@ template <
     for (short iq = 0; iq < TQ; iq++) {
       STEEL_PRAGMA_UNROLL
       for (short ik = 0; ik < TK; ik += 2) {
-        STEEL_PRAGMA_UNROLL
+        // Unrolling the head-dim loop by 4 rather than fully lets the
+        // compiler interleave the next K-tile loads with the running mma
+        // chain instead of hoisting all TD loads up front; measures +11%
+        // at head_dim 128 on M5 Max and is a no-op at head_dim 64 (TD == 4,
+        // already a full unroll).
+#pragma clang loop unroll_count(4)
         for (short id = 0; id < TD; id++) {
           NAXTile<T, 1, 1> Qtile;
           NAXTile<T, 2, 1> Ktile;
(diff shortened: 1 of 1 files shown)


angeloskath (approved):
Very nice!

PR #3084, opened 2026-01-29
Add tiled unfolding to fix Metal maxBufferLength failures in conv_transpose

itsmedonttell:
Title: Add tiled unfolding to fix Metal maxBufferLength failures in conv_transpose

## Summary

Fixes transposed convolution failures on Metal when unfolding buffers exceed `maxBufferLength`.

Previously, `conv_transpose1d/2d/3d` with `stride > 1` would allocate full im2col buffers that could exceed Metal's `maxBufferLength`, causing allocation failures. This PR implements tiled unfolding to process the operation in chunks that fit within hardware limits.

**Changes:**
- Add row-offset support to Metal unfold kernels
- Tile explicit GEMM conv unfold + GEMM along `implicit_M` dimension
- Add validation guard for edge case where a single unfold row exceeds max buffer size

## Impact

**Before:** Operations requiring unfold buffers larger than `maxBufferLength` failed with Metal allocation errors

**After:** Same operations succeed by processing in tiles. Example on M4 Max:
- Input: `(1, 5158, 1)`, kernel: `(1, 4096, 1)`, stride: 1024
- Unfold buffer required: ~156 GB (exceeds device's `maxBufferLength` of ~86.6 GB) → tiled into chunks within `maxBufferLength`
- Previously: FAIL with `maxBufferLength` error
- Now: SUCCESS

## Implementation Details

- `explicit_gemm_conv_ND_gpu` now tiles along rows and writes directly into row-window views of output (`mlx/backend/metal/conv.cpp`)
- `explicit_gemm_conv_group_ND_gpu` applies same tiling logic for grouped convolutions (`mlx/backend/metal/conv.cpp`)
- Unfold kernels (`naive_unfold_Nd` / `naive_unfold_transpose_Nd`) accept a row offset parameter for correct global indexing (`mlx/backend/metal/kernels/conv.metal`)
- Tile size is computed from `maxBufferLength / row_bytes`
- Returns error if a single unfold row exceeds the device's `maxBufferLength` (Metal API limit)

## Performance Note

Tiling introduces additional kernel launches and a temporary buffer per tile, which may slightly reduce performance for very large outputs. However, this tradeoff enables operations that would otherwise fail due to `maxBufferLength` constraints.

## Validation

Tested on Apple M4 Max (macOS 26.2, MLX 0.30.5.dev20260129+590b4f1c):
```python
import mlx.core as mx

print("mlx", mx.__version__)
print("device_info", mx.device_info())

[...]

Issue #3082, opened 2026-01-29
[BUG] conv_transpose* on Metal exceeds maxBufferLength due to full im2col buffer allocation


itsmedonttell:
**Describe the bug**
On Apple Silicon (Metal backend), `conv_transpose1d/2d/3d` with `stride > 1` maps `stride -> input_dilation` and routes to the explicit GEMM path, which allocates a full im2col/unfold buffer. For large outputs this exceeds Metal’s `maxBufferLength` and fails with a Metal allocation error. This affects transposed convs (and any use of `conv_general` with `input_dilation > 1`) on Metal.

**To Reproduce**

```python
import mlx.core as mx

print("mlx", mx.__version__)
print("device_info", mx.device_info())

def try_n(n_frames, kernel_size=4096, stride=1024):
    x = mx.ones((1, n_frames, 1), dtype=mx.float32)
    w = mx.ones((1, kernel_size, 1), dtype=mx.float32)
    y = mx.conv_transpose1d(x, w, stride=stride, padding=0)
    mx.eval(y)
    return y.shape

for n in [5150, 5155, 5156, 5157, 5158, 5160]:
    try:
        out = try_n(n)
        print("OK", n, out)
    except Exception as e:
        print("FAIL", n, e)
```

Observed output (M4 Max, mlx 0.30.4):

```
mlx 0.30.4
device_info {'device_name': 'Apple M4 Max', 'max_recommended_working_set_size': 115448725504, 'memory_size': 137438953472, 'architecture': 'applegpu_g16s', 'max_buffer_length': 86586540032, 'resource_limit': 499000}
OK 5150 (1, 5276672, 1)
OK 5155 (1, 5281792, 1)
OK 5156 (1, 5282816, 1)
OK 5157 (1, 5283840, 1)
FAIL 5158 [metal::malloc] Attempting to allocate 86587211776 bytes which is greater than the maximum allowed buffer size of 86586540032 bytes.
FAIL 5160 [metal::malloc] Attempting to allocate 86620766208 bytes which is greater than the maximum allowed buffer size of 86586540032 bytes.
```

**Expected behavior**
The operation should complete successfully when the output size fits in memory, or MLX should provide a clear, MLX-level error before Metal allocation fails (explaining the unfolding buffer size requirement).

**Desktop (please complete the following information):**
 - OS Version: macOS 26.2
 - Version: mlx 0.30.4
[...]

PR #3845, opened 2026-07-14
Fix conv_transpose maxBufferLength failures on Metal via tiled unfold

eyupcanakman:
## Proposed changes

`conv_transpose1d/2d/3d` (and `conv_general` with `input_dilation > 1`) on Metal go through the explicit GEMM path, which allocates one full unfold buffer of `implicit_M` by `implicit_K`. For large outputs that buffer exceeds the device `maxBufferLength` and the allocation throws, even when the output itself fits in memory (#3082).

The explicit GEMM now unfolds and runs in row tiles that each stay under `maxBufferLength`, reusing a single unfold buffer so peak memory is one tile. Both the grouped and non-grouped paths are covered. The approach follows #3084.

`MLX_CONV_UNFOLD_TILE_ROWS` caps the tile size so the tiled path can be exercised on small inputs, which the added test uses.

## Checklist

### DIFF
--- a/mlx/backend/metal/conv.cpp
+++ b/mlx/backend/metal/conv.cpp
@@ -30,6 +30,23 @@ ensure_row_contiguous(const array& x, metal::Device& d, const Stream& s) {
   return result;
 }
 
+inline int max_unfold_rows(metal::Device& d, size_t row_bytes, int total_rows) {
+  size_t max_buffer = d.mtl_device()->maxBufferLength();
+  size_t max_rows = row_bytes == 0 ? total_rows : max_buffer / row_bytes;
+  // Force a smaller tile (tests, or to bound the unfold buffer further).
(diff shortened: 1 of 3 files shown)


mlx/backend/metal/conv.cpp
```diff
@@ -82,23 +75,73 @@ void explicit_gemm_conv_ND_gpu(
...
-      /*K = */ implicit_K,
-      /*batch_size_out = */ 1,
-      /*a_cols = */ implicit_K,
-      /*b_cols = */ implicit_K,
-      /*a_transposed = */ false,
-      /*b_transposed = */ true,
-      /*copies = */ copies);
+  // 2D view of the output; each tile writes a row window of it.
+  Strides out_2d_strides{
+      out.strides()[out.ndim() - 2], out.strides()[out.ndim() - 1]};
```
zcbenz: You can use `out.strides(-2)`

zcbenz (approved):
Nice fix!

PR #3848, opened 2026-07-16
Fix installed static MLX package on Windows

yingjiacai:
## Proposed changes

Fix the installed static MLX package on Windows so downstream CMake
projects can link against `mlx.lib` without requiring `mlx.dll`.

The changes:

- Compile the dlfcn-win32 implementation directly into the static MLX
  library instead of exposing a FetchContent-only target to installed
  consumers.
- Link the required Windows `Psapi.lib` system library.
- Represent the downloaded OpenBLAS import library as an imported CMake
  target and install the `.lib` file.
- Keep `cudnn_frontend` as a build-only dependency.
- Reconstruct the required CUDA, cuDNN, and OpenBLAS dependencies in the
  installed MLX CMake package.

## Testing

Manually verified on Windows using MSVC, Ninja, CUDA 13.0, and cuDNN 9.19:

- Configured and built MLX with `BUILD_SHARED_LIBS=OFF`
- Installed MLX successfully
- Verified that `mlx.lib` is installed without requiring `mlx.dll`
- Rebuilt and ran the downstream `mlx_cpp_inference` project against the
  installed static MLX package
- Ran `pre-commit run --all-files` successfully

## Checklist

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -261,7 +261,15 @@ if(WIN32)
   FetchContent_MakeAvailable(dlfcn-win32)
   endblock()
   target_include_directories(mlx PRIVATE "${dlfcn-win32_SOURCE_DIR}/src")
-  target_link_libraries(mlx PRIVATE dl)
+  if(BUILD_SHARED_LIBS)
+    target_link_libraries(mlx PRIVATE dl)
+  else()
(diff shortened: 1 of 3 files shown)


zcbenz (approved):
This is a very nice improvement, thanks for making the static build work.

PR #3855, opened 2026-07-16
Add metadata to exported functions

### LINKED ISSUE #2410: [Feature requires] Support saving meta data to mlxfn files and reading them back

eyupcanakman:
Closes #2410. Exported `.mlxfn` files can now store a metadata string, so a model's config can live with the function instead of in a separate JSON file.

`export_function` and `exporter` take an optional `metadata` string and `import_function(file, return_metadata=True)` returns the function along with it. Structured config can go into the string as JSON, which is what the docs example does. Metadata is only supported when exporting to a file. Passing it with a callback raises.

String fields in the header are now written in one bulk write instead of a write per byte. Without that, a 1 MB metadata string took seconds per trace in a multi-trace export.

The metadata lives in the file header, so a `.mlxfn` written before this change has no metadata section and won't load in a metadata-aware build. That matches the existing note that exported functions aren't compatible across MLX versions.

### DIFF
--- a/docs/src/usage/export.rst
+++ b/docs/src/usage/export.rst
@@ -109,6 +109,33 @@ keyword arguments when calling the imported function.
   out, = imported_fun(x, z=y)
 
 
+Saving Metadata
+---------------
+
+You can save metadata, such as a model configuration, alongside an exported
+function. The metadata is a dictionary of string keys with string, integer, or
+float values:
+
+.. code-block:: python
+
+  def fun(x, y):
+    return x + y
+
+  x = mx.array(1.0)
+  y = mx.array(1.0)
+  metadata = {"description": "adds two arrays", "version": 1, "scale": 0.5}
+  mx.export_function("add.mlxfn", fun, x, y, metadata=metadata)
+
+Pass ``return_metadata=True`` to read the metadata back when importing:
+
+.. code-block:: python
+
+  imported_fun, metadata = mx.import_function("add.mlxfn", return_metadata=True)
+
+  # Prints: adds two arrays
+  print(metadata["description"])
+
+
 Exporting Modules
 -----------------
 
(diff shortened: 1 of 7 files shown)


zcbenz (commented):
I think we can just make the metadata a plain string, it is trivial in high level languages to encode arbitrary data into JSON string. Serializing a dictionary of rich types feels like something that would bite us later and does not provide much more convenience.

eyupcanakman:
Changed it to a plain string and rebased.

zcbenz (approved):
Looks good to me, thanks!

csukuangfj:
Thanks!

PR #3867, opened 2026-07-17
Skip non-quantizable modules in nn.quantize with a custom predicate

### LINKED ISSUE #3865: [BUG] nn.quantize: a class_predicate that selects a module without `to_quantized` makes the whole call fatal — apply the documented "defines to_quantized" floor to explicit predicates too
### Affected project / version

`mlx` 0.31.2 (Apple Silicon). Relevant code: `mlx.nn.quantize`
(`mlx/nn/layers/quantized.py`).

### Minimal reproduction (self-contained)

```python
import mlx.core as mx
import mlx.nn as nn

class MoEGate(nn.Module):            # router-style module: has params,
    def __init__(self):              # is not quantizable, no to_quantized
        super().__init__()
        self.weight = mx.zeros((8, 64))

class Block(nn.Module):
    def __init__(self):
        super().__init__()
        self.proj = nn.Linear(64, 64)
        self.gate = MoEGate()

# Default predicate: fine — gate is skipped, proj becomes QuantizedLinear.
nn.quantize(Block(), group_size=64, bits=4)

# Explicit predicate that (correctly for proj, incidentally for gate)
# returns True: the WHOLE call dies.
nn.quantize(Block(), group_size=64, bits=4, class_predicate=lambda p, m: True)
# ValueError: Unable to quantize model of type <class '__main__.MoEGate'>
```

(Verified on mlx 0.31.2. On older releases this surfaced as
`AttributeError: '<cls>' object has no attribute 'to_quantized'`.)

### Actual behavior

With `class_predicate=None` the docstring's rule applies: "all layers that
define a `to_quantized()` method will be quantized" — non-quantizable modules
are skipped. But the moment a caller passes an explicit predicate, that floor
disappears: a truthy result for a module without `to_quantized` raises and
aborts the entire call (typically a whole model load).

### Expected behavior

A caller-provided predicate should narrow the set of otherwise quantizable
modules; it should not make an intrinsically non-quantizable module eligible.
A truthy predicate on a module without `to_quantized` should skip that module,
consistent with the documented default rule.

### Why existing public APIs are insufficient / ecosystem evidence

Every careful call site currently reimplements the identical guard:

- `mlx_lm.utils.load_model`'s config predicate starts with
  `if not hasattr(m, "to_quantized"): return False`;
- `mlx_lm.utils.quantize_model`'s `wrapped_predicate` does the same;
- downstream loaders for custom bundle formats do the same in several places —
  and the ones that miss it (predicates built from on-disk key suffixes that
  fall through to `return True`) turn a config/suffix mismatch into a fatal
  load error. Some embedding applications wrap `nn.quantize` process-wide just
  to AND this floor onto every predicate.

When the same guard is hand-rolled at N call sites and its absence is fatal,
it belongs in the function.

### Smallest proposed change (proposal, not a requirement)

Compose the documented floor with the caller's predicate:

```python
effective_predicate = (
    lambda path, module:
        hasattr(module, "to_quantized")
        and class_predicate(path, module)
)
```

If you consider the explicit failure intentional (predicate-as-config-
validation), preserving it in some form is of course your call — but the
default experience of "predicate selected one extra module ⇒ entire model load
aborts" seems unlikely to be what callers want.

### Backward compatibility

The only behavioral delta is fatal-crash → skip for modules that could never
be quantized; standard layers all define `to_quantized`, so no existing
successful call changes.

### Tests upstream could add

- The repro above with the explicit predicate: `proj` quantized, `gate`
  untouched, no exception.
- A predicate returning a params dict for a non-quantizable module: skipped.

### What this replaces downstream

Our launcher's process-wide `nn.quantize` wrapper (whose entire purpose is
injecting this floor), carried since early 2026 to keep MoE-gate-bearing
hybrid bundles loadable behind third-party predicates. Happy to PR.

### Prior art / existing issues (searched)

None found on ml-explore/mlx covering predicate-selected non-quantizable
modules; the mlx-lm-side guards cited above are the visible prior art.

shubhxho:
Fixes #3865.

With the default predicate, `nn.quantize` skips modules that don't define `to_quantized()`. Passing a custom `class_predicate` loses that behavior: if it returns truthy for such a module, the call raises `Unable to quantize model of type ...` and the whole model load fails.

```python
class Gate(nn.Module):        # has params, no to_quantized()
    def __init__(self):
        super().__init__()
        self.weight = mx.zeros((8, 64))

class Block(nn.Module):
    def __init__(self):
        super().__init__()
        self.proj = nn.Linear(64, 64)
        self.gate = Gate()

nn.quantize(Block(), group_size=64, bits=4)                                  # ok
nn.quantize(Block(), group_size=64, bits=4, class_predicate=lambda p, m: True)  # raises
```

This is why call sites like `mlx_lm` start every predicate with `if not hasattr(m, "to_quantized"): return False`.

The fix checks for `to_quantized()` before calling the predicate, so a module without it is skipped whether the predicate is custom or the default. Standard layers all define `to_quantized()`, so existing calls are unaffected. Updated the docstring and added test coverage for a custom predicate returning `True` and a params dict over a non-quantizable module.

### DIFF
--- a/python/mlx/nn/layers/quantized.py
+++ b/python/mlx/nn/layers/quantized.py
@@ -50,8 +50,13 @@ def quantize(
         class_predicate (Optional[Callable]): A callable which receives the
            :obj:`Module` path and :obj:`Module` itself and returns ``True`` or a
            dict of params for ``to_quantized`` if it should be quantized and
-           ``False`` otherwise. If ``None``, then all layers that define a
-           ``to_quantized()`` method are quantized. Default: ``None``.
+           ``False`` otherwise. A module selected by the predicate but without a
+           ``to_quantized()`` method (for example a normalization or activation
+           layer) is skipped rather than quantized, so a broad predicate can be
+           applied to a whole model. If the predicate selects *only* such
+           modules and nothing is quantized, a :obj:`ValueError` is raised, since
+           that request cannot be satisfied. If ``None``, then all layers that
+           define a ``to_quantized()`` method are quantized. Default: ``None``.
 
     Example:
         Weight only quantization for all layers that define a ``to_quantized()`` method:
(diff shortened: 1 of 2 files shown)


Pablosinyores (commented):
Reproduced the reported failure on current main. Because `quantize` walks `leaf_modules()`, any leaf that does not define `to_quantized` reaches the predicate, so a broad predicate hits the raise on ordinary layers:

```python
class Block(nn.Module):
    def __init__(self):
        super().__init__()
        self.proj = nn.Linear(64, 64)
        self.norm = nn.LayerNorm(64)

nn.quantize(Block(), group_size=64, bits=4, class_predicate=lambda p, m: True)
# ValueError: Unable to quantize model of type <class 'mlx.nn.layers.normalization.LayerNorm'>
```

`leaf_modules()` returns `['proj', 'norm']`, so this is not an exotic case, any model with a norm or activation layer trips it. Skipping those rather than raising looks like the right direction to me, and doing the `hasattr` check before calling the predicate reads cleaner than the nested version it replaces. `black` is clean on both files.

One thing I would want a maintainer to weigh explicitly: the change removes the `Unable to quantize model of type ...` error entirely, so a predicate that *deliberately* selects a type that cannot be quantized now silently does nothing. `class_predicate=lambda p, m: isinstance(m, nn.LayerNorm)` used to say why nothing happened; after this it returns a model that is simply unquantized, with no signal.

Skipping incidental leaves and staying silent when the user asked for something impossible are two different things, and this collapses them into one behavior. If you want to keep a signal without bringing back the old over-eager raise, one option is to raise only when the predicate returned true for at least one module and nothing was quantized at all. Happy either way, but it seems worth deciding rather than dropping the diagnostic as a side effect.

Also worth updating the docstring line about the default: with `class_predicate` now defaulting to `lambda _, m: True`, the "all layers that define a `to_quantized()` method are quantized" sentence is still accurate, but it now describes the `hasattr` guard rather than the default predicate. The added sentence covers it, just noting the two are doing different jobs now.

nastya236:
Thank you for flagging this! Will look at it today

nastya236:
Thanks for looking into this and for the detailed issue.

That said, I think the current behaviour is intended:

- If `class_predicate` is not provided -- quantize every module that defines `to_quantized()`, and skip the rest.
- If `class_predicate` is provided -- quantize every module the predicate selects (`True` or a params dict). If a selected module does not define `to_quantized()`, we raise. This is deliberate: it guarantees that anything explicitly selected for quantization actually gets quantized, rather than being silently skipped.

In other words, `class_predicate` is an explicit selection, not a filter over "quantizable" modules. Silently skipping a selected by a predicate but unquantizable module would hide real mistakes, for example a custom layer meant to be quantizable but missing `to_quantized()`.

If you want to quantize a custom module, the intended way is to define a `to_quantized()` method. And if you specifically want a broad predicate to skip non-quantizable modules, you can compose that yourself:

`nn.quantize(model, class_predicate=lambda p, m: hasattr(m, "to_quantized") and my_predicate(p, m))`

Given that, I don't think we should change this behaviour, so I'll close the PR.

Issue #3346, opened 2026-03-31
Kernel panic: "completeMemory() prepare count underflow" in IOGPUMemory.cpp on macOS 26.4 M3 Ultra during MLX inference


yoyaku155:
## Summary

Running MLX inference on Mac Studio M3 Ultra (96GB) with macOS 26.4 (25E246) causes **repeated kernel panics** in Apple's IOGPU kernel extension. We have observed **9 kernel panics in 6 days**, with 4 occurring in a single 2-hour window.

This appears to be a bug in Apple's \`IOGPU.kext\` GPU memory management, not in MLX itself, but MLX's Metal usage pattern triggers it reliably.

## System

- **Hardware**: Mac Studio, Apple M3 Ultra, 96 GB Unified Memory
- **OS**: macOS 26.4 (Build 25E246)
- **Kernel**: Darwin 25.4.0 (xnu-12377.101.15~1/RELEASE_ARM64_T6031)
- **MLX**: Latest via pip (mlx, mlx-lm)
- **Model**: Qwen3.5-122B-A10B-Vision-MLX-Mixed-4bit (~50GB VRAM)

## Panic Signatures

Two distinct bugs in IOGPU.kext:

### Bug 1: IOGPUMemory.cpp:550 (7 occurrences, March 31 - April 1)
\`\`\`
panic(cpu N caller 0xXXXX): "completeMemory() prepare count underflow" @IOGPUMemory.cpp:550
\`\`\`
Reference count underflow in \`completeMemory()\` — a double-free or premature release of GPU memory objects.

| Date/Time (JST) | CPU |
|---|---|
| 2026-03-31 09:43 | cpu 23 |
| 2026-03-31 10:17 | cpu 25 |
| 2026-03-31 15:10 | cpu 21 |
| 2026-04-01 00:02 | cpu 9 |
| 2026-04-01 00:24 | cpu 7 |
| 2026-04-01 01:32 | cpu 26 |
| 2026-04-01 01:42 | cpu 7 |

### Bug 2: IOGPUGroupMemory.cpp:219 (2 occurrences, March 26)
\`\`\`
panic(cpu N caller 0xXXXX): "Memory object unexpectedly not found in fPendingMemorySet" @IOGPUGroupMemory.cpp:219
\`\`\`
Race condition or tracking inconsistency in group memory management.

| Date/Time (JST) | CPU |
|---|---|
| 2026-03-26 13:27 | cpu 11 |
[...]

PR #3868, opened 2026-07-17
Don't reuse a Metal buffer while a command buffer is still using it

shubhxho:
Under continuous batching with `async_eval`, the small int32 KV-cache bookkeeping arrays (`offset` / `left_padding`) sometimes come back as garbage that looks like bf16 bits in int32 slots -- e.g. `16256` (`0x3F80`, bf16 `1.0`). A buffer is being reused while a kernel is still writing it. Full write-up and repro in the linked issue; likely related to #3346.

`MetalAllocator::free` returns a buffer to the reuse pool without knowing which command buffer last used it. Fine on one stream (in order), but `async_eval` adds a second stream, so a buffer freed on one can be reused on the other before the first kernel finishes.

Fix: count how many in-flight command buffers reference each buffer -- marked at `end_encoding`, retired from the completion handler. `free()` holds a still-in-use buffer aside and recycles it once the work is done.

### DIFF
--- a/mlx/backend/metal/allocator.cpp
+++ b/mlx/backend/metal/allocator.cpp
@@ -187,6 +187,14 @@ void MetalAllocator::free(Buffer buffer) {
   }
   std::unique_lock lk(mutex_);
   active_memory_ -= buf->length();
+  // Don't hand this buffer back out while a command buffer we haven't seen
+  // finish still uses it, or a concurrent (or cross-stream) allocation could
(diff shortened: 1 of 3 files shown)


zcbenz (changes requested):
A buffer still being used by a command buffer would never be released, usually this kind of crash was caused by custom extensions, and and please reference https://github.com/ml-explore/mlx/pull/3462 on how you could find out the root cause on your side.

PR #3869, opened 2026-07-18
[Metal] Avoid regex in custom kernel name generation

aleroot:
## Proposed changes

Templated custom Metal kernels generate a specialized kernel name on every invocation. This previously constructed a `std::regex` and called `std::regex_replace` to convert the template definition into a valid kernel-name suffix.

This cost becomes measurable in workloads that repeatedly invoke templated custom kernels. In an Instruments capture of Qwen 3.5 gated-delta decoding, regex processing accounted for approximately 3.5% of the generation worker's sampled CPU cycles.

This change replaces the regex with a single linear pass over the template definition while preserving the generated kernel names byte-for-byte. It also removes the unused `<regex>` dependency.

### Performance

An isolated C++20 `-O3` benchmark using a representative template definition:

`<bfloat16_t, float, 128, 64, 8, 8>`

Results over 20,000 iterations:

- Previous regex implementation: approximately 6.86 µs per call
- New linear implementation: approximately 0.077 µs per call
- Isolated speedup: approximately 88.6×  

This benchmark measures only template-name transformation and does not imply an equivalent end-to-end inference speedup.

## Checklist

- [x] I have updated the necessary documentation (not needed: this is an internal performance change with no API or behavior change)

### DIFF
--- a/mlx/backend/common/metal_kernel.cpp
+++ b/mlx/backend/common/metal_kernel.cpp
@@ -1,7 +1,6 @@
 // Copyright © 2024 Apple Inc.
 
 #include <iostream>
-#include <regex>
 #include <sstream>
 
 #include "mlx/backend/common/compiled.h"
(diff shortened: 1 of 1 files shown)


angeloskath (commented):
This looks great! Do you mind moving it out to a simple helper function? I would also reserve `template_def.size()` instead of 1 less cause maybe there is no `, ` in the definition.

aleroot:
> This looks great! Do you mind moving it out to a simple helper function? I would also reserve `template_def.size()` instead of 1 less cause maybe there is no `, ` in the definition.

Done, look if is ok. Thanks.

angeloskath (approved):
Very nice!

PR #3872, opened 2026-07-19
Zero-copy CPU import: mx.array(host_buffer, copy=False) on unified memory

HaoXuAI:
## Motivation
When I tried loading Parquet/Arrow columns into MLX on Apple Silicon: `mx.array(host_buffer)` always copies the host bytes into MLX-owned memory. But on unified memory the CPU and GPU share the same DRAM, so that copy is avoidable — the GPU can address the existing bytes. Pure overhead for large resident columns (feature tables, embeddings).

## What
A keyword-only `copy` on the `array` constructor. `copy=False` adopts a CPU host buffer (via Metal `newBufferWithBytesNoCopy`) instead of copying it.

```python
a = np.random.rand(64_000_000).astype(np.float32)   # 256 MB

mx.array(a)              # copy  -> 4.2 ms
mx.array(a, copy=False)  # adopt -> 0.02 ms   (no data movement; shares the bytes)
```
(M4 Max. Default is `copy=None` — unchanged behavior.)

## How
In the CPU branch of `nd_array_to_mlx`, wrap the host pointer with `make_buffer` and `set_data`, using a deleter that keeps the source alive and releases the wrapper via `allocator::release`. It raises (rather than silently copying) when Metal is unavailable, the source element size != dtype size, or the buffer can't be adopted.

## Validation
All tests pass on a local Metal build (M4 Max): values, zero-copy sharing, dtype-mismatch raise, source-lifetime, and adopt-in-a-loop. Fork Linux CI green (lint / Fedora / ASAN / UBSAN); the macOS/Metal matrix only runs on the upstream repo.

## Review questions
- CPU adopt is Metal-build-only by design — acceptable, or would you prefer a different fallback than raising?

### DIFF
--- a/python/src/array.cpp
+++ b/python/src/array.cpp
@@ -306,18 +306,13 @@ void init_array(nb::module_& m) {
       nb::pooled(/* capacity = */ 128))
       .def(
           "__init__",
-          [](mx::array* aptr,
-             nb::object v,
-             std::optional<mx::Dtype> t,
-             std::optional<bool> copy) {
-            new (aptr) mx::array(create_array(v, t, copy));
+          [](mx::array* aptr, nb::object v, std::optional<mx::Dtype> t) {
+            new (aptr) mx::array(create_array(v, t, true));
           },
           "val"_a,
           "dtype"_a = nb::none(),
-          nb::kw_only(),
-          "copy"_a = nb::none(),
           nb::sig(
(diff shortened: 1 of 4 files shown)


angeloskath (approved):
That's great!

I refactored it a bit, the main point being to remove the copy argument from `mx.array` following also the discussion in #3531.  We can use the no-copy cast via `mx.asarray` or `mx.from_dlpack` and `mx.array` always copies. If there is need we can add the copy in the array constructor later.

PR #3878, opened 2026-07-21
Fix integer keys collision in tree_unflatten

### LINKED ISSUE #3877: tree_unflatten silently misplaces values when integer-like keys collide (e.g. "01" vs "1")
## Describe the bug

`mlx.utils.tree_unflatten` silently misplaces values when two keys parse to the same integer index, for example `"01"` and `"1"`. Instead of raising or handling the collision, every value from the collision onward gets shifted one slot and a phantom `{}` placeholder appears in the list. No error is raised, so the corruption is easy to miss.

## To Reproduce

```python
from mlx.utils import tree_unflatten

print(tree_unflatten([("01", "a"), ("1", "b"), ("2", "c")]))
# observed: [{}, 'a', 'b', 'c']
```

Observed: `[{}, 'a', 'b', 'c']`. The value from key `"1"` lands at index 2, the value from key `"2"` lands at index 3, and index 0 is a fabricated empty dict.

`tree_flatten` itself always emits canonical `str(i)` keys, so a normal flatten/unflatten round trip is fine. This bites when the flat dict comes from somewhere else, e.g. externally produced checkpoints or hand-edited key paths that use zero-padded indices.

## Expected behavior

Either a `ValueError` on colliding indices, or keys treated consistently so nothing silently shifts. Happy to put up a small PR with a fix plus a regression test if you have a preference on which behavior it should be (my instinct is raising on collision, since silently guessing seems worse).

## Root cause

`python/mlx/utils.py:234` sorts `(int(idx), idx)` pairs and the padding loop at lines 236-239 uses `range(i - len(l))`, which assumes the string to int mapping is injective. When two keys map to the same `i`, the second `append` just lands one slot further right with no check.

## Desktop

- OS: macOS 26.5 (M4)
- Version: 0.32.1.dev source build at ce307338 (CPU)

---

quick honest note: i found this while stress testing edge cases with help from Claude Code, then verified the repro and read through the source myself. i'm a freshman still learning my way around, so if this is actually intended behavior i'd love to understand why :)

Solaris-star:
## Proposed changes

`tree_unflatten` treated all integer-like keys as list indices via `int(idx)`. Keys such as `"01"` and `"1"` both parse as `1`, which silently shifted later values and inserted a phantom `{}` (issue #3877).

This PR rejects non-canonical integer keys (`str(i) != k`) and duplicate indices by raising `ValueError`, so the existing `except ValueError` path falls back to a dict tree and preserves values.

Fixes #3877

## Checklist

### DIFF
--- a/python/mlx/utils.py
+++ b/python/mlx/utils.py
@@ -232,6 +232,17 @@ def tree_unflatten(tree: Union[List[Tuple[str, Any]], Dict[str, Any]]) -> Any:
     # Assume they are a list and fail to dict if the keys are not all integers
     try:
         keys = sorted((int(idx), idx) for idx in children.keys())
+        # Reject integer-like keys that are not the canonical str(i) form, e.g.
+        # "01" and "1" both parse as 1 and would silently shift later values.
+        seen_indices = set()
+        for i, k in keys:
+            if str(i) != k:
+                raise ValueError(
+                    f"Non-canonical integer key {k!r} collides with index {i}"
+                )
+            if i in seen_indices:
+                raise ValueError(f"Duplicate integer key for index {i}")
+            seen_indices.add(i)
         l = []
         for i, k in keys:
             # if i <= len(l), no {} will be appended.
--- a/python/mlx/utils.py
+++ b/python/mlx/utils.py
@@ -229,20 +229,16 @@ def tree_unflatten(tree: Union[List[Tuple[str, Any]], Dict[str, Any]]) -> Any:
         next_idx = "" if not next_idx else next_idx[0]
         children[current_idx].append((next_idx, value))
 
-    # Assume they are a list and fail to dict if the keys are not all integers
+    # Assume they are a list and fail to dict if the keys are not all integers.
+    # Non-canonical integer-like keys (e.g. "01") also fall back to a dict:
+    # "01" and "1" both parse as 1 and would otherwise silently shift later
+    # list slots. Bare ValueError is intentional so the existing except path
+    # keeps the fall-back-to-dict contract for malformed external keys.
     try:
         keys = sorted((int(idx), idx) for idx in children.keys())
-        # Reject integer-like keys that are not the canonical str(i) form, e.g.
-        # "01" and "1" both parse as 1 and would silently shift later values.
-        seen_indices = set()
         for i, k in keys:
             if str(i) != k:
-                raise ValueError(
-                    f"Non-canonical integer key {k!r} collides with index {i}"
-                )
-            if i in seen_indices:
-                raise ValueError(f"Duplicate integer key for index {i}")
-            seen_indices.add(i)
+                raise ValueError
         l = []
         for i, k in keys:
             # if i <= len(l), no {} will be appended.
--- a/python/mlx/utils.py
+++ b/python/mlx/utils.py
@@ -229,22 +229,25 @@ def tree_unflatten(tree: Union[List[Tuple[str, Any]], Dict[str, Any]]) -> Any:
         next_idx = "" if not next_idx else next_idx[0]
         children[current_idx].append((next_idx, value))
 
-    # Build a list only when every key is a canonical, non-negative index.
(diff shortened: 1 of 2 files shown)


0lekz (approved):
LGTM. Change makes tree_unflatten reject ambiguous keys instead of silently producing a corrupted list, and the regression test covers both the edge case and the canonical behavior. I don't see any compatibility concerns since tree_flatten already emits canonical integer keys

Solaris-star:
Thanks @0lekz for the review and approval — much appreciated.

Happy to address any further feedback. If this looks good from a maintainer perspective, a merge when convenient would be great.

Pablosinyores (commented):
Confirmed the bug and the fix behaves as described. Checked out the branch and ran `test_tree`, all 5 pass, and the collision case now falls back to a dict instead of the shifted `[{}, 'a', 'b', 'c']`.

Three things worth looking at before this goes in.

**`black` will fail the lint job.** CI has not been approved to run on this fork PR yet, so it is not visible yet, but `black --check python/tests/test_tree.py` reformats it:

```
-
     def test_tree_unflatten_integer_key_collision(self):
...
+
 if __name__ == "__main__":
```

One blank line too many before the new method and one too few before `if __name__`. Running `black python/tests/test_tree.py` fixes both.

**The two `ValueError` messages can never reach a user.** They are raised inside the `try` whose `except ValueError` is the existing fall-back-to-dict path, so they are immediately swallowed and the caller just gets a dict. That is a reasonable outcome, and it matches how the function already treats non-integer keys, but it means the carefully worded "collides with index" and "Duplicate integer key" text is unreachable. Either drop the messages and use a bare `raise ValueError` with a comment saying the fall-back is deliberate, or, if you want the user to actually see the diagnosis, raise outside the `try`.

Worth flagging that the issue you filed leaned toward raising rather than falling back. Those are genuinely different contracts for someone loading an externally produced checkpoint: raising tells them their keys are malformed, falling back hands them a dict where they expected a list and lets it fail somewhere further away. That is a maintainer call, but it would be good to have it made explicitly rather than inherited from where the `raise` happens to sit.

**The duplicate-index check looks unreachable.** `children` is a dict, so its keys are already distinct strings. If two distinct keys map to the same `i`, at least one of them is not `str(i)`, so the canonical check raises first. If both satisfy `str(i) == k` they are the same string and cannot both be present. So `seen_indices` never fires and could come out, which would make the intent of the remaining check clearer.

Nice find on the root cause, the `range(i - len(l))` analysis in the issue is exactly right.

Solaris-star:
Thanks @Pablosinyores — addressed all three points in the latest commit:

1. **black**: reformatted `python/tests/test_tree.py` (blank-line spacing around the new method / `if __name__`)
2. **unreachable messages**: dropped the carefully-worded `ValueError` text and use a bare `raise ValueError` so the existing `except ValueError` fall-back-to-dict path stays the deliberate contract for malformed external keys (aligned with how non-integer keys already behave)
3. **`seen_indices`**: removed — as you noted it is unreachable once the canonical `str(i) == k` check is in place

Happy to switch to an explicit raise-outside-try contract if maintainers prefer that over fall-back-to-dict.

zcbenz (approved):
A decent fix should refactor the code like below rather than relying on exceptions being thrown.

```
if all_keys_are_integer:
  return list
else:
  return dict

**Review by zcbenz — CHANGES_REQUESTED** [review]
A decent fix should refactor the code like below rather than relying on exceptions being thrown.

```
if all_keys_are_integer:
  return list
else:
  return dict
```

Solaris-star:
Addressed the requested control-flow refactor in signed commit `5093c82`. `tree_unflatten` now classifies all keys up front and only builds a list when every key is a canonical non-negative integer; otherwise it directly returns a dict, without using exceptions for branching. Verification: all 5 `TestTreeUtils` tests pass against the local `utils.py`, Black passes for both touched files, and `git diff --check` passes.

zcbenz (approved):

PR #3881, opened 2026-07-21
[CUDA][Improvement] RMSNorm backward 

nastya236:
Currently `RMSNorm::vjp` kernel consists of 2 kernels:
- compute `dx` and `dw` per row -> (M, N), (M, N)
- column reduction over (M, N) to sum gradients over rows (batch axis) for `dw`

Since rmsnorm is memory-bandwidth bounded, in theory we should benefit a lot from making dw intermediates smaller (especially since batch dim can be up M=32768 or even M=65536) to decrease the traffic. 

Therefore to make backward faster, we:
1) make kernel persistent: we launch as many CTAs as number of SMs * x (where x is heuristic, I set it to 2). Each block covers 1 tile = 1 row. To hide loads latency we pipeline using shared memory double buffering, overlapping row reduction with `cp.async` of the next row. dw partials now has a size (num_blocks, N) instead of (M, N), where num_blocks = 256 for h100/b200. 
2) also decrease block size to 128 instead of 1024 -> increase work per thread  (similar optimization to #3850)

```python 
M = 8192 * 2  # seq_len * batch size
N = 4096 # model dim
x = mx.random.uniform(shape=(M, N)).astype(mx.bfloat16)
w = mx.random.uniform(shape=(N,)).astype(mx.bfloat16)
g = mx.random.uniform(shape=(M, N)).astype(mx.bfloat16)
mx.eval(x, w, g)

def vjp_fn(x, w, g):
    return mx.vjp(lambda a, b: [mx.fast.rms_norm(a, b, 1e-5)], [x, w], [g])[1]

time_fn(vjp_fn, x, w, g, msg="rms_norm vjp (dx, dw)")
```

Before:  0.59904 msec
After: 0.32864 msec

For a chain rms_norm + linear forward+backward for M = 8192 * 2, N = 4096:
This PR backward + forward (#3850) results in 2.90942 while before (latest release) 3.54703. 

Highly inspired by https://github.com/Dao-AILab/quack.


zcbenz (approved):
Very impressive improvement!

PR #3882, opened 2026-07-21
Reuse Metal WAR tracking hash tables

neilmehta24:
## Proposed changes

Metal’s write-after-read tracking moves `next_inputs_` and `next_outputs_` into the previous sets at every barrier. This discards the old hash tables, so the next sets must rebuild their bucket lists during the following barrier epoch. This change was introduced in this PR https://github.com/ml-explore/mlx/pull/3630.

Qwen 35B hits this path 1,056 times per decoded token. The input side alone caused 1,859 bucket-table growths per token, while output tracking followed the same allocation pattern.

This change swaps both pairs of sets instead. The existing `clear()` calls remove old entries while keeping all four bucket tables for reuse. Synchronization behavior and tracked resources are unchanged.

Controlled Qwen runs with input-table reuse improved throughput by about 1.9%, with identical output and unchanged peak memory. MLX’s RoPE benchmark also showed no regression. The output-table reuse applies the same optimization but was not benchmarked separately.

## Checklist

### DIFF
--- a/mlx/backend/metal/device.cpp
+++ b/mlx/backend/metal/device.cpp
@@ -392,9 +392,9 @@ void CommandEncoder::maybeInsertBarrier() {
   if (needs_barrier_) {
     get_command_encoder()->memoryBarrier(MTL::BarrierScopeBuffers);
     needs_barrier_ = false;
-    // Preserve both hash tables' buckets for reuse across barrier epochs.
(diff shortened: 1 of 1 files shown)


angeloskath (approved):
Interesting if that makes such a difference. Can you please also provide a benchmark to showcase the 2% eg running mlx_lm.benchmark or something similar.

Issue #3860, opened 2026-07-17
[BUG] fp32 matmul silently defaults to TF32-class precision (`MLX_ENABLE_TF32=1`), undocumented on both backends


stoyoda0012-cyber:
## Describe the bug

On the CUDA backend, fp32 `mx.matmul` runs in TF32 (10-bit mantissa)
by default: `enable_tf32()` in `mlx/utils.h` defaults `MLX_ENABLE_TF32`
to `1`, and `dtype_to_compute_type()` in
`mlx/backend/cuda/gemms/cublas_gemm.cpp` then selects
`CUBLAS_COMPUTE_32F_FAST_TF32` for `float32` (and `complex64`) inputs.

The switch exists, but nothing tells the user about it: there is no
warning, and `MLX_ENABLE_TF32` appears nowhere in the documentation —
the only occurrences in the repo are `mlx/utils.h` and
`python/tests/mlx_tests.py`. We spent a full-pipeline bisection tracing
a ~9 dB accuracy regression before finding it, and initially worked
around it with NVIDIA's driver-level `NVIDIA_TF32_OVERRIDE=0` because
we could not find any MLX-side control.

Relative Frobenius error of a 512×512 fp32 GEMM against a float64
reference:

| backend | rel. error |
|---|---|
| NumPy fp32 | 2.9e-07 |
| MLX CPU stream | 4.1e-07 |
| MLX Metal (M-series) | fp32-class |
| **MLX CUDA (default, `MLX_ENABLE_TF32=1`)** | **2.9e-04** |
| MLX CUDA + `MLX_ENABLE_TF32=0` | 2.1e-07 |
| MLX CUDA + `NVIDIA_TF32_OVERRIDE=0` (driver-level) | 2.1e-07 |

~1000× the error of fp32 — the expected TF32 mantissa signature.
`MLX_ENABLE_TF32=0` is measured equivalent to the driver-level
override (2.08e-07 vs 2.077e-07 on the same seed). Because the flag is
read lazily on first use, `os.environ["MLX_ENABLE_TF32"] = "0"` before
the first kernel call also works in-process (measured: 2.08e-07) —
useful, but only if the user knows the flag exists.

The practical problem is the **silent divergence from Metal**: the same
MLX program produces fp32-accurate results on Apple silicon and
TF32-accurate results on CUDA. In our fitting engine this flipped
near-tie `argmax` decisions over correlation scores (1.4–2.5 % of
[...]

PR #3883, opened 2026-07-21
Warn once when float32 ops silently run at TF32 precision

pierre427:
Addresses the diagnostic half of #3860.

float32 matmul-family ops default to TF32-class reduced precision (`MLX_ENABLE_TF32` defaults to 1) on both backends on current hardware — CUDA tensor cores and the M5 Metal NAX route — and the default is undocumented. #3860 has two independent downstream reports of the resulting failure mode: op-level tests stay green while downstream results shift (near-tie argmax flips on sm_120; mlx-lm's `test_ssm` state comparison on M5), and the bisection points at hardware/environment rather than any commit. This repo's own test harness already pins `MLX_ENABLE_TF32=0` (`python/tests/mlx_tests.py`) for the same reason.

This adds `env::tf32_active_for_fp32()`: returns exactly `env::enable_tf32()`, but prints a one-line warning to stderr, once per process, and **only when `MLX_ENABLE_TF32` is unset** — anyone who set `=1` or `=0` explicitly has already made the choice and never sees it:

> [mlx] float32 matmul-family ops are running at reduced (TF32) precision, which is the default when MLX_ENABLE_TF32 is unset. Set MLX_ENABLE_TF32=0 before the first operation for full float32 precision, or set MLX_ENABLE_TF32=1 explicitly to silence this warning.

It is called at exactly the points where a float32 op takes the reduced-precision route, by reordering the existing short-circuit `(env::enable_tf32() || dtype != float32)` → `(dtype != float32 || env::tf32_active_for_fp32())`, so processes that never engage TF32 (matvec-only, non-fp32, opted out) never print:

- **Metal:** the NAX gates in steel matmul, gather_mm_rhs, segmented_mm, qmm, gather_qmm, gather_qmm_rhs, and SDPA full attention
- **CUDA:** cuBLAS compute type (float32 case), cuDNN conv graph, grouped CUTLASS GEMM

No behavior change beyond the log — every gate computes the same boolean as before. The `complex64` cuBLAS case (also `FAST_TF32` today) is intentionally left untouched; probably worth folding into the docs follow-up discussed in #3860.

### Testing

On M5 Max (macOS 26.5), source build with NAX kernels active:

| scenario | warnings |
|---|---|
| default env, fp32 GEMM ×2 in one process | exactly 1 |
| default env, fp32 matvec only | 0 |
| default env, bf16 GEMM | 0 |
| `MLX_ENABLE_TF32=0`, fp32 GEMM | 0 |
| explicit `MLX_ENABLE_TF32=1`, fp32 GEMM | 0 |
| default env, fp32 quantized_matmul (NAX qmm route) | exactly 1 |
| default env, fp32 quantized_matmul (split-K route, not reduced) | 0 |

Numerics unchanged: default fp32 GEMM max rel err vs an fp64 reference is 8.3e-4 (TF32 signature), and 2.5e-7 with `MLX_ENABLE_TF32=0`. `test_blas.py` (24 tests), `test_quantized.py` (32), and `test_fast_sdpa.py` (16) all pass.

The CUDA side compiles the same helper but I have no NVIDIA hardware to run it — @stoyoda0012-cyber offered an sm_120 review in #3860.

### DIFF
--- a/mlx/backend/cuda/conv.cpp
+++ b/mlx/backend/cuda/conv.cpp
@@ -136,7 +136,7 @@ std::optional<DnnGraph> build_conv_graph(
     return std::nullopt;
   }
   graph.deselect_numeric_notes({fe::NumericalNote_t::DOWN_CONVERT_INPUTS});
-  if (dtype == float32 && !env::tf32_active_for_fp32()) {
+  if (dtype == float32 && !env::tf32_active_for_fp32("convolution")) {
     graph.deselect_numeric_notes({fe::NumericalNote_t::TENSOR_CORE});
   }
   CHECK_CUDNN_ERROR(graph.build());
--- a/mlx/backend/cuda/conv.cpp
+++ b/mlx/backend/cuda/conv.cpp
@@ -136,7 +136,7 @@ std::optional<DnnGraph> build_conv_graph(
     return std::nullopt;
   }
   graph.deselect_numeric_notes({fe::NumericalNote_t::DOWN_CONVERT_INPUTS});
-  if (dtype == float32 && !env::enable_tf32()) {
+  if (dtype == float32 && !env::tf32_active_for_fp32()) {
     graph.deselect_numeric_notes({fe::NumericalNote_t::TENSOR_CORE});
   }
   CHECK_CUDNN_ERROR(graph.build());
--- a/mlx/backend/cuda/gemms/cublas_gemm.cpp
+++ b/mlx/backend/cuda/gemms/cublas_gemm.cpp
@@ -19,8 +19,9 @@ cublasComputeType_t dtype_to_compute_type(Dtype dtype) {
     case bfloat16:
       return CUBLAS_COMPUTE_32F;
     case float32:
-      return mlx::core::env::enable_tf32() ? CUBLAS_COMPUTE_32F_FAST_TF32
-                                           : CUBLAS_COMPUTE_32F;
+      return mlx::core::env::tf32_active_for_fp32()
+          ? CUBLAS_COMPUTE_32F_FAST_TF32
+          : CUBLAS_COMPUTE_32F;
     case float64:
       return CUBLAS_COMPUTE_64F;
     case complex64:
--- a/mlx/backend/cuda/gemms/cublas_gemm.cpp
+++ b/mlx/backend/cuda/gemms/cublas_gemm.cpp
@@ -19,9 +19,11 @@ cublasComputeType_t dtype_to_compute_type(Dtype dtype) {
     case bfloat16:
       return CUBLAS_COMPUTE_32F;
     case float32:
-      return mlx::core::env::tf32_active_for_fp32()
-          ? CUBLAS_COMPUTE_32F_FAST_TF32
-          : CUBLAS_COMPUTE_32F;
+      // Compute-type selection only; the once-per-process TF32 warning is
+      // emitted from the CublasGemm constructor, where the shape is known and
+      // rank-1 (matvec) false positives can be suppressed.
+      return mlx::core::env::enable_tf32() ? CUBLAS_COMPUTE_32F_FAST_TF32
+                                           : CUBLAS_COMPUTE_32F;
     case float64:
       return CUBLAS_COMPUTE_64F;
     case complex64:
@@ -82,6 +84,18 @@ CublasGemm::CublasGemm(
       CUBLASLT_MATMUL_DESC_POINTER_MODE,
       &pointer_mode,
       sizeof(pointer_mode)));
+
+  // Warn once (see mlx/utils.h) that fp32 is running at reduced TF32 precision,
+  // but only for a genuine matrix-matrix product. cuBLASLt leaves rank-1 shapes
+  // (M == 1 || N == 1 -- e.g. a matvec whose layout missed the dedicated GEMV
+  // path and fell through to cuBLAS) numerically exact even when FAST_TF32 is
+  // requested, so warning on them is a false positive. Every cuBLAS fp32 entry
+  // -- plain matmul, AddMM (both constructors), and the GEMM-conv fallbacks --
+  // funnels through this constructor, so coverage matches the previous
+  // dtype-only gate in dtype_to_compute_type() while dropping the matvec noise.
+  if (dtype == float32 && M_ > 1 && N_ > 1) {
+    env::tf32_active_for_fp32();
+  }
 }
 
 CublasGemm::CublasGemm(
--- a/mlx/backend/cuda/gemms/grouped_gemm_unaligned.cu
+++ b/mlx/backend/cuda/gemms/grouped_gemm_unaligned.cu
@@ -302,10 +302,12 @@ auto* get_grouped_mm_funcion(Dtype dtype, int N, cu::Device& device) {
       using Arch = MLX_GET_TYPE(arch_tag);
       dispatch_bool(N % 8 == 0, [&](auto is_out_aligned) {
         constexpr int kAlignmentC = is_out_aligned ? 8 : 1;
-        dispatch_bool(env::enable_tf32(), [&](auto kEnableTF32) {
-          fun = grouped_gemm_v2<
-              GemmConfiguration<DataType, Arch, kAlignmentC, kEnableTF32>>;
-        });
+        dispatch_bool(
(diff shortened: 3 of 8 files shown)


stoyoda0012-cyber:
Running the sm_120 pass this week — source build on WSL2, results with numbers to follow.

One design question I'd like to resolve before merge, since it's the item I most want hardware to answer: the Metal gates sit inside the NAX dispatch (shape-aware), so matvec never reaches the helper — consistent with your 0-warning row. The CUDA main gate is `dtype_to_compute_type()`, which selects the compute type without consulting shape. If sm_120 gemv still passes through it, a matvec-only fp32 workload would warn while staying exact fp32 — a false positive, and the one case where "fires on actual engagement" wouldn't hold on the CUDA half. My scenario 2 is exactly this; I'll report either way, and if it does fire I'll propose a shape guard.

Same question likely applies to the cuDNN conv path: if an fp32-conv-only workload warns, the message text ("matmul-family ops") won't help the reader locate it.

pierre427:
Both points are well taken.

**On the gemv/matvec asymmetry**: you've read the structure exactly right — the Metal gates live inside the NAX dispatch, which is shape-aware, so matvec (and any shape that routes to a non-NAX kernel) never reaches the helper; that's where the 0-warning gemv row comes from. The CUDA gate at `dtype_to_compute_type()` is dtype-only by construction. If your scenario-2 run shows a matvec-only fp32 workload warning while staying exact, that's a real false positive on the CUDA half and I'll take your shape guard (or hoist the helper call below a shape check at the cublas site — whichever you prefer, since you have the hardware to validate it). Your measurement should settle it; no need to guess ahead of the sm_120 numbers.

**On the message text**: agreed, and fixed now rather than waiting — `d3d6c38a` threads the op family from the call site into the helper (default `"matmul"`; the conv site passes `"convolution"`, SDPA `"attention"`, quantized `"quantized matmul"`, grouped CUTLASS `"grouped matmul"`), so the once-per-process message names whichever family actually engaged first. No change to *when* the warning fires. M5-verified: GEMM-first warns "float32 matmul ops", NAX-shape quantized-matmul-first warns "float32 quantized matmul ops", gemv and non-NAX qmm shapes stay silent, explicit `=0`/`=1` stay silent; test_blas / test_quantized / test_fast_sdpa green. The CUDA sites are compile-only from me as before — worth an eye during your pass.

stoyoda0012-cyber:
Ran the CUDA half on real sm_120 hardware (RTX 5070 Laptop, WSL2
Ubuntu 24.04, CUDA 13.3, branch @ d3d6c38): warning behaves as
designed for GEMM (fires once per process, silent for bf16 and for
either explicit setting), the cuDNN conv site fires with family
"convolution", and numerics match the measurements in #3860
(2.9e-04 default / 2.1e-07 with `MLX_ENABLE_TF32=0`). Full 774-test
parity suite: exactly 1 warning, failure set identical to the
0.32.0 wheel — no behavioral change.

One asymmetry vs the M5 measurement: **fp32 matvec fires the warning
on CUDA** (your Metal run had 0), while the matvec result itself
stays fp32-exact (rel err 9.7e-08 vs fp64, same process). The CUDA
gate in `dtype_to_compute_type()` is shape-blind, so matvec-only
workloads get warned about a reduction that doesn't happen for that
shape. Suggest either shape-gating that call site or documenting the
requested-mode semantics.

Minor observation for the discussion: under pytest's default capture
the once-per-process warning is invisible (swallowed with whichever
test ran the first fp32 GEMM); `--capture=no` shows it. May be worth
a line in the docs.

pierre427:
Confirmed — thanks for the hardware run. The 9.7e-08 vs 2.9e-04 split is consistent with cuBLASLt not selecting a TF32 tensor-core kernel for that rank-1 shape while the compute descriptor still *requests* `FAST_TF32`; since the CUDA gate at `dtype_to_compute_type()` is dtype-only, it warns on the request regardless of the kernel cuBLAS actually runs. That's the false positive.

(The Metal half stays silent on matvec for a different reason than I first implied: vector shapes are routed to a dedicated `gemv` kernel *before* `steel_matmul_axpby`, where the fp32/TF32 gate lives — so they never reach the warn, rather than being filtered by a shape-aware `use_nax`.)

Fix pushed (`57466a7`): `dtype_to_compute_type()` goes back to silent `enable_tf32()` (pure compute-type selection, unchanged), and the warning moves into the `CublasGemm` constructor, which already holds the output dims (`M_ = a_rows`, `N_ = b_cols`, set by `init_base`), gated on `M_ > 1 && N_ > 1`. I kept it at the constructor rather than a dispatch site on purpose: `AddMM` has a second direct `CublasGemm` construction that bypasses the shared `gemm_and_bias` helper, and the GEMM-conv fallbacks build `CublasGemm` directly too — the constructor is the one point every cuBLAS fp32 entry funnels through, so warning coverage matches the old dtype-only gate exactly.

To be precise about what the guard buys: `M_ > 1 && N_ > 1` suppresses the rank-1 shapes your run showed staying fp32-exact — it is **not** a claim to detect actual tensor-core engagement, since cuBLASLt picks the algorithm by runtime heuristic. The warning stays deliberately conservative: it fires when reduced precision is *requested* on a genuine matrix-matrix shape.

CUDA is still compile-only from me, so would you mind re-checking on the 5070: (1) your scenario-2 matvec now warns 0×, (2) a GEMM-first workload still warns exactly once (family "matmul"), (3) an `AddMM` / `beta != 0` fp32 workload still warns, and (4) the 774-test parity set is unchanged. I'll also add your `--capture=no` observation to the PR notes — good catch.

stoyoda0012-cyber:
Re-ran the matrix on real sm_120 hardware against 57466a7
(RTX 5070 Laptop, WSL2, CUDA 13.3, incremental rebuild):

| scenario | result |
|---|---|
| fp32 matvec → in-process control GEMM | **(0, 1)** — matvec silent, control still warns |
| fp32 GEMM ×2 | 1, family "matmul" |
| fp32 `addmm` | 1, family "matmul" |
| 774-test parity suite (env unset / `MLX_ENABLE_TF32=0`) | 1 warning, failures identical to wheel / 0 warnings, failures identical to wheel |
| fp32 conv2d | 1, family "convolution" |

The shape gate resolves the matvec false positive from my earlier
comment — matvec-only workloads are now silent while the in-process
control GEMM proves the machinery still engages, and the addmm
constructor path warns as intended. No behavioral change in either
suite configuration. LGTM from the hardware side.

PhilipJohnBasile:
Current-base revalidation from a second M5 Max checkout (macOS 27.0 build 26A5388g, Xcode 27.0):

- Replayed the three PR commits cleanly onto current `origin/main` at `973e27f8`.
- Clean source build: `mlx 0.32.1.dev20260726`.
- Runtime warning matrix on `applegpu_g17s`:
  - default fp32 GEMM twice: exactly one warning
  - default fp32 matvec: silent
  - default bf16 GEMM: silent
  - fp32 CPU GEMM: silent
  - explicit `MLX_ENABLE_TF32=0`: silent
  - explicit `MLX_ENABLE_TF32=1`: silent
- Numerical control against a float64 reference for 512x512 fp32 GEMM:
  - default: relative Frobenius error `7.6947e-4`, one warning
  - `MLX_ENABLE_TF32=0`: `2.8766e-7`, no warning
- `test_blas.py`: 27 passed.
- `test_quantized.py` + `test_fast_sdpa.py`: 48 passed, 1 skipped.

The current pre-commit clang-format hook finds two wrapping-only violations in the PR diff. The mechanical correction is isolated at [`2b5f6eec`](https://github.com/PhilipJohnBasile/mlx/commit/2b5f6eec98e32d8d9a71471d27ec8cd269333f77) and can be cherry-picked; all pre-commit hooks pass afterward.

Security classification for clarity: based on the reported behavior and reproductions, this is a numerical-correctness and runtime-observability issue, **not a cybersecurity vulnerability**. It changes floating-point precision and can alter downstream numerical decisions, but does not introduce an access-control bypass, secret exposure, arbitrary code execution, or a demonstrated memory-safety boundary violation. The warning is still important because the application-level accuracy impact can be substantial.

I do not have NVIDIA hardware locally; the sm_120 execution evidence remains the real-hardware validation reported above by @stoyoda0012-cyber.

zcbenz (commented):
Thanks for the PR but printing warning for TF32 is not really common practice as far as I know, and I think documentation + programmable switch would be a more reasonable choice.

PR #3889, opened 2026-07-22
Fix Adamax betas docstring and MultiOptimizer filters type

ayaangazali:
## Proposed changes

Two docstring fixes in `optimizers.py`, no behavior change:

1. `Adamax` betas: the doc says the coefficients compute "running averages of the gradient and its square", but that describes Adam. Adamax's second moment is the exponentially weighted infinity norm (`v = max(beta2 * v, |g|)`, matching the math block and `apply_single`), nothing is squared. Reworded to match.
2. `MultiOptimizer` `filters`: the type had unbalanced brackets `list[Callable[[str, array], bool]` (missing the outer closing `]`).

## Checklist

(docstring-only; test_optimizers.py still passes 25/25)

---

honesty note: freshman here, i lean on Claude Code to spot doc/code mismatches but i checked the Adamax update math (max of the infinity norm, not a squared term) against the code myself before writing this. happy to be corrected if i phrased anything poorly.

### DIFF
--- a/python/mlx/optimizers/optimizers.py
+++ b/python/mlx/optimizers/optimizers.py
@@ -165,7 +165,7 @@ class MultiOptimizer(Optimizer):
 
     Args:
         optimizers (list[Optimizer]): A list of optimizers to delegate to
-        filters (list[Callable[[str, array], bool]): A list of predicates that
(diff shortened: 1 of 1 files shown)


angeloskath (approved):
Thanks!

PR #3891, opened 2026-07-22
Avoid unaligned writes in CPU RandomBits

markjouh:
## Summary

- avoid unaligned 32-bit stores in the CPU `RandomBits` implementation
- preserve the generated bytes and Threefry sequence by using `memcpy` for complete words
- leave the existing partial-word copy path unchanged

## Root cause

For batched keys, each key's output begins `bytes_per_key` bytes after the
previous one. When that stride is not divisible by four, casting the start of a
later key's output to `uint32_t*` and writing through it is undefined behavior.

The existing `test random bits` case covers an affected layout: vmapping five
`uint16` values over two keys makes each per-key output 10 bytes wide. Before
this change, that test aborts under UBSan with a misaligned `uint32_t` store.

## Validation

- `UBSAN_OPTIONS=halt_on_error=1:print_stacktrace=1 build-ubsan/tests/tests -tc='test random bits'` (1 test, 28 assertions)
- release C++ suite (244 tests, 3,238 assertions)
- UBSan suite excluding two unrelated pre-existing GGUF alignment failures (242 tests, 3,187 assertions)
- `clang-format --dry-run --Werror mlx/backend/cpu/primitives.cpp`
- `git diff --check`

### DIFF
--- a/mlx/backend/cpu/primitives.cpp
+++ b/mlx/backend/cpu/primitives.cpp
@@ -3,6 +3,7 @@
 #include <algorithm>
 #include <cassert>
 #include <cmath>
+#include <cstring>
 #include <numeric>
 #include <sstream>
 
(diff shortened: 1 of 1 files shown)


zcbenz:
Can you paste the UBSAN logs?

zcbenz:
Did some research on unaligned memory access in CPU and my conclusion is that we should be fine when compiler thinks it is safe to generate unaligned access code, the majority of software including linux kernel does not care about this https://gcc.gnu.org/bugzilla/show_bug.cgi?id=93031.

Issue #2915, opened 2025-12-16
[CI] linux sanitizer tests - task list for future improvements


incertum:
See initial PR https://github.com/ml-explore/mlx/pull/2860

- [ ] enable ASAN for GPU backend tests
- [ ] enable TSAN tests (may require a few fixes first)
- [ ] consider enabling `halt_on_error` for UBSAN tests once it's the right time. Sometimes UBSAN can be tricky to address. Currently we have the logs in the CI as the tests are enabled, but we do not error out (as of 12/6/2025)

CC @angeloskath and @awni
CC @madrob

zcbenz:
Closing since there is no one planning on working on this.

PR #3892, opened 2026-07-22
Handle packed GGUF metadata without unaligned access

markjouh:
# Handle packed GGUF metadata without unaligned access

## Summary

- read scalar values, array headers, and string lengths from packed GGUF metadata with `memcpy`
- copy numeric array payloads into aligned, owning MLX buffers before constructing arrays
- serialize array headers and string lengths without casting byte buffers to aligned GGUF structs

## Background

GGUF metadata is tightly byte-packed, so metadata values are not guaranteed to satisfy the host alignment requirements of `gguf_value`, `gguf_string`, or their member types. Dereferencing those typed views is undefined behavior. The existing `test gguf metadata` case reproduces this immediately when UBSan is configured with `halt_on_error=1`.

This addresses one of the UBSan failures tracked by #2915.

## Validation

- Before the change: `test gguf metadata` aborts in `set_mx_value_from_gguf` on a misaligned `gguf_value` access.
- UBSan targeted tests with `halt_on_error=1`: 2 test cases and 51 assertions passed.
- UBSan full suite with `halt_on_error=1`, excluding only `test random bits` (addressed separately by #3891): 243 test cases and 3,210 assertions passed.
- Release full suite: 244 test cases and 3,238 assertions passed.
- ASan targeted GGUF tests with leak detection enabled: 2 test cases and 51 assertions passed.
- ASan full suite with leak detection disabled: 244 test cases and 3,238 assertions passed. With leak detection enabled, all assertions still pass before LeakSanitizer reports unrelated graph-object leaks from arithmetic tests.
- `clang-format --dry-run --Werror mlx/io/gguf.cpp`
- `git diff --check`

## Risk

The serialized GGUF byte layout and public API are unchanged. The patch is limited to `mlx/io/gguf.cpp`, and the existing save/load round-trip tests cover scalar, numeric-array, string, and string-array metadata paths.

### DIFF
--- a/mlx/io/gguf.cpp
+++ b/mlx/io/gguf.cpp
@@ -13,6 +13,30 @@ namespace mlx::core {
 // https://github.com/antirez/gguf-tools/blob/af7d88d808a7608a33723fba067036202910acb3/gguflib.h#L102-L108
 constexpr int gguf_array_header_size = 12;
 
+namespace {
+
+// GGUF metadata is byte-packed, so values need not satisfy host alignment.
+template <typename T>
+T read_unaligned(const void* src) {
+  T value;
+  std::memcpy(&value, src, sizeof(value));
+  return value;
+}
+
+template <typename T>
+void write_unaligned(void* dst, T value) {
+  std::memcpy(dst, &value, sizeof(value));
+}
+
+array array_from_bytes(const void* src, int size, Dtype dtype) {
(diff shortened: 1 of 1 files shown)


zcbenz:
Did some research on unaligned memory access in CPU and my conclusion is that we should be fine when compiler thinks it is safe to generate unaligned access code, the majority of software including linux kernel does not care about this https://gcc.gnu.org/bugzilla/show_bug.cgi?id=93031.

Issue #3858, opened 2026-07-17
[BUG] [CUDA] Batched matrix-vector product crashes when batch > 65,535 (gridDim limit)


stoyoda0012-cyber:
## Describe the bug

On the CUDA backend, a batched matrix-vector `matmul` crashes with
`cudaGraphAddKernelNode(...) failed: invalid argument` when the batch
dimension exceeds 65,535. Bisected exactly: **B = 65,535 works,
B = 65,536 crashes** — the CUDA `gridDim.y/z` limit, suggesting the
batched-GEMV kernel maps the batch axis to a grid dimension without
tiling.

Square batched matmul at the same batch size is fine
(`(65536, 8, 8) @ (65536, 8, 8)` works), so this is specific to the
matvec (n=1 output column) path.

With `MLX_USE_CUDA_GRAPHS=0` the same shape still fails, just later:
`cudaLaunchKernelExC(&config, func, params) failed: invalid argument` —
so it is the kernel launch configuration itself, not the graph capture.

After the failure the CUDA context appears unusable: subsequent
unrelated `mx.eval` calls abort the process.

Metal has no equivalent grid limit, so code ported from Apple silicon
hits this only on CUDA, and only at production batch sizes (our parity
test suite passed everywhere because test batches are small).

## To Reproduce

```python
import mlx.core as mx

for B in (65535, 65536):
    A = mx.zeros((B, 225, 4))
    v = mx.zeros((B, 4, 1))
    try:
        mx.eval(A @ v)
        print(f"B={B}: OK")
    except RuntimeError as e:
        print(f"B={B}: {e}")
```

Output:

```
B=65535: OK
B=65536: cudaGraphAddKernelNode(&node, graph_, NULL, 0, &params) failed: invalid argument
```

## Expected behavior

Batched matmul should either tile the batch axis across grid launches
(as large-size elementwise ops already do) or raise a clear
shape-limit error at trace time.

## Desktop

- OS: Windows 11 + WSL2 (Ubuntu 24.04), kernel 6.18.33.2-microsoft-standard-WSL2
- GPU: NVIDIA GeForce RTX 5070 Laptop (Blackwell, sm_120), 8 GB, driver 596.13
- MLX: 0.32.0 (`pip install "mlx[cuda13]"`, `mlx-cuda-13` 0.32.0)
- Python 3.12.3, NumPy 2.3.5

## Additional context

Looks like the same class as the (fixed) #3666 ("fix grid for large
[...]

PR #3893, opened 2026-07-22
[CUDA] Fix grid overflow in gemm conv unfold kernels for >= 65,536 output positions

AdamDLuz:
The naive unfold kernels in `backend/cuda/conv/gemm_conv.cu` and `gemm_grouped_conv.cu` launch with

```cpp
num_blocks.x = cuda::ceil_div(wt_spatial_size, block_dims.x);
num_blocks.y = params.C;
num_blocks.z = mat_M;          // mat_M = N * out_pixels
```

`gridDim.z` is limited to 65,535, so any convolution with ≥ 65,536 output positions — e.g. a batch-1 **256×256** image, common in VAE decoders — fails with `cudaGraphAddKernelNode(...): invalid argument` (or `cudaLaunchKernelExC(...): invalid argument` with `MLX_USE_CUDA_GRAPHS=0`). Meanwhile `grid.x` (limit 2³¹−1) carries only `ceil_div(wt_spatial_size, block_dims.x)`, which is 1 for typical filters.

This PR swaps the mapping so the unbounded matrix-row index rides `grid.x` and the filter-spatial blocks ride `grid.z`, updating the two `tid.x`/`tid.z` consumers in each kernel accordingly. Per-block thread layout and memory access patterns are unchanged.

Same bug class as #3858 (batched matvec, batch > 65,535) and the merged #3666 (binary ops).

Repro (before):
```python
import mlx.core as mx
x = mx.zeros((1, 256, 256, 512), dtype=mx.bfloat16)
w = mx.zeros((512, 3, 3, 512), dtype=mx.bfloat16)
mx.eval(mx.conv2d(x, w, stride=1, padding=1))   # invalid argument
# (1, 128, 128, 512) works: 16,384 <= 65,535
```

Found while debugging ollama/ollama#14843 (FLUX.2-klein image generation on Linux/CUDA): the VAE decoder's upsample-to-256 conv is the first launch to cross the limit. Verified indirectly on an NVIDIA L4 (CUDA 13) via an equivalent host-side chunking workaround producing correct 512×512 and 1024×1024 images; the patch itself is a pure grid-axis remap (not compile-tested here — missing local dep tree).

Note: `num_blocks.y = params.C` retains a theoretical >65,535-channel limit, untouched here.

### DIFF
--- a/mlx/backend/cuda/conv/gemm_conv.cu
+++ b/mlx/backend/cuda/conv/gemm_conv.cu
@@ -25,17 +25,19 @@ __global__ void naive_unfold_nd(
   auto tid = block.group_index();
   auto lid = block.thread_index();
 
-  int index_batch = tid.z / out_pixels; // [0, N)
-  int index_out_spatial = tid.z % out_pixels; // [0, H_out * W_out)
+  // The matrix-row index (N * out_pixels values) rides grid.x, which allows
+  // up to 2^31-1 blocks; grid.y and grid.z are limited to 65,535.
+  int index_batch = tid.x / out_pixels; // [0, N)
+  int index_out_spatial = tid.x % out_pixels; // [0, H_out * W_out)
   int index_wt_spatial =
-      tid.x * block.dim_threads().x + lid.x; // [0, H_wt * W_wt)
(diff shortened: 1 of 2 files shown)


nastya236 (approved):
Looks great! THanks for the fix.

PR #3894, opened 2026-07-23
docs: document the reduced-precision float32 default and MLX_ENABLE_TF32

stoyoda0012-cyber:
Documents the reduced-precision default discussed in #3860.

Adds a short "Numerical Precision" usage page stating two things:

* float32 matmul-family operations (matmul, quantized matmul, grouped
  matmul, convolution and attention) may run at reduced precision by
  default on hardware with dedicated matrix-multiplication units;
* `MLX_ENABLE_TF32=0` at launch keeps them in full float32.

Per review, per-backend descriptions and implementation-specific
behaviour are left out, so the page does not need updating when either
changes.

### Testing

Built the docs with `sphinx-build -b html` against
`docs/requirements.txt`: the page renders, is reachable from the Usage
toctree, and none of the six build warnings originate from it (they are
the pre-existing `nn.distributed` autosummary imports against the pip
wheel, the doxygen-generated C++ group, and the `printoptions` stub).

### DIFF
--- a/docs/src/usage/precision.rst
+++ b/docs/src/usage/precision.rst
@@ -1,0 +1,53 @@
+.. _precision:
+
+Numerical Precision
+===================
+
+By default MLX runs ``float32`` matrix-multiplication family operations
+(matmul, quantized matmul, grouped matmul, convolution, and attention)
+at reduced, TF32-class precision on hardware with dedicated
+matrix-multiplication units:
+
+* On the CUDA backend, cuBLAS and cuDNN are asked for TF32 tensor-core
+  math.
+* On Apple silicon with a neural accelerator (the M5 generation), the
+  Metal backend dispatches these operations to accelerator kernels of
+  comparable precision. Earlier Apple silicon (M1–M4) runs them in full
+  ``float32``.
+
+TF32-class math keeps the ``float32`` exponent range but rounds the
+significand to about 10 bits, so results carry roughly three decimal
+digits instead of seven. For a 512x512 ``float32`` matrix product the
+relative error against a ``float64`` reference is on the order of
+``1e-4`` to ``1e-3``, compared with about ``2e-7`` in full ``float32``
+(measured on both backends). Individual operations still look correct,
+but workloads that compare nearly equal values downstream — for
+example an ``argmax`` over correlation scores, or a kernel-vs-reference
+test with tight tolerances — can change results.
+
+To run these operations in full ``float32`` precision, set the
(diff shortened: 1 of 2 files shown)


docs/src/usage/precision.rst
```diff
@@ -0,0 +1,64 @@
...
+  math.
+* On Apple silicon with a neural accelerator (the M5 generation and
+  newer, on macOS 26.2 or later), the Metal backend dispatches these
+  operations to accelerator kernels of comparable precision. Earlier
+  Apple silicon (M1–M4) always runs them in full ``float32``: on those
+  machines ``MLX_ENABLE_TF32`` has no observable effect because the
+  reduced-precision path is never taken, not because the setting is
+  ignored.
+
+TF32-class math keeps the ``float32`` exponent range but rounds the
```
zcbenz: The technical details should be ignored since Apple silicon might not do exactly same.

docs/src/usage/precision.rst
```diff
@@ -0,0 +1,64 @@
...
+
+  import os
+
+  os.environ["MLX_ENABLE_TF32"] = "0"  # before the first matmul
+
+  import mlx.core as mx
+
+  # float32 matmuls now run at full float32 precision
+
+A few details worth knowing:
```
zcbenz: I prefer omitting the details because they could change at any time and we might not remember to update the docs here.

docs/src/usage/precision.rst
stoyoda0012-cyber: Dropped, along with the TF32-format paragraph. Agreed that the docs
shouldn't assert equivalence between the Apple silicon path and NVIDIA's
TF32 -- that was inferred from measurements rather than from anything
either vendor specifies, and it is exactly the kind of claim that goes
stale silently.

docs/src/usage/precision.rst
stoyoda0012-cyber: Dropped. I also removed the in-process os.environ example, which is what
made the latching caveat necessary in the first place -- the page now
shows only the launch-time form, so there is no first-use ordering left
to explain.

stoyoda0012-cyber:
Reduced the page to the two things that hold regardless of backend and
hardware generation: float32 matmul-family ops may run at reduced
precision by default, and MLX_ENABLE_TF32=0 opts out. Anything that
depends on a vendor's current behaviour is gone; those measurements are
recorded in #3860 instead, where they carry their date and device.

zcbenz (approved):
Thanks!

PR #3898, opened 2026-07-23
Fix prod dtype promotion when reducing a size-1 axis

eyupcanakman:
## Proposed changes

`prod` computes its promoted output type but discards it on the no-op branch, so the result dtype depends on the length of the reduced axis. Reducing a size-1 axis returns the input dtype while a size-2 axis promotes:

```python
mx.prod(mx.array([2], dtype=mx.int8), axis=0).dtype     # int8   (before)
mx.prod(mx.array([2, 2], dtype=mx.int8), axis=0).dtype  # int32
```

`sum` casts on the same branch (`is_noop ? astype(a, out_type, s) : ...`), but `prod` returned bare `a`. The fix applies the same `astype`, so `prod` matches `sum` and the other promoting reductions. The promotion is there to keep the product from overflowing the narrow input type, so the size-1 case was also silently keeping a narrow accumulator.

Added a test that the output dtype does not depend on the reduced axis length.

## Checklist

### DIFF
--- a/mlx/ops.cpp
+++ b/mlx/ops.cpp
@@ -2397,7 +2397,7 @@ array prod(
     out_type = int32;
   }
   auto out = (is_noop)
-      ? a
+      ? astype(a, out_type, s)
       : array(
             std::move(out_shape),
             out_type,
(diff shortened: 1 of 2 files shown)


angeloskath (approved):
👍

PR #3899, opened 2026-07-23
Making JACCL coordinator optional

angeloskath:
This allows the user to define their own side channel used to exchange RDMA metadata to initialize the connection.

This can be used for apps that have a different way to establish a slow connection between the nodes that can be used to exchange metadata. It avoids the need for opening a port and listening for connections etc. It could use a shared filesystem or iCloud or any centralized service etc

Here is an example I asked Claude to write that uses ZeroMQ to handle connection retries etc.

<details>
<summary>ZeroMQ all gather example</summary>

```python
# Copyright © 2025 Apple Inc.

"""A minimal ZeroMQ based all-gather side channel for the jaccl backend.

The jaccl backend bootstraps its RDMA connections by performing a couple of
byte-level all-gathers to exchange connection metadata. This module provides a
simple :func:`zmq_all_gather_factory` that implements that all-gather over TCP
using ZeroMQ. Rank 0 acts as a coordinator: every other rank connects to it,
sends its chunk and receives the fully assembled buffer back.

Usage:

    import mlx.core as mx
    from mlx._distributed_utils.zmq_all_gather import zmq_all_gather_factory

    group = mx.distributed.init(
        backend="jaccl",
        all_gather_factory=zmq_all_gather_factory(),
    )
"""

import os
import struct

import zmq

# The environment variable set by ``mlx.launch`` that holds the ``ip:port`` of
# rank 0 which acts as the all-gather coordinator.
_COORDINATOR_ENV = ("JACCL_COORDINATOR", "MLX_JACCL_COORDINATOR")


def _coordinator_address(coordinator):
    if coordinator is not None:
        return coordinator
    for name in _COORDINATOR_ENV:
        value = os.environ.get(name)
        if value:
            return value
    raise RuntimeError(
        "No coordinator address provided and neither of "
        f"{_COORDINATOR_ENV} is set in the environment."
    )


def zmq_all_gather_factory(coordinator=None):
    """Create an all-gather factory for the jaccl backend backed by ZeroMQ.

    Args:
        coordinator (str, optional): The ``ip:port`` of rank 0. If ``None`` it
            is read from the ``MLX_JACCL_COORDINATOR`` environment variable.

    Returns:
        Callable[[int, int], Callable[[bytes, int], bytes]]: A factory that
        ``mx.distributed.init`` calls once per rank with ``(rank, size)`` and
        returns the actual all-gather callable.
    """
    ip, port = _coordinator_address(coordinator).rsplit(":", 1)
    endpoint = f"tcp://{ip}:{port}"

    def factory(rank, size):
        context = zmq.Context.instance()

        if rank == 0:
            # The coordinator uses a ROUTER socket so it can gather every
            # rank's chunk before replying with the assembled buffer.
            socket = context.socket(zmq.ROUTER)
            socket.bind(f"tcp://*:{port}")

            def all_gather(src, n_bytes):
                chunks = [None] * size
                chunks[0] = src
                envelopes = []
                for _ in range(size - 1):
                    frames = socket.recv_multipart()
                    # REQ sockets add an empty delimiter frame; everything
                    # before the payload is the routing envelope.
                    envelope, msg = frames[:-1], frames[-1]
                    peer = struct.unpack("<i", msg[:4])[0]
                    chunks[peer] = msg[4:]
                    envelopes.append(envelope)
                assembled = b"".join(chunks)
                for envelope in envelopes:
                    socket.send_multipart(envelope + [assembled])
                return assembled

            return all_gather

        # Non-coordinator ranks send their chunk and receive the full buffer.
        socket = context.socket(zmq.REQ)
        socket.connect(endpoint)

        def all_gather(src, n_bytes):
            socket.send(struct.pack("<i", rank) + src)
            return socket.recv()

        return all_gather

    return factory
```

</details>

### DIFF
--- a/mlx/distributed/distributed.cpp
+++ b/mlx/distributed/distributed.cpp
@@ -138,9 +138,19 @@ Group Group::split(int color, int key /* = -1 */) const {
   return Group(group_->split(color, key));
 }
 
-Group init(bool strict /* = false */, const std::string& bk /* = "any" */) {
+namespace {
+
+std::unordered_map<std::string, std::shared_ptr<detail::GroupImpl>>&
+get_backends() {
   static std::unordered_map<std::string, std::shared_ptr<detail::GroupImpl>>
       backends;
+  return backends;
+}
+
+} // namespace
+
+Group init(bool strict /* = false */, const std::string& bk /* = "any" */) {
+  auto& backends = get_backends();
 
   // Already initialized so return the group.
   if (auto g = backends.find(bk); g != backends.end()) {
@@ -192,6 +202,10 @@ Group init(bool strict /* = false */, const std::string& bk /* = "any" */) {
   }
   backends.insert({std::move(bk_), group});
   return Group(group);
+}
+
+void clear_backends() {
+  get_backends().clear();
 }
 
 } // namespace mlx::core::distributed
--- a/mlx/distributed/distributed.cpp
+++ b/mlx/distributed/distributed.cpp
@@ -147,6 +147,17 @@ get_backends() {
   return backends;
 }
 
+Group register_group(std::shared_ptr<detail::GroupImpl> group, std::string bk) {
+  auto& backends = get_backends();
+  if (group == nullptr) {
+    group = std::make_shared<detail::EmptyGroup>();
+  } else {
+    backends.insert({"any", group});
+  }
+  backends.insert({std::move(bk), group});
+  return Group(group);
+}
+
 } // namespace
 
 Group init(bool strict /* = false */, const std::string& bk /* = "any" */) {
@@ -195,13 +206,27 @@ Group init(bool strict /* = false */, const std::string& bk /* = "any" */) {
     throw std::invalid_argument(msg.str());
   }
 
-  if (group == nullptr) {
-    group = std::make_shared<detail::EmptyGroup>();
-  } else {
-    backends.insert({"any", group});
+  return register_group(std::move(group), std::move(bk_));
+}
+
+Group init(bool strict, const std::string& bk, AllGatherFactory factory) {
+  if (bk != "jaccl") {
+    std::ostringstream msg;
+    msg << "[distributed] An all gather factory is only supported with the "
+        << "'jaccl' backend but '" << bk << "' was provided.";
+    throw std::invalid_argument(msg.str());
   }
-  backends.insert({std::move(bk_), group});
-  return Group(group);
+
+  auto& backends = get_backends();
+
+  // Already initialized so return the cached group. The factory is ignored in
+  // this case since the backend is already up.
+  if (auto g = backends.find(bk); g != backends.end()) {
+    return Group(g->second);
+  }
+
+  auto group = jaccl::init(strict, std::move(factory));
+  return register_group(std::move(group), bk);
 }
 
 void clear_backends() {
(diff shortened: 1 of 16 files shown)


nastya236:
Looks awesome! Thanks for putting me as a reviewer, looking at it today!

python/src/distributed.cpp
```diff
@@ -75,10 +79,60 @@ void init_distributed(nb::module_& parent_module) {
...
+              msg << "Custom all-gather returned " << result.size()
+                  << " bytes but expected " << expected;
+              throw std::runtime_error(msg.str());
+            }
+            std::memcpy(dst, result.data(), expected);
+          };
+        };
+
+        auto group =
+            mx::distributed::jaccl::init(strict, std::move(cpp_factory));
```
nastya236: Based on my understanding, `mx::distributed::init(strict, backend)` (line 87) with register initialised backend and put it into a map, then we could do `mx.distributed.all_sum(mx.ones((9,))` without a group being specified. But for all gather factory path if you call init directly for jccl, would it still work as expected ?

mlx/distributed/distributed.cpp
```diff
@@ -194,4 +204,8 @@ Group init(bool strict /* = false */, const std::string& bk /* = "any" */) {
   return Group(group);
 }
 
+void clear_backends() {
```
nastya236, line 232: This is related to my question about the jaccl group with a custom all_gather not being added to the backends map. If that's the case, then clear_backends() won't actually clear it, will it?🤔

mlx/distributed/jaccl/lib/jaccl/jaccl.cpp
```diff
@@ -142,6 +174,12 @@ bool Config::is_valid_ring() const {
   return true;
 }
 
+bool Config::is_valid() const {
+  return size_ >= 1 && rank_ < size_ && rank_ >= 0 &&
```
nastya236, line 178: Nit, seems like `size_ >= 1` is redundant because `is_valid_mesh() || is_valid_ring()` => `size_ > 2`?

mlx/distributed/jaccl/no_jaccl.cpp
```diff
@@ -17,4 +17,18 @@ std::shared_ptr<GroupImpl> init(bool strict /* = false */) {
   return nullptr;
 }
 
+std::shared_ptr<GroupImpl> init(bool strict, AllGatherFn /* fn */) {
```
nastya236: Nit. I guess this is leftover from the first draft where all_gather was a function not a factory?:)

nastya236 (commented):
Beautiful pull request. 

In the old version `JACCL_COORDINATOR` was required: rank 0 listens on a port, every rank connects to it (except rank 0 of course) and the RDMA connection metadata is exchanged through that TCP communication. This PR makes it possible throw an alternative side channel: user pass a factory that returns an all_gather, and jaccl uses that. This is amazing because 1) we don't need to setup second network 2) rank 0 doesn't need a reachable IP and an open listening port. 

I left a couple of questions, but looks awesome. 
I have a question regarding a barrier: why barrier is implemented with 2 all_gather?

mlx/distributed/jaccl/lib/jaccl/jaccl.cpp
angeloskath: `size_ >= 2` yes indeed. Just there as an early exit for the potentially common case of singleton configs.

python/src/distributed.cpp
angeloskath: Yes of course it is thanks for catching that. It 'd be a total mess 🙈

Moved the init in mlx/distributed/distributed.cpp where it should be and where it registers to the backends. Not sure where this was lost in translation 😬

nastya236 (approved):
Looks great!

PR #3900, opened 2026-07-23
Refactor the JACCL ring and add threads for the multiple rings

angeloskath:
This is on top of the coordinator changes and I will rebase it when that is merged, for now focus on the changes in the reduction ops and more importantly the ring implementation.

The TL;DR is the following:
- Refactor the ring to extract the ring pass in a standalone function (with copy op or reduction)
- Refactor the reduction ops to support out of place reduction that allows us to avoid the initial copy
- Write a per-wire all reduce and all gather using the ring pass
- Copy the MLX threadpool and use it to dispatch the other wires to different threads

The performance on a single wire is maintained and there should be 0 overhead since that wire runs in the calling thread.

The following is a 2-node multi-wire all_sum before and after
ring-bandwidth

And a 4 node all_sum with 2 rings before and after
 
ring-4-bench

### DIFF
--- a/mlx/distributed/jaccl/lib/jaccl/reduction_ops.h
+++ b/mlx/distributed/jaccl/lib/jaccl/reduction_ops.h
@@ -8,11 +8,24 @@
 
 namespace jaccl {
 
+// Each reduction op has an in place form out[i] OP= in[i] and an out of place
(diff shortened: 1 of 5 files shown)


zcbenz (approved):
👍

PR #3901, opened 2026-07-23
Add scatter reduce for JACCL

angeloskath:
Like #3900 this is on top of the coordinator changes and the threads changes. I will be rebasing as these are merged.

This PR adds scatter reduce to the JACCL backend. It also removes the ring from the mesh implementation and switches to a scatter gather approach for the high BW all reduce which makes better use of the many cables than the ring reduce. We now use the naive all reduce only for sizes <= 32KB which are completely latency bound.

For the ring backend we can't unfortunately use the ring_pass for the scatter reduce because it requires a buffer of the same size as the input to avoid having race conditions in the read/write location. The approach in the scatter reduce is to only write in the output buffer at parts that have already been copied to the RDMA buffers.

Here is a 4-node all reduce comparison of before and after this PR.

allreduce_benchmark

### DIFF
--- a/mlx/backend/cpu/distributed.cpp
+++ b/mlx/backend/cpu/distributed.cpp
@@ -98,6 +98,24 @@ void Recv::eval_cpu(
 void ReduceScatter::eval_cpu(
     const std::vector<array>& inputs,
     std::vector<array>& outputs) {
-  throw std::runtime_error("[ReduceScatter] Not implemented yet.");
+  assert(inputs.size() == 0);
+  assert(outputs.size() == 1);
+
+  auto [in, copied] = ensure_row_contiguous(inputs[0], stream());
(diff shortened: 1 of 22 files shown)


mlx/backend/cpu/distributed.cpp
```diff
@@ -98,6 +98,24 @@ void Recv::eval_cpu(
...
+  auto [in, copied] = ensure_row_contiguous(inputs[0], stream());
+  outputs[0].set_data(allocator::malloc(outputs[0].nbytes()));
+
+  switch (reduce_type_) {
+    case Sum:
+      distributed::detail::sum_scatter(group(), in, outputs[0], stream());
+      break;
+    default:
+      throw std::runtime_error(
+          "Only all reduce sum, min and max are supported for now");
```
zcbenz: Error message needs an update.

zcbenz (approved):
👍

PR #3903, opened 2026-07-23
Fix InstanceNorm Shape docstring to require at least 3 dimensions

ayaangazali:
## Proposed changes

The `InstanceNorm` docstring Shape section says the input is `(..., C)`, which reads like a 2D `(N, C)` input is allowed, but `__call__` raises for anything with fewer than 3 dimensions:

```python
import mlx.core as mx
import mlx.nn as nn

nn.InstanceNorm(4)(mx.zeros((2, 4)))
# ValueError: InstanceNorm expects inputs with at least 3 dimensions (N, ..., C) ...
```

Updated the Shape line to `(N, ..., C)` and noted the at-least-3-dimensions requirement, matching the error message and the reduction over the spatial axes. Same spirit as the earlier BatchNorm shape doc fix (#3782).

## Checklist

(docstring-only, test_nn.py passes 69/69)

---

being honest: freshman here, Claude Code helps me find these mismatches but i reproduced the 3-dimension error above myself and read the reduction axes to make sure the constraint is real before opening this. corrections welcome :)

### DIFF
--- a/python/mlx/nn/layers/normalization.py
+++ b/python/mlx/nn/layers/normalization.py
@@ -25,7 +25,8 @@ class InstanceNorm(Module):
         affine (bool): Default: ``False``.
 
     Shape:
-      - Input: :math:`(..., C)` where :math:`C` is equal to :attr:`dims`.
+      - Input: :math:`(N, ..., C)` where :math:`C` is equal to :attr:`dims`.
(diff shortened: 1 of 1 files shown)


zcbenz (approved):
Thanks!

PR #3905, opened 2026-07-24
add runnable data parallel example

tylergibbs1:
## proposed changes

#2930 asks for standalone distributed examples. #2973 added the explanation, but `examples/python` still has no data parallel script users can run end to end.

this adds a linear regression example where each rank creates its own synthetic data shard, starts from the same parameters, and averages gradients with `nn.average_gradients`. the final loss is evaluated on every rank before rank zero prints. without that step, lazy evaluation can leave one process waiting on a collective after its peer exits.

run it with:

```sh
mlx.launch -n 2 python examples/python/data_parallel.py
```

on my m4 mac, the one rank run finished with an l2 distance of `0.00971`. the two rank run finished at `0.00464`.

## checks

- `python examples/python/data_parallel.py`
- `mlx.launch -n 2 python examples/python/data_parallel.py`

## checklist

- [x] i have read the contributing document
- [x] i have run `pre-commit run --all-files`
- [ ] i have added automated tests. this pr adds a runnable example and the checks above exercise both paths
- [ ] i have updated the necessary documentation. no api or existing docs changed

### DIFF
--- a/examples/python/data_parallel.py
+++ b/examples/python/data_parallel.py
@@ -1,0 +1,59 @@
+# Copyright © 2026 Apple Inc.
+
+"""train linear regression with data parallel workers.
+
+run locally with:
+
+    mlx.launch -n 2 python examples/python/data_parallel.py
+"""
+
+import time
+
+import mlx.core as mx
+import mlx.nn as nn
+
+num_features = 100
+num_examples = 1_000
+num_iters = 100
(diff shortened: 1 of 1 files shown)


zcbenz:
Thanks for the PR but I don't think this is a very useful example as the data is just generated separately in each node.

PR #3907, opened 2026-07-24
Accumulate low-precision float scans in float32 on the CPU

v-code01:
### Problem

The CPU scan keeps its running value in the output dtype and reads it back from the output buffer each step (`op(*(output - 1), *input)`). For `float16` / `bfloat16` a long cumulative sum stops growing once the running value's ULP exceeds the increment:

```python
mx.cumsum(mx.ones((5000,), mx.float16))[-1]   # CPU: 2048.0,  GPU: 5000.0
mx.cumsum(mx.ones((5000,), mx.bfloat16))[-1]  # CPU:  256.0,  GPU: 4992.0
```

The float16 sum saturates at 2048 and bfloat16 at 256, regardless of length. The GPU's blocked scan stays accurate, so the **same call returns different results on the two backends**. (numpy has the same saturation summing in-dtype, but the CPU/GPU divergence within MLX is the real issue.)

The file already anticipated this:

```cpp
// TODO: If we add the option to accumulate floats in higher precision
//       floats perhaps we should add the full all-to-all dispatch.
```

### Fix

Carry the scan accumulator in a separate `AccT` type and narrow to the output dtype only on store, mirroring the existing `bool -> int32` accumulation. `AccT` is `float32` for `float16`/`bfloat16` and the output type for every other dtype, so all non-low-precision dtypes are bit-identical and only the float16/bfloat16 paths change. The contiguous and strided scans both carry the accumulator (the strided path uses one accumulator per lane).

After the fix the CPU matches the GPU and the true value:

```python
mx.cumsum(mx.ones((5000,), mx.float16))[-1]   # 5000.0
mx.cumsum(mx.ones((5000,), mx.bfloat16))[-1]  # 4992.0  (nearest bfloat16, == GPU)
```

### Tests

Extended `test_scans` with a float16/bfloat16 `cumsum`-of-ones check past the stall point (sizes 4096/5000/20000, contiguous and strided), pinned to the CPU stream. It fails on the current CPU backend (stalls at 2048) and passes with the fix. Full `test_ops.py` (146 tests / 309 subtests) and the autograd cumprod-grad test stay green; `cumsum`/`cumprod` for int and float32 remain bit-identical to numpy.

Verified with a CPU-only build (`MLX_BUILD_METAL=OFF`) on an M4.

### DIFF
--- a/mlx/backend/cpu/scan.cpp
+++ b/mlx/backend/cpu/scan.cpp
@@ -1,6 +1,7 @@
 // Copyright © 2023 Apple Inc.
 
 #include <cassert>
+#include <vector>
 
 #include "mlx/backend/common/utils.h"
 #include "mlx/backend/cpu/binary_ops.h"
@@ -13,7 +14,7 @@ namespace mlx::core {
 
 namespace {
 
-template <typename T, typename U, typename Op>
+template <typename T, typename U, typename AccT, typename Op>
 void contiguous_scan(
     const T* input,
     U* output,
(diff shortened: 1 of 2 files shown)


mlx/backend/cpu/scan.cpp
```diff
@@ -88,30 +102,41 @@ void strided_scan(
     bool reverse,
     bool inclusive,
     const Op& op,
-    U init) {
+    AccT init) {
   // TODO: Vectorize the following naive implementation
+  // One running accumulator per lane, kept in AccT and narrowed to U on store
+  // (see the note in contiguous_scan); bit-identical when AccT == U.
+  std::vector<AccT> acc(stride);
```
zcbenz, line 109: This is doing an implicit heap allocation that escaped our buffer cache, it should allocate a new temporary array and add it to the command encoder's temporaries.

mlx/backend/cpu/scan.cpp
```diff
@@ -88,30 +102,41 @@ void strided_scan(
...
   if (!reverse) {
     if (inclusive) {
       for (int i = 0; i < count; i++) {
-        std::copy(input, input + stride, output);
+        for (int k = 0; k < stride; k++) {
```
zcbenz, line 113: The code should still use std::copy when `if constexpr (std::is_same_v<T, AccT>)` which would be a lot faster.

zcbenz (changes requested):
Can you also do some benchmarking? I'm mostly worried about the strided_scan getting affected by the extra allocation.

zcbenz:

PR #3909, opened 2026-07-24
Accumulate low-precision float reductions in float32 on the CPU

v-code01:
### Problem

`Sum` / `Prod` reductions of `float16` / `bfloat16` accumulate in the input dtype (`reduction_op<InT, InT, ...>`), so a long reduction saturates the accumulator:

```python
with mx.stream(mx.cpu):
    mx.sum(mx.ones((20000,), mx.float16))    # 16384.0  (true 20000, numpy: 20000)
    mx.sum(mx.ones((70000,), mx.bfloat16))   #   256.0  (true 70000)
    mx.mean(mx.ones((70000,), mx.bfloat16))  #     0.0  (true 1.0)
```

The CPU result is both wrong and worse than numpy, which reduces in a wider type. Because `mean` and `var` are built on `sum`, they inherit the error (a large bfloat16 mean of ones returns 0). This is the reduce-path sibling of the scan issue in #3907.

### Fix

Integer reductions already widen to `int32`; do the same for low-precision floats. Reduce into a `float32` accumulator through the existing `reduction_op` machinery — which already supports an accumulator wider than the input (used for `int8`/`int16 -> int32`) — and narrow back to the output dtype once at the end. Every other dtype is untouched, so `float32` / int / complex results stay bit-identical.

```python
with mx.stream(mx.cpu):
    mx.sum(mx.ones((20000,), mx.float16))    # 20000.0
    mx.sum(mx.ones((70000,), mx.bfloat16))   # 70144.0  (nearest bfloat16)
    mx.mean(mx.ones((70000,), mx.bfloat16))  #     1.0
```

A `float16` sum still overflows to `inf` once the true value exceeds float16's ~65504 max — that is unavoidable for a `float16` output — but within range it now matches numpy, and `mean`/`var` are correct.

### Tests

Added `test_low_precision_float_accumulation` (float16/bfloat16 `sum`/`mean`/`var`, all-reduce and a strided axis reduction), pinned to the CPU stream. It fails on the current backend (stall/saturation) and passes with the fix. Full `test_reduce.py` + `test_ops.py` (157 tests / 5371 subtests) stay green; float32/int reductions remain bit-identical to numpy.

Verified with a CPU-only build (`MLX_BUILD_METAL=OFF`) on an M4.

### DIFF
--- a/mlx/backend/cpu/reduce.cpp
+++ b/mlx/backend/cpu/reduce.cpp
@@ -410,21 +410,49 @@ void reduce_dispatch_and_or(
   }
 }
 
+// Reduce into a float32 accumulator and narrow back to the low-precision output
+// dtype. A float16/bfloat16 accumulator saturates on a long reduction the same
+// way a scalar chain would (e.g. a float16 sum of 70000 ones overflows to inf
+// and a bfloat16 one stalls at 256), which also propagates into mean/var. This
+// keeps the reduction in float32 and only narrows once at the end.
+template <typename InT, typename Op>
+void reduce_float_in_fp32(
+    const array& in,
+    array& out,
+    Op,
+    const std::vector<int>& axes,
+    float init) {
+  array acc(out.shape(), float32, nullptr, {});
+  acc.set_data(allocator::malloc(acc.nbytes()));
+  reduction_op<InT, float, Op>(in, acc, axes, init);
+  auto acc_ptr = acc.data<float>();
+  auto out_ptr = out.data<InT>();
+  for (size_t i = 0; i < out.size(); i++) {
+    out_ptr[i] = static_cast<InT>(acc_ptr[i]);
+  }
+}
+
 template <typename InT>
 void reduce_dispatch_sum_prod(
     const array& in,
     array& out,
     Reduce::ReduceType rtype,
     const std::vector<int>& axes) {
+  constexpr bool is_low_precision_float =
+      std::is_same_v<InT, float16_t> || std::is_same_v<InT, bfloat16_t>;
   if (rtype == Reduce::Sum) {
     if constexpr (std::is_integral_v<InT> && sizeof(InT) <= 4) {
       reduction_op<InT, int32_t, SumReduce>(in, out, axes, 0);
+    } else if constexpr (is_low_precision_float) {
+      reduce_float_in_fp32<InT>(in, out, SumReduce{}, axes, 0.0f);
     } else {
       reduction_op<InT, InT, SumReduce>(in, out, axes, 0);
     }
   } else {
     if constexpr (std::is_integral_v<InT> && sizeof(InT) <= 4) {
       reduction_op<InT, int32_t, ProdReduce>(in, out, axes, 1);
+    } else if constexpr (is_low_precision_float) {
+      reduce_float_in_fp32<InT>(in, out, ProdReduce{}, axes, 1.0f);
     } else {
       reduction_op<InT, InT, ProdReduce>(in, out, axes, 1);
     }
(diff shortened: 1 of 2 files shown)


jagrit06:
Thanks for putting this together 
Have you had a chance to run the CPU reduction benchmarks to see if the change has any implications on performance ?

v-code01:
Thanks! Yes, I benchmarked CPU reductions before and after on an M4 (`MLX_BUILD_METAL=OFF`, same `time_fn` methodology: 5 warmup + 200 timed iterations, ms per call). `f16`/`bf16`/`f32`, a 4M-element all-reduce plus 4096x4096 last-axis (contiguous), first-axis (strided), and `mean`:

| case | main | this PR | ratio |
|---|---:|---:|---:|
| bf16 sum, all (4M) | 19.32 | 2.42 | **8.0x faster** |
| bf16 sum, axis=-1 | 76.76 | 9.22 | **8.3x faster** |
| bf16 mean, axis=-1 | 76.98 | 9.22 | **8.3x faster** |
| bf16 sum, axis=0 (strided) | 3.51 | 1.22 | **2.9x faster** |
| f16 sum, all (4M) | 0.33 | 0.38 | 1.15x slower |
| f16 sum, axis=-1 | 1.09 | 1.34 | 1.23x slower |
| f16 sum, axis=0 (strided) | 0.68 | 1.27 | 1.86x slower |
| f16 mean, axis=-1 | 1.10 | 1.34 | 1.22x slower |
| f32 (all cases) | 0.35 / 1.20 / 1.53 | unchanged | 1.00x |

Two takeaways:

- **bfloat16 gets 3-8x faster.** Apple Silicon has no native SIMD arithmetic for bfloat16, so the old in-dtype accumulation fell back to a slow scalar-ish path; accumulating in float32 uses the hardware float SIMD, so it is both correct and much faster. This is the common low-precision dtype, so it is a large net win.
- **float16 has a small slowdown** from the extra float32 scratch buffer and the narrowing pass (float16 *does* have native SIMD, so widening costs a bit without a compute win). Worst on the strided axis=0 case, still sub-2ms for a 16M-element reduction. float32 and int paths are untouched and measure unchanged.

If you would like to trim the float16 overhead, I am happy to drop the scratch array for the all-reduce (scalar-output) case, or gate the widening more narrowly. Let me know which direction you prefer.

mlx/backend/cpu/reduce.cpp
```diff
@@ -410,21 +410,49 @@ void reduce_dispatch_and_or(
...
+    Op,
+    const std::vector<int>& axes,
+    float init) {
+  array acc(out.shape(), float32, nullptr, {});
+  acc.set_data(allocator::malloc(acc.nbytes()));
+  reduction_op<InT, float, Op>(in, acc, axes, init);
+  auto acc_ptr = acc.data<float>();
+  auto out_ptr = out.data<InT>();
+  for (size_t i = 0; i < out.size(); i++) {
+    out_ptr[i] = static_cast<InT>(acc_ptr[i]);
```
zcbenz, line 431: Can you try using std::copy which is cleaner and might generate faster code:
```
std::copy_n(acc.data<float>(), out.size(), out.data<InT>());
```

zcbenz (commented):
There were some similar conversations before and our conclusion was that we should only do high-precision reduction when the implementation does not hurt performance https://github.com/ml-explore/mlx/pull/3488#issuecomment-4404747230, since this implementation simply allocates a temporary array I think this is probably not the ideal implementation that we would like.

Though for bfloat16 this apparently improves performance a lot so I think I'm +1 on doing it for bfloat16 only.

zcbenz:

PR #3913, opened 2026-07-24
[Metal] Faster logsumexp for short rows via a simdgroup-per-row kernel

qx555:
One simdgroup per row (8 rows per 256-thread threadgroup) with the online max/normalizer rescaling already used by logsumexp_looped: reductions never leave the simdgroup, so the kernel needs no threadgroup memory and no barriers (the block kernel uses 5). Dispatched for axis_size <= 2048; the block kernel is unchanged and still serves 2048 < axis_size <= 4096.

On M4 Max: fp16 up to 2.8x, bf16 up to 2.6x, fp32 up to 1.5x on affected sizes; parity at the 2048 boundary and no change on block/looped paths.

### DIFF
--- a/mlx/backend/metal/jit_kernels.cpp
+++ b/mlx/backend/metal/jit_kernels.cpp
@@ -320,6 +320,8 @@ MTL::ComputePipelineState* get_logsumexp_kernel(
     kernel_source +=
         get_template_definition("block_" + lib_name, "logsumexp", t_str);
     kernel_source += get_template_definition(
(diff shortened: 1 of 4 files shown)


zcbenz (commented):
The chance of this PR being actually good is low: added logsumexp kernel but no softmax, no benchmarking code, no tests, no mention which real world model would be helped. So I don't think we would allocate time to review this PR and I'm closing it.

PR #3919, opened 2026-07-25
Fuse the causal mask into the SDPA fallback softmax

spokvulcan:
Long-context prefill takes the SDPA ops fallback for shapes the fast kernels don't cover. With a causal mask, that chain is greater_equal -> where(finfo.min) -> softmax(precise): at 32K it materializes 512 MB of masked scores plus a 32 MB bool mask per layer chunk, all to feed a softmax that immediately reduces over the same data. The chain costs about 6.5 ms per chunk, of which the mask build and the where are about 3 ms.

This folds the causal condition into the looped softmax kernel: the same precise reduction tree with N_READS=4, with the causal select applied at the load sites. Scores get read once and probabilities written once, and the two mask passes plus the giant intermediate disappear.

The result is bit-identical to the chain it replaces, not approximately. The masked value is exactly bf16 finfo.min upcast to f32, and exp(min - max) underflows to exact zero, so per-element arithmetic is unchanged. Verified bitwise on 14 of 14 configurations (3 seeds, sequence lengths 8192 and 32768, several mask offsets), and token-identical over 48 of 48 interleaved end-to-end A/B pairs on a 35B model. A test in test_fast_sdpa.py forces the fallback and asserts exact equality against the reference chain, so the bit-identity claim is enforced in CI rather than taken on faith.

The gate is deliberately conservative: causal only, no array mask, no sinks, bf16, softmax axis > 4096, GPU stream with Metal available, and never inside a function transform, so grad, vmap, and compile keep the differentiable stock chain. Anything outside the gate takes the stock chain exactly as before.

Measured: the mask+softmax stage drops 45% at 32K and 61% at 8K, which lands on the softmax's memory-traffic floor. End to end that is +2.8% prefill at 32K on Qwen3.6-35B-A3B (M3 Max), plus the memory-pressure relief from never allocating the masked-scores tensor.

The kernel body is deliberately a frozen copy of softmax_looped's body with the select applied at the load sites, since that is what the bitwise verification covered. If you would rather see this as a causal template parameter on softmax_looped itself (or a proper primitive with per-backend gating), happy to rework it that way as a follow-up and re-verify.

### DIFF
--- a/mlx/backend/metal/kernels/defines.h
+++ b/mlx/backend/metal/kernels/defines.h
@@ -12,6 +12,7 @@ static MTL_CONST constexpr int MAX_REDUCE_SPECIALIZED_DIMS = 4;
 static MTL_CONST constexpr int REDUCE_N_READS = 4;
 static MTL_CONST constexpr int REDUCE_N_WRITES = 4;
 static MTL_CONST constexpr int SOFTMAX_N_READS = 4;
+static MTL_CONST constexpr int SOFTMAX_LOOPED_LIMIT = 4096;
 static MTL_CONST constexpr int RMS_N_READS = 4;
 static MTL_CONST constexpr int RMS_LOOPED_LIMIT = 4096;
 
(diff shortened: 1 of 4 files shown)


zcbenz (changes requested):
The implementation simply ignores how the code is organized in the project.

PR #3920, opened 2026-07-25
Faster eval walk and cached gather_mm identity indices

spokvulcan:
LLM decode evaluates a graph of a few thousand nodes per token, and in an M3 Max profile the CPU side of eval was a visible slice of the generation thread. This trims it in two independent places.

The first is eval_impl. Building the degree counts and the tape does several hash map operations per graph edge on every eval, and the unordered_map node allocations dominated that walk. The map is now a small open-addressing table. It is probed by key and never iterated, so the tape order it produces is unchanged. The tape loop also stops redoing the stream, event, and fence bookkeeping for every node while consecutive nodes run on one stream, which is the whole loop in the common single-stream case.

The second is gather_mm and gather_qmm. When no lhs indices are passed they rebuild the identity row indices (an arange plus a reshape) on every call. An MoE layer only passes rhs expert ids, so a 256-expert model rebuilds them a few hundred times per decoded token. The evaluated identity array is now cached per shape. The cache only ever holds evaluated arrays, so every consumer sees an immutable constant leaf. Keys are exact shapes, the cache is bounded and mutex guarded, and tracing bypasses it so function transforms never capture a cached leaf.

Numerics are untouched. Output is token-identical over 18 of 18 and 20 of 20 interleaved A/B pairs on a 4B dense and a 35B MoE model. Measured on the MoE model, decode is +2.0% at short context and +2.8% at 8K, prefill is +3.6% at 8K from the index cache, and peak memory is exactly flat.

Happy to split this into two PRs if the map and the cache are easier to review separately.

### DIFF
--- a/mlx/ops.cpp
+++ b/mlx/ops.cpp
@@ -5,10 +5,13 @@
 #include <algorithm>
 #include <climits>
 #include <cmath>
+#include <mutex>
 #include <numeric>
 #include <set>
 #include <sstream>
+#include <unordered_map>
 
+#include "mlx/allocator.h"
 #include "mlx/backend/cuda/cuda.h"
 #include "mlx/backend/metal/metal.h"
 #include "mlx/fast_primitives.h"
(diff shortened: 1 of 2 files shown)


zcbenz (changes requested):
Not saying cache does not help, but the implementation is dirty enough that I can't even suggest changes.

PR #3923, opened 2026-07-26
Add BitLinear: BitNet b1.58 quantization-aware training layer

peterlodri-sec:
## Summary
Adds `nn.BitLinear`, implementing BitNet b1.58 ([arXiv:2402.17764](https://arxiv.org/abs/2402.17764)) ternary weight quantization (plus a provably-binary variant) as a quantization-*aware training* layer, so BitNet-style models can be trained natively in MLX.

**Honestly scoped, read this before reviewing:** this is *not* a packed/memory-efficient quantized format like the existing `QuantizedLinear`/`QQLinear` (backed by real C++/Metal kernels via `mx.quantize`). `BitLinear` keeps full-precision weights and re-quantizes them on every forward pass via a straight-through estimator — forward sees the quantized value, `mx.stop_gradient` makes backward treat that as identity so gradients still reach the full-precision weight the optimizer actually updates. It does **not** reduce memory or improve inference speed the way a packed ternary format with a fused Metal kernel would. A real benchmark (`benchmarks/python/bitlinear_bench.py`) confirms this plainly: `BitLinear` is 1.2–1.8x *slower* than `nn.Linear` at dim=768/4096 on CPU, not faster — expected, since it does strictly more compute (an RMSNorm plus two quantization passes) with zero compute/memory savings from actual bit-packing yet. That packed format is real, separate, harder follow-on work; this PR lays the groundwork with something that already works for training today, not a claim about inference speed it doesn't back.

## What's in this PR
- `python/mlx/nn/layers/bitlinear.py`:
  - `weight_quant(w)`: `sign(w - mean(w)) * mean(|w|)`, nominally ternary per the paper. Measured against a real trained BitNet b1.58 checkpoint: `sign()` on a continuous float essentially never lands on exact zero (9 zeros out of 162,129,408 weight elements measured — 0.0000055%), so this is ~binary in practice, not truly ternary.
  - `binary_weight_quant(w)`: `where(w >= mean(w), scale, -scale)`. Provably 1-bit — the boundary case is deterministic, zero is never possible by construction (not just empirically rare like `weight_quant`).
  - `activation_quant(x)`: per-token simulated int8 quantization, same as the original BitNet paper's activation path.
  - `BitLinear(Module)`: `RMSNorm` + straight-through-estimated weight/activation quantization + matmul. `binary=True` switches to `binary_weight_quant`.
  - Every hot-path function follows this file's existing performance idiom (`@partial(mx.compile, shapeless=True)`, matching `activations.py`), including a dedicated `_quantize_ste` helper so the subtract/stop_gradient/add sequence fuses into one dispatched kernel instead of three separate ops.
- `python/tests/test_bitlinear.py`: 7 tests covering quantization correctness (binary is provably two-valued and never zero; ternary matches binary except at measure-zero ties), forward shape (with/without bias, both quant modes), gradient flow to the full-precision weight, and repr.
- `benchmarks/python/bitlinear_bench.py`: real timing vs `nn.Linear`, following the existing `benchmarks/python/` convention (`time_utils.time_fn`, `--gpu` flag).
- Registered in `mlx.nn.layers.__init__` and `docs/src/python/nn/layers.rst` alongside the other layers (alphabetical placement).

## Test plan
- [x] `python/tests/test_bitlinear.py` — 7/7 pass, verified against a real MLX runtime (installed the published `mlx` wheel in a clean venv, dropped this file into the installed package tree, ran it for real — not just written and assumed correct)
- [x] Real forward + `nn.value_and_grad` check outside the test file too: gradient reaches `weight` with the correct shape and nonzero values, confirming the STE actually works
- [x] `benchmarks/python/bitlinear_bench.py` — real numbers, reported honestly above
- [x] `black --check` — clean on all three new files
- [ ] `pre-commit run --all-files` (clang-format n/a, no C++ changes) / full `swift test`-equivalent CI — happy to address any CI feedback

## Not in this PR (real, separate follow-up work)
A packed ternary storage format with a fused Metal kernel, giving actual memory and inference-speed benefits the way `QuantizedLinear` does for int4/int8. That's genuine new low-level C++/Metal work, not something to fake a quick version of here.

### DIFF
--- a/benchmarks/python/bitlinear_bench.py
+++ b/benchmarks/python/bitlinear_bench.py
@@ -1,0 +1,44 @@
+# Copyright © 2026 8b-is
+
+import argparse
+
+import mlx.core as mx
+import mlx.nn as nn
+from time_utils import time_fn
+
+from mlx.nn.layers.bitlinear import BitLinear
+
+B = 32
+T = 256
+
+
+def time_bitlinear(dim):
+    mx.random.seed(3)
+    x = mx.random.normal((B, T, dim))
+    linear = nn.Linear(dim, dim)
+    ternary = BitLinear(dim, dim)
+    binary = BitLinear(dim, dim, binary=True)
+    mx.eval(x, linear.parameters(), ternary.parameters(), binary.parameters())
+
+    time_fn(linear, x, msg=f"nn.Linear(dim={dim})")
+    time_fn(ternary, x, msg=f"BitLinear(dim={dim}, binary=False)")
+    time_fn(binary, x, msg=f"BitLinear(dim={dim}, binary=True)")
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser("MLX benchmarks.")
+    parser.add_argument("--gpu", action="store_true", help="Use the Metal back-end.")
+    args = parser.parse_args()
+    if args.gpu:
+        mx.set_default_device(mx.gpu)
+    else:
+        mx.set_default_device(mx.cpu)
+
+    # BitLinear does strictly more work per forward pass than nn.Linear (an
+    # RMSNorm plus two quantization passes, all still in full precision --
+    # this is a quantization-*aware-training* layer, not a packed/memory-
+    # efficient format). Expect it to be slower, not faster; that is the
+    # honest baseline a future packed ternary format should be measured
+    # against, not a claim this benchmark is trying to win.
+    for dim in (768, 4096):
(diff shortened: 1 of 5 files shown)


zcbenz:
Thanks for the pull request! Unfortunately due to maintenance burden we do not plan to support every quantization method, as for BitNet b1.58 we currently don't consider it adopted widely enough to have builtin support.

PR #3929, opened 2026-07-27
Fix CUDA batched GEMV grid overflow

### LINKED ISSUE #3858: [BUG] [CUDA] Batched matrix-vector product crashes when batch > 65,535 (gridDim limit)

jasp-nerd:
## Proposed changes

Fixes #3858.

CUDA batched GEMV used `grid.y` as a linear batch index, so valid batches above
65,535 exceeded the CUDA grid-dimension limit before kernel execution.
`gemv_gather` used the same launch geometry.

This change splits the batch dimension over `grid.y` and `grid.z`, reconstructs
the logical batch index in both kernels, and ignores the one padded slot when
the rectangular grid is not exact. It also keeps empty gathered outputs from
reaching a zero-sized launch.

The regression tests cover ordinary and gathered GEMV at batch sizes 65,535,
65,536, and 65,537 using small analytical inputs, plus an empty gather.

Validation:

```text
pre-commit hooks on changed files: passed
Python test syntax: passed
independent grid-mapping and value-oracle checks: passed
CPU C++ suite: 229 test cases, 3,070 assertions passed
```

This machine does not have a CUDA toolkit or NVIDIA GPU, so CUDA compilation
and execution are left to the repository CI matrix.

## Checklist

- [x] I have updated the necessary documentation (not needed for this backend-only fix)

### DIFF
--- a/mlx/backend/cuda/gemms/gemv.cu
+++ b/mlx/backend/cuda/gemms/gemv.cu
@@ -13,6 +13,14 @@ namespace cg = cooperative_groups;
 
 static constexpr int rows_per_block = 8;
 
+// Split batches across grid.y/z because both dimensions are capped at 65,535.
+static dim3 get_gemv_grid_dims(uint32_t num_blocks_x, uint32_t batch_size) {
+  constexpr uint32_t max_grid_yz_dim = 65535;
+  uint32_t num_blocks_z = cuda::ceil_div(batch_size, max_grid_yz_dim);
+  uint32_t num_blocks_y = cuda::ceil_div(batch_size, num_blocks_z);
+  return {num_blocks_x, num_blocks_y, num_blocks_z};
+}
+
 // Accumulator type selection per input element type T.
 template <typename T>
 struct GemvAccType {
@@ -88,12 +96,17 @@ __global__ void gemv_batched(
     T* out,
     int rows,
     int cols,
+    uint32_t batch_size,
     const __grid_constant__ Shape batch_shape,
     const __grid_constant__ Strides mat_batch_strides,
     const __grid_constant__ Strides vec_batch_strides,
     int batch_ndim) {
-  auto block = cg::this_thread_block();
-  auto batch_idx = block.group_index().y;
+  auto grid = cg::this_grid();
+  auto batch_idx =
+      grid.block_index().z * grid.dim_blocks().y + grid.block_index().y;
+  if (batch_idx >= batch_size) {
+    return;
+  }
   auto [vec_offset, mat_offset] = elem_to_loc(
       batch_idx,
       batch_shape.data(),
@@ -113,6 +126,7 @@ __global__ void gemv_gather(
     uint32_t* vec_indices,
     int rows,
     int cols,
+    uint32_t batch_size,
     const __grid_constant__ Shape mat_batch_shape,
     const __grid_constant__ Strides mat_batch_strides,
     int mat_batch_ndim,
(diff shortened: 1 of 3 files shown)


zcbenz (commented):
Sorry but we do not review code that never got tested.

sashko-zakharchuk:
@jasp-nerd I have an RTX 5050 (sm_120) and verified this patch on it (CUDA 12.8), since it was closed for lack of a hardware run.

It applies cleanly on current main. Before the patch, `mx.matmul(mx.zeros((65536, 225, 4)), mx.zeros((65536, 4, 1)))` aborts with `cudaGraphAddKernelNode ... invalid argument` (with graphs off, `cudaLaunchKernelExC ... invalid configuration argument`), while `65535` and a square `(65536, 8, 8) @ (65536, 8, 8)` are fine. With the patch, batch 65535 / 65536 / 65537 / 200000 all run and match the CPU result for both `matmul` and `gather_mm`, and the empty `gather_mm` case returns an empty array instead of crashing. Both added tests pass.

Would it be possible to reopen this? Happy to help if anything else is needed.

jasp-nerd:
Thanks so much for testing this, @sashko-zakharchuk! @zcbenz, would this hardware verification be enough to reopen the PR?

zcbenz (approved):
LGTM

PR #3931, opened 2026-07-27
Fix Metal column reduction grid overflow

### LINKED ISSUE #3784: mx.sum over a non-last axis returns all zeros for certain large tensors (e.g. [1, 32, 4096, 131072])
## Summary

`mx.sum(x, axis=1)` returns all zeros instead of the correct sum for `x` of shape `[1, 32, 4096, 131072]`. It happens deterministically, for every dtype (float32, float16, bfloat16). The same reduction over the last axis is correct, and smaller shapes are correct.

## Repro

```python
import mlx.core as mx

x = mx.ones((1, 32, 4096, 131072), dtype=mx.float32)
y = mx.sum(x, axis=1)
mx.eval(y)
print(float(y.min()), float(y.max()))   # prints 0.0 0.0, expected 32.0 32.0
```

100% reproducible (5 of 5 fresh runs). Same result with `bfloat16` and `float16`.

## What I found

Reducing the last axis is fine: `mx.sum(x, axis=3)` on the same tensor returns the correct `131072.0`. Only the non last axis reduction is affected.

It is size gated. `[1, 32, 2048, 131072]` through `[1, 32, 4032, 131072]` are all correct. It first breaks at `[1, 32, 4096, 131072]` (2^34 elements). Just above that boundary the output is partially zeroed (for example `[1, 32, 4352, 131072]` gives `min 0, max 32`), which looks like a 32 bit index overflow in the strided reduction (2^34 is 4 times 2^32).

It is specific to a reduction code path, not a generic element count threshold. At the same 2^34 total, `[1, 16, 8192, 131072]`, `[1, 64, 2048, 131072]`, and `[1, 128, 1024, 131072]` are all correct. Only the shapes that reduce 32 elements with a large inner size break, e.g. `[1, 32, 4096, 131072]`, `[1, 32, 8192, 65536]`, `[1, 32, 16384, 32768]`.

## Expected vs actual

Expected: the correct per position sum (`32.0` for the ones tensor above).
Actual: all zeros, or partially zeros just above the boundary.

## Workaround

Reducing with elementwise adds (`x[:, 0:1] + x[:, 1:2] + ...`) or moving the reduced axis to the end both give the correct result.

## Environment

MLX `0.32.0.dev20260609+cc3f3e60`
Apple M3 Ultra, macOS 26.5.1 (build 25F80)

jasp-nerd:
## Proposed changes

Fixes #3784.

The affected Metal column-reduction launches express the x dimension as a
total thread count:

```text
256 * ceil(reduction_stride / 32)
```

For the reported reduction stride of `2^29`, that becomes `2^32` and overflows
the dispatch dimension. The kernels consume threadgroup coordinates, and the
thread count is always an exact multiple of the 256-thread group size.

This change expresses the four affected launches directly as threadgroup
counts. It preserves the kernel-visible grid for ordinary inputs while avoiding
the overflowing multiplication. The shared grid-size helper uses
quotient/remainder arithmetic so it remains valid at `SIZE_MAX`, and its test
covers the reported boundary without allocating the 32+ GiB reproducer.

Validation:

```text
focused grid-size test: 1 test case, 8 assertions passed
CPU C++ suite: 245 test cases, 3,246 assertions passed
host syntax check for metal/reduce.cpp: passed
pre-commit hooks on changed files: passed
```

The full Metal compiler is unavailable on this machine, and the original
large-memory runtime reproducer was not run. Repository CI and validation on
the reporter's large-memory Metal host remain the final runtime gates.

## Checklist

- [x] I have updated the necessary documentation (not needed for this backend-only fix)

### DIFF
--- a/mlx/backend/common/reduce.h
+++ b/mlx/backend/common/reduce.h
@@ -46,6 +46,11 @@ struct ReductionPlan {
   ReductionPlan(ReductionOpType type_) : type(type_) {}
 };
 
+inline size_t col_reduce_grid_size(size_t reduction_stride, size_t block_size) {
+  return reduction_stride / block_size +
+      static_cast<size_t>(reduction_stride % block_size != 0);
+}
+
 ReductionPlan get_reduction_plan(const array& x, const std::vector<int>& axes);
 
 std::pair<Shape, Strides> shapes_without_reduction_axes(
--- a/mlx/backend/metal/reduce.cpp
+++ b/mlx/backend/metal/reduce.cpp
@@ -710,7 +710,7 @@ void strided_reduce_longcolumn(
   second_args.reduce_strides.push_back(out.size());
   second_args.reduce_ndim++;
   int BN = 32;
-  grid_dims = MTL::Size(256 * ((out.size() + BN - 1) / BN), 1, 1);
+  grid_dims = MTL::Size(col_reduce_grid_size(out.size(), BN), 1, 1);
   group_dims = MTL::Size(256, 1, 1);
 
   // Set the 2nd kernel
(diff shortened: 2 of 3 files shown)


zcbenz (changes requested):
The repro in original issue still fails with this PR.

PR #3933, opened 2026-07-27
Fix crashes in the ring and jaccl distributed backends

jzdziarski:
## Proposed changes

Three unrelated stability issues in the CPU-side distributed backends, all of which show up as SIGSEGVs on long-running multi-node jobs.

Use-after-free in the collectives. `Encoder::dispatch` only enqueues the lambda, and `set_input_array`/`set_output_array` are no-ops on the CPU backend, so nothing keeps the arrays' buffers alive until the task runs. `cpu::eval` queues a keep-alive holding the inputs, but excludes the output, which the collectives allocate themselves. Under memory pressure the allocator can reclaim and reuse the output block before the collective runs, and the stream thread then writes through a stale pointer. Capture `data_shared_ptr()` alongside the raw pointers in the jaccl and ring collectives so the buffers are pinned for the task's lifetime.

Unvalidated work completions in the jaccl ring. A failed or spurious completion carries an undefined `wr_id`, and the wire/buff decoded from it was used unchecked to index the per-wire counters, offsets, limits and the buffer pools. Out-of-range values produced a wild pointer in `reduce_op`/`std::copy`. Check `wc.status` and range-check the decoded indices, then drop the completion and keep draining. The in-flight counter is decremented before the check so the loop still terminates.

32-bit offset overflow. `read_offset`/`write_offset` in the mesh all_gather and `n_steps` in the ring all_gather were `int` while indexing payloads sized by `int64_t`, wrapping past 2GB and ~4GB per rank respectively. Widen them to `int64_t`.

## Checklist

NOTE: Only reproducible around 420GB, so was not able to write a test for this fix.

### DIFF
--- a/mlx/distributed/jaccl/lib/jaccl/ring_impl.h
+++ b/mlx/distributed/jaccl/lib/jaccl/ring_impl.h
@@ -2,6 +2,7 @@
 
 #pragma once
 
+#include <iostream>
 #include <span>
 
 #include "jaccl/rdma.h"
@@ -130,13 +131,25 @@ class RingImpl {
         ibv_wc wc[WC_NUM];
         int n = poll(left_, right_, WC_NUM, wc);
         for (int i = 0; i < n; i++) {
+          in_flight--;
+
+          // A failed or spurious completion carries an undefined wr_id, so
+          // the decoded wire and buff would index the counters and buffer
+          // pools out of bounds. Drop it and keep draining; the in-flight
+          // slot is already retired above.
           int work_type = wc[i].wr_id >> 16;
           int buff = (wc[i].wr_id >> 8) & 0xff;
           int wire = wc[i].wr_id & 0xff;
           int lr = wire / RING_MAX_CONNS;
           int lw = wire % RING_MAX_CONNS;
-
-          in_flight--;
+          if (wc[i].status != IBV_WC_SUCCESS || lr >= MAX_DIR ||
+              lw >= n_wires || buff >= PIPELINE) {
+            std::cerr << IBV_TAG << " rank " << rank_
+                      << ": dropped bad all_reduce completion (status="
+                      << wc[i].status << ", wr_id=0x" << std::hex << wc[i].wr_id
+                      << std::dec << ")" << std::endl;
+            continue;
+          }
 
           if (work_type == SEND_WR && send_count[wire] < n_steps) {
             int64_t offset = lw * N + send_count[wire] * n_wires * N +
@@ -229,13 +242,25 @@ class RingImpl {
         ibv_wc wc[WC_NUM];
         int n = poll(left_, right_, WC_NUM, wc);
         for (int i = 0; i < n; i++) {
+          in_flight--;
+
+          // A failed or spurious completion carries an undefined wr_id, so
+          // the decoded wire and buff would index the counters and buffer
+          // pools out of bounds. Drop it and keep draining; the in-flight
+          // slot is already retired above.
           int work_type = wc[i].wr_id >> 16;
           int buff = (wc[i].wr_id >> 8) & 0xff;
           int wire = wc[i].wr_id & 0xff;
           int lr = wire / RING_MAX_CONNS;
           int lw = wire % RING_MAX_CONNS;
-
-          in_flight--;
+          if (wc[i].status != IBV_WC_SUCCESS || lr >= MAX_DIR ||
+              lw >= n_wires || buff >= PIPELINE) {
+            std::cerr << IBV_TAG << " rank " << rank_
+                      << ": dropped bad all_reduce completion (status="
+                      << wc[i].status << ", wr_id=0x" << std::hex << wc[i].wr_id
+                      << std::dec << ")" << std::endl;
+            continue;
+          }
 
           if (work_type == SEND_WR && send_count[wire] < n_steps) {
             int64_t offset = lw * N + send_count[wire] * n_wires * N +
@@ -306,7 +331,9 @@ class RingImpl {
     size_t n_bytes_per_wire = (n_bytes + (2 * n_wires) - 1) / (2 * n_wires);
     size_t out_bytes = n_bytes * size_;
     auto [sz, N] = buffer_size_from_message(n_bytes_per_wire);
-    int n_steps = (n_bytes_per_wire + N - 1) / N;
+    // 64-bit to match all_reduce above: the offsets derived from the step
+    // count wrap an int once the per-rank payload reaches ~4GB.
+    int64_t n_steps = (n_bytes_per_wire + N - 1) / N;
 
     // Counters to maintain the state of transfers
     int in_flight = 0;
@@ -354,13 +381,22 @@ class RingImpl {
         ibv_wc wc[WC_NUM];
         int n = poll(left_, right_, WC_NUM, wc);
         for (int i = 0; i < n; i++) {
+          in_flight--;
+
+          // Same guard as all_reduce. all_gather is always 2-directional.
           int work_type = wc[i].wr_id >> 16;
           int buff = (wc[i].wr_id >> 8) & 0xff;
           int wire = wc[i].wr_id & 0xff;
           int lr = wire / RING_MAX_CONNS;
           int lw = wire % RING_MAX_CONNS;
-
-          in_flight--;
+          if (wc[i].status != IBV_WC_SUCCESS || lr >= 2 || lw >= n_wires ||
+              buff >= PIPELINE) {
+            std::cerr << IBV_TAG << " rank " << rank_
+                      << ": dropped bad all_gather completion (status="
+                      << wc[i].status << ", wr_id=0x" << std::hex << wc[i].wr_id
+                      << std::dec << ")" << std::endl;
+            continue;
+          }
 
           if (work_type == SEND_WR && send_count[wire] < n_steps) {
             int64_t offset = lw * N + send_count[wire] * n_wires * N +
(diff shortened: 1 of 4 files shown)


guruswami-ai:
**Controlled A/B on this PR's exact base at ~420 GB/rank: the lifetime capture did not prevent the failure.**

You note the problem only reproduced around 420 GB. We run at **420.8 GB/rank**,
so we built a paired experiment against your branch. The result is negative, and
since this PR has no automated reproduction I think that is more useful to you
than another untested pass.

### Setup

- 4× Mac Studio M3 Ultra 512 GB, macOS 26.6 (25G72), Thunderbolt 5 full mesh, JACCL/RDMA
- Kimi-K3 (2.78 T, MXFP4) tensor-parallel across 4 ranks, **420.8 GB/rank**, 445.7 GB peak
- `MLX_METAL_FAST_SYNCH=0`, `prefill_step_size=512`, prompt cache size 1 with a 24 GiB byte cap
- Base commit `973e27f82ffe68dbd626cda31ba34997045d1eb7`; PR head `ae7d19bf9e0a228317b575e854d5dc2a349d4af1`

Both arms differ **only** in the two MLX libraries. Same model files, same
scripts, same soak workload and ordering, fresh reboot and verified-quiescent
fleet before each arm, interface/MTU symmetry verified and recorded.

| | control | candidate |
|---|---|---|
| contents | pinned base, clean | base + #3933 + #3152 (mesh `check_wc_status` only) |
| `libmlx.dylib` | `6abacfb3376b94d27613e4da7d4083a15d592cc1144b072d6cbb5282b9ff2fe5` | `248decd0b2acf8bf0afbd0ffca081a6e4dadd9c75fe62c3f1fd41cee33270d71` |
| `libjaccl.dylib` | `8ee82aae82996938435bcebe724d819296712b8b34fac021ce7aa66073730e1e` | `1ee43f31dc8732b43697d88b2511400afec2b4142489a48c529df74389ae26f6` |

Note both builds report the identical `mx.__version__`
(`0.32.1.dev20260729+973e27f8`), so we identify variants by binary markers, not
version. We verify the lifetime capture is present by the closure-object
destructor symbol for the dispatched `all_sum` lambda — upstream that closure
captures only raw pointers and is trivially destructible, so no destructor is
emitted; capturing `data_shared_ptr()` forces one:

```
nm -U libmlx.dylib | grep -c \
  'JACCLGroup7all_sumERKNS0_5arrayERS5_NS0_6StreamEENUlvE_D1Ev'
# control 0, candidate 1
```

### Control — SIGSEGV, request 7 of 14

```
SIGSEGV / EXC_BAD_ACCESS
  libthunderboltrdma.dylib  tbt_post_recv
  libjaccl.dylib            jaccl::Connection::post_recv(jaccl::SharedBuffer const&, ...)
  libjaccl.dylib            jaccl::MeshImpl::all_reduce<int, jaccl::SumOp<int>>
  libmlx.dylib              mlx::core::scheduler::StreamThread::thread_fn()
```

No kernel panic and no reboot — the host stayed up throughout; the rank died and
the launcher tore down the survivors. The crash landed **~49 s into a 60 s idle
window, ~11 s before the next request was issued**, on the `int` reduce. A
collective either remained queued from earlier work or was initiated while
nominally idle; our evidence does not distinguish these.

### Candidate — request 12 of 14, MIXED failure modes across ranks

Three ranks aborted after #3152 reported a bad work completion:

All three reported `status=4 (LOC_PROT_ERR)` with an **identical
`vendor_err` and `wr_id`**, across two different queue pairs:

```
qp_num=2320  vendor_err=131331  wr_id=0x20003     x1
qp_num=2336  vendor_err=131331  wr_id=0x20003     x2

SIGABRT   (3 ranks, 17:06:23)
  libjaccl.dylib  jaccl::check_wc_status(ibv_wc const&)
  libjaccl.dylib  jaccl::MeshImpl::all_reduce<jaccl::bfloat16_t, jaccl::SumOp<...>>
  libmlx.dylib    mlx::core::scheduler::StreamThread::thread_fn()
```

A **fourth rank segfaulted ~3 s later** on the same `bfloat16_t` reduce:

```
SIGSEGV   (1 rank, 17:06:26)
  libthunderboltrdma.dylib  tbt_post_recv
  libjaccl.dylib            jaccl::Connection::post_recv
  libjaccl.dylib            jaccl::MeshImpl::all_reduce<jaccl::bfloat16_t, ...>
```

This may be a secondary peer-loss cascade, but **without ordered per-rank
instrumentation we cannot identify the initiating event**.

### What we take from this

1. **The buffer-lifetime capture did not prevent the failure in this workload.**
   The capture is present in the candidate binary and the RDMA work completion
   still failed.
2. The failures are a **strongly matching failure class**, not provably the same
   fault: the control never exposed a work-completion status, so it cannot be
   shown to have been `LOC_PROT_ERR`. What matches is the collective family, the
   RDMA receive path and the scale.
3. **Not confined, in our observations, to a single** node, element dtype or
   load state — seen on two different nodes, on `int` and `bfloat16_t` reduces,
   both idle-associated and under load. Two incidents establish breadth, not
   statistical independence.
4. **No silently incorrect output was observed in either arm.** Every completed
   request returned the exact expected answer, and a 4-fixture wire-level API
   suite passed identically on both builds.

### Limitations, stated plainly

- **Two arms, not three.** The candidate carries **both** #3933 and #3152, so
  this does not isolate #3933 from #3152. It shows the combined build did not
  survive; it cannot attribute the change in presentation.
- It does not show #3933 is wrong in general — only that it is **insufficient
  here**. There may be a second cause, or reclamation on a path the patch does
  not cover.
- `LOC_PROT_ERR` is a local memory-protection violation, which can arise from an
  invalid or stale `lkey`, a deregistered MR, an access-permission violation, or
  an address outside MR bounds. We have not determined which applies.
- We did **not** include #3141; the fleet runs `MLX_METAL_FAST_SYNCH=0`.
- Timings matched between arms to within ~0.5 s per request across the **common
  completed pre-idle requests** — the control never completed the post-idle
  portion, so the comparison covers only the overlap. It was not powered as a
  latency benchmark and does not establish that #3152 is free of cost.

### On #3152

Worth separating from this PR. In the candidate arm #3152 made the bad work
completion attributable **on three ranks**, naming the queue pair, vendor error
and work-request id. A fourth rank still segfaulted in `tbt_post_recv` shortly
afterwards.

We therefore regard #3152 as materially useful **diagnostic / fail-fast coverage
where its check executes** — not as a reliability fix, and not as a guarantee
that every rank fails cleanly. A failed completion still means the collective
did not deliver.

### Happy to provide

We can provide **sanitised/redacted** copies of the control crash report, the
four candidate crash reports, run manifests (commit, variant, library hashes,
hostfile hash, per-node interface snapshot), the hostfile, per-request results
and the build scripts.

Redaction applies to **all** of these, not only the crash reports: the
manifests, hostfile and interface snapshots contain exactly the hostnames,
management addresses and Thunderbolt link-local addresses removed from this
comment, and raw `.ips` files additionally carry process paths, arguments and
machine metadata.

If useful, we can share those artifacts and discuss targeted instrumentation
feasible on this fleet — for example `qp_num` → rank/peer/direction mapping,
`wr_id` decoding, and first-bad-completion capture of opcode, length, buffer
address, MR bounds, `lkey`/`rkey` and registration generation.

guruswami-ai:
**Correction to my earlier comment: that A/B was confounded and its negative conclusion should not be relied on.**

I reported a controlled A/B on this PR's base and concluded the buffer-lifetime capture "did not prevent the failure". Two uncontrolled variables have since come to light on our side. I no longer think the experiment supports that conclusion, and I would rather say so than leave it standing.

### 1. The configuration I stated was wrong

I wrote that both arms ran `MLX_METAL_FAST_SYNCH=0`. They did not. The value was never declared in the hostfile, so each rank inherited it from the login shell — and `/etc/zshenv` exported `=1` on three of the four nodes while the fourth had no such file at all.

So both arms ran an **asymmetric** collective-sync configuration (`1/1/1/unset`), not the uniform `0` I claimed. That is my error, not a subtlety: an undeclared environment variable is an untracked variable.

### 2. A second uncontrolled variable: one node was attempting to sleep

The same odd-one-out node was configured with a 1-minute idle-sleep timer while the others had sleep disabled. It dispatched `systemWillSleep` during both arms, 4–6 s before the first crash in each.

I want to be careful here, because the obvious inference is not supported by the evidence:

- In both arms the host was **awake and logging** thousands of events per second at the crash instants.
- The candidate window does contain a completed sleep/wake cycle (`IOPlatformWakeAction -> AppleThunderboltHALType7`), but at **27 s after** the crash, not before it.
- So the crashes landed *inside* a sleep transition — after drivers were told to quiesce, before the platform actually slept.

That is a plausible mechanism for a `LOC_PROT_ERR` on the RDMA receive path, but it is a hypothesis. I cannot show it caused the failures, and I am not claiming it did.

### What I think is actually true now

- The **negative result should be withdrawn.** An experiment with an asymmetric sync configuration and a node attempting to suspend mid-run cannot distinguish "the patch is insufficient" from "our fleet was misconfigured".
- Nothing here suggests this PR is wrong. It says our test of it was not controlled.
- The observation that the crashes were **idle-associated** — which I offered as evidence — is better explained by an idle-sleep timer firing during idle windows than by anything about the collective.

### What still stands

The failures themselves were real and are unchanged: `LOC_PROT_ERR` with identical `vendor_err` and `wr_id` across two queue pairs, on both `int` and `bfloat16_t` reduces, plus a `tbt_post_recv` segfault. Whatever the trigger, the RDMA path did fail at ~420 GB/rank.

Separately, and independent of any of the above: when a peer disappears, the ring backend does not recover. Surviving ranks block indefinitely in `recvfrom` inside `RingGroup::all_reduce_impl<bfloat16>` and in `SocketThread::worker()` on a condition variable, with no peer-loss timeout. That is reproducible and does not depend on why the peer left.

### Offer

We have since rebuilt the fleet from scratch with the configuration declared per run, sleep disabled and audited, and power telemetry recorded. If a clean three-arm run (base / base+#3933 / base+#3933+#3152) at ~420 GB/rank would be useful, I am happy to do it properly this time and report whatever it shows.

Apologies for the noise.

mlx/distributed/jaccl/lib/jaccl/ring_impl.h
```diff
@@ -130,13 +131,25 @@ class RingImpl {
...
+          // slot is already retired above.
           int work_type = wc[i].wr_id >> 16;
           int buff = (wc[i].wr_id >> 8) & 0xff;
           int wire = wc[i].wr_id & 0xff;
           int lr = wire / RING_MAX_CONNS;
           int lw = wire % RING_MAX_CONNS;
-
-          in_flight--;
+          if (wc[i].status != IBV_WC_SUCCESS || lr >= MAX_DIR ||
+              lw >= n_wires || buff >= PIPELINE) {
```
zcbenz, line 146: Since `wr_id` is undefined on failure, checking lr/lw/buff would not be meaningful?

I think the code should:
1. ensure `wc[i].status` is not `IBV_WC_SUCCESS` before polling;
2. only check `wc[i].status != IBV_WC_SUCCESS` for error.

zcbenz (changes requested):
Do you happen to still have the stack trace of the crashes? It would help diagnose whether the fix is correct.

PR #3934, opened 2026-07-27
metal: reduce NVFP4 scales per 16-lane group

jessegross:
Use the SIMD lane rather than the global grid index when splitting each SIMD group. This prevents every SIMD group after the first from sharing one 32-value maximum across two independently scaled NVFP4 groups.

Wikitext-2 test perplexity, Qwen3.6-35B-A3B quantized to NVFP4 with mlx-lm (297k tokens, Apple M5): 6.848 -> 6.778, bf16 baseline 6.432.


nastya236:
Looks great! Thank you for fixing this bug!

nastya236 (approved):

PR #3937, opened 2026-07-28
Metal allocator: round buffer cache lookups to a size class so growing allocations get reused

PhilipJohnBasile:
The Metal buffer cache (\`reuse_from_cache\`) looks up by exact size via
\`lower_bound\`, so a sequence of monotonically growing allocations (e.g.
repeated concatenate in a loop) misses the cache almost every time even
though the freed buffers are still resident and would fit with trivial
overallocation. Adds \`round_to_size_class()\` (1/16-octave granularity)
applied after page-alignment in \`malloc()\`, clamped to
\`maxBufferLength()\`. Regression test \`test_growing_concatenate_reuses_cache\`
pins the fixed behavior at well above the observed unpatched ~96x blowup.

### DIFF
--- a/mlx/backend/metal/allocator.cpp
+++ b/mlx/backend/metal/allocator.cpp
@@ -7,6 +7,8 @@
 
 #include <mach/vm_page_size.h>
 #include <unistd.h>
+#include <algorithm>
+#include <bit>
 #include <cassert>
 #include <cstdlib>
 
@@ -14,6 +16,35 @@ namespace mlx::core {
 
 constexpr size_t resource_options =
     MTL::ResourceStorageModeShared | MTL::ResourceHazardTrackingModeUntracked;
+
+namespace {
+
+// Round `size` up to a coarse, scale-relative size class for large buffers,
+// so a monotonically GROWING allocation sequence (e.g. per-token KV-cache
+// append built on `concatenate`) reuses the same over-provisioned buffer
+// across several consecutive steps instead of missing BufferCache on every
+// single request. `BufferCache::reuse_from_cache` anchors on
+// `std::multimap::lower_bound(size)`, so it only ever returns a same-or-
+// larger cached buffer -- widening its acceptance window cannot help a
+// strictly growing query sequence, since every already-freed buffer is
+// smaller than the next request by construction. Over-provisioning the
+// actual allocation is the lever that works: within one octave
+// [2^k, 2^(k+1)), granularity is fixed at 2^k/16 (independent of the exact
+// query, so nearby sizes round to the same class), bounding internal
+// fragmentation to <= 1/16 (6.25%) of the class boundary. See
+// https://github.com/ml-explore/mlx/issues/3886.
+constexpr size_t kLargeSizeClassThreshold = 1 << 20; // 1 MiB
+
+size_t round_to_size_class(size_t size) {
+  if (size < kLargeSizeClassThreshold) {
+    return size;
+  }
+  int octave = std::bit_width(size) - 1; // 2^octave <= size < 2^(octave + 1)
+  size_t step = size_t(1) << (octave - 4); // octave width / 16
+  return step * ((size + step - 1) / step);
+}
+
+} // namespace
 
 namespace allocator {
 
@@ -124,6 +155,12 @@ Buffer MetalAllocator::malloc(size_t size) {
   if (size > vm_page_size) {
     size = vm_page_size * ((size + vm_page_size - 1) / vm_page_size);
   }
+  // Never round past the device's hard ceiling -- the size checked against
+  // it above is the pre-rounding value, so a request near the limit must be
+  // clamped rather than over-provisioned.
+  size = std::min(
+      round_to_size_class(size),
+      static_cast<size_t>(device_->maxBufferLength()));
 
   // Try the cache
   std::unique_lock lk(mutex_);
--- a/python/tests/test_memory.py
+++ b/python/tests/test_memory.py
@@ -72,6 +72,40 @@ class TestMemory(mlx_tests.MLXTestCase):
         del a
         self.assertEqual(init_mem, mx.get_active_memory())
 
+    @unittest.skipIf(not mx.metal.is_available(), "Metal is not available")
+    def test_growing_concatenate_reuses_cache(self):
+        # Regression test for #3886: BufferCache.reuse_from_cache anchors on
+        # std::multimap::lower_bound(size), so it only ever returns a
+        # same-or-larger cached buffer. A monotonically GROWING allocation
+        # sequence -- the natural pattern of a per-token KV-cache built on
+        # `concatenate` -- freed a buffer smaller than every subsequent
+        # request, so it could never be reused: every step missed the cache,
+        # and the never-reused buffers piled up in the pool. Measured before
+        # this fix, on this exact pattern: ~2129 MB left in the pool (96x the
+        # final array) after 100 growth steps. Large buffers are now rounded
+        # up to a coarse, scale-relative size class before the cache lookup,
+        # so consecutive small growth steps land on the same over-provisioned
+        # class and actually get reused.
+        mx.synchronize()
+        mx.clear_cache()
+
+        base_rows, step_rows, cols = 5000, 4, 1024
+        a = mx.zeros((base_rows, cols))
+        mx.eval(a)
+
+        n_steps = 100
+        for _ in range(n_steps):
+            a = mx.concatenate([a, mx.zeros((step_rows, cols))], axis=0)
+            mx.eval(a)
+        mx.synchronize()
+
+        # Measured with the fix: ~131 MB (5.9x final_bytes). Measured on the
+        # unpatched allocator with this identical pattern: ~2129 MB (96x).
+        # 20x is a generous margin above the fixed number and decisively
+        # below the unpatched one, so this fails loudly if the reuse
+        # regresses without being sensitive to small measurement noise.
+        self.assertLess(mx.get_cache_memory(), 20 * a.nbytes)
+
 
 if __name__ == "__main__":
     mlx_tests.MLXTestRunner()


angeloskath:
I am not totally against this change, we normally deal with this at the programmer level ie we tend to over allocate manually and slice update into the over allocated buffer (see KV caches in mlx-lm). It offers potentially significantly more benefits including avoiding the full buffer copy for adding one small part to a big buffer.

Do you mind running a comprehensive benchmark suite training/inference and comparing memory use and speeds?

PhilipJohnBasile:
Happy to run it — and your point about the programmer-level pattern is the one I want to address directly, because I think it's right and it changes what the benchmark has to show.

Where the growth site is known, manual over-allocation plus slice-update is strictly better than anything the allocator can do. It avoids the copy entirely; size-class rounding only makes the *reallocation* cheaper to satisfy, it doesn't remove it. So this change should not be read as an alternative to the mlx-lm KV-cache approach.

The population it targets is growth the programmer doesn't control: `concatenate` in a loop inside user code or a library, where there is no single site to over-allocate at. `reuse_from_cache` anchors on `lower_bound(size)`, so a strictly-growing query sequence misses every time by construction — every freed buffer is smaller than the next request. Widening the acceptance window can't fix that; over-provisioning the allocation is the only lever. Bounded at 1/16-octave, worst-case internal fragmentation is 6.25% of the class boundary.

Which means the interesting question isn't whether it helps the pathological case — it's whether a global rounding rule **taxes workloads that don't have the growing pattern**. That's the risk in this patch and it's what I'll measure.

Plan, main vs patched, same machine, fans pinned:

1. **Inference** — mlx-lm generate across a few model sizes and quantizations, short and long context, reporting peak memory and tok/s separated into prefill and decode.
2. **Training** — a real training loop, peak memory and step time.
3. **Control** — workloads with no monotonic growth, to detect a regression from rounding alone. This is the arm that decides it.
4. **The motivating case** — growing `concatenate`, as the upper bound on benefit.
5. **The comparison you implied** — the manual over-allocate + slice-update pattern versus this, on the same workload, so the tradeoff is explicit rather than assumed.

I'll report variance rather than single runs, publish the harness, and state adverse results plainly — if the control arm regresses, that's the finding and this should be closed or gated behind a flag.

One thing I'd like your read on: rounding is currently unconditional above 1 MiB. If the control arm shows a memory cost you consider unacceptable, would you rather see it behind an opt-in (env var or `MetalAllocator` setting), or not at all? That changes what I optimize toward.

PhilipJohnBasile:
Correcting my previous comment — I ran an adversarial review of it against the source and four of my claims were overstated. All four errors ran the same direction: overstating this patch's necessity and understating the alternatives. Worst first.

**1. "It avoids the copy entirely" — wrong about your own suggestion.** `SliceUpdate` calls `copy_gpu` on the *whole source* first ([indexing.cpp:749](https://github.com/ml-explore/mlx/blob/main/mlx/backend/metal/indexing.cpp#L749)), and only skips it when the input is donateable (unique ownership + compatible storage, `common/copy.h`). The update is then still copied into the destination slice. So manual over-allocate + slice-update avoids the growing-prefix copy **in the single-owner donation case**, not unconditionally.

**2. "Strictly better" and "the only lever" — neither absolute holds.** And one alternative is worth more attention than my patch, because MLX already has the machinery for it: a growth-aware `concatenate` could reuse a unique buffer when its *actual* capacity suffices, via `array::buffer_size()` ([array.h:359](https://github.com/ml-explore/mlx/blob/main/mlx/array.h#L359)) plus the existing donation path. That's a targeted fix where the growth happens, rather than a global allocation policy — plausibly the better design. Others that would also work: an arena/suballocator that coalesces adjacent free spans (impossible with standalone `MTL::Buffer` entries, so a different allocator design), a segmented/chunked KV representation that retains earlier buffers instead of needing one contiguous grow, and geometric allocation at the call site.

**3. The 6.25% figure was ambiguous.** The tight bound is `waste < step = 2^k/16`. Against the request or the lower octave boundary that is `<6.25%`; against *allocated* size — the conventional denominator for internal fragmentation — the supremum is below `1/17`, about **5.88%**. I should have said which denominator. Separately, the `maxBufferLength` clamp truncates the top class and breaks regular class spacing near the ceiling; that isn't a correctness bug (the buffer still covers every requested byte) but it's a wart worth noting.

**4. "Misses every time by construction" was too broad.** That holds for a strictly increasing sequence of *post-page-alignment* lookup sizes against an isolated cache. Page rounding happens before lookup ([allocator.cpp:123](https://github.com/ml-explore/mlx/blob/main/mlx/backend/metal/allocator.cpp#L123)), so growing raw requests can collapse to equal lookup sizes, and a populated global cache may already hold an older, larger buffer.

**What survives:** for growing `concatenate` specifically, `concatenate_gpu` always allocates a fresh output ([slicing.cpp:26](https://github.com/ml-explore/mlx/blob/main/mlx/backend/metal/slicing.cpp#L26)) and copies every input into it, and `reuse_from_cache` itself performs no copy — the copy is imposed by `concatenate`, not by allocator reuse. So the mechanism this patch targets is real; my characterisation of its uniqueness was not.

Given (2), a question before I spend the benchmark time: would you rather see a growth-aware `concatenate` using `buffer_size()` + donation than a global rounding rule in `malloc()`? If so this PR is probably the wrong shape and I'd rather find that out now than after producing numbers for it. If you'd still like the comparison measured, the suite I described stands and I'll add the manual-over-allocate arm with the donation caveat above made explicit.

PhilipJohnBasile:
Partial results — memory arm complete, timing arm deferred (this machine currently has loadavg >5 with a simulator and an emulator live; I won't publish throughput numbers taken under that).

**Both arms built and verified comparable.** Baseline is `3d9e72df^`, not `main` — worth flagging as a trap for anyone reproducing this: 164/166 object files are bit-identical between arms, only `allocator.cpp.o` and `version.cpp.o` differ, and `mlx.metallib` is bit-identical. Discriminator probe: a 1,064,960-byte request yields an active-memory delta of 1,064,960 (base) vs 1,114,112 (patch).

### The measurement problem comes first

`get_peak_memory()` cannot see this bug. `peak_memory_ = std::max(peak_memory_, active_memory_)` ([allocator.cpp:167](https://github.com/ml-explore/mlx/blob/main/mlx/backend/metal/allocator.cpp#L167), and again at :215) tracks `active_memory_` only — the buffer pool is excluded. On the PR's own workload, base reports `get_peak_memory() = 64,716,808` while simultaneously holding `get_cache_memory() = 2,155,675,652`. **The standard metric understates the condition by 33x.** `mlx_lm`'s `GenerationResponse.peak_memory` is this same counter.

So the honest instrument is `max(get_active_memory() + get_cache_memory())`, sampled in-process. Anyone who benchmarks this with the obvious metric will correctly conclude nothing is happening.

### Benefit, measured

Growing-concatenate workload, `cache_end`: **2,155,675,652 → 110,116,868** bytes. That is 95.3x the final tensor size dropping to 5.9x — a **19.6x reduction** in retained pool.

### Cost, measured — and this is the part that argues against my patch

Since `active_memory_ += buf->length()`, the rounding tax is directly observable at weight-load time. Static safetensors-header arithmetic predicted +231.9 MiB for Qwen3-14B-4bit; measured `active_after_load` went 8,307,800,072 → 8,550,938,632, i.e. **+243,138,560 B = +231.87 MiB = +2.927%**. Prediction matched measurement essentially to the byte.

| model | tax |
|---|---|
| Qwen3-0.6B-4bit | +0.635% |
| **Qwen3-14B-4bit** | **+2.927%** |
| Qwen3-32B-4bit | +2.011% |
| Qwen3.6-35B-A3B-4bit | +0.116% |
| Qwen3-Next-80B-A3B-4bit | +0.019% |

The worst offenders are `embed_tokens.weight` / `lm_head.weight` at `[151936, 640] U32`: 388,956,160 → 402,653,184, +13.06 MiB each. Page alignment costs those tensors **zero** bytes, so the entire delta is attributable to the patch. Power-of-two-shaped tensors round to exactly 0. Dense models pay; MoE models are nearly free.

A ~3% unconditional footprint tax on dense models, to fix a pathology that only manifests in unmanaged growth, is a poor trade — which reinforces what I said above about a growth-aware `concatenate` being the better shape.

### One observation relevant to the programmer-level argument

`mlx_lm`'s `KVCache` uses `step = 256` and still calls `mx.concatenate` on every capacity exhaustion ([cache.py:347-348](https://github.com/ml-explore/mlx-lm/blob/main/mlx_lm/models/cache.py#L347)). So the over-allocate pattern bounds the *frequency* of reallocation rather than eliminating it. I'm running a `step = 4096` arm to test whether a one-line userspace change captures most of the benefit — if it does, that is a stronger argument than either of ours for not touching the allocator at all.

I'll follow up with timing and training once the machine is quiet. If you'd rather I stop here and redirect at `concatenate`, say so and I'll close this.

PhilipJohnBasile:
Partial results. I'll lead with the part that changes the PR rather than the benchmark you asked for, because **the patch as posted has a blocking memory regression** and I no longer think the allocator-wide design should merge.

## The regression

`is_donatable()` ([common/utils.h](https://github.com/ml-explore/mlx/blob/main/mlx/backend/common/utils.h)) tests

```cpp
in.buffer_size() <= out.nbytes() + 16384
```

`array::buffer_size()` is the **allocated** length (`MTL::Buffer::length()`). Once `malloc()` rounds up to a 1/16-octave class, the candidate can exceed the output by up to 1/16, blowing past the fixed 16 KiB slack. Donation is refused, the intermediate stays live, and an O(1)-memory elementwise chain becomes O(n).

To be clear about whose contract broke: donation inspecting the *physical* allocator size is deliberate — it exists to stop a small logical output inheriting an oversized buffer. The rounding change silently violates that contract rather than tripping over something surprising.

Six shapes × four chain lengths, plain `abs`/`add` chains, `float32`, 7 runs per cell (deterministic counters — the runs agreed exactly, so treat the reps as a reproducibility check, not statistics). The signature is that the peak ratio **grows with chain length**; a flat ratio would just be the footprint tax:

| shape | chain=1 | chain=8 | growth | |
|---|---|---|---|---|
| (1024, 1025) | 1.588x | 8.996x | **5.667** | donation lost |
| (2048, 2064) | 1.581x | 8.961x | **5.667** | donation lost |
| (3000, 3001) | 1.572x | 8.910x | **5.667** | donation lost |
| (7000, 7000) | 1.541x | 6.677x | **4.333** | donation lost |
| (2048, 2048) | 1.000x | 1.000x | 1.000 | intact |
| (2048, 2560) | 1.000x | 1.000x | 1.000 | intact |

The two unaffected shapes are the ones whose byte size lands **exactly on a class boundary** — `2048*2560*4 = 20,971,520 B = 16 MiB + 4 steps`. Note that (2048, 2560) is not a power-of-two shape, so "test a power-of-two tensor" is not the escape hatch; exact class alignment is. Everything else loses donation.

This is also why the PR's own repro didn't catch it: `concatenate_gpu` always allocates a fresh output, so that path never exercises donation at all.

## A one-hunk diagnostic (not a proposed fix)

Scaling the slack to cover the allocator's own bound —

```cpp
const size_t extra = std::max(out.nbytes() >> 4, size_t{16384});
```

— restores every shape to donation-intact, which confirms the diagnosis. I am **not** proposing it for merge:

- `is_donatable()` is shared — metal (15 call sites), common (9), cpu (6), cuda (3) — so this changes CPU and CUDA donation too.
- It weakens the deliberate anti-retention heuristic above, making tolerated excess proportional to tensor size.
- `nbytes >> 4` is looser than the actual class padding through most of an octave.

A real fix should keep the logical 16 KiB rule and account exactly for Metal's rounding — e.g. compare against an allocator "good size", or track requested size separately from physical length.

## The cost, measured

Weight-load tax, 3 fresh processes per cell, zero within-arm variance:

| model | tax |
|---|---|
| Qwen3-0.6B-4bit | +0.635% |
| **Qwen3-14B-4bit** | **+2.927%** |
| Qwen3-32B-4bit | +2.011% |
| Qwen3.6-35B-A3B-4bit | +0.116% |
| Qwen3-Next-80B-A3B-4bit | +0.019% |

The 14B delta (243,138,560 B) matches a static safetensors-header prediction and an earlier independent run byte-for-byte. Worst offenders are `embed_tokens`/`lm_head` at `[151936, 640] U32`, +13.06 MiB each; page alignment costs those zero, so the delta is entirely attributable to rounding. The dense checkpoints I tested pay up to 2.9%; the two MoE checkpoints I tested were nearly unaffected — I wouldn't generalise further than those five.

The donation fix does **not** change this tax, so the ~3% is purely the rounding rule.

## Where I land

The strongest argument against the design isn't the regression — that's fixable. It's the **unconditional cost**: up to 2.9% of active model bytes on tested dense checkpoints, permanently, to fix a pathology that only appears in unmanaged growth. The regression matters mainly because it shows the change interacts with more of the system than a `malloc()`-local rule appears to.

So I no longer recommend merging the allocator-wide version, and I'd suggest closing it in favour of investigating something targeted at the growth site. I'm deliberately *not* promising a `concatenate` PR as the replacement, because I haven't prototyped it and it has at least three unsolved problems: the current donation predicate doesn't establish capacity; there's no concatenate-local over-allocation policy (upstream page slack alone gives little reusable room); and for row-contiguous concatenation on axes after 0 the leading strides change, so prior contents aren't already in position — which is exactly the `axis=-2` multi-head KV case.

Your call on sequencing. If you'd still like the timing and training arms on the rounding approach before deciding, they're queued and I'll post them. Either way this is a **partial stop-gate, not the comprehensive suite you asked for** — timing and training remain unmeasured. Harness, arm-comparability manifests (164/166 object files bit-identical between arms) and all raw rows are published.

zcbenz:
Thanks for running the experiment, I'm closing this PR since it does not seem to deliver ideal results.

Issue #3611, opened 2026-06-01
Convolution shape and GEMM arithmetic overflow int32 for large or extreme-parameter inputs


uqio:
While building a Rust binding over `mlx-c`, a static review of the convolution path surfaced several `int` (int32) quantities that overflow for large or extreme-parameter convolutions. Signed-integer overflow is UB in C++, and in practice yields wrong shapes/sizes. Reporting in case these are worth widening to `int64_t` or guarding with a raised error.

Quantities computed in `int`:

1. **Forward output shape** (`conv_out_shape`): per axis, `dilation * (kernel - 1)`, the effective input `input_dilation * (in - 1) + 1`, and `(effective_in + pad_lo + pad_hi - dilated_kernel) / stride + 1`. A `dilation`/`padding`/`stride` near `INT_MAX` overflows before validation.

2. **Transposed prelude** (`conv_transpose_general`): computes `1 + dilation * (weight_dim - 1)`, `2 * padding`, and `(in - 1) * stride + ...` over the weight and parameters *before* the nested `conv_general` validates the rank, so an `output_padding`/`dilation` near `INT_MAX` overflows even for a tiny input.

3. **Negative-padding normalization**: slicing every dimension forms `dim + 1` / `0 - pad_lo` in int32, overflowing for a dimension near `INT_MAX`.

4. **Metal implicit GEMM** (`implicit_gemm_conv_2D_gpu` and the 3D path): `implicit_M = N * oH * oW` is an int32 product of the *output* spatial dims. For a conv whose batch x output-spatial exceeds `INT_MAX` this overflows — e.g. `conv2d` on input `[1, 46341, 46341, 1]` gives `implicit_M = 46341 * 46341 = 2147488281 > 2147483647`. The K dimension `product(weight_spatial) * channels_per_group` and the `lcm(input_dilation, stride)` jump-table sizes are similarly int32.

Most need either multi-GB tensors or extreme parameters, so they're unlikely in normal use, but they are reachable from the safe API and are UB rather than a clean error. Would you accept widening these to `int64_t` (or adding overflow checks)? Happy to help pin down exact locations.

zcbenz:
Lot of the cases were already fixed, for example https://github.com/ml-explore/mlx/pull/3524, so I'm closing this issue. Please feel free to open pull requests for cases that still affected by UB.

PR #3938, opened 2026-07-28
Fix signed-integer overflow in convolution shape arithmetic

eyupcanakman:
Follow-up to #3611, which was closed as mostly fixed with an invitation to open PRs for whatever still hit UB. #3524 handled the Metal and CUDA sides. The shape arithmetic itself is still 32-bit.

`conv_out_shape` computes `1 + dilation * (dim - 1)` and the padding sum in 32 bits. `conv_transpose_general` works out all of its padding in 32 bits before the nested `conv_general` validates anything, and the backward pass repeats those same formulas in `Convolution::vjp` and `conv_weight_backward_patches`. Extreme but in-range int32 parameters wrap rather than raise. On an 8x8 input with a 3x3 weight, `mx.conv_general(..., kernel_dilation=2**31-1)` returns a 10x10 output.

These now compute in `int64_t` and narrow through `safe_cast`, the same way #3524 and #3604 did. `Convolution::vjp` calls the 64-bit `dilate_size` helper rather than repeating the formula, which is what #2601 pulled it out for.

Most of those inputs raised nothing before. The ones that did move from `std::invalid_argument` to `std::overflow_error`, so Python raises `OverflowError` where it used to raise `ValueError`. In-range convolutions are unchanged.

Built CPU-only with `-DUSE_UBSAN=ON`. The added test drives 15 signed-overflow reports before the fix and none after, and 9 of its 11 assertions fail without it. Full C++ suite passes 245/245, Python 784 passed.

Two things #3611 also lists are left alone. The lcm jump-table sizes are in the Metal backend, and the `dim + 1` half of its negative-padding item is in `normalize_slice`, reachable from `mx.slice` with no convolution involved.

### DIFF
--- a/mlx/ops.cpp
+++ b/mlx/ops.cpp
@@ -4343,7 +4343,6 @@ array conv_transpose_general(
     StreamOrDevice s) {
   std::vector<int> padding_lo(padding.size());
   std::vector<int> padding_hi(padding.size());
-  // Computed in 64 bits: this runs before conv_general validates anything.
   for (int i = 0; i < padding.size(); ++i) {
     int64_t wt_size =
         1 + static_cast<int64_t>(dilation[i]) * (weight.shape(1 + i) - 1);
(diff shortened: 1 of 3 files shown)


zcbenz (approved):
Thanks!

PR #3930, opened 2026-07-27
Add CUDA 13 runtime headers

jasp-nerd:
## Proposed changes

Fixes #3859.

The CUDA 13 backend wheel does not depend on the package containing
`cuda_runtime.h`. After that package is installed, its CUDA 13 headers are
under `nvidia/cu13/include`, while MLX only checks the CUDA 12
`nvidia/cuda_runtime/include` layout.

This change:

- adds the CUDA 13 runtime package dependency;
- resolves both the CUDA 12 and CUDA 13 package layouts while preserving the
  existing toolkit fallback;
- requires the runtime header pair before accepting an include directory; and
- verifies the unconditional dependency in the built CUDA 13 wheel metadata.

The path tests use isolated package trees so a system CUDA installation cannot
mask a missing package lookup. CCCL is not added because MLX already bundles
the CCCL headers used by its JIT kernels.

Validation:

```text
focused path tests: 1 test case, 5 assertions passed
CPU C++ suite: 245 test cases, 3,243 assertions passed
mocked CUDA 13 package metadata: passed
pre-commit hooks on changed files: passed
```

This machine does not have a CUDA toolkit or NVIDIA GPU, so the real CUDA 13
wheel build, JIT smoke test, and Windows coverage are left to repository CI.

I used an AI coding assistant to help research and prepare this change; I
remain responsible for the submitted patch.

## Checklist

- [x] I have read the [CONTRIBUTING](https://github.com/ml-explore/mlx/blob/main/CONTRIBUTING.md) document
- [x] I have run `pre-commit run --all-files` to format my code / installed pre-commit prior to committing changes
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have updated the necessary documentation (not needed for this packaging fix)

Issue #3859, opened 2026-07-17
[PACKAGING] `mlx[cuda13]` missing CUDA runtime/CCCL headers — first GPU op fails; silent abort under pytest


stoyoda0012-cyber:
## Describe the bug

A fresh `pip install "mlx[cuda13]"` cannot execute any GPU op. The
extra pulls `nvidia-cublas`, `nvidia-cuda-nvrtc`, `nvidia-cudnn-cu13`,
`nvidia-cufft`, `nvidia-nccl-cu13`, `nvidia-nvjitlink` — whose combined
`include/` holds only 13 headers (cuBLAS/cuFFT/NVRTC API headers). The
CUDA runtime headers (`cuda_runtime.h`, ...) and CCCL (`cub`/`thrust`)
that NVRTC needs to JIT MLX's kernels are absent, so the first kernel
compile fails:

```
RuntimeError: Can not find locations of CUDA headers, please set
environment variable CUDA_HOME or CUDA_PATH.
```

Two aggravations:

1. On a machine with no system CUDA toolkit (the presumable target
   audience of a pip-wheel backend), there is no valid `CUDA_HOME` to
   point at — the fix is discovering that two more pip packages exist.
2. **Under pytest the error is invisible**: the suite dies with a bare
   `Fatal Python error: Aborted` and a C-level stack dump, no mention
   of headers. We only found the real error by running a single op
   outside the test harness.

## To Reproduce

On a machine with no system CUDA toolkit:

```bash
python -m venv v && . v/bin/activate
pip install "mlx[cuda13]" numpy
python -c "import mlx.core as mx; mx.eval(mx.zeros((4,)) + 1)"
# RuntimeError: Can not find locations of CUDA headers ...
```

## Workaround / suggested fix

```bash
pip install nvidia-cuda-runtime nvidia-cuda-cccl
export CUDA_HOME="$VIRTUAL_ENV/lib/python3.12/site-packages/nvidia/cu13"
```

(Note the unsuffixed package names; `nvidia-cuda-runtime-cu13` /
`nvidia-cuda-cccl-cu13` are deprecated stubs that fail to build.)

Suggested fixes, any of:

1. Add `nvidia-cuda-runtime` and `nvidia-cuda-cccl` to the `[cuda13]`
   extra, and default the header search to the venv's
   `site-packages/nvidia/cu13` before requiring `CUDA_HOME`.
2. Vendor the needed headers in the `mlx-cuda-13` wheel.
[...]

PR #3944, opened 2026-07-30
[CUDA] Define MLX_CCCL_DIR for every build tree, not only tests

GoodOlClint:
## Proposed changes

`cccl_dir()` backs the fallback branch of `include_path_args()` in `jit_module.cpp`: when the installed `include/cccl` directory is not next to the binary, the JIT falls back to the CCCL tree that `FetchContent` downloaded at build time. That is exactly the situation for any binary running out of a build tree.

But the `MLX_CCCL_DIR` define that `cccl_dir()` reads is applied only under `if(MLX_BUILD_TESTS)`. In any other CUDA build tree `cccl_dir()` returns `nullptr`, the CCCL `--include-path` is dropped silently — the branch is guarded by `std::filesystem::exists`, so nothing throws — and NVRTC then fails to find `cuda/std/*` for every JIT'd kernel that includes it.

The comment above the guard already states the rationale: *"The binary of C++ tests will not be installed so it can not find the CCCL headers, and we have to hard-code the path."* That is true of every binary in the build tree, not only the test binary. This PR widens the define to match, and rewords the comment.

Dropping the guard is cheap:

* CCCL is fetched unconditionally ~10 lines above, so `cccl_SOURCE_DIR` is always populated by the time this runs.
* The define is already `PRIVATE` to `mlx_dirs`, the small object library that exists to hold these path defines, so nothing else recompiles.

## Verification

Configured with `-DMLX_BUILD_CUDA=ON -DMLX_BUILD_TESTS=OFF` (CUDA 13.0.88, aarch64 Linux, `MLX_CUDA_ARCHITECTURES=80`), built the `mlx_dirs` target, and linked `dirs.cpp.o` against a small driver that prints `cccl_dir()`:

|  | `dirs.cpp` compile line | `cccl_dir()` returns |
|---|---|---|
| before | no `-DMLX_CCCL_DIR` | `(nullptr)` |
| after | `-DMLX_CCCL_DIR="<build>/_deps/cccl-src/include"` | `<build>/_deps/cccl-src/include` |

and that directory does hold the fetched `cuda/std/` headers.

Adjacent but not overlapping: #3930 (for #3859) fixes the CUDA *runtime* header lookup for the pip wheel layouts. This is the CCCL branch of the same function, and the two changes touch different code.

## Checklist

On tests: the define is missing precisely when `MLX_BUILD_TESTS=OFF`, i.e. when the test binary is not built, so no in-tree test can observe the defect — which is also why CI does not catch it. The before/after above is the closest equivalent I could run. Happy to add something if you have a preferred way to cover build-configuration behaviour.

### DIFF
--- a/mlx/backend/cuda/CMakeLists.txt
+++ b/mlx/backend/cuda/CMakeLists.txt
@@ -236,12 +236,10 @@ install(DIRECTORY ${cccl_SOURCE_DIR}/include/cuda
 install(DIRECTORY ${cccl_SOURCE_DIR}/include/nv
         DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}/cccl)
 
-# The binary of C++ tests will not be installed so it can not find the CCCL
-# headers, and we have to hard-code the path.
-if(MLX_BUILD_TESTS)
-  target_compile_definitions(mlx_dirs
-                             PRIVATE MLX_CCCL_DIR="${cccl_SOURCE_DIR}/include")
-endif()
+# Binaries in the build tree are not installed, so they can not find the CCCL
+# headers next to them and have to fall back to this hard-coded path.
+target_compile_definitions(mlx_dirs
+                           PRIVATE MLX_CCCL_DIR="${cccl_SOURCE_DIR}/include")
 
 # Use fixed version of NVTX.
 FetchContent_Declare(


zcbenz:
The default path is an absolute local path that would not exist on other machines when shipping the built binaries.

PR #3947, opened 2026-07-30
[CUDA] Ce fused kernel

nastya236:
Fused backward + forward kernels for cross entropy. 

There is an alternative to make this a custom kernel and put it into pretraining pipeline (as in mlx-lm for kl and js losses). But after given it a thought, I am leaning towards having a fast primitive for it. I feel that ce loss is an essential part of next token prediction and it deserves to be a fast primitive:) 

forward: compute `loss = logsumexp(x) - x_n` -> logits and loss are saved for backward
backward: compute `exp(loss + x_n) - onehot`, donate logits buffer for dx (we can do it in most cases) 

`logits = model(x).astype(mx.float32)` <- now after forward we don't need to cast to `float32` and it saves crazy amount of memory: for 4B pretraining with `input.shape = (2, 8192)` and `vocab = 128256` it results in more than 17GB decrease of peak memory! 

I am happy to close this PR and make it a custom kernel if you think it is a preferable option.


zcbenz (approved):
The dropped peak memory is amazing!

PR #3952, opened 2026-07-31
added eye(0) support and tests

aaishwarymishra:
## Proposed changes
This pull request improves the behavior of the `eye` function to correctly handle cases where the number of rows or columns is zero, and adds corresponding tests to ensure this behavior is correct.

**Bug fix and behavior improvement:**

* Updated the `eye` function in `mlx/ops.cpp` to return an empty array when either the number of rows or columns is zero, instead of throwing an error or producing invalid output.

**Testing enhancements:**

* Added new test cases to `python/tests/test_ops.py` for `mx.eye` to verify correct handling of zero rows and/or columns.
* Added C++ tests in `tests/ops_tests.cpp` to check that `eye(0, 0)` returns an empty array with the correct shape, size, and data type.

## Checklist

Hi, this was part of the https://github.com/data-apis/array-api-compat/issues/452 where we were trying to add support of MLX in `array-api-compat` where we found some of the issues when running tests.

### DIFF
--- a/mlx/ops.cpp
+++ b/mlx/ops.cpp
@@ -360,10 +360,15 @@ array ones_like(const array& a, StreamOrDevice s /* = {} */) {
 }
 
 array eye(int n, int m, int k, Dtype dtype, StreamOrDevice s /* = {} */) {
-  if (n <= 0 || m <= 0) {
+  if (n < 0 || m < 0) {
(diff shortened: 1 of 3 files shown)


zcbenz (approved):
Thanks!

PR #3955, opened 2026-07-31
Sample randint directly instead of through a float32 uniform

### LINKED ISSUE #3926: random.randint loses integer range semantics through float32

axiom-of-choice:
## Proposed changes

Fixes #3926.

## Problem

`randint` sampled through a float32 uniform and cast the result:

```cpp
```

float32 has a 24-bit mantissa, so once the bounds reach 2^24 the integer interval is gone before the cast. Three distinct symptoms, all on `main`:

```python
>>> key = mx.random.key(42)
>>> x = mx.random.randint(2**24, 2**24 + 2, (10_000,), dtype=mx.int32, key=key)
>>> sorted(set(x.tolist()))
[16777216, 16777218]      # returns `high` (excluded); never returns 2**24 + 1

>>> y = mx.random.randint(2**40, 2**40 + 1024, (10_000,), dtype=mx.int64, key=key)
>>> len(set(y.tolist()))
1                         # a 1024-wide interval collapses to one value
```

At 2^24 adjacent float32 values are two integers apart; at 2^40 the spacing is 2^17, so the whole interval sits between two representable floats.

## Change

Sample the integers directly. The interval width is computed in int64 and reinterpreted as **uint64**, then 64 uniform bits (two `bits()` draws) are reduced onto it with a remainder.

The residual modulo bias is bounded by `range / 2^64`, below 1e-9 for any range under 2^32, and exactly zero when the width is a power of two. That is against the current behaviour of making values unreachable outright.

Empty intervals (`high <= low`) keep returning `low`, as before.

## Relationship to #3936

#3936 proposes the same integer-domain approach, and I want to be clear that the idea is not mine alone; I reached it independently and [said so there](https://github.com/ml-explore/mlx/pull/3936#issuecomment-5144707879), offering the fix below as a patch to that branch rather than a competing PR. Opening this now so there is a reviewable version that closes the remaining case; happy for it to be closed in favour of #3936 if that branch picks the fix up.

The gap is that #3936 clamps the **signed** difference:

```cpp
auto range = subtract(hi, lo, stream);                                   // int64
auto safe_range = maximum(range, array(int64_t(1), int64), stream);      // signed compare
```

For an interval wider than 2^63 the true width overflows int64 and wraps negative, so `maximum(range, 1)` reads it as negative and clamps to `1`. Every draw then lands on `low`.

I built that branch (`6de5378`, CPU-only, macOS/arm64) and measured 20,000 draws per interval against this one:

| interval | #3936 | this PR |
|---|---|---|
| `[2^24, 2^24+2)` | 2 distinct ✅ | 2 distinct ✅ |
| `[2^40, 2^40+1024)` | 1024 ✅ | 1024 ✅ |
| `[0, 2^63-1)` | 20000 ✅ | 20000 ✅ |
| `[-2^62, 2^62)` | **1** ❌ | 20000 ✅ |
| `[-2^63, 2^63-1)` | **1** ❌ | 20000 ✅ |

```python
# on #3936
>>> x = mx.random.randint(-(2**62), 2**62, (20_000,), dtype=mx.int64)
>>> min(x.tolist()), max(x.tolist()), len(set(x.tolist()))
(-4611686018427387904, -4611686018427387904, 1)   # every draw == low
```

So `randint(-2^62, 2^62)` returns a constant, the same failure this work sets out to fix, relocated to a wider magnitude. An int64 interval can be up to 2^64 - 1 wide, and uint64 is the only type that holds every width `randint` can legitimately be asked for, which is why the clamp has to happen after the reinterpretation:

```cpp
auto empty = less_equal(hi, lo, stream);      // signed, where the compare is meaningful
auto range = where(
    empty,
    array(1, uint64),
    astype(subtract(hi, lo, stream), uint64, stream),
    stream);
```

One smaller point: #3936's comment argues numpy `remainder` semantics keep the 64-bit path exact when the combined draw goes negative. The *sign* of the result is right, but `remainder(raw, r)` for negative `raw` is `raw % r + r`, which skews the distribution toward the low end of the interval by roughly `2^63 mod r`. Reducing in uint64 removes the need for that argument.

## Verification

Reported cases, and the overflow edges:

```
[2^24, 2^24+2)          in_bounds=True  distinct=2
[2^40, 2^40+1024)       in_bounds=True  distinct=1024
[-2^62, 2^62)           in_bounds=True  distinct=20000
[-2^63, 2^63-1)         in_bounds=True  distinct=20000
[0, 2^63-1)             in_bounds=True  distinct=20000
```

Uniformity over 600,000 draws. Every value reachable, worst-case deviation from uniform ~1%:

```
[0, 6)                  coverage=6/6      max_rel_dev=0.0036
[2^24, 2^24+2)          coverage=2/2      max_rel_dev=0.0003
[-10, 10)               coverage=20/20    max_rel_dev=0.0147
```

Existing behaviour preserved: `randint(10, -10, ...)` still returns `low`, same-key draws still reproduce, broadcasting of array bounds unchanged, and `bool` / `uint8` dtypes unchanged.

Suites, CPU-only build on macOS/arm64:

- `python/tests/test_random.py`: 16 passed
- `python/tests/`: 727 passed, 66 skipped, 9718 subtests passed
- `./build-tests/tests/tests`: 244 cases, 3245 assertions, all passed

I could not exercise the Metal path (no full Xcode locally, so no `metal` compiler). The change is in `mlx/random.cpp` above any backend dispatch and the issue reproduces identically on `mx.cpu` and `mx.gpu`, but the GPU path is the part worth a second look.

## Tests added

- `python/tests/test_random.py::test_randint_exact_integer_range`: the 2^24 and 2^40 cases plus a width above 2^63.
- `python/tests/test_random.py::test_randint_uniform_coverage`: every value in a small interval is reachable at roughly equal frequency.
- `tests/random_tests.cpp`: the same three intervals in the existing `test random randint` case.

Both Python tests fail on `main` and pass with the change.

## Checklist

- [x] I have run `pre-commit` (on the changed files: clang-format, black, isort all pass) prior to committing changes
- [ ] I have updated the necessary documentation (if needed). The documented contract ("integers in `[low, high)` with equal probability") is unchanged; this makes the implementation match it

### DIFF
--- a/mlx/random.cpp
+++ b/mlx/random.cpp
@@ -279,8 +279,49 @@ array randint(
     throw std::invalid_argument(
         "[randint] randint only accepts integer dtypes and bool.");
   }
-  auto u = uniform(low, high, shape, float32, key, s);
-  return astype(maximum(u, low, s), dtype, s);
+  auto stream = to_stream(s);
+
+  // Sample the integers directly rather than by casting a float32 uniform.
+  // Routing through float32 loses the integer interval whenever the bounds are
+  // not exactly representable: at 2^24 consecutive float32 values are already
+  // two apart, so values become unreachable, `high` itself can be returned,
+  // and a wide int64 interval can collapse to a single value.
+  auto lo = astype(low, int64, stream);
+  auto hi = astype(high, int64, stream);
+
+  // Width of the half-open interval, as uint64: an int64 interval can be up to
+  // 2^64 - 1 wide, which only uint64 holds. The subtraction is done in int64
+  // and reinterpreted, which is correct even when the difference overflows
+  // int64 — the wrapped bit pattern is the intended unsigned width. Note the
+  // clamp below therefore has to act on the *unsigned* value: clamping the
+  // signed difference would read a genuine width above 2^63 as negative and
+  // collapse the whole interval onto `low`.
+  //
+  // An empty interval (high <= low) is documented to return `low`, and a
+  // remainder by zero is undefined, so those entries get a divisor of 1, which
+  // yields an offset of 0.
+  auto empty = less_equal(hi, lo, stream);
+  auto range = where(
+      empty,
+      array(1, uint64),
+      astype(subtract(hi, lo, stream), uint64, stream),
+      stream);
+
+  // 64 uniform bits per sample, assembled from two 32-bit draws.
+  auto wide_shape = shape;
+  wide_shape.push_back(2);
+  auto words = astype(bits(wide_shape, 4, key, stream), uint64, stream);
+  auto halves = split(words, 2, -1, stream);
+  auto u = bitwise_or(
+      left_shift(squeeze(halves[0], -1, stream), array(32, uint64), stream),
+      squeeze(halves[1], -1, stream),
+      stream);
+
+  // Reduce onto [0, range). The residual modulo bias is bounded by
+  // range / 2^64: below 1e-9 for any range under 2^32, and exactly zero
+  // whenever range is a power of two.
+  auto offset = astype(remainder(u, range, stream), int64, stream);
+  return astype(add(lo, offset, stream), dtype, stream);
 }
 
 array bernoulli(
--- a/python/tests/test_random.py
+++ b/python/tests/test_random.py
@@ -234,6 +234,41 @@ class TestRandom(mlx_tests.MLXTestCase):
             mx.random.randint(0, 1).dtype, mx.random.randint(0, 1, dtype=None).dtype
         )
 
+    def test_randint_exact_integer_range(self):
+        # Bounds that are not exactly representable in float32 must not make
+        # values unreachable, return `high`, or collapse the interval.
+        key = mx.random.key(42)
+
+        # At 2**24 adjacent float32 values are two integers apart.
+        a = mx.random.randint(2**24, 2**24 + 2, (10_000,), dtype=mx.int32, key=key)
+        values = set(a.tolist())
+        self.assertNotIn(2**24 + 2, values)  # `high` is exclusive
+        self.assertEqual(values, {2**24, 2**24 + 1})  # both are reachable
+
+        # At 2**40 the float32 spacing is 2**17, which collapsed this interval
+        # to a single value.
+        b = mx.random.randint(2**40, 2**40 + 1024, (20_000,), dtype=mx.int64, key=key)
+        self.assertTrue(mx.all(b >= 2**40).item())
+        self.assertTrue(mx.all(b < 2**40 + 1024).item())
+        self.assertEqual(len(set(b.tolist())), 1024)
+
+        # A width above 2**63 is representable only as unsigned.
+        c = mx.random.randint(-(2**62), 2**62, (10_000,), dtype=mx.int64, key=key)
+        self.assertTrue(mx.all(c >= -(2**62)).item())
+        self.assertTrue(mx.all(c < 2**62).item())
+        self.assertGreater(len(set(c.tolist())), 1)
+
+    def test_randint_uniform_coverage(self):
+        # Every value in a small interval should be reachable, with roughly
+        # equal probability.
+        n = 200_000
+        for low, high in [(0, 6), (-10, 10), (3, 15)]:
+            a = mx.random.randint(low, high, (n,), dtype=mx.int64)
+            counts = [mx.sum(a == v).item() for v in range(low, high)]
+            self.assertTrue(all(c > 0 for c in counts))
+            expected = n / (high - low)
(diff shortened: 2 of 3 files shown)


PhilipJohnBasile:
Replied in detail on #3936 (https://github.com/ml-explore/mlx/pull/3936#issuecomment-5148875160) rather than here, which left this PR with no response at all — correcting that.

Short version, so this thread stands on its own:

- Your report is confirmed on a real CPU-only build at 20,000 draws per interval. The wide-`int64` collapse you found is real, your diagnosis of the signed clamp is right, and your fix is now on #3936's head.
- **This PR has one remaining case**, the same one #3936's head has, because both use `less_equal(hi, lo)` *after* the int64 cast: a `uint64` interval that straddles 2^63 (`low < 2^63 <= high`) reads as empty and collapses to `low`. Its mirror image also bites — a genuinely inverted `uint64` interval straddling 2^63 reads as non-empty and emits a spread instead of the contracted constant.
- I first proposed a one-line fix for it and have since **retracted** that: comparing the raw bounds promotes signed × `uint64` to float32, which reintroduces the same precision bug in the predicate and additionally opens a modulo-by-zero on float bounds. Details and measurements are in the #3936 comment.
- What does survive both directions is comparing in the output domain:

```cpp
auto cmp_dtype = issubdtype(dtype, unsignedinteger) ? uint64 : int64;
auto empty = less_equal(
    astype(high, cmp_dtype, stream), astype(low, cmp_dtype, stream), stream);
```

Happy either way on which PR lands it — yours with variant C folded in and I close #3936, or mine and you close this. Your reinterpret-as-`uint64` core is the right approach regardless, and it's yours.

axiom-of-choice:
Superseded by #3936, which is the earlier PR and now carries everything here plus more.

That branch has the uint64 reinterpretation this PR was built around, and beyond it: a value-based empty-interval predicate that is exact for cross-sign and mixed signed/unsigned bounds, an unconditional clamp on the divisor so a wrong predicate can never expose raw PRNG bits through `remainder(x, 0)`, 64 bits of entropy for every output dtype, and the C++ coverage from this branch. Two bugs this PR still has — a `uint64` interval straddling 2^63 collapsing to `low`, and its inverted mirror emitting a spread instead of the contracted constant — are fixed there. Closing rather than porting, since the work has converged on that branch.

## GPU verification

Worth recording here because both PRs carried a "Metal not exercised locally" caveat, and that gap is now closed. I installed Xcode 26.6 and the Metal toolchain and ran #3936's head (`243dc2fc4`) on an M2 Pro, macOS 26.5.2 — the first run of any of this work on real GPU hardware.

Every interval executed on both `mx.gpu` and `mx.cpu`, with the `(distinct, min, max)` tuples compared: **16/16 correct, and bit-identical across the two backends**, including the cross-sign pairs, the `uint64` intervals straddling 2^63, the full-width `[int64_min, int64_max)` case, and the saturating float32 bounds. `uint32 [0, 3*2^30)` gives bands `[66770, 66866, 66364]` on both.

Suites with Metal enabled:

- `python/tests/test_random.py`: 20/20
- full Python discovery: **749 passed, 45 skipped, 10584 subtests** (vs 728/66 CPU-only — the 21 additional tests are Metal-only paths)
- C++: **260 cases, 3500 assertions** (vs 244/3248 CPU-only)
- warnings-as-errors build: clean

Entropy checks on the combined 64-bit draw, 200k samples of `u64 [0, 2^64-1)`: `corr(hi_half, lo_half) = -0.0015`, `hi == lo` in zero samples, all 64 bit positions within 1% of 0.5.

One cost note, reported in full on #3936: on GPU the new sampler runs **~6x** the old float32 path for large `int32` draws (10M: 16.17 ms vs 2.73 ms), because uint64 `remainder` costs 6.9x its uint32 counterpart on Metal while `divide`/`multiply` are only 1.2–1.8x. I looked at gating the second draw on a range threshold and the arithmetic does not support it — at `2^24 + 1` a single 32-bit draw is already 0.4% off uniform — so it is a trade for correctness, not an avoidable regression.

I have the Metal toolchain set up now, so if #3936 picks up further revisions I'm happy to re-run GPU verification on request.

PR #3956, opened 2026-07-31
Fix incorrect examples and outputs in the usage docs

ayaangazali:
## Proposed changes

A few things in the usage docs do not match what the code actually does. I ran every example below on current main to check.

**`compile.rst`**

- The implicit-state example mixes 3 and 4 space indentation, so copying it into a file gives `IndentationError: unexpected indent`. Aligned the trailing lines with the rest of the block.
- Typo: "Firt call compiles the function".

**`indexing.rst`**

- The `arr[:, :, 0]` block had a stray `array(3, dtype=int32)` line that does not belong to either expression, and both outputs were missing their closing paren. Actual result for both is `array([[0, 2], [4, 6]], dtype=int32)`.
- `arr.shape` was shown as `[8]` and `arr[None].shape` as `[1, 8]`. `shape` is a tuple, so these print as `(8,)` and `(1, 8)`.
- The two boolean-mask assignment examples showed `array([5.0, 2.0, 6.0], ...)` and `array([[1.0, 0.0, 1.0], ...])`, but MLX drops the trailing `.0` for whole valued floats, so they print as `array([5, 2, 6], dtype=float32)` and `array([[1, 0, 1], [0, 0, 1]], dtype=float32)`.

**`quick_start.rst`**

- Same shape issue: `a.shape` shown as `[4]`, actually `(4,)`.

## Checklist

(docs only; every example above was executed against current main and the new outputs are copied from the actual results)

---

small note on how i work: i'm a freshman contributor and i use Claude Code to help me spot this kind of drift, but for this one i pulled each code block out into a scratch file, ran it, and pasted back what it really printed. if any of these outputs look different on your machine i would like to know, since that would be more interesting than the typo.

### DIFF
--- a/docs/src/usage/compile.rst
+++ b/docs/src/usage/compile.rst
@@ -207,9 +207,9 @@ You have two options to deal with this. The first option is to simply return
       state.append(z)
       return mx.exp(z), state
 
-    _, state = fun(mx.array(1.0), mx.array(2.0))
-    # Prints [array(3, dtype=float32)]
-    print(state)
+   _, state = fun(mx.array(1.0), mx.array(2.0))
+   # Prints [array(3, dtype=float32)]
+   print(state)
 
 In some cases returning updated state can be pretty inconvenient. Hence,
 :func:`compile` has a parameter to capture implicit outputs:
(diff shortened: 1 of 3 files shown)


angeloskath (approved):
Thanks!

PR #3959, opened 2026-07-31
docs: remove references to removed --no-verify-script launch flag

latent-9:
The distributed-launch guide documents a `--no-verify-script` flag for `mlx.launch`, but that flag was removed from the launcher.

`python/mlx/_distributed_utils/launch.py` (the `mlx.launch` entry point) defines no `--no-verify-script` option and contains no script-verification logic — grep for `verify` there returns nothing. The flag and its logic were removed in the launcher cleanup (PR #3513); this doc page was last edited before that and went stale.

Because `launch.py` uses `parse_known_args()`, the documented command does not error cleanly — `--no-verify-script` falls into the leftover args, and since it is not `--`, the launcher treats `--no-verify-script` as the program to run instead of `./my-job.sh`, so the job does not run as documented.

This removes the flag from the example command and deletes the paragraph describing the no-longer-existent feature. Docs-only, no functional change.

### DIFF
--- a/docs/src/usage/launching_distributed.rst
+++ b/docs/src/usage/launching_distributed.rst
@@ -160,11 +160,6 @@ host and on the same path. A good checklist to debug errors is the following:
   ``mlx.launch --print-python`` to see what that path is.
 * the script you want to run is available on all hosts at the same path
 
-If you are launching from a node with a completely different setup than the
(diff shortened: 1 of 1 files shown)


zcbenz (approved):
Thanks!

PR #3962, opened 2026-08-01
Fix Transformer ignoring a custom encoder or decoder with no parameters

ayaangazali:
## Proposed changes

`Transformer` picks its submodules with `custom_encoder or TransformerEncoder(...)`. Since `Module` subclasses `dict`, a module that holds no arrays is an empty dict and therefore falsy, so a parameterless custom encoder or decoder gets silently thrown away and the default one is built instead:

```python
import mlx.core as mx
import mlx.nn as nn

bool(nn.Identity())   # False

t = nn.Transformer(dims=8, num_heads=2, num_encoder_layers=3, num_decoder_layers=1,
                   custom_encoder=nn.Identity())
type(t.encoder)       # TransformerEncoder, not Identity
```

Same for `custom_decoder`. It only shows up when the custom module has no parameters of its own, so a custom encoder that happens to hold a weight works fine and this looks inconsistent rather than broken. Switched both to an explicit `is not None` check.

Added a small test for it. It fails on main with the current truthiness check and passes with this change.

## Checklist

(`test_nn.py` passes; happy to drop the test if you would rather keep this to the one line change)

---

context on me: freshman contributor, and i lean on Claude Code when hunting for things like this, but i confirmed the empty-module-is-falsy behaviour myself and checked that a custom module holding a parameter behaves differently, which is what made it click. if you would rather express the fix as a conditional expression than an if/else, say the word.

### DIFF
--- a/python/mlx/nn/layers/transformer.py
+++ b/python/mlx/nn/layers/transformer.py
@@ -327,27 +327,33 @@ class Transformer(Module):
     ):
         super().__init__()
 
-        self.encoder = custom_encoder or TransformerEncoder(
-            num_encoder_layers,
-            dims,
-            num_heads,
-            mlp_dims,
-            dropout,
-            activation,
-            norm_first,
-            checkpoint,
-        )
+        if custom_encoder is not None:
+            self.encoder = custom_encoder
+        else:
+            self.encoder = TransformerEncoder(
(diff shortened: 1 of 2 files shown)


zcbenz (approved):
Nice catch!

PR #3963, opened 2026-08-01
Fix implicit `thread` address space qualifier becoming explicit in metal 4.1 

louen:
## Proposed changes

Metal 4.1 introduced the concept of "generic address space" and defaults pointers and references to this address space (where it used to implicitly be `thread` before) 

In particular this means that `this` which previously was `thread` is now `generic` which causes code with functions that return a thread reference type to fail to compile, for example:
```metal
struct Foo {
  float v[4];
  thread float& at(short i) {
    return v[i];
  }
};
```
starting from **Xcode 27.0 Beta 4** this will fail with 
```
error: reference to type 'thread float' could not bind to an lvalue of type '__metal_generic float'
```
on previous versions of the toolchain, this would emit a warning (if enabled)
```
warning: missing 'thread' address space qualifier [-Wmetal-addr-spaces]
```

To fix the code under the new compiler rules the `thread` qualifier has to be added to such functions returning a reference to, but also to all member functions of classes that call them, propagating virally similarly to `const` member functions in C++.

This propagation stops at the creation of the object since as an automatic variable
```metal
Foo f;
foo.at(42);
```
`Foo` is still implicitly delclared as `thread`.

The fix has been applied automatically in most instances by adding `-Xclang -fixit -Wmetal-addr-spaces` to the `metal` compiler flags in CMake (the warning flag remains in this PR, as discussed below).
Extra fixes were necessary especially in the instancing macros which clang could not fix automatically. 

To verify the fixes, I have taken the following steps
- **The code compiles** without warning, and the **tests succeed** under Xcode 27.0 Beta 3  (`Apple metal version 32023.918 (metalfe-32023.918.1)`) and Beta 4 toolchains (`Apple metal version 32023.921 (metalfe-32023.921)`).
- **The behavior of the metal kernels is not altered** : as the `thread` keyword only makes explicit what was previously implicit behavior. I checked produced `air` files are binary-identical under both toolchains.

As said above in addition to the addition of the keyword I have left the `-Wmetal-addr-spaces` so that older toolchains still warn about the construct.

Note that mlx-swift will have to upgrade its submodule to compile under the new Xcode toolchain

## Checklist

- [ x] I have read the [CONTRIBUTING](https://github.com/ml-explore/mlx/blob/main/CONTRIBUTING.md) document

### DIFF
--- a/docs/src/dev/extensions.rst
+++ b/docs/src/dev/extensions.rst
@@ -330,6 +330,32 @@ GPU kernels in MLX are written using Metal.
     * Documentation for metal shading language: `Metal Specification`_
     * Using metal from C++: `Metal-cpp`_
 
+.. note::
+
+    As of Metal 4.1, the implicit address space of ``this`` in a member function
+    is ``__metal_generic`` rather than ``thread``. Member functions in Metal
+    shading code should therefore carry an explicit address space qualifier,
+    written after the parameter list:
+
+    .. code-block:: C++
+
+        struct Accumulator {
+            float vals[4];
+            thread float& at(short i) thread {
+                return vals[i];
+            }
+        };
+
+    Without the trailing ``thread``, returning a ``thread``-qualified reference
+    to a member no longer compiles::
+
+        error: reference to type 'thread float' could not bind to an lvalue
+        of type '__metal_generic float'
+
+    Metal libraries built with ``mlx_build_metallib`` are compiled with
+    ``-Wmetal-addr-spaces``, which reports a missing qualifier as a warning on
+    toolchains where it is not yet an error.
+
 Let's keep the GPU kernel simple. We will launch exactly as many threads as
 there are elements in the output. Each thread will pick the element it needs
 from ``x`` and ``y``, do the point-wise operation, and update its assigned
--- a/mlx/backend/metal/kernels/quantized_nax.h
+++ b/mlx/backend/metal/kernels/quantized_nax.h
@@ -615,12 +615,12 @@ struct QuantizedBlockLoader {
       ushort simd_lane_id [[thread_index_in_simdgroup]]) thread
       : src_ld(src_ld_),
         tile_stride(
-            reduction_dim ? BCOLS_PACKED * bytes_per_pack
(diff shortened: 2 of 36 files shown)


mlx/backend/metal/kernels/quantized_nax.h
```diff
@@ -752,14 +752,14 @@ struct QuantizedBlockLoader<
...
       : src_ld(src_ld_),
         tile_stride(
-            reduction_dim ? BCOLS_PACKED * bytes_per_pack
+            reduction_dim ? BCOLS_PACKED* bytes_per_pack
                           : BROWS * src_ld * bytes_per_pack / pack_factor),
-        group_stride(BROWS * src_ld / group_size),
+        group_stride(BROWS* src_ld / group_size),
         thread_idx(simd_group_id * 32 + simd_lane_id),
-        bi(n_reads * thread_idx / BCOLS_PACKED),
+        bi(n_reads* thread_idx / BCOLS_PACKED),
```
zcbenz: Can you revert changes like this? The automatic tool probably has some bugs handling `*`.

docs/src/dev/extensions.rst
```diff
@@ -330,6 +330,32 @@ GPU kernels in MLX are written using Metal.
     * Documentation for metal shading language: `Metal Specification`_
     * Using metal from C++: `Metal-cpp`_
 
+.. note::
+
+    As of Metal 4.1, the implicit address space of ``this`` in a member function
```
zcbenz, line 335: Since `extension.cmake‎`  adds the `-Wmetal-addr-spaces` flag, I think there is no need document the error, users should be able to find out the information in this PR by searching the error.

mlx/backend/metal/kernels/quantized_nax.h
louen: Fixed it in 0e3920ca746bbd42ca68c4a7eb4158eca0b00651. I'm going to file a bug to `clang-format`

angeloskath (approved):
Thanks!

zcbenz:
Sorry I didn't realize the it is our own linter messing up with the spaces around `*`, would you mind reverting the change and we can merge.

Issue #2112, opened 2025-04-23
Complex32 (and BComplex32) support


djphoenix:
As title, should we have "complex half" and "complex bfloat16" types? These may be implemented by converting `complex64_t` into template with backing float/half/bfloat16 types. Almost all ops may be just specialized then.

PR #3969, opened 2026-08-01
Template Metal C2C FFT scalar lanes

PhysicistJohn:
## Motivation

MLX's Metal C2C FFT core hard-codes `float2`. This PR is the narrow Metal-kernel foundation for [#2112](https://github.com/ml-explore/mlx/issues/2112): it removes that source-level coupling so a later, end-to-end reduced-precision complex proposal can be evaluated without rewriting the FFT implementation first.

The target use case is complex I/Q and signal-processing workloads, where storing real and imaginary lanes in FP16 or BF16 halves storage and bytes transferred relative to two FP32 lanes. This PR does **not** claim that lower precision is appropriate for every workload; public dtype semantics, host dispatch, operation coverage, accuracy policy, and new static instantiations remain separate work.

## Existing accelerator-library precedent

This is not a novel representation:

- [NVIDIA cuFFT](https://docs.nvidia.com/cuda/cufft/index.html) supports half-complex `CUDA_C_16F` and bfloat16-complex `CUDA_C_16BF` transforms through its type-generic planning API.
- [AMD rocFFT](https://rocm.docs.amd.com/projects/rocFFT/en/develop/reference/api.html) exposes `rocfft_precision_half` together with complex interleaved and planar array layouts.

Those are direct GPU-vendor FFT libraries. This patch only prepares MLX's C2C Metal implementation for equivalent storage lanes; it does not add a public reduced-precision complex type.

## Scope

- Template C2C radix arithmetic, shared memory, and I/O conversion over the scalar lane type.
- Preserve the current public API, host dispatch, static kernel-instantiation table, and packed float32 RFFT/IRFFT paths.
- Generalize the C2C four-step second pass so every eligible lane type retains the current float32 indexing, inversion, and normalization behavior.
- Derive twiddle phase evaluation precision from the scalar lane and Metal's pi constant through the usual arithmetic promotion, rather than naming `float` inside generic FFT code.

## Binary-size impact

No new static kernels or public exports are instantiated.

- Local release `mlx.metallib`: 130,977,720 B → 130,994,264 B
- Delta: **+16,544 B (+0.01263%)**
- `metal-nm -g` exported-symbol output: exact match to the base build

## Validation

- Native `xcrun metal` compile and `metallib` link of the changed `fft.metal` using MLX's build flags.
- Native CMake build with deployment target macOS 15.0 (NAX kernels excluded by the existing build gate): `ctest` **261/261** passed, including all FFT cases.
- Branch-local Metal Python extension: `python -m unittest test_fft -v` — **16 passed, 2 expected skips**.
- Compile-only C2C instantiations for `half2` and `bfloat2` across Stockham, Rader, Bluestein, and four-step planners.
- The phase-typing refinement produces byte-identical production and reduced-precision AIR relative to the previously reviewed generic FFT tree.
- `pre-commit` (clang-format) and `git diff --check` passed.

## Focused float32 regression checks

| Path | Correctness | Generic / base median |
| --- | --- | --- |
| C2C, N=256, batch=4096 | Bit-identical | 0.2952 / 0.2954 ms (0.9993x) |
| Four-step second pass | Bit-identical | 0.2384 / 0.2350 ms (1.0145x) |

The second measurement is a focused dispatch-scale check, not a claim of a material performance change. There is no end-to-end reduced-precision performance claim until a future public dtype and dispatch path exists.

## Downstream half-complex FFT result

[PhysicistJohn/mlx#7](https://github.com/PhysicistJohn/mlx/pull/7) adds one
dependent commit on top of #3969, #3970, and the storage adapter in
[PhysicistJohn/mlx#6](https://github.com/PhysicistJohn/mlx/pull/6). It
instantiates packed `complex_t<half>` C2C Stockham, Rader, Bluestein, and
four-step kernels.

A matched M5 Max benchmark compared the half kernels directly with untouched
`main` FP32 (`fb5133e1`) using identical plans, geometry, inputs, batches,
warmups, samples, and order-balanced runs:

| Plan | Forward half/main | Inverse half/main |
| --- | ---: | ---: |
| Stockham | 2.151x | 2.151x |
| Rader | 1.273x | 1.326x |
| Bluestein | 1.165x | 1.242x |
| Four-step | 2.242x | 2.220x |

The candidate FP32 control was bit-identical to original FP32 on all four paths
and measured 0.992x-1.065x its throughput across forward and inverse. Packed
half-complex storage and external I/O are 50% smaller. A 39-length correctness
sweep through 1,048,576 points remained finite, with worst normalized RMSE of
0.2585% forward and 0.4101% round trip.

### DIFF
--- a/mlx/backend/metal/kernels/fft/radix.h
+++ b/mlx/backend/metal/kernels/fft/radix.h
@@ -16,49 +16,63 @@ them into (n-1)=6,10,12 codelets. */
 #include <metal_math>
 #include <metal_stdlib>
 
-METAL_FUNC float2 complex_mul(float2 a, float2 b) {
-  return float2(a.x * b.x - a.y * b.y, a.x * b.y + a.y * b.x);
+// The codelets are templated over the scalar lane type so reduced-precision
+// complex paths can reuse them; complex values are plain two-lane vectors.
+
+template <typename T>
+METAL_FUNC metal::vec<T, 2> complex_mul(
+    metal::vec<T, 2> a,
+    metal::vec<T, 2> b) {
+  return {a.x * b.x - a.y * b.y, a.x * b.y + a.y * b.x};
 }
 
 // Complex mul followed by conjugate
-METAL_FUNC float2 complex_mul_conj(float2 a, float2 b) {
-  return float2(a.x * b.x - a.y * b.y, -a.x * b.y - a.y * b.x);
+template <typename T>
+METAL_FUNC metal::vec<T, 2> complex_mul_conj(
+    metal::vec<T, 2> a,
+    metal::vec<T, 2> b) {
+  return {a.x * b.x - a.y * b.y, -a.x * b.y - a.y * b.x};
 }
 
 // Compute an FFT twiddle factor
-METAL_FUNC float2 get_twiddle(int k, int p) {
-  float theta = -2.0f * k * M_PI_F / p;
-
-  float2 twiddle = {metal::fast::cos(theta), metal::fast::sin(theta)};
-  return twiddle;
+template <typename T>
+METAL_FUNC metal::vec<T, 2> get_twiddle(int k, int p) {
+  // Derive phase evaluation precision from the scalar lane and Metal's pi
+  // constant. Reduced lanes currently promote to float for fast trig.
+  using phase_T = decltype(M_PI_F * T(0));
+  phase_T theta = -phase_T(2) * phase_T(k) * phase_T(M_PI_F) / phase_T(p);
+  return {metal::fast::cos(theta), metal::fast::sin(theta)};
 }
 
-METAL_FUNC void radix2(thread float2* x, thread float2* y) {
+template <typename T>
+METAL_FUNC void radix2(thread metal::vec<T, 2>* x, thread metal::vec<T, 2>* y) {
   y[0] = x[0] + x[1];
   y[1] = x[0] - x[1];
 }
 
-METAL_FUNC void radix3(thread float2* x, thread float2* y) {
-  float pi_2_3 = -0.8660254037844387;
+template <typename T>
+METAL_FUNC void radix3(thread metal::vec<T, 2>* x, thread metal::vec<T, 2>* y) {
+  T pi_2_3 = -0.8660254037844387;
 
-  float2 a_1 = x[1] + x[2];
-  float2 a_2 = x[1] - x[2];
+  metal::vec<T, 2> a_1 = x[1] + x[2];
+  metal::vec<T, 2> a_2 = x[1] - x[2];
 
   y[0] = x[0] + a_1;
-  float2 b_1 = x[0] - 0.5 * a_1;
-  float2 b_2 = pi_2_3 * a_2;
+  metal::vec<T, 2> b_1 = x[0] - 0.5 * a_1;
+  metal::vec<T, 2> b_2 = pi_2_3 * a_2;
 
-  float2 b_2_j = {-b_2.y, b_2.x};
+  metal::vec<T, 2> b_2_j = {-b_2.y, b_2.x};
   y[1] = b_1 + b_2_j;
   y[2] = b_1 - b_2_j;
 }
 
-METAL_FUNC void radix4(thread float2* x, thread float2* y) {
-  float2 z_0 = x[0] + x[2];
-  float2 z_1 = x[0] - x[2];
-  float2 z_2 = x[1] + x[3];
-  float2 z_3 = x[1] - x[3];
-  float2 z_3_i = {z_3.y, -z_3.x};
+template <typename T>
+METAL_FUNC void radix4(thread metal::vec<T, 2>* x, thread metal::vec<T, 2>* y) {
+  metal::vec<T, 2> z_0 = x[0] + x[2];
+  metal::vec<T, 2> z_1 = x[0] - x[2];
+  metal::vec<T, 2> z_2 = x[1] + x[3];
+  metal::vec<T, 2> z_3 = x[1] - x[3];
+  metal::vec<T, 2> z_3_i = {z_3.y, -z_3.x};
 
   y[0] = z_0 + z_2;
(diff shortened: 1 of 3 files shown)


mlx/backend/metal/kernels/fft/radix.h
```diff
@@ -16,75 +16,128 @@ them into (n-1)=6,10,12 codelets. */
...
-
-  float2 twiddle = {metal::fast::cos(theta), metal::fast::sin(theta)};
-  return twiddle;
+template <typename T>
+METAL_FUNC fft_complex_t<T> get_twiddle(int k, int p) {
+  // Derive phase evaluation precision from the scalar lane and Metal's pi
+  // constant. Reduced lanes currently promote to float for fast trig.
+  using phase_T = decltype(M_PI_F * T(0));
+  phase_T theta = -phase_T(2) * phase_T(k) * phase_T(M_PI_F) / phase_T(p);
+  return fft_make_complex<T>(metal::fast::cos(theta), metal::fast::sin(theta));
```
zcbenz: Most of the `fft_make_complex` should not be needed, you can just do
```suggestion
  return {metal::fast::cos(theta), metal::fast::sin(theta)};
```

mlx/backend/metal/kernels/fft/radix.h
```diff
@@ -16,75 +16,128 @@ them into (n-1)=6,10,12 codelets. */
...
+METAL_FUNC void radix2(thread fft_complex_t<T>* x, thread fft_complex_t<T>* y) {
   y[0] = x[0] + x[1];
   y[1] = x[0] - x[1];
 }
 
-METAL_FUNC void radix3(thread float2* x, thread float2* y) {
-  float pi_2_3 = -0.8660254037844387;
+template <typename T>
+METAL_FUNC void radix3(thread fft_complex_t<T>* x, thread fft_complex_t<T>* y) {
+  auto pi_2_3 = fft_splat<T>(-0.8660254037844387);
```
zcbenz: Similarly I think you don't need `fft_splat` for most code:
```suggestion
  T pi_2_3 = -0.8660254037844387;
```

mlx/backend/metal/kernels/fft/radix.h
```diff
@@ -16,75 +16,128 @@ them into (n-1)=6,10,12 codelets. */
...
+template <typename T>
+struct fft_complex {
+  static_assert(
+      metal::is_floating_point_v<T>,
+      "FFT complex lanes must use a floating-point scalar type");
+  using type = metal::vec<T, 2>;
+};
+
+template <typename T>
+using fft_complex_t = typename fft_complex<T>::type;
```
zcbenz: `fft_complex_t<T>` has the same length with `metal::vec<T, 2>`, and `complex_t` implies it being a complex class but it is actually a vector type which makes the code a lot confusing.

mlx/backend/metal/kernels/fft/readwrite.h
```diff
@@ -74,21 +148,22 @@ struct ReadWriter {
...
   // ifft(x) = 1/n * conj(fft(conj(x)))
-  METAL_FUNC float2 post_in(float2 elem) const {
-    return inv ? float2(elem.x, -elem.y) : elem;
-  }
-
-  // Handle float case for generic RFFT alg
-  METAL_FUNC float2 post_in(float elem) const {
-    return float2(elem, 0);
+  METAL_FUNC complex_T post_in(in_T elem) const {
+    auto value = FFTValueTraits<in_T>::load(elem);
```
zcbenz: I don't quite get the need of `FFTValueTraits`, the old code should just need minimal change?

```suggestion
    return inv ? complex_T(elem.x, -elem.y) : elem;
```

PhysicistJohn:
Thanks for the review! Applied all four: dropped `fft_make_complex` and `fft_splat`, with `get_twiddle` returning the brace-initialized pair per your suggestion and the radix constants back to plain `T` scalars, so the codelets read like the originals again (`x[0] - a_1 / 2`, `a_5 / 4`); dropped the `fft_complex_t<T>` alias in favor of spelling `vec<T, 2>` directly; and dropped `FFTValueTraits`, with `post_in`/`pre_out` back to the original shape and the inverse inlined per your suggestion. The only trait left is the small scalar-type derivation the kernels need to get the lane type from the storage type.

Also rebased onto main to pick up #3970, which resolves the conflict, and folded in the explicit `thread` qualifiers from #3963 along the way. One consequence of restoring the original `post_in`: real-input loads under `inv` now yield `(x, +0)` rather than `(x, −0)`, matching the pre-#3969 code.

mlx/backend/metal/kernels/fft/radix.h
PhysicistJohn: Done — returning the brace-initialized pair directly. The other call sites go back to the original spellings too: brace-init for the lane swaps, `vec<T, 2>(...)` for the Rader constants.

mlx/backend/metal/kernels/fft/radix.h
PhysicistJohn: Done — plain `T` constants throughout, and the places `fft_splat` stood in are back to the original scalar arithmetic (`x[0] - a_1 / 2`, `a_5 / 4`).

mlx/backend/metal/kernels/fft/radix.h
PhysicistJohn: Agreed — dropped it; everything uses `vec<T, 2>` directly now.

mlx/backend/metal/kernels/fft/readwrite.h
PhysicistJohn: Done — `post_in`/`pre_out` are back to the original shape with the inverse inlined per your suggestion; the only trait left is the scalar-type derivation the kernels need. One consequence: real-input loads under `inv` now yield `(x, +0)` instead of `(x, −0)`, matching the pre-#3969 code.

zcbenz (approved):
Looks good to me!

PR #3970, opened 2026-08-01
Template Metal complex scalar lanes

PhysicistJohn:
## Motivation

MLX Metal currently defines complex storage and arithmetic only through the float-specific `complex64_t`, while MLX CUDA already uses the lane-generic `complex_t<T> = cuda::std::complex<T>`. This PR independently gives the Metal backend the same extensible type foundation while preserving all current complex64 behavior and kernel entry points.

It complements upstream [#3969](https://github.com/ml-explore/mlx/pull/3969), which independently templates the Metal C2C FFT scalar lane. Together, the two narrowly scoped changes establish the orthogonal type and FFT foundations needed to evaluate reduced-precision complex support in MLX. That closes a capability gap with accelerator stacks that already provide reduced-precision complex transforms: [NVIDIA cuFFT](https://docs.nvidia.com/cuda/cufft/index.html) supports FP16 and BF16 complex transforms, and [AMD rocFFT](https://rocm.docs.amd.com/projects/rocFFT/en/develop/reference/api.html) supports FP16 complex transforms.

Neither PR introduces a user-facing dtype or commits MLX to promotion, dispatch, or accuracy semantics; those remain separate design decisions.

## Summary

- Replace the float-specific Metal complex structure with lane-generic `complex_t<T>`.
- Preserve every existing kernel entry point through `using complex64_t = complex_t<float>`.
- Generalize scalar and cross-complex conversions, arithmetic, and `Limits` for packed half and bfloat lane types.
- Add compile-time layout and reduced-lane arithmetic checks.

## Scope boundaries

This patch intentionally does not add public complex32/complex16 dtypes, host promotion rules, Python API, production half/BF16 kernel instantiations, FFT integration, or FP8. Those should be separate proposals after this primitive is accepted.

The current Metal compiler accepts the template implementation but rejects C++20 `concept` declarations, so mixed scalar operators use narrow `enable_if` constraints.

## Compatibility checks

| Check | Result |
| --- | --- |
| Release Metal build | Pass |
| Minimum-target Metal compile/link (macOS 14.0 deployment target) | Pass |
| CTest, serial | 261/261 pass on rebased `a79332de` |
| Focused conversion probe | float, half, and bfloat cross-lane conversion, same-lane copying, mixed scalar operations, and complex-to-scalar conversion compiled successfully |
| Existing complex64 output (original `fb5133e1` base) | Bit-identical for 8 operators across 65,536 deterministic values |
| Exported Metal symbols | Exact match to rebased `a79332de`: 15,809 lines and identical SHA-256 |
| Existing complex64 latency (original `fb5133e1` base) | Add -0.49%, multiply +0.06%, divide -0.10%; no material regression |
| Metallib size | 132,415,080 B -> 132,428,520 B; +13,440 B (+0.01015%) |

No new public kernel symbols are emitted.

## Reduced-lane proof

Both `complex_t<half>` and `complex_t<bfloat16_t>` are exactly 4 bytes. A linked on-device probe produced the expected `(-2, 14)` result for both types.

LLVM AIR inspection shows native half and bfloat loads, stores, arithmetic, division, and FMA instructions. The reduced paths do not promote their internal arithmetic to float.

A focused on-device streaming complex multiply-add benchmark over 2^24 elements measured:

| Lane type | Median kernel time | Relative to float |
| --- | ---: | ---: |
| float | 0.8095 ms | 1.00x |
| half | 0.3819 ms | 2.12x faster |
| bfloat | 0.3826 ms | 2.12x faster |

This is a raw-type bandwidth-oriented microbenchmark, not an end-to-end MLX dtype performance claim.

## Downstream half-complex FFT result

[PhysicistJohn/mlx#7](https://github.com/PhysicistJohn/mlx/pull/7) adds one
dependent commit on top of #3969, #3970, and the storage adapter in
[PhysicistJohn/mlx#6](https://github.com/PhysicistJohn/mlx/pull/6). It
instantiates packed `complex_t<half>` C2C Stockham, Rader, Bluestein, and
four-step kernels.

A matched M5 Max benchmark compared the half kernels directly with untouched
`main` FP32 (`fb5133e1`) using identical plans, geometry, inputs, batches,
warmups, samples, and order-balanced runs:

| Plan | Forward half/main | Inverse half/main |
| --- | ---: | ---: |
| Stockham | 2.151x | 2.151x |
| Rader | 1.273x | 1.326x |
| Bluestein | 1.165x | 1.242x |
| Four-step | 2.242x | 2.220x |

The candidate FP32 control was bit-identical to original FP32 on all four paths
and measured 0.992x-1.065x its throughput across forward and inverse. Packed
half-complex storage and external I/O are 50% smaller. A 39-length correctness
sweep through 1,048,576 points remained finite, with worst normalized RMSE of
0.2585% forward and 0.4101% round trip.

### DIFF
--- a/mlx/backend/metal/kernels/complex.h
+++ b/mlx/backend/metal/kernels/complex.h
@@ -4,164 +4,263 @@
 
 #include <metal_stdlib>
 
+#include "mlx/backend/metal/kernels/bf16.h"
+
 using namespace metal;
 
-struct complex64_t;
+template <typename T>
+struct complex_t;
 
 template <typename T>
-static constexpr constant bool can_convert_to_complex64 =
-    !is_same_v<T, complex64_t> && is_convertible_v<T, float>;
+static constexpr constant bool is_complex_v = false;
 
 template <typename T>
-static constexpr constant bool can_convert_from_complex64 =
-    !is_same_v<T, complex64_t> &&
-    (is_convertible_v<float, T> || is_convertible_v<bfloat16_t, T>);
+static constexpr constant bool is_complex_v<complex_t<T>> = true;
 
-struct complex64_t {
-  float real;
-  float imag;
+// Metal accepts explicit bfloat casts that is_convertible_v reports as false.
+template <typename From, typename To>
+static constexpr constant bool is_lane_convertible_v =
+    is_convertible_v<From, To> ||
+    (is_same_v<To, bfloat16_t> && is_convertible_v<From, float>) ||
+    (is_same_v<From, bfloat16_t> && is_convertible_v<float, To>);
+
+template <typename T>
+struct complex_t {
+  using value_type = T;
+
+  T real;
+  T imag;
 
   // Constructors
-  constexpr complex64_t(float real, float imag) thread : real(real),
-                                                         imag(imag) {};
-  constexpr complex64_t() thread : real(0), imag(0) {};
-  constexpr complex64_t() threadgroup : real(0), imag(0) {};
+  constexpr complex_t(T real, T imag) thread : real(real), imag(imag) {};
+  constexpr complex_t() thread : real(0), imag(0) {};
+  constexpr complex_t() threadgroup : real(0), imag(0) {};
 
-  // Conversions to complex64_t
+  // Conversions from scalar types
   template <
-      typename T,
-      typename = typename enable_if<can_convert_to_complex64<T>>::type>
-  constexpr complex64_t(T x) thread : real(x), imag(0) {}
-
-  template <
-      typename T,
-      typename = typename enable_if<can_convert_to_complex64<T>>::type>
-  constexpr complex64_t(T x) threadgroup : real(x), imag(0) {}
+      typename U,
+      typename = typename enable_if<
+          !is_complex_v<U> && is_lane_convertible_v<U, T>>::type>
+  constexpr complex_t(U x) thread : real(static_cast<T>(x)),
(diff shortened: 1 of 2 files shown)


mlx/backend/metal/kernels/complex.h
```diff
@@ -4,163 +4,251 @@
...
+    is_convertible_v<From, To> ||
+    (is_same_v<To, bfloat16_t> && is_convertible_v<From, float>) ||
+    (is_same_v<From, bfloat16_t> && is_convertible_v<float, To>);
+
+template <typename From, typename To>
+static constexpr constant bool can_convert_to_complex =
+    !is_complex_v<From> && can_convert_scalar<From, To>;
+
+template <typename From, typename To>
+static constexpr constant bool can_convert_complex =
```
zcbenz: I'm confused about this, what is the difference between this and `can_convert_to_complex`/`can_convert_from_complex`?

The `can_convert_scalar` is also confusing in the same way.

PhysicistJohn: Thanks, good catch. I agree this was more indirect than it needed to be. These were just overload guards around one lane conversion rule. I collapsed them into `is_lane_convertible_v` and moved the checks for complex values and same-type copies to the relevant overloads. The `bfloat` exception remains because Metal accepts those explicit casts even though `is_convertible_v` is false.

zcbenz (approved):
This looks a good refactoring, thanks!

PR #3971, opened 2026-08-02
Export C++20 requirement to CMake consumers

PhysicistJohn:
## Proposed changes

MLX's installed public C++ headers use C++20 defaulted comparisons in
`Device` and `Stream`, but the exported CMake target does not advertise that
requirement.

This exports `cxx_std_20` to CMake C++ consumers, aligns the three shipped
external CMake examples, and removes the stale, non-transitive
`CXX_STANDARD 17` property from the installed target configuration.

Runtime CPU JIT and compiled-preamble compilation remain on C++17.

## Validation

- An installed-package consumer declaring C++17 inherited `-std=c++20`,
  linked, and ran
- A strict C++17 include-only probe fails at `mlx/device.h:28`; the same probe
  passes under C++20
- All three external CMake examples configured and built; the sample ran and
  the extension imported
- CPU-only Release: 244/244 native cases and 3,238/3,238 assertions
- Static Metal Release: 260/260 native cases and 3,490/3,490 assertions
- CPU JIT remains on `g++ -std=c++17`; native compile coverage passed 31/31
  cases and 136/136 assertions
- `pre-commit run --all-files` and `git diff --check` passed

## Checklist

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -151,6 +151,7 @@ cmake_policy(SET CMP0135 NEW)
 
 add_library(mlx)
 
+target_compile_features(mlx INTERFACE cxx_std_20)
 target_compile_options(mlx PUBLIC ${SANITIZER_COMPILE_FLAGS})
 target_link_options(mlx PUBLIC ${SANITIZER_LINK_FLAGS})
 
--- a/examples/cmake_project/CMakeLists.txt
+++ b/examples/cmake_project/CMakeLists.txt
@@ -2,7 +2,7 @@ cmake_minimum_required(VERSION 3.27)
 
 project(example LANGUAGES CXX)
 
-set(CMAKE_CXX_STANDARD 17)
+set(CMAKE_CXX_STANDARD 20)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
 
 # Comment the following two commands only the MLX C++ library is installed and
--- a/examples/export/CMakeLists.txt
+++ b/examples/export/CMakeLists.txt
@@ -2,7 +2,7 @@ cmake_minimum_required(VERSION 3.27)
 
 project(import_mlx LANGUAGES CXX)
 
-set(CMAKE_CXX_STANDARD 17)
+set(CMAKE_CXX_STANDARD 20)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
 
 find_package(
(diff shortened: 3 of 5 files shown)


zcbenz (changes requested):
The project requiring C++20 does not mean all the dependents would also need C++20 to use it.

PhysicistJohn:
Agreed that MLX being built as C++20 does not, by itself, make C++20 a downstream usage requirement. Let's table the JIT/preamble question for now.

Setting that aside, the current installed public interface raises a separate, narrow usage-requirement question. At this PR's base, [`mlx/device.h`](https://github.com/ml-explore/mlx/blob/fb5133e1049cc1482a330cd3a7135fda62a74b0d/mlx/device.h#L27-L28) contains:

```cpp
bool operator==(const Device&) const = default;
```

That form of defaulted comparison requires C++20. `mlx/device.h` is included by [`mlx/mlx.h`](https://github.com/ml-explore/mlx/blob/fb5133e1049cc1482a330cd3a7135fda62a74b0d/mlx/mlx.h#L5-L10), and the package [installs these public headers](https://github.com/ml-explore/mlx/blob/fb5133e1049cc1482a330cd3a7135fda62a74b0d/CMakeLists.txt#L424-L440).

An include-only probe against the exact base commit:

```sh
printf '#include "mlx/device.h"\n' |
  clang++ -std=c++17 -pedantic-errors -I. -x c++ -fsyntax-only -
```

fails at `device.h:28` with `defaulted comparison operators are a C++20 extension`; the same probe passes under C++20. Plain Apple Clang accepts it in C++17 mode as an extension, so this is specifically a standard-conforming/portable C++17 issue.

[CMake's compile-feature model](https://cmake.org/cmake/help/latest/manual/cmake-compile-features.7.html#compile-feature-requirements) uses `PUBLIC` or `INTERFACE` requirements when a library's public interface requires a feature, including when a compiler happens to accept it as an extension.

That appears to leave two coherent downstream contracts:

1. If portable C++17 consumers are supported, these public comparisons need C++17-compatible implementations and the C++17 examples should test that contract strictly.
2. If the public headers remain as written, the exported target should advertise `cxx_std_20`; `INTERFACE` would be sufficient for that narrower purpose.

Which contract is intended? Any follow-up can be scoped specifically to that decision, leaving JIT/preamble out.

zcbenz:
That's a valid point thanks for clarification. I agree we should require C++20 since device.h uses the feature, but we should still not require C++20 for JIT compilation which is not yet available in many client machines.

PhysicistJohn:
Thanks, understood. The distinction between the installed public C++ interface and the runtime JIT compiler makes sense.

I will narrow this PR to the public usage requirement: export cxx_std_20 through the target interface, retain only the consumer-facing CMake/example cleanup, and remove the JIT/preamble changes and associated regression test. I will also update the title, description, and validation to match that scope.

zcbenz (approved):
Thanks!

PR #3975, opened 2026-08-02
python: fix bfloat16 buffer format itemsize mismatch

### LINKED ISSUE #3974: [BUG] bfloat16 buffer protocol format string itemsize mismatch
**Describe the bug**
Accessing a `bfloat16` MLX array via `np.asarray(x)` or `memoryview(x)` raises a `RuntimeError`:

```text
RuntimeError: Item size 2 for PEP 3118 buffer format string B does not match the dtype B item size 1.
```

**To Reproduce**

```python
import numpy as np
import mlx.core as mx

x = mx.ones((5,), dtype=mx.bfloat16)
mx.eval(x)
arr = np.asarray(x)
```

**Expected behavior**
The buffer view is created without error.

 - OS Version: macOS 15.7.3 
 - Version 0.32.1

**Additional context**
In [`python/src/buffer.h#L45-L46`](https://github.com/ml-explore/mlx/blob/main/python/src/buffer.h#L45-L46), `buffer_format()` returns `"B"` for `mx::bfloat16`:
    https://github.com/ml-explore/mlx/blob/fb5133e1049cc1482a330cd3a7135fda62a74b0d/python/src/buffer.h#L45-L46

`"B"` represents 1-byte unsigned char under PEP 3118 while bfloat16 has an itemsize of 2 bytes, this can be fixed by using type-code `"H"`.

reckylurker:
## Proposed changes
Fixes #3974 

## Checklist

### DIFF
--- a/python/src/array.cpp
+++ b/python/src/array.cpp
@@ -539,6 +539,17 @@ void init_array(nb::module_& m) {
               return nb::make_tuple(1, 0);
             }
           })
+      .def(
+          "__array__",
+          [](const mx::array& self, nb::object dtype, nb::object copy) {
+            if (self.dtype() == mx::bfloat16) {
+              throw nb::type_error(
+                  "bfloat16 arrays cannot be converted to NumPy.");
+            }
+            return mlx_to_np_array(self);
+          },
+          "dtype"_a = nb::none(),
+          "copy"_a = nb::none())
       .def("__copy__", [](const mx::array& self) { return mx::array(self); })
       .def(
           "__deepcopy__",
--- a/python/src/array.cpp
+++ b/python/src/array.cpp
@@ -542,10 +542,6 @@ void init_array(nb::module_& m) {
       .def(
           "__array__",
           [](const mx::array& self, nb::object dtype, nb::object copy) {
-            if (self.dtype() == mx::bfloat16) {
-              throw nb::type_error(
-                  "bfloat16 arrays cannot be converted to NumPy.");
-            }
             return mlx_to_np_array(self);
           },
           "dtype"_a = nb::none(),
--- a/python/src/buffer.h
+++ b/python/src/buffer.h
@@ -43,7 +43,7 @@ std::string buffer_format(const mx::array& a) {
     case mx::float32:
       return "f";
     case mx::bfloat16:
-      return "H";
+      return "bfloat16";
     case mx::float64:
       return "d";
     case mx::complex64:
--- a/python/src/buffer.h
+++ b/python/src/buffer.h
@@ -43,7 +43,7 @@ std::string buffer_format(const mx::array& a) {
     case mx::float32:
       return "f";
     case mx::bfloat16:
-      return "B";
+      return "H";
     case mx::float64:
       return "d";
     case mx::complex64:
--- a/python/tests/test_array.py
+++ b/python/tests/test_array.py
@@ -1868,10 +1868,10 @@ class TestArray(mlx_tests.MLXTestCase):
         mv_mx = memoryview(a_mx)
         self.assertEqual(mv_mx.strides, (8, 2))
         self.assertEqual(mv_mx.shape, (3, 4))
-        self.assertEqual(mv_mx.format, "H")
-        with self.assertRaises(TypeError) as cm:
+        self.assertIn(mv_mx.format, "bfloat16")
+        with self.assertRaises(ValueError) as cm:
             np.array(a_mx)
-        self.assertIn("bfloat16 arrays cannot be converted to NumPy", str(cm.exception))
+        self.assertIn("bfloat16", str(cm.exception))
 
         # Test buffer protocol with non-arrays ie bytes
         a = ord("a") * 257 + mx.arange(10).astype(mx.int16)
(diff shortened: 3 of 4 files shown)


zcbenz (approved):
Nice catch and good fix, thanks!

zcbenz:
It seems that cpu backend needs some fixes.

reckylurker:
Hi @zcbenz, Thanks for the approval. 
Looks like the memory format was asserted to be equal to 'B' (1 byte) instead of 'H' (2 bytes) and conversion to numpy for bfloat16 was treated as an error.

https://github.com/ml-explore/mlx/blob/a136dc8ca80bc3b65c637f5f04e9e2892567a840/python/tests/test_array.py#L1866-L1875

Since Numpy doesn't have bfloat16, will just check with `np.uint16`. 
[Edit] Let's just leave numpy conversion as an error instead since bfloat16 -> uint16 will cause errors.

reckylurker:
Raised a type error to prevent conversion from `bfloat16` to `numpy` since numpy doesn't have native support.
Should pass the tests now.

python/src/array.cpp
```diff
@@ -539,6 +539,17 @@ void init_array(nb::module_& m) {
...
             }
           })
+      .def(
+          "__array__",
+          [](const mx::array& self, nb::object dtype, nb::object copy) {
+            if (self.dtype() == mx::bfloat16) {
+              throw nb::type_error(
+                  "bfloat16 arrays cannot be converted to NumPy.");
+            }
+            return mlx_to_np_array(self);
```
zcbenz, line 545: mlx_to_np_array already throws for bfloat16. Also I think there is no need to implement the `__array__`, it should throw in buffer protocol instead.

reckylurker: without implementing `__array__` and throwing `TypeError` in buffer protocol raises `TypeError` in `memoryview` and numpy creates an iterable sequence of `mx.bfloat16`

To avoid `TypeError` in `memoryview` and raise it for Numpy conversion explicitly requires implementing `__array__`

zcbenz: That makes sense, thanks for testing. I think here is no need to throw in this function though since `mlx_to_np_array` itself already throws for bfloat16?

reckylurker: Thanks for your reply @zcbenz, yes the `TypeError` is redundant and can be removed.

Upon further testing, I found that raising `TypeError` in `__array__` alone is not sufficient for Numpy to error out. In Numpy 2.0+, `np.array` call works as follows: 

1) The buffer protocol is called first and any error raised during buffer export is [cleared](https://github.com/numpy/numpy/blob/57ae4890237c92ce0f07bf6954f97564d5aa0899/numpy/_core/src/multiarray/ctors.c#L1389)!
```cpp
if (PyObject_CheckBuffer(op) && !PyBytes_Check(op) && !PyUnicode_Check(op)) {
        PyObject *memoryview = PyMemoryView_FromObject(op);
        if (memoryview == NULL) {
            /* TODO: Should probably not blanket ignore errors. */
            PyErr_Clear();
        }
```
2) Next, NumPy checks if `__array__` is [implemented](https://github.com/numpy/numpy/blob/57ae4890237c92ce0f07bf6954f97564d5aa0899/numpy/_core/src/multiarray/array_coercion.c#L1033)
```cpp
arr = (PyArrayObject *)_array_from_array_like(obj,
                requested_descr, 0, copy, &was_copied_by__array__);
        if (arr == NULL) {
            return -1;
        }
        else if (arr == (PyArrayObject *)Py_NotImplemented) {
            Py_DECREF(arr);
            arr = NULL;
        }
```
3) if `__array__` raises an error, only then Numpy raises an error (in the above code snippet, checking if `arr==NULL`)

To allow `memoryview` call to pass but `numpy` call to fail, buffer protocol must backtrace the call stack to detect "numpy" which can have a significant performance impact. On my M2 MacBook Air, I found that backtrace checking (for the last 32 frames in the stack) had a 4x slowdown in `mx.array` to `np.array` conversion than the present version.

Instead a cleaner fix will be to use a type format other than PEP 3118. Let's still retain `__array__` to allow `mlx.core.array.__array__` calls where a `TypeError` would be thrown from `mlx_to_np_array` upon `bfloat16` conversion

zcbenz: I'm good with current approach, in the meanwhile would it make sense returning uint64 array for bfloat16 dtype? Some existing code does that:
https://github.com/ml-explore/mlx/blob/0268c81b62b135dc99773565016edf2b35a67898/python/src/array.cpp#L481-L488

reckylurker: Yes, returning `uint16` would work for this case. Because bfloat16 is a 2-byte type, viewing the array as
  `uint16` allows dumping binary data to disk without creating temp copies. Note that the `dtype` is packed along with the memory view in the tuple. 

When the data is read from disk again via `__setstate__`, the values are cast from uint16 array back to bfloat16 using the `dtype` packed in the tuple.
 https://github.com/ml-explore/mlx/blob/0268c81b62b135dc99773565016edf2b35a67898/python/src/array.cpp#L489-L512

PR #3976, opened 2026-08-02
Fix mx.remainder floored-mod for float16/bfloat16 on CPU

sashko-zakharchuk:
`mx.remainder` returns the wrong result for `float16` and `bfloat16` on the CPU backend.

The SIMD kernel sign-corrects the remainder so it takes the sign of the divisor, like `float32` and the GPU backends. The correction is gated on `std::is_signed_v<T>`, which is false for the custom `float16_t` and `bfloat16_t`, so the half types skip it and return the raw IEEE remainder.

Example with `a = [-1, -1, -1]`, `b = [3, 5, 7]`:

- `float32`: `[2, 4, 6]` (correct)
- `float16` and `bfloat16`: `[-1, -1, -1]` (wrong)

The correction now gates on `!std::is_unsigned_v<T> && !is_complex<T>`, in the scalar path (`base_simd.h`) and the Accelerate SIMD path (`accelerate_simd.h`). For every type that was already correct this is the same branch, so only the half types change. `test_remainder` is extended to `float16` and `bfloat16`.

Verified on macOS (M3 Ultra, Metal + Accelerate) and Linux (CPU): `float16`/`bfloat16` remainder now matches `float32` on both CPU and GPU, and `test_ops.py` passes.

`complex64` remainder is also wrong on CPU (it drops the imaginary part), which I'll address separately.

### DIFF
--- a/mlx/backend/cpu/simd/accelerate_simd.h
+++ b/mlx/backend/cpu/simd/accelerate_simd.h
@@ -241,7 +241,7 @@ Simd<T, N> remainder(Simd<T, N> a, Simd<T, N> b) {
   } else {
     r = a - b * (a / b);
   }
-  if constexpr (std::is_signed_v<T>) {
+  if constexpr (is_signed_v<T>) {
     auto mask = r != 0 && (r < 0 != b < 0);
     r = select(mask, r + b, r);
   }
--- a/mlx/backend/cpu/simd/base_simd.h
+++ b/mlx/backend/cpu/simd/base_simd.h
@@ -13,6 +13,8 @@
 #include <intrin.h> // For _BitScanReverse
 #endif
 
+#include "mlx/types/half_types.h"
+
 namespace mlx::core::simd {
 template <typename T, int N>
 struct Simd;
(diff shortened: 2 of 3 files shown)


zcbenz (changes requested):
The correct way to fix this is to define our own `is_signed_v` utility that works with the half types.

Also there are too many unrelated changes in docs.

sashko-zakharchuk:
Done. Added an `is_signed_v` in `base_simd.h` (next to `is_complex`) that treats non-complex floating types as signed, covering `float16_t`/`bfloat16_t`, and gated the floored-mod correction on it in `base_simd.h` and `accelerate_simd.h`. Dropped the docs changes and rebased on main.

zcbenz:
You can check the definition of `is_array_v` or `is_vector_v` for an idea how to do it correctly.

sashko-zakharchuk:
Thanks. Reworked `is_signed_v` to an explicit trait like `is_array_v`: `std::is_signed_v<T> || std::is_same_v<T, float16_t> || std::is_same_v<T, bfloat16_t>`, with `half_types.h` included in `base_simd.h` for the type names.

zcbenz (approved):
Nice catch and good fix!

Issue #2712, opened 2025-10-28
[CUDA][PERF] Matmuls with non-multiple inputs are slow


awni:
The following benchmark if you slice the `k` and `v` to 2047 before applying the matmuls you get a 3x reduction in perf because you hit a very bad kernel. It should be possible to avoid this by padding the output.

```python
import time
import mlx.core as mx

H = 16
D = 128
T = 2048
Ts = 2047

q = mx.random.normal(shape=(H, Ts, D))
k = mx.random.normal(shape=(H, T, D))[:, :Ts, :]
v = mx.random.normal(shape=(H, T, D))[:, :Ts, :]

def fun(q):
    for _ in range(10):
        scores = q @ k.swapaxes(-1, -2)
        q = scores @ v
    return q

for _ in range(10):
    mx.eval(fun(q))

tic = time.time()
for _ in range(10):
    mx.eval(fun(q))
toc = time.time()

ms = 1e3 * (toc - tic)
print(f"{ms=:.3f}")
```

zcbenz:
At least on DGX with CUDA 13 the perf penalty has been reduced to within 20%, since this is something that cuBlas keeps improving there is no need to track this issue ourselves.

PR #3977, opened 2026-08-02
fix(cuda): pad non-aligned matmul inputs for cublasLt performance (#2712)

alphapibeta:
## Summary

This PR is a draft candidate for GitHub issue #2712. It implements a safe zero-padding path for non-aligned GEMMs: A and B are padded to multiples of 32, cuBLASLt runs on aligned working dimensions, and only the logical M×N result is copied back.

This is not claimed as a complete performance fix. The GPU and cuBLASLt configuration affected by #2712 are not specified in the issue, and the workaround must demonstrate a speedup on that configuration before merge.

## Implementation

- Pads A and B into zero-filled arrays with `pad_gpu()`.
- Preserves batch dimensions, transposes, leading dimensions, and batch strides.
- Pads the bias epilogue input when N is padded.
- Copies only the logical result back to the requested output.
- Keeps temporary arrays registered with the command encoder.
- Adds broad non-aligned M/N/K, bias, transpose, batch, and boundary coverage.

### RTX 2060 / Turing

- CUDA library build passed with CUDA 12.6 and SM 7.5.
- Focused PR test: 38/38 assertions passed.
- Clean baseline focused test: 3/3 assertions passed.
- The padding path regressed performance on this GPU, so the current code bypasses padding for Turing and newer devices.
- The exact 16-head issue graph exceeds the 6 GiB allocator budget; no exact RTX 2060 issue-loop timing is claimed.

### RTX PRO 4000 / Blackwell

- Focused PR test: 38/38 assertions passed.
- The exact issue-shaped workload completed, but padding showed no measurable benefit and is bypassed.

### Quadro P2000 / Pascal SM 6.1

The full MLX CUDA library and tests were built after applying temporary local SM 6.1 compatibility workarounds outside this PR. The PR focused test passed with 38/38 assertions; the clean baseline passed with 3/3 assertions.

The individual GEMM comparison showed a substantial regression from input padding:

| Case | Baseline | PR padded path |
|---|---:|---:|
| 2047×2047, K=1 | 271 µs | 1,191 µs |
| 2047×2047, K=128 | 1,056 µs | 5,441 µs |
| 2047×4095, K=1 | 443 µs | 2,147 µs |
| 4095×4095, K=1 | 752 µs | 4,045 µs |

Increasing the cuBLASLt workspace from 4 MiB to 32 MiB did not remove this regression. These results show that the current Pascal-wide policy is not suitable for production.

### Apple CPU

The synchronized CPU-only build passed 244/244 test cases and 3238/3238 assertions. This is portability validation, not CUDA validation.

## Limitations and status

- The current Pascal-wide padding policy is known to regress on the tested P2000 and must not be treated as a general Pascal fix.
- Turing and Blackwell did not show a benefit from padding.
- The affected GPU from issue #2712 remains unknown.
- The exact repeated issue-loop baseline could not be measured on the P2000 because the temporary Pascal compatibility build crashed in the benchmark harness; the individual-GEMM matrix above is valid.
- `git diff --check` passes.

This PR remains draft-only and is not merge-ready. Further work requires reproducing the original slowdown on the affected GPU, then narrowing the workaround to that validated configuration without regressing other architectures.

### DIFF
--- a/mlx/backend/cuda/matmul.cpp
+++ b/mlx/backend/cuda/matmul.cpp
@@ -93,7 +93,14 @@ void gemm_and_bias(
   // Safe fix: zero-pad inputs, run GEMM on aligned dimensions, slice back.
   int Mp = round_up_align(M);
   int Np = round_up_align(N);
-  bool needs_padding = (Mp != M) || (Np != N);
+
+  // Blackwell's cuBLASLt kernels do not exhibit the non-aligned-dimension
+  // regression that motivated this workaround for sufficiently large K.
+  // Avoiding the two input padding copies matters there because they cost
+  // more than the non-aligned GEMM itself.  Keep padding for small K values,
+  // where cuBLASLt may not find a native algorithm at all.
+  bool needs_padding = ((Mp != M) || (Np != N)) &&
+      (encoder.device().compute_capability_major() < 12 || K < 64);
 
   array a_work = a;
   array b_work = b;
--- a/mlx/backend/cuda/matmul.cpp
+++ b/mlx/backend/cuda/matmul.cpp
@@ -162,8 +162,8 @@ void gemm_and_bias(
   auto [batch_shape, a_batch_strides, b_batch_strides] =
       collapse_batches(a_work, b_work);
 
-  int Mwork = Mp;
-  int Nwork = Np;
+  int Mwork = needs_padding ? Mp : M;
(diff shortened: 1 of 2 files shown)


zcbenz:
I'm closing this PR since it is not showing performance gain, and there is no need to do padding ourselves: https://github.com/ml-explore/mlx/issues/2712#issuecomment-5185504133 .

PR #3978, opened 2026-08-02
Add Stream and Event APIs with Metal timing

XXXXRT666:
## Proposed changes

This PR adds Python APIs for stream management, event synchronization, and
Metal GPU timing.

- Add Stream construction, context-manager support, completion queries, and
  synchronization methods.
- Add a GPU Event API with recording, waiting, querying, synchronization, and
  elapsed-time measurement.
- Implement Metal event ordering with `MTLSharedEvent` and timing with
  command-buffer `GPUEndTime`.
- Document the interaction between events, streams, and MLX lazy evaluation.
- Add C++ and Python tests for the new APIs.

CUDA Event functionality is not implemented. The CUDA and no-GPU backend
changes only report unsupported operations through the shared interface.

## Checklist

### DIFF
--- a/docs/src/python/devices_and_streams.rst
+++ b/docs/src/python/devices_and_streams.rst
@@ -10,6 +10,7 @@ Devices and Streams
 
    Device
    Stream
+   Event
    default_device
    set_default_device
    default_stream
--- a/docs/src/usage/using_streams.rst
+++ b/docs/src/usage/using_streams.rst
@@ -1,18 +1,111 @@
 .. _using_streams:
 
-Using Streams
-=============
+Using Streams and Events
+========================
 
 .. currentmodule:: mlx.core
 
-Specifying the :obj:`Stream`
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Specifying a :obj:`Stream`
+--------------------------
 
 All operations (including random number generation) take an optional
-keyword argument ``stream``. The ``stream`` kwarg specifies which
-:obj:`Stream` the operation should run on. If the stream is unspecified then
-the operation is run on the default stream of the default device:
-``mx.default_stream(mx.default_device())``.  The ``stream`` kwarg can also
-be a :obj:`Device` (e.g. ``stream=my_device``) in which case the operation is
-run on the default stream of the provided device
-``mx.default_stream(my_device)``.
+keyword argument ``stream``. The argument specifies the :obj:`Stream` on which
+the operation runs. If it is unspecified, the operation uses the default stream
+of the default device, ``mx.default_stream(mx.default_device())``. A
+:obj:`Device` can also be used as the ``stream`` argument, in which case the
+operation uses that device's default stream.
+
+Streams are in-order queues. Evaluated operations on the same stream run in
+order, while operations on different streams may overlap. Create a stream
+directly with :class:`Stream`, or use a stream as a context manager to make it
+the default for operations created in the context:
+
+.. code-block:: python
+
+   compute_stream = mx.Stream(mx.gpu)
+
+   with compute_stream:
+       y = mx.exp(x)
+       mx.async_eval(y)
+
+Using ``with mx.stream(compute_stream):`` is equivalent to using
+``with compute_stream:``.
+
+.. note::
+
+   An event records a point in a stream's GPU execution timeline; it does not
+   evaluate lazy arrays. :meth:`Stream.record_event` submits this point even if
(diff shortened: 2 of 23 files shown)


zcbenz (commented):
I do not oppose adding Event APIs since they are useful for measuring kernel time, but the APIs would be very tricky to implement in CUDA backend because there are 2 different kinds of events due to CPU/GPU synchronization. So I think we should not add such APIs without experimenting the CUDA implementations first.

Also the PyTorch style Event API is quite light wrapper of the native event APIs, it would be cumbersome to use for profiling graphs in my opinion. It should be possible to have some better APIs like:

```python
a = ...
b = ...
start = mx.timer_start(a, b)
c = layer(a, b)
end = mx.timer_end(start, c)

mx.eval(c, end)
print(mx.elapsed_time(end))
```

So you could just measure the different parts of a graph in one run. Though it is just a rough idea and I didn't think more about the details.

zcbenz:
I'm closing this PR since we are not sure on adding an Event API. We definitely need much ways for profiling, but it could be exporting trace files for external viewers like [what JAX does](https://docs.jax.dev/en/latest/profiling.html), or some specialized profiling APIs. The PyTorch style Event API is hardly the best choice. This is something that should be discussed before starting a PR.

XXXXRT666:
> It should be possible to have some better APIs like:
> 
> ```python
> a = ...
> b = ...
> start = mx.timer_start(a, b)
> c = layer(a, b)
> end = mx.timer_end(start, c)
> 
> mx.eval(c, end)
> print(mx.elapsed_time(end))
> ```
> 
> So you could just measure the different parts of a graph in one run. Though it is just a rough idea and I didn't think more about the details.

I see. What do you think about this piece of code? I have implemented it for both CUDA and Metal. The timer starts measuring only after both x and y have been evaluated following `timer.start()`, and stops after both a and b have been evaluated, ensuring that the evaluation of `tmp` does not affect the measured time.

```python
import mlx.core as mx

x = mx.random.uniform(shape=(1024, 1024))
y = mx.random.uniform(shape=(1024, 1024))
mx.eval(x, y)

timer = mx.Timer()
x, y = timer.start(x, y)

tmp = mx.random.uniform(shape=(1024, 1024))
tmp = tmp @ tmp
mx.eval(tmp)

a = x @ y
b = some_func(y)
a, b = timer.stop(a, b)

mx.async_eval(a, b) # can be replaced by mx.eval
print(timer.elapsed_time())
```

XXXXRT666:
> I'm closing this PR since we are not sure on adding an Event API. We definitely need much ways for profiling, but it could be exporting trace files for external viewers like [what JAX does](https://docs.jax.dev/en/latest/profiling.html), or some specialized profiling APIs. The PyTorch style Event API is hardly the best choice. This is something that should be discussed before starting a PR.

I proposed this PR mainly to enable more accurate timing for small ops or function-level workloads. I have run some benchmarks to evaluate the overhead and timing difference.

The results show that for small workloads, using `perf_counter + eval` introduces a noticeable overhead, especially on `CUDA` where kernel execution time is very short. The GPU timer can measure the actual execution time more accurately without including unrelated synchronization or evaluation overhead.

<details>
<summary>Apple M4</summary>

Workload | GPU Timer | perf_counter + eval | Additional Wall Time | Ratio
-- | -- | -- | -- | --
1024² Matmul | 1.047 ms | 1.310 ms | 0.263 ms | 1.25×
MLP (4096) | 1.433 ms | 1.721 ms | 0.288 ms | 1.20×
Unfused Attention | 0.973 ms | 1.287 ms | 0.314 ms | 1.32×
Transformer Layer | 2.504 ms | 2.908 ms | 0.403 ms | 1.16×

</details>

<details>
<summary>CUDA: RTX 5070 / CUDA 13.0</summary>

Workload | GPU Timer | perf_counter + eval | Additional Wall Time | Ratio
-- | -- | -- | -- | --
1024² Matmul | 0.040 ms | 0.100 ms | 0.060 ms | 2.51×
MLP (4096) | 0.089 ms | 0.169 ms | 0.079 ms | 1.89×
Unfused Attention | 0.079 ms | 0.211 ms | 0.133 ms | 2.68×
Transformer Layer | 0.191 ms | 0.412 ms | 0.221 ms | 2.16×

</details>

zcbenz:
I think the timer API you proposed looks very promising, if you want to open a pull request I can keep it open for discussion. Just notice that profiling API is currently not on our roadmap and it requires a lot of discussions to be sure about the API, so it will take quite a while before we can review it.

PR #3983, opened 2026-08-02
Fix matmul precision with broadcast RHS

### LINKED ISSUE #3953: float32 matmul loses precision to float16 when the right operand has a broadcast batch dimension (also corrupts batch-1 SDPA fallback)
### Summary

`mx.matmul` on **float32** inputs silently loses precision to **float16** when the right operand has a broadcast (stride-0) batch dimension. Multiplying by an exact `1.0` changes the values:

```python
import mlx.core as mx
mx.random.seed(9)
v    = mx.random.normal((1, 1, 1, 512)).astype(mx.float32)
ones = mx.ones((1, 16, 1, 1)).astype(mx.float32)
out  = ones @ v            # K = 1: out must equal v exactly
print(mx.abs(out - v).max().item())   # 1.9e-03  (should be 0.0)
```

Every output element lands on the fp16 grid. Two confirmations:

```python
# fp16 ulp at 1.0 is 2^-10
b = mx.full((1,1,1,512), 1 + 2**-10).astype(mx.float32)   # fp16-representable
print(mx.abs(ones @ b - (1 + 2**-10)).max().item())        # 0.0
b = mx.full((1,1,1,512), 1 + 2**-12).astype(mx.float32)   # NOT representable
print(mx.abs(ones @ b - (1 + 2**-12)).max().item())        # 2.441406e-04 == 2^-12 exactly
```

### Shape dependence

| shape | broadcast? | max err |
|---|---|---|
| `[1,1] @ [1,512]` | – | 0.0 |
| `[1,1,1,1] @ [1,1,1,512]` | no | 0.0 |
| `[1,16,1,1] @ [1,1,1,512]` | **yes** | **1.9e-03** |
| `[1,16,1,1] @ [1,16,1,512]` (materialized) | no | 0.0 |
| `[2,16,1,1] @ [1,1,1,512]` | **yes** | **1.9e-03** |

Materializing the broadcast (`mx.contiguous(mx.broadcast_to(...))`) restores exactness, so the loss is specific to the stride-0 batched path.

### Real-world impact: `fast.scaled_dot_product_attention` at batch 1

For head dims outside the fused kernels' support list, SDPA runs the unfused fallback, whose `softmax @ values` matmul hits this path via the GQA `expand_dims` broadcast. At L=1 the output must equal V exactly (softmax of a single score is 1); it does not:

```python
mx.random.seed(9)
q = mx.random.normal((1, 16, 1, 512)).astype(mx.float32)
k = mx.random.normal((1, 1, 1, 512)).astype(mx.float32)
v = mx.random.normal((1, 1, 1, 512)).astype(mx.float32)
out = mx.fast.scaled_dot_product_attention(q, k, v, scale=1.0, mask="causal")
print(mx.abs(out - v).max().item())   # ~1.9e-03; 20/20 random seeds deviate
```

At B >= 2 the same call is exact (different matmul dispatch). Batch 1 is the single-user interactive case, so every batch-1 decode step through the fallback carries fp16-scale noise on f32 activations — we found it because two independent attention implementations disagreed only at batch 1, and the L=1 ground-truth case (out must equal V, no arithmetic involved) showed which side was wrong.

### Environment

- MLX 0.32.0 (pip), Python 3.14
- Apple M5 Max, macOS 26 (Darwin 25.5.0)

Happy to test patches. Related report from the same project: #3949 (rope batch rows, now fixed in 0.32.0 as far as we can measure — thank you!).

Sarves1911:
## Proposed changes

Fixes #3953.

The Metal matmul backend was collapsing a broadcast batch dimension into `M` even when the original operation had `M == 1`.

For a float32 batched matrix-vector product with a stride-zero broadcasted right-hand operand, this changed the operation from the GEMV path to the TF32-enabled GEMM path. As a result, multiplying by exactly `1.0` could reduce the precision of the right-hand operand.

This change prevents the batch-into-`M` collapse when the original `M` dimension is `1`, allowing the operation to remain on the existing GEMV path.

A regression test is included that runs the operation with TF32 enabled in a separate process.

## Testing

* Focused regression test passed:
  `PYTHONPATH=python/tests python -m unittest test_blas.TestBlas.test_matmul_broadcast_rhs_k1_precision -v`
* Complete BLAS test suite passed:
  `PYTHONPATH=python/tests python -m unittest test_blas -v`
* 29 BLAS tests passed.
* Pre-commit hooks passed for the modified files.
* The original issue reproduction now reports a maximum error of `0.0` with `MLX_ENABLE_TF32=1`.

## Checklist

* [x] I have read the [CONTRIBUTING](https://github.com/ml-explore/mlx/blob/main/CONTRIBUTING.md) document
* [x] I have run `pre-commit run --all-files` to format my code / installed pre-commit prior to committing changes
* [x] I have added tests that prove my fix is effective or that my feature works
* [x] I have updated the necessary documentation (if needed)

### DIFF
--- a/mlx/backend/metal/matmul.cpp
+++ b/mlx/backend/metal/matmul.cpp
@@ -1441,7 +1441,7 @@ void Matmul::eval_gpu(const std::vector<array>& inputs, array& out) {
   auto batch_size_out = out.size() / (size_t(M) * size_t(N));
 
   // Collapse batches into M if needed
-  if (batch_size_out > 1 && !a_transposed && batch_shape.size() == 1 &&
+  if (M > 1 && batch_size_out > 1 && !a_transposed && batch_shape.size() == 1 &&
       a.strides()[a.ndim() - 2] == K && A_batch_stride.back() == M * K &&
       B_batch_stride.back() == 0) {
     M *= batch_shape.back();
--- a/python/tests/test_blas.py
+++ b/python/tests/test_blas.py
@@ -1,6 +1,9 @@
 # Copyright © 2023-2024 Apple Inc.
 
 import math
+import os
+import subprocess
+import sys
 import unittest
 from itertools import permutations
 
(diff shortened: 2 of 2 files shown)


zcbenz:
Closing per https://github.com/ml-explore/mlx/issues/3953#issuecomment-5185582688.

PR #3984, opened 2026-08-02
added __complex__ support

aaishwarymishra:
This pull request adds support for converting `mx::array` objects to Python complex numbers by implementing the 

`__complex__` method. It also extends the test suite to verify this new functionality, ensuring that both real and complex arrays behave as expected when converted to complex numbers.

#### New feature: Python complex conversion support

* Added a `__complex__` method to the `mx::array` Python bindings, allowing `mx::array` instances to be converted to Python complex numbers using the built-in `complex()` function.

#### Test improvements

* Extended the `test_to_scalar` test to check that `complex(mx.array(...))` returns the correct complex value for real and complex arrays, and raises a `ValueError` for non-scalar arrays.

## Checklist

Part of the https://github.com/data-apis/array-api-compat/issues/452 .

### DIFF
--- a/python/src/array.cpp
+++ b/python/src/array.cpp
@@ -1035,11 +1035,7 @@ void init_array(nb::module_& m) {
       .def(
           "__complex__",
           [](mx::array& a) {
-            auto value = PyComplex_AsCComplex(to_scalar(a).ptr());
-            if (PyErr_Occurred()) {
-              throw nb::python_error();
-            }
-            return std::complex<double>(value.real, value.imag);
+            return nb::cast<std::complex<double>>(to_scalar(a));
           })
       .def(
           "__format__",
--- a/python/src/array.cpp
+++ b/python/src/array.cpp
@@ -1036,6 +1036,9 @@ void init_array(nb::module_& m) {
           "__complex__",
           [](mx::array& a) {
             auto value = PyComplex_AsCComplex(to_scalar(a).ptr());
+            if (PyErr_Occurred()) {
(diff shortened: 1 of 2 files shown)


aaishwarymishra:
While working on test my linter were showing me warning that complex is not valid input for mlx array do you think we need to update mlx array arguments?

python/src/array.cpp
```diff
@@ -1025,6 +1025,12 @@ void init_array(nb::module_& m) {
           nb::rv_policy::none)
       .def("__int__", [](mx::array& a) { return nb::int_(to_scalar(a)); })
       .def("__float__", [](mx::array& a) { return nb::float_(to_scalar(a)); })
+      .def(
+          "__complex__",
+          [](mx::array& a) {
+            auto value = PyComplex_AsCComplex(to_scalar(a).ptr());
```
zcbenz: Can it be done with just `return nb::handle(PyComplex_AsCComplex(to_scalar(a).ptr()))`?

zcbenz:
> While working on test my linter were showing me warning that complex is not valid input for mlx array do you think we need to update mlx array arguments?

I'm open to support complex as array input.

python/src/array.cpp
aaishwarymishra: Hi, I have checked as nb does not provide complex binding we need to convert it to c complex value first which nb automatically casts to python complex.

I ran your suggested code, it failed to compiler :(

zcbenz (approved):

PR #3791, opened 2026-07-02
Raise qmv batch limit for large matrices on M5-class GPUs

pierre427:
## Proposed changes

`get_qmv_batch_limit` currently sends gen-17 (M5-class) GPUs down the generic
non-`'d'` fallback, which caps the qmv/qmv_wide path at M<10 for large
matrices. Measured on an M5 Max (`applegpu_g17s`), the `qmv_wide` kernels stay
ahead of the `qmm` path well past that limit on large shapes. This PR raises
the large-matrix limit to 16 for `arch_gen >= 17` non-`'d'` devices; small and
mid shapes keep the existing fallback values (no data to justify changing
them).

This mitigates the small-M quantized-matmul cost step discussed in #3553 for
the M=11–16 band, which matters for speculative decoding: verification of
multi-candidate / tree / long prompt-lookup proposals lands exactly in that
window. #2031 notes these limits are empirical and machine-dependent; gen-17
hardware postdates the tunings there.

## Measurements

Apple M5 Max, 128GB (`applegpu_g17s`), macOS 26. Whole-model forward cost vs
row count M against a warm 400-token KV cache, min of 6 reps. Model: 70B dense
llama-arch (LLM360 K2-V2), affine 4-bit, group_size 64 (D=8192; O spans 1024 /
8192 / 28672 / 250112 across layers).

| M | main @ e9463bbf (limit 10) | this PR (limit 16) |
|---|---|---|
| 1 | 76.4 ms | 75.9 ms |
| 3 | 85.0 | 84.8 |
| 5 | 108.2 | 112.7 |
| 8 | 167.9 | 170.8 |
| 12 | **294.6** (qmm) | **241.5** (qmv_wide, −18%) |
| 16 | — | 296.0 (qmv_wide) |

Same model at affine 8-bit: M=12 326.7 → 304.1 (−7%).

For reference, the 0.31.2 release (no qmv_wide kernels) costs 298.5 ms at
M=12 on the same 4-bit model, and end-to-end speculative decoding on this
hardware improves from ~18 tok/s (release) to 24.5 tok/s (main + k=3 chain
with a 0.6B draft), with the M=11–16 window enabling wider verification
schemes this change dispatches to the faster path.

## Repro

```python
import time
import mlx.core as mx
from mlx_lm import load
from mlx_lm.models.cache import make_prompt_cache

model, _ = load("<4-bit 70B model>")
caches = make_prompt_cache(model)
[...]

PR #3987, opened 2026-08-03
Make get_qmv_batch_limit consider which vector kernel it guards

katlun-lgtm:
## Proposed changes

Follow-up from #3863 (@zcbenz: "get_qmv_batch_limit does not consider use_qmv_wide").

The limit picks the vector→matmul crossover from (D, O) and the architecture alone, while the vector kernel it guards differs by quantization mode (`use_qmv_wide`: fp modes everywhere, affine on gen 15+, plain `qmv` otherwise). This passes the mode in and retunes the one cell I can measure — affine with `qmv_wide` active, gen-15 non-`d` — leaving fp modes, the gen ≤ 14 tables, and the `d` tiers unchanged for measurements on other hardware.

Measured crossovers (M3 Max `g15s`, both routings forced at the same M via a local env-override build, cold weights round-robin so nothing stays cache-resident, min of 8 trials × 30 reps, M ∈ 3–20, 4-bit gs 128, bf16 activations):

| shape | D·O | old limit | measured crossover | new limit |
|---|---|---|---|---|
| 2048×2048 | 4.2M | 18 | 17 | 17 |
| 2048×4096 | 8.4M | 12 | 16 | 16 |
| 3072×4096 | 12.6M | 12 | 13 | 13 |
| 4096×4096 | 16.8M | 12 | 13 | 13 |
| 3072×8192 | 25.2M | 10 | 13 | 13 |
| 4096×8192 | 33.6M | 10 | 13 | 13 |
| 2048×16384 | 33.6M | 10 | 13 | 13 |
| 4096×14336 | 58.7M | 10 | 13 | 13 |
| 5120×17408 | 89.1M | 10 | 13 | 13 |

The crossover tracks D·O for affine — 4096×8192 and 2048×16384 (same D·O) cross at the same M — and settles on a floor of 13 that holds across a 5× range of sizes. Below it qmv_wide's margin is large (1.2–3.5× at M ≤ 10 on these shapes); above it qmm pulls ahead steadily (1.5–1.7× by M=20).

End-to-end effect (default routing, this branch vs 0.32.0, same harness):

| shape class | M=10 | M=11–12 | elsewhere |
|---|---|---|---|
| ≥ 25M elements (4 shapes) | **1.22–1.25×** | 1.04–1.08× | parity |
| 4096×4096 | — | **1.11×** at M=12 | parity |
| 2048×2048 | — | — | 1.10× at M=17, parity else |

No cell regresses beyond ±2% noise (full sweep M ∈ 3–20 × 7 shapes).

I ran mxfp4 through the same harness: its crossover is K-dependent rather than D·O-dependent (4096×8192 crosses at 11 where 2048×16384 crosses at 13), so fp wants its own cell rather than inheriting this one — data below if useful.

<details>
<summary>mxfp4 crossovers (same harness)</summary>

2048×2048: 17 · 4096×4096: 13 · 3072×8192: 13 · 4096×8192: 11 · 2048×16384: 13 · 4096×14336: 11 · 5120×17408: 11
</details>

One correction to the sweep I posted on #3863: it put the large-shape crossover near 7. This finer forced-routing sweep on current main puts the floor at 13; the earlier number was inconsistent with its own native-qmv_wide measurement at M=10 (0.49 ms measured, vs the ~0.72 ms at M=9 the crossover claim implied). The table above supersedes it.

`test_quantized.py` passes in full (32 tests, 2896 subtests). Routing-only change — both kernels are already covered by those tests.

## Checklist

- [x] I have run `pre-commit run --all-files` to format my code / installed pre-commit prior to committing changes (clang-format v21.1.8 as pinned)
- [ ] I have added tests that prove my fix is effective or that my feature works (routing-only change; existing `test_quantized.py` covers both kernels)

### DIFF
--- a/mlx/backend/metal/quantized.cpp
+++ b/mlx/backend/metal/quantized.cpp
@@ -92,17 +92,19 @@ get_qmv_batch_limit(int D, int O, const std::string& mode, metal::Device& d) {
   auto arch_gen = d.get_architecture_gen();
   // Measured on M3 Max (g15s) for affine mode with qmv_wide active: the
   // qmv_wide -> qmm crossover tracks D * O rather than the dims separately
-  // (4096x4096 and 2048x8192 cross at the same M), and sits well below the
-  // old tiers on MLP-sized layers. fp modes and the gen <= 14 tables are
-  // left unchanged pending measurements on those paths.
+  // (same-D*O shapes cross at the same M), falling from 17 at 2048x2048 to
+  // a floor of 13 that holds from ~12M elements through the largest shapes
+  // measured (5120x17408). fp modes and the gen <= 14 tables are left
+  // unchanged pending measurements on those paths (fp_qmv_wide's crossover
+  // is K-dependent, so it needs its own cell, not this one).
   if (mode == "affine" && arch_gen >= 15 && arch_size != 'd') {
     int64_t elements = int64_t(D) * O;
     if (elements <= int64_t(2048) * 2048) {
       return 17;
-    } else if (elements <= int64_t(4096) * 4096) {
-      return 13;
+    } else if (elements <= int64_t(2048) * 4096) {
+      return 16;
     } else {
-      return 7;
+      return 13;
     }
   }
   if (arch_gen == 13 || arch_gen == 14) {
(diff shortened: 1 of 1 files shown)


zcbenz:
Thanks for the PR, this has a lot overlap with #3791 and I think we will focus on it first.

katlun-lgtm:
Makes sense — #3791 first. If the gen-15 crossover data from the #3863 thread is useful when the other tiers get tuned, it's all there.

PR #3988, opened 2026-08-03
Fix build error caused by TMA macro guard

zcbenz:
The [release build is failing](https://github.com/ml-explore/mlx/actions/runs/30613074114/job/91100039025) because the macros guarding TMA related PTX code are not strictly correct, reuse the macro guard of cutlass to save time defining our own.

### DIFF
--- a/mlx/backend/cuda/ptx.cuh
+++ b/mlx/backend/cuda/ptx.cuh
@@ -1,14 +1,12 @@
 #pragma once
 
-#include <cuda.h>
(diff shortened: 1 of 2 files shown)


nastya236:
Thank you @zcbenz ! Not sure why it was missed #3157

nastya236 (approved):

zcbenz:
The daily tests do not build for all cuda arches so failures for specific arches only happen in nightly build where all supported targets would be built.

PR #3991, opened 2026-08-03
Fix uniform distribution bias in randint (#3926)

### LINKED ISSUE #3926: random.randint loses integer range semantics through float32

alphapibeta:
## Summary

Fixes #3926 by sampling integer ranges directly instead of routing `mx.random.randint` through float32 uniform sampling.

## Root cause

The previous implementation converted integer bounds to float32, sampled `uniform(low, high, ...)`, and cast the result. Float32 has a 24-bit significand, so ranges above `2**24` could return the excluded upper bound, omit valid integers, or collapse wide int64 intervals to a constant. Boolean output could also lose `1` because `high=2` was cast to bool before sampling.

## Implementation

- Generate integer random words with the existing Threefry machinery.
- Use unsigned-width rejection sampling to preserve `[low, high)` semantics and avoid modulo bias.
- Compare empty ranges in the output integer domain.
- Avoid signed overflow for negative, wide, and near-full-width ranges.
- Preserve scalar and array bounds, broadcasting, shapes, streams, deterministic keys, signed and unsigned integer dtypes, and bool behavior.

## Files changed

- `mlx/random.cpp`
- `mlx/primitives.h`
- `mlx/primitives.cpp`
- `mlx/backend/cpu/primitives.cpp`
- `mlx/backend/cuda/random.cu`
- `mlx/backend/metal/kernels/random.metal`
- `mlx/backend/metal/primitives.cpp`
- `mlx/backend/no_gpu/primitives.cpp`
- `tests/random_tests.cpp`
- `python/tests/test_random.py`

## Validation

- **RTX 2060 CUDA, SM 7.5 (`eth`):** focused randint test passed with 20/20 assertions; full CUDA test binary passed with 260/260 tests and 3385/3385 assertions.
- **Mac CPU:** focused C++ randint test passed with 20/20 assertions.
- **Mac Metal:** focused C++ randint test passed with 20/20 assertions.
- **Mac Python bindings:** `python/tests/test_random.py -k randint` passed with 2 tests; 13 were deselected.
- **Mac full suite:** 256/260 test cases passed; the four failures were optional GGUF/safetensors tests disabled by configuration. All 3432 executed assertions passed.
- **P2000:** the full MLX build is blocked by the pre-existing SM 6.1 `event.cu` atomic-wait error. The changed `random.cu` compiles for `compute_61/sm_61`; no P2000 runtime result is claimed.

## Checks and limitations

- Final pushed commit: `170a604a0b8d90f0dca2f0addddfdb00c991e39c`.
- `clang-format --dry-run --Werror` passes for all changed files.
- `git diff --check` passes.
- The repository's `pre-commit` command was unavailable on the validation hosts, so the hooks were not run.
- No benchmark was added because this is a correctness fix; no performance claim is made.
- This PR remains Draft pending maintainer review.


zcbenz:
Closing per https://github.com/ml-explore/mlx/issues/3926#issuecomment-5187326553.

PR #3999, opened 2026-08-04
Increase ccache size in CI

zcbenz:
When the cache is full and `ccache` does cleanup, it does not necessarily remove the oldest cache because it does not keep precise records. So if the cache limit is not big enough `ccache` would keep erasing cache of recent build and end up with bad cache hit rate.


nastya236:
thanks!

nastya236 (approved):

PR #4002, opened 2026-08-04
Fix mlx.launch --python: flag is parsed but never forwarded to the launch script

jonathan308:
### Proposed changes

`mlx.launch --python` is documented and parsed, but has no effect: #3536 removed automatic interpreter prepending and added the flag for users who still need it, but `RemoteProcess` never forwards it to `make_launch_script`, so the launched command always runs without an interpreter prefix on the remote side.

This threads the flag through `RemoteProcess` → `make_launch_script`. The no-prepend default from #3536 is unchanged; an explicit `--python` is now honored.

### Checklist

- [x] I have run `black`/`isort` on the touched Python files (clean; no C++ touched)
- [x] I have added tests (`python/tests/test_launch.py`) covering the default no-prepend path, an explicit `--python`, and interpreter-path handling


zcbenz (approved):
Thanks!

Issue #3994, opened 2026-08-04
mx.divmod truncates while `//` and `%` floor, and on Metal `mx.divmod` is internally inconsistent (`q*b + r != a`)


sashko-zakharchuk:
## Summary

`mx.divmod` rounds its quotient toward zero (truncates), but `mx.floor_divide` (for floats) and `mx.remainder` round toward negative infinity (floor). This produces several inconsistencies:

1. `mx.divmod(a, b) != (a // b, a % b)` for mixed-sign operands, even though the docstring says divmod "is equivalent to but faster than `(a // b, a % b)`" (`python/src/ops.cpp:458-471`).
2. `mx.divmod` disagrees with `numpy.divmod` / Python `divmod` on negatives (both floor).
3. On the Metal backend, `mx.divmod` is internally inconsistent: `q*b + r != a`, and its remainder disagrees with the CPU backend for the same call.
4. For integers, `mx.floor_divide` truncates while `mx.remainder` floors, so `(a // b) * b + (a % b) != a`.

Measured on MLX `0.32.1.dev20260803`, numpy `2.5.1`: CPU rows on Linux x86, GPU rows on Apple Metal (M-series).

## Matrix (mixed-sign cases, measured)

| dtype | a | b | backend | `mx.divmod` | `q*b+r == a`? | `(a//b, a%b)` in MLX | `numpy.divmod` |
|---|---|---|---|---|---|---|---|
| float32 | -7 | 3 | CPU | `(-2, -1)` | yes | `(-3, 2)` | `(-3, 2)` |
| float32 | -7 | 3 | Metal | `(-2, 2)` | **no** (`-4`) | `(-3, 2)` | `(-3, 2)` |
| float32 | 7 | -3 | CPU | `(-2, 1)` | yes | `(-3, -2)` | `(-3, -2)` |
| float32 | 7 | -3 | Metal | `(-2, -2)` | **no** (`4`) | `(-3, -2)` | `(-3, -2)` |
| int32 | -7 | 3 | CPU | `(-2, -1)` | yes | `(-2, 2)` | `(-3, 2)` |
| int32 | -7 | 3 | Metal | `(-2, 2)` | **no** (`-4`) | `(-2, 2)` | `(-3, 2)` |

Positive/positive and negative/negative cases all agree, so this only bites on mixed signs.

Three problems:

[...]

PR #4003, opened 2026-08-04
Fix fp16/bf16 divmod and floor_divide precision (divide in float before rounding)

sashko-zakharchuk:
On fp16/bf16, `mx.divmod` and `mx.floor_divide` divided `x / y` in the input precision
and narrowed the quotient to fp16/bf16 before `trunc`/`floor`. A true quotient just
below an integer rounds up when narrowed, so the result is one too high.

```python
a = mx.array([86.5625], dtype=mx.float16)    # a / b = 2.99946
b = mx.array([28.859375], dtype=mx.float16)
mx.divmod(a, b)[0]        # before: 3    after: 2
mx.floor_divide(a, b)     # before: 3    after: 2
```

The fix computes the division in float32 for the half types before rounding: on CPU in
`DivMod`'s `trunc`/`fmod` op, on Metal and CUDA in the `FloorDivide` functor that
`divmod` reuses, and in the `floor_divide` lowering in `ops.cpp`. float32, float64,
integer, and sign behavior are untouched (`divmod` still truncates). For non-negative
fp16/bf16 operands `divmod` and `floor_divide` now agree with each other, and with numpy
for same-dtype operands; for negatives `divmod` still truncates while `floor_divide`
floors.

This is also why `test_divmod` is flaky: it compares fp16 `mx.divmod` against numpy with
unseeded random inputs, and near an integer boundary they disagreed (about 0.04% of
fp16 pairs, more for bf16). With the division in float they match. The test gets a
deterministic boundary case for fp16 and bf16, plus a seeded bf16 batch checked against
a float32 reference (numpy has no bf16).

Part of #3994 (the fp16/bf16 precision part); the floored-vs-truncated sign defects are
a separate change.

## Verification

- CPU (Linux x86): fp16/bf16 divmod and floor_divide match a float-wide reference on
  0/200000 random pairs; `1e300 / 1e299` float64 stays `10` (no downcast); `-7 / 3`
  stays `(-2, -1)` (sign unchanged); full `test_ops.py` passes.
- CUDA (RTX 5050, sm_120): fp16/bf16 divmod and floor_divide give `2` at the boundary,
  0/100000 divergence, sign unchanged.
- Metal (M3 Ultra): fp16/bf16 divmod and floor_divide give `2` at the boundary,
  0/100000 divergence, sign unchanged; `test_divmod` passes.

### DIFF
--- a/mlx/backend/cpu/binary.cpp
+++ b/mlx/backend/cpu/binary.cpp
@@ -48,7 +48,14 @@ void DivMod::eval_cpu(
       return std::make_pair(x / y, x % y);
     };
     auto float_op = [](auto x, auto y) {
-      return std::make_pair(std::trunc(x / y), std::fmod(x, y));
+      // fp16/bf16: divide in float so the quotient isn't narrowed before trunc
+      using T = std::decay_t<decltype(x)>;
+      using C = std::conditional_t<sizeof(T) == 2, float, T>;
+      auto xc = static_cast<C>(x);
+      auto yc = static_cast<C>(y);
+      return std::make_pair(
+          static_cast<T>(std::trunc(xc / yc)),
+          static_cast<T>(std::fmod(xc, yc)));
     };
 
     switch (out_a.dtype()) {
--- a/mlx/backend/cuda/device/binary_ops.cuh
+++ b/mlx/backend/cuda/device/binary_ops.cuh
@@ -18,6 +18,10 @@ struct FloorDivide {
   __device__ T operator()(T x, T y) {
     if constexpr (cuda::std::is_integral_v<T>) {
       return x / y;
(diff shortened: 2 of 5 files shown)


zcbenz (commented):
Since this is going to add overhead I think we should follow what PyTorch does, which has the same behavior:

```python
a = torch.tensor([86.5625], dtype=torch.float16)
b = torch.tensor([28.859375], dtype=torch.float16)

print(torch.div(a, b), torch.remainder(a, b))
```

```console
tensor([3.], dtype=torch.float16) tensor([28.8438], dtype=torch.float16)
```

Generally when in doubt we would prefer speed over precision as long as there is a way for users to use high precision computation.

Issue #3830, opened 2026-07-10
[BUG] Distributed pipeline generation: Metal fence handoff deadlocks under MLX_METAL_FAST_SYNCH=1 (TCP ring; orphaned fence_wait kernel locks GPU until reboot) and hits the ~5s GPU watchdog when unset (kIOGPUCommandBufferCallbackErrorTimeout at ~7.3k tokens)


geekybraindev:
## Summary
Running pipeline-style distributed inference over the **plain TCP ring backend** silently deadlocks when large activation tensors cross the GPU-stream ↔ CPU-communication-stream fence handoff with `MLX_METAL_FAST_SYNCH=1`. Both ranks sit at ~100% CPU forever, memory flat, no error. Killing the wedged processes releases their memory but leaves the GPU compute engine occupied by the orphaned `fence_wait` spin kernel — after that, **any** Metal eval from **any** process on the machine hangs, and only a reboot recovers (in one instance the machine escalated to a full userspace freeze). We believe this is the same family as #3142 / #3141, but note the fence layer is transport-independent (`ring.cpp` and `jaccl.cpp` declare the identical CPU communication stream), so the "JACCL" framing of #3142 undersells the blast radius: plain-Ethernet TCP-ring users hit it too. This matches silent multi-node hangs long reported in the wild (e.g. `mlx_lm.server --pipeline` over 10 GbE wedging mid-generation after ~2k tokens on ~60-layer models; exo-explore/exo#1666 / #1853 / #1726).

With the env var **unset**, the deadlock disappears — but long generations instead die at the macOS GPU watchdog (details below). Distributed pipeline users currently choose between an untimed spin that deadlocks and a driver watchdog that kills their process.

## Environment
- Mac Studio M2 Ultra 192 GB + MacBook Pro M5 Max 128 GB, macOS 26.5.1 (25F80)
- mlx 0.31.2, mlx-lm 0.31.3 (PyPI), Python 3.13
- Reproduces on 2 ranks on ONE machine over a loopback TCP ring, and across two Macs over 10 GbE

## Minimal repro (no model needed, ~30 s)
`fence_stress.py` (~90 lines, happy to attach): 2-rank loopback ring; per iteration it replays mlx-lm's per-token pipeline pattern — GPU compute → CPU-stream `send`/`recv_like`/`all_gather` → GPU compute, with `mx.async_eval` double-buffering exactly like `mlx_lm/generate.py`.

```
[...]

PR #4005, opened 2026-08-05
docs: Do not pass MLX_METAL_FAST_SYNCH=1 by default

katlun-lgtm:
Following up on #3830: since fast synch is not reliable and there is no fix in sight (per the discussion there and in #3142), this adds a one-time stderr warning the first time a fence is actually created in fast mode, so users hitting the wedge can find the cause without bisecting env vars. It fires on any multi-stream workload, not just distributed, matching the single-node report in #3830. One file, +12/−1, no behavior change beyond the message.

### DIFF
--- a/docs/src/usage/distributed.rst
+++ b/docs/src/usage/distributed.rst
@@ -335,8 +335,7 @@ of a gigantic model using MLX LM.
 
 .. code-block::
 
-   mlx.launch --verbose --backend jaccl --hostfile m3-ultra-jaccl.json \
-        --env MLX_METAL_FAST_SYNCH=1 -- \  # <--- important
+   mlx.launch --verbose --backend jaccl --hostfile m3-ultra-jaccl.json -- \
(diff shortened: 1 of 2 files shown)


zcbenz (commented):
I think instead of printing a warning we should remove it from the `mlx.launch` command in `distributed.rst` and only mention it in the side note as a faster but buggy flag. The plan was to make it default after fixing the hanging bug, but at the moment I don't see any hope of that and we probably would have to end up removing it.

katlun-lgtm:
That's better, done — the runtime warning is gone and `fence.cpp` is back to identical with main. The flag is dropped from the `mlx.launch` example and the side note now says it's faster but unreliable, can deadlock and leave the GPU wedged in distributed runs and in any cross-stream workload, linking #3142 and #3830.

Now docs-only, one file, +9/−4.

zcbenz (approved):

PR #4012, opened 2026-08-05
Keep randint samples inside [low, high)

### LINKED ISSUE #3926: random.randint loses integer range semantics through float32

adityasingh2400:
Fixes #3926.

`mx.random.randint` draws a float32 uniform on `[low, high)` and casts it to the output dtype. Two things go wrong on the way.

Rounding in float32 can land a sample on `high` itself, so the excluded upper bound comes back:

```python
key = mx.random.key(42)
x = mx.random.randint(2**24, 2**24 + 2, (10_000,), dtype=mx.int32, key=key)
sorted(set(x.tolist()))    # [16777216, 16777218], and 16777218 is `high`
```

The cast also truncates towards zero instead of flooring, so for an interval that spans negative values `low` is never produced and zero collects twice the weight of every other value:

```python
a = mx.random.randint(-5, 5, (100_000,), key=key)
sorted(set(a.tolist()))    # [-4, -3, -2, -1, 0, 1, 2, 3, 4], -5 is missing
```

That is the same reason `mx.random.randint(0, 1, dtype=mx.bool_)` returns `True` almost always: any nonzero float casts to `True`, and `0` is the only value the interval allows.

The fix takes the floor before the cast and clamps the result to `high - 1`. The clamp bound is computed in the integer domain rather than in float32, because `high - 1` is not always representable there. At `2**24` the float32 spacing is already 2, so clamping in float would round the bound back onto `high` and also drop `2**24 + 1`, which clamping after the cast keeps. The clamp to `low` is applied after the clamp to `high - 1` so an empty interval still collapses to `low`, which is the existing behaviour and is covered by a test.

This keeps the uniform based algorithm and only clamps, per the direction on the issue, which is also what JAX documents for `jax.random.randint`. Wide intervals and bounds beyond the float32 integer resolution still cannot reach every integer, for example `[2**40, 2**40 + 1024)` still collapses to a single value, so I added a note to the docstring saying so. Every sample is at least in range now.

Verified on an M3 Pro running macOS 26, on both the CPU and the Metal stream.

Before, on 0.31.2:

```
mlx 0.31.2  default_device=Device(gpu, 0)

[1] issue repro: bounds must hold in [low, high)
  DeviceType.cpu         int32 [2**24, 2**24+2) distinct=[16777216, 16777218] out_of_range=[16777218]
  DeviceType.cpu         int64 [2**40, 2**40+1024) n_distinct=1 out_of_range=[]
  DeviceType.gpu         int32 [2**24, 2**24+2) distinct=[16777216, 16777218] out_of_range=[16777218]
  DeviceType.gpu         int64 [2**40, 2**40+1024) n_distinct=1 out_of_range=[]

[2] negative interval [-5, 5): every integer reachable, uniform
  DeviceType.cpu         distinct=[-4, -3, -2, -1, 0, 1, 2, 3, 4] min_count=9844 max_count=19971
  DeviceType.gpu         distinct=[-4, -3, -2, -1, 0, 1, 2, 3, 4] min_count=9844 max_count=19971

[3] bool output
  DeviceType.cpu         randint(0,2,bool) true_frac=1.000 | randint(0,1,bool) any_true=True
  DeviceType.gpu         randint(0,2,bool) true_frac=1.000 | randint(0,1,bool) any_true=True

[4] behaviour that must not change
  empty interval randint(10, -10) all == 10: True
  randint(-10, 10) in bounds: True min=-9 max=9
  randint(0, 3) counts: {0: 20204, 1: 19953, 2: 19843}

[5] cpu and gpu agree bit for bit
  randint(0, 10, mlx.core.int32) cpu == gpu: True
  randint(-5, 5, mlx.core.int32) cpu == gpu: True
  randint(16777216, 16777218, mlx.core.int32) cpu == gpu: True
  randint(0, 2, mlx.core.bool) cpu == gpu: True
```

After:

```
mlx 0.32.1.dev20260805+2c46b953  default_device=Device(gpu, 0)

[1] issue repro: bounds must hold in [low, high)
  DeviceType.cpu         int32 [2**24, 2**24+2) distinct=[16777216, 16777217] out_of_range=[]
  DeviceType.cpu         int64 [2**40, 2**40+1024) n_distinct=1 out_of_range=[]
  DeviceType.gpu         int32 [2**24, 2**24+2) distinct=[16777216, 16777217] out_of_range=[]
  DeviceType.gpu         int64 [2**40, 2**40+1024) n_distinct=1 out_of_range=[]

[2] negative interval [-5, 5): every integer reachable, uniform
  DeviceType.cpu         distinct=[-5, -4, -3, -2, -1, 0, 1, 2, 3, 4] min_count=9844 max_count=10127
  DeviceType.gpu         distinct=[-5, -4, -3, -2, -1, 0, 1, 2, 3, 4] min_count=9844 max_count=10127

[3] bool output
  DeviceType.cpu         randint(0,2,bool) true_frac=0.496 | randint(0,1,bool) any_true=False
  DeviceType.gpu         randint(0,2,bool) true_frac=0.496 | randint(0,1,bool) any_true=False

[4] behaviour that must not change
  empty interval randint(10, -10) all == 10: True
  randint(-10, 10) in bounds: True min=-10 max=9
  randint(0, 3) counts: {0: 20204, 1: 19953, 2: 19843}

[5] cpu and gpu agree bit for bit
  randint(0, 10, mlx.core.int32) cpu == gpu: True
  randint(-5, 5, mlx.core.int32) cpu == gpu: True
  randint(16777216, 16777218, mlx.core.int32) cpu == gpu: True
  randint(0, 2, mlx.core.bool) cpu == gpu: True
```

I extended `test_randint` in `python/tests/test_random.py` and the C++ `test random randint` case with the bound, negative interval and boolean cases above.

### DIFF
--- a/mlx/random.cpp
+++ b/mlx/random.cpp
@@ -285,7 +285,16 @@ array randint(
         "[randint] randint only accepts integer dtypes and bool.");
   }
   auto u = uniform(low, high, shape, float32, key, s);
-  return astype(maximum(u, low, s), dtype, s);
+  // Take the floor before casting: `astype` truncates towards zero, which
+  // makes `low` unreachable and doubles the weight of zero when the interval
+  // spans negative values.
+  auto out = astype(floor(u, s), dtype, s);
+  // Rounding in the float32 domain can place a sample on `high` itself, so
+  // clamp to `high - 1`. The bound is computed in the integer domain since it
+  // is not always representable in float32. Clamping to `low` last keeps an
+  // empty interval collapsing to `low`.
+  auto hi = astype(subtract(high, array(1, high.dtype()), s), dtype, s);
+  return maximum(minimum(out, hi, s), astype(low, dtype, s), s);
 }
 
 array bernoulli(
--- a/mlx/random.cpp
+++ b/mlx/random.cpp
@@ -285,14 +285,10 @@ array randint(
         "[randint] randint only accepts integer dtypes and bool.");
   }
   auto u = uniform(low, high, shape, float32, key, s);
-  // Take the floor before casting: `astype` truncates towards zero, which
-  // makes `low` unreachable and doubles the weight of zero when the interval
-  // spans negative values.
+  // astype truncates decimal parts so -1.7 becomes -1, use floor.
   auto out = astype(floor(u, s), dtype, s);
-  // Rounding in the float32 domain can place a sample on `high` itself, so
-  // clamp to `high - 1`. The bound is computed in the integer domain since it
-  // is not always representable in float32. Clamping to `low` last keeps an
-  // empty interval collapsing to `low`.
+  // low/high may not be representable in float32 so actual range of uniform
+  // may be larger than [low, high) and we have to clamp to [low, high - 1].
   auto hi = astype(subtract(high, array(1, high.dtype()), s), dtype, s);
   return maximum(minimum(out, hi, s), astype(low, dtype, s), s);
 }
--- a/python/src/random.cpp
+++ b/python/src/random.cpp
@@ -322,6 +322,12 @@ void init_random(nb::module_& parent_module) {
         half-open interval ``[low, high)``. The lower and upper bound can be
         scalars or arrays and must be broadcastable to ``shape``.
 
+        .. note::
+           The samples are drawn from a ``float32`` uniform and clamped to
+           ``[low, high - 1]``, so not every integer in the range is reachable
+           once the bounds or the width of the interval go beyond the
+           ``2**24`` integer resolution of ``float32``.
+
         Args:
             low (scalar or array): Lower bound of the interval.
             high (scalar or array): Upper bound of the interval.
--- a/python/tests/test_random.py
+++ b/python/tests/test_random.py
@@ -233,6 +233,26 @@ class TestRandom(mlx_tests.MLXTestCase):
         a = mx.random.randint(10, -10, [1000, 1000])
         self.assertTrue(mx.all(a == 10).item())
 
+        # Bounds hold when the interval is not exactly representable in float32
+        for dtype, low, high in [
+            (mx.int32, 2**24, 2**24 + 2),
+            (mx.uint32, 2**24, 2**24 + 2),
+            (mx.int64, 2**40, 2**40 + 1024),
+        ]:
+            a = mx.random.randint(low, high, [10000], dtype=dtype, key=key)
+            self.assertTrue(mx.all(a >= low).item())
+            self.assertTrue(mx.all(a < high).item())
+
+        # The lower bound is reachable when the interval spans negative values
+        a = mx.random.randint(-5, 5, [20000], key=key)
+        self.assertEqual(sorted(set(a.tolist())), list(range(-5, 5)))
+
+        # Booleans use the whole interval
+        a = mx.random.randint(0, 2, [1000], dtype=mx.bool_, key=key)
+        self.assertEqual(sorted(set(a.tolist())), [False, True])
+        a = mx.random.randint(0, 1, [1000], dtype=mx.bool_, key=key)
+        self.assertFalse(mx.any(a).item())
+
         self.assertEqual(
             mx.random.randint(0, 1).dtype, mx.random.randint(0, 1, dtype=None).dtype
         )
--- a/tests/random_tests.cpp
+++ b/tests/random_tests.cpp
@@ -512,6 +512,19 @@ TEST_CASE("test random randint") {
   x = random::randint(low, high, {3, 3});
   CHECK(all(equal(x, array(low))).item<bool>());
 
+  // Check bounds hold when the interval is not representable in float32
+  auto ilow = 1 << 24;
+  auto ihigh = ilow + 2;
+  x = random::randint(ilow, ihigh, {10000}, int32);
+  CHECK(
+      (all(array(ilow) <= x).item<bool>() &&
+       all(x < array(ihigh)).item<bool>()));
+
+  // Check the lower bound is reachable for intervals spanning negative values
+  x = random::randint(-5, 5, {20000}, int32);
+  CHECK_EQ(min(x).item<int>(), -5);
+  CHECK_EQ(max(x).item<int>(), 4);
+
   // Check wrong key type or shape
   auto key = array({0, 0}, {1, 2});
   CHECK_THROWS_AS(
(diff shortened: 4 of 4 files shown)


zcbenz (approved):
Nice fix, thanks!

mlx/random.cpp
```diff
@@ -280,7 +280,16 @@ array randint(
...
+  // Take the floor before casting: `astype` truncates towards zero, which
+  // makes `low` unreachable and doubles the weight of zero when the interval
+  // spans negative values.
+  auto out = astype(floor(u, s), dtype, s);
+  // Rounding in the float32 domain can place a sample on `high` itself, so
+  // clamp to `high - 1`. The bound is computed in the integer domain since it
+  // is not always representable in float32. Clamping to `low` last keeps an
+  // empty interval collapsing to `low`.
+  auto hi = astype(subtract(high, array(1, high.dtype()), s), dtype, s);
+  return maximum(minimum(out, hi, s), astype(low, dtype, s), s);
```
zcbenz, line 293: Should we pass `high - 1` to `uniform` instead of using `minimum(out, high - 1)`?

adityasingh2400: I tried that first and it loses the top value, because `uniform(low, high - 1)` samples `[low, high - 1)` and the floor then tops out at `high - 2`:

```
low=0 high=4   this PR -> [0, 1, 2, 3]      high-1 arg -> [0, 1, 2]
low=-3 high=3  this PR -> [-3 .. 2]         high-1 arg -> [-3 .. 1]
low=0 high=2   this PR -> [0, 1]            high-1 arg -> [0]
```

The last row is the clearest, the two value case collapses to a constant.

There is a second reason not to push the bound through `uniform`. The argument would go in as float32, and `high - 1` is not always representable there, so it can round back up to `high` or past it:

```
high-1 = 1073741823  -> float32 1073741824
high-1 = 2147483646  -> float32 2147483648
```

That is why the clamp is built in the integer domain instead.

One thing I should be straight about. I could not get `floor(uniform(low, high))` to actually reach `high` in testing, six seeds at 200k samples across `2**24+1`, `2**30`, `[-2**30, 2**30)` and `INT32_MAX`, all stayed in range. So the `minimum` is defensive rather than something I can show firing. Happy to drop it if you would rather not carry a guard without a reproducer, though the `floor` change is the part that fixes the reported bug either way.

zcbenz: Hmm I think we should just return `floor(uniform(low, high))`, since `uniform` is guarenteed to return between [low, high). I guess the bug was caused by `astype` rounding the result to `high`, and `floor` should have fixed it.

I wonder if we should remove `maximum(u, low)` too, I dont really see a way that we would get a number below `low`.

adityasingh2400: I have to correct myself. Earlier I said I could not get `floor(uniform(low, high))` to reach `high`. That was wrong, I was sweeping wide ranges, which is the wrong place to look. It reproduces immediately on a narrow range straddling the float32 integer limit:

```
randint(2**24, 2**24 + 2)  without the upper clamp
  int32   observed max 16777218   high = 16777218
  uint32  observed max 16777218   high = 16777218
```

`uniform` is in `[low, high)` as a real interval, but `high` here is not representable in float32, so `low + range * u` rounds up to exactly `high` and the floor keeps it. With `minimum(out, high - 1)` the same 12 seeds stay in range. Those two cases are in this PR's `test_randint`, so dropping the clamp fails its own test.

On `maximum(u, low)`, that one is load bearing too, and it is your existing tests that pin it rather than anything I added:

```
python/tests/test_random.py:233   mx.random.randint(10, -10, ...)  asserts all == 10
tests/random_tests.cpp:509        "Check high < low => all equals to low"
```

I tried the bare version locally and both of those fail, `randint(10, -10)` returns values in `[-10, 9]` instead of all `10`.

So I have left the code as it is. You are right that `floor` is what actually fixes the reported bug, the two clamps are only guarding the unrepresentable-bound case and the inverted-range contract. Happy to reshape or comment them differently if you would prefer that spelled out another way.

zcbenz: Thanks for experimenting, I wonder if the algorithm proposed by JAX would be better:
https://docs.jax.dev/en/latest/_autosummary/jax.random.randint.html
```python
def randint_via_uniform(key, shape, minval, maxval, dtype):
  u = jax.random.uniform(key, shape, minval=minval - 0.5, maxval=maxval - 0.5)
  return u.round().astype(dtype)
```

Current approach feels like spending too much efforts patching around floating number details.

adityasingh2400: I tried it, and it does not get us out of the floating point details, it lands in the same trap one step earlier.

On the exact case this PR is about, `randint(2**24, 2**24 + 2)`, the rounding form is out of range on every seed I ran:

```
jax-style   out-of-range seeds 12/12   values seen [16777216, 16777218]
this PR     out-of-range seeds  0/12   values seen [16777216, 16777217]
```

`16777218` is `high`, and `16777217` never appears at all, so it both escapes the interval and drops a valid value. The reason is that the shift is applied in float32 before sampling:

```
float32 spacing at 2**24 = 2.0
low  - 0.5 = 16777215.5 -> float32 16777216.0
high - 0.5 = 16777217.5 -> float32 16777218.0
```

At that magnitude the half cannot be represented, so both bounds move up by 0.5 and `round` has nothing left to correct.

It also changes behavior your tests pin. `randint(10, -10)` and `randint(20, -10)` both stop returning `low`:

```
python/tests/test_random.py:233   randint(10, -10) all == 10   jax-style False
tests/random_tests.cpp:509        randint(20, -10) all == 20   jax-style False
```

On everything else the two agree exactly, same values and same counts over `[-3, 3)`, so the difference really is only at the boundary.

I do take the point that this reads as patching around float behavior. That is fair, and if you want it out of `randint` entirely the honest fix is to draw integers rather than route through `uniform`, generating the range directly from `bits` and rejecting the overhang. Happy to do that instead if you would prefer it, though it is a bigger change than this bug needs.

zcbenz: Thanks for trying, this is probably the best we can get with current algorithm and I'm good with it. I will do some research to have a fully understanding of the root cause and then merge.

Issue #3885, opened 2026-07-21
Metal SDPA: no fused path for head_dim=512 (vector or full kernel), and the fallback is silent


apocryphx:
## Summary

`mx.fast.scaled_dot_product_attention` has no fused path for **head_dim = 512** — neither the vector (decode) kernel nor the full (prefill) kernel accepts it — and the fallback to the composite eager path is **silent**, with no way to query or log it. Models with d=512 attention (e.g. Gemma 4's global layers) run every such attention op, at every sequence length, on the unfused path with materialized scores.

This complements #3658 / #3293, which cover extending the **full** kernel to d=192/256. This issue is about (a) d=512 having no coverage anywhere, including the vector kernel where decode lives, and (b) fallback discoverability.

## Current routing (`use_fallback`, [scaled_dot_product_attention.cpp#L621-L633](https://github.com/ml-explore/mlx/blob/0c537a413a39616c1d19b96689705719e67be4c1/mlx/backend/metal/scaled_dot_product_attention.cpp#L621-L633))

- vector (qL ≤ 8): `query_head_dim ∈ {64, 96, 128, 256}` or `(192, 128)` — no 512
- full (qL > 8): `query_head_dim ∈ {64, 80, 128}` — no 192/256 (tracked in #3658), no 512

Gemma 4 31B (dense) has a 5:1 hybrid layout: 50 sliding layers (n_q=32, n_kv=16, d=256) and 10 global layers (n_q=32, n_kv=4, **d=512**, 25% partial RoPE). The sliding layers get the vector kernel at decode; the global layers never fuse, and at prefill **no layer of this model fuses at all**.

## Measured impact (M4 Max 128 GB, bf16, Gemma-4-31B QAT Q4; MLX pin 51b2768, routing re-verified against 0c537a41)

Decode (qL=1), isolated `mx.fast.scaled_dot_product_attention` calls, additive mask:

| shape | seqK=1024 | seqK=4096 |
|---|---:|---:|
| d=512, 32 q / 4 kv heads (**fallback**) | 168 µs | 260 µs |
| d=256, 32 q / 8 kv heads (same KV bytes, vector kernel) | 146 µs | 184 µs |

≈ +15% @1K and +41% @4K per call for the same bytes, plus ~8 extra dispatches per call. Small per token (10 layers ≈ +0.8 ms at 4K), but it is pure routing loss.

[...]

PR #4018, opened 2026-08-05
Check threadgroup size in the 1-pass sdpa_vector dispatch

apocryphx:
`sdpa_vector` dispatches a fixed 1024-thread threadgroup but never validates it against the pipeline's `maxTotalThreadsPerThreadgroup`. Both dispatches in `sdpa_vector_2pass` already call `check_kernel_threadgroup_size`; this adds the same call to the 1-pass path.

Today this is latent — I'm not aware of a head dim currently routed to the vector kernel that trips the limit on any shipping device. What makes it worth a line is *how* the failure presents when it does trip. An over-limit `dispatchThreadgroups` is dropped by the driver: the kernel never runs, no error is raised at any level, and the output buffer is returned untouched — all zeros. There is no signal from MLX at all; `MTL_DEBUG_LAYER=1` is what eventually surfaced it for me.

I hit it while prototyping a d=512 instantiation for #3885 on an M1 Max, where register pressure caps both d=512 pipelines at 832 threads (`maxTotalThreadsPerThreadgroup` = 832 for `sdpa_vector_bfloat16_t_512_512`, vs 1024 for `..._256_256` on the same device). The 2-pass path threw a clear exception from the check that's already there. The 1-pass path silently returned zeros, and I spent a while trusting benchmark numbers from a kernel that never ran.

The cost is one comparison per dispatch on an already-cached pipeline object. The benefit is that a register-limited pipeline fails the same loud way on both vector paths instead of one of them producing wrong results quietly. That protection is independent of #3885 — it applies to any future head dim, kernel change, or device whose register allocation lands under 1024.

No test: triggering this requires a pipeline that register-limits below 1024, which is device- and head-dim-specific and not reproducible in CI. The two existing `check_kernel_threadgroup_size` call sites are untested for the same reason.

### DIFF
--- a/mlx/backend/metal/scaled_dot_product_attention.cpp
+++ b/mlx/backend/metal/scaled_dot_product_attention.cpp
@@ -380,6 +380,7 @@ void sdpa_vector(
   // Get the kernel
   auto& compute_encoder = metal::get_command_encoder(s);
   auto kernel = d.get_kernel(kname, hash_name, func_consts);
+  check_kernel_threadgroup_size(kernel, group_dims, hash_name);
   compute_encoder.set_compute_pipeline_state(kernel);
 
   // Set its arguments


zcbenz (approved):
It makes sense adding the check, thanks!

Issue #3886, opened 2026-07-21
Metal BufferCache: reuse window [size, size+2*page_size) defeats caching for monotonically growing allocations (naive KV-cache append pattern)


apocryphx:
## Summary

`BufferCache::reuse_from_cache` accepts a cached buffer only if its size is in `[size, min(2*size, size + 2*page_size))` — for any buffer larger than ~32 KB the reuse window is just **[size, size + 32 KB)**. Any allocation pattern whose sizes grow monotonically (the naive per-token KV-cache append: each step requests a few KB *more* than what it just freed) therefore **misses the cache on every single request**: every step pays a real `newBuffer` + wire, and the just-freed buffers pile up in the pool until GC pressure evicts them.

This is the exact pattern a straightforward C++ KV cache built on `mx::concatenate` produces, and it silently dominates decode overhead. (mlx-lm's Python `KVCache` sidesteps it by preallocating in 256-step chunks with `slice_update` — the trap mainly catches C++ / custom-cache users, and nothing in the docs warns about it.)

## Code ([buffer_cache.h#L30-L38](https://github.com/ml-explore/mlx/blob/0c537a413a39616c1d19b96689705719e67be4c1/mlx/backend/common/buffer_cache.h#L30-L38))

```cpp
T* reuse_from_cache(size_t size) {
  auto it = buffer_pool_.lower_bound(size);
  if (it == buffer_pool_.end() ||
      it->first >= std::min(2 * size, size + 2 * page_size_)) {
    return nullptr;
  }
  ...
```

A growing sequence `size_t = base + t * per_token_bytes` requests strictly more than anything freed at `t-1`, and the freed buffer is below `lower_bound(size)` — guaranteed miss, forever.

## Measured (M4 Max 128 GB, macOS, MLX @ 51b2768; behavior identical in 0c537a41 source)

Per-"decode-step" cost of appending one position to 10 pairs of `[4, N, 512]` bf16 buffers (K+V for 10 layers), 30-step steady state, three patterns:

| pattern | ctx=512 | 1024 | 2048 | 4096 |
|---|---:|---:|---:|---:|
| A: `concatenate` **grow** (monotonic sizes → cache-miss) | 1.19 ms | 1.28 | 1.74 | 3.12 |
| B: `concatenate` at **constant** size (same copies, sizes repeat → cache-hit) | 0.58 | 0.82 | 1.13 | 1.96 |
[...]

PR #4019, opened 2026-08-05
docs: add a usage page on growing arrays

apocryphx:
Adds a short usage page documenting the preallocate-and-update pattern for arrays that grow a step at a time, and why growing with `concatenate` in a loop is worse than it looks.

The copying cost is the obvious half. The half worth documenting is that the pattern defeats buffer reuse. `BufferCache` reuses a pooled buffer only whole — it does not split one to serve a smaller request or combine several to serve a larger one — so a loop whose requests grow each step never matches what it just freed. Every step allocates from the driver, and the freed buffers accumulate at sizes nothing will ask for again.

The cost is CPU-side, so it profiles as GPU idle time between kernels rather than as a slow kernel, which is what makes it hard to find.

mlx-lm sidesteps all of this by preallocating in 256-step chunks, so Python users of it never see the problem. Anyone writing a cache directly, especially in C++, hits it with no warning.

Measured on an M4 Max (mlx 0.32.0), appending one position per step to 20 buffers of `[1, 4, N, 512]` bfloat16:

| context | concatenate | preallocate + update |
|---|---|---|
| 512 | 0.90 ms/step | 0.24 ms/step |
| 1024 | 1.11 ms/step | 0.21 ms/step |
| 4096 | 3.73 ms/step | 0.22 ms/step |

The preallocated version is flat in context length; the concatenating one is not.

The page also notes that the chunk size should be a multiple of 256, since CUDA's cuDNN attention routing requires the key/value arrays to be slices of a cache whose capacity is a multiple of `kv_cache_step` (`mlx/backend/cuda/scaled_dot_product_attention.cpp`). That requirement is currently only discoverable by reading the routing predicate.

Background and fuller measurements in #3886. This documents the existing behavior rather than changing it — I looked at fixing it in the allocator and could not find a version that doesn't pad every allocation, including weights that never recycle, so that seemed like the wrong trade to propose.

Verification: docs build clean with sphinx, no warnings attributable to the new page. The C++ example compiles against the current headers, and both Python forms were run against mlx 0.32.0 and checked to produce results identical to the naive `concatenate` loop.


docs/src/usage/growing_arrays.rst
```diff
@@ -0,0 +1,126 @@
...
+      cache = mx::concatenate({cache, pad}, 1);
+    }
+    cache = mx::slice_update(cache, x, {0, offset, 0}, {1, offset + 1, d});
+    offset++;
+    mx::eval(cache);
+  }
+
+  auto keys = mx::slice(cache, {0, 0, 0}, {1, offset, d});
+
+Why It Matters
```
zcbenz: This section is too long and verbose, it should just briefly explain MLX's buffer cache policy, and then list the numbers showing the difference.

docs/src/usage/growing_arrays.rst
```diff
@@ -0,0 +1,126 @@
...
+
+  keys = cache[:, :offset]
+
+Indexed assignment works as well and reads more naturally:
+
+.. code-block:: python
+
+  cache[:, offset : offset + 1, :] = x
+
+The same pattern in C++:
```
zcbenz: Let's just ignore C++ in the tutorial.

zcbenz (commented):
Thanks for writing this, it is very useful, especially for documenting the behavior of cuDNN kernel.

I wonder if it would make sense to slightly change it to explicitly be a guide on writing performant KV cache, which would attract much more readers.

apocryphx:
Thanks @zcbenz — reframed it along those lines.

- Retitled to "Writing a Fast KV Cache" and renamed the page to `usage/kv_cache.rst`. It now opens on autoregressive generation rather than arrays in general.
- Dropped the C++ example; the page is Python-only.
- Moved the measurements up to directly after the two code samples and cut the buffer-pool explanation down to the policy itself plus why it surfaces as GPU idle time. The numbers lead now, the mechanism supports them.
- Promoted the chunk-size section to the end, so the cuDNN requirement is the last thing a reader takes away.

Say the word if you'd rather I keep the original filename.

docs/src/usage/growing_arrays.rst
apocryphx: Removed — Python only now.

docs/src/usage/growing_arrays.rst
apocryphx: Cut down to the pool policy and the page-size caveat, and moved the table up ahead of it so the numbers come first.

zcbenz (approved):
Thanks!

PR #4021, opened 2026-08-05
Fix segfault in expand_dims for out of bounds negative axes

### LINKED ISSUE #4011: [BUG] `expand_dims()` seg faults for invalid inputs.
**Describe the bug**

`expand_dims()` gives a segmentation fault on invalid inputs. Found here: https://github.com/data-apis/array-api-compat/issues/450#issuecomment-5192379764

**To Reproduce**

```python
import mlx.core as mlx

x = mlx.zeros(())
mlx.expand_dims(x, axis=(-3, -2))
```

Gives:

```
[1]    87881 segmentation fault
```

**Expected behavior**

Should raise a `ValueError` instead.

**Desktop**
 - OS Version: MacOS 26.5.1
 - MLX Version: 0.32.0

Gusanidas:
## Proposed changes

Closes #4011

`mx.expand_dims(mx.zeros(()), axis=(-3, -2))` segfaults.

### Cause

`expand_dims(a, axes)` normalized negative axes against the output rank and
then passed them to `expand_dims_impl`, which normalized them a second time
(`-3 + 2 = -1`, still negative and never range checked, then `-1 + 2 = 1`). An
out of bounds axis therefore turned silently into a valid but wrong one, and
the axes ended up as `{1, 0}` instead of sorted. `ExpandDims::output_shape`
assumes sorted, in bounds axes, so this inserted past the end of an empty shape
and crashed. `out_ndim` was also `size_t`, so the `new_ax < 0` guard could
never fire.

### Fix

Normalize, validate and sort the axes exactly once, in a single helper per op,
reusing the existing `normalize_axis_index`. Error wording now comes from that
helper; no tests asserted on the old strings.

`squeeze` had the same double normalization, but failed silently rather than
crashing: `mx.squeeze(mx.zeros((1, 1, 1)), (-5,))` returned shape `(1, 1)`
instead of raising, because `-5` mapped to axis `1`. Fixed the same way, though
happy to split it out if you would prefer this PR stay scoped to the crash.

### Testing

Extended the C++ `test squeeze and expand` case and added
`test_squeeze_expand_invalid_axes` in Python, covering out of bounds negative
axes for both ops plus the valid negative axis cases that must keep working.
Both segfault against unpatched `ops.cpp` and pass with the fix.

Full C++ and Python suites pass on a Metal build; the only failures are torch
MPS DLPack tests, which fail identically without this change.

## Checklist

### DIFF
--- a/mlx/ops.cpp
+++ b/mlx/ops.cpp
@@ -562,26 +562,50 @@ array hadamard_transform(
       {astype(a, dtype, s)});
 }
 
+namespace {
+
+std::vector<int> normalize_squeeze_axes(
+    const array& a,
+    const std::vector<int>& axes) {
+  int ndim = a.ndim();
+  std::set<int> unique_axes;
+  for (auto ax : axes) {
+    int new_ax = normalize_axis_index(ax, ndim, "[squeeze] ");
+    if (a.shape(new_ax) != 1) {
+      std::ostringstream msg;
+      msg << "[squeeze] Cannot squeeze axis " << ax << " with size "
+          << a.shape(new_ax) << " which is not equal to 1.";
+      throw std::invalid_argument(msg.str());
+    }
+    unique_axes.insert(new_ax);
+  }
+  if (unique_axes.size() != axes.size()) {
+    throw std::invalid_argument("[squeeze] Received duplicate axes.");
+  }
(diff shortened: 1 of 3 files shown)


zcbenz (approved):
Nice fix, thanks!

PR #4025, opened 2026-08-06
Fix C++ benchmarks failing to build on overloaded astype

ayaangazali:
## Proposed changes

The C++ benchmarks do not compile:

```
$ cmake -B build -DMLX_BUILD_BENCHMARKS=ON && cmake --build build
benchmarks/cpp/single_ops.cpp:31:3: error: no matching function for call to 'time_fn'
   31 |   TIMEM("mx::float32 to mx::int32", mx::astype, a, mx::int32, device);
benchmarks/cpp/time_utils.h:26:8: note: candidate template ignored: couldn't infer template argument 'F'
   26 | double time_fn(F fn, Args&&... args) {
```

`mx::astype` has two overloads (`ops.h:51` and `ops.h:52`), so the bare name cannot be deduced as the template argument of `time_fn`. There are 11 of these across `single_ops.cpp` and `irregular_strides.cpp`, and they are the only errors in the target.

Wrapped the three argument form in a small lambda per file, which is the same style the rest of these files already use for ops that need adapting, and left the call sites otherwise unchanged.

With that, `MLX_BUILD_BENCHMARKS=ON` builds all four binaries and they run:

```
Timing (mx::float32 to mx::int32) astype ... 0.079649 msec
Timing (1D strided) astype ... 0.14269 msec
```

Since the benchmarks default to `OFF`, this looks like it went unnoticed for a while rather than being a recent regression.

## Checklist

(benchmark sources, so no test; verified by building with benchmarks enabled and running both binaries)

---

honest note: freshman contributor and Claude Code helps me hunt, but i found this by turning the benchmarks on and building them, read the actual overload resolution error, checked `ops.h` to confirm astype really is overloaded, and then reran both binaries to make sure they produce numbers rather than just compiling. built CPU only here, so i have not exercised the Metal paths of these benchmarks.

### DIFF
--- a/benchmarks/cpp/irregular_strides.cpp
+++ b/benchmarks/cpp/irregular_strides.cpp
@@ -9,6 +9,12 @@
 
 namespace mx = mlx::core;
 
+// mx::astype is overloaded, so it cannot be passed directly to the timing
+// helpers. Wrap the three argument form instead.
+auto astype = [](const mx::array& a, mx::Dtype dtype, mx::StreamOrDevice s) {
+  return mx::astype(a, dtype, s);
+};
+
 void time_irregular_binary_ops_1D() {
   auto device = mx::default_device();
   int size = 1000000;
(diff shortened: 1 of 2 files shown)


zcbenz (approved):
Nice catch!

PR #4030, opened 2026-08-06
Fix mx.longsumexp output shape issue.

### LINKED ISSUE #4029: [BUG] `mx.logsumexp` returns the wrong output shape for inputs with a size-1 leading dim
**Describe the bug**

When `logsumexp` is called **without an explicit axis** (full reduction) on an
array whose leading dimension has size 1 (e.g. shape `(1, N)`, `(1, 1)`, …),
the result has shape `(1,)` instead of the scalar shape `()`. Every other MLX
reduction (`sum`, `prod`, `max`, …) returns a scalar `()` for a full reduction,
so `logsumexp` is inconsistent, and the result fails the universal reduction
invariant: *a reduction over all axes with `keepdims=False` is 0-dimensional*.

**To Reproduce**
The following code is the PoC 

```python
import mlx.core as mx
import numpy as np

for shp in [(1, 2), (1, 1), (3, 4), (2, 1)]:
    r = mx.logsumexp(mx.zeros(shp))
    s = mx.sum(mx.zeros(shp))
    print(f"logsumexp(zeros{shp}) -> {np.array(r).shape}   "
          f"sum(zeros{shp}) -> {np.array(s).shape}")
```

**Expected behavior**

```python
>>> np.array(mx.logsumexp(mx.array([[0.0, 0.0]]))).shape
(1,)        # actual — wrong
>>> np.array(mx.sum(mx.array([[0.0, 0.0]]))).shape
()          # expected (consistent with every other reduction)
```

For your reference 
```
op            input (1,2)   (1,1)   (1,5)   (2,2)   (2,1)   (3,4)
sum              ()          ()      ()      ()      ()      ()
prod             ()          ()      ()      ()      ()      ()
mean             ()          ()      ()      ()      ()      ()
max              ()          ()      ()      ()      ()      ()
min              ()          ()      ()      ()      ()      ()
std              ()          ()      ()      ()      ()      ()
var              ()          ()      ()      ()      ()      ()
all              ()          ()      ()      ()      ()      ()
any              ()          ()      ()      ()      ()      ()
count_nonzero    ()          ()      ()      ()      ()      ()
logsumexp        (1,)        (1,)    (1,)    ()      ()      ()     
```

 - OS Version: 15.5
 - Version:  MLX `0.32.1`,

JasonHonKL:
## Proposed changes
Fix #4029

## Checklist

### DIFF
--- a/mlx/ops.cpp
+++ b/mlx/ops.cpp
@@ -2864,7 +2864,7 @@ array logsumexp(
         std::make_shared<LogSumExp>(to_stream(s)),
         {astype(a, dtype, s)});
     if (!keepdims) {
-      out = squeeze(out, -1, s);
+      out = squeeze(out, axes, s);
     }
     return out;
   }
--- a/python/tests/test_ops.py
+++ b/python/tests/test_ops.py
@@ -840,6 +840,47 @@ class TestOps(mlx_tests.MLXTestCase):
         x = mx.broadcast_to(mx.random.uniform(shape=(2, 1, 8)), (2, 2, 8))
         self.assertTrue(mx.allclose(mx.logsumexp(x), logsumexp(x)))
 
+    def test_logsumexp_shape(self):
+        # A reduction over all axes with keepdims=False must yield a scalar
+        # array (), consistent with every other reduction (sum, prod, max, ...).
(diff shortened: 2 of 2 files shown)


zcbenz (approved):
Thanks!

PR #4033, opened 2026-08-06
Fix empty matrix case in cholesky

### LINKED ISSUE #4032: [BUG] `linalg.cholesky` gives runtime error on empty matrices.
**Describe the bug**

Calling `linalg.cholesky` on an empty (0, 0) matrix aborts the process. Found here: https://github.com/data-apis/array-api-compat/issues/450#issuecomment-5202873179

**To Reproduce**

```python
import mlx.core as mx
a = mx.zeros((0, 0))
r = mx.linalg.cholesky(a)
mx.eval(r)   # process aborts here
```
Output:

```
** On entry to  POTRF, parameter number  4 had an illegal value
libc++abi: terminating due to uncaught exception of type std::runtime_error: [Cholesky::eval_cpu] Cholesky decomposition failed with error code -4
[1]    31154 abort  
```

**Expected behavior**

Return an empty (0, 0) array

- OS Version: MacOS 26.5.1
- MLX Version: 0.32.0

prady0t:
## Proposed changes

Fixes #4032 

## Checklist

### DIFF
--- a/mlx/backend/cpu/cholesky.cpp
+++ b/mlx/backend/cpu/cholesky.cpp
@@ -26,15 +26,15 @@ void cholesky_impl(const array& a, array& factor, bool upper, Stream stream) {
       a.flags().row_contiguous ? CopyType::Vector : CopyType::General,
       stream);
 
+  if (a.shape(-1) == 0) {
+    return;
+  }
   auto& encoder = cpu::get_command_encoder(stream);
   encoder.set_output_array(factor);
   encoder.dispatch([matrix = factor.data<T>(),
                     upper,
                     N = a.shape(-1),
                     size = a.size()]() mutable {
-    if (N == 0) {
-      return;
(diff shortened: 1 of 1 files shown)


mlx/backend/cpu/cholesky.cpp
```diff
@@ -32,6 +32,9 @@ void cholesky_impl(const array& a, array& factor, bool upper, Stream stream) {
                     upper,
                     N = a.shape(-1),
                     size = a.size()]() mutable {
+    if (N == 0) {
```
zcbenz: I think this can be called earlier, i.e. before `get_command_encoder`.

prady0t: Or maybe even before `copy_cpu`?

zcbenz (approved):
Thanks!

PR #4036, opened 2026-08-06
Fix copy bug in mlx.core.asarray

aaishwarymishra:

This pull request refines how the `mx.asarray` and `mx.array` functions handle the `copy=False` option when given an existing `mx::array`. Now, if the input array already matches the requested dtype, a copy is avoided as expected. If a dtype conversion would be needed, an error is raised if `copy=False` is specified. The test suite is updated to verify this improved behavior.

### Improvements to array creation logic

* Updated `create_array` in `python/src/convert.cpp` to only raise an error for `copy=False` when a dtype conversion is required, otherwise returning the original array without copying.
* Removed redundant code handling for `mx::array` instances in the type dispatch, consolidating logic for clarity and correctness.

### Test suite updates

* Changed `test_asarray_copy` and `test_asarray` in `python/tests/test_array.py` to confirm that no error is raised when `copy=False` is used with an array of the same dtype, and that errors are still raised if a dtype conversion is required without copying. [[1]](diffhunk://#diff-cfd104f77ff7a0bdc64e75551b6e228eefd980865b0083bde29203951f1dc6adL2613-R2621) [[2]](diffhunk://#diff-cfd104f77ff7a0bdc64e75551b6e228eefd980865b0083bde29203951f1dc6adL2641-L2642)## Proposed changes

## Checklist

### DIFF
--- a/python/src/convert.cpp
+++ b/python/src/convert.cpp
@@ -707,37 +707,37 @@ mx::array create_array(
     return mx::astype(arr, dtype, copy);
   }
 
-  if (copy.has_value() && copy.value() == false)
+  if (copy.has_value() && copy.value() == false) {
     throw std::invalid_argument(
         "Unable to avoid copy while creating an array as requested.");
-}
-
-if (nb::isinstance<nb::bool_>(v)) {
-  return mx::array(nb::cast<bool>(v), t.value_or(mx::bool_));
-} else if (nb::isinstance<nb::int_>(v)) {
-  auto val = nb::cast<int64_t>(v);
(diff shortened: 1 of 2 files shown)


zcbenz (commented):
The reason we throw for `copy=False` is that, semantically it means the copy would be shallow:

```python
b = array(1)
a = asarray(b, copy=False)
b += 1
assert(a == 2)
```

which is not true in MLX because we don't have in-place updates, and accepting `copy=False` would just confuse users.

PR #4028, opened 2026-08-06
Add optional dtype parameter to zeros_like and ones_like

reckylurker:
## Proposed changes
Solves #4001 

## Checklist

- [x] I have read the [CONTRIBUTING](https://github.com/ml-explore/mlx/blob/main/CONTRIBUTING.md) document
- [x] I have run `pre-commit run --all-files` to format my code / installed pre-commit prior to committing changes
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have updated the necessary documentation (if needed)

PR #4037, opened 2026-08-06
Fix: Add dtype to empty_like function

### LINKED ISSUE #4001: Adding dtype parameter to creation methods like the zeroes_like and ones_like
Hi, I have checked that the current implementation of like methods including `zeros_like`, `ones_like` and `empty_like` does not take `dtype` as parameter unlike non like methods `zeros` for example.

```
array zeros_like(const array &a, StreamOrDevice s = {})
```

I think adding dtype will make them array api compatible, I seemed like an simple addition, but I ran into issue that many internal methods that use them, use positional arguments which breaks the code and need much refactoring.

Similarly with the users there code can be broken too after this addition, a simple work around could be to add it after stream, which can avoid all that.

I just wanted to get your opinion on this before I made any changes.

JasonHonKL:
## Proposed changes
(Please merge previous PR) @reckylurker 
Fix #4001 

## Update
Added test with the empty_like function. Remove public APIs of empty like. 

## Checklist

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -262,7 +262,8 @@ if(WIN32)
   FetchContent_MakeAvailable(dlfcn-win32)
   endblock()
   target_include_directories(mlx PRIVATE "${dlfcn-win32_SOURCE_DIR}/src")
-  target_link_libraries(mlx PRIVATE dl)
+  target_sources(mlx PRIVATE "${dlfcn-win32_SOURCE_DIR}/src/dlfcn.c")
+  target_link_libraries(mlx PRIVATE Psapi.lib)
 endif()
 
 if(MLX_BUILD_CPU)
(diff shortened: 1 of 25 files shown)


zcbenz (changes requested):
Can you make the C++ `zeros_like` take additional dtype while keeping `empty_like` unchanged as an alias? Also we prefer hiding `empty`/`empty_like` from public APIs as we don't want to have multiple APIs doing the same thing.

JasonHonKL:
@zcbenz Sure no problem.

JasonHonKL:
@zcbenz Removed the public APIs. Since at the beginning zero_likes is the same as empty_like, and zero_likes is updated, this PR just adds tests to verify empty_like. Thanks for the reminder !

zcbenz:
Ah sorry I forgot https://github.com/ml-explore/mlx/pull/4028 just updated the zeros_like. Since empty_like is a trivial alias of zeros_like we don't really need to add more tests, so I'm closing this PR.

PR #4046, opened 2026-08-07
Reject negative dimensions in broadcast_to and random shapes

ayaangazali:
## Proposed changes

`mx.broadcast_to` and every `mx.random.*` sampler accept negative dimensions and hand back an array whose shape is negative. The size is then computed as an unsigned product, so it underflows:

```python
>>> mx.broadcast_to(mx.array(1.0), (2, -3)).size
18446744073709551610
>>> mx.random.uniform(shape=(2, -3)).shape
(2, -3)
```

Reachable without writing a negative literal, since scale factors flow into shapes:

```python
>>> nn.Upsample(scale_factor=-2)(mx.zeros((1, 4, 4, 1))).shape
(1, -8, -8, 1)
```

Everything else in the library already rejects this. `mx.zeros`, `mx.ones` and `mx.full` raise `[full] Negative dimensions not allowed.`, `mx.reshape` raises, and numpy raises `all elements of broadcast shape must be non-negative` for the same `broadcast_to` call. These two paths are the ones that missed the check.

Added the same guard `full` already uses, at the two places a user supplied shape enters:

- `broadcast_to` in `mlx/ops.cpp`
- `random::bits` in `mlx/random.cpp`, which every sampler (`uniform`, `normal`, `randint`, `gumbel`, `laplace`, `truncated_normal`) routes through

```python
>>> mx.broadcast_to(mx.array(1.0), (2, -3))
ValueError: [broadcast_to] Negative dimensions not allowed.
>>> mx.random.uniform(shape=(2, -3))
ValueError: [bits] Negative dimensions not allowed.
```

Zero sized dimensions are untouched: `mx.broadcast_to(mx.ones((3,)), (0, 3))` and `mx.random.uniform(shape=(0,))` still work.

## Checklist

Tested on a CPU only build (`MLX_BUILD_METAL=OFF`). `test_ops.py`, `test_random.py`, `test_array.py`, `test_nn.py` and `test_autograd.py` all pass; the two new assertions fail on main.

---

freshman contributor, i use Claude Code alongside my own digging. found this by feeding deliberately bad shapes to the array constructors and comparing which ones complained, then traced the two that did not back to the missing check.

### DIFF
--- a/mlx/ops.cpp
+++ b/mlx/ops.cpp
@@ -304,6 +304,10 @@ array as_strided(
     Strides strides,
     size_t offset,
     StreamOrDevice s /* = {} */) {
+  if (std::any_of(shape.begin(), shape.end(), [](auto i) { return i < 0; })) {
+    throw std::invalid_argument(
+        "[as_strided] Negative dimensions not allowed.");
+  }
   auto copied_shape = shape; // |shape| will be moved
   auto dtype = a.dtype(); // |a| will be moved
   return array(
--- a/mlx/ops.cpp
+++ b/mlx/ops.cpp
@@ -1692,6 +1692,11 @@ array broadcast_to(
     const array& a,
     const Shape& shape,
     StreamOrDevice s /* = {} */) {
+  if (std::any_of(shape.begin(), shape.end(), [](auto i) { return i < 0; })) {
+    throw std::invalid_argument(
+        "[broadcast_to] Negative dimensions not allowed.");
+  }
+
   if (a.shape() == shape) {
     return a;
   }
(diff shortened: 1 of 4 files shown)


ayaangazali:
Rebased on main and added `as_strided`, which had the same hole:

```python
>>> mx.as_strided(mx.arange(6.0), (-2, 3), (3, 1), 0).size
18446744073709551610
```

Only the shape is checked, negative strides are still allowed since `as_strided` uses them for reversal (there is an existing test for that at test_ops.py:2226).

zcbenz (approved):

PR #4048, opened 2026-08-07
Remove the per-op completion handler in gpu::eval

kernelpool:
## Proposed changes

`gpu::eval` registers an empty `addCompletedHandler` and builds an `unordered_set` of `shared_ptr`s for every operation, in order to keep that op's inputs alive until the command buffer completes. `CommandEncoder::commit()` already installs one handler per command buffer, so the per-op handlers are redundant and the buffers can go there instead. On an M3 Ultra that takes `mlx-community/Qwen3.5-0.8B-4bit` decode from 280 to 367 tok/s. Bandwidth-bound models are unchanged.

### Before

```
mlx_lm.benchmark --model mlx-community/Qwen3.5-0.8B-4bit -p 2048 -g 128 -n 3
Running warmup..
Timing with prompt_tokens=2048, generation_tokens=128, batch_size=1.
Trial 1:  prompt_tps=13108.539, generation_tps=284.374, peak_memory=1.886, total_time=0.744
Trial 2:  prompt_tps=13404.757, generation_tps=278.961, peak_memory=1.886, total_time=0.749
Trial 3:  prompt_tps=13181.042, generation_tps=277.251, peak_memory=1.887, total_time=0.753
Averages: prompt_tps=13231.446, generation_tps=280.195, peak_memory=1.886
```

### After

```
mlx_lm.benchmark --model mlx-community/Qwen3.5-0.8B-4bit -p 2048 -g 128 -n 3
Running warmup..
Timing with prompt_tokens=2048, generation_tokens=128, batch_size=1.
Trial 1:  prompt_tps=13288.070, generation_tps=365.077, peak_memory=1.886, total_time=0.637
Trial 2:  prompt_tps=13271.215, generation_tps=361.761, peak_memory=1.886, total_time=0.650
Trial 3:  prompt_tps=13402.103, generation_tps=376.355, peak_memory=1.887, total_time=0.629
Averages: prompt_tps=13320.463, generation_tps=367.731, peak_memory=1.886
```

## Checklist

### DIFF
--- a/mlx/backend/metal/device.h
+++ b/mlx/backend/metal/device.h
@@ -54,6 +54,14 @@ class MLX_API CommandEncoder {
   void add_temporary(array arr);
   void add_temporaries(std::vector<array> arrays);
 
+  // Keep a buffer alive until the current command buffer completes. Deduped
+  // because a primitive may bind the same array in several slots.
+  void hold_buffer(const std::shared_ptr<array::Data>& buf) {
+    if (buf && retained_ptrs_.insert(buf.get()).second) {
+      retained_buffers_.push_back(buf);
+    }
+  }
+
   void dispatch_threadgroups(MTL::Size grid_dims, MTL::Size group_dims);
   void dispatch_threads(MTL::Size grid_dims, MTL::Size group_dims);
   void maybeInsertBarrier();
@@ -129,6 +137,8 @@ class MLX_API CommandEncoder {
   bool needs_barrier_{false};
   bool concurrent_{false};
   std::vector<array> temporaries_;
+  std::vector<std::shared_ptr<array::Data>> retained_buffers_;
+  std::unordered_set<const array::Data*> retained_ptrs_;
   std::unordered_set<MTL::Resource*> prev_inputs_;
   std::unordered_set<MTL::Resource*> prev_outputs_;
   std::unordered_set<MTL::Resource*> next_inputs_;
--- a/mlx/backend/metal/eval.cpp
+++ b/mlx/backend/metal/eval.cpp
@@ -1,6 +1,4 @@
 // Copyright © 2023-2024 Apple Inc.
-#include <memory>
-
 #include "mlx/backend/gpu/eval.h"
 #include "mlx/backend/metal/device.h"
 #include "mlx/backend/metal/utils.h"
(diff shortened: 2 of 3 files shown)


mlx/backend/metal/device.h
```diff
@@ -129,6 +137,8 @@ class MLX_API CommandEncoder {
   bool needs_barrier_{false};
   bool concurrent_{false};
   std::vector<array> temporaries_;
+  std::vector<std::shared_ptr<array::Data>> retained_buffers_;
+  std::unordered_set<const array::Data*> retained_ptrs_;
```
zcbenz, line 141: I think you can just store buffers in the set.

```suggestion
  std::unordered_set<std::shared_ptr<array::Data>> retained_buffers_;
```

mlx/backend/metal/eval.cpp
```diff
@@ -44,27 +42,23 @@ void eval(array& arr) {
...
 
   if (encoder.needs_commit()) {
     encoder.end_encoding();
     scheduler::notify_new_task(s);
-    encoder.commit([s, buffers = std::move(buffers)]() {
-      scheduler::notify_task_completion(s);
-    });
-  } else {
-    command_buffer->addCompletedHandler(
-        [buffers = std::move(buffers)](MTL::CommandBuffer* cbuf) {});
```
zcbenz, line 67: The downside of this approach is that buffers are now retained until the command encoder is committed, which I think could increase peak memory a lot? Can you check the memory usage in a few popular models?

I support reducing the completion handler to have lower overhead but we probably need to add some heuristics.

zcbenz:
Closing due to lack of response.

avlp12:
I came across this PR while looking into launch overhead on my own stack, and
had the patch built anyway, so here is the missing data, in case it is useful.

Setup: Apple M3 Ultra (512 GB), macOS 26, mlx v0.32.2, GLM-5.3-Flash
(320B-A18B MoE, 4-bit, ~172 GB of weights), batch-1 greedy decode.

Throughput: no measurable change — 34.78 vs 34.76 tok/s (medians of three
alternating stock/patched pairs, −0.06%). The +31% on Qwen3.5-0.8B looks real
but size-dependent: at ~29 ms/step the per-op handler is already in the noise
here. So this reads as overhead reduction for small, fast-stepping models
rather than a universal speedup.

Peak memory: unchanged. 199.2330 / 199.2329 GB stock vs 199.2326 GB patched —
the patched arm was never higher, and the two agree to within 0.0005 GB out of
199 GB. Across a 300 s / 6067-step decode soak, peak settled to the same value
in both arms within ~20 s and did not move again, and RSS did not drift. I run
with `MLX_MAX_OPS_PER_BUFFER=100000` and `MLX_MAX_MB_PER_BUFFER=2048`, well
above the defaults, which should be the adverse direction for buffer retention.

Caveats: the diff does not apply cleanly to 0.32.2, so I hand-adapted it — a
rebase on main is the artifact worth trusting, and my numbers are only
indicative. The soak restarts the cache periodically and uses a bounded context,
so it does not probe long-single-context growth. My test suite passed and
greedy output was byte-identical between arms.

Hope this is useful.

PR #4054, opened 2026-08-07
Fix hadamard_transform on the GPU for n = m with no power-of-2 factor

### LINKED ISSUE #4049: [BUG] mx.hadamard_transform` crashes on GPU for the documented sizes n ∈ {12, 20, 28}
**Describe the bug**

`mx.hadamard_transform` is documented to support sizes `n = m·2^k` for
`m ∈ {1, 12, 20, 28}` and `2^k ≤ 8192`. On the **Metal (GPU) backend** the three
sizes where the power-of-2 factor is exactly 1 — i.e. `n ∈ {12, 20, 28}` — raise
at runtime:

```
runtime_error: [metal::Device] Unable to build metal library from source
```

**To Reproduce**

```python
import mlx.core as mx
import numpy as np

x = mx.array(np.arange(12, dtype=np.float32))

with mx.stream(mx.cpu):
    mx.eval(mx.hadamard_transform(x))   # OK

with mx.stream(mx.gpu):
    mx.eval(mx.hadamard_transform(x))   # raises runtime_error
```

**Expected behavior**
Just no crush.

FYI
```
   m   k     n  GPU
  12   1    12  CRASH      <- n = m · 2^0
  12   2    24  OK
  12   4    96  OK
  12   8  1536  OK
  20   1    20  CRASH
  20   2    40  OK
  20   4   160  OK
  28   1    28  CRASH
  28   2    56  OK
  28   4   224  OK
```

axiom-of-choice:
Fixes #4049.

## Summary

`hadamard_transform` is documented for `n = m*2^k` with `m in (1, 12, 20, 28)`, but on Metal the three sizes where the power-of-2 factor is exactly 1 (`n = 12, 20, 28`) failed to build the Metal library:

```
hadamard.h:44:19: error: constexpr variable 'num_steps' must be initialized by a constant expression
note: division by zero
  constexpr short num_steps = logN / logR;
```

`decompose_hadamard` returns `(n=1, m)` for those sizes, so the n2 stage is instantiated with `N=1`, giving `logR = ctz(max_radix_2) = ctz(1) = 0` and a `0/0`. The n1 and m stages are already guarded by `> 1`; n2 was not.

## Changes

**Emit and launch the n2 stage only when `n2 > 1`.** When `n2 == 1` the stage is the identity and carries no scale: `scale_n2` is `scale` only when `m == 1`, and `n == 1 && m == 1` returns before the primitive is created (`ops.cpp:530`), so `n2 == 1` inside the backend always implies `m > 1` and `scale_n2 == 1.0`. The skipped stage was also what copied `x` into `y`, so the m stage now reads `x` directly when `n == 1`; that matters on the non-donatable path, where `y` is freshly allocated and otherwise never written.

**Cap `read_width_m` at `n`.** The m kernel strides by `n / read_width`, so with `n = 1` and `read_width = 4` the group width was `1/4 == 0` and **zero threads were dispatched**, leaving the output untouched. Fixing only the build error surfaces this as a silent all-zero result for any batched input, so both changes are needed.

## Verification

Built with Metal on an **M2 Pro (Mac mini, macOS 26.5.2)** and compared against the CPU backend. Bit-exact, `max |gpu - cpu| = 0`, for `m in (12, 20, 28)` across shapes `(m,)`, `(4, m)`, `(3, 5, m)`, with and without an explicit `scale`, on both the donatable and freshly-allocated output paths. Sizes with `k >= 1` (24, 40, 48, 56, 80, 112, ...) are unchanged and still bit-exact.

`pytest python/tests/test_ops.py python/tests/test_autograd.py python/tests/test_vmap.py`: 227 passed. The one failure, `test_unstack`, is unrelated and pre-existing here (`np.unstack` needs numpy >= 2.1, this env has 1.26.4); it fails identically without this patch. `clang-format` 21.1.8 and `black` clean.

## Why it was not caught

`test_hadamard` builds its cases from `range(1, 14)`, so `k` starts at 1 and `n = m*2^0` was never tested; `m = 12` and `m = 20` do not appear in its `[1, 28]` list at all. The new `test_hadamard_m_only` pins all three sizes, and includes a batch dimension deliberately: the 1-D case alone still passes with the zero-thread dispatch when the input happens to be donated, so a 1-D-only test would not have caught the second half of this.

## Risks

Low, and confined to `n == 1`. For every other size the emitted kernels, launch order, grid dimensions and `read_width_m` are byte-identical to before, so `k >= 1` paths cannot change behavior. The one judgement call is reading from `x` in the m stage when `n == 1`; it is guarded on `n > 1 ? y : x` rather than changing the general path, so the multi-stage cases keep chaining through `y` exactly as they did.

## Pre-merge / post-merge

None. No API or build changes.

### DIFF
--- a/mlx/backend/metal/hadamard.cpp
+++ b/mlx/backend/metal/hadamard.cpp
@@ -69,9 +69,7 @@ void hadamard_mn_contiguous(
   int n = n1 * n2;
   int read_width_n1 = n1 == 2 ? 2 : 4;
   int read_width_n2 = n2 == 2 ? 2 : 4;
-  // n == 1 (m in (12, 20, 28) with no power-of-2 factor) leaves the m stage as
-  // the only transform, and it strides by n / read_width_m, so the read width
-  // cannot exceed n.
+  // The m stage strides by n / read_width_m, so cap the read width at n.
   int read_width_m = (n == 1) ? 1 : ((n == 2 || m == 28) ? 2 : 4);
   int max_radix_1 = std::min(n1, 16);
   int max_radix_2 = std::min(n2, 16);
@@ -158,8 +156,7 @@ void hadamard_mn_contiguous(
   if (m > 1) {
     auto kernel = d.get_kernel("m" + kname, lib);
     compute_encoder.set_compute_pipeline_state(kernel);
-    // n = 1 means neither power-of-2 stage ran, so x has not been copied into y
-    // yet and the only transform is this one; read straight from x.
+    // With n == 1 no earlier stage copied x into y, so read from x.
     compute_encoder.set_input_array(n > 1 ? y : x, 0);
     compute_encoder.set_output_array(y, 1);
     compute_encoder.set_bytes(scale_m, 2);
--- a/mlx/backend/metal/hadamard.cpp
+++ b/mlx/backend/metal/hadamard.cpp
@@ -69,7 +69,10 @@ void hadamard_mn_contiguous(
   int n = n1 * n2;
   int read_width_n1 = n1 == 2 ? 2 : 4;
   int read_width_n2 = n2 == 2 ? 2 : 4;
-  int read_width_m = (n == 2 || m == 28) ? 2 : 4;
+  // n == 1 (m in (12, 20, 28) with no power-of-2 factor) leaves the m stage as
+  // the only transform, and it strides by n / read_width_m, so the read width
+  // cannot exceed n.
+  int read_width_m = (n == 1) ? 1 : ((n == 2 || m == 28) ? 2 : 4);
   int max_radix_1 = std::min(n1, 16);
   int max_radix_2 = std::min(n2, 16);
   float scale_n1 = 1.0;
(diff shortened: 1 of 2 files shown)


mlx/backend/metal/hadamard.cpp
```diff
@@ -143,18 +145,22 @@ void hadamard_mn_contiguous(
...
+    compute_encoder.set_bytes(scale_n2, 2);
+    compute_encoder.dispatch_threads(grid_dims_n2, group_dims_n2);
+  }
 
   // Launch the strided transform for m
   if (m > 1) {
     auto kernel = d.get_kernel("m" + kname, lib);
     compute_encoder.set_compute_pipeline_state(kernel);
-    compute_encoder.set_input_array(y, 0);
+    // n = 1 means neither power-of-2 stage ran, so x has not been copied into y
```
JasonHonKL: Just personally thinks these comment are a bit nasty haha. (Maybe the author likes it idk)

mlx/backend/metal/hadamard.cpp
```diff
@@ -69,7 +69,10 @@ void hadamard_mn_contiguous(
   int n = n1 * n2;
   int read_width_n1 = n1 == 2 ? 2 : 4;
   int read_width_n2 = n2 == 2 ? 2 : 4;
-  int read_width_m = (n == 2 || m == 28) ? 2 : 4;
+  // n == 1 (m in (12, 20, 28) with no power-of-2 factor) leaves the m stage as
```
JasonHonKL: yea just a bit too much comment to me hahaha.

mlx/backend/metal/hadamard.cpp
axiom-of-choice: 🤣
Can short them for sure

JasonHonKL: Yea! Great !

axiom-of-choice: Done!

mlx/backend/metal/hadamard.cpp
axiom-of-choice: Done

zcbenz (approved):
Thanks!

PR #4056, opened 2026-08-07
Normalize biases before encoding in gather_qmm_rhs

### LINKED ISSUE #4055: [BUG] gather_qmm with sorted_indices=True silently returns wrong results for a row-sliced quantized weight
### Describe the bug

A row slice of a quantized expert weight is non-contiguous. Passing it to
`mx.gather_qmm` with `sorted_indices=True` returns wrong values — no exception,
no warning. The same slice is fine with `sorted_indices=False`, and a
contiguous copy of it is fine either way.

Easy to hit in practice: a MoE FFN that stores `gate` and `up` as one
`[E, 2H, D]` tensor and slices the halves does exactly this.

### To Reproduce

```python
import mlx.core as mx

E, R, D, N = 8, 64, 256, 64
mx.random.seed(0)
qw, qs, qb = mx.quantize((mx.random.normal((E, 2 * R, D)) * 0.05).astype(mx.float16),
                         group_size=64, bits=4)
x = (mx.random.normal((N, 1, D)) * 0.5).astype(mx.float16)
idx = mx.sort(mx.random.randint(0, E, (N,)).astype(mx.uint32))
mx.eval(qw, qs, qb, x, idx)

view = (qw[:, :R], qs[:, :R], qb[:, :R])        # non-contiguous row slice
copy = tuple(mx.contiguous(a) for a in view)    # identical values

kw = dict(rhs_indices=idx, transpose=True, group_size=64, bits=4)
print(mx.abs(mx.gather_qmm(x, *view, sorted_indices=True, **kw)
           - mx.gather_qmm(x, *copy, sorted_indices=True, **kw)).max())   # 1.41
print(mx.abs(mx.gather_qmm(x, *view, sorted_indices=False, **kw)
           - mx.gather_qmm(x, *copy, sorted_indices=False, **kw)).max())  # 0.0
```

Against an fp32 dequantized reference the contiguous result is correct
(4.7e-04, i.e. quantization noise) and the sliced one is not (1.41e+00).

### Expected behavior

Same result as the contiguous copy, or an explicit error.

### Where it comes from

`ensure_row_contiguous_matrix` (`mlx/backend/metal/quantized.cpp:64-83`), used
by `GatherQMM::eval_gpu` for `w`/`scales`/`biases` (`quantized.cpp:1566-1570`),
validates only the last two axes:

```cpp
auto stride_0 = x.strides()[x.ndim() - 2];
auto stride_1 = x.strides()[x.ndim() - 1];
if (stride_0 == x.shape(-1) && stride_1 == 1) {
  return x;                       // accepted, no copy
}
```

A `[E, 2R, Dw]` weight sliced to `[E, R, Dw]` satisfies both conditions, so no
copy is made — but its leading stride is still `2R * Dw`, not `R * Dw`. The
sorted path (`gather_qmm_rhs`, `quantized.cpp:1587`) walks experts itself and
takes the expert offset from the logical shape, so every expert after the first
reads from the wrong place. The helper's name is the tell: it checks that the
*matrix* is row-contiguous, which is enough for the 2D quantized ops but not
for the gather ops, which also index the leading axis.

### Environment

Apple M5 Pro 24 GB, macOS 26.5.1, `mlx 0.32.0` (pip).

<details>
<summary>Evidence that the leading (expert) stride is the cause</summary>

If it is, the failure must disappear when there is only one expert to cross:

| weight | view vs contiguous copy, `sorted=True` |
|---|---|
| `E=1`, first half `w[:, :R]` | 0.0 |
| `E=1`, second half `w[:, R:]` | 0.0 |
| `E=8`, first half | 1.41e+00 |
| `E=8`, second half | 1.98e+00 |
| `E=8`, natively contiguous `[E, R, D]` (control) | 1.2e-03 (quantization noise) |

Only the sorted path is affected, and only once the row count is large enough
to select it:

| shape | rows | `sorted=False` | `sorted=True` |
|---|---|---|---|
| E=8, R=64, D=256 | 64 | ok | **wrong** |
| E=128, R=768, D=2048 | 8 / 32 / 64 / 128 | ok | ok |
| E=128, R=768, D=2048 | 512 | ok | **wrong** |

The wrong values are not stable across runs (seed 1 gave 1.99e+00 and 1.46e+00
on consecutive invocations), which suggests reads landing outside the intended
region rather than a fixed indexing slip.

</details>

### Possible fixes

1. Require full row-contiguity for the gather variants' quantized inputs —
   contained, costs a copy when a caller passes a slice.
2. Pass the leading stride into `gather_qmm_rhs` and index experts with it — no
   copy, more kernel surface.

I have (1) plus a regression test ready and will open a PR right after this;
happy to redo it as (2) if you prefer.

Related but distinct: #3200, #3444.

dudududukim:
## Proposed changes

Fixes #4055.

`GatherQMM::eval_gpu` normalizes `w`, `scales` and `biases` with
`ensure_row_contiguous_matrix`, which validates only the last two axes. A row
slice of an `[E, 2R, D]` expert weight passes that check while keeping the
leading stride of the original array, so the sorted path reads every expert
after the first from the wrong offset — silently, with no error.

The gather kernels index the leading axes themselves, so this switches those
three inputs to `ensure_row_contiguous`, which already exists in the file and
is already used for `indices`. `x` is left as is; the failure I can demonstrate
is in the weight path.

Verified on M5 Pro / macOS 26.5.1: the new test fails before the change and
passes after, and `Ran 794 tests ... OK (skipped=46)` for the full suite.

## Checklist


zcbenz (commented):
Thanks for the fix, I can verify the bug and the fix, but it is strange that `gather_qmm_rhs` actually uses `ensure_row_contiguous` inside it:

https://github.com/ml-explore/mlx/blob/f599c0209f606ed7faad5c78a6c761e222da4e63/mlx/backend/metal/quantized.cpp#L1612-L1613

Do you have an idea what might go wrong?

dudududukim:
Thank you for pointing that out — it led to the actual root cause.

The biases `ensure_row_contiguous` runs *mid-encode*: after
`set_compute_pipeline_state` and the first three `set_input_array` calls. For
a non-contiguous biases it encodes a copy kernel on the same encoder right
there, clobbering the state of the kernel being set up. `w` and `scales` are
normalized at the top of the function, so they were never affected — isolating
the inputs on 0.32.0 confirms only strided biases reproduces the bug
(w only: 0.0, scales only: 0.0, biases only: 1.41).

Updated the PR: the biases normalization is hoisted next to `w`/`scales` in
both `gather_qmm_rhs` and `gather_qmm_rhs_nax`, and the `eval_gpu` change is
reverted. Test unchanged, red before / green after, full `test_quantized`
passes on M5 Pro.

zcbenz (approved):
Nice fix!

PR #4059, opened 2026-08-07
Fix DeviceType annotations for device and stream arguments

XXXXRT666:
## Proposed changes

`mx.cpu` and `mx.gpu` are `DeviceType` values. Nanobind accepts them at runtime through the existing implicit conversion from `DeviceType` to `Device`, but the generated Python stubs only exposed `Device`. This caused type checkers to reject valid calls such as:

```python
mx.exp(x, stream=mx.gpu)
```

This PR:

- Adds `DeviceType` as an explicit `StreamOrDevice` alternative.
- Updates stream argument annotations across core, random, linalg, fast, and distributed APIs.
- Includes `ThreadLocalStream`, which is already accepted by `StreamOrDevice`, in those annotations.
- Adds `DeviceType`-aware annotations to device and stream APIs that rely on nanobind's implicit conversion.

## Checklist

### DIFF
--- a/mlx/utils.cpp
+++ b/mlx/utils.cpp
@@ -18,6 +18,8 @@ Stream to_stream(StreamOrDevice s) {
     return default_stream(default_device());
   } else if (std::holds_alternative<Device>(s)) {
     return default_stream(std::get<Device>(s));
+  } else if (std::holds_alternative<Device::DeviceType>(s)) {
+    return default_stream(std::get<Device::DeviceType>(s));
   } else if (std::holds_alternative<ThreadLocalStream>(s)) {
     return stream_from_thread_local_stream(std::get<ThreadLocalStream>(s));
   } else {
@@ -30,6 +32,10 @@ Stream to_stream(StreamOrDevice s, Device default_) {
     return default_stream(default_);
   } else if (std::holds_alternative<Device>(s)) {
     return default_stream(std::get<Device>(s));
+  } else if (std::holds_alternative<Device::DeviceType>(s)) {
+    return default_stream(std::get<Device::DeviceType>(s));
+  } else if (std::holds_alternative<ThreadLocalStream>(s)) {
+    return stream_from_thread_local_stream(std::get<ThreadLocalStream>(s));
   } else {
     return std::get<Stream>(s);
   }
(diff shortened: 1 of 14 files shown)


python/src/distributed.cpp
```diff
@@ -170,7 +170,7 @@ void init_distributed(nb::module_& parent_module) {
       "group"_a = nb::none(),
       "stream"_a = nb::none(),
       nb::sig(
-          "def all_sum(x: array, *, group: Optional[Group] = None, stream: Union[None, Stream, Device] = None) -> array"),
+          "def all_sum(x: array, *, group: Optional[Group] = None, stream: Union[None, Stream, ThreadLocalStream, Device, DeviceType] = None) -> array"),
```
zcbenz: Is it possible to define a `StreamOrDevice` alias so we can just do `stream: StreamOrDevice = None`? Current format is horrible to write, and makes the docs a mess.

mlx/utils.cpp
```diff
@@ -18,6 +18,8 @@ Stream to_stream(StreamOrDevice s) {
     return default_stream(default_device());
   } else if (std::holds_alternative<Device>(s)) {
     return default_stream(std::get<Device>(s));
+  } else if (std::holds_alternative<Device::DeviceType>(s)) {
```
zcbenz, line 21: This adds a little runtime overhead, though negligible, if the implicit conversion already works perfectly I think there is no need to do it explicitly.

python/src/distributed.cpp
XXXXRT666: Done. I added a `StreamOrDevice` alias and updated all signatures to use it

mlx/utils.cpp
XXXXRT666: The issue is that nanobind derives generated signatures from the variant alternatives, and implicit conversions are not reflected there. Keeping `DeviceType` in `StreamOrDevice` makes the generated signatures match the inputs accepted at runtime and avoids maintaining per-function stub overrides.

image
image


image
image

zcbenz: That makes sense, thanks for clarification.

XXXXRT666:
Another question: Do you prefer using `Union/Optional`, or the Python 3.10-style `A | B` type annotation?

Because nanobind automatically generates the `A | B` syntax, while the existing codebase still uses the legacy `Union/Optional` style.

zcbenz:
I don't really have a preference, but it feels more reasonable following what nanobind does.

XXXXRT666:
I think the CI is failing because the non-Metal export cannot recognize `DeviceType`. Trying to fix it.

zcbenz (approved):
Thanks!

Issue #3341, opened 2026-03-31
rocm / cuda / metal - bfloat8/float8 support


Geramy:
Hi guys, I been working on the rocm support and I was wondering if you object to adding float8 support and types for native optimizations in backends?

zcbenz:
I think it is a nice thing to have but as you might have already noticed that we are bottlenecked at the number of reviewers.

So for adding new dtypes I think the only reasonable way to make progress is to refactor existing code to use the [`dispatch_types` utils](https://github.com/ml-explore/mlx/blob/117b4f18061be46acd91c2d50ac45bfd3094f32f/mlx/dtype_utils.h#L47) instead of `if`/`switch` first, which can be done in batches of small PRs that we would be happy to review and merge. And then we can gradually add the new dtype for each backend without worrying about missing a function.

If you just start a PR touch hundreds of files I think it would not go anywhere.

Geramy:
> I think it is a nice thing to have but as you might have already noticed that we are bottlenecked at the number of reviewers.
> 
> So for adding new dtypes I think the only reasonable way to make progress is to refactor existing code to use the [`dispatch_types` utils](https://github.com/ml-explore/mlx/blob/117b4f18061be46acd91c2d50ac45bfd3094f32f/mlx/dtype_utils.h#L47) instead of `if`/`switch` first, which can be done in batches of small PRs that we would be happy to review and merge. And then we can gradually add the new dtype for each backend without worrying about missing a function.
> 
> If you just start a PR touch hundreds of files I think it would not go anywhere.

Perfect that sounds great, and the fun begins haha! This will go great with the rocm implementation too! Can metal take advantage on the hardware level too? Example rocm / RDNA 3.5 has two 8-bit capable FPU per CU which should greatly increase speed. I would assume.

zcbenz:
I have no idea about float8 support in metal actually, @angeloskath @jagrit06 should know about it.

Geramy:
Pushed a PR regarding this, not implementing functions but just the base needed to start.
The absolute minimum to not create any compile errors.
https://github.com/ml-explore/mlx/pull/3374

zcbenz:
I just realized there was already an issue on this, I'm merging this to https://github.com/ml-explore/mlx/issues/1670.

PR #4064, opened 2026-08-08
Refactor NumberOfElements dtype dispatch

PhysicistJohn:
Replace the exhaustive dtype switch in `NumberOfElements::eval` with the existing `dispatch_all_types` utility.

This preserves the supported dtypes and existing conversion behavior while centralizing the type mapping. It follows the small-PR dtype-dispatch cleanup suggested in [#3341](https://github.com/ml-explore/mlx/issues/3341#issuecomment-4160792368).

### Testing

- Metal-enabled Python wheel build
- Serial native C++ test suite
- Shapeless compiled mean on CPU and GPU
- `pre-commit` formatting checks

### DIFF
--- a/mlx/backend/common/common.cpp
+++ b/mlx/backend/common/common.cpp
@@ -3,6 +3,7 @@
 
 #include "mlx/backend/common/broadcasting.h"
 #include "mlx/backend/common/utils.h"
+#include "mlx/dtype_utils.h"
 #include "mlx/primitives.h"
 
 namespace mlx::core {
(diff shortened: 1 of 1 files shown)


mlx/backend/common/common.cpp
```diff
@@ -99,50 +100,9 @@ void NumberOfElements::eval(const std::vector<array>& inputs, array& out) {
...
-      *out.data<bfloat16_t>() = static_cast<bfloat16_t>(numel);
-      break;
-    case float64:
-      *out.data<double>() = static_cast<double>(numel);
-      break;
-    case complex64:
-      *out.data<complex64_t>() = static_cast<complex64_t>(numel);
-      break;
-  }
+  dispatch_all_types(out.dtype(), [&]<typename T>(type_identity<T>) {
```
zcbenz: Thanks for doing the refactoring, though I'm not sure this is correct, it does not match other `dispatch_all_types` calls.

PhysicistJohn: Good catch. I updated this to match the existing `auto type_tag` / `MLX_GET_TYPE(type_tag)` pattern, and made the same change in #4065 and #4066. The focused builds and tests still pass.

zcbenz (approved):

PR #4067, opened 2026-08-08
Fix `mx.maximum` / `mx.minimum` / `mx.max` / `mx.min` handle signed zeros (`+0.0` vs `-0.0`) incorrectly

### LINKED ISSUE #4063: [BUG]`mx.maximum` / `mx.minimum` / `mx.max` / `mx.min` handle signed zeros (`+0.0` vs `-0.0`) incorrectly
**Describe the bug**
IEEE 754 (`maxNum`/`minNum`), C99 (`fmax`/`fmin`), and NumPy mandate explicit
rules for the `+0.0` / `-0.0` case:

  - `maximum(+0.0, -0.0) == +0.0` (and the operation is **commutative**)
  - `minimum(+0.0, -0.0) == -0.0`

MLX implements `Maximum`/`Minimum` (`mlx/backend/metal/kernels/binary_ops.h`)
and the reductions `Max`/`Min` (`mlx/backend/metal/kernels/reduction/ops.h`)
with a plain ordered comparison and **no `+0`/`-0` rule**:

```cpp
// binary_ops.h Maximum (float overload)
if (metal::isnan(x)) return x;
return x > y ? x : y;
```

When `x == y` (in particular for `+0.0` vs `-0.0`, since `+0 == -0`), the result
takes the **second operand's sign**, so `maximum`/`minimum` become
non-commutative on signed zeros and the `max` reduction returns `-0.0` where
`+0.0` is required. This is observable: `1 / mx.max([-0.0, +0.0])` returns
`-inf` instead of `+inf`.

**To Reproduce**

Include code snippet
```python
import mlx.core as mx
import numpy as np, struct

def neg0(x):
    a = np.asarray(np.array(x)).astype(np.float32)
    return (struct.unpack('<I', a.tobytes())[0] & 0x80000000) != 0

print(neg0(mx.maximum(mx.array(np.float32(0.0)), mx.array(np.float32(-0.0)))))  # True  (-0.0)   BUG: np -> +0.0
print(neg0(mx.max(mx.array(np.array([-0.0, 0.0, -0.0], dtype=np.float32)))))   # True  (-0.0)   BUG: np -> +0.0
print(float(np.array(mx.reciprocal(
    mx.max(mx.array(np.array([-0.0, 0.0], dtype=np.float32))))).item()))        # -inf            BUG: np -> +inf
```

**Expected behavior**
```
  maximum(-0, +0)   : mlx=+0.0  np=+0.0  OK
  maximum(+0, -0)   : mlx=-0.0  np=+0.0  WRONG
  minimum(-0, +0)   : mlx=+0.0  np=-0.0  WRONG
  minimum(+0, -0)   : mlx=-0.0  np=-0.0  OK
  max([-0,+0,-0])   : mlx=-0.0  np=+0.0  WRONG
  min([-0,+0,-0])   : mlx=-0.0  np=-0.0  OK

  1 / max([-0, +0]) : mlx=-inf  np=+inf  WRONG (sign flip)
```

JasonHonKL:
## Proposed changes

Fix #4063 

## Checklist

### DIFF
--- a/mlx/backend/cpu/simd/base_simd.h
+++ b/mlx/backend/cpu/simd/base_simd.h
@@ -235,7 +235,11 @@ Simd<T, 1> maximum(Simd<T, 1> a_, Simd<T, 1> b_) {
       return a;
     }
   }
-  return (a > b) ? a : b;
+  if constexpr (std::is_floating_point_v<T>) {
+    return (a > b || (a == b && !std::signbit(a))) ? a : b;
+  } else {
+    return (a > b) ? a : b;
+  }
 }
 
 template <typename T>
@@ -247,7 +251,11 @@ Simd<T, 1> minimum(Simd<T, 1> a_, Simd<T, 1> b_) {
       return a;
     }
   }
-  return (a < b) ? a : b;
+  if constexpr (std::is_floating_point_v<T>) {
+    return (a < b || (a == b && std::signbit(a))) ? a : b;
+  } else {
+    return (a < b) ? a : b;
+  }
 }
 
 template <typename T>
--- a/mlx/backend/metal/kernels/binary_ops.h
+++ b/mlx/backend/metal/kernels/binary_ops.h
@@ -163,7 +163,7 @@ struct Maximum {
     if (metal::isnan(x)) {
       return x;
     }
-    return x > y ? x : y;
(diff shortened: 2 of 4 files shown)


zcbenz:
Closing per https://github.com/ml-explore/mlx/issues/4063#issuecomment-5230910009.

PR #4068, opened 2026-08-08
Handle zero-size inputs in eig and qr on the CPU

ayaangazali:
## Proposed changes

#4033 and #3834 fixed zero-size inputs for `cholesky`, `eigh` and `svd`. `eig` and `qr` have the same hole and were not covered by either.

`eig` aborts the process:

```python
>>> mx.eval(mx.linalg.eig(mx.zeros((0, 0)), stream=mx.cpu))
** On entry to SGEEV , parameter number  5 had an illegal value
[Eig::eval_cpu] Eigenvalue decomposition failed with error code -5
abort
```

`qr` returns the right shapes but still calls LAPACK with a zero leading dimension, so it prints on every call:

```python
>>> q, r = mx.linalg.qr(mx.zeros((0, 0)), stream=mx.cpu); mx.eval(q, r)
** On entry to SGEQRF, parameter number  4 had an illegal value
** On entry to SORGQR, parameter number  5 had an illegal value
```

Both come from the same place the earlier two did. `num_matrices` is `a.size() / (M * N)`, which divides by zero when a dimension is zero, and the workspace query runs before the loop either way, so LAPACK gets `lda = 0`. numpy returns empty results for all of these.

Guards added in the same spirit as the merged ones:

- `eig.cpp`, after `values.set_data`. The input is square, so `n = 0` means both outputs are empty and there is nothing to write. No identity fill is needed here, unlike `svd`.
- `qrf.cpp`, before `num_matrices` is computed, so the zero division never happens. `q` is `(..., M, k)` and `r` is `(..., k, N)` with `k = min(M, N)`, so if either dimension is zero then `k` is zero and both outputs are empty whichever one it was.

Shapes now match numpy for `(0, 0)`, `(3, 0, 0)`, `(0, 4, 4)`, and for `qr` also `(5, 0)` and `(0, 5)`.

## Checklist

Worth being precise about the two tests, since they are not equally strong. The `eig` one aborts the interpreter on main, so it fails there in the loudest possible way. The `qr` one passes on main, because the shapes were already correct and the only visible symptom was the LAPACK output above; it is a regression guard for the new early return rather than proof of the old break. Say the word if you would rather I drop it.

CPU only build (`MLX_BUILD_METAL=OFF`) at f599c020, `test_linalg.py` 18 tests OK, and non-empty `eig` and `qr` verified unchanged against their existing reconstruction checks.

---

freshman contributor, i lean on Claude Code while digging. i had actually built a version of this a day earlier that also touched cholesky, eigh and svd, then found #4033 and #3834 already open and threw it away rather than compete. this is only the part those two left behind.

### DIFF
--- a/mlx/backend/cpu/eig.cpp
+++ b/mlx/backend/cpu/eig.cpp
@@ -241,6 +241,15 @@ void Eig::eval_cpu(
 
   values.set_data(allocator::malloc(values.nbytes()));
 
+  // Nothing to decompose for n = 0; LAPACK rejects lda = 0. The input is
+  // square, so both outputs are empty and there is nothing to write.
+  if (a.shape(-1) == 0) {
+    if (compute_eigenvectors_) {
+      vectors.set_data(allocator::malloc(vectors.nbytes()));
+    }
+    return;
+  }
+
   if (compute_eigenvectors_) {
     // Set the strides and flags so the eigenvectors
     // are in the columns of the output
--- a/mlx/backend/cpu/eig.cpp
+++ b/mlx/backend/cpu/eig.cpp
@@ -228,6 +228,19 @@ void Eig::eval_cpu(
   const auto& a = inputs[0];
   auto& values = outputs[0];
 
+  values.set_data(allocator::malloc(values.nbytes()));
+
+  // Nothing to decompose for n = 0; LAPACK rejects lda = 0. The input is
+  // square, so both outputs are empty and there is nothing to write. Return
+  // before the copy below, since the temporary it produces is only kept alive
+  // by the add_temporary in eig_impl.
+  if (a.shape(-1) == 0) {
+    if (compute_eigenvectors_) {
+      outputs[1].set_data(allocator::malloc(outputs[1].nbytes()));
+    }
+    return;
+  }
+
   auto vectors = compute_eigenvectors_
       ? outputs[1]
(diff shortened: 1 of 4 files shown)


zcbenz (approved):
Thanks!

ayaangazali:
The cuda-12.6 x86_64 failure was mine, pushed a fix.

```
test_eig (test_linalg.TestLinalg.test_eig) ... terminate called after throwing an instance of 'std::runtime_error'
  what():  cudaMemcpy(data, buf.data, buf.size, kind) failed: invalid argument
```

My guard in `Eig::eval_cpu` returned after `copy_cpu(a, a_copy, ...)` had already queued work, but the `encoder.add_temporary(a_copy)` that keeps that buffer alive lives inside `eig_impl`, which the guard skipped. So `a_copy` was released with its copy task still pending, and the freed buffer reached `CudaAllocator::move_to_unified_memory` (allocator.cpp:296). That is why it showed up on one runner and not the others: cuda-12.9 and cuda-13.0 x86_64 both ran `test_eig` and reported ok.

Moved the guard above the copy so no temporary is created at all, which is the same ordering the merged eigh guard uses. `qr` was already returning before its copy so it did not have this problem.

I only have a CPU box, so I could not reproduce the CUDA abort locally. What I can say is that the dangling temporary is gone rather than just no longer observable.

PR #4071, opened 2026-08-08
Fix: mx.distributed.sum_scatter crashes on a scalar(0-d)

### LINKED ISSUE #4070: [BUG] `mx.distributed.sum_scatter` crashes (assertion failure / UB) on a scalar (0-d) input
**Describe the bug**

`mx.distributed.sum_scatter` accesses `x.shape()[0]` to check divisibility by
the group size **without first checking that the array has any dimensions**.
For a 0-dimensional (scalar) input `x.shape()` is empty, so `x.shape()[0]` is
an out-of-bounds vector access and the process aborts:

```
Assertion failed: (size() > index), function operator[], file small_vector.h, line 315.
```

instead of raising a graceful `std::invalid_argument` like every other
distributed op does (e.g. `send`/`recv` validate `dst`/`src`, and
`sum_scatter`'s *own* non-divisible-size check throws a clear message).

Reproduced with the bundled 2-rank local ring backend; the crash happens during
graph construction (`mlx/distributed/ops.cpp`), so it fires identically on
nccl/mpi backends before reaching the backend implementation.

Include code snippet
```python
import mlx.core as mx
# on a multi-rank group:
g = mx.distributed.init()                       
sc = mx.distributed.sum_scatter(mx.array(1.0))  
mx.eval(sc)
```

**Expected behavior**
Gracefully return the error.

JasonHonKL:
## Proposed changes

Fix #4070 

## Checklist

### DIFF
--- a/mlx/distributed/ops.cpp
+++ b/mlx/distributed/ops.cpp
@@ -165,11 +165,11 @@ array sum_scatter(
   if (group.size() == 1) {
     return x;
   }
-  if (x.shape()[0] % group.size() != 0) {
+  if (x.ndim() == 0 || x.shape()[0] % group.size() != 0) {
     std::ostringstream msg;
     msg << "[sum_scatter] Invalid shape=" << x.shape()
         << " for a group of size " << group.size()
-        << ". The first dimension (axis 0) must be divisible by the group size.";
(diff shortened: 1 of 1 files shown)


zcbenz (approved):
Thanks!

PR #4073, opened 2026-08-08
Fix distributed NaN Issue 

### LINKED ISSUE #4072: [BUG] `mx.distributed.all_max` / `all_min` silently drop `NaN` (inconsistent with `mx.max`/`mx.min`)
**Describe the bug**

The distributed cross-rank reduce uses `std::max` / `std::min`
(`mlx/distributed/reduction_ops.h:17-35`), which are **not** NaN-propagating
(`std::max(nan, x)` returns `x` when `nan` is the first argument). So if
**exactly one rank** has a `NaN` at a position, `all_max` / `all_min` return
the other ranks' non-`NaN` value and the `NaN` vanishes — the failure is
silently swallowed.

**To Reproduce**

```python
import mlx.core as mx
# 2-rank group
nan = float("nan")
x = mx.array([1.0, nan if rank == 0 else 2.0, 3.0])  # rank 0 has NaN at idx 1
m = mx.distributed.all_max(x)
mx.eval(m)   # -> array([1.0, 2.0, 3.0])   WRONG: should be [1.0, nan, 3.0]
```

**Expected behavior**
```
[rank 0] AMAX [1.0, 2.0, 3.0]
[rank 0] AMIN [1.0, 0.5, 3.0]
[rank 1] AMAX [1.0, 2.0, 3.0]
[rank 1] AMIN [1.0, 0.5, 3.0]
local mx.max([1,nan,3]) = nan (propagates NaN)
local mx.min([1,nan,3]) = nan (propagates NaN)

BUG REPRODUCED: all_max dropped the single-rank NaN (mx.max would propagate it).
```

JasonHonKL:
## Proposed changes
Fix #4072 

@zcbenz This is quite a massive change. More specifically it fix the NaN issue.  Not sure if it is suitable to open a file or multiple issue. 

## Checklist

### DIFF
--- a/mlx/distributed/jaccl/lib/jaccl/reduction_ops.h
+++ b/mlx/distributed/jaccl/lib/jaccl/reduction_ops.h
@@ -2,6 +2,7 @@
 
 #pragma once
 
+#include <cmath>
 #include <cstddef>
 
 #include "jaccl/types.h"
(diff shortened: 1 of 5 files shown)


zcbenz:
I'm leaving this for @angeloskath to decide.

zcbenz:
So this is a real bug but this PR is way too messy and there is no way we would accept the change as is. At the moment we don't really have time to cleanup things like this and we would like to keep the bug open until there are actual real cases requiring the bug to be fixed.

JasonHonKL:
sure !

PR #4074, opened 2026-08-08
Fix hardcoding pow2=10 in get_block_dims_common.

### LINKED ISSUE #4062: get_block_dims_common ignores its pow2 argument (hard-coded sum == 10)
### Describe the bug

`get_block_dims_common` accepts a `pow2` budget but hard-codes `sum == 10` in two of its three break conditions, so a caller asking for a smaller block is frequently ignored and can receive up to a full 1024-thread block.

`mlx/backend/common/utils.cpp`:

```cpp
Dims get_block_dims_common(int dim0, int dim1, int dim2, int pow2 /* = 10 */) {
  int pows[3] = {0, 0, 0};
  int sum = 0;
  while (true) {
    int presum = sum;
    if (dim0 >= (1 << (pows[0] + 1))) { pows[0]++; sum++; }
    if (sum == 10) { break; }            // <-- ignores pow2
    if (dim1 >= (1 << (pows[1] + 1))) { pows[1]++; sum++; }
    if (sum == 10) { break; }            // <-- ignores pow2
    if (dim2 >= (1 << (pows[2] + 1))) { pows[2]++; sum++; }
    if (sum == presum || sum == pow2) { break; }
  }
  return std::make_tuple(1ul << pows[0], 1ul << pows[1], 1ul << pows[2]);
}
```

Two failure modes:

1. `sum` can pass `pow2` between the intra-iteration checks. Only the third condition tests `pow2`, and it tests equality, so an exceeded budget is never caught.
2. The first two checks compare against the literal `10`, so any `pow2 < 10` is only honoured at one of three exit points.

### To Reproduce

Faithful transcription of the loop, swept over `pow2` in `1..9` and dims in `{1, 2, 64, 1024, 2^20}`, yields **473** combinations that exceed the requested budget. Smallest illustrative cases:

| Requested `pow2` | Budget | dims | Returned block | Threads |
|---|---|---|---|---|
| 3 | 8 | (1, 64, 64) | (1, 32, 32) | **1024** |
| 5 | 32 | (1, 64, 64) | (1, 32, 32) | **1024** |
| 9 | 512 | (1, 64, 64) | (1, 32, 32) | **1024** |
| 1 | 2 | (1, 2, 2) | (1, 2, 2) | 4 |

So `get_block_dims(1, 64, 64, /*pow2=*/3)` requests 8 threads and returns 1024.

### Expected behavior

Comparing against `pow2` and using `>=` rather than `==` keeps every exit inside the requested budget:

```cpp
    if (dim0 >= (1 << (pows[0] + 1))) { pows[0]++; sum++; }
    if (sum >= pow2) { break; }
    if (dim1 >= (1 << (pows[1] + 1))) { pows[1]++; sum++; }
    if (sum >= pow2) { break; }
    if (dim2 >= (1 << (pows[2] + 1))) { pows[2]++; sum++; }
    if (sum == presum || sum >= pow2) { break; }
```

With this change all 473 cases fall within budget, and behaviour at the default `pow2 = 10` is unchanged.

### Why it matters

This matters for callers sizing a block against a pipeline's `maxTotalThreadsPerThreadgroup`. A kernel whose register or threadgroup-memory usage lowers its occupancy reports a maximum below 1024; passing a correspondingly reduced `pow2` is silently ignored, and the oversized dispatch then trips `-[MTLDebugComputeCommandEncoder _validateThreadsPerThreadgroup:]`.

### Desktop (please complete the following information)

- OS Version: macOS 27.0
- Version: current `main` (`47bbfe8`); also present in `ce45c525`

FeliGame:
## Proposed changes
Fix [4062](https://github.com/ml-explore/mlx/issues/4062)
Fix hardcoding pow2=10 process logic to a more reasonable and robust way.

## Checklist

### DIFF
--- a/mlx/backend/common/utils.cpp
+++ b/mlx/backend/common/utils.cpp
@@ -124,21 +124,21 @@ Dims get_block_dims_common(int dim0, int dim1, int dim2, int pow2 /* = 10 */) {
       pows[0]++;
       sum++;
     }
-    if (sum == 10) {
+    if (sum >= pow2) {
       break;
     }
     if (dim1 >= (1 << (pows[1] + 1))) {
       pows[1]++;
       sum++;
     }
-    if (sum == 10) {
+    if (sum >= pow2) {
       break;
     }
     if (dim2 >= (1 << (pows[2] + 1))) {
       pows[2]++;
       sum++;
     }
-    if (sum == presum || sum == pow2) {
+    if (sum == presum || sum >= pow2) {
       break;
     }
   }


zcbenz:
Closing per https://github.com/ml-explore/mlx/issues/4062#issuecomment-5225614043.

PR #4077, opened 2026-08-08
Read each K/V byte once in gqa-8 decode attention

### LINKED ISSUE #4076: sdpa_vector_2pass_1 loads K/V once per query head instead of once per kv head
On M-series decode, `sdpa_vector_2pass_1` assigns one simdgroup per query head, and the `gqa_factor` simdgroups of a threadgroup all stream the same K/V block (the `keys`/`values` offsets don't depend on the simdgroup index). Loads therefore scale with query heads, not kv heads, and for models with few kv heads the kernel saturates load-issue before DRAM.

Measured effective bandwidth of `mx.fast.scaled_dot_product_attention` (M5 Pro 24GB, fp16, q_len 1, KV bytes counted once, distinct KV buffers cycled between calls to keep SLC out of the numbers):

| geometry | ctx 2K | ctx 8K | ctx 32K |
|---|---|---|---|
| 32q/32kv (gqa 1) | 239 GB/s | 282 GB/s | 281 GB/s |
| 32q/8kv (gqa 4) | 234 GB/s | 263 GB/s | 264 GB/s |
| 32q/4kv/hd128 (gqa 8, Qwen3-30B/235B) | 130 GB/s | 184 GB/s | **190 GB/s** |
| 64q/8kv/hd64 (gqa 8, gpt-oss-20b) | 76 GB/s | 113 GB/s | **119 GB/s** |

A dense fp16 qmv on the same machine streams at 266 GB/s, so the gqa-8 rows are kernel-shape, not hardware.

At 32K context this is most of the decode step: Qwen3-30B-A3B-4bit (truncated to 24 layers so 32K KV fits in 24 GB) spends ~60% of its 15 ms/step streaming KV at this reduced rate.

Reassigning the work so that each simdgroup owns a contiguous token sub-chunk and computes several of its group's query heads from registers (so each K/V byte is read `gqa_factor / heads_per_simd` times instead of `gqa_factor`) recovers most of the gap: 1.14-1.27x kernel-level at 32K, and +9.5% / +9.2% end-to-end on Qwen3-30B-A3B and Qwen2.5-3B, with the second pass unchanged.

dudududukim:
Closes #4076.

## Proposed changes

`sdpa_vector_2pass_1` has every simdgroup of a threadgroup stream the same K/V block, so loads scale with query heads instead of kv heads and gqa-8 models decode well below the machine's bandwidth (195 GB/s for 32q/4kv/hd128 and 120 GB/s for 64q/8kv/hd64 at 32K context, against 266 GB/s for a dense qmv on the same M5 Pro).

This adds `sdpa_vector_2pass_1_gqa`: each simdgroup owns a contiguous token sub-chunk of the block and computes `HPT` of its group's query heads from registers, so each K/V byte is read `gqa_factor / HPT` times instead of `gqa_factor`. The simdgroups' partials are merged once through threadgroup memory at the end; the partials layout and `sdpa_vector_2pass_2` are unchanged. `HPT` is a template parameter because register pressure decides it: 8 heads per simdgroup wins at head_dim 64 but spills and loses at head_dim 128, where 4 is right.

Dispatch is deliberately narrow: single-token queries without mask or sinks, `gqa_factor == 8`, head_dim 64 or 128, `N >= 8192`. Everything else stays on the existing kernel. Sinks and other gqa factors can follow if this direction looks good.

All numbers below are from an M5 Pro 24 GB on AC, bf16, with K/V passed as strided cache slices and distinct buffers cycled between calls so the SLC doesn't flatter either arm. Both arms run in one process with the dispatch flipped between them (through a temporary env check that is not part of this diff), interleaved ABBA over several rounds; the tables give the median.

Kernel level:

| geometry | ctx | main | this PR | speedup |
|---|---|---|---|---|
| 32q/4kv/hd128 | 8192 | 90.8 us (185 GB/s) | 84.8 us (198 GB/s) | 1.07x |
| 32q/4kv/hd128 | 16384 | 176.9 us (190 GB/s) | 162.4 us (207 GB/s) | 1.09x |
| 32q/4kv/hd128 | 32768 | 343.5 us (195 GB/s) | 301.5 us (223 GB/s) | 1.14x |
| 64q/8kv/hd64 | 8192 | 147.1 us (114 GB/s) | 127.7 us (131 GB/s) | 1.15x |
| 64q/8kv/hd64 | 16384 | 288.5 us (116 GB/s) | 247.5 us (136 GB/s) | 1.17x |
| 64q/8kv/hd64 | 32768 | 561.6 us (120 GB/s) | 440.6 us (152 GB/s) | 1.27x |

Per-round spread is within 0.01x of the median for every row above.

End-to-end decode, ms/step, on Qwen3-30B-A3B-4bit truncated to 24 layers (so 32K of KV fits in 24 GB; 32q/4kv/hd128) and on the full 36-layer Qwen2.5-3B-Instruct-4bit (16q/2kv/hd128). Context 4096 is below the dispatch gate, so both arms run the identical kernel there and it serves as a control:

| ctx | 30B main | 30B this PR | delta | 3B main | 3B this PR | delta |
|---|---|---|---|---|---|---|
| 4096 (control) | 7.289 | 7.292 | -0.0% | 9.626 | 9.626 | -0.0% |
| 8192 | 8.596 | 8.301 | +3.5% | 10.451 | 10.178 | +2.7% |
| 16384 | 10.757 | 9.936 | +8.3% | 12.152 | 11.479 | +5.9% |
| 32760 | 15.025 | 13.726 | +9.5% | 15.674 | 14.355 | +9.2% |

Individual control rounds span -0.2%..+0.3% (30B) and -0.6%..+1.6% (3B), so the gated contexts sit well outside the noise floor.

Logits stay within bf16 rounding of the current kernel with argmax stable over greedy decode, and outputs match an fp32 reference to 1e-5 (fp16) and 2e-4 (bf16) on both gated and ungated shapes. The added test covers both instantiated head dims at a context that is not a multiple of the block count.

## Checklist

### DIFF
--- a/mlx/backend/metal/scaled_dot_product_attention.cpp
+++ b/mlx/backend/metal/scaled_dot_product_attention.cpp
@@ -506,6 +506,53 @@ void sdpa_vector_2pass(
   compute_encoder.add_temporary(sums);
   compute_encoder.add_temporary(maxs);
 
+  if (!mask && !sinks && q.shape(2) == 1 && gqa_factor == 8 &&
+      q.shape(-1) == v.shape(-1) && (q.shape(-1) == 64 || q.shape(-1) == 128) &&
+      N >= 8192) {
+    std::string gname;
+    gname.reserve(64);
+    gname += "sdpa_vector_2pass_1_gqa_";
+    gname += get_type_string(q.dtype());
+    gname += "_";
+    gname += std::to_string(q.shape(-1));
+    gname += "_";
+    gname += std::to_string(v.shape(-1));
+    auto kernel = d.get_kernel(gname);
+    check_kernel_threadgroup_size(kernel, group_dims, gname);
+    compute_encoder.set_compute_pipeline_state(kernel);
+    compute_encoder.set_input_array(q, 0);
+    compute_encoder.set_input_array(k, 1);
+    compute_encoder.set_input_array(v, 2);
+    compute_encoder.set_output_array(intermediate, 3);
+    compute_encoder.set_output_array(sums, 4);
+    compute_encoder.set_output_array(maxs, 5);
+    compute_encoder.set_bytes(N, 7);
+    compute_encoder.set_bytes(k_head_stride, 8);
+    compute_encoder.set_bytes(k_seq_stride, 9);
+    compute_encoder.set_bytes(v_head_stride, 10);
+    compute_encoder.set_bytes(v_seq_stride, 11);
+    compute_encoder.set_bytes(scale, 12);
+    compute_encoder.dispatch_threadgroups(grid_dims, group_dims);
+
+    kname.clear();
+    kname = "sdpa_vector_2pass_2_";
+    kname += get_type_string(q.dtype());
+    kname += "_";
+    kname += std::to_string(v.shape(-1));
+    kernel = d.get_kernel(kname);
+    compute_encoder.set_compute_pipeline_state(kernel);
+    compute_encoder.set_input_array(intermediate, 0);
+    compute_encoder.set_input_array(sums, 1);
+    compute_encoder.set_input_array(maxs, 2);
+    compute_encoder.set_output_array(out, 3);
+    compute_encoder.set_bytes(blocks, 4);
+    group_dims = MTL::Size(1024, 1, 1);
+    grid_dims = MTL::Size(q.shape(0) * q.shape(1), q.shape(2), 1);
+    check_kernel_threadgroup_size(kernel, group_dims, kname);
+    compute_encoder.dispatch_threadgroups(grid_dims, group_dims);
+    return;
+  }
+
   bool has_mask = mask.has_value();
   bool bool_mask = has_mask && (*mask).dtype() == bool_;
   bool float_mask = has_mask && !bool_mask;
--- a/mlx/backend/metal/scaled_dot_product_attention.cpp
+++ b/mlx/backend/metal/scaled_dot_product_attention.cpp
@@ -430,7 +430,13 @@ void sdpa_vector_2pass(
   // Set the kernel name
   std::string kname;
   kname.reserve(64);
-  kname += "sdpa_vector_2pass_1_";
+  kname += "sdpa_vector_2pass_1";
+  if (!mask && !sinks && q.shape(2) == 1 && q.shape(1) == 8 * k.shape(1) &&
+      q.shape(-1) == v.shape(-1) && (q.shape(-1) == 64 || q.shape(-1) == 128) &&
+      k.shape(2) >= 8192) {
+    kname += "_gqa";
+  }
+  kname += "_";
   kname += get_type_string(q.dtype());
   kname += "_";
   kname += std::to_string(q.shape(-1));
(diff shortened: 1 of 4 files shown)


erwinzhang7:
Verified on an M5 Max, 40 GPU cores, 128 GB, macOS 26.2. Your numbers are from an
M5 Pro 24 GB, so this is a different part rather than a bigger one, and the 128 GB
removes the truncation your 30B run needed.

`base` throughout is `6a0dd0fd2`, this branch's own parent, rather than current
main, since the branch carries a merge commit and diffing against main would
credit it with three weeks of unrelated work.

**The kernel win holds on wider silicon.** bf16, K/V as
strided cache slices with eight buffer sets cycled so the SLC cannot hold them,
12 rounds of 20 calls with one eval per round:

| geometry | ctx | base | this PR | speedup | yours |
|---|---:|---:|---:|---:|---:|
| 32q/4kv/hd128 | 8192 | 53.4 us | 46.2 us | 1.16x | 1.07x |
| 32q/4kv/hd128 | 16384 | 87.6 us | 80.9 us | 1.08x | 1.09x |
| 32q/4kv/hd128 | 32768 | 156.2 us | 139.3 us | 1.12x | 1.14x |
| 64q/8kv/hd64 | 8192 | 73.5 us | 61.4 us | 1.20x | 1.15x |
| 64q/8kv/hd64 | 16384 | 130.2 us | 109.4 us | 1.19x | 1.17x |
| 64q/8kv/hd64 | 32768 | 244.0 us | 186.6 us | 1.31x | 1.27x |

The baseline on this machine is 1.7x to 2.3x faster than yours on the same kernel
before any change. The worry with a bandwidth argument is that it stops applying on a part with
bandwidth to spare. The argument holds: larger on four of the six
geometries, and within a couple of points on the other two.

Nothing outside the gate moves at all. Both builds scored against each other
and against an fp32 reference:

| case | max abs diff vs base | vs fp32 |
|---|---:|---:|
| ungated, ctx 4096 below the gate | **0** | 3.5e-4 |
| ungated, gqa_factor 4 | **0** | 3.3e-4 |
| ungated, head_dim 80 | **0** | 4.6e-4 |
| ungated, 2 query tokens | **0** | 3.4e-4 |
| gated 32q/4kv/hd128 | 2.4e-4 | 2.8e-4 |
| gated 64q/8kv/hd64 | 2.4e-4 | 1.9e-4 |
| gated, ctx 8192+137 | 4.9e-4 | 2.4e-4 |

Bit-identical on every ungated shape, so the dispatch condition fires exactly
where you say and nowhere else. On the gated shapes the differences are bf16
rounding and the new kernel sits as close to fp32 as the old one, so it isn't
sacrificing accuracy for speed. The 8192+137 row would expose an off-by-one in the sub-chunk
partitioning, and it is clean.

**End to end, on all 48 layers rather than 24.** Same model, `Qwen3-30B-A3B-4bit`,
untruncated, 128 GB holds 32K of KV for the whole thing. 24 decode steps per
context after a full prefill, median:

| ctx | base | this PR | delta | yours, 24 layers |
|---:|---:|---:|---:|---:|
| 4096 | 9.303 ms | 9.260 ms | +0.5% | -0.0% |
| 8192 | 10.682 ms | 10.264 ms | **+3.9%** | +3.5% |
| 16384 | 12.757 ms | 12.177 ms | **+4.5%** | +8.3% |
| 32760 | 17.724 ms | 16.491 ms | **+7.0%** | +9.5% |

The win is real at full depth and clears the control comfortably. My 4096 control
comes out at +0.5% rather than zero (the two arms are separate processes with
separate model loads) so read 0.5% as this harness's noise floor.

However, the magnitude is smaller than the truncated run shows, and consistently:
+4.5% against +8.3% at 16K, +7.0% against +9.5% at 32K. That's from halving the
layer count: it removes half the non-attention work per step, so attention
occupies a larger share of what is left and any speedup measured against the step
scales up with it.

I'd quote +7.0% at 32760 on the full model rather than +9.5%. It is still a
good number for a decode path and it is one nobody can argue with. Good work!

zcbenz (commented):
~Can you add the affected shapes to `sdpa_bench.py`?~

Sorry just realized this is sdpa_vector, ignore this comment.

mlx/backend/metal/scaled_dot_product_attention.cpp
```diff
@@ -506,6 +506,53 @@ void sdpa_vector_2pass(
   compute_encoder.add_temporary(sums);
   compute_encoder.add_temporary(maxs);
 
+  if (!mask && !sinks && q.shape(2) == 1 && gqa_factor == 8 &&
+      q.shape(-1) == v.shape(-1) && (q.shape(-1) == 64 || q.shape(-1) == 128) &&
+      N >= 8192) {
+    std::string gname;
```
zcbenz: It seems that most of the host code is duplicated, I think you can just change the `kname`?

dudududukim: Right! 
Fixed in d1ab9ded8, the gate now just appends `_gqa` to the kname.
Tests pass and the dispatch still switches at the boundary.

zcbenz (approved):
Nice improvement, thanks!

Issue #1214, opened 2024-06-17
[Feature] GPU scatter support for complex64


credwood:
I'm trying to implement a VQ-VAQ with a loss function which uses the log Mel spectrogram. I can train my network using MSE and L1 loss with float32 values, but when I apply the Mel transform to these arrays I get the following error

```python
ValueError: [scatter] GPU scatter does not yet support complex64 for the input or updates.
```
Is this on the short-term roadmap? If not, I might try to implement the feature or maybe a short-term workaround.

System information:
 - OS Version: MacOS Sonoma 14.5
 - Version 0.15.0

barronalex:
Do you know which op within the log mel spectrogram is causing the error? My guess would be it's in padding the spectrogram?

Probably the quickest fix is to run the scatter op on the cpu with:
```python
with mx.stream(cpu):
  # scatter op
```

Or you could re-implement in terms of `mx.pad` which is how the [Whisper MLX example](https://github.com/ml-explore/mlx-examples/blob/a7598e9456c6455a07ff4905712c2ea3cfcd52db/whisper/mlx_whisper/audio.py#L104) does it.

We should definitely support the complex64 scatter op directly too, but hopefully the above unblocks you.

credwood:
> Do you know which op within the log mel spectrogram is causing the error? My guess would be it's in padding the spectrogram?
> 
> Probably the quickest fix is to run the scatter op on the cpu with:
> 
> ```python
> with mx.stream(cpu):
>   # scatter op
> ```
> 
> Or you could re-implement in terms of `mx.pad` which is how the [Whisper MLX example](https://github.com/ml-explore/mlx-examples/blob/a7598e9456c6455a07ff4905712c2ea3cfcd52db/whisper/mlx_whisper/audio.py#L104) does it.
> 
> We should definitely support the complex64 scatter op directly too, but hopefully the above unblocks you.

Thanks for the quick reply and the help! I swapped out my ```stft``` function for the one in the Whisper example linked above, and that triggered the same problem (I was already using ```mx.pad``` in my padding function). I adapted the ```log_mel_spectrogram```  function from that same Whisper example code and was able to get my model training by running the following lines in that function on CPU:

```
with mx.stream(mx.cpu):
        freqs = stft(audio, window, nperseg=n_fft, noverlap=hop_length)
        magnitudes = freqs[:, :-1, :].abs().square()
        filters = mel_filters(n_mels, n_fft)
        mel_spec = magnitudes @ filters.T
[...]

PR #4078, opened 2026-08-08
Enable complex64 scatter addition on GPU

PhysicistJohn:
## Summary

Enable the existing GPU scatter-add path for `complex64`. This supports
`.at[...].add(...)` and complex gather backward, addressing the additive case
reported in #1214.

Complex addition is componentwise. On Metal, `Scatter::Sum` accumulates the
real and imaginary FP32 lanes independently with native atomics. The shared
eight-byte GPU restriction is relaxed only for the existing `Scatter::Sum`
mode with `complex64`.

CUDA already has a CAS-based complex atomic-add implementation, so the same
guard change makes that path reachable. Assignment, product, min/max, other
eight-byte dtypes, and `ScatterAxis` remain unchanged.

## Testing

Coverage includes:

- 4,096 colliding `complex64` GPU updates
- Complex gather `value_and_grad`
- Python array tests on CPU and Metal
- Native scatter tests on CPU and Metal
- The focused regression with Metal API validation enabled

### DIFF
--- a/mlx/backend/metal/indexing.cpp
+++ b/mlx/backend/metal/indexing.cpp
@@ -230,7 +230,8 @@ void Gather::eval_gpu(const std::vector<array>& inputs, array& out) {
 }
 
 void Scatter::eval_gpu(const std::vector<array>& inputs, array& out) {
-  if (size_of(out.dtype()) == 8) {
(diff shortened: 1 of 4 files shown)


zcbenz (approved):
Thanks!

PR #4079, opened 2026-08-08
Only reject max and min when a reduced axis is empty

ayaangazali:
## Proposed changes

`max` and `min` reject any zero-size array, even when every axis being reduced is non-empty:

```python
>>> mx.max(mx.zeros((0, 2)), axis=-1)
ValueError: [max] Cannot max reduce zero size array.
>>> np.max(np.zeros((0, 2)), axis=-1)
array([], dtype=float64)
```

The reduction is well defined here. The array is empty because of axis 0, which is not being reduced, so the answer is an empty result of shape `(0,)`. What actually makes `max` undefined is reducing over an empty axis, since it has no identity element, and numpy draws the line exactly there:

```python
>>> np.max(np.zeros((2, 0)), axis=-1)
ValueError: zero-size array to reduction operation maximum which has no identity
```

`sum` and `prod` already get this right in mlx, because they have identities and never had the blanket check. So `mx.sum(mx.zeros((0, 2)), axis=-1)` returns shape `(0,)` today while `mx.max` on the same input raises.

The guard now checks the reduced axes rather than the whole array, after `compute_reduce_shape` has normalized them:

```python
>>> [mx.max(mx.zeros(s), axis=a).shape
...  for s, a in [((0, 2), -1), ((2, 0), 0), ((3, 0, 2), -1), ((0, 3, 2), [1, 2])]]
[(0,), (0,), (3, 0), (0,)]
```

Those match numpy. Reducing an empty axis still raises, with a message that now names the axis:

```
[max] Cannot max reduce over axis 1 with size 0.
```

Scope note worth stating up front: I left `argmax` and `argmin` alone even though they carry the same blanket check. Every case this change newly allows has `in.size() == out.size() == 0`, so `Reduce::eval_gpu` takes the existing `out.size() == in.size()` identity-copy branch on both Metal and CUDA before any kernel is dispatched, and both copy paths already return early on an empty output. `ArgReduce::eval_gpu` has no such branch and would dispatch a grid built from an empty output shape. I only have a CPU box, so rather than guess at the GPU behaviour I kept the arg reductions as they are. Happy to extend it if you would prefer them consistent.

## Checklist

CPU only build (`MLX_BUILD_METAL=OFF`) at 8d66629. `test_reduce.py`, `test_ops.py`, `test_linalg.py` and `test_compile.py` all pass; the new shape assertions raise `ValueError` on main. No test asserted the old message text.

---

freshman contributor, i use Claude Code while working through these. i got here from the other direction: i was checking which ops disagree with numpy on zero-size input and kept landing on the same reduction check rather than on the ops themselves, so this seemed like the place to fix it once.

### DIFF
--- a/mlx/ops.cpp
+++ b/mlx/ops.cpp
@@ -22,9 +22,14 @@ namespace mlx::core {
 
 namespace {
 
+// Pass the reduction's name in `op_without_identity` when it has no identity
+// element, as max and min do not. Those are undefined over an empty axis, so
+// naming them here has that checked. Reductions with an identity, such as sum
+// and any, reduce an empty axis fine and leave it unset.
 std::tuple<Shape, std::vector<int>, bool> compute_reduce_shape(
     const std::vector<int>& axes,
-    const Shape& shape) {
+    const Shape& shape,
+    std::string_view op_without_identity = {}) {
   bool is_noop = true;
   std::set<int> axes_set;
   auto ndim = shape.size();
@@ -57,24 +62,17 @@ std::tuple<Shape, std::vector<int>, bool> compute_reduce_shape(
   if (is_noop && !sorted_axes.empty() && detail::in_dynamic_tracing()) {
     is_noop = false;
   }
-  return {out_shape, sorted_axes, is_noop};
-}
-
-// max and min have no identity element, so they are only undefined when an
-// axis being reduced over is itself empty. An array that is empty because of
-// some other axis still reduces fine, into an empty result.
-void check_reduce_axes_not_empty(
-    const array& a,
-    const std::vector<int>& sorted_axes,
-    std::string_view op) {
-  for (auto ax : sorted_axes) {
-    if (a.shape(ax) == 0) {
-      std::ostringstream msg;
-      msg << "[" << op << "] Cannot " << op << " reduce over axis " << ax
-          << " with size 0.";
-      throw std::invalid_argument(msg.str());
+  if (!op_without_identity.empty()) {
+    for (auto ax : sorted_axes) {
+      if (shape[ax] == 0) {
+        std::ostringstream msg;
+        msg << "[" << op_without_identity << "] Cannot " << op_without_identity
+            << " reduce over axis " << ax << " with size 0.";
+        throw std::invalid_argument(msg.str());
+      }
     }
   }
+  return {out_shape, sorted_axes, is_noop};
 }
 
 Dtype at_least_float(const Dtype& d) {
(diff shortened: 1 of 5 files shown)


zcbenz (commented):
Maybe just add a check in compute_reduce_shape?

ayaangazali:
Done, moved it in. It cannot be unconditional there though, since `compute_reduce_shape` is shared with `all`, `any`, `sum` and `prod`, and those have identities so they reduce an empty axis fine:

```python
>>> [f(mx.zeros((2, 0)), axis=-1).tolist() for f in (mx.sum, mx.prod, mx.any, mx.all)]
[[0.0, 0.0], [1.0, 1.0], [False, False], [True, True]]
```

which is what numpy gives too. So it is an opt-in parameter that the caller sets only for a reduction with no identity:

```cpp
compute_reduce_shape(axes, a.shape(), "max")
```

That drops the separate helper and puts the check right next to where the axes get normalized, so it reads off `sorted_axes` directly instead of the caller doing it. Rebased on main, tests still pass.

mlx/ops.cpp
zcbenz, line 52: would it be possible to check here? also should we just check for all ops that calls the function as we usually don't want to have 0 in output shape?

zcbenz: ah sorry just noticed that your previous comment already answered my second question.

ayaangazali:
Yes, moved it into that loop. It is the branch that already identifies a reduced axis, so the check sits there and the extra pass over `sorted_axes` is gone:

```cpp
} else {
  if (!op_without_identity.empty() && shape[i] == 0) {
    ...throw...
  }
  out_shape.push_back(1);
}
```

Same behaviour, and `sum`, `prod`, `any` and `all` still reduce an empty axis since they leave the parameter unset. Rebased and tests pass.

zcbenz (approved):
Thanks!

ayaangazali:
CI caught a real gap in this change on the GPU backends, so I pushed a follow-up commit.

Allowing `(0, 2)` reduced over the last axis means the output is empty *and* a different shape than the input. Both GPU `Reduce::eval_gpu` implementations short-circuit to a cast-copy when `out.size() == in.size()`, which is also true at zero, and `copy_gpu_inplace` asserts `in.shape() == out.shape()` — so `test_zero_size` aborted:

```
Assertion failed: (in.shape() == out.shape()), function copy_gpu_inplace, file copy.cpp, line 20.
```

That is why the CUDA x86_64 and macOS jobs failed while the CPU jobs passed.

The fix excludes an empty input from that fast path so it falls through to the empty-reduction handling that both backends already have (CUDA's `if (in.size() == 0)` branch was unreachable for exactly this reason). It also returns early from `init_reduce` when the output is empty, since the dispatch would otherwise compute a zero-sized threadgroup on Metal and a zero-sized block on CUDA.

One caveat worth stating: I could not build the Metal path locally (no Xcode on this machine, so no `metal` compiler), so I am relying on CI to confirm both backends.

PR #4086, opened 2026-08-09
Fix Device and Stream lexicographic ordering

### LINKED ISSUE #4083: [BUG] Device::operator< and Stream::operator< are not strict weak orderings
**Describe the bug**

`Device::operator<` (`mlx/device.h`) and `Stream::operator<` (`mlx/stream.h`) are not strict weak orderings:

```cpp
bool operator<(const Device& rhs) const {
  return type < rhs.type || index < rhs.index;
}

bool operator<(const Stream& rhs) const {
  return device < rhs.device || index < rhs.index;
}
```

The second clause is evaluated even when the first field differs, so both `a < b` and `b < a` can be true. It needs to be a lexicographic compare (the second field should only be consulted when the first is *equal*).

This matters because `eval_impl` uses the ordering: `mlx/transforms.cpp` keys `std::set<Stream> open_streams` on the whole `Stream` (it was `std::unordered_set<int>` on `stream.index` before 0.31.2), so every `eval()` runs the comparator. A `std::set` with an invalid comparator can hold duplicate streams and fail to find ones it holds, and the standard places these calls outside the domain of the algorithm.

**To Reproduce**

C++ rather than Python, since these are C++ operators:

```cpp
#include "mlx/device.h"
#include "mlx/stream.h"
#include <cstdio>

using namespace mlx::core;

int main() {
  Device cpu1(Device::cpu, 1);
  Device gpu0(Device::gpu, 0);
  printf("Device{cpu,1} < Device{gpu,0} : %d\n", cpu1 < gpu0);
  printf("Device{gpu,0} < Device{cpu,1} : %d\n", gpu0 < cpu1);

  Stream s_cpu1(1, Device(Device::cpu, 0));
  Stream s_gpu0(0, Device(Device::gpu, 0));
  printf("Stream{idx1,cpu} < Stream{idx0,gpu} : %d\n", s_cpu1 < s_gpu0);
  printf("Stream{idx0,gpu} < Stream{idx1,cpu} : %d\n", s_gpu0 < s_cpu1);
}
```

```
$ c++ -std=c++20 -I <mlx> repro.cpp -o repro && ./repro
Device{cpu,1} < Device{gpu,0} : 1
Device{gpu,0} < Device{cpu,1} : 1
Stream{idx1,cpu} < Stream{idx0,gpu} : 1
Stream{idx0,gpu} < Stream{idx1,cpu} : 1
```

Both directions are `true` in both cases. Generally: any pair where one field compares less and the other compares greater.

It also aborts at runtime. libc++ validates comparators under `_LIBCPP_HARDENING_MODE_DEBUG`, which the Xcode 27 SDK enables, so any `eval()` with two such streams open dies:

```
__utility/lazy_synth_three_way_comparator.h:47: libc++ Hardening assertion
  __result ? !static_cast<bool>(__comp_(__rhs_, __lhs_)) : true failed:
  Comparator does not induce a strict weak ordering
```

```
frame #3: std::__1::__tree<mlx::core::Stream, std::__1::less<mlx::core::Stream>, ...>::__emplace_unique<...>
frame #7: std::__1::set<mlx::core::Stream, ...>::insert
frame #8: mlx::core::eval_impl(...) at transforms.cpp:227
frame #9: mlx::core::eval(...) at transforms.cpp:344
```

In mlx-swift this reproduces by running an op on an explicit `cpu` stream while the default device is `gpu` (two streams, `Stream{0, gpu}` and `Stream{1, cpu}`) and calling `eval`.

**Expected behavior**

`a < b` and `b < a` are never both true, and `std::set<Stream>` / `std::set<Device>` behave correctly. A lexicographic compare:

```cpp
bool operator<(const Device& rhs) const {
  if (type != rhs.type) {
    return type < rhs.type;
  }
  return index < rhs.index;
}

bool operator<(const Stream& rhs) const {
  if (!(device == rhs.device)) {
    return device < rhs.device;
  }
  return index < rhs.index;
}
```

Happy to send a PR if useful.

 - OS Version: macOS 27.0 (Apple Silicon), Xcode 27.0 SDK
 - Version: introduced in 0.31.2 (both operators are new in that release), still present in 0.32.0 and on `main` (8056817bd)

**Additional context**

The `// TODO: Use default three-way comparison when it gets supported in XCode.` comment above both operators suggests these were hand-written stand-ins for `<=>`; a defaulted `operator<=>` would have produced the lexicographic order.

wgu9:
## Proposed changes

Fixes #4083.

This makes `Device` and `Stream` comparisons lexicographic, so they satisfy strict weak ordering when the first and second fields sort in opposite directions.

I also added regression coverage for asymmetry, transitivity, and `std::set` ordering for both types.

## Testing

- `pre-commit run --files mlx/device.h mlx/stream.h tests/device_tests.cpp`
- `cmake -S . -B build -DMLX_BUILD_METAL=OFF -DMLX_BUILD_EXAMPLES=OFF -DMLX_BUILD_PYTHON_BINDINGS=OFF`
- `cmake --build build --target tests -j 8`
- `ctest --test-dir build --output-on-failure -j 1` (248/248 passed)

## Checklist

### DIFF
--- a/mlx/device.h
+++ b/mlx/device.h
@@ -5,6 +5,7 @@
 #include "mlx/api.h"
 
 #include <string>
+#include <tuple>
 #include <unordered_map>
 #include <variant>
 
@@ -27,7 +28,7 @@ struct MLX_API Device {
   // TODO: Use default three-way comparison when it gets supported in XCode.
   bool operator==(const Device&) const = default;
   bool operator<(const Device& rhs) const {
-    return type < rhs.type || index < rhs.index;
+    return std::tie(type, index) < std::tie(rhs.type, rhs.index);
   }
 };
 
--- a/mlx/stream.h
+++ b/mlx/stream.h
@@ -2,6 +2,7 @@
 
 #pragma once
 
+#include <tuple>
 #include <vector>
 
 #include "mlx/api.h"
@@ -17,7 +18,7 @@ struct MLX_API Stream {
   // TODO: Use default three-way comparison when it gets supported in XCode.
   bool operator==(const Stream&) const = default;
   bool operator<(const Stream& rhs) const {
-    return device < rhs.device || index < rhs.index;
+    return std::tie(device, index) < std::tie(rhs.device, rhs.index);
   }
 };
 
--- a/tests/device_tests.cpp
+++ b/tests/device_tests.cpp
@@ -2,11 +2,56 @@
 
 #include "doctest/doctest.h"
 
+#include <algorithm>
+#include <array>
 #include <cstdlib>
+#include <set>
 
 #include "mlx/mlx.h"
 
 using namespace mlx::core;
+
+template <typename T, size_t N>
+void check_strict_weak_ordering(const std::array<T, N>& ordered_values) {
+  for (const auto& value : ordered_values) {
+    CHECK_FALSE(value < value);
+  }
+
+  for (const auto& lhs : ordered_values) {
+    for (const auto& rhs : ordered_values) {
+      if (lhs < rhs) {
+        CHECK_FALSE(rhs < lhs);
+      }
+      for (const auto& other : ordered_values) {
+        if (lhs < rhs && rhs < other) {
+          CHECK(lhs < other);
+        }
+      }
+    }
+  }
+
+  std::set<T> values(ordered_values.begin(), ordered_values.end());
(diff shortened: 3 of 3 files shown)


zcbenz (approved):
This is a very beautiful fix, thanks!

PR #4090, opened 2026-08-09
Fix any and all treating -0.0 as nonzero

### LINKED ISSUE #4075: [BUG]  `mx.any` returns `True` for all-zero arrays that mix `+0.0` and `-0.0`
**Describe the bug**
`mx.any` is supposed to be `True` iff at least one element is nonzero. But for
an array of all zeros that contains both `+0.0` and `-0.0`, it returns `True`

**To Reproduce**

Include code snippet
```python
import mlx.core as mx
import numpy as np
a = mx.array(np.array([0.0, -0.0], dtype=np.float32))
np.array(mx.any(a)).item()   # True   WRONG (should be False)
np.any(np.array([0.0, -0.0]))   # False
```

**Expected behavior**
```
mx.any on all-zero arrays that mix +0.0 and -0.0 (must be False):
  any([0.0, -0.0]):      mlx=True  np=False  <- BUG
  any([0.0, -0.0, 0.0]): mlx=True  np=False  <- BUG
  any([-0.0, 0.0, -0.0]):mlx=True  np=False  <- BUG
  all([0,-0]): mlx=False (correct=False)
  count_nonzero([0,-0]): mlx=0 (correct=0)
```

ayaangazali:
## Proposed changes

Fixes #4075. `mx.any` reports a nonzero element in an array that is all zeros, as long as one of them is `-0.0`:

```python
>>> a = mx.array(np.array([0.0, -0.0], dtype=np.float32))
>>> mx.any(a).item()
True
>>> np.any(np.array([0.0, -0.0]))
False
```

`mx.all` is wrong the same way, which the issue did not mention:

```python
>>> mx.all(mx.array(np.array([-0.0, -0.0], dtype=np.float32))).item()
True     # every element is zero, so this should be False
```

The cause is in the `And` / `Or` dispatch in `mlx/backend/cpu/reduce.cpp`, which picks the reduction type by width rather than by value:

```cpp
case uint32:
case int32:
case float32:
  reduce_dispatch_and_or<int32_t>(in, out, reduce_type_, axes_);
```

Reading the floats as integers is fine for deciding truthiness of every value except `-0.0`, whose bit pattern is `0x80000000`. It is zero, but not zero *bits*, so the reduction counts it as set. That is also why a single `-0.0` looks correct: a one element reduction is a no-op and goes through `astype` instead.

Floats now dispatch as themselves so `y != 0` is a real float comparison. Integers keep sharing a type by width, since only their bits matter.

```python
>>> [mx.any(mx.array(np.array(v, dtype=np.float32))).item()
...  for v in ([0.0, -0.0], [-0.0] * 70, [-0.0, 0.0, 1.0])]
[False, False, True]
```

This costs nothing at runtime. `simd::max_size<bool>` is 1, so `N` in `contiguous_reduce` was already 1 for these reductions and the wider integer type was never vectorized over; it only shared template instantiations.

`complex64` deliberately stays on the integer path, with a comment saying why. Testing it as a complex would go through `complex64_t`'s conversion to `float`, which returns only the real part, so `mx.any` would stop seeing `1j`. The integer path gets that right.

## Checklist

CPU only build (`MLX_BUILD_METAL=OFF`) at 6a0dd0f. `test_reduce.py`, `test_ops.py`, `test_array.py` and `test_linalg.py` pass, and the new test fails on main. I checked float32, float16, float64 and bfloat16 against numpy over signed zeros, subnormals, nan, inf and ordinary values, and confirmed int8 through uint64 and bool are unchanged. Metal's `And` and `Or` take a `bool` that is converted before the reduction, so there was nothing matching to fix there.

---

freshman contributor, i lean on Claude Code while digging. worth admitting the detour: i first blamed the simd overloads in `OrReduce`, whose parameters are named the opposite way round from the scalar one, and only after rewriting them and seeing the output not budge did i find the real cause a layer up in the dtype switch. the parameter naming is confusing but harmless, so i left it alone.

### DIFF
--- a/mlx/backend/cpu/reduce.cpp
+++ b/mlx/backend/cpu/reduce.cpp
@@ -460,6 +460,9 @@ void Reduce::eval_cpu(const std::vector<array>& inputs, array& out) {
     switch (reduce_type_) {
       case Reduce::And:
       case Reduce::Or: {
+        // Integers can be reduced as whatever type has the same width, since
+        // only their bits matter. Floats cannot: -0.0 compares equal to zero
+        // but has a bit set, so it has to be tested as a float.
         switch (in.dtype()) {
           case bool_:
           case uint8:
@@ -468,20 +471,31 @@ void Reduce::eval_cpu(const std::vector<array>& inputs, array& out) {
             break;
           case int16:
           case uint16:
-          case float16:
-          case bfloat16:
             reduce_dispatch_and_or<int16_t>(in, out, reduce_type_, axes_);
+            break;
+          case float16:
+            reduce_dispatch_and_or<float16_t>(in, out, reduce_type_, axes_);
+            break;
+          case bfloat16:
+            reduce_dispatch_and_or<bfloat16_t>(in, out, reduce_type_, axes_);
             break;
           case uint32:
           case int32:
-          case float32:
             reduce_dispatch_and_or<int32_t>(in, out, reduce_type_, axes_);
+            break;
+          case float32:
+            reduce_dispatch_and_or<float>(in, out, reduce_type_, axes_);
             break;
           case uint64:
           case int64:
-          case float64:
+          // complex64 stays on the integer path. Testing it as a complex
+          // would go through complex64_t's conversion to float and only look
+          // at the real part, which would miss 1j.
           case complex64:
             reduce_dispatch_and_or<int64_t>(in, out, reduce_type_, axes_);
+            break;
+          case float64:
+            reduce_dispatch_and_or<double>(in, out, reduce_type_, axes_);
             break;
         }
         break;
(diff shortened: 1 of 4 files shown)


ayaangazali:
The macOS job caught that this only fixed the CPU backend — pushed a follow-up for Metal.

`remap_reduce_types` picks the and/or input type by width, so a float was tested as the integer of the same size and `-0.0` (sign bit set, value zero) read as nonzero:

```
FAIL: test_and_or_negative_zero (dtype='float32')
AssertionError: True != False : any float32 [0.0, -0.0]
```

Same root cause as the CPU path, one layer down. `float16`, `bfloat16` and `float32` now map to themselves and the and/or kernels are instantiated for them, so the value is what gets tested. `complex64` stays on the integer path for the reason already in the CPU comment (testing it as a complex would only look at the real part and miss `1j`), and `float64` is not a Metal type.

Note I could not build the Metal path locally — no Xcode on this machine, so no `metal` compiler for the shader instantiations. The C++ side compiles clean under `-Werror`; the new kernel instantiations mirror the existing `sum_prod` and `min_max` lines, but CI is the real check on those.

zcbenz (approved):
Nice fix, thanks!

PR #4094, opened 2026-08-09
Enable complex64 take_along_axis backward on Metal

PhysicistJohn:
## Summary

Enable the existing Metal `ScatterAxis::Sum` path for `complex64`. This supports `take_along_axis` backward when duplicate indices require additive accumulation.

The eight-byte restriction is relaxed only for `complex64` in sum mode, reusing the componentwise atomic addition merged in #4078. Assignment and other eight-byte outputs remain unchanged. CUDA already supports this path; there is no public API or Metal-kernel change.

## Testing

- Complex `take_along_axis` VJP with duplicate indices under Metal API validation
- Full autograd module on CPU and Metal: 52 passed, 1 environment-dependent Torch test skipped on each
- Existing `take_along_axis` and `put_along_axis` controls on CPU and Metal
- Inherited #4078 complex scatter-add regression and unsupported assignment guard
- Release C++20 warnings-as-errors build, all eight JIT layout/location compile-link variants, and formatting checks

### DIFF
--- a/python/tests/test_autograd.py
+++ b/python/tests/test_autograd.py
@@ -519,24 +519,22 @@ class TestAutograd(mlx_tests.MLXTestCase):
         grad = mx.grad(scatter_axis_fun)(mx.ones((3, 3)))
         self.assertTrue(mx.array_equal(grad, mx.ones((3, 3))))
 
-    @unittest.skipIf(not mx.metal.is_available(), "Metal is not available")
-    def test_take_along_axis_complex_vjp_metal(self):
-        with mx.stream(mx.gpu):
-            x = mx.zeros((4,), dtype=mx.complex64)
-            indices = mx.array([3, 3], dtype=mx.int32)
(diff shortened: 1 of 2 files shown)


python/tests/test_autograd.py
```diff
@@ -519,6 +519,25 @@ def scatter_axis_fun(w):
         grad = mx.grad(scatter_axis_fun)(mx.ones((3, 3)))
         self.assertTrue(mx.array_equal(grad, mx.ones((3, 3))))
 
+    @unittest.skipIf(not mx.metal.is_available(), "Metal is not available")
```
zcbenz: Since this works in other backends we should just run the test for all without restricting metal and gpu.

PhysicistJohn: Thanks, good call. I updated the test to use the default device and removed the Metal-only skip, so it now runs across backends. I also ran the full autograd test file with both DEVICE=cpu and DEVICE=gpu (53 tests on each, with the same existing skip).

zcbenz (approved):

PR #4098, opened 2026-08-09
Fix SinusoidalPositionalEncoding silently overriding explicit scale=0.0

### LINKED ISSUE #4097: [BUG] SinusoidalPositionalEncoding: explicit scale=0.0 silently overridden by default
**Describe the bug**
`nn.SinusoidalPositionalEncoding` resolves its `scale` argument with
`scale or (2 / dims) ** 0.5`. Since `0.0` is falsy in Python, passing
`scale=0.0` explicitly is indistinguishable from not passing `scale` at
all, so it gets silently replaced by the default `sqrt(2/dims)` instead
of being respected.

**To Reproduce**

```python
import mlx.core as mx
import mlx.nn as nn

layer = nn.SinusoidalPositionalEncoding(dims=64, scale=0.0)
print(layer.scale)  # 0.1767766952966369, not 0.0

x = mx.array([0.0, 1.0, 2.0])
print(layer(x)[1])  # nonzero, even though scale=0.0 was requested
```

**Expected behavior**
`layer.scale` should be exactly `0.0`, and `layer(x)` should return the
zero vector for every position, since `scale=0.0` is a valid explicit
float (e.g. to fully disable the positional signal for an ablation).

 - OS Version: macOS 26.2
 - Version: 0.32.0 (also present on main, in
   `python/mlx/nn/layers/positional_encoding.py`)

**Additional context**
Root cause:
```python
self.scale = scale or (2 / dims) ** 0.5
```
should check `scale is not None` rather than truthiness. I have a fix
+ regression test ready and will open a PR referencing this issue.

YassineMA03:
Fixes #4097 ...

### DIFF
--- a/python/mlx/nn/layers/positional_encoding.py
+++ b/python/mlx/nn/layers/positional_encoding.py
@@ -107,7 +107,7 @@ class SinusoidalPositionalEncoding(Module):
             self._sigmas = self._sigmas * (2 * math.pi)
 
         # Save some constants that define the implementation
-        self.scale = scale or (2 / dims) ** 0.5
+        self.scale = scale if scale is not None else (2 / dims) ** 0.5
         self.cos_first = cos_first
 
     def __call__(self, x):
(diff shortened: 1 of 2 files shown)


zcbenz (approved):
Thanks!

PR #4100, opened 2026-08-09
Stop integer power from hanging on a negative exponent

ayaangazali:
## Proposed changes

Raising an integer to a negative power hangs the CPU backend:

```python
>>> x = mx.power(mx.array([2]), mx.array([-1]), stream=mx.cpu)
>>> mx.eval(x)      # never returns
```

The integer branch of `pow` in `mlx/backend/cpu/simd/base_simd.h` is exponentiation by squaring:

```cpp
T res = 1;
while (exp) {
  if (exp & 1) {
    res *= base;
  }
  exp >>= 1;
  base *= base;
}
```

`exp >>= 1` on a negative signed value is an arithmetic shift, so `-1 >> 1` is `-1` and the loop never reaches zero. Any negative exponent spins, for every signed integer width.

Both GPU backends already handle this and settled on returning 0, so this is the CPU falling out of step rather than an open question:

```cpp
// mlx/backend/metal/kernels/binary_ops.h
// Undefined to raise integer to negative power
if (exp < 0) { ...; return 0; }

// mlx/backend/cuda/device/binary_ops.cuh
// Raising an integer to a negative power is undefined
if constexpr (cuda::std::is_signed_v<T>) {
  if (exp < 0) { return 0; }
}
```

Added the same guard to the CPU path, kept behind `if constexpr (std::is_signed_v<T>)` so the unsigned instantiations do not get an always-false comparison under `-DCMAKE_COMPILE_WARNING_AS_ERROR=ON`.

```python
>>> mx.power(mx.array([2, -2, 1]), mx.array([-1, -3, -9])).tolist()
[0, 0, 0]
```

`test_integer_power` already had a case for this, added with the comment `# Doesn't hang`:

```python
# Doesn't hang
x = mx.power(2, -1)
```

It never hung because nothing evaluates `x`, so the graph was built and dropped. The eval is now there, which is what makes it fail on main, along with the same check across the four signed widths.

## Checklist

CPU only build (`MLX_BUILD_METAL=OFF`) at 8c28c38, `test_ops.py` 149 tests pass. Checked that unsigned integer powers, positive exponents, a zero exponent and float powers are all unchanged. Worth flagging that on main the new assertions hang rather than fail, since that is the bug.

---

freshman contributor, i use Claude Code while working through these. found it by running ops against numpy for odd inputs and having the whole script time out with no output, which took a couple of passes to narrow down to one line. numpy raises for this instead of returning 0, but i matched what Metal and CUDA already do here rather than introduce a third behaviour.


.claude/settings.local.json
zcbenz, line 1: Can you remove this file?

ayaangazali: Removed, sorry about that. It was an editor settings file my `git add -A` swept in. The branch is now just `base_simd.h` and `test_ops.py`, and I have excluded the directory locally so it cannot happen again.

zcbenz (approved):
Nice catch!

Issue #3979, opened 2026-08-02
conv2d on Metal silently returns all-zero output for large inputs (channel-count dependent, ~2e9 elements)


kayadibi1:
## Summary

`mlx.nn.Conv2d` / `mx.conv2d` on Metal **silently returns an all-zero tensor** for
large batched inputs. No exception, no warning — the op completes and produces
zeros. The failure depends on both total size and channel count, which suggests a
size-limited kernel variant selected for certain channel configurations.

This is distinct from (and much more dangerous than) the loud
`metal::malloc ... maximum allowed buffer size` error for >30 GB buffers: that one
fails the run; this one corrupts results silently.

## Environment

- mlx **0.32.0** (via pip)
- Apple **M5 Pro, 48 GB**, macOS **27.0** (build 26A5378n)
- Python 3.14.6

## Minimal repro

```python
import mlx.core as mx
import mlx.nn as nn
import numpy as np

def make(BT, H, W, C):  # deterministic non-constant input
    bt = (mx.arange(BT).reshape(BT, 1, 1, 1) % 13).astype(mx.float32) * 0.31
    h  = (mx.arange(H).reshape(1, H, 1, 1) % 7).astype(mx.float32) * 0.17
    w  = (mx.arange(W).reshape(1, 1, W, 1) % 5).astype(mx.float32) * 0.11
    c  = (mx.arange(C).reshape(1, 1, 1, C) % 3).astype(mx.float32) * 0.07
    x = mx.broadcast_to(bt + h + w + c - 1.0, (BT, H, W, C)).astype(mx.float32)
    mx.eval(x)
    return x

def test(BT, H, W, C):
    x = make(BT, H, W, C)
    conv = nn.Conv2d(C, C // 2, 3, padding=1)
    mx.eval(conv.parameters())
    y = conv(x); mx.eval(y)
    for b in (0, BT - 1):
        got  = np.array(y[b, :4, :4])
        want = np.array(conv(x[b : b + 1])[0, :4, :4])   # same op, single slice
        d = float(np.abs(got - want).max())
        print(f"BT={BT} C={C} elems={x.size:,}: batch {b} "
              f"maxdiff={d:.4f} out_std={got.std():.4f}"
              f"{'   <-- ALL ZEROS' if got.std() < 1e-6 else ''}")

test(32, 512, 512, 192)   # 1.61e9 elements -> correct
test(40, 512, 512, 192)   # 2.01e9 elements -> ALL ZEROS, silently
[...]

PR #4102, opened 2026-08-09
Bound winograd conv2d working set by tiling the batch

Gusanidas:
Addresses the conv2d manifestation of #3979 — the conv-side guard, not a fix for the general silent-error mechanism.

## Problem

`conv2d` on Metal silently returns all zeros for large batched inputs. The Winograd path keeps the padded input and both GEMM workspaces alive at once (~3.7x the input for the reported shapes), all referenced by one command buffer; once that working set exceeds `recommendedMaxWorkingSetSize` the driver fails the command buffer. As [@jasonge27's instrumentation](https://github.com/ml-explore/mlx/issues/3979#issuecomment-5231678147) shows, Metal reports the OOM but errors from mid-eval commits are dropped, so nothing surfaces. Winograd's selection criteria (`C % 32 == 0 && O % 32 == 0 && C + O >= 256`) explain the channel dependence in the report. That dropped-error mechanism is general and needs a separate, complementary fix in the eval machinery; this PR avoids the OOM for conv2d, which also bounds its peak memory.

## Fix

Run the batch in tiles sized from the available working set, reusing one set of scratch buffers across tiles — mirroring the row tiling in `explicit_gemm_conv_ND_gpu`. The budget is 3/4 of `recommendedMaxWorkingSetSize` minus the conv's own tensors (input, output, and transformed filter), making the tile choice a deterministic function of the conv and the device. Measured failures start at ~99% of the reported limit; the quarter held back is headroom for what is not charged — GPU memory held by other processes and allocations unrelated to the conv.

When the whole batch fits, the common case, the emitted work is unchanged. It falls back to the scratch-free implicit GEMM when not even one batch element fits, or when tiles would carry too few GEMM rows to amortize their fixed cost (small tiles can be far slower than the fallback).

## Testing

The real threshold needs tens of GB, so tests shrink the budget instead:

- `MLX_CONV_WINOGRAD_WORKING_SET` injects the byte budget: covers automatic tiling and both fallbacks.
- `MLX_CONV_WINOGRAD_TILE_BATCH` forces a tile size (capped by the budget): covers uniform tiles, a short final tile, and a consumer op reading the tiled output in the same eval.

Tiling keeps the per-element reduction order, so tiled Winograd is bit-identical to untiled while the implicit GEMM fallback never is — the tests use exact equality to pin each run to its intended path, with no memory measurements.

At real sizes on a 96 GB M2 Max: the issue's repro passes (previously all-zero at batch 80 on this machine; the reporter hit it at batch 40 on 48 GB), tiled vs untiled outputs are bit-identical at ~50 GB scale, CPU cross-checks agree, and the batch-80 repro peaks at 58 GB where untiled Winograd would need ~84 GB. On the reported fp32 shape (40x512x512, 192 -> 96 channels) Winograd runs ~1.7x faster than implicit GEMM (310 ms vs 536 ms), while a spatially tiny conv (256x4x4, 64 -> 192) cut into one-element Winograd tiles is ~40x slower than implicit GEMM — the two sides of the tile-versus-fallback trade-off. Mid-size Winograd shapes select a single tile and show no regression.

### DIFF
--- a/mlx/backend/metal/conv.cpp
+++ b/mlx/backend/metal/conv.cpp
@@ -1,10 +1,7 @@
 // Copyright © 2023-2024 Apple Inc.
 #include <algorithm>
 #include <cassert>
-#include <cerrno>
-#include <cstdlib>
 #include <numeric>
-#include <stdexcept>
 
 #include "mlx/backend/gpu/copy.h"
 #include "mlx/backend/gpu/slicing.h"
@@ -15,7 +12,6 @@
 #include "mlx/backend/metal/kernels/steel/conv/params.h"
 #include "mlx/backend/metal/matmul.h"
 #include "mlx/backend/metal/utils.h"
-#include "mlx/memory.h"
 #include "mlx/primitives.h"
 #include "mlx/utils.h"
 
@@ -53,10 +49,9 @@ inline int max_unfold_rows(metal::Device& d, size_t row_bytes, int total_rows) {
 }
 
 // Winograd's scratch is a multiple of the input, all referenced by one
-// command buffer; past the GPU's working set limit the kernels silently
-// produce zeros. Run the batch in tiles to bound it. Returns the batch
-// elements per tile, or 0 if not even one fits (the caller must then pick a
-// path that needs no scratch).
+// command buffer, which fails past the GPU's working set limit. Run the
+// batch in tiles to bound it. Returns the batch elements per tile, or 0 if
+// not even one fits.
 inline int winograd_batch_step(
     metal::Device& d,
     const array& in,
@@ -64,17 +59,18 @@ inline int winograd_batch_step(
   int total_n = conv_params.N;
 
   size_t itemsize = in.itemsize();
-  int padded_h =
-      6 * ((conv_params.iS[0] + 2 * conv_params.pad[0] - 2 + 5) / 6) + 2;
-  int padded_w =
-      6 * ((conv_params.iS[1] + 2 * conv_params.pad[1] - 2 + 5) / 6) + 2;
+  int64_t padded_h = static_cast<int64_t>(conv_params.iS[0]) +
+      2 * static_cast<int64_t>(conv_params.pad[0]);
+  int64_t padded_w = static_cast<int64_t>(conv_params.iS[1]) +
+      2 * static_cast<int64_t>(conv_params.pad[1]);
+  padded_h = 6 * ((padded_h - 2 + 5) / 6) + 2;
+  padded_w = 6 * ((padded_w - 2 + 5) / 6) + 2;
   int tiles_per_n =
       ((conv_params.oS[0] + 5) / 6) * ((conv_params.oS[1] + 5) / 6);
 
   // The padded input and both gemm workspaces scale with the tile; the
-  // transformed filter does not. That is all of the scratch: the batched gemm
-  // allocates nothing here (split-k needs a batch size of one, winograd always
-  // runs 8 * 8, and its operands are contiguous so it never copies one).
+  // transformed filter does not. That is all of the scratch: the batched
+  // gemm allocates nothing here.
   size_t filt_bytes =
       static_cast<size_t>(8 * 8) * conv_params.C * conv_params.O * itemsize;
   size_t bytes_per_n =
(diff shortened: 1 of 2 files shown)


erwinzhang7:
Verified on an **M4 Pro Mac mini with 48 GB**, macOS 26.5 — the reporter's
configuration in #3979, which you note you could not test at 96 GB.

Control is `8c28c385`, this PR's own parent, since the branch is cut from an
older main. Same sweep on both, 512x512, 192 -> 96, fp32, checked against a CPU
convolution of the same input:

```
control 8c28c385                      this PR 413078a0
 batch   peak GB   max err  result     batch   peak GB   max err  result
    32      33.7  1.14e-05  ok            32      30.0  1.18e-05  ok
    40      42.1      3.34  ALL ZERO      40      30.1  1.02e-05  ok
    48      50.6      5.64  ALL ZERO      48      29.5  1.69e-05  ok
    64      67.4      5.18  ALL ZERO      64      29.9  2.03e-05  ok
```

**Failure starts at batch 40**, the number @kayadibi1 reported on 48 GB. Peak
climbs unbounded on the control, to 67.4 GB asked for on a 48 GB machine. With
the patch it plateaus around 30 GB from batch 32 and stays correct through 64.

Also on an M5 Max via the injected budget: at 512 MB peak drops 1.10 -> 0.36 GB
and the output is **bit-identical** to untiled, `np.array_equal` on the uint32
view of 25,165,824 elements. At 1 MB it falls back to implicit GEMM with a max
diff of 0.0398 against untiled winograd, that is the two algorithms differing,
not the tiling. `test_conv`
passes.

No opinion on the tile-versus-fallback heuristic, which I have not stressed
beyond these shapes.

mlx/backend/metal/conv.cpp
```diff
@@ -47,6 +51,86 @@ inline int max_unfold_rows(metal::Device& d, size_t row_bytes, int total_rows) {
...
+// path that needs no scratch).
+inline int winograd_batch_step(
+    metal::Device& d,
+    const array& in,
+    const MLXConvParams<2>& conv_params) {
+  int total_n = conv_params.N;
+
+  size_t itemsize = in.itemsize();
+  int padded_h =
+      6 * ((conv_params.iS[0] + 2 * conv_params.pad[0] - 2 + 5) / 6) + 2;
```
zcbenz: Can you rebase and fix the conflict with https://github.com/ml-explore/mlx/pull/4258/changes#diff-d89046a37f9b2d939417b57314e679e1d9aa2f64a00b0a45b7e4d9656a75056dR903?

mlx/backend/metal/conv.cpp
```diff
@@ -47,6 +51,86 @@ inline int max_unfold_rows(metal::Device& d, size_t row_bytes, int total_rows) {
...
+      static_cast<size_t>(8 * 8) * conv_params.C * conv_params.O * itemsize;
+  size_t bytes_per_n =
+      static_cast<size_t>(padded_h) * padded_w * conv_params.C * itemsize +
+      static_cast<size_t>(8 * 8) * tiles_per_n *
+          (conv_params.C + conv_params.O) * itemsize;
+
+  size_t working_set = d.mtl_device()->recommendedMaxWorkingSetSize();
+  // Test hook: pretend the GPU can only keep this many bytes resident, so the
+  // budget arithmetic below can be exercised at sizes that fit in CI.
+  if (auto env_ws = env::get_var("MLX_CONV_WINOGRAD_WORKING_SET", "");
```
zcbenz: I think you can just do

```suggestion
  if (int env_ws = env::get_var("MLX_CONV_WINOGRAD_WORKING_SET", 0);
```

mlx/backend/metal/conv.cpp
```diff
@@ -47,6 +51,86 @@ inline int max_unfold_rows(metal::Device& d, size_t row_bytes, int total_rows) {
...
+  // Results went wrong at ~99% of the working set, so hold a quarter of it
+  // back: this process cannot see the GPU memory other processes hold, nor
+  // the allocator's cache of freed-but-unreturned buffers.
+  size_t limit = working_set / 4 * 3;
+
+  // Deliberately over-counts: only buffers this command buffer binds matter,
+  // but charging all live allocations only errs toward a smaller tile. The
+  // input, weights and output are already allocated and inside this figure,
+  // so only winograd's scratch is left to make room for.
+  size_t used = get_active_memory() + filt_bytes;
```
zcbenz: We don't dispatch kernels depending on current memory, it is not reliable and make behavior unpredictable.

python/tests/test_conv.py
```diff
@@ -1214,6 +1214,156 @@ def test_conv2d_unaligned_channels(self):
...
+
+        def run(x, w, **env):
+            # Returns the result plus the bytes the conv itself allocated.
+            for k in (tile_key, ws_key):
+                os.environ.pop(k, None)
+            os.environ.update(env)
+            # Get the previous run's temporaries off the books first.
+            mx.synchronize()
+            mx.reset_peak_memory()
+            base = mx.get_active_memory()
```
zcbenz: I would prefer dropping the memory measurements in tests, empirically they tend to be very flaky.

zcbenz (changes requested):

zcbenz (approved):
Nice work, thanks!

PR #4110, opened 2026-08-10
Fix intermittent crashes at thread and interpreter teardown

dhiltgen:
## Proposed changes

MLX sometimes dies at thread or process teardown with "terminate called without an active exception" or SIGSEGV, on Linux/Windows.  The general pattern: a destructor reachable from __call_tls_dtors() must not acquire the GIL, nor throw, nor drop a Python reference.

Acquiring the GIL is fatal because take_gil() answers a request from a non-finalizing thread, while another thread finalizes, by calling PyThread_exit_thread(); the forced unwind then crosses the noexcept destructor frame.

Throwing is equally fatal: CUDA destructors ran CHECK_*_ERROR on teardown calls that legitimately fail -- every Windows run with two threads died.

Adds python/tests/test_teardown.py. The failure kills the process, so the cases run in subprocesses and repeat; both fail without this change.

## Checklist

### DIFF
--- a/mlx/backend/cuda/cublas_utils.cpp
+++ b/mlx/backend/cuda/cublas_utils.cpp
@@ -45,8 +45,10 @@ auto& cublas_handles_cache() {
   struct CublasHandles {
     ~CublasHandles() {
       if (handle) {
-        CHECK_CUBLAS_ERROR(cublasLtDestroy(handle));
-        CHECK_CUBLAS_ERROR(cublasLtMatmulPreferenceDestroy(pref));
+        // Unchecked: destroy can fail once the library is unloading, and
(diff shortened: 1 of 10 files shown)


zcbenz (commented):
Our current recommended pattern is to let users call `clear_streams` at the end of threads and the process, so they can do cleanup before cuda and python shutdown, which is something that we could not do on the library side. Have you tried that? We should have a multi-threading doc for all these APIs though.

https://github.com/ml-explore/mlx/blob/e78d894c8f718a805341e1aa1cc835d9c9b8462e/tests/scheduler_tests.cpp#L89-L101

dhiltgen:
I can close this PR, these are just crashes that I was noticing while working on other improvements.

PR #4116, opened 2026-08-10
Fix vmap of partition/argpartition dropping the kth argument

### LINKED ISSUE #4113: vmap of partition/argpartition drops kth and flattens the array; topk fails under vmap
### Describe the bug

`mx.vmap` of `mx.partition` / `mx.argpartition` silently returns a **flattened** array partitioned around the **wrong element**, and `mx.vmap` of `mx.topk` raises an internal error.

`Partition::vmap` and `ArgPartition::vmap` in `mlx/primitives.cpp` call

```cpp
return {{partition(inputs[0], axis_ + axis_left, stream())}, axes};
```

There are two overloads:

```cpp
array partition(const array& a, int kth, StreamOrDevice s = {});            // flattens
array partition(const array& a, int kth, int axis, StreamOrDevice s = {});  // along axis
```

`partition(a, int, stream)` matches the **first** one, so `axis_ + axis_left` is passed as `kth` and the array is flattened. `kth_` is dropped entirely. `Sort::vmap` just below is correct only because `sort` has no `kth` parameter.

`topk` is implemented on top of `partition`, so it inherits the bug: the traced `slice` is then applied to a 1-D array and fails.

### To Reproduce

```python
import mlx.core as mx

a = mx.random.uniform(shape=(2, 3, 4))

out = mx.vmap(lambda x: mx.partition(x, 1, axis=-1))(a)
ref = mx.stack([mx.partition(x, 1, axis=-1) for x in a])
print(out.shape, ref.shape)   # (24,) (2, 3, 4)   <-- silently wrong

mx.vmap(lambda x: mx.topk(x, 2, axis=-1))(a)
# ValueError: [slice] Invalid number of indices or strides for array with dimension 1.
```

With `in_axes=1` the `topk` case fails differently:

```
IndexError: SmallVector out of range.
```

`mx.argpartition` behaves the same way. This reproduces for every `kth` / `axis` / `in_axes` combination I tried.

### Expected behavior

`vmap` should match the equivalent loop, i.e. `mx.vmap(f)(a)` should agree with `mx.stack([f(x) for x in a])` in both shape and values.

### Desktop

- OS: macOS 26.2
- Version: `main` @ 8c28c385 (also reproduces on 0.32.0)

### Additional context

There is currently no `vmap` test for `partition`, `argpartition` or `topk` in `python/tests/test_vmap.py`, which is likely why this went unnoticed.

I have a fix and a regression test ready and will link a PR.

Adityaj0:
Fixes #4113.

### Proposed changes

`Partition::vmap` and `ArgPartition::vmap` called the two-argument overload

```cpp
partition(inputs[0], axis_ + axis_left, stream())
```

which is the one that partitions the **flattened** array, with `axis_ + axis_left` bound to `kth`. `kth_` was dropped entirely, so under `vmap` these ops returned a 1-D array partitioned around the wrong element. `Sort::vmap` immediately below is correct only because `sort` has no `kth` parameter.

`topk` is built on `partition`, so it inherited the bug and failed outright once the traced `slice` was applied to the flattened result.

Passing `kth_` explicitly selects the `(a, kth, axis, stream)` overload.

Before, on `main`:

```
vmap(partition)    -> (24,)   expected (2, 3, 4)
vmap(argpartition) -> (24,)   expected (2, 3, 4)
vmap(topk)         -> ValueError: [slice] Invalid number of indices or strides
                                 for array with dimension 1.
vmap(topk, in_axes=1) -> IndexError: SmallVector out of range.
```

After:

```
vmap(partition)    -> (2, 3, 4), kth element matches the loop reference
vmap(argpartition) -> (2, 3, 4)
vmap(topk)         -> (2, 3, 2)
vmap(topk, in_axes=1) -> (2, 3, 2)
```

### Tests

Added `test_vmap_partition` and `test_vmap_topk` to `python/tests/test_vmap.py`; there was previously no `vmap` coverage for these ops. They sweep every `in_axes`, inner `axis` and every valid `kth` / `k`, comparing against the equivalent loop.

Because `partition` only pins the `kth` element and leaves the two sides in an arbitrary order, the tests compare the sorted result against the sorted input and check the `kth` element against the reference, rather than asserting element-wise equality that the API does not promise. `argpartition` is checked by gathering with the returned indices.

Verified with a CPU-only build (`-DMLX_BUILD_METAL=OFF`); `test_vmap`, `test_ops` and `test_autograd` pass.

### DIFF
--- a/mlx/primitives.cpp
+++ b/mlx/primitives.cpp
@@ -616,7 +616,7 @@ std::pair<std::vector<array>, std::vector<int>> ArgPartition::vmap(
   assert(axes.size() == 1);
 
   int axis_left = axes[0] >= 0 && axes[0] <= axis_;
-  return {{argpartition(inputs[0], axis_ + axis_left, stream())}, axes};
+  return {{argpartition(inputs[0], kth_, axis_ + axis_left, stream())}, axes};
 }
 
 std::vector<array> ArgPartition::vjp(
@@ -3417,7 +3417,7 @@ std::pair<std::vector<array>, std::vector<int>> Partition::vmap(
   assert(axes.size() == 1);
 
   int axis_left = axes[0] >= 0 && axes[0] <= axis_;
-  return {{partition(inputs[0], axis_ + axis_left, stream())}, axes};
+  return {{partition(inputs[0], kth_, axis_ + axis_left, stream())}, axes};
 }
 
 bool Partition::is_equivalent(const Primitive& other) const {
--- a/python/tests/test_vmap.py
+++ b/python/tests/test_vmap.py
@@ -252,6 +252,85 @@ class TestVmap(mlx_tests.MLXTestCase):
         expected = mx.array([2, 1])
         self.assertTrue(mx.array_equal(out, expected))
 
+    def _unstack(self, x, axis):
+        return [s.squeeze(axis) for s in mx.split(x, x.shape[axis], axis=axis)]
+
+    def test_vmap_partition(self):
+        # Distinct values so each lane has a single valid kth element
+        a = mx.random.permutation(2 * 3 * 4).reshape(2, 3, 4).astype(mx.float32)
+
+        for in_axis in (0, 1, 2):
+            slices = self._unstack(a, in_axis)
+            # Axis of the batched output that the inner axis maps onto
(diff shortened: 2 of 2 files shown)


zcbenz (approved):
Looks good to me, thanks!

PR #4117, opened 2026-08-10
Fix second order gradients for sort, partition, topk and cummax/cummin

### LINKED ISSUE #4114: Second order gradients fail for sort, partition, topk, cummax and cummin
### Describe the bug

Second order derivatives (Hessians, Hessian-vector products, anything of the form `mx.grad(mx.grad(...))`) fail for `mx.sort`, `mx.partition`, `mx.topk`, `mx.cummax` and `mx.cummin` with:

```
ValueError: [scatter_axis] Cannot calculate VJP with respect to indices.
Use stop_gradient on indices to stop gradients from being computed.
```

The first derivative works fine; only the second fails.

The cause is that these rules build an index array *from the primal* and then feed it to a gather/scatter. For example `Sort::vjp`:

```cpp
auto sort_idx = argsort(primals[0], axis_, stream());
return {put_along_axis(
    zeros_like(primals[0], stream()), sort_idx, cotangents[0], axis_, stream())};
```

When this is differentiated again, `primals[0]` is a tracer, so `sort_idx` becomes a node in the graph and the engine tries to compute a VJP with respect to the *indices* argument of `ScatterAxis`, which is explicitly unsupported.

Sorting/partitioning applies a permutation that is locally constant in the input, so the indices carry no gradient and should be cut off with `stop_gradient` — exactly what the error message recommends.

Affected sites in `mlx/primitives.cpp`:

- `Sort::vjp` / `Sort::jvp` (`sort_idx`)
- `Partition::vjp` / `Partition::jvp` (`sort_idx`)
- `Scan::vjp`, the `Max` / `Min` branch (`owner`)

### To Reproduce

```python
import mlx.core as mx

def hvp(f, x):
    return mx.grad(lambda a: mx.sum(mx.grad(f)(a)))(x)

x = mx.array([3.0, 1.0, 2.0, 5.0])

mx.grad(lambda a: mx.sum(mx.sort(a) ** 2))(x)   # fine
hvp(lambda a: mx.sum(mx.sort(a) ** 2), x)       # raises
```

```
ValueError: [scatter_axis] Cannot calculate VJP with respect to indices.
```

Same for:

```python
hvp(lambda a: mx.sum(mx.partition(a, 2) ** 2), x)
hvp(lambda a: mx.sum(mx.topk(a, 2) ** 2), x)
hvp(lambda a: mx.sum(mx.cummax(a, axis=0) ** 2), x)
hvp(lambda a: mx.sum(mx.cummin(a, axis=0) ** 2), x)
```

For comparison, these all work, which is what makes the failures look like an oversight rather than a deliberate limitation:

```python
hvp(lambda a: mx.sum(mx.max(a) ** 2), x)                     # array([0, 0, 0, 2])
hvp(lambda a: mx.sum(mx.take(a, mx.array([0, 2])) ** 2), x)  # array([2, 0, 2, 0])
hvp(lambda a: mx.sum(mx.take_along_axis(a, mx.array([0, 1, 2, 3]), axis=0) ** 2), x)
```

Note also that `mx.jvp` through `mx.sort` currently works, so forward and reverse mode disagree on whether this is supported.

### Expected behavior

Second order gradients should work, with the permutation treated as a constant. For `f(x) = sum(sort(x)**2)` the Hessian is `2 * I`, so the HVP against a vector of ones should be `[2, 2, 2, 2]`.

### Desktop

- OS: macOS 26.2
- Version: `main` @ 8c28c385 (also reproduces on 0.32.0)

Adityaj0:
Fixes #4114.

### Proposed changes

`Sort::vjp` / `Sort::jvp`, `Partition::vjp` / `Partition::jvp` and the `Max` / `Min` branch of `Scan::vjp` build an index array from the primal and feed it to a gather/scatter:

```cpp
auto sort_idx = argsort(primals[0], axis_, stream());
return {put_along_axis(
    zeros_like(primals[0], stream()), sort_idx, cotangents[0], axis_, stream())};
```

When those rules are themselves differentiated, `primals[0]` is a tracer, so the index array becomes part of the graph and the engine tries to compute a VJP with respect to the indices argument of `ScatterAxis` / `GatherAxis`. That is explicitly unsupported, so any second order derivative through these ops failed with

```
[scatter_axis] Cannot calculate VJP with respect to indices.
Use stop_gradient on indices to stop gradients from being computed.
```

even though the first derivative was fine.

The permutation these ops apply is locally constant in the input, so the indices carry no gradient. This wraps them in `stop_gradient`, which is what the error message itself recommends.

Before, on `main`, with `x = [3, 1, 2, 5]` and `hvp(f, x) = grad(lambda a: sum(grad(f)(a)))(x)`:

```
sort       -> ValueError: [scatter_axis] Cannot calculate VJP with respect to indices.
partition  -> ValueError: ...
topk       -> ValueError: ...
cummax     -> ValueError: ...
max        -> array([0, 0, 0, 2])          (already worked)
```

After:

```
sort       -> array([2, 2, 2, 2])
partition  -> array([2, 2, 2, 2])
topk       -> array([2, 0, 0, 2])
cummax     -> array([6, 0, 0, 2])
cummin     -> array([2, 6, 0, 0])
max        -> array([0, 0, 0, 2])          (unchanged)
```

### Tests

Added `test_second_order_permutation_ops` to `python/tests/test_autograd.py`, covering `sort`, `partition`, `topk`, `cummax`, `cummin`, the `reverse` and `inclusive` scan variants, and a non-trailing axis.

Rather than hardcoding expected Hessian values, the test compares the analytic HVP against a central-difference of the first gradient. The inputs are well separated so the permutation does not change across the finite-difference step. (I originally hand-computed the constants and got `cummax` wrong, so the self-checking form seemed safer.)

Verified with a CPU-only build (`-DMLX_BUILD_METAL=OFF`); `test_autograd`, `test_ops` and `test_vmap` pass.

### DIFF
--- a/mlx/primitives.cpp
+++ b/mlx/primitives.cpp
@@ -3390,7 +3390,10 @@ std::vector<array> Partition::vjp(
     const std::vector<array>& cotangents,
     const std::vector<int>& argnums,
     const std::vector<array>&) {
-  auto sort_idx = argpartition(primals[0], kth_, axis_, stream());
+  // The permutation is locally constant in the input, so cut the gradient
+  // there to keep higher order derivatives working.
+  auto sort_idx =
+      stop_gradient(argpartition(primals[0], kth_, axis_, stream()), stream());
   return {put_along_axis(
       zeros_like(primals[0], stream()),
       sort_idx,
@@ -3405,7 +3408,8 @@ std::vector<array> Partition::jvp(
     const std::vector<int>& argnums) {
   assert(primals.size() == 1);
   assert(tangents.size() == 1);
-  auto sort_idx = argpartition(primals[0], kth_, axis_, stream());
+  auto sort_idx =
+      stop_gradient(argpartition(primals[0], kth_, axis_, stream()), stream());
   auto out = take_along_axis(tangents[0], sort_idx, axis_, stream());
   return {out};
 }
@@ -4390,10 +4394,14 @@ std::vector<array> Scan::vjp(
         iota,
         array(reverse_ ? n : -1, int32),
         s);
-    auto owner = astype(
-        reverse_ ? cummin(masked, axis_, /* reverse = */ true, true, s)
-                 : cummax(masked, axis_, /* reverse = */ false, true, s),
-        uint32,
+    // The owner indices are locally constant in the input, so cut the
+    // gradient there to keep higher order derivatives working.
+    auto owner = stop_gradient(
+        astype(
+            reverse_ ? cummin(masked, axis_, /* reverse = */ true, true, s)
+                     : cummax(masked, axis_, /* reverse = */ false, true, s),
+            uint32,
+            s),
         s);
 
     if (!inclusive_) {
@@ -5432,7 +5440,9 @@ std::vector<array> Sort::vjp(
   // Sort applies a permutation to the input, so the cotangents must be
   // scattered back to the original positions (the transpose of the
   // permutation), not gathered forward as in the jvp.
-  auto sort_idx = argsort(primals[0], axis_, stream());
+  // The permutation is locally constant in the input, so cut the gradient
+  // there to keep higher order derivatives working.
+  auto sort_idx = stop_gradient(argsort(primals[0], axis_, stream()), stream());
   return {put_along_axis(
       zeros_like(primals[0], stream()),
       sort_idx,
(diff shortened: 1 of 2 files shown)


zcbenz (approved):
Nice catch, thanks!

PR #4125, opened 2026-08-10
Fix einsum not broadcasting batch dimensions in batched tensordot

### LINKED ISSUE #4124: einsum fails to broadcast size 1 batch dimensions when the smaller operand comes first
### Describe the bug

`mx.einsum` fails when an operand has a size 1 batch dimension that needs to broadcast against a larger one, but only when the smaller operand comes **first**.

```
ValueError: [reshape] Cannot reshape array of size 30 into shape (1,3,5).
```

`batch_tensordot` in `mlx/einsum.cpp` broadcasts the contracting dimensions of the two operands to a common size, but not the batch dimensions:

```cpp
// Broadcast contracting dimensions
for (int i = 0; i < a_contract.size(); ++i) {
  auto d = std::max(a.shape(a_contract[i]), b.shape(b_contract[i]));
  a_shape[a_contract[i]] = d;
  b_shape[b_contract[i]] = d;
}
```

It then builds the output shape from the first operand alone:

```cpp
for (auto ax : a_batch) {
  out_shape.push_back(a.shape(ax));
}
```

So when `a` has a batch dimension of 1, `matmul` broadcasts it up to `b`'s size while `out_shape` keeps the 1, and the final reshape gets the wrong number of elements. The reverse order happens to work because the output shape is then taken from the larger operand, which is why this went unnoticed.

### To Reproduce

```python
import mlx.core as mx

a = mx.zeros((1, 3, 4))
b = mx.zeros((2, 4, 5))

mx.einsum("...ij,...jk->...ik", a, b)
# ValueError: [reshape] Cannot reshape array of size 30 into shape (1,3,5).
```

NumPy gives `(2, 3, 5)`. Swapping the operands works in MLX:

```python
mx.einsum("...ij,...jk->...ik", mx.zeros((2, 3, 4)), mx.zeros((1, 4, 5))).shape
# (2, 3, 5)   <- fine
```

It is not specific to ellipses; explicit batch labels hit it too:

```python
mx.einsum("bij,bjk->bik", mx.zeros((1, 3, 4)), mx.zeros((2, 4, 5)))
# same error
```

And with a reduced output:

```python
mx.einsum("...i,...i->...", mx.zeros((1, 4)), mx.zeros((5, 4)))
# ValueError: [reshape] Cannot reshape array of size 5 into shape (1)
```

### Expected behavior

Batch dimensions should broadcast, matching NumPy:

| operands | NumPy | MLX (current) |
| --- | --- | --- |
| `(1,3,4)`, `(2,4,5)` | `(2,3,5)` | error |
| `(2,3,4)`, `(1,4,5)` | `(2,3,5)` | `(2,3,5)` |
| `(1,4)`, `(5,4)` -> `...` | `(5,)` | error |
| `(5,4)`, `(1,4)` -> `...` | `(5,)` | `(5,)` |

This pattern comes up whenever one shared matrix is applied across a batch, which is common enough that the asymmetry is surprising.

### Desktop

- OS Version: macOS 26.2
- Version: `main` @ 8c28c385

### Additional context

While looking at the tests I noticed `test_broadcasting` in `python/tests/test_einsum.py` has an unconditional `return` after the first call, so the assertions after it never run. Those assertions do pass on `main`, so it is unrelated to this bug, but it is dead code sitting right next to it.

Adityaj0:
Fixes #4124.

### Proposed changes

`batch_tensordot` broadcast the contracting dimensions of its two operands to a common size but left the batch dimensions alone, then built the output shape from the first operand only:

```cpp
for (auto ax : a_batch) {
  out_shape.push_back(a.shape(ax));
}
```

When the first operand had a size 1 batch dimension, `matmul` broadcast it against the second operand while `out_shape` kept the 1, so the final reshape got the wrong element count:

```python
mx.einsum("...ij,...jk->...ik", mx.zeros((1, 3, 4)), mx.zeros((2, 4, 5)))
# ValueError: [reshape] Cannot reshape array of size 30 into shape (1,3,5).
```

The reverse operand order worked, because the output shape was then taken from the larger operand. That asymmetry is why it went unnoticed. It is not specific to ellipses either, `"bij,bjk->bik"` with the same shapes fails identically.

The fix broadcasts the batch dimensions the same way the contracting dimensions already are, in the same block. `a_batch` and `b_batch` are parallel by construction (`b_batch` is built by looking up each `a_batch` label in `b`'s subscript), so they can be zipped exactly like `a_contract` / `b_contract`.

| operands | NumPy | before | after |
| --- | --- | --- | --- |
| `(1,3,4)`, `(2,4,5)` | `(2,3,5)` | error | `(2,3,5)` |
| `(2,3,4)`, `(1,4,5)` | `(2,3,5)` | `(2,3,5)` | `(2,3,5)` |
| `(1,4)`, `(5,4)` -> `...` | `(5,)` | error | `(5,)` |
| `(5,4)`, `(1,4)` -> `...` | `(5,)` | `(5,)` | `(5,)` |

### Tests

Added `test_ellipses_broadcast` to `python/tests/test_einsum.py` covering both operand orders, multiple leading batch dimensions, a reduced (`->...`) output, and the explicit-label spelling, each checked against NumPy for both shape and values.

I also removed an unconditional `return` in `test_broadcasting` that made the four lines after it dead code. To be clear about scope: those assertions pass on `main` as well, so that is an unrelated cleanup rather than something this fix enables. I checked before touching it. It was simply hiding coverage right next to the bug.

Verified with a CPU-only build (`-DMLX_BUILD_METAL=OFF`): `test_einsum`, `test_ops`, `test_autograd`, `test_vmap`, `test_linalg`, `test_blas`, `test_nn`, `test_array` and `test_reduce` pass (464 tests).

### DIFF
--- a/mlx/einsum.cpp
+++ b/mlx/einsum.cpp
@@ -356,9 +356,7 @@ array batch_tensordot(
     std::vector<int> b_batch,
     std::vector<int> b_concat,
     StreamOrDevice s) {
-  // Broadcast contracting and batch dimensions. The batch dimensions have to
-  // be broadcast as well, otherwise out_shape below is taken from ``a`` alone
-  // and does not match the batch size matmul produces.
+  // Broadcast contracting and batch dimensions.
   {
     auto a_shape = a.shape();
     auto b_shape = b.shape();
--- a/mlx/einsum.cpp
+++ b/mlx/einsum.cpp
@@ -356,7 +356,9 @@ array batch_tensordot(
     std::vector<int> b_batch,
     std::vector<int> b_concat,
     StreamOrDevice s) {
-  // Broadcast contracting dimensions
+  // Broadcast contracting and batch dimensions. The batch dimensions have to
+  // be broadcast as well, otherwise out_shape below is taken from ``a`` alone
+  // and does not match the batch size matmul produces.
   {
     auto a_shape = a.shape();
     auto b_shape = b.shape();
@@ -364,6 +366,11 @@ array batch_tensordot(
       auto d = std::max(a.shape(a_contract[i]), b.shape(b_contract[i]));
       a_shape[a_contract[i]] = d;
       b_shape[b_contract[i]] = d;
+    }
+    for (int i = 0; i < a_batch.size(); ++i) {
+      auto d = std::max(a.shape(a_batch[i]), b.shape(b_batch[i]));
+      a_shape[a_batch[i]] = d;
+      b_shape[b_batch[i]] = d;
     }
     a = broadcast_to(a, a_shape, s);
     b = broadcast_to(b, b_shape, s);
--- a/python/tests/test_einsum.py
+++ b/python/tests/test_einsum.py
@@ -188,7 +188,6 @@ class TestEinsum(mlx_tests.MLXTestCase):
         a = mx.full((5, 1), 1.0)
         b = mx.full((8, 2), 1.0)
         a_mx = mx.einsum("ab,bc->c", a, b)
-        return
         a_np = np.einsum("ab,bc->c", a, b)
         self.assertTrue(np.array_equal(a_mx, a_np))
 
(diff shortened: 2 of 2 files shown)


JasonHonKL:
Will this cause some performance issue ? @Adityaj0

Adityaj0:
Hi @JasonHonKL no, because the fix only adds lazy batch dimension broadcasting in the existing batched tensordot path, so it should have negligible overhead beyond a small amount of shape bookkeeping.

zcbenz (approved):
Looks good to me, thanks!

PR #4126, opened 2026-08-10
Deduplicate slow CPU convolution dtype dispatch

PhysicistJohn:
## Summary

The 1D, 2D, and 3D slow CPU convolution paths repeat the same
`float32`/`float16`/`bfloat16` routing. Move that mapping into one file-local,
positive dispatcher using C++20 templated callables.

Unsupported and future dtypes still fail closed with the existing diagnostic,
and no unsupported slow-convolution templates are instantiated. This removes
63 net lines without changing kernels, route selection, or public behavior.

## Testing

- Byte-identical PR-base parity for 900 supported cases: 1D/2D/3D, all
  three supported dtypes, four forced slow routes per dimension, and every
  pairing of five input and five weight layouts
- Exact unsupported-type diagnostics for `float64`, `int32`, `complex64`, and
  `bool`
- CPU convolution tests (9 passed, 9 skipped, 16 subtests passed; skips require
  Torch, which was unavailable in the validation environment)
- External symbol-name lists unchanged; convolution object, archive, and linked
  text sections shrink
- `pre-commit` and `git diff --check`

### DIFF
--- a/mlx/backend/cpu/conv.cpp
+++ b/mlx/backend/cpu/conv.cpp
@@ -6,6 +6,7 @@
 #include "mlx/backend/cpu/copy.h"
 #include "mlx/backend/cpu/encoder.h"
 #include "mlx/backend/cpu/lapack.h"
+#include "mlx/dtype_utils.h"
 #include "mlx/primitives.h"
 #include "mlx/utils.h"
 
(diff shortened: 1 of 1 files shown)


mlx/backend/cpu/conv.cpp
```diff
@@ -670,6 +671,24 @@ void slow_conv_3D(
   });
 }
 
+template <typename F>
+void dispatch_slow_conv_type(Dtype dtype, F&& f) {
```
zcbenz, line 674: Since this is adding a new utility, we should use the C++20 style and you can find an example at https://github.com/ml-explore/mlx/blob/94c29eedf4c082815d43e3e2b59ce2130b435b16/mlx/backend/cuda/gemms/gather_gemm.cu#L25-L37

(At the moment we don't plan to update all existing utilities.)

PhysicistJohn: Thanks for the pointer. I updated the new dispatcher and all three call sites to use the C++20 templated-callable style, and removed the now-unused type-tag include. I reran the 900-case parity matrix, focused CPU convolution tests, and full native suite; all pass.

zcbenz (approved):

PR #4129, opened 2026-08-10
Use common dispatch for compiled constants

PhysicistJohn:
## Summary

Replace `print_constant`'s exhaustive dtype switch with `dispatch_all_types`.
Compile-time capability branches retain the existing bool, byte-integer,
integer, floating, and complex formatting paths for every current dtype.

This removes sixteen net lines from one function.

## Testing

- Exact `compiled.cpp` Release translation-unit replay with warnings as errors;
  the replay matched the stored 41,120-byte object with a 14,836-byte text
  section
- Baseline/candidate generated-text parity for all 14 current dtypes, including
  the int8 and uint8 numeric formatting paths
- Native compile/JIT suite (24/24 cases, 117/117 assertions)
- Full native C++ suite (247/247 cases, 3,326/3,326 assertions)
- Release object and static library each shrink by 512 bytes; defined global
  object and library symbol-name lists are unchanged, while the now-unused
  `std::runtime_error(char const*)` import is removed
- Same-file overlap check against experimental #2300: it changes
  `compiled_check_contiguity` and its call in
  `compiled_collapse_contiguous_dims`, not `print_constant`; this change adds
  no new conflict to #2300's existing current-main conflict

### DIFF
--- a/mlx/backend/common/compiled.cpp
+++ b/mlx/backend/common/compiled.cpp
@@ -8,9 +8,7 @@
 namespace mlx::core {
 
 void print_constant(std::ostream& os, const array& x) {
-  bool printed = false;
   dispatch_all_types(x.dtype(), [&](auto type_tag) {
-    printed = true;
     using T = MLX_GET_TYPE(type_tag);
     if constexpr (std::is_same_v<T, bool>) {
       os << std::boolalpha << x.item<bool>();
@@ -28,9 +26,6 @@ void print_constant(std::ostream& os, const array& x) {
       throw std::runtime_error("Unsupported constant type");
     }
   });
-  if (!printed) {
-    throw std::runtime_error("Unsupported constant type");
-  }
 }
 
 std::string get_type_string(Dtype d) {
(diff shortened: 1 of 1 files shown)


mlx/backend/common/compiled.cpp
```diff
@@ -2,45 +2,34 @@
...
+      print_complex_constant<T>(os, x);
+    } else if constexpr (is_floating_point_v<T>) {
+      print_float_constant<T>(os, x);
+    } else if constexpr (std::is_integral_v<T>) {
+      print_int_constant<T>(os, x);
+    } else {
       throw std::runtime_error("Unsupported constant type");
+    }
+  });
+  if (!printed) {
```
zcbenz: I think this check is redundant?

PhysicistJohn: Thanks — yes, good catch. Given the array dtype invariant, `dispatch_all_types` covers every current `array::dtype()` value, so the outer sentinel was redundant. I removed it and reran the 14-dtype text parity, compile/JIT tests (24 cases, 117 assertions), and full C++ suite (247 cases, 3,326 assertions); all pass.

zcbenz (approved):

PR #4134, opened 2026-08-10
Fix silent Metal OOM error by sticky stream

jasonge27:
## Proposed changes

Addressing issue #3979, where the MLX drops OOM errors silently in an edge case. 

```

a = mx.full((3100, 1048576), 3.0)
b = mx.full((3100, 1048576), 4.0)
mx.eval(a, b)
c = a + b
mx.eval(c)
print(float(np.array(c[:1, :1])[0, 0]))   # returns 0.0 (M4 Pro, 48GB)
```

**What goes wrong today:**
1. Mid-eval, MLX hits `needs_commit()` and submits a buffer with no signal events (eval.cpp around the mid-eval commit).
2. That buffer fails → error is parked on the encoder.
3. Nothing is poisoned, because poison only happens for events in `signal_events`.
3. A later healthy buffer signals the eval synchronizer → mx.eval returns “success.”
4. Output was never written **→ we get 0.0.**

Later, `get_command_encoder(`) / `error_.reset()` discards the error. In the end, GPU failed, **user got silence + wrong data.**. I'm also [linking a cpp gist](https://gist.github.com/jasonge27/881afa51d1bd32e0c873e01564191998) to reproduce the error (without having to install python library).

**Changes**: 
Handling error in lazy-eval / async systems is hard. CUDA uses sticky last-error and PyTorch surfaces async errors at the next sync.  I think following industry best practices, we should **park the error per stream rather than per encoder**, and report it from synchronization points rather than from events:
- `StreamError` keyed by stream index, so it outlives the encode cycle. `get_command_encoder()` no longer resets it.
- `commit()` captures the holder by `shared_ptr` instead of `&error_`, which also removes a reference that could outlive the encoder.
- `Event::wait` checks the stream after waiting; `EventImpl::wait` no longer throws, leaving the event error to propagate across streams via wait_events as before.
- `array::wait` stops taking the`is_available()` shortcut, which detached the event as soon as the GPU signaled and skipped the check.
- The cpu-stream branch of `Event::wait(Stream)` reports nothing. It runs on a scheduler worker with no handler above it, so throwing there terminated the process rather than reaching the user.

With our changes, now the test script above would throw error `[METAL] Command buffer execution failed: Insufficient Memory (00000008:kIOGPUCommandBufferCallbackErrorOutOfMemory)`.
 
**Testing**
I added four cases into`tests/gpu_tests.cpp`. It injects the error into the completion handler, so that we can reproduce the real one needs more memory than a test machine can assume. Reverting either source change fails the matching test. Full suite passes (267 cases).
 
## Checklist

cc @zcbenz

### DIFF
--- a/mlx/array.cpp
+++ b/mlx/array.cpp
@@ -143,13 +143,18 @@ bool array::is_available() const {
 }
 
 void array::wait() {
-  if (!is_available()) {
-    if (event().valid()) {
-      event().wait();
-      detach_event();
-    }
-    set_status(Status::available);
+  if (status() == Status::available) {
+    return;
   }
+  // Wait on the event even when it is already signaled. Going through
+  // Event::wait is what reports an asynchronous backend failure, and taking
+  // the is_available() shortcut here would skip it in exactly the case the
+  // GPU raced ahead of the error being recorded.
+  if (event().valid()) {
+    event().wait();
+    detach_event();
+  }
+  set_status(Status::available);
 }
 
 void array::eval() {
--- a/mlx/backend/metal/device.cpp
+++ b/mlx/backend/metal/device.cpp
@@ -348,12 +348,6 @@ void check_stream_error(Stream s) {
   }
 }
 
-void clear_stream_error(Stream s) {
-  if (s.device == mlx::core::Device::gpu) {
-    get_stream_error(s.index)->clear_error();
-  }
-}
-
 CommandEncoder::CommandEncoder(
     Device& d,
     int index,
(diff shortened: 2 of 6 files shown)


zcbenz:
Thanks for the PR, while I think your solution works, we are in the middle of refactoring the event handling also for the cpu side (https://github.com/ml-explore/mlx/pull/3742) and I do need to have a minimal fix that would not block too much things, so I'm closing this in favor of #4174.

PR #4142, opened 2026-08-10
Fix wired limit documentation grammar

PhysicistJohn:
## Summary

Fix a missing article in the public wired-memory-limit documentation and
reflow the surrounding sentence.

## Testing

- Release Metal build with C++20 and warnings as errors
- Focused Python memory tests (3/3)
- Native C++ suite (263/263 cases, 3,581/3,581 assertions)


zcbenz:
For minor grammar errors I would prefer leaving them be.

PR #4143, opened 2026-08-10
Deduplicate zero-dimensional indexing checks

PhysicistJohn:
## Summary

Factor the identical zero-dimensional input check shared by six Python
indexing paths into a local helper.

The condition, exception type, diagnostic, and call sites are unchanged. This
removes 14 net lines from the binding implementation.

## Testing

- Release Metal build with C++20 and warnings as errors
- Exact baseline/candidate parity across all six zero-dimensional get/set
  routes; every route retains the same `ValueError` and diagnostic
- Focused Python indexing tests (4/4)
- Native C++ suite (263/263 cases, 3,581/3,581 assertions)
- Metallib bytes and linked extension exports unchanged

### DIFF
--- a/python/src/indexing.cpp
+++ b/python/src/indexing.cpp
@@ -84,12 +84,16 @@ bool is_valid_index_type(const nb::object& obj) {
       nb::ellipsis().is(obj) || nb::isinstance<nb::list>(obj);
 }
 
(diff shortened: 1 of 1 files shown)


zcbenz:
For really trivial one-condition checks it is usually more readable just keeping copies in each function.

PR #3164, opened 2026-02-24
Add all_to_all collective primitive

0xDaizz:
## Summary

Add `all_to_all` collective primitive for exchanging equal-sized chunks between all ranks. First in a series of PRs adding Expert Parallelism support.

- `mx.distributed.all_to_all(x)` — splits input along axis 0, sends chunk *i* to rank *i*, concatenates received chunks
- CPU eval via backend-specific `GroupImpl::all_to_all`
- MPI and JACCL backends implemented
- VJP support (`all_to_all` is its own transpose)

> **Follow-ups:**
> - Previous: #3158 (umbrella, now draft)
> - Next: MoE fused dispatch/combine C++ primitives (coming soon)

## Test plan

- [x] `pytest python/tests/mlx_distributed_tests.py -k all_to_all` — 4 passed, 1 skipped (shape validation requires ws>1)
- [x] 2-rank JACCL integration verified
- [ ] Linux CPU-only build (CI)
- [ ] MPI multi-rank (CI)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #4144, opened 2026-08-10
Deduplicate distributed all-reduce construction

PhysicistJohn:
## Summary

Factor the identical construction shared by distributed `all_sum`, `all_max`,
and `all_min` into a local helper.

The helper is defined immediately below `to_group`, without a forward
declaration.

The three public wrappers still select their corresponding `AllReduce`
operation. Optional-group handling, the singleton fast path, communication
stream selection, output shape, and output dtype are unchanged. This removes
16 net lines and reduces the Release object and static library by 3,896 bytes.

## Testing

- Eight-rank Ring distributed suite on both baseline and candidate (13/13
  tests per rank), including 120 all-reduce cases per rank across ten dtypes,
  four shapes, and sum/max/min
- Exact artifact and global-symbol comparison
- Same-file review of older #3158 and #3164: the candidate adds no conflict to
  their independently conflicting current-main merges, and `ops.cpp` merges
  cleanly in both cases

### DIFF
--- a/mlx/distributed/ops.cpp
+++ b/mlx/distributed/ops.cpp
@@ -12,7 +12,13 @@ namespace mlx::core::distributed {
 
 namespace {
 
-Group to_group(std::optional<Group> group);
+Group to_group(std::optional<Group> group) {
+  if (group.has_value()) {
+    return group.value();
+  } else {
+    return distributed::init();
+  }
+}
 
 array all_reduce(
     const array& x,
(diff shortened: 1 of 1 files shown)


mlx/distributed/ops.cpp
```diff
@@ -12,6 +12,25 @@ namespace mlx::core::distributed {
 
 namespace {
 
+Group to_group(std::optional<Group> group);
```
zcbenz: There is no need for forward declaration you can just put all_reduce below to_group.

PhysicistJohn: Thanks, Cheng — agreed. I moved `all_reduce` below `to_group` and removed the forward declaration. The eight-rank Ring suite passes 13/13 tests on every rank, and the native suite passes 247/247.

zcbenz (approved):

PR #4147, opened 2026-08-10
Remove redundant Metal bitwise dispatch

PhysicistJohn:
## Summary

- remove the switch whose five Metal bitwise operations all take the same path
- call `binary_op_gpu` directly from `BitwiseBinary::eval_gpu`
- remove 16 net production lines

## Testing

- focused native GPU binary tests (4 cases, 122 assertions)
- focused Python bitwise tests (2 passed)
- full native suite matches the exact-main beta-Accelerate result (257/263 cases; the same 6 linalg failures)
- changed object shrinks by 16 bytes; unchanged defined global names and byte-identical Metal library

### DIFF
--- a/mlx/backend/metal/binary.cpp
+++ b/mlx/backend/metal/binary.cpp
@@ -235,7 +235,7 @@ BINARY_GPU(Power)
 BINARY_GPU(Subtract)
 
 void BitwiseBinary::eval_gpu(const std::vector<array>& inputs, array& out) {
-  return binary_op_gpu(inputs, out, name());
+  binary_op_gpu(inputs, out, name());
 }
 
 } // namespace mlx::core
--- a/mlx/backend/metal/binary.cpp
+++ b/mlx/backend/metal/binary.cpp
@@ -235,15 +235,7 @@ BINARY_GPU(Power)
 BINARY_GPU(Subtract)
 
 void BitwiseBinary::eval_gpu(const std::vector<array>& inputs, array& out) {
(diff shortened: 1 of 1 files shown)


zcbenz (commented):
Maybe just return `binary_op_gpu` without the switch at all?

PhysicistJohn:
Thanks, Cheng — good call. I removed the switch and now return `binary_op_gpu` directly. Focused GPU tests (4 cases, 122 assertions), Python bitwise tests (2/2), and the non-linalg native suite (248/248) all pass.

mlx/backend/metal/binary.cpp
```diff
@@ -235,23 +235,7 @@ BINARY_GPU(Power)
...
-      binary_op_gpu(inputs, out, name());
-      break;
-    case BitwiseBinary::LeftShift:
-      binary_op_gpu(inputs, out, name());
-      break;
-    case BitwiseBinary::RightShift:
-      binary_op_gpu(inputs, out, name());
-      break;
-  }
+  return binary_op_gpu(inputs, out, name());
```
zcbenz: return is not needed

PhysicistJohn: Thanks, Cheng — agreed. I removed the unnecessary return and kept the direct binary_op_gpu call. The focused native GPU binary tests (4 cases, 122 assertions) and Python bitwise tests (2/2) pass.

zcbenz (approved):

PR #4148, opened 2026-08-10
Group equivalent Metal integer reduction cases

PhysicistJohn:
## Summary

- group equivalent Metal integer sum/product accumulator mappings
- return the input dtype directly for the four identity mappings
- retain the existing default diagnostic while removing 5 net production lines

## Testing

- 96-case GPU parity against exact main across all eight integer dtypes, sum/product, axes, and `keepdims`
- full Python reduce tests (12 passed)
- byte-identical Metal library and unchanged defined global names; changed object is 408 bytes smaller

### DIFF
--- a/mlx/backend/metal/reduce.cpp
+++ b/mlx/backend/metal/reduce.cpp
@@ -249,21 +249,16 @@ std::pair<Dtype, Dtype> remap_reduce_types(
     if (issubdtype(in.dtype(), integer)) {
       switch (in.dtype()) {
         case uint8:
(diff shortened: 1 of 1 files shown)


PhysicistJohn:
Closing this one after an additional edge-case review: returning `in.dtype()` preserves a manually constructed noncanonical size, while the original switch normalizes recognized values to the canonical dtype constants. Canonical arrays behave identically, but that is still a behavior change and does not meet the exact-preservation bar for this cleanup.

PR #4150, opened 2026-08-10
Deduplicate scalar-vector output allocation

PhysicistJohn:
## Summary

- share the identical scalar-vector and vector-scalar output-allocation path
- select the corresponding vector operand and its already-computed donation flag
- remove 6 net production lines without changing allocation or donation behavior

## Testing

- 80-case scalar/vector GPU parity against exact main across four dtypes and five binary operations
- focused Python binary tests (3 passed)
- byte-identical Metal library and unchanged defined global names; `libmlx.a` is 1,920 bytes smaller


zcbenz:
I think this would reduce the readability of the code.

PR #4153, opened 2026-08-10
Fix stream API documentation grammar

PhysicistJohn:
## Summary

Correct the grammar and spelling in the public `new_thread_unsafe_stream` Python docstring.

The sentence now says that it is the user's responsibility to ensure there is no race condition. Runtime stream behavior is unchanged.

## Testing

- Release build with Metal and Python bindings using C++20 and warnings as errors
- Direct import of the candidate extension and exact public docstring check
- Unsafe-stream array evaluation smoke test
- Python-extension artifact and exported-symbol check
- Fully paginated open-PR overlap and collision scan


zcbenz:
For minor grammar errors I would prefer leaving them be.

PR #4164, opened 2026-08-10
Decode the e4m3 NaN encodings in from_fp8

### LINKED ISSUE #4135: [BUG] Regression: `mx.from_fp8` decodes SafeTensors `F8_E4M3` NaNs as finite values
## Summary

`mx.from_fp8()` does not preserve the two canonical NaN encodings of the
SafeTensors `F8_E4M3` / PyTorch `float8_e4m3fn` format:

| Raw byte | Expected SafeTensors/PyTorch value | Actual MLX `from_fp8` value |
| --- | --- | --- |
| `0x7f` | `NaN` | `+480.0` |
| `0xff` | `NaN` | `-480.0` |

This is a regression from the behavior tested in [PR #1859](https://github.com/ml-explore/mlx/pull/1859), introduced when [PR #2985](https://github.com/ml-explore/mlx/pull/2985) moved FP8 conversion to the explicit `mx.from_fp8()` API.

This report is self-contained: the complete minimal reproducer is included
below. It creates one local, four-byte SafeTensors file; it does not download
a model, execute remote code, or run model inference.

## Environment

- MLX `0.32.0`
- Python `3.14`
- SafeTensors `0.8.0`
- PyTorch `2.13.0`
- Apple M2 Pro, 16-core GPU, Metal backend
- Current MLX `main` source was inspected at
  `e78d894c8f718a805341e1aa1cc835d9c9b8462` (2026-08-10); it has the same
  CPU and Metal decode logic.

## Complete minimal reproduction

Save the following as `repro.py` and run `python repro.py` in an environment
with `mlx`, `torch`, and `safetensors` installed:

```python
import json
import math
import struct
import tempfile
from pathlib import Path

import mlx.core as mx
import torch
from safetensors import safe_open


def write_fixture(directory: Path) -> Path:
    header = {
        "w": {
            "dtype": "F8_E4M3",
            "shape": [4],
            "data_offsets": [0, 4],
        }
    }
    header_bytes = json.dumps(header, separators=(",", ":")).encode("utf-8")
    path = directory / "f8_nan.safetensors"
    path.write_bytes(
        struct.pack("<Q", len(header_bytes))
        + header_bytes
        + bytes([0x00, 0x3C, 0x7F, 0xFF])
    )
    return path


with tempfile.TemporaryDirectory(prefix="mlx-f8-nan-decode-") as tmp:
    path = write_fixture(Path(tmp))

    # MLX's documented native-FP8 path.
    raw = mx.load(path)["w"]
    raw_values = raw.tolist()
    mlx_values = mx.from_fp8(raw, dtype=mx.float32).tolist()

    # Canonical SafeTensors/PyTorch interpretation of the exact same bytes.
    with safe_open(path, framework="pt") as tensors:
        reference_values = tensors.get_tensor("w").to(torch.float32).tolist()

assert raw.dtype == mx.uint8
assert raw_values == [0, 60, 127, 255]
assert mlx_values[:2] == [0.0, 1.5]
assert reference_values[:2] == [0.0, 1.5]
assert mlx_values[2:] == [480.0, -480.0]
assert all(math.isnan(value) for value in reference_values[2:])

print("raw:        ", raw_values)
print("MLX:        ", mlx_values)
print("SafeTensors:", reference_values)
```

Actual output on the Metal system above:

```text
raw:         [0, 60, 127, 255]
MLX:         [0.0, 1.5, 480.0, -480.0]
SafeTensors: [0.0, 1.5, nan, nan]
```

## Why `NaN` is the expected result

The SafeTensors PyTorch binding maps `F8_E4M3` to
`torch.float8_e4m3fn`. For that format, `0x7f` and `0xff` are the two NaN
encodings; the maximum finite values are `0x7e` and `0xfe` (`+448` and
`-448`). The reproducer obtains the reference through the SafeTensors binding
itself rather than reimplementing the format.

MLX previously implemented and tested the same expectation. [PR #1859](https://github.com/ml-explore/mlx/pull/1859), titled “Support loading F8_E4M3 from safetensors,” added a fixture containing `0x7f` and `0xff` and expected both entries to be `mx.nan`.

## Regression provenance

[PR #2985](https://github.com/ml-explore/mlx/pull/2985) intentionally changed
`mx.load()` to retain FP8 data as a `uint8` carrier and exposed explicit
`mx.from_fp8()` conversion. That API change is reasonable and is **not** the
bug reported here.

However, the same PR replaced the prior `FromFP8` bit-level conversion, which
handled the NaN encodings, with the current float16-based conversion. It also
changed the prior loader test from the NaN bytes `0x7f`/`0xff` to the ordinary
maximum-finite bytes `0x7e`/`0xfe`; the NaN cases were no longer tested.

Therefore, this can be fixed while keeping PR #2985's intended `uint8`
carrier API unchanged.

## Root cause

The current CPU decoder is effectively:

```cpp
auto v = Simd<uint16_t, N>(x & 127) << 7;
auto converted = *(Simd<float16_t, N>*)(&v);
out = converted * 256.0;
return select(x & 128, -out, out);
```

The Metal decoder uses the same transform. Neither special-cases values for
which `(x & 0x7f) == 0x7f`, so the reserved NaN encodings are interpreted as
finite half-derived values instead.

Relevant current source locations:

- [`mlx/ops.cpp`](https://github.com/ml-explore/mlx/blob/main/mlx/ops.cpp#L5411-L5429)
  constructs `fast::ConvertFP8` for `from_fp8`.
- [`mlx/backend/cpu/unary_ops.h`](https://github.com/ml-explore/mlx/blob/main/mlx/backend/cpu/unary_ops.h#L154-L172)
  contains the CPU decoder.
- [`mlx/backend/metal/kernels/fp8.h`](https://github.com/ml-explore/mlx/blob/main/mlx/backend/metal/kernels/fp8.h#L32-L46)
  contains the equivalent Metal conversion.

## Verified upper-layer impact (MLX-LM)

This is not only a raw-array API difference. In an MLX-LM checkout, the
following normal model-loading path was exercised with an inert local fixture:

```text
mlx_lm.utils.load_model
  -> mx.load(model.safetensors)
  -> MiniMax Model.sanitize(weights)
  -> mx.from_fp8(weight, dtype=mx.bfloat16)
  -> model.load_weights(...)
```

The fixture contained one `128 x 128` `F8_E4M3` weight block, one `F32` scale
of `1.0`, and only its first weight byte set to the indicated value. It did
not download or run an external model. The actual result was:

```text
0x7f: loaded=480.0; q_proj MLX-LM=480.0; canonical=nan
0xff: loaded=-480.0; q_proj MLX-LM=-480.0; canonical=nan
```

Thus a standard MLX-LM sanitizer materializes a finite parameter where the
same immutable SafeTensors artifact has NaN semantics in the canonical reader.
This downstream check is supplementary; the four-byte reproduction above is
complete for the MLX core bug.

## Suggested fix

Before or after the ordinary conversion, preserve the two NaN bit patterns:

```cpp
const auto is_nan = (x & 0x7f) == 0x7f;
// Decode ordinary values as today, then select quiet_NaN for is_nan.
```

Apply the equivalent behavior to CPU and Metal, and check other backends for
their intended `F8_E4M3` semantics.

## Acceptance test

Add `0x7f` and `0xff` back to a `mx.from_fp8()` regression test, asserting
`NaN` with `equal_nan=True`, alongside maximum-finite `0x7e` and `0xfe`:

```text
bytes:    0x7e    0xfe    0x7f   0xff
expected: +448.0  -448.0  NaN    NaN
```

The supplied four-byte reproduction can be used verbatim as the test input.

## Scope

This is a correctness/interoperability regression, not a report that
`mx.load()` returning `uint8` is unsafe. The raw `uint8` carrier is the
intentional API introduced by PR #2985, and raw-byte use (including `0x80`) is
outside this report.

The issue can become security-relevant only in a separate workflow that
validates/rejects F8 NaNs using a canonical reader and then deploys the same
artifact through MLX. This report does not claim code execution, a generic
validation bypass, or a security severity.

axiom-of-choice:
## Proposed changes

Fixes #4135.

`0x7f` and `0xff` are e4m3's only NaN encodings, and `from_fp8` decoded both as `+/-480.0`.

The decode reinterprets `(bits & 127) << 7` as a float16 and scales by 256, which is exact for every finite e4m3 value. It does not carry NaN: the shift leaves those two patterns on exponent field 15 rather than 31, so float16 reads them as a normal number, and `1.875 * 256 = 480`.

Worth noting beyond the issue: 480 is above the format's largest finite magnitude of 448, so a NaN weight did not merely lose its NaN-ness, it silently became an out-of-range value that reads as ordinary data.

Both decoders had the same defect. `mlx/backend/metal/kernels/fp8.h`'s `fp8_e4m3` is not a native Metal type, it performs the same shift-and-scale as the CPU path, so CPU and GPU agreed on the wrong answer.

## Verification

Exhaustive over all 256 byte values against `torch.float8_e4m3fn`, on both streams:

| stream | mismatches before | mismatches after |
| --- | --- | --- |
| CPU | 2 (`0x7f`, `0xff`) | 0 |
| Metal | 2 (`0x7f`, `0xff`) | 0 |

Those two bytes were the only ones that ever disagreed, so nothing finite moves.

The new test fails without the change with `array([480, -480], dtype=float16)`. It pins both NaN encodings across float16, bfloat16 and float32 on both streams, and pins 448 as finite so a future change cannot swallow the maximum. The pre-existing round-trip test only covers finite values, which is why this went unnoticed.

Full suites on an M2 Pro (macOS 26.5.2, Metal build): 759 passed / 46 skipped / 10804 subtests in Python, and 261 C++ cases with 3521 assertions.

## Out of scope

`to_fp8` does not preserve NaN either, saturating it to `0x7e` (448) the way it saturates overflow, where torch maps both to NaN. That is a separate call and its saturating behaviour looks deliberate, so this PR leaves it alone. Happy to follow up if you want it changed.

## Checklist

### DIFF
--- a/mlx/backend/cpu/unary_ops.h
+++ b/mlx/backend/cpu/unary_ops.h
@@ -5,6 +5,7 @@
 #include <stdint.h>
 #include <cmath>
 #include <complex>
+#include <limits>
 
 #include "mlx/backend/cpu/simd/simd.h"
 
@@ -165,6 +166,13 @@ struct FromFP8 {
         out[i] = converted * 256.0;
       }
     }
+    // 0x7f/0xff are e4m3's only NaN encodings. Shifted into float16 they land
+    // on a finite exponent (15, not 31), so the reinterpret above decodes them
+    // as 480.
+    out = select(
+        Simd<bool, N>((x & 127) == 127),
+        Simd<float, N>(std::numeric_limits<float>::quiet_NaN()),
+        out);
     auto sign = Simd<bool, N>(x & 128);
     return select(sign, -out, out);
   }
--- a/mlx/backend/metal/kernels/fp8.h
+++ b/mlx/backend/metal/kernels/fp8.h
@@ -33,6 +33,12 @@ struct fp8_e4m3 {
     uint16_t v = (bits & 127) << 7;
     half converted = as_type<half>(v);
     converted *= 256.0;
+    // 0x7f/0xff are e4m3's only NaN encodings. Shifted into half they land on a
+    // finite exponent (15, not 31), so the reinterpret above decodes them as
+    // 480.
+    if ((bits & 127) == 127) {
+      converted = NAN;
+    }
     auto sign = bits & 128;
     return (sign ? -converted : converted);
   }
--- a/python/tests/test_ops.py
+++ b/python/tests/test_ops.py
@@ -3609,6 +3609,23 @@ class TestOps(mlx_tests.MLXTestCase):
         self.assertTrue(mx.array_equal(mx.from_fp8(mx.to_fp8(vals)), vals))
         self.assertTrue(mx.array_equal(mx.from_fp8(mx.to_fp8(-vals)), -vals))
 
+    def test_from_fp8_nan(self):
+        # 0x7f/0xff are e4m3's only NaN encodings. Shifted into float16 they land on
+        # a finite exponent, so a plain reinterpret decodes them as +/-480.
+        nans = mx.array([0x7F, 0xFF], mx.uint8)
+        for dtype in [mx.float16, mx.bfloat16, mx.float32]:
+            for stream in [mx.cpu, mx.gpu]:
+                if stream == mx.gpu and not mx.metal.is_available():
+                    continue
+                out = mx.from_fp8(nans, dtype=dtype, stream=stream)
+                self.assertTrue(
+                    mx.all(mx.isnan(out)).item(), msg=f"{dtype} {stream}: {out}"
+                )
+
+        # the largest finite magnitude is 448 (0x7e), and must stay finite
+        finite = mx.from_fp8(mx.array([0x7E, 0xFE], mx.uint8), dtype=mx.float32)
+        self.assertTrue(mx.array_equal(finite, mx.array([448.0, -448.0])))
+
     def test_zeros_ones_empty_like_dtype(self):
         x = mx.array([1, 2, 3], dtype=mx.int32)
 


erwinzhang7:
Confirmed on an M5 Max, macOS 26.5, against `c2bcf47`:

```
from_fp8(0x7f)   480.0  ->  nan
from_fp8(0xff)  -480.0  -> -nan
```

This is one of four open PRs fixing NaN handling in different ops. I applied all
four together and measured the result in
https://github.com/ml-explore/mlx/pull/4146#issuecomment-5309688331 — five ops
disagree with numpy on main, none do with the four applied.

zcbenz (changes requested):
Adding a branch is likely to cause heavy performance regression. Practically it is harmless bug so unless there is an efficient implementation I would deny fixing it.

PR #4168, opened 2026-08-10
add complex to scalar to avoid raising warnings by linters

aaishwarymishra:
## Proposed changes

This pull request makes a small change to the `python/mlx/_stub_patterns.txt` file by updating the `scalar` type alias to include the `complex` type in addition to `int`, `float`, and `bool`. This allows the codebase to support complex numbers where `scalar` is used.

## Checklist


zcbenz (approved):
Thanks!

PR #4172, opened 2026-08-11
docs: fix stale CLI invocations in LLM inference example

deBrian07:
## Proposed changes## Proposed changes

The example commands in the LLM inference guide no longer run as written, tried copying and pasting them and all failed.

Both scripts in `mlx-examples/llms/llama` now accept flags only:

- `llama.py`: `--model-path` (default `mlx_model`), `--prompt`, `--few-shot`,
  `--max-tokens`/`-m`, `--write-every`, `--temp`, `--seed`. No positional arguments.
- `convert.py`: `--torch-path`, `--mlx-path`, `--model-name`, `--quantize`,
  `--q-group-size`, `--q-bits`, `--dtype`. Writes a directory (default `mlx_model`),
  not a single `.npz` file.

`docs/src/examples/llama-inference.rst` still passes the weights and tokenizer
positionally, and uses `--num-tokens`, which no longer exists. This updates the
`convert.py` call, the three `llama.py` calls, and the surrounding sentence to
match the current CLI. Prompts, sample output, and timings are unchanged.

This is only a doc change, found when following the guide, no corresponding issues. 

## Checklist

### DIFF
--- a/docs/src/examples/llama-inference.rst
+++ b/docs/src/examples/llama-inference.rst
@@ -322,14 +322,14 @@ several unnecessary copies from disk to numpy and then from numpy to MLX. It
 will be replaced in the future with direct loading to MLX.
 
 You can download the full example code in `mlx-examples`_. Assuming, the
-existence of ``weights.pth`` and ``tokenizer.model`` in the current working
-directory we can play around with our inference script as follows (the timings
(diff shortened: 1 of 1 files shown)


zcbenz (approved):
Thanks!

PR #4174, opened 2026-08-11
Fix metal error escaping handler

zcbenz:
When a kernel is committed separately from event signaling, the event would be signaled even when the kernel failed, and it is possible that the waiter of the event would be signaled before the completion handler sets the error, ends up with the error getting silently ignored.

Using the example from #4134:

```python

a = mx.full((3100, 1048576), 3.0)
b = mx.full((3100, 1048576), 4.0)
mx.eval(a, b)
c = a + b
mx.eval(c)
```

While `a+b` failed due to OOM, the synchronizer event would not get the error when eval ends.

This PR makes the error persistent in command encoder, until it is thrown either by event, or when users tries to use the stream. 

Using the above example, with this PR while `eval(c)` still does not throw, following calls will eventually receive the error.

```python
c = a + b
mx.eval(c) # do not throw
mx.arange(10) # throws
```

### DIFF
--- a/mlx/backend/common/error.h
+++ b/mlx/backend/common/error.h
@@ -1,0 +1,49 @@
+// Copyright © 2026 Apple Inc.
+
+#pragma once
+
+#include <atomic>
+#include <memory>
+#include <string>
+
+namespace mlx::core {
+
+class Error {
+ public:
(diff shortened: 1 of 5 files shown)


zcbenz:
Merging the change to #3742 since they overlap too much.

PR #4176, opened 2026-08-11
Refactor CPU ScatterAxis and MaskedScatter dtype dispatch

PhysicistJohn:
## Summary

Replace the remaining exhaustive payload dtype switches in CPU `ScatterAxis`
and `MaskedScatter` with the existing `dispatch_all_types` helper. This keeps
the restricted integer index dispatch, reduction mode selection, broadcasting,
source traversal, and diagnostics unchanged while removing 83 net lines.

## Testing

- Release CPU-only build with global `-Werror`
- `python/tests/test_ops.py` (153/153 tests)
- focused `vmap` and autograd coverage for both operations
- 1,456 ScatterAxis parity cases and 14 exact unsupported-index diagnostics
- 602 MaskedScatter parity cases and 14 exact short-source diagnostics
- baseline/candidate symbol and linked-export sets match; the indexing object is 416 bytes smaller

### DIFF
--- a/mlx/backend/cpu/indexing.cpp
+++ b/mlx/backend/cpu/indexing.cpp
@@ -555,53 +555,10 @@ void ScatterAxis::eval_cpu(const std::vector<array>& inputs, array& out) {
                     idx = array::unsafe_weak_copy(idx),
                     updates = array::unsafe_weak_copy(updates),
                     out = array::unsafe_weak_copy(out)]() mutable {
-    switch (out.dtype()) {
-      case bool_:
-        dispatch_scatter_axis<bool>(out, idx, updates, axis_, reduce_type_);
-        break;
-      case uint8:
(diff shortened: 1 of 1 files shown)


zcbenz:
Would you mind updating all the dispatch code in the file in one pull request? Our CI is throttled to handle many pull requests.

PhysicistJohn:
Thanks, Cheng — agreed. I folded the remaining `MaskedScatter` payload switch
into this PR, so the last manual exhaustive payload-dtype router in
`indexing.cpp` now uses `dispatch_all_types`. The restricted integral-index
routers and reduction-mode switches remain unchanged because they encode
narrower type and operation semantics, including the existing diagnostics. The
combined diff is +8/-91. The Release `-Werror` build, 247/247 native cases,
153/153 Python ops tests, and paired ScatterAxis/MaskedScatter parity and
diagnostic checks all pass.

zcbenz (approved):

PR #4182, opened 2026-08-11
Compute float64 exp, sin, cos and erf in double

### LINKED ISSUE #4158: [BUG] `mx.sin` / `mx.cos` / `mx.exp` on float64 return float32-class accuracy (max rel err 3.2e-04 for cos)
**Describe the bug**
For a `float64` input, `mx.sin`, `mx.cos` and `mx.exp` return results whose **relative** error is of order
`1e-7`  (i.e. float32-class) while the returned dtype is `float64`. `mx.tan`, `mx.log` and `mx.sqrt` on
the same inputs are correctly rounded (~1e-16), so this is specific to these three functions rather than a
general float64 limitation.

`mx.cos` is the worst at **3.2e-04** max relative error, which is ~1e12x float64 epsilon.

**To Reproduce**

```python
import mlx.core as mx, numpy as np, mpmath as mp
mp.mp.dps = 60
rng = np.random.default_rng(0)
xs = rng.uniform(-3, 3, 2000)

with mx.stream(mx.cpu):
    A = mx.array(xs, dtype=mx.float64)
    for name, fn, ref in (("sin", mx.sin, mp.sin), ("cos", mx.cos, mp.cos), ("exp", mx.exp, mp.exp),
                          ("tan", mx.tan, mp.tan), ("log", mx.log, mp.log), ("sqrt", mx.sqrt, mp.sqrt)):
        src = np.abs(xs) if name in ("log", "sqrt") else xs
        B = np.asarray(fn(mx.abs(A) if name in ("log", "sqrt") else A))
        rel = max(abs(float((mp.mpf(float(B[i])) - ref(mp.mpf(float(src[i])))) / ref(mp.mpf(float(src[i])))))
                  for i in range(len(xs)))
        print(f"mx.{name:5s} float64 max relative error = {rel:.3e}")
```

```
mx.sin   float64 max relative error = 7.746e-07     <-- float32-class
mx.cos   float64 max relative error = 3.206e-04     <-- float32-class, and 1e12x float64 eps
mx.exp   float64 max relative error = 2.932e-07     <-- float32-class
mx.tan   float64 max relative error = 3.077e-16     ok
mx.log   float64 max relative error = 2.452e-16     ok
mx.sqrt  float64 max relative error = 1.094e-16     ok
```

Single-value form, if that is easier to triage:

```python
import mlx.core as mx, numpy as np, mpmath as mp
with mx.stream(mx.cpu):
    print(float(np.asarray(mx.cos(a))[0]))   # actual:  0.5403022766113281   <- float32 value
print(float(mp.cos(mp.mpf(1))))              # correct: 0.5403023058681398
```

**Expected behavior**
A `float64` input should give a `float64`-accurate result, within a few ULP (~1e-16 relative), as `tan`,
`log` and `sqrt` already do.

**Environment**
`mlx 0.32.0` · python 3.12 · Apple M2 (arm64), macOS · float64 evaluated under `with mx.stream(mx.cpu)`
(float64 is CPU-only) · reference: `mpmath` at 60 decimal digits.

AKnassa:
## Proposed changes

Fixes #4158.

`exp`, `sin`, `cos` and `erf` in `mlx/backend/cpu/simd/math.h` are hand-rolled
float32 approximations, and they convert their input to `Simd<float, N>`
unconditionally. float64 therefore gets float32 accuracy:

| max rel. error, float64 | before | after |
| --- | --- | --- |
| `exp` | 3.00e-07 | 0.00e+00 |
| `sin` | 7.75e-07 | 0.00e+00 |
| `cos` | 3.42e-04 | 0.00e+00 |
| `erf` | 9.44e-04 | 0.00e+00 |

`exp` is worse than imprecise, because it also inherited float's saturation
points, so a float64 array returns the wrong value outright:

```
exp(100.0)  -> inf   (should be 2.688e+43)
exp(709.0)  -> inf   (should be 8.218e+307)
exp(-200.0) -> 0     (should be 1.384e-87)
```

This routes double through libm and leaves float32 untouched.

Worth flagging: the existing `test_unary_ops` in `python/tests/test_double.py`
asserts `allclose(y, y_double.astype(mx.float32, mx.cpu))` — it compares the
float64 result against the float32 one, so it could not catch any of this.

### Performance

This is a real cost. 2^20 elements, M2 Pro, best of 7 runs:

| | float32 (unchanged) | float64 before | float64 after |
| --- | --- | --- | --- |
| `exp` | 0.437 -> 0.438 ms | 0.573 ms | 5.908 ms |
| `sin` | 0.804 -> 0.807 ms | 1.067 ms | 6.659 ms |
| `cos` | 0.786 -> 0.791 ms | 1.032 ms | 6.743 ms |
| `erf` | 0.877 -> 0.878 ms | 1.160 ms | 11.641 ms |

float32 is unchanged, as expected from a compile-time branch. The old float64
speed was the speed of computing the wrong answer.

I checked whether that speed can be recovered. Accelerate's vectorized double
math (`simd::exp` / `simd::erf` on `double4`) measured no faster than scalar
libm on this machine — 2.456 vs 2.558 ms for `exp`, 9.653 vs 9.951 ms for `erf`,
with bit-identical results — so the cost looks intrinsic to evaluating these at
double precision rather than a poor choice of implementation. Happy to take
direction if you would prefer a different trade-off here.

`erfinv` still narrows to float. It has no libm equivalent and would need a
double precision polynomial, so I left it for a separate change rather than
half-do it.

### Testing

Two new tests in `python/tests/test_double.py`: one pins the four ops to double
accuracy against numpy, one pins the range of `exp`. Both fail before this
change and pass after.

- Full Python suite: 821 tests, no new failures.
- Full C++ suite: 247 cases, 3326 assertions, all pass.

Built CPU-only (`MLX_BUILD_METAL=OFF`). float64 is CPU-only in MLX, so that is
where this behaviour lives, but note the GPU suite was not run on my machine.

## Checklist

- [ ] I have updated the necessary documentation (if needed) — no documentation change needed, there is no API change

### DIFF
--- a/mlx/backend/cpu/simd/math.h
+++ b/mlx/backend/cpu/simd/math.h
@@ -9,6 +9,21 @@ namespace mlx::core::simd {
 constexpr float inf = std::numeric_limits<float>::infinity();
 
 /**
+ * Evaluate a scalar function on each lane.
+ *
+ * The approximations in this file are tuned for float32. They are not accurate
+ * enough for float64, so double precision falls back to libm through here.
+ */
+template <typename T, int N, typename Op>
+Simd<T, N> lanewise(Simd<T, N> in, Op op) {
+  Simd<T, N> out;
+  for (int i = 0; i < N; ++i) {
+    out[i] = op(in[i]);
+  }
+  return out;
+}
+
+/**
  * Compute exp(x) in an optimizer friendly way as follows:
  *
  * First change the problem to computing 2**y where y = x / ln(2).
@@ -28,6 +43,10 @@ template <typename T, int N>
 Simd<T, N> exp(Simd<T, N> in) {
   if constexpr (is_complex<T>) {
     return Simd<T, 1>{std::exp(in.value)};
+  } else if constexpr (std::is_same_v<T, double>) {
+    // The polynomial below also saturates outside of float's range, which
+    // would clamp double results to zero or infinity beyond +-88.
+    return lanewise(in, [](double v) { return std::exp(v); });
   } else {
     Simd<float, N> x_init = in;
     auto x = x_init * 1.442695f; // multiply with log_2(e)
@@ -119,6 +138,8 @@ template <typename T, int N>
 Simd<T, N> sin(Simd<T, N> x) {
   if constexpr (is_complex<T>) {
     return std::sin(x.value);
+  } else if constexpr (std::is_same_v<T, double>) {
+    return lanewise(x, [](double v) { return std::sin(v); });
   } else {
     return sincos<true>(x);
   }
@@ -128,6 +149,8 @@ template <typename T, int N>
 Simd<T, N> cos(Simd<T, N> x) {
   if constexpr (is_complex<T>) {
     return std::cos(x.value);
+  } else if constexpr (std::is_same_v<T, double>) {
+    return lanewise(x, [](double v) { return std::cos(v); });
   } else {
     return sincos<false>(x);
   }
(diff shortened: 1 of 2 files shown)


zcbenz (changes requested):
At the moment we don't plan to taking non-critical simd changes until we merge https://github.com/ml-explore/mlx/pull/3019.

PR #4183, opened 2026-08-11
Fix compile leaking multi-output arrays

### LINKED ISSUE #3932: Memory leak when compiled function contains multi-output op
## Summary

When a function that is compiled with `mx.compile` contains a **multi-output primitive**
(`mx.split`, and by construction any other `make_arrays` op) and closes over arrays that were
**evaluated outside the compiled function**, dropping the compiled function and the captured
arrays does not release their memory. `mx.get_active_memory()` stays elevated permanently and
accumulates without bound across repetitions.

Replacing the single `mx.split` with two equivalent single-output slices removes the leak
entirely, leaving the numerics bit-identical.

## Environment

| | |
|---|---|
| MLX | 0.31.1 (`mlx==0.31.1`, `mlx-metal==0.31.1`, from PyPI) |
| Hardware | Apple M4 Max |
| OS | macOS 26.5.2 |
| Python | 3.12 |
| Backend | Metal (`Device(gpu, 0)`) |

### Minimal reproduction

The script below is the whole report. The fuller ablation harness that produced the tables — same
model, one flag per ingredient — is committed beside this document as
[`1M3R3-09-T19-repro.py`](1M3R3-09-T19-repro.py).

```python
import gc
import mlx.core as mx

BLOCKS, DIM, BATCH = 24, 512, 8          # 24 MiB of "weights"
x0 = mx.random.normal((BATCH, DIM)); mx.eval(x0)

def make_forward(weights):
    def forward(x):
        for w in weights:
            h = x @ w
            a, b = mx.split(h, 2, axis=-1)          # <-- multi-output primitive
            x = mx.concatenate([b, a], axis=-1)
        return x
    return forward

for cycle in range(10):
    mx.synchronize(); base = mx.get_active_memory()

    weights = [mx.random.normal((DIM, DIM)) for _ in range(BLOCKS)]
    fwd = make_forward(weights)

    mx.eval(weights)                                # <-- evaluated OUTSIDE the compiled function

    fn = mx.compile(fwd)
    out = fn(x0); mx.eval(out); del out

    del fn, fwd, weights
    gc.collect()

    mx.synchronize()
    print(cycle, (mx.get_active_memory() - base) // (1024 * 1024), "MiB retained")
```

## Expected vs actual

**Expected:** each iteration prints `0 MiB retained`.

**Actual:** each iteration prints `24 MiB retained` — the entire weight set — and the process
footprint grows by 24 MiB per iteration without limit.

Measured over 200 iterations: **200/200 leaked, 24.00 MiB every time**, process residual
4800 MiB. Nothing is ever reclaimed; `mx.clear_cache()` does not help (the buffers were never
freed to the allocator, so they never enter the recycle pool).

### All three ingredients are required

n = 30 iterations per configuration, same machine, same script, one flag changed at a time:

| `mx.split` | `mx.compile` | weights evaluated outside | leaked / 30 |
|---|---|---|---|
| ✅ | ✅ | ✅ (`mx.eval(weights)`) | **30 / 30** |
| ✅ | ✅ | ✅ (an eager uncompiled `fwd(x0)`) | **30 / 30** |
| ❌ two slices | ✅ | ✅ | 0 / 30 |
| ✅ | ❌ | ✅ | 0 / 30 |
| ✅ | ✅ | ❌ | 0 / 30 |

The "two slices" row replaces

```python
a, b = mx.split(h, 2, axis=-1)
```

with

```python
half = h.shape[-1] // 2
a = h[..., :half]
b = h[..., half:]
```

and nothing else. It is a complete fix in this reproduction.

Note the third ingredient: if the captured arrays are *never* evaluated outside the compiled
function, they are owned by the compile-cache tape and released with it, so the leak does not
appear. Any real model that materializes its weights before or alongside a compiled call —
loading them, running a warm-up, computing a reference — satisfies it.

## Heap evidence

`leaks(1)` against the parked reproducer (3 iterations, `MallocStackLogging=1`):

```
Process 83244: 50 leaks for 8816 total leaked bytes.

STACK OF 1 INSTANCE OF 'ROOT CYCLE: <malloc in mlx::core::array::make_arrays(
    std::vector<mlx::core::SmallVector<int, 10ul>>, std::vector<mlx::core::Dtype> const&,
    std::shared_ptr<mlx::core::Primitive> const&, std::vector<mlx::core::array> const&)>':

    48 (8.53K) ROOT CYCLE: <malloc in mlx::core::array::make_arrays(...) 0xa0b0addd0> [48]
       44 (7.97K) ROOT CYCLE: <std::__shared_ptr_emplace<mlx::core::array::ArrayDesc> 0xa0b823480> [448]
          CYCLE BACK TO <malloc in mlx::core::array::make_arrays(...) 0xa0b0addd0> [48]
          1 (160 bytes) <std::__shared_ptr_emplace<mlx::core::Split> 0xa0a8f6580> [160]
       3 (528 bytes) ROOT CYCLE: <std::__shared_ptr_emplace<mlx::core::array::ArrayDesc> 0xa0b8239c0> [448]
```

Unreachable, cyclic, rooted at `make_arrays`, primitive `mlx::core::Split`.

`leaks` reports kilobytes rather than megabytes because it only walks malloc zones: the leaked
nodes are the `ArrayDesc` / `Split` / sibling-vector control blocks, while the buffers they pin
are Metal VM regions. `vmmap` on the same process attributes the retained bytes to
`IOAccelerator (graphics)`.

## Analysis

`array::make_arrays` (`mlx/array.cpp:42-58`) gives every output of a multi-output primitive a
**by-value copy of each of its siblings**:

```cpp
for (size_t i = 0; i < outputs.size(); ++i) {
  auto siblings = outputs;
  siblings.erase(siblings.begin() + i);
  outputs[i].set_siblings(std::move(siblings), i);
}
```

That is a reference cycle by construction: each `ArrayDesc` holds a `shared_ptr` to the others.
`~ArrayDesc` is aware of this and breaks it — but only for nodes it first judges *deletable*
(`mlx/array.cpp:299-316`):

```cpp
ad.inputs.clear();                                    // :299  unconditional
for (auto& [_, a] : input_map) {
  bool is_deletable =
      (a.array_desc_.use_count() <= a.siblings().size() + 1);   // :301-302
  for (auto& s : a.siblings()) { ... }
  if (is_deletable) {
    for_deletion.push_back(std::move(a.array_desc_));
  }
}
```

and the cycle-breaking `siblings.clear()` runs only inside the `for_deletion` drain
(`mlx/array.cpp:328-334`).

The failure is the combination of those two lines:

1. `ad.inputs.clear()` at `:299` runs **unconditionally**, so the parent releases its reference to
   the group before deletability is decided.
2. If `use_count()` at `:301` exceeds the threshold — the count is read without synchronization,
   and the copy in `input_map` plus any transient reference elsewhere is enough — the group is
   skipped, never enters `for_deletion`, and therefore never reaches `siblings.clear()`.
3. The group is now referenced by nothing but itself, and it still holds its own `inputs`, which
   is what pins the upstream weights.

A single-output node cannot hit this: with `siblings().size() == 0` the threshold is
`use_count() <= 1`, and a node rejected there is merely deferred — whoever holds the extra
reference will release it later and `~ArrayDesc` will run normally. Only a sibling group can be
orphaned in a state where nothing will ever come back for it.

Compilation is required because the compile-cache tape is what keeps the traced group alive past
the forward pass; without it the graph is transient and torn down while its parents still hold it.
**(Marked as inference: the precise reference that inflates the count at `:301` was not isolated —
what is measured is that the cycle survives, is unreachable, and is rooted at `make_arrays`.)**

## Suggested fix

Break the sibling cycle unconditionally rather than only on the deletable path — i.e. clear
`siblings` for every entry in `input_map` that the parent has just released, not only for those
that pass the `use_count()` test. Alternatively, hold siblings as `weak_ptr`, which removes the
cycle at the source.

## Impact

This affects any long-lived process that builds and drops compiled models: the leak is the full
captured parameter set per drop, permanent, and unbounded. In our case (a 210 MiB speech model
behind `mx.compile`) a repeating build/drop cycle reached 1.18 GiB retained after 100 iterations
and was still climbing linearly.

## Confirmation from the C++ side

The same three ingredients, and the same fix, reproduce through `mlx-rs` on a real 210 MiB model
(MossFormer2_SE). There the race is intermittent rather than deterministic, which makes the
ablations statistically rather than trivially conclusive:

| change | n (drops) | leaking | rate |
|---|---|---|---|
| baseline (`split` in the compiled closure) | 4000 | 234 | 5.85% |
| **`split` → two single-output slices** | 4000 | **0** | **0.00%** |
| baseline, second model co-resident | 600 | 37 | 6.17% |
| **slices, second model co-resident** | 600 | **0** | **0.00%** |
| weights never evaluated outside the compiled fn | 4600 | 0 | 0.00% |
| weights evaluated by a bare `eval` (no forward) | 2000 | 103 | 5.15% |

Fisher p = 1.1 × 10⁻⁷² and 4.1 × 10⁻¹² for the two slice ablations. Note the last two rows: it is
**materialization** of the captured arrays that arms the bug, not the shape of whatever computation
did the materializing — a bare `mx.eval` is sufficient.

Replacing the multi-output op cost nothing measurable (median iteration 138 ms vs 139 ms) and left
every output bit-identical.

Also, unrelated to the leak but noticed alongside it: a second compiled *shape* entry for the same
function costs a full extra copy of the captured parameters in live memory (638.6 MiB vs 427.4 MiB
for one shape, same model). That looks like intended tape behaviour rather than a defect, but it is
worth knowing before compiling several shapes per model.

## AI Disclosure

This issue was identified by Claude Opus 5 as the root cause of a memory leak in a downstream project that depends on MLX. We have reworked our code to avoid this issue, and I'm submitting the report Claude wrote for your information and any follow-up you might want to take. Thanks for your work on this library!

AKnassa:
## Proposed changes

Fixes #3932.

A compiled function containing a multi-output primitive leaks every array the
traced function captured, once per call:

```python
def make_forward(weights):
    def forward(x):
        for w in weights:
            a, b = mx.split(x @ w, 2, axis=-1)
            x = mx.concatenate([b, a], axis=-1)
        return x
    return forward

# 8 x (256, 256) float32 weights, evaluated outside the compiled function
# before: 2097152 bytes retained per call, forever
# after:  0
```

Replacing the `mx.split` with plain slicing (a single-output op) leaks nothing,
which is what pointed at the multi-output path.

### Cause

`compile_dfs` deep copies the traced tape, so the originals of any multi-output
node should die once the new tape and the parents map are in place. They cannot:

- The outputs of a multi-output primitive hold each other through `siblings`
  (`array::make_arrays`), which is a strong reference cycle by construction.
- `array::~array()` is the only thing that breaks that cycle, and only when the
  destructing array is the last external holder (`use_count() == n + 1`).
- At `tape = std::move(new_tape)` the old arrays are destroyed, but the parents
  map still holds copies of them, so the count is too high and nothing is
  detached.
- The parents map then drops its references through `array::operator=`, which
  reseats `array_desc_` and has no cycle-breaking counterpart.

The group ends up orphaned at `use_count() == 1` each, referenced only by each
other, still holding `ArrayDesc::inputs` — which pins the captured weights.

That also explains why all three ingredients are needed to see it: single-output
originals have no cycle, `compile_dfs` is the only place doing this deep copy
plus `operator=` rewiring, and an already-evaluated capture is shared into the
new tape rather than copied, so the orphan holds a real buffer.

### Fix

Rather than adding a new liveness rule, keep one member of each replaced group
alive until both the tape and the parents map have been rewired, then drop it,
so the existing destructor test runs at a point where its arithmetic is valid.

This is deliberately conservative. If the caller still holds the group — for
example a lazy `mx.split` created before tracing and captured by the closure —
the reference count stays above the threshold and nothing is mutated at all. I
checked that case explicitly against `main` and against this branch, and both
behave identically. Nothing is force-detached, so the compile cache's record of
each node's primitive is untouched.

Also worth a line: `compile_simplify`'s `merge_one` rewires parents through the
same `array::operator=` and `split_one` clones nodes, so a similar shape may
exist there. I have not investigated it and am not claiming simplify is clean.

### Testing

New `test_multi_output_leak` in `python/tests/test_compile.py`, which loops so a
single-cycle fluke cannot pass it. It fails before this change and passes after.
The existing `test_leaks` gates on `mx.metal.is_available()`, so it is a no-op
on a CPU-only build; this one does not.

- Full Python suite: 820 tests, no new failures.
- Full C++ suite: 247 cases, all pass.

Built CPU-only (`MLX_BUILD_METAL=OFF`); the GPU suite was not run on my machine.

## Checklist

- [ ] I have updated the necessary documentation (if needed) — no documentation change needed, this is an internal lifetime fix

### DIFF
--- a/mlx/compile.cpp
+++ b/mlx/compile.cpp
@@ -469,6 +469,9 @@ std::pair<std::vector<array>, ParentsMap> compile_dfs(
 
   // Deep copy the tape and parents map while preserving inputs and outputs
   std::vector<array> new_tape;
+  // One member of each multi-output group that gets replaced below, see the
+  // comment where this is cleared
+  std::vector<array> replaced_multi_outputs;
   std::unordered_set<uintptr_t> io_set;
   std::unordered_map<uintptr_t, array> old_to_new;
   for (auto& o : outputs) {
@@ -510,6 +513,7 @@ std::pair<std::vector<array>, ParentsMap> compile_dfs(
         old_to_new.insert({out[i].id(), as[i]});
       }
       new_tape.push_back(as[arr.sibling_position()]);
+      replaced_multi_outputs.push_back(arr);
     } else {
       auto a = array(
           arr.shape(), arr.dtype(), arr.primitive_ptr(), std::move(inputs));
@@ -543,6 +547,15 @@ std::pair<std::vector<array>, ParentsMap> compile_dfs(
     new_parents_map[old_to_new.find(id)->second.id()] = std::move(vec);
   }
   parents_map = std::move(new_parents_map);
+
+  // The replaced multi-output arrays reference each other through their
+  // siblings, so they only go away when array::~array() breaks that cycle, and
+  // it does so only for the last external reference. Both the tape and the
+  // parents map still held them until just above, so drop them here instead.
+  // Any group the caller still holds is left alone, since the reference count
+  // stays above the threshold and nothing is detached.
+  replaced_multi_outputs.clear();
+
   return {tape, parents_map};
 }
 
--- a/python/tests/test_compile.py
+++ b/python/tests/test_compile.py
@@ -1244,6 +1244,41 @@ class TestCompile(mlx_tests.MLXTestCase):
 
         self.assertEqual(mem_pre, mem_post)
 
+    def test_multi_output_leak(self):
+        # A multi-output primitive inside a compiled function used to leave the
+        # traced arrays orphaned in their sibling reference cycle, which pinned
+        # every array they had captured.
+        dim, blocks = 128, 4
+        x = mx.random.normal((2, dim))
+        mx.eval(x)
+
+        def run():
+            weights = [mx.random.normal((dim, dim)) for _ in range(blocks)]
+            mx.eval(weights)
+
+            def forward(y):
+                for w in weights:
+                    a, b = mx.split(y @ w, 2, axis=-1)
+                    y = mx.concatenate([b, a], axis=-1)
+                return y
+
+            fn = mx.compile(forward)
+            mx.eval(fn(x))
+            del fn, forward, weights
+
+        # Warm up so the allocator is not growing for unrelated reasons
+        run()
+        gc.collect()
+        mx.synchronize()
+        mem_pre = mx.get_active_memory()
+
+        for _ in range(10):
+            run()
+            gc.collect()
+        mx.synchronize()
+
+        self.assertEqual(mx.get_active_memory(), mem_pre)
+
     def test_double_constant(self):
         with mx.stream(mx.cpu):
             x = mx.array(1.0, dtype=mx.float64)


zcbenz (changes requested):
This holds the arrays too long which increase memory usage a lot.

Issue #4161, opened 2026-08-10
[BUG] `mx.tan` (and `**`, `arctan`) on float64 give different results depending on ARRAY LENGTH and LANE INDEX


kyrollosyanny:
**Describe the bug**
`mx.tan` applied to a float64 array returns a value that depends on **how many elements the array has** and
on **which lane** the element sits in, not only on the element's value. It looks like a vectorised body
plus a scalar tail that disagree by 1 ULP.

Concretely, for `x = 0.6719517620178169` (= 38.5 degrees in radians):

```
tan(x) = 0x1.974360604de0bp-1    when N <= 3, and for the TAIL lanes of a longer array
         0x1.974360604de0cp-1    for the vectorised body                    (1 ULP higher)
```

At `N = 5`, lanes 0–3 return `...e0c` and lane 4 returns `...e0b`  **both values in one array**.

**To Reproduce**

```python
import mlx.core as mx, numpy as np
x = 0.6719517620178169
with mx.stream(mx.cpu):
    for n in (1, 2, 3, 4, 5, 6, 8, 16, 1000):
        v = np.asarray(mx.tan(mx.array(np.full((n,), x), dtype=mx.float64)))
        print(f"N={n:5d}  lane0={float(v[0]).hex()}  distinct values in this array: "
              f"{sorted({float(z).hex() for z in v})}")
```
```
N=    1  lane0=0x1.974360604de0bp-1  distinct values in this array: ['0x1.974360604de0bp-1']
N=    2  lane0=0x1.974360604de0bp-1  distinct values in this array: ['0x1.974360604de0bp-1']
N=    3  lane0=0x1.974360604de0bp-1  distinct values in this array: ['0x1.974360604de0bp-1']
N=    4  lane0=0x1.974360604de0cp-1  distinct values in this array: ['0x1.974360604de0cp-1']
N=    5  lane0=0x1.974360604de0cp-1  distinct values in this array: ['0x1.974360604de0bp-1', '0x1.974360604de0cp-1']
N=    6  lane0=0x1.974360604de0cp-1  distinct values in this array: ['0x1.974360604de0bp-1', '0x1.974360604de0cp-1']
N=    8  lane0=0x1.974360604de0cp-1  distinct values in this array: ['0x1.974360604de0cp-1']
...
```

Replicating each of many random values to N ∈ {1,2,3,4,5,6,7,8,9,16,17,64,1000} and collecting the distinct
results:

[...]

PR #4187, opened 2026-08-11
Fix elementwise unary ops giving different results for strided vs contiguous inputs (CPU) (#4163)

### LINKED ISSUE #4163: [BUG] `mx.tan(a[:, 0])` differs from `mx.tan(contiguous_copy)` on float64 arrays
**Describe the bug**
Taking a **strided (non-contiguous) view** of a float64 array and applying `mx.tan` gives a different
result by ~1 ULP  than applying it to a contiguous array holding the same values. So an upstream slice
such as `d[:, 0]` silently changes the arithmetic. Not sure if this would be considered a duplicate of #4161 

**To Reproduce**

```python
rng = np.random.default_rng(0)
bad = 0
with mx.stream(mx.cpu):
    for _ in range(300):
        v = rng.uniform(-1.5, 1.5, 8)
        M = mx.array(np.stack([v, v], 1), dtype=mx.float64)      # (8, 2)
        strided    = M[:, 0]                                     # a non-contiguous view
        contiguous = mx.array(np.ascontiguousarray(v), dtype=mx.float64)
        t1, t2 = mx.tan(strided), mx.tan(contiguous)
        mx.eval(t1, t2)
        if not np.array_equal(np.asarray(t1), np.asarray(t2)):
            bad += 1
print(f"mx.tan(strided) != mx.tan(contiguous) on {bad}/300 arrays")
```
```
mx.tan(strided) != mx.tan(contiguous) on 296/300 arrays
```

`x ** 2` shows the same behaviour on ~256/300.

**Expected behavior**
`f(view)` and `f(copy_of_view)` should be bitwise identical for an elementwise `f`. Memory layout is not
part of the mathematical input.

AxelNoun:
## Summary
- On Apple Silicon, CPU unary ops used Accelerate SIMD for contiguous inputs and scalar libm for strided views (and for the contiguous residual), so the same values could differ by ~1 ULP depending on memory layout / lane index.
- Route strided loads and the contiguous residual through the same padded SIMD kernel so results are bit-identical to the contiguous fast path.
- Commits are split: (1) strided gather, (2) contiguous residual, (3) regression tests so the residual-only fix remains submittable if bit-identity across layouts is considered out of scope (NumPy does not guarantee it either).

Fixes #4163.

Possibly related to #4161: this PR fixes the **unary** lane-index / length dependence shown there (	an, rctan, etc.). Binary cases in that issue (x**2, x**0.5, …) share the same Accelerate-vs-libm pattern in inary.h and are **not** fixed here happy to follow up the same way in inary.h/	ernary.h if this approach is accepted. Leaving #4161 open for the binary remainder.

## Validation (no local Mac GitHub Actions macos-26 / Accelerate)

| | Run | Result |
|---|---|---|
| Before (main) | [baseline](https://github.com/AxelNoun/mlx/actions/runs/31515581697) | **296/300** mismatches (issue repro) |
| After (this fix) | [fix](https://github.com/AxelNoun/mlx/actions/runs/31515577187) | **0/300**, position-independence OK, **full** `python -m unittest discover python/tests` green |

## Micro-benchmark (strided **before vs after**, float64, n=1e6, 50 iters, same runner)

The useful comparison is strided-before vs strided-after (not strided vs contiguous). Trivial ops included because gather overhead is most visible there:

| Op (strided) | cols | Before (main) | After (fix) |
|---|---|---|---|
| `abs` | 2 | 14.14 ms | **7.91 ms** |
| `negative` | 2 | 11.86 ms | **3.23 ms** |
| `tan` | 2 | 20.03 ms | **9.29 ms** |
| `abs` | 8 | 11.01 ms | **8.37 ms** |
| `negative` | 8 | 10.52 ms | **3.38 ms** |
| `tan` | 8 | 17.28 ms | **9.06 ms** |

No regression observed: SIMD+gather is faster than the previous pure-scalar strided path for both trivial and expensive ops on this runner.

## Test plan
- [x] Issue #4163 repro script: 296/300 on main → 0/300 with fix (CI links above)
- [x] Position-independence check from #4161 (unary `tan`)
- [x] Full `python -m unittest discover python/tests` on macos-26 CPU/Accelerate
- [x] `pre-commit` (clang-format / black) on touched files

### DIFF
--- a/mlx/backend/cpu/unary.h
+++ b/mlx/backend/cpu/unary.h
@@ -2,6 +2,8 @@
 
 #pragma once
 
+#include <algorithm>
+
 #include "mlx/backend/common/unary.h"
 #include "mlx/backend/cpu/encoder.h"
 #include "mlx/backend/cpu/simd/simd.h"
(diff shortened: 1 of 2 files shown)


zcbenz:
Thanks for the PR but we plan to switch to use google/highway for SIMD and at the moment (https://github.com/ml-explore/mlx/pull/3019), so we don't accept non-critical changes on SIMD code.

PR #4193, opened 2026-08-12
Jaccl ring multi wire fix

angeloskath:
Fixes a bug where if the size per wire is not a multiple of the rdma buffer we overwrite some data in the chunk that is handled by the other wire.

It finally showed up when I was testing `mlx_lm.share` over multi wire rings.


zcbenz (approved):
👍

PR #4196, opened 2026-08-12
Make CPU JIT-generated code consistently C++20

PhysicistJohn:
## Proposed changes

The CPU JIT and generated preamble still select C++17 on MSVC and GNU-like
toolchains even though MLX requires C++20 for CMake consumers. Use C++20
consistently across those runtime compilation paths and add a regression test
that compiles and links a C++20 concept through the real CPU JIT. AppleClang
compiler selection and the public API remain unchanged.

## Checklist

- [x] I have run the relevant `clang-format --dry-run --Werror`, shell syntax, and diff checks
- [x] I have added a focused regression test; it passes 1/1 case and 4/4 assertions
- [x] I have run the native suite (248/248 cases, 3,335/3,335 assertions) and CTest (249/249)
- [ ] Documentation update not needed; this does not change a public API

### DIFF
--- a/mlx/backend/cpu/jit_compiler.cpp
+++ b/mlx/backend/cpu/jit_compiler.cpp
@@ -248,7 +248,7 @@ std::string JitCompiler::build_command(
   return fmt::format(
       "\""
       "cd /D \"{0}\" && "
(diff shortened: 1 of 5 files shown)


zcbenz:
For the moment we want to target C++17 in JIT compilation since C++20 is still not so widely available especially in servers.

PR #4199, opened 2026-08-12
Preserve components in complex-like conversions

PhysicistJohn:
## Proposed changes

Complex-like inputs that also expose a scalar conversion currently select the scalar constructor, replacing their real component with the scalar value and dropping the imaginary component. Construct those inputs from `real()` and `imag()` instead, keep scalar and complex-like constructor paths mutually exclusive, and reject complex-like inputs whose components cannot be converted. Existing conversion traits and wrapper layout remain unchanged.

The focused regression passes 12/12 assertions. The full native suite passes 249/249 tests with 3,354 assertions, and CTest passes 250/250.

## Checklist

### DIFF
--- a/mlx/types/complex.h
+++ b/mlx/types/complex.h
@@ -2,12 +2,31 @@
 
 #pragma once
 #include <complex>
+#include <concepts>
+#include <type_traits>
+
 #include "mlx/types/half_types.h"
 
 namespace mlx::core {
 
 struct complex64_t;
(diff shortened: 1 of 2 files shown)


zcbenz:
`requires` requires C++20 which we don't want to enforce for JIT compilation yet, and while it is nice to support complex like structures I think it is not worthwhile at the moment.

PR #4202, opened 2026-08-12
Re-disable qmm_n_nax and fix group_size < 64

### LINKED ISSUE #4201: quantized_matmul(transpose=False) wrong results on Metal (gs 32/64, bits 2/4/8)
### Summary
On current `main` (`31b5cbb09`), the non-transposed path of `mx.quantized_matmul` produces garbage on Metal. `transpose=True` is fine; `transpose=False` fails for `group_size` 32 and 64 and bits 2/4/8. This also shows up as 48 failing subtests in `TestQuantized::test_qmm`.

### Repro (M5 Max, MLX built from main, Metal backend)
```python
import mlx.core as mx
import numpy as np

mx.random.seed(1)
M, N, K = 8, 128, 128
x = mx.random.normal((M, K)).astype(mx.float16)
for bits in (4, 8):
    for gs in (32, 64):
        wq, s, b = mx.quantize(w, group_size=gs, bits=bits)
        refq = mx.dequantize(wq, s, b, group_size=gs, bits=bits)
        y_ref = (x.astype(mx.float32) @ refq.astype(mx.float32).T)
        for tp in (False, True):
            y_q = mx.quantized_matmul(x, wq, s, b, transpose=tp, group_size=gs, bits=bits)
            mx.eval(y_q, y_ref)
            err = np.abs(np.array(y_ref) - np.array(y_q)).max()
            print(f"bits={bits} gs={gs} transpose={tp}: max_err={err:.2e}")
```
Output:
```
bits=4 gs=32 transpose=False: max_err=5.90e+02
bits=4 gs=32 transpose=True : max_err=2.92e-02
bits=4 gs=64 transpose=False: max_err=6.79e+01
bits=4 gs=64 transpose=True : max_err=2.87e-02
bits=8 gs=32 transpose=False: max_err=8.95e+03
bits=8 gs=32 transpose=True : max_err=1.87e-02
bits=8 gs=64 transpose=False: max_err=5.07e+01
bits=8 gs=64 transpose=True : max_err=1.61e-02
```

### Notes / hypothesis
- Native an already `transpose=True` path is correct, so this is scoped to the non-transposed quantized matmul Metal implementation (the path reworked by #4051). It is not caused by the open #3922 fix (verified by running the same suite with and without it).
- `TestQuantized::test_qmm` FAILED/SUBFAILED shapes: {M∈(8,33,64)} x {N∈(128,256)} x {K∈(128,256)}, bits 2/4/8, group_size 32, transpose=False (same class at gs=64).

Happy to dig into the kernel (local Metal build available) and bring a fix PR.

PhilipJohnBasile:
Fixes #4201. Root-cause fix (supersedes the earlier workaround in this PR).

## Root cause
The `group_size=32` `QuantizedBlockLoader` specialization
(`quantized_nax.h`) advanced the scale/bias pointer by
`n_groups * group_stride` when stepping between reduction tiles. For the
non-transposed NAX kernel (#4051) the scale plane is `[K, N / group_size]`,
so a 64-row reduction tile is `group_stride = 64 * N / group_size` scale
entries — advancing by `n_groups *` that (4N for gs=32) reads out of bounds
from the second K-tile on. Result: `quantized_matmul(transpose=False)` with
`group_size=32` produced garbage (and NaN under some inputs) on Metal for any
`K > 64`. `K = 64` (a single reduction tile) was correct — which is why the
bug was localized only after comparing K=64 vs K=128.

`group_size` 64/128 and `transpose=True` were unaffected (they use the
generic loader / the `reduction_dim == 1` branch respectively).

The same over-advance existed in the fp (fp8/MX) loader and is fixed here too.

## Fix
Advance by `group_stride` (not `n_groups * group_stride`) in the
`reduction_dim == 0` branch of both loaders.

## Tests / verification
- New regression `test_qmm_non_transposed_group_size_lt_64`: gs=32, bits
  2/4/8, M 8/33/65, K 64 (single-tile control) / 128 / 256 (multi-tile).
  Fails on main at K>64 (garbage/NaN), passes with the fix.
- Full `test_quantized` on a local Metal build (M5 Max, macOS 27, Xcode 27,
  mlx from source): previously 48 failed subtests -> now **37 passed,
  3106 subtests passed**.
- Manual sweep matches fp64 reference (`~1e-2`) for all bits/gs, both
  transpose flags, and odd-M partial tiles.

### DIFF
--- a/mlx/backend/metal/kernels/fp_quantized_nax.h
+++ b/mlx/backend/metal/kernels/fp_quantized_nax.h
@@ -178,7 +178,10 @@ struct QuantizedBlockLoader {
     if (reduction_dim == 1) {
       scales += n_groups;
     } else {
-      scales += n_groups * group_stride;
+      // Advance from one BROWS-row reduction tile to the next in a
+      // [K, N / group_size] scale plane; see the identical fix in
+      // quantized_nax.h (QuantizedBlockLoader<..., 32, bits>).
+      scales += group_stride;
     }
   }
 };
--- a/mlx/backend/metal/kernels/fp_quantized_nax.h
+++ b/mlx/backend/metal/kernels/fp_quantized_nax.h
@@ -178,9 +178,6 @@ struct QuantizedBlockLoader {
     if (reduction_dim == 1) {
       scales += n_groups;
     } else {
-      // Advance from one BROWS-row reduction tile to the next in a
-      // [K, N / group_size] scale plane; see the identical fix in
-      // quantized_nax.h (QuantizedBlockLoader<..., 32, bits>).
       scales += group_stride;
     }
   }
--- a/mlx/backend/metal/kernels/quantized_nax.h
+++ b/mlx/backend/metal/kernels/quantized_nax.h
@@ -826,8 +826,14 @@ struct QuantizedBlockLoader<
       biases += n_groups;
       // }
     } else {
-      scales += n_groups * group_stride;
-      biases += n_groups * group_stride;
+      // Advance from one BROWS-row reduction tile to the next. The scale
+      // plane is [K, N / group_size], so the step is BROWS rows of N/gs
+      // entries = group_stride. Advancing by n_groups * group_stride skips
+      // past n_groups weight tiles of scales, reading out of bounds from the
+      // second reduction tile on (visible at group_size=32 where one 64-wide
+      // tile contains two 32-value groups).
+      scales += group_stride;
+      biases += group_stride;
     }
   }
 };
--- a/mlx/backend/metal/kernels/quantized_nax.h
+++ b/mlx/backend/metal/kernels/quantized_nax.h
@@ -826,12 +826,6 @@ struct QuantizedBlockLoader<
       biases += n_groups;
       // }
     } else {
-      // Advance from one BROWS-row reduction tile to the next. The scale
-      // plane is [K, N / group_size], so the step is BROWS rows of N/gs
-      // entries = group_stride. Advancing by n_groups * group_stride skips
-      // past n_groups weight tiles of scales, reading out of bounds from the
-      // second reduction tile on (visible at group_size=32 where one 64-wide
-      // tile contains two 32-value groups).
       scales += group_stride;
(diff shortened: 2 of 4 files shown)


zcbenz (approved):
Thanks for the fix, I should have ran the full test suite when testing the PR 🤦 

So apparently the whole qmm_n is still WIP and I'm re-disabling it until a full review and benchmarking.

gordofreemo:
Flagging a footgun this left behind, since #4051 was mine.

After this PR the guard is:
```cpp
bool has_nax_kernel =
    metal::is_nax_available() && (transpose || mode == "affine");
if (has_nax_kernel && transpose && (K % 64 == 0) && ...
```
`has_nax_kernel && transpose` makes the `(transpose || mode == "affine")` disjunct unreachable, so it's really just `is_nax_available() && transpose`. What worries me is that the `N % 64 == 0` term is gone while the variable now reads as "a NAX kernel exists for this config" so the natural way to re-enable later is to drop the && transpose, and that hands you the non-transposed path with no N constraint.

It needs one. `qmm_n_nax_tgp_impl` loads a 64×64 weight tile via `load_unsafe()`, and the store is clamped to the simdgroup tile rather than the matrix  `Dtile.store_safe(y + tm * N + tn, N, short2(SN, sgp_sm))` passes `lim_y = SN = 32`. So for `N % 64 != 0` the last N-tile reads w/scales past the row end and writes y out of bounds. N is a multiple of group_size for any validly quantized w, so this only breaks at `group_size == 32`.

Folding the precondition into the variable makes the name true and keeps it from re-arming, without changing behaviour:
```cpp
bool has_nax_kernel = metal::is_nax_available() &&
    (transpose || (mode == "affine" && N % 64 == 0));
if (has_nax_kernel && transpose && (K % 64 == 0) && ...
```
Unrelated but adjacent: `fp_qmm_n_impl` still has the [N, K] weight addressing and the missing M-tile bounds that #4051 fixed on the affine side. This is unreachable today for the same reason the affine one was but worth knowing before that guard widens. I can open a separate issue.

PR #4206, opened 2026-08-12
Require pytorch 2.12 for metal dlpack tests

zcbenz:
The tests added by #3531 assumes pytorch 2.12 which supports sharing metal buffers between other frameworks, older versions of pytorch still work it is just we could not do non-copy data sharing. It does not really worth the efforts to refactor the tests to support old versions of pytorch so just skip them.


angeloskath (approved):
👍

Issue #3442, opened 2026-04-22
init(backend="any") selects ring singleton instead of JACCL when launched via mlx.launch --backend jaccl


machiabeli:
## Observed Behavior

When using `mlx.launch --backend jaccl --hostfile hosts.json -- python3 -m mlx_lm.server ...`, the server process calls `mx.distributed.init()` (which defaults to `backend="any"`). The init function tries ring before jaccl, and ring creates a singleton group of size 1 because `MLX_RANK` is set but `MLX_HOSTFILE` is not. JACCL is never attempted.

The result is that distributed inference silently falls back to single-node, even though JACCL env vars (`MLX_JACCL_COORDINATOR`, `MLX_IBV_DEVICES`) are correctly set by the launcher.

## Expected Behavior

When `mlx.launch --backend jaccl` sets JACCL-specific env vars, `init(backend="any")` should initialize JACCL, not ring.

## Steps to Reproduce

```bash
mlx.launch --backend jaccl --hostfile hosts.json -- \
  python3 -c "import mlx.core as mx; g = mx.distributed.init(); print(f'rank={g.rank()} size={g.size()}')"

# Expected: size=N (number of nodes in hostfile)
# Actual: size=1 on each rank (ring singleton)
```

## Root Cause

In `distributed.cpp`, `init(backend="any")` tries backends in order: nccl -> ring -> mpi -> jaccl. On Mac, `ring::init(false)` sees `MLX_RANK` (set by `launch_jaccl`) but no `MLX_HOSTFILE`, creates a size-1 group, and returns non-null. The init function treats this as success and never tries JACCL.

## Possible Approaches

1. **Env var hint from launcher**: `launch_jaccl()` sets `MLX_DISTRIBUTED_BACKEND=jaccl`, and `init(backend="any")` checks this env var first before the default try order.

2. **Reorder try order**: Try jaccl before ring when JACCL env vars are present.

3. **Fix ring singleton**: `ring::init()` could return nullptr when `MLX_HOSTFILE` is not set, instead of creating a singleton.

Each has trade-offs. Would love to hear which approach (if any) the team prefers.

## Workaround

We have been using a Python import hook (`_mlx_backend_fix.py` + `.pth` file) on all cluster nodes since March 2026 that patches `mx.distributed.init` to read `MLX_DISTRIBUTED_BACKEND` from the environment. It works but must be redeployed after every `pip install --upgrade mlx`.

## Environment

[...]

PR #4210, opened 2026-08-12
Only use MPI backend for world size > 1

erwinzhang7:
Related to #3442.

`init(backend="any")` tries backends in order and takes the first that initializes. Open MPI
initializes successfully for a program that was not started with `mpirun`, giving an
`MPI_COMM_WORLD` of size 1. That counted as success, so the chain stopped at mpi and never
reached the backends after it. A correctly configured JACCL job would then run single node and
report nothing.

`mpi::init` now returns `nullptr` when the world holds a single process, so `any` carries on to
the next backend. Asking for it directly with `backend="mpi"` and `strict=true` raises instead:

```
[mpi] The world has a single process. Launch with mpirun to initialize the mpi backend.
```

The size comes from `mpi().world()` rather than from a constructed `MPIGroup`, because building
one and discarding it runs `~MPIGroup()`, which calls `finalize_safe()` when `global_` is set,
and `MPI_Init` cannot be called again after that.

The second hunk is needed to make the first safe. `register_group` only caches under `"any"` for
a non-null group, so once `mpi::init` can decline, nothing is cached and the next `init()` call
re-runs the whole chain. `init_safe()` called `MPI_Init` unconditionally, and Open MPI treats a
second `MPI_Init` as fatal:

```
Open MPI has detected that this process has attempted to initialize
MPI (via MPI_INIT or MPI_INIT_THREAD) more than once.  This is erroneous.
*** MPI_ERRORS_ARE_FATAL
```

That is unreachable on main today, because a successful mpi group is cached and `init_safe` never
runs twice. `init_safe` is idempotent now.

#### Verified

Open MPI 5.0.9, macOS 26.6.

| case | result |
|---|---|
| libmpi present, no mpirun, `init()` | mpi not selected |
| libmpi present, no mpirun, `init(strict=true)` | `[distributed] Couldn't initialize any backend` |
| libmpi present, `init(backend="mpi", strict=true)` | raises the message above |
| libmpi present, `init()` twice | no crash |
| `mpirun -n 2`, `init()` | size 2, ranks 0 and 1 |
| `mpirun -n 2`, `init(backend="mpi", strict=true)` | size 2, no raise |
| `mpirun -n 2`, `all_sum` | `[3.0, 3.0, 3.0, 3.0]` |

The first two rows match running with `MLX_MPI_LIBNAME` pointed at a missing file, which is the
intent: with Open MPI installed but no `mpirun`, MLX behaves as though it were not there.

### DIFF
--- a/mlx/distributed/mpi/mpi.cpp
+++ b/mlx/distributed/mpi/mpi.cpp
@@ -166,6 +166,11 @@ struct MPIWrapper {
     if (!is_available()) {
       return false;
     }
+    // MPI_Init is an error to call twice, and init() can run more than once
+    // when it returns without a group.
+    if (initialized_) {
+      return true;
+    }
     bool success = init(nullptr, nullptr) == MPI_SUCCESS;
 
     // Initialize custom types and ops
@@ -491,6 +496,19 @@ std::shared_ptr<GroupImpl> init(bool strict /* = false */) {
   if (!mpi().init_safe()) {
     if (strict) {
       throw std::runtime_error("Cannot initialize MPI");
+    }
+    return nullptr;
+  }
+
+  // Open MPI initializes a world of size 1 for a program that was not started
+  // with mpirun, which is not a distributed group.
+  int size = 1;
+  mpi().size(mpi().world(), &size);
+  if (size <= 1) {
+    if (strict) {
+      throw std::runtime_error(
+          "[mpi] The world has a single process. Launch with mpirun to "
+          "initialize the mpi backend.");
     }
     return nullptr;
   }


angeloskath (dismissed):
I don't think that this is the correct fix. I would instead only return successful MPI initialization if the world size is larger than 1.

erwinzhang7:
Reworked: `mpi::init` now returns `nullptr` when the
world has a single process, and the warning is gone.

One thing falls out of it. Returning `nullptr` there made a latent crash reachable:
`register_group` only caches under `"any"` for a non-null group, so nothing is cached, the next
`init()` re-runs the chain, and `init_safe()` called `MPI_Init` a second time, which Open MPI
treats as fatal. `init_safe()` is idempotent now, which is the second hunk.

Checked with Open MPI 5.0.9: without `mpirun`, `init()` no longer selects mpi and behaves the
same as with libmpi absent; `init(backend="mpi", strict=True)` raises; under `mpirun -n 2`,
`init()` still gives size 2 and `all_sum` is correct.

zcbenz (approved):

PR #4211, opened 2026-08-12
Break monolithic MTLResidencySet into smaller sets

JamesMcCarthy44:
Fix for performance issues observed when running large LLM models on MacOS 27. Issue was tracked to non-related GPU operations (desktop update etc) causing the entire model to be continuously re-wired to GPU memory.

MacOS 27 appears to make GPU residency decisions per `MTL::ResidencySet`, so with everything wired in one large set, losing residency once means the whole working set has to be made resident again — decode can't continue until it is. This spreads the wired allocations across several size-capped sets, so one such event costs one set instead of everything.

`set_wired_limit` is unchanged: still the total wired budget, still `0` by default. Only the distribution across sets changes, never how many bytes are wired. No public API or documented behaviour change.

`MLX_RESIDENCY_SET_MAX_PCT` caps each set as a percentage of `recommendedMaxWorkingSetSize` (default `5`); `<= 0` or `>= 100` gives a single set, the previous layout. The set count is capped at 32, the per-command-queue limit from the Metal feature set tables.

`tests/residency_tests.cpp` adds 10 cases for the wired-limit contract and the set behaviour; the suite passes with `MLX_RESIDENCY_SET_MAX_PCT` at 5, 1, 2, 0 and 200.

Decode Performance Before -> After as measured on M3 Ultra (512Gb)
0.46 -> 13.19 tokens/sec - (mlx-community/GLM-5.1-MXFP4-Q8)
0.39 -> 21.96 tokens/sec - (mlx-community/Kimi-K2.6-mlx-DQ3_K_M-q8) 
0.71 -> 24.48 tokens/sec - (mlx-community/MiniMax-M2.7)

## Checklist

### DIFF
--- a/mlx/backend/metal/resident.cpp
+++ b/mlx/backend/metal/resident.cpp
@@ -1,6 +1,7 @@
 // Copyright © 2024 Apple Inc.
 
 #include <algorithm>
+#include <cassert>
 #include <cstdio>
 #include <sstream>
 #include <stdexcept>
@@ -141,7 +142,8 @@ void ResidencySets::insert(MTL::Allocation* buf) {
 
   auto [it, inserted] = buf_to_set_.try_emplace(buf, Placement{kNoSet, bytes});
   if (!inserted) {
-    return; // already tracked
+    assert(false && "allocation is already tracked");
+    return;
   }
   // Stay within the wired limit. The excess is tracked but left out of any
   // set, and is added to one by resize() if the limit is raised later.
@@ -159,7 +161,8 @@ void ResidencySets::erase(MTL::Allocation* buf) {
   std::lock_guard<std::mutex> lk(mtx_);
   auto it = buf_to_set_.find(buf);
   if (it == buf_to_set_.end()) {
-    return; // a heap child, or never inserted
+    assert(false && "erasing an allocation that was never inserted");
+    return;
   }
   if (it->second.set_id != kNoSet) {
     const uint32_t idx = it->second.set_id;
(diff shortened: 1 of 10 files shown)


mlx/backend/metal/resident.cpp
```diff
@@ -1,95 +1,236 @@
...
-void ResidencySet::resize(size_t size) {
-  if (!wired_set_) {
+void ResidencySets::erase(MTL::Allocation* buf) {
+  if (!enabled_) {
     return;
   }
+  std::lock_guard<std::mutex> lk(mtx_);
+  auto it = buf_to_set_.find(buf);
+  if (it == buf_to_set_.end()) {
+    return; // a heap child, or never inserted
```
zcbenz: It does not seem that this would ever happen, maybe add a debug time assertion like `assert(false)` for this and the failed insertion in `insert`?

zcbenz (approved):
Could not really find an issue with the implementation, solid and awesome work!

mlx/backend/metal/resident.cpp
JamesMcCarthy44: Thanks for the feedback, you're quite right there should be asserts there.  Amended the branch.

JamesMcCarthy44:
It's probably also worth saying that it's not just 'large' models that can hit this, it's any model that gets us close to `iogpu.wired_limit_mb`. So 'large' is relative to the amount of GPU memory you have!
 The change was tested against 30 of the most popular models to check for any performance regressions caused by the overhead of the additional residency sets and none was observed.

PR #4212, opened 2026-08-12
Bound GGUF metadata string/array values against the file mapping

### LINKED ISSUE #4136: OOB read in the GGUF loader: tensor data offset and size are not bounded against the file mapping
Loading a crafted GGUF file makes the loader read outside the mapped file. Filing this publicly as a robustness issue, as suggested when the advisory was closed.

The vendored gguflib builds a tensor's data pointer from fields read straight out of the file. In `gguf_get_tensor()`:

```c
tensor->offset = ctx->data_off + *offset;
tensor->weights_data = ctx->data + tensor->offset;
```

`*offset` and `bsize` come from the file and are never checked against `ctx->size`. A crafted tensor places `weights_data` outside the mapping, and any later read of `bsize` bytes reads out of bounds. MLX reaches this through `extract_tensor_data`.

Control versus crafted under AddressSanitizer, only the offset field differs:

- offset 0: loads normally, exits clean
- offset 1<<40: SEGV inside `__asan_memcpy` on the tensor read

Re-checked on main on 2026-07-17, after #3436 force-enabled the gguflib asserts in release builds. Those asserts cover the metadata and tensor-info walk. They do not cover the data region.

Related but not the same: CVE-2025-62609 added a `weights_data == nullptr` check in `load_gguf()`. That does not catch a non-null pointer landing outside the mapping.

Upstream, antirez/gguf-tools#33 proposes the bound, checking offset plus size against the mapping. An earlier PR there, #28, bounded the metadata walk only.

Happy to share the crafted file.

x14ngch3n:
## Summary

Bounds the length of `STRING` and `ARRAY` **metadata** KV values against the mmap'd file mapping in `load_gguf`, mirroring the existing `check_tensor_in_file()` guard that protects the tensor data path. A crafted GGUF file could otherwise force an out-of-bounds read past the file mapping.

## Problem

`gguf_get_key()` returns a pointer (`val`) into the mmap'd file but performs **no bounds checking**, and the metadata value lengths are read straight from the file. In `set_mx_value_from_gguf`, these attacker-controlled lengths were fed directly to a copy/string constructor with no check against the mapping size:

**`STRING` KV:**
```cpp
value = std::string(val->string.string, static_cast<int>(val->string.len));
```
`val->string.len` is a `uint64_t` from the file, narrowed to `int`, then `std::string(ptr, n)` memmoves `n` bytes out of the mmap.

**`ARRAY` KV:**
```cpp
auto size = static_cast<int>(val->array.len);
...
value = array(reinterpret_cast<uint32_t*>(data), {size}, uint32);
```
`array(T*, Shape, Dtype)` allocates `size*itemsize` and `std::copy`s `size` elements from the mmap — a pure source over-read.

The tensor path was already bounded by `check_tensor_in_file()` (added in #4179 / fixes #4136), but the **metadata path had no equivalent guard**.

### Impact

Out-of-bounds read past the mmap'd GGUF file when loading an untrusted `.gguf`. Confirmed with AddressSanitizer (4 MB read on a ~50-byte file, both sinks). In a non-ASAN process the read crosses into an unmapped page → SEGV (denial of service). No write primitive (the allocation is derived from the same `len`, so it is a source over-read, not an over-write).

Reachable from the default-config public Python API on every platform mlx ships:
```python
import mlx.core as mx
mx.load("evil.gguf", format="gguf")
```
`MLX_BUILD_GGUF` is `ON` by default.

## Fix

Adds `check_metadata_value_in_file(ctx, val, value_bytes)`, which verifies the value region lies within `[0, ctx->size)`. Each metadata branch now bounds its value length against the mapping before any copy/string construction:

- **STRING:** rejects `len` that doesn't fit in an `int` or exceeds `ctx->size`, then bounds `sizeof(gguf_string) + len`.
- **ARRAY:** computes `elt_size` from the declared element type and checks `arr_len * elt_size <= remaining` (overflow-safe division form) before the element loop.
- **STRING array elements:** each inner `str_val->len` is bounded individually.

The `static_cast<int>` narrowing is also guarded so a length above `INT_MAX` is rejected explicitly rather than wrapping to a negative `int`.

## Testing

Built with clang `-O1 -g -fsanitize=address -fno-omit-frame-pointer` (LLVM clang, since libmlx is built with it). Three crafted GGUFs (over-long STRING, over-long UINT32 ARRAY, and STRING with `len` > `INT_MAX`):

**Before** (on `main`), `STRING` sink:
```
==95370==ERROR: AddressSanitizer: unknown-crash on address 0x000103ab0000
READ of size 4194304 at 0x000103ab0000 thread T0
    #6 mlx::core::set_mx_value_from_gguf(...) gguf.cpp:128
    #7 mlx::core::load_metadata(gguf_ctx*) gguf.cpp:209
    #8 mlx::core::load_gguf(...) gguf.cpp:279
```

**After** (this branch): all three crafted files are rejected cleanly:
```
exception: [load_gguf] String metadata value length exceeds file size.
```
no ASAN violation, process exits 0/1 via exception.

**No regression:** a legitimate roundtrip (`save_gguf` of a small float tensor + a STRING and an INT32 ARRAY metadata value, then `load_gguf`) still loads correctly with the new bounds in place.

## Prior art

Distinct from #4136/#4179 (tensor data offset/bsize — tensor only), #3436 (gguflib `-UNDEBUG` asserts — mlx bypasses `gguf_do_with_value`), and CVE-2025-62609 (different sink). None of these bound the metadata KV value length.

### DIFF
--- a/mlx/io/gguf.cpp
+++ b/mlx/io/gguf.cpp
@@ -88,12 +88,34 @@ std::tuple<allocator::Buffer, Dtype> extract_tensor_data(gguf_tensor* tensor) {
   return {buffer, float16};
 }
 
-// Mirror check_tensor_in_file(): gguf_get_key() leaves key.val / ctx->off
-// pointing at the value but performs no bounds checking, and the value lengths
-// are read straight from the file. Bound the whole value (and, for arrays,
-// every element) against the mmap'd file once per key in load_metadata(),
-// before set_mx_value_from_gguf consumes it. Lengths that would not fit in the
-// int the downstream array() / std::string constructors take are rejected here.
+// gguf_get_key() leaves key.val / ctx->off pointing at the value but performs
+// no bounds checking, and the value lengths are read straight from the file.
+// Bound the whole value (and, for arrays, every element) against the mmap'd
+// file once per key in load_metadata(), before set_mx_value_from_gguf
+// consumes it. Array lengths are also capped at int max because the
+// downstream array() shape is int-valued.
+size_t gguf_value_type_size(uint32_t type) {
+  switch (type) {
+    case GGUF_VALUE_TYPE_BOOL:
+    case GGUF_VALUE_TYPE_UINT8:
+    case GGUF_VALUE_TYPE_INT8:
+      return 1;
+    case GGUF_VALUE_TYPE_UINT16:
+    case GGUF_VALUE_TYPE_INT16:
+      return 2;
+    case GGUF_VALUE_TYPE_UINT32:
+    case GGUF_VALUE_TYPE_INT32:
+    case GGUF_VALUE_TYPE_FLOAT32:
+      return 4;
+    case GGUF_VALUE_TYPE_UINT64:
+    case GGUF_VALUE_TYPE_INT64:
+    case GGUF_VALUE_TYPE_FLOAT64:
+      return 8;
+    default:
+      return 0;
+  }
+}
+
 void check_metadata_value_in_file(
     const gguf_ctx* ctx,
     uint32_t type,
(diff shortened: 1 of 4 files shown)


zcbenz (changes requested):
Can you do simple check in `load_metadata` like what #4179 did?

x14ngch3n:
Thanks @zcbenz. Done in the latest push: the bounds check now lives in a single `check_metadata_value_in_file(ctx, type, val)` call inside `load_metadata()` (one call per key, before `set_mx_value_from_gguf` consumes it), mirroring `check_tensor_in_file()` on the tensor path. `set_mx_value_from_gguf` is back to reading lengths straight from the file — the STRING/ARRAY validation (fixed scalars, length-prefixed strings, fixed-size arrays, and each element of a string array) is centralized in the validator. Added `"test gguf metadata value validation"` next to the tensor-offset test. All four `gguf` cases pass under ASAN (`-O1`).

mlx/io/gguf.cpp
```diff
@@ -87,6 +88,139 @@ std::tuple<allocator::Buffer, Dtype> extract_tensor_data(gguf_tensor* tensor) {
...
+          "[load_gguf] Metadata value extends past the end of the file. "
+          "Perhaps an incomplete download or corrupt file?");
+    }
+    return;
+  }
+
+  // gguf_string = { uint64_t len; char string[] }.
+  if (type == GGUF_VALUE_TYPE_STRING) {
+    if (sizeof(uint64_t) > avail(base) ||
+        val->string.len > static_cast<uint64_t>(std::numeric_limits<int>::max()) ||
```
zcbenz: feels meaningless to test int_max?

mlx/io/gguf.cpp
```diff
@@ -87,6 +88,139 @@ std::tuple<allocator::Buffer, Dtype> extract_tensor_data(gguf_tensor* tensor) {
...
+        elt_size = 2;
+        break;
+      case GGUF_VALUE_TYPE_UINT32:
+      case GGUF_VALUE_TYPE_INT32:
+      case GGUF_VALUE_TYPE_FLOAT32:
+        elt_size = 4;
+        break;
+      case GGUF_VALUE_TYPE_UINT64:
+      case GGUF_VALUE_TYPE_INT64:
+      case GGUF_VALUE_TYPE_FLOAT64:
```
zcbenz: can you deduplicate the code

x14ngch3n:
Addressed both comments in e9b3ac1:

- Extracted `gguf_value_type_size()` shared by the scalar and array-element paths, and folded the string validation (single + array walk) into one `check_string` lambda.
- The per-string `int` max checks were indeed redundant once the file-mapping bound is checked, so instead of defending the narrowing I removed it: `std::string` now takes the `uint64_t` length directly. The array-length `int` max check stays because the downstream `array()` shape is `int`-valued.

zcbenz (approved):

PR #4215, opened 2026-08-12
Fix MultiOptimizer on models containing empty modules

erenmenges:
## Description
MultiOptimizer raised ValueError on any model that had parameterless modules (e.g. ReLU/GELU). I hit this while writing a ternary transformer that included something like a MLP = Sequential(BitLinear, GELU, BitLinear), where BitLinear is a custom module with both 2D weights and RMSNorm inside, therefore requiring different optimizers. I've reproduced this issue with stock MLX without any custom modules as well.

This is because MultiOptimizer uses tree_merge underneath to merge each optimizer's updates during optimizer.update. Normally, tree_merge treats empty containers ({}, [], ()) as None so it can merge with a non-empty tree. This works. But when both containers at the same position are empty, they both become None, and there is no condition in the function to handle that. Therefore it defaults to the `else` branch which acts as a conflict resolver with a passed function merge_fn. The `else` branch attempts to call merge_fn, but fails since a merge_fn isn't passed by MultiOptimizer.

This happens with parameterless modules, since their optimizer slice is empty.

MLX already treats empty containers as empty subtrees:
```python
>>> tree_flatten({"a": {}})
[]
```
This being absent in tree_merge is inconsistent, since it would be the only util that wouldn't treat an empty container as an empty subtree.

## Reproduction

```python
import mlx.core as mx
import mlx.nn as nn

model = nn.Sequential(nn.Linear(4, 4), nn.ReLU(), nn.Linear(4, 4))
mx.eval(model.parameters())
opt = optim.MultiOptimizer(
    [optim.Muon(learning_rate=0.01), optim.AdamW(learning_rate=0.01)],
    [lambda path, w: w.ndim >= 2],
)
loss_and_grad = nn.value_and_grad(model, lambda m, x: m(x).sum())
loss, grads = loss_and_grad(model, mx.zeros((1, 4)))
opt.update(model, grads)
```

## Proposed changes

My fix makes tree_merge return an empty container of the same type if two empty containers of the same type are being merged. This fixes the MultiOptimizer issue, as evidenced by the two tests I wrote. It throws a ValueError if two empty containers of different types are being merged, since there's no way to decide which one is to use, other than their ordering. Also adds my name and fix description to the acknowledgments.

## Tests

`test_optimizers.py/test_multi_optimizer_with_parameterless_layers`
`test_tree.py/test_empty_subtree_merge`

## Checklist

### DIFF
--- a/ACKNOWLEDGMENTS.md
+++ b/ACKNOWLEDGMENTS.md
@@ -22,7 +22,6 @@ MLX was developed with contributions from the following individuals:
 - Gökdeniz Gülmez: Added the `Muon (MomentUm Orthogonalized by Newton-schulz)` optimizer, and the `ReLU²` activation function.
 - katlun-lgtm: Added `reflect` and `symmetric` padding modes.
 - Erwin Zhang: Added `searchsorted`. Fixed the ring backend hanging when a peer disconnects. Improved JACCL error reporting.
-- Eren Menges: Fixed `MultiOptimizer` not working on models with parameter-free layers (like `ReLU`).
 
 <a href="https://github.com/ml-explore/mlx/graphs/contributors">
   <img class="dark-light" src="https://contrib.rocks/image?repo=ml-explore/mlx&anon=0&columns=20&max=100&r=true" />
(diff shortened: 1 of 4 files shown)


zcbenz (approved):
This looks good to me, thanks! I'm reverting the changes on ACKNOWLEDGMENTS.md though since it is mostly for non-trivial features.

PR #4216, opened 2026-08-12
Optimize BF16 quantized matmul with FP16 tiles

justinchuby:
## Summary
- add an opt-in affine INT4/group-32 QMM path with BF16 device I/O, FP16 threadgroup/MMA tiles, and FP32 accumulation
- vectorize BF16-to-FP16 activation loads and use 64x64 output tiling
- keep AOT and Metal JIT tile instantiations consistent and test both wide/narrow variants

Enable with `MLX_BF16_QMM_FP16=1`; `MLX_BF16_QMM_FP16_TILE=32` selects the narrow comparison path.

## Performance
M1 Max, M=512, K=6656, N=19968, affine INT4 group size 32:

| Path | Time |
|---|---:|
| Stock BF16 | 30.6 ms |
| Stock FP16 | 17.2 ms |
| BF16 I/O + FP16 tiles | 17.32 ms |

The mixed path is ~1.77x faster than stock BF16 and effectively matches pure FP16 while preserving BF16 I/O.

## Validation
- `python/tests/test_quantized.py`: 33 tests / 2896 subtests pass
- AOT and `MLX_METAL_JIT=ON` builds
- wide/narrow tiles, small M/N, non-64 K, and batched inputs
- JIT regression test reproduces the pre-fix tile mismatch and passes with the fix

### DIFF
--- a/mlx/backend/metal/kernels/quantized.h
+++ b/mlx/backend/metal/kernels/quantized.h
@@ -690,6 +690,263 @@ struct QuantizedBlockLoader {
   }
 };
 
+// Experimental variant of QuantizedBlockLoader used by the
+// MLX_BF16_QMM_FP16 path (see mlx/backend/metal/quantized.cpp). The
+// device-resident scales/biases are TExt (bfloat16_t) while the
+// dequantized weights are written into TTile (float16_t) threadgroup
+// memory so that the per-element dequantize math and the downstream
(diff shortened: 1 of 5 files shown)


zcbenz:
Thanks for the PR, while it is a nice improvement we prefer users to convert to fp16 themselves rather than us implicitly do the conversion.

justinchuby:
Thanks - explicit cast verified to have matching performance.

PR #4218, opened 2026-08-13
Implement Group.split() for the ring backend

erwinzhang7:
Part of #3205. `RingGroup::split` threw, so `mx.distributed.init().split(...)` was unavailable
on ring. MPI and NCCL both implement it, so this brings ring to the same contract:

- ranks sharing a color end up in the same group,
- new ranks are ordered by ascending key, with the parent rank breaking ties,
- `key < 0` (the default) means keep the parent's order.

Two things made this smaller than expected.

`TCPSocket::listen` already sets `SO_REUSEADDR` and `SO_REUSEPORT`, and the parent's listening
sockets are closed once `accept()` returns, so a subgroup rebinds the same hostfile addresses.
No port negotiation is needed.

`RingGroup` is constructed from the full address table, so once every rank knows every color it
can work out the subgroups by itself. The constructor was dropping that table, it is kept now.
The colors and keys rotate around the existing ring, so the connection side channel is not
involved.

`RingGroup` also returns early at size one instead of trying to connect to itself, which is what
a color nobody else shares produces. Collectives over a group that size are resolved in
`ops.cpp` before they reach a backend.

`test_groups` in `ring_test_distributed.py` asserted that split raises. It now covers the
behaviour instead.

### Verified

Ring backend on localhost at 2, 4 and 8 ranks. Beyond rank and size, every case checks an
`all_sum` inside the subgroup to confirm the collective is carried by the subgroup and not the
parent.

| | 2 ranks | 4 ranks | 8 ranks |
|---|---|---|---|
| even/odd split | pass | pass | pass |
| uneven blocks (`rank // 3`) | pass | pass | pass |
| nested split of a subgroup | pass | pass | pass |
| identity split reproduces the world | pass | pass | pass |
| singleton subgroups | pass | pass | pass |
| key-driven reordering | pass | pass | pass |
| `ring_test_distributed.py` | 13/13 | 13/13 | 13/13 |

### DIFF
--- a/mlx/distributed/ring/ring.cpp
+++ b/mlx/distributed/ring/ring.cpp
@@ -6,6 +6,7 @@
 #include <unistd.h>
 
 #include <algorithm>
+#include <array>
 #include <chrono>
 #include <fstream>
 #include <future>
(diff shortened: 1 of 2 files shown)


zcbenz (changes requested):
To split a ring you would need to bind the unconnected nodes, this is far from correct.

erwinzhang7:
The verification in the description was localhost only so I re-ran it across machines at `3f2be4bd`.

Four ranks, one per Mac mini on a LAN. Ring edges are 0-1, 1-2, 2-3, 3-0, so a parity split puts `{0,2}` on machines A and C and `{1,3}` on B and D. Neither pair is ring-adjacent, both span physical hosts, so each subgroup has to bind nodes the parent ring never connected. Each rank contributes `10**rank`, so the parent sums to 1111 and the subgroups to 101 and 1010, and wrong groupings are easy to spot.

| case | result |
|---|---|
| parent `all_sum` before any split | ok |
| parity split | ok |
| parent `all_sum` after split, three times interleaved with subgroup use | ok |
| subgroup `all_sum`, 1 MB payload | ok |
| subgroup `all_gather` | ok |
| key-driven reordering, rank reversed | ok |
| nested split of a subgroup | ok |
| parent and first subgroup still correct after nesting | ok |
| singleton subgroup | ok |
| identity split reproduces the world | ok |
| uneven blocks (`rank // 3`) | ok |
| parent `all_sum` after all splits | ok |

20 checks, three repeats, 0 failures. `ring_test_distributed.py` in full also passed on the same four hosts twice, 13 tests on every rank.

Yes it has to bind the unconnected nodes, that is what this does. `TCPSocket::listen` already sets `SO_REUSEADDR` and `SO_REUSEPORT`, and the parent's listening sockets are closed once `accept()` returns (TCPSocket is a stack local, so it closes when it goes out of scope) so a subgroup rebinds the same hostfile addresses. The parent-after-split rows are there because rebinding while the parent is still live is the failure I would expect first; the parent stays correct after being split twice, interleaved with subgroup collectives.

Not litigating if the feature isn't wanted, just providing deeper verification than localhost.

Used Claude Code for verification. The implementation is mine, I am responsible for every line.

PR #4219, opened 2026-08-13
Add runnable distributed examples

### LINKED ISSUE #2930: [Docs] Simple examples of using mlx.distributed

erwinzhang7:
Closes #2930. The docs added in #2973 explain data and tensor parallelism, but `examples/python`
still has nothing you can launch, so this adds two scripts that run as they are:

```
mlx.launch -n 4 python examples/python/distributed_data_parallel.py
mlx.launch -n 4 python examples/python/distributed_tensor_parallel.py
```

Both work on any number of ranks including one, and both end on a number you can check rather
than on output you have to take on trust.

**`distributed_data_parallel.py`** trains a small MLP with SGD. The dataset is drawn from fixed
keys so every rank starts from the same data, then each rank keeps its own slice and drops the
rest. Averaging the gradients of equal slices gives the gradient of the whole dataset, so the
run is equivalent to single process training and the final loss does not move with the number of
ranks:

| ranks | loss |
|---|---|
| 1 | 0.017677 |
| 2 | 0.017679 |
| 4 | 0.017677 |

The sixth decimal moves because the additions happen in a different order, which the docstring
says.

**`distributed_tensor_parallel.py`** shards an MLP across ranks, `AllToShardedLinear` for the up
projection and `ShardedToAllLinear` for the down projection, one all reduce per block. It runs
the full model first and prints the largest difference against the sharded one: `0.000e+00` at 1
and 2 ranks, `2.086e-07` at 4, which is float32 reassociation from summing four partial results.

Both scripts call `mx.eval` on every rank before the rank 0 print. Leaving that to `.item()`
inside the rank check means only rank 0 forces the collective and the others exit while it
waits, which surfaces as `[ring] connection to a peer was lost` and does not point anywhere near
the cause. There is a comment saying so in both files, since it is an easy thing to reproduce by
accident.

There is no fully sharded example because `RingGroup::sum_scatter` throws, so `nn.fully_shard`
cannot run on the ring backend and `mlx.launch` uses ring locally.

### Verified

macOS 26.6, M5 Max, ring backend via `mlx.launch` at 1, 2 and 4 ranks.

### DIFF
--- a/examples/python/distributed_data_parallel.py
+++ b/examples/python/distributed_data_parallel.py
@@ -1,0 +1,102 @@
+# Copyright © 2026 Apple Inc.
+
+"""
+Data parallel training of a small MLP.
+
+Every rank holds a different slice of the same dataset and averages the
+gradients each step, so the training is equivalent to running on one process
+and the final loss is the same at any number of ranks, up to the order the
+floating point additions happen in:
+
+    python examples/python/distributed_data_parallel.py
+    mlx.launch -n 2 python examples/python/distributed_data_parallel.py
+    mlx.launch -n 4 python examples/python/distributed_data_parallel.py
+
+The model is replicated and the batch is split, which is what you reach for
+when the model fits on one machine but the data is large.
+"""
+
+import time
+
+import mlx.core as mx
+import mlx.nn as nn
+import mlx.optimizers as optim
+
+num_features = 100
+num_examples = 1_000
+hidden = 64
+num_iters = 200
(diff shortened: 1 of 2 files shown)


zcbenz (approved):
Thanks!

PR #4222, opened 2026-08-13
Fix nan_to_num replacing inf with 0 for float16 and bfloat16

### LINKED ISSUE #4220: [BUG] nan_to_num replaces inf with 0 instead of the dtype's max finite value for float16/bfloat16
**Describe the bug**

`mx.nan_to_num` documents that `posinf` / `neginf` default to the largest / smallest finite value of the input dtype. That holds for `float32`, but for `float16` and `bfloat16` the defaults are `±0`, so infinities are silently zeroed instead of clamped.

**To Reproduce**

```python
import mlx.core as mx

a = mx.array([float("nan"), float("inf"), float("-inf"), 1.5], dtype=mx.float16)
print(mx.nan_to_num(a))
# array([0, 0, -0, 1.5], dtype=float16)

print(mx.nan_to_num(a.astype(mx.bfloat16)))
# array([0, 0, -0, 1.5], dtype=bfloat16)

print(mx.nan_to_num(a.astype(mx.float32)))  # float32 is correct
# array([0, 3.40282e+38, -3.40282e+38, 1.5], dtype=float32)
```

**Expected behavior**

Match the docstring and numpy: `inf` maps to the dtype's largest finite value (`65504` for float16, `3.38953e+38` for bfloat16). `np.nan_to_num` on the same float16 input returns `[0, 65504, -65504, 1.5]`.

- OS Version: macOS 26.5.2 (Apple Silicon)
- Version: reproduced with the released 0.32.0 wheel (CPU and Metal give the same result) and with a CPU-only source build at 7729d58

**Additional context**

The defaults come from `type_to_max` in `mlx/ops.cpp`, which calls `std::numeric_limits<float16_t>::max()` / `std::numeric_limits<bfloat16_t>::max()`. `std::numeric_limits` has no specialization for those types, so it falls back to the primary template (`is_specialized = false`) whose `max()` returns `T()`, i.e. `0`. The replacement constant is baked into the graph at construction time, which is why every backend is affected. The correct specializations already exist in `mlx/types/limits.h` (`mlx::core::numeric_limits`, used by `finfo`, `softmax`, `logsumexp`). The existing `test_nan_to_num` misses this because its default-argument case never casts the input to the loop dtype.

I have a minimal fix with a regression test ready and will open a PR shortly.

codeAnqiang-ma:
## Proposed changes

`nan_to_num` documents that `posinf` / `neginf` default to the largest / smallest finite value of the input dtype, but for `float16` and `bfloat16` the defaults come out as `±0`, so infinities are silently zeroed:

```python
>>> a = mx.array([float("nan"), float("inf"), float("-inf"), 1.5], dtype=mx.float16)
>>> mx.nan_to_num(a)
array([0, 0, -0, 1.5], dtype=float16)
>>> mx.nan_to_num(a.astype(mx.bfloat16))
array([0, 0, -0, 1.5], dtype=bfloat16)
>>> mx.nan_to_num(a.astype(mx.float32))
array([0, 3.40282e+38, -3.40282e+38, 1.5], dtype=float32)
```

`np.nan_to_num` on the same float16 input returns `[0, 65504, -65504, 1.5]`.

The defaults are computed by `type_to_max` in `mlx/ops.cpp`:

```cpp
} else if (dtype == bfloat16) {
  return std::numeric_limits<bfloat16_t>::max();
} else if (dtype == float16) {
  return std::numeric_limits<float16_t>::max();
```

`std::numeric_limits` has no specialization for `float16_t` / `bfloat16_t` (neither the ARM `__fp16` / `__bf16` typedefs nor the fallback structs), so this instantiates the primary template: `is_specialized` is `false` and `max()` returns `T()`, which is `0`. The wrong constant is baked into the graph at construction time, so every backend is affected — the released 0.32.0 wheel gives the same output on CPU and Metal.

The fix switches to `mlx::core::numeric_limits` from `mlx/types/limits.h`, which carries the correct bit patterns (`0x7BFF` = 65504 for float16, `0x7F7F` ≈ 3.38953e38 for bfloat16) and is what `finfo`, `softmax` and `logsumexp` already use. After the change float16 maps `±inf` to `±65504` and bfloat16 to `±3.38953e+38`, matching numpy and `mx.finfo`.

The existing `test_nan_to_num` never caught this because its default-argument case builds the input without `.astype(t)`, so that path only ever ran in float32. The test now casts to the loop dtype — that one-line change alone fails on main for `float16` — and adds a bfloat16 check against `mx.finfo(mx.bfloat16).max`.

Fixes #4220

## Checklist

CPU-only build (`MLX_BUILD_METAL=OFF`) at 7729d58: `python/tests/test_ops.py` passes 155 tests with 1 skip (a Metal-only case, expected on a CPU build), same as main; the updated `test_nan_to_num` fails on main and passes with this change.

---

Disclosure: I used an AI assistant while investigating and preparing this change. I reproduced the bug, reviewed every line of the diff, and ran the builds and tests locally on an Apple Silicon Mac myself.

### DIFF
--- a/mlx/ops.cpp
+++ b/mlx/ops.cpp
@@ -16,6 +16,7 @@
 #include "mlx/primitives.h"
 #include "mlx/transforms.h"
 #include "mlx/transforms_impl.h"
+#include "mlx/types/limits.h"
 #include "mlx/utils.h"
 
 namespace mlx::core {
@@ -2103,9 +2104,9 @@ array nan_to_num(
     if (dtype == float32) {
       return std::numeric_limits<float>::max();
     } else if (dtype == bfloat16) {
-      return std::numeric_limits<bfloat16_t>::max();
+      return static_cast<float>(numeric_limits<bfloat16_t>::max());
     } else if (dtype == float16) {
-      return std::numeric_limits<float16_t>::max();
+      return static_cast<float>(numeric_limits<float16_t>::max());
     } else {
       std::ostringstream msg;
       msg << "[nan_to_num] Does not yet support given type: " << dtype << ".";
--- a/mlx/ops.cpp
+++ b/mlx/ops.cpp
@@ -2102,11 +2102,11 @@ array nan_to_num(
 
   auto type_to_max = [](const auto& dtype) -> float {
     if (dtype == float32) {
-      return std::numeric_limits<float>::max();
+      return numeric_limits<float>::max();
     } else if (dtype == bfloat16) {
-      return static_cast<float>(numeric_limits<bfloat16_t>::max());
+      return numeric_limits<bfloat16_t>::max();
     } else if (dtype == float16) {
-      return static_cast<float>(numeric_limits<float16_t>::max());
+      return numeric_limits<float16_t>::max();
     } else {
       std::ostringstream msg;
       msg << "[nan_to_num] Does not yet support given type: " << dtype << ".";
--- a/python/tests/test_ops.py
+++ b/python/tests/test_ops.py
@@ -2262,7 +2262,7 @@ class TestOps(mlx_tests.MLXTestCase):
         self.assertTrue(np.allclose(out_mx, out_np))
 
         for t in [mx.float32, mx.float16]:
-            a = mx.array([float("inf"), 6.9, float("nan"), float("-inf")])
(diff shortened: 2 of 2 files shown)


zcbenz (approved):
Thanks!

PR #4224, opened 2026-08-13
Preserve subnormal float values when casting to bool

### LINKED ISSUE #4205: [BUG] any/all and astype(bool) flush float subnormals to zero on Metal
**Describe the bug**

On Metal, a float array holding only subnormal values reduces as if it were all zeros, and casting it to
bool gives False. CPU and numpy both say True. float32 (`0x00000001` .. `0x007FFFFF`) and bfloat16
(`0x0001` .. `0x007F`) are affected; the smallest normal of each is fine, and float16 is unaffected since
its subnormals are normal in float32.

**To Reproduce**

```python
a = mx.array(np.array([0x00000001] * 4, dtype=np.uint32)).view(mx.float32)  # 1.4e-45
print(np.array(a.view(mx.uint32)))                       # [1 1 1 1] -- data is intact
mx.set_default_device(mx.cpu); print(bool(mx.any(a)))    # True
mx.set_default_device(mx.gpu); print(bool(mx.any(a)))    # False
print(np.array(a.astype(mx.bool_)))                      # [False False False False]
```

**Expected behavior**

`any` should be True for any nonzero value, subnormal included, and match the CPU backend and numpy.

**Desktop**
 - OS Version: macOS 26.6, M3 Pro
 - Version: main (also reproduces on 0.32.0 via `astype`, see below)

**Additional context**

The reduce path coerces with `static_cast<U>(in[i])`, `U = bool`, at
`mlx/backend/metal/kernels/reduction/reduce_all.h:40`, and a float->bool cast in-shader appears to flush
subnormals. `-fno-fast-math` is already passed (`kernels/CMakeLists.txt:18`), so it is not that flag.
`astype(mx.bool_)` shows the same thing outside any reduce, which suggests the cast rather than the
reduction is where to fix it.

This predates the recent any/all change: on 0.32.0 `astype(mx.bool_)` on a Metal subnormal is already
False. It only became visible in `any`/`all` once they stopped reinterpreting floats as integers, since
the int path saw the nonzero bits.

Testing the bit pattern instead of the value -- `(bits << 1) != 0`, floating types only -- fixes it
without reintroducing the signed-zero bug, but I have not benchmarked that against the cast.

reckylurker:
Fixes #4205

## Proposed changes

In Metal Shading Language (MSL), casting a floating-point number directly to `bool` via `static_cast<bool>(x)` evaluates in float registers, which flushes subnormal numbers to zero in hardware. So `static_cast<bool>(subnormal)` ends up evaluating as `0.0f != 0.0f` -> `false`.

This was affecting `.astype(mx.bool_)` (in `copy.h`) as well as `any()`, `all()`, and other reductions whenever `U = bool`.

To fix this, introduced a `mlx_cast<U, T>` helper in `utils.h` that checks the raw bit pattern for floats (ignoring the sign bit, e.g. `(as_type<uint32_t>(val) & 0x7FFFFFFF) != 0`). This checks the exponent and fraction bits as raw integers so subnormals don't get flushed by Metal hardware units, while still keeping `-0.0` as `False`.

updated `copy.h`, `reduce_all.h`, `reduce_col.h`, and `reduce_row.h` to use `mlx_cast<U>`.

Added `test_subnormal_bool_cast` in `python/tests/test_ops.py` to check `astype(bool)`, `any()`, and `all()` on `float32` and `bfloat16` subnormals. All Python and C++ tests pass!

## Checklist

### DIFF
--- a/mlx/backend/metal/kernels/utils.h
+++ b/mlx/backend/metal/kernels/utils.h
@@ -446,3 +446,32 @@ template <typename T, typename U>
 struct ConditionalType<true, T, U> {
   using type = T;
 };
+
+///////////////////////////////////////////////////////////////////////////////
+// Type casting utils
+///////////////////////////////////////////////////////////////////////////////
+
+template <typename U, typename T>
+inline U mlx_cast(T val) {
+  return static_cast<U>(val);
+}
+
+template <>
+inline bool mlx_cast<bool, float>(float val) {
+  return (as_type<uint32_t>(val) & 0x7FFFFFFF) != 0;
+}
+
+template <>
+inline bool mlx_cast<bool, bfloat16_t>(bfloat16_t val) {
+  return (as_type<uint16_t>(val) & 0x7FFF) != 0;
+}
+
+template <>
+inline bool mlx_cast<bool, half>(half val) {
+  return (as_type<uint16_t>(val) & 0x7FFF) != 0;
+}
+
+template <>
+inline bool mlx_cast<bool, complex64_t>(complex64_t val) {
+  return mlx_cast<bool, float>(val.real) || mlx_cast<bool, float>(val.imag);
+}
--- a/mlx/backend/metal/kernels/utils.h
+++ b/mlx/backend/metal/kernels/utils.h
@@ -452,26 +452,21 @@ struct ConditionalType<true, T, U> {
 ///////////////////////////////////////////////////////////////////////////////
 
 template <typename U, typename T>
-inline U mlx_cast(T val) {
+inline U cast_to(T val) {
   return static_cast<U>(val);
 }
 
 template <>
-inline bool mlx_cast<bool, float>(float val) {
+inline bool cast_to<bool, float>(float val) {
   return (as_type<uint32_t>(val) & 0x7FFFFFFF) != 0;
 }
 
 template <>
-inline bool mlx_cast<bool, bfloat16_t>(bfloat16_t val) {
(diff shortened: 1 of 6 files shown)


mlx/backend/metal/kernels/utils.h
```diff
@@ -446,3 +446,32 @@ template <typename T, typename U>
 struct ConditionalType<true, T, U> {
   using type = T;
 };
+
+///////////////////////////////////////////////////////////////////////////////
+// Type casting utils
+///////////////////////////////////////////////////////////////////////////////
+
+template <typename U, typename T>
+inline U mlx_cast(T val) {
```
zcbenz: In cuda backend there is a similar utility named `cast_to`, I think we can use the same name here.

mlx/backend/metal/kernels/utils.h
```diff
@@ -446,3 +446,32 @@ template <typename T, typename U>
...
+}
+
+template <>
+inline bool mlx_cast<bool, bfloat16_t>(bfloat16_t val) {
+  return (as_type<uint16_t>(val) & 0x7FFF) != 0;
+}
+
+template <>
+inline bool mlx_cast<bool, half>(half val) {
+  return (as_type<uint16_t>(val) & 0x7FFF) != 0;
```
zcbenz: float16 does not seem to need this, should add a test.

reckylurker: yes, looks like half values are not affected in MSL. Thanks for the suggestion.

zcbenz (approved):
Thanks!

PR #4226, opened 2026-08-13
Fix Metal sort for arrays with negative strides

### LINKED ISSUE #4225: [BUG] mx.sort / topk / partition on Metal GPU return zeros for rows with negative strides
**Describe the bug**

When calling `mx.sort`, `mx.topk`, or `mx.partition` on Metal GPU with a 2D/ND array that has negative strides on a non-sorted axis (such as `x[::-1, :]`), only row 0 is sorted correctly. All subsequent rows (rows 1, 2, ...) return all zeros. CPU device and NumPy both give the correct output.

**To Reproduce**

```python

x_np = np.arange(12, dtype=np.float32).reshape(3, 4)
x_strided = mx.array(x_np)[::-1, :]

print("CPU sort:\n", mx.sort(x_strided, axis=-1, stream=mx.cpu))
print("GPU sort:\n", mx.sort(x_strided, axis=-1, stream=mx.gpu))
print("GPU topk:\n", mx.topk(x_strided, 2, axis=-1, stream=mx.gpu))
```

Output:
```text
CPU sort:
 array([[8, 9, 10, 11],
        [4, 5, 6, 7],
        [0, 1, 2, 3]], dtype=float32)

GPU sort:
 array([[8, 9, 10, 11],
        [0, 0, 0, 0],
        [0, 0, 0, 0]], dtype=float32)

GPU topk:
 array([[10, 11],
        [ 0,  0],
        [ 0,  0]], dtype=float32)
```

**Expected behavior**

GPU results should match CPU and NumPy across all rows.

**Desktop:**
 - OS Version: macOS 26.6 (M2 MacBook Air)
 - Version: main @ c071987

**Additional context**

In `mlx/backend/metal/sort.cpp`, `single_block_sort` binds the input buffer using `set_input_array(in, 0)`, which sets Metal's buffer binding window start to `a_buf + in.offset()`.

For arrays with negative strides (e.g. `x[::-1, :]`), `in.offset()` points to the start of physical row 2 (index 8). In the Metal kernel `block_sort_nc` (`mlx/backend/metal/kernels/sort.h`), `in_block_idx` computes negative relative offsets for subsequent rows (`tid.y = 1 -> -4 elements`, `tid.y = 2 -> -8 elements`). MSL attempts to read memory before Metal's bound buffer offset (`a_buf + in.offset()`), triggering GPU out-of-bounds memory protection which returns `0.0f`.

reckylurker:
Fixes #4225.

## Proposed changes

`mx.sort`, `mx.topk`, and `mx.partition` on Metal GPU returned zeros for non-first rows when called on arrays with negative strides (e.g. `x[::-1, :]`).

In `single_block_sort` and `multi_block_sort`, `set_input_array` binds Metal buffer 0 starting at `a_buf + in.offset()`. When non-sorted axes have negative strides, `elem_to_loc` in MSL computes negative relative pointer offsets for subsequent rows (`tid.y = 1 -> -4 elements`), attempting to read memory prior to Metal's bound buffer window. Metal's GPU memory protection traps these out-of-bounds reads and outputs `0.0f`.

Matching NumPy's behavior when sorting negative-strided views, `gpu_merge_sort` now checks for negative strides and creates a temporary contiguous GPU copy via `contiguous_copy_gpu(in, s)`. Zero-copy execution is fully preserved for standard contiguous arrays, transposed arrays (`x.T`), and positive strided views (`x[::2, :]`).

Added `test_strided_negative_stride_sort` in `python/tests/test_ops.py`.

## Checklist

### DIFF
--- a/mlx/backend/metal/sort.cpp
+++ b/mlx/backend/metal/sort.cpp
@@ -278,9 +278,27 @@ void gpu_merge_sort(
     array& out,
     int axis_,
     bool argsort) {
+  // Negative strides cause MSL elem_to_loc to compute negative relative pointer
+  // offsets, attempting to read memory prior to Metal's bound buffer window
+  // (a_buf + in.offset()). To prevent GPU memory protection traps while
+  // matching NumPy behavior, we create a temporary contiguous copy for negative
+  // strides.
+  bool has_negative_stride = false;
+  for (auto st : in.strides()) {
+    if (st < 0) {
+      has_negative_stride = true;
+      break;
+    }
+  }
+
+  array in_dense = in;
+  if (has_negative_stride) {
+    in_dense = contiguous_copy_gpu(in, s);
+  }
+
   // Get size info
-  int axis = axis_ < 0 ? axis_ + in.ndim() : axis_;
-  int size_sorted_axis = in.shape(axis);
+  int axis = axis_ < 0 ? axis_ + in_dense.ndim() : axis_;
+  int size_sorted_axis = in_dense.shape(axis);
 
   // Get kernel size
   int tn = 4;
@@ -299,7 +317,7 @@ void gpu_merge_sort(
     bn = 32;
   }
 
-  if (bn == 512 && size_of(in.dtype()) > 4) {
+  if (bn == 512 && size_of(in_dense.dtype()) > 4) {
     bn = 256;
   }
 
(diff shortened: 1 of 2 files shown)


zcbenz (changes requested):
The bug is true but a correct fix should get the root cause fixed rather than working around the problem by doing a copy.

zcbenz:
Closing in favor of #4252.

PR #4229, opened 2026-08-13
Return tuple in meshgrid

### LINKED ISSUE #4034: Returning a tuple instead of list for meshgrid to make it compliant with Array API
Hi, currently [`mlx.core.meshgrid`](https://ml-explore.github.io/mlx/build/html/python/_autosummary/mlx.core.meshgrid.html)  returns a list. I am proposing to make it return tuple for making it compliant with [Array API](https://data-apis.org/array-api/latest/API_specification/generated/array_api.meshgrid.html), and how [numpy](https://numpy.org/doc/stable/reference/generated/numpy.meshgrid.html), and [cupy](https://github.com/cupy/cupy/blob/v14.1.1/cupy/_creation/ranges.py#L260) do it.

## Impact
Low risk of breaking real-world code. Common workflows are fully compatible with tuples:

- Unpacking: X, Y = mx.meshgrid(x, y)

- Indexing & Iteration: grids[0], for g in grids:

- Direct passing: mx.stack(mx.meshgrid(...))

Only rare cases relying on isinstance(..., list) or container mutation (.append()) would be affected.

JasonHonKL:
## Proposed changes

Fix #4034 

## Checklist

### DIFF
--- a/python/src/ops.cpp
+++ b/python/src/ops.cpp
@@ -3386,14 +3386,14 @@ void init_ops(nb::module_& m) {
          mx::StreamOrDevice s) {
         std::vector<mx::array> arrays =
             nb::cast<std::vector<mx::array>>(arrays_);
-        return mx::meshgrid(arrays, sparse, indexing, s);
+        return nb::tuple(nb::cast(mx::meshgrid(arrays, sparse, indexing, s)));
       },
       "arrays"_a,
       "sparse"_a = false,
       "indexing"_a = "xy",
       "stream"_a = nb::none(),
       nb::sig(
-          "def meshgrid(*arrays: array, sparse: bool | None = False, indexing: str | None = 'xy', stream: StreamOrDevice = None) -> array"),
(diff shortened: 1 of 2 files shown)


python/src/ops.cpp
```diff
@@ -3386,14 +3386,14 @@ void init_ops(nb::module_& m) {
          mx::StreamOrDevice s) {
         std::vector<mx::array> arrays =
             nb::cast<std::vector<mx::array>>(arrays_);
-        return mx::meshgrid(arrays, sparse, indexing, s);
+        return nb::tuple(nb::cast(mx::meshgrid(arrays, sparse, indexing, s)));
```
zcbenz, line 3389: would `nb::cast<nb::tuple>()` work?

JasonHonKL: I don't think so.
```
error: no matching function for call to 'cast'
note: candidate not viable: no known conversion from 'std::vector<array>' to 'nb::tuple'
```

zcbenz (approved):
Thanks!

PR #4230, opened 2026-08-13
Stabilize reduced-precision InstanceNorm

### LINKED ISSUE #4228: InstanceNorm FP16 variance overflow collapses finite feature maps
### Problem

`mlx.nn.InstanceNorm` computes `mean` and `var` in the input dtype. For FP16 feature maps, the variance can overflow even when every input value is finite. The normalization then collapses a non-constant channel to zero. In a trained 3D segmentation model, larger feature maps also produced non-finite outputs.

### Minimal reproducer

Tested with MLX 0.32.0, Apple M4, macOS 26.4.1:

```python
import mlx.core as mx
import mlx.nn as nn
import numpy as np

checkerboard = np.indices((4, 4, 4)).sum(0) % 2
x = np.stack(
    (
        np.where(checkerboard, -512, 512),
        np.where(checkerboard, -256, 256),
    ),
    axis=-1,
)[None].astype(np.float16)
x = mx.array(x)

direct = nn.InstanceNorm(2)(x)
fp32_stats = nn.InstanceNorm(2)(x.astype(mx.float32)).astype(mx.float16)
mx.eval(direct, fp32_stats)

print(np.asarray(mx.var(x, axis=(1, 2, 3))))
print(np.unique(np.asarray(direct)))
print(np.asarray(fp32_stats).min(), np.asarray(fp32_stats).max())
```

Output:

```text
[[inf inf]]
[-0.]
-1.0 1.0
```

### Expected behavior

For FP16 and BF16 inputs, `InstanceNorm` should calculate reduction statistics in FP32, normalize in FP32, then return the input dtype. This matches the mixed-precision policy normally used for normalization layers and prevents a finite, non-constant feature map from collapsing because the variance reduction overflowed.

### Real model impact

The issue appeared while porting a 29.7M-parameter 3D nnU-Net-v1 inference graph from PyTorch MPS to MLX. Full FP16 produced non-finite logits with trained weights. Casting only the `InstanceNorm` input to FP32 for statistics and casting its output back to the feature dtype removed the non-finite values.

On a complete `128 x 96 x 96` FramePack decode-transfer-inference-host route, this mixed MLX path still processed 6.47 patches/s versus 3.48 patches/s for PyTorch MPS FP16 autocast on the same M4.

I can submit a focused PR with an FP16 regression test and before/after benchmark if this dtype policy is acceptable.

ternaus:
## Proposed changes

Fixes #4228.

This PR stabilizes reduced-precision `InstanceNorm` without changing its public `(N, ..., C)` layout or output dtype:

- reshape each input to `(N, C, spatial_voxels)` and use the stable fused `mx.fast.layer_norm` reduction;
- restore the original channels-last shape before the optional affine transform;
- add an FP16 regression case whose direct variance overflows even though every input value is finite.

## Why

The current FP16 path computes `mx.var` in FP16. A channel alternating between `-512` and `512` therefore gets `var=inf`, and `InstanceNorm` collapses the channel to zero instead of returning values near `-1` and `1`.

`mx.fast.layer_norm` already performs the same reduction needed by `InstanceNorm` after the spatial dimensions are flattened. It avoids a full FP32 input cast and uses the optimized normalization kernel.

## Performance

Apple M4, MLX 0.32.0 core, FP16 input, five warmups and 20 measured calls:

| NDHWC shape | Existing p50 | FP32-cast p50 | This PR p50 |
| --- | ---: | ---: | ---: |
| `1 x 32 x 64 x 64 x 30` | `0.785 ms` | `0.789 ms` | `0.196 ms` |
| `1 x 128 x 96 x 96 x 30` | `2.357 ms` | `5.145 ms` | `1.098 ms` |

The larger shape is the first-stage feature map from the 3D segmentation route that exposed the overflow.

## Validation

- `PYTHONPATH=python:python/tests python -m unittest test_nn.TestLayers`
- `PYTHONPATH=python:python/tests python -m unittest test_nn`

## Checklist

- [x] I have updated the necessary documentation (if needed; no public API or documented behavior changed)

### DIFF
--- a/python/mlx/nn/layers/normalization.py
+++ b/python/mlx/nn/layers/normalization.py
@@ -64,12 +64,19 @@ class InstanceNorm(Module):
                 f"InstanceNorm expects inputs with at least 3 dimensions"
                 f" (N, ..., C) but the input has {x.ndim} dimensions."
             )
-        reduction_axes = tuple(range(1, x.ndim - 1))
-        # Compute stats
-        mean = mx.mean(x, axis=reduction_axes, keepdims=True)
-        var = mx.var(x, axis=reduction_axes, keepdims=True)
-        # Normalize
-        x = (x - mean) * mx.rsqrt(var + self.eps)
+        batch_size, features = x.shape[0], x.shape[-1]
+        spatial_shape = x.shape[1:-1]
+        channels_first = mx.transpose(x, (0, x.ndim - 1, *range(1, x.ndim - 1)))
+        x = mx.fast.layer_norm(
+            channels_first.reshape(batch_size, features, -1),
+            None,
+            None,
+            self.eps,
+        )
+        x = mx.transpose(
+            x.reshape(batch_size, features, *spatial_shape),
+            (0, *range(2, len(spatial_shape) + 2), 1),
+        )
         # Scale and shift if necessary
         return (self.weight * x + self.bias) if "weight" in self else x
 
--- a/python/tests/test_nn.py
+++ b/python/tests/test_nn.py
@@ -693,6 +693,21 @@ class TestLayers(mlx_tests.MLXTestCase):
         ]
         self.assertTrue(x.shape == y.shape)
         self.assertTrue(np.allclose(y, expected_y, atol=1e-5))
+        # Reduced-precision statistics must not overflow for finite feature maps.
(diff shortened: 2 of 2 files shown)


zcbenz (approved):
Nice fix, thanks!

PR #4235, opened 2026-08-13
benchmarks: add Apple I8MM affine-QMM evidence harness

### LINKED ISSUE #4197: RFC: enable NEON I8MM for CPU quantized matmul (8-bit) — 7-24x measured on M5
## RFC: enable NEON I8MM for the CPU quantized matmul path (8-bit)

**tl;dr** Apple Silicon CPUs expose the NEON I8MM extension (UMMLA/SMMLA),
which mlx's CPU quantized kernels never use. A tiled I8MM micro-kernel for the
existing 8-bit quantized matmul measures **7-24x faster than the current CPU
kernel on an M5 Max**, single-threaded, with ~1e-2 max error (from dynamic
int8 activation quantization). Proposal: add it as an opt-in fast path for
the CPU 8-bit path, gated on `FEAT_I8MM`.

### Why I8MM specifically

We audited the M5 feature set to find "on-the-silicon" compute mlx doesn't
use. Results that shape scope:

- **NEON I8MM (`vusmmlaq_s32`) works in user space** (verified on M5). This is
  the clean, immediately-writable target.
- **SVE2 / SME2 (incl. SME2 B16B16/F16F16/I16I64) is present in the silicon
  but SIGILLs in user space on macOS** (verified empirically) — mlx should not
  build CPU work on it.
- NEON bf16-dot (`vdotq_bf16`) and fp16-FML (`FMLAL`) are the same non-
  streaming family and could cover the fp16/bf16 gemm paths later.

### Measured (M5 Max 128 GB, single thread, clang `-march=armv8.6-a+i8mm`)

Same semantics as `mlx/backend/cpu/quantized.cpp`: fp16 activations, uint8
group-64 weights, per-column group scale+bias. Correct against an fp64
reference. Error column: tiled-I8MM vs fp64 (int8 activation quantization is
the only approximation; the current kernel is ~3-9e-3).

| M | K | N | g | current (GFLOPS) | I8MM (GFLOPS) | speedup | err |
|---|---|---:|---:|---:|---:|---:|---:|
| 1 | 4096 | 4096 | 64 | 3.9 | 28.1 | x7.2 | 6.6e+0 |
| 2 | 4096 | 4096 | 64 | 3.9 | 47.8 | x12.2 | 6.4e+0 |
| 8 | 4096 | 4096 | 64 | 3.9 | ~54 | x~14 | 8.6e+0 |
| 64 | 4096 | 4096 | 64 | 3.9 | 53.3 | x13.8 | 9.3e+0 |
| 64 | 2048 | 4096 | 64 | 4.1 | 98.5 | x23.9 | 6.8e+0 |
| 8 | 8192 | 8192 | 32 | 3.9 | 22.8 | x5.9 | 1.0e+1 |

Caveat: the "current" column is a scalar reader standing in for
`_qmm_t_simd`; the in-tree simd kernel may be faster, so the final claim
belongs to an in-tree benchmark, not this micro-table. Micro-kernel: 2 rows
x 2 cols x 8 K-lanes per `vusmmlaq`, activations quantized once per
(row, 64-block) and reused across all columns; scale + zero-point applied
per group in fp32.

Design used the public **vLLM int8 micro-gemm**
(`csrc/cpu/micro_gemm/cpu_micro_gemm_int8_neon.hpp`, Apache-2.0) as the
reference for tiling, adapted to mlx's layout and unsigned weights.

### Proposed integration (open to your steer)

1. New CPU kernel variant for the 8-bit quantized matmul in
   `mlx/backend/cpu/quantized.cpp` (+ 4-bit via pack-expand if wanted),
   transposed and non-transposed, T in {fp32, fp16, bf16}.
2. **Opt-in fast path** — dynamic int8 activations change results ~1e-2 (the
   FP8/MX precedent keeps such paths opt-in). Suggest an env/flag until
   validated on real models.
3. Gating: compile a targeted TU with `-march=armv8.6-a+i8mm` and guard with
   `__ARM_FEATURE_MATMUL_INT8`; runtime fallback via
   `sysctlbyname("hw.optional.arm.FEAT_I8MM", ...)` on the non-specialized
   path.
4. Tests: `TestQuantized` gets the int8-activation variant + the in-tree
   benchmark.

### Ask

Is a PR in this shape welcome? Specifically:
1. Opt-in fast path vs default-on-with-tolerance — which do you prefer?
2. 8-bit first (this kernel), or both 8-bit + 4-bit in the first PR?
3. Runtime gate: `sysctlbyname("hw.optional.arm.FEAT_I8MM")` acceptable, or do
   you prefer compile-time-only (single-binary on all Apple silicon)?
4. Any concern about the ~1e-2 activation-quant error class for mlx's
   numerics guarantees?

Happy to bring the full POC source + in-tree benchmark to the PR.

PhilipJohnBasile:
## Summary

This draft adds an opt-in Apple arm64 evidence harness for comparing MLX affine 8-bit QMM against two dynamic-int8 activation controls:

- a U8×I8 NEON dot-product implementation
- a 2×2 U8×I8 I8MM implementation using `vusmmlaq_s32`

It follows the closed #4197 discussion and is intended only to inform a possible future ARM path alongside Highway work in #3019.

## Scope

This draft:

- does not alter MLX production dispatch
- is excluded from normal builds unless explicitly enabled
- records activation preparation separately from same-preparation compute
- separates numerical-correctness runs from timing-only runs
- embeds the candidate source SHA256 and rejects a mismatched candidate
- records MLX binary/source identities as unbound provenance
- permanently reports `throughput: null`

It is not evidence of a public speedup, model-level quality, production readiness, Highway integration, or MLX source/binary provenance.

## Why this shape

The earlier microbenchmark compared against a scalar stand-in and mixed activation-reuse gains with I8MM gains. This harness adds an actual MLX baseline, a same-preparation dot-product control, alternating measurement order, tail/edge coverage, and explicit fail-closed claim boundaries.

## Validation

- fresh opt-in CMake configure and target build
- source-identity check and mismatch rejection
- normal self-test passed
- ASan and UBSan self-test passed
- diagnostic M=1/2/3 comparison passed
- CMake, C++, Python, and diff formatting checks passed

## Possible extraction path

If the evidence eventually supports production work, activation reuse, the I8MM tile, and ARM Highway/CMake dispatch should be developed and reviewed as separate changes.

### DIFF
--- a/benchmarks/cpp/CMakeLists.txt
+++ b/benchmarks/cpp/CMakeLists.txt
@@ -9,3 +9,29 @@ build_benchmark(single_ops.cpp)
 build_benchmark(irregular_strides.cpp)
 build_benchmark(compare_devices.cpp)
 build_benchmark(autograd.cpp)
+
+# This is intentionally opt-in: it is an ARM I8MM evidence harness for mlx#4197,
+# not a production MLX CPU backend or a portability requirement.
+option(MLX_BUILD_I8MM_EVIDENCE
+       "Build the Apple ARM I8MM affine-QMM evidence prototype" OFF)
+if(MLX_BUILD_I8MM_EVIDENCE)
+  if(NOT APPLE OR NOT CMAKE_HOST_SYSTEM_PROCESSOR MATCHES "arm64")
+    message(FATAL_ERROR "MLX_BUILD_I8MM_EVIDENCE requires an Apple arm64 host")
+  endif()
+  set_property(
+    DIRECTORY
+    APPEND
+    PROPERTY CMAKE_CONFIGURE_DEPENDS
+             "${CMAKE_CURRENT_SOURCE_DIR}/quantized_i8mm_evidence.cpp")
+  file(SHA256 "${CMAKE_CURRENT_SOURCE_DIR}/quantized_i8mm_evidence.cpp"
+       QUANTIZED_I8MM_EVIDENCE_SOURCE_SHA256)
+  add_executable(quantized_i8mm_evidence quantized_i8mm_evidence.cpp)
+  target_compile_features(quantized_i8mm_evidence PRIVATE cxx_std_17)
+  target_compile_definitions(
+    quantized_i8mm_evidence
+    PRIVATE
+      MLX_I8MM_EVIDENCE_SOURCE_SHA256="${QUANTIZED_I8MM_EVIDENCE_SOURCE_SHA256}"
+  )
+  target_compile_options(quantized_i8mm_evidence
+                         PRIVATE "-march=armv8.6-a+i8mm")
+endif()
--- a/benchmarks/cpp/README_i8mm_evidence.md
+++ b/benchmarks/cpp/README_i8mm_evidence.md
@@ -1,0 +1,151 @@
+# CPU I8MM affine-QMM evidence prototype
+
+This directory contains a local, evidence-only exploration following the
+closed [mlx#4197](https://github.com/ml-explore/mlx/issues/4197) discussion. It
+may inform a future Apple I8MM follow-up to Highway
+[PR #3019](https://github.com/ml-explore/mlx/pull/3019). It is not compiled into
+MLX, does not change a production dispatch, and is not a claim that dynamic-int8
+activations preserve model quality.
+
+## What it measures
+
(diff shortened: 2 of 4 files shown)


zcbenz:
Thanks for sharing this! It will help us when benchmarking the highway impl.

PR #4236, opened 2026-08-13
Add sum_scatter to the ring backend

erwinzhang7:
Fixes part of #4285.

`all_reduce_impl` already runs its send/recv plan twice, and the code says why:
"Two times the same send/recv operations, first scatter reduce and then gather."
The `k == 0` half is a reduce-scatter, so sum_scatter is that phase with the local
segment kept instead of continuing into the gather.

Verified on two M4 Pro minis over a direct Thunderbolt link: the call returns
cleanly on both ranks where current main hangs indefinitely, and the ring
distributed test suite gains a sum_scatter case.

Note that this does not address the second half of the issue. An error raised out of
`ReduceScatter::eval_cpu` still does not reach the caller; this change means the
common path no longer raises at all.

### DIFF
--- a/mlx/backend/cpu/distributed.cpp
+++ b/mlx/backend/cpu/distributed.cpp
@@ -98,7 +98,7 @@ void Recv::eval_cpu(
 void ReduceScatter::eval_cpu(
     const std::vector<array>& inputs,
     std::vector<array>& outputs) {
-  assert(inputs.size() == 0);
+  assert(inputs.size() == 1);
   assert(outputs.size() == 1);
 
   auto [in, copied] = ensure_row_contiguous(inputs[0], stream());
--- a/mlx/distributed/ring/ring.cpp
+++ b/mlx/distributed/ring/ring.cpp
@@ -541,7 +541,10 @@ class RingGroup : public GroupImpl {
   }
 
   void sum_scatter(const array& input, array& output, Stream stream) override {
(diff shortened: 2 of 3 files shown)


erwinzhang7:
Adding the end-to-end evidence this was missing, now that a real two-node ring is
set up.

Two M4 Pro minis on the LAN, one rank each, same probe on both builds:

| build | result |
|---|---|
| upstream `9ab977b56` | both ranks enter `sum_scatter`, neither returns, neither raises |
| this branch | both ranks return cleanly |

Without the patch the group stops. Not an error a user can catch and work around
— the job just stops, at 0% CPU, and stays that way. Four ranks from an earlier
run were still alive 23 hours 30 minutes later.

That is the practical case for this PR. `mlx.launch` uses ring locally, and
`nn.fully_shard` reduce-scatters gradients in the backward pass, so FSDP on ring
does not fail with a message saying the op is missing, it hangs. This makes it
run instead.

0.32.0 raised `[ReduceScatter] Not implemented yet.` and it reached
the caller, so the hang is a regression somewhere after that, tracked separately
in #4285. Landing this removes the hang through this path either way, but the
underlying error-propagation problem is its own thing and outlives it.

zcbenz (commented):
Verifying this PR is going to take a lot of work and we don't really have that time at the moment and I'm closing this as won't fix at the moment.

PR #4237, opened 2026-08-13
Fix M5 split-K NAX dispatch cliff

### LINKED ISSUE #4198: qmm_splitk's 512-threadgroup target routes M<=32 transposed qmm to the non-NAX split-K kernel on M5 (M=33 is 21% faster than M=32)
## Summary

On M5, transposed 4-bit `quantized_matmul` at N=6656 (gs64, B=1) runs ~21% slower for M=24..32 than for M=33..40. The M<=32 side dispatches `qmm_t_splitk`, which has no NAX variant; M>=33 dispatches the NAX `qmm_t`. The routing comes from the split-K heuristic's 512-threadgroup target:

- Initial `split_k = max(1, 512 / (ceil(N/32) * ceil(M/32)))` ([quantized.cpp#L1132-L1137](https://github.com/ml-explore/mlx/blob/31b5cbb0933e6d3adc49f148946cd2c3ccb9c292/mlx/backend/metal/quantized.cpp#L1132-L1137)). For N=6656 (208 column tiles), any M<=32 gives an initial `split_k=2`, which survives the K-alignment adjustments when K is divisible by 128 — so the split-K path is taken; `split_k <= 1` would fall back to `qmm` ([#L1150-L1153](https://github.com/ml-explore/mlx/blob/31b5cbb0933e6d3adc49f148946cd2c3ccb9c292/mlx/backend/metal/quantized.cpp#L1150-L1153)).
- The split-K kernel is plain Metal ([#L1178-L1188](https://github.com/ml-explore/mlx/blob/31b5cbb0933e6d3adc49f148946cd2c3ccb9c292/mlx/backend/metal/quantized.cpp#L1178-L1188)); the `qmm` it displaces would hit the NAX gate ([#L1039-L1041](https://github.com/ml-explore/mlx/blob/31b5cbb0933e6d3adc49f148946cd2c3ccb9c292/mlx/backend/metal/quantized.cpp#L1039-L1041)).
- At M=33, `m_tiles=2` → `split_k=1` → NAX `qmm_t`.

`quantized.cpp` is byte-identical between v0.32.0 and current main (36fde275); measurements below are from both.

## Single-op measurement

M5 Max (128 GB), macOS 26.5.2. Chained 64 dependent ops per timed step, bf16 activations, 4-bit gs64, transpose=True, idle GPU, warmup + 6 reps. Band A is the affected shape; band B pads N past 256 column tiles so the initial `split_k` is 1 at every M (never split-K).

| µs/op | M=24 | 28 | 31 | 32 | **33** | 34 | 36 | 40 | M=33/M=32 |
|---|---:|---:|---:|---:|---:|---:|---:|---:|---:|
| A: N=6656, main | 258 | 253 | 256 | 256 | **203** | 200 | 203 | 205 | **0.79** |
| A: N=6656, v0.32.0 | 252 | 252 | 252 | 247 | **194** | 194 | 194 | 196 | **0.79** |
| B: N=8224, main | 212 | 211 | 212 | 201 | 201 | 202 | 202 | 203 | 1.00 |
| B: N=8224, v0.32.0 | 212 | 211 | 213 | 200 | 201 | 202 | 201 | 202 | 1.00 |

Two things stand out: the 21% step at the `m_tiles` boundary exists only in band A, and at M<=32 the N=8224 op is *absolutely faster* (≈212 µs) than the N=6656 op (≈252-258 µs) despite computing 24% more output columns — the smaller matmul loses because it is routed to split-K.

## Where this lands in model shapes

A 52-distinct-layer 30B-decoder-shaped stack (hidden 6656, intermediate 19968, vocab 202048, ~13.7 GB of 4-bit weights per step) on current main: step cost is 1.0-1.2x the M=1 cost through M=4, ~3.9-4.35x across M=16..32, and drops to 3.35x at M=33. On main the qmv/qmv_wide path covers M<=12 for these shapes (vector limit 13), so the split-K window is M=13..32 — where speculative-decoding verification blocks (M = draft+1) and small-batch decode operate.

<details><summary>Repro script (single-op probe)</summary>

```python
import time
import mlx.core as mx

K, CHAIN, REPS = 6656, 64, 6

def qw(n, k):
    w = mx.random.normal((n, k)).astype(mx.bfloat16)
    q = mx.quantize(w, group_size=64, bits=4)
    mx.eval(*q)
    return q

for N in (6656, 8224):
    wq = qw(N, K)
    for M in (24, 28, 30, 31, 32, 33, 34, 36, 40):
        x0 = mx.random.normal((M, K)).astype(mx.bfloat16)
        mx.eval(x0)

        def step(x):
            for _ in range(CHAIN):
                x = mx.quantized_matmul(
                    x, *wq, transpose=True, group_size=64, bits=4
                )[:, :K]
            return x

        mx.eval(step(x0)); mx.synchronize()
        t0 = time.perf_counter()
        for _ in range(REPS):
            mx.eval(step(x0))
        mx.synchronize()
        us = (time.perf_counter() - t0) / REPS / CHAIN * 1e6
        print(f"N={N} M={M:2d} {us:7.1f} us/op")
```

</details>

## Proposed fix (happy to PR)

Bypass split-K when the fallback `qmm` would take the NAX path and `m_tiles == 1`, for the demonstrated band first (N around 6656..8192, K divisible by 128), with an N/K sweep on M5 hardware before any broader threshold change — at much smaller N the heuristic selects far more K partitions and may well be winning there. Can validate candidates on M5 Max, as with #3791.

## Prior art

- #3553 / #3839 (closed) and #3852 (open) cover the qmv/qmv_wide small-M regime; on main the vector limit for these shapes is 13 (#3791), which keeps M<=12 out of the window reported here.
- #4171 changes NAX `qmm_t` tiling for N>8192 — the band-B side. This report concerns the N<=8192 split-K routing.
- #3584 (split-K on NAX): its no-regression sweep began at M=64 / N=2048 and did not cover the `m_tiles==1` region where `split_k=2` engages against a NAX-capable `qmm`.

PhilipJohnBasile:
Fixes #4198.

## Summary

On M5, the split-K heuristic routes one-M-tile transposed QMMs in the
N=6656..8192 band to the plain Metal `qmm_t_splitk` kernel. The fallback
`qmm` path is NAX-capable and is faster for the measured two-way split.

This change:

- factors the existing fallback-QMM NAX eligibility check so the split-K
  decision uses the exact same predicate;
- bypasses split-K only on the measured `applegpu_g17s` M5 Max class for
  single-batch, transposed affine 4-bit/group-64 QMM with one M tile,
  N=6656..8192, K divisible by 128, and a finalized split count of two;
- leaves all other devices, modes, shapes, and split counts on the existing
  path.

The narrow shape/device gate is intentional. Prior measurements show split-K
still wins in other regions, and this PR only changes the band demonstrated in
#4198. It composes with #4171 if that NAX small-M tiling change lands.

## Validation

- Clean editable build from `a8e24f202`: passed.
- `git diff --check`, clang-format, Black, Python syntax: passed.
- CPU `test_quantized`: 37 passed, 3 skipped.
- Pure C++ dispatch-policy test: 18 assertions passed.
- Targeted M5 numerical regression, including M=32/N=6656/K=6656 and adjacent
  N/M boundaries: passed.
- Full M5 Max GPU `test_quantized`: 37 passed.
- M5 Max ABBA performance sweep: passed; details below.

## M5 Max performance

Apple M5 Max (`applegpu_g17s`), macOS 27.0, bfloat16 activations, affine
4-bit/group-64 weights. Baseline and candidate were separate clean builds from
the same `a8e24f202` source. The Metal library hash was identical; only the host
`libmlx` differed. Each cell used 64 dependent operations, warmup, and seven
samples per arm in baseline/candidate/candidate/baseline order.

| Shape | M | baseline | candidate | speedup |
|---|---:|---:|---:|---:|
| K=6656, N=6656 | 32 | 251.4 us | 196.4 us | 1.28x |
| K=6656, N=7168 | 32 | 275.1 us | 199.8 us | 1.38x |
| K=6656, N=7680 | 32 | 277.4 us | 202.6 us | 1.37x |
| K=6656, N=8192 | 32 | 299.6 us | 201.8 us | 1.48x |
| K=4096, N=8192 | 32 | 187.7 us | 126.1 us | 1.49x |
| K=5120, N=8192 | 32 | 232.0 us | 158.5 us | 1.46x |
| K=8192, N=8192 | 32 | 361.1 us | 241.8 us | 1.49x |
| K=6656, N=8224 (control) | 32 | 204.0 us | 204.6 us | 1.00x |

Across all 33 affected M<=32 cells, speedup was **1.22–1.49x** (median
**1.37x**). Across 24 M>=33 or N=8224 controls, the ratio was **0.992–1.012x**
(median **0.999x**).

The M=33/M=32 ratio at N=6656/7168/7680/8192 moved from
0.781/0.733/0.732/0.682 on the baseline to 1.005/1.002/1.000/1.007 on the
candidate, eliminating the dispatch cliff. Baseline and candidate ABBA arm
drift stayed below 1.4% and 1.9%, respectively.

### DIFF
--- a/mlx/backend/metal/quantized.cpp
+++ b/mlx/backend/metal/quantized.cpp
@@ -6,6 +6,7 @@
 #include "mlx/backend/gpu/copy.h"
 #include "mlx/backend/metal/device.h"
 #include "mlx/backend/metal/kernels.h"
+#include "mlx/backend/metal/quantized_dispatch.h"
 #include "mlx/backend/metal/reduce.h"
 #include "mlx/backend/metal/unary.h"
 #include "mlx/backend/metal/utils.h"
@@ -1020,6 +1021,17 @@ void gather_qmm_nax(
   compute_encoder.dispatch_threadgroups(grid_dims, group_dims);
 }
 
+bool qmm_nax_eligible(
+    const array& x,
+    bool transpose,
+    int K,
+    const std::string& mode) {
+  bool has_nax_kernel =
+      metal::is_nax_available() && (transpose || mode == "affine");
+  return has_nax_kernel && transpose && (K % 64 == 0) &&
+      (env::enable_tf32() || x.dtype() != float32);
+}
+
 void qmm(
     const array& x,
     const array& w,
@@ -1035,10 +1047,7 @@ void qmm(
     metal::Device& d,
     const Stream& s,
     const std::string& mode) {
-  bool has_nax_kernel =
-      metal::is_nax_available() && (transpose || mode == "affine");
-  if (has_nax_kernel && transpose && (K % 64 == 0) &&
-      (env::enable_tf32() || x.dtype() != float32)) {
+  if (qmm_nax_eligible(x, transpose, K, mode)) {
     return qmm_nax(
         /* const array& x = */ x,
         /* const array& w = */ w,
@@ -1129,7 +1138,8 @@ void qmm_splitk(
     int K,
     metal::Device& d,
     const Stream& s,
-    const std::string& mode) {
+    const std::string& mode,
+    bool fallback_qmm_is_nax_eligible) {
   // Choose split_k to target ~512 threadgroups
   int bm = 32, bn = 32;
   int n_tiles = (N + bn - 1) / bn;
@@ -1147,6 +1157,22 @@ void qmm_splitk(
   // Ensure K divides evenly by split_k * k_align
   while (split_k > 1 && (K % (split_k * k_align) != 0)) {
     split_k--;
+  }
+  if (metal::qmm_t_splitk_should_use_nax(
+          fallback_qmm_is_nax_eligible,
+          d.get_architecture_gen(),
+          d.get_architecture().back(),
+          /* transpose = */ true,
+          /* single_batch = */ true,
+          /* affine = */ mode == "affine",
+          group_size,
+          bits,
+          m_tiles,
+          N,
+          K,
+          split_k)) {
+    return qmm(
+        x, w, scales, biases, out, true, group_size, bits, M, N, K, d, s, mode);
   }
   if (split_k <= 1) {
     return qmm(
@@ -1807,8 +1833,22 @@ void QuantizedMatmul::eval_gpu(const std::vector<array>& inputs, array& out) {
     // Use split-K qmm for small M with transposed weights (non-batched only)
     int B = out.size() / M / N;
     if (transpose_ && B == 1) {
+      bool fallback_qmm_is_nax_eligible = qmm_nax_eligible(x, true, K, mode);
       qmm_splitk(
-          x, w, scales, biases, out, group_size_, bits_, M, N, K, d, s, mode);
+          x,
+          w,
+          scales,
+          biases,
+          out,
+          group_size_,
+          bits_,
+          M,
+          N,
+          K,
+          d,
+          s,
+          mode,
+          fallback_qmm_is_nax_eligible);
       return;
     }
     qmm(x,
--- a/mlx/backend/metal/quantized_dispatch.h
+++ b/mlx/backend/metal/quantized_dispatch.h
@@ -1,0 +1,33 @@
+// Copyright © 2026 Apple Inc.
+
+#pragma once
+
+namespace mlx::core::metal {
+
+// The #4198 measurements cover the applegpu_g17s M5 Max class's two-way split-K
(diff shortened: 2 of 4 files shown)


zcbenz:
Thanks for the PR but currently we are only focusing on real cases happened during inference otherwise we wouldn't be able to review all the improvements, so I'm closing this PR.

PR #4249, opened 2026-08-14
Add mmap_weights: zero-copy file-backed weight arrays

beatakouchnir:
## Proposed changes

Adds `mx.mmap_weights(file, byte_offset, shape, dtype)`: an array whose
data is an mmap'd, page-cache-backed file region wrapped through the
existing `allocator::make_buffer` no-copy seam (the DLPack import path),
with an explicit release+munmap deleter so the mapping is never recycled
into the buffer cache.

## What it does, measured (M5 Max 128 GB; MoE expert streaming testbed)

- **Bit-identical to `mx.load`, at resident speed.** gemma-4-26B with all
  270 expert tensors served as views: 132 tok/s = indistinguishable from
  resident, outputs bit-identical, through `gather_qmm` on quantized
  experts. The same holds unquantized: the bf16 checkpoint (53 GB) streams
  at **101% of resident decode** (56.7 vs 56.0 tok/s), bit-identical.
- **Load becomes a ~1–2 s reversible mapping instead of a weights read.**
  Cold start pays one pay-as-you-go prefill (12.85 GB faulted at ~5 GB/s);
  the page cache survives process exit, so restarts and repeat CLI
  invocations start near-instantly; releasing all views and re-creating
  them takes 0.05 s — a working residency-control primitive for
  multi-model serving (hot model at full speed, idle models at ~zero wired
  cost after release + ~1 s deferred unwire).
- **Hardened beyond the happy path.** Donation windows (view's last
  reference dropped before eval; eager/chained/compiled/reduction) never
  mutate the mapping — backing-file SHA verified. 2,512 map/use/release
  cycles: zero memory drift. Epoch swaps under in-flight decode
  (pipelined async_eval holding old views): crash-free, bit-identical.
  Cross-family: every weight of a dense 7B (4-bit: 735 tensors; bf16: 339)
  and a per-expert-layout MoE served as views, bit-identical; mixed
  per-projection quantization exercised at the kernel level.

## Two Metal facts reviewers should know (both measured, probes included)

1. **Whole-buffer wiring.** Metal wires the entire bytesNoCopy buffer on
   first kernel use, not touched pages: a one-expert `gather_qmm` (3 MB
   logical) faults its full 1.2 GB stacked tensor. Page-granular laziness
   does not exist — the unit of residency is the buffer. Consequence:
   `mmap_weights` gives zero-copy load and buffer-granular residency
   control, NOT transparent partial-tensor streaming.
2. **The wired ceiling applies to mapped buffers.** A model whose touched
   buffers exceed `max_recommended_working_set_size` fails with a hard
   kIOGPUCommandBufferCallbackErrorOutOfMemory mid-pass (measured at
   ~112 GB on a 128 GB machine; Metal churns residency at ~300 MB/s on
   the way there). So the primitive's domain is models whose weights fit
   under that ceiling — where it delivers resident speed with the
   load/switch/restart economics above. Past the ceiling, applications
   need sub-buffer residency management (we run an expert slot-cache
   there; out of scope for this PR).

## Notes

- safetensors provides no alignment guarantee (real checkpoints place
  tensors at odd offsets), so element-aligned callers need a one-time
  aligned repack today. A follow-up aligning `save_safetensors` output
  would make future checkpoints mappable in place.
- Accounting: mapped bytes appear in `mx.get_active_memory` (make_buffer
  registers size) but in NO process ledger (RSS/footprint/wired) — worth a
  doc note; it confuses first-time measurement.
- Tests included mirror the hardening list above; the measurement harness
  and probes live in the linked comment on #2878.

## Checklist

### DIFF
--- a/docs/src/python/ops.rst
+++ b/docs/src/python/ops.rst
@@ -132,6 +132,7 @@ Operations
    mean
    median
    meshgrid
+   mmap_weights
    min
    minimum
    moveaxis
--- a/mlx/io.h
+++ b/mlx/io.h
@@ -34,6 +34,12 @@ load(std::shared_ptr<io::Reader> in_stream, StreamOrDevice s = {});
 /** Load array from file in .npy format */
 MLX_API array load(std::string file, StreamOrDevice s = {});
 
+/** Zero-copy view of a tensor stored in a file: the data is an mmap'd,
+ * page-cache-backed region wrapped via the allocator's no-copy path
+ * (Metal only). The returned array is read-only by contract. */
+MLX_API array
+mmap_weights(const std::string& file, int64_t byte_offset, Shape shape, Dtype dtype);
+
 /** Load array map from .safetensors file format */
 MLX_API SafetensorsLoad
 load_safetensors(std::shared_ptr<io::Reader> in_stream, StreamOrDevice s = {});
--- a/mlx/io.h
+++ b/mlx/io.h
@@ -35,10 +35,13 @@ load(std::shared_ptr<io::Reader> in_stream, StreamOrDevice s = {});
 MLX_API array load(std::string file, StreamOrDevice s = {});
 
 /** Zero-copy view of a tensor stored in a file: the data is an mmap'd,
- * page-cache-backed region wrapped via the allocator's no-copy path
- * (Metal only). The returned array is read-only by contract. */
-MLX_API array
-mmap_weights(const std::string& file, int64_t byte_offset, Shape shape, Dtype dtype);
+ * page-cache-backed region wrapped via the allocator's no-copy path.
+ * The returned array is read-only by contract. */
+MLX_API array mmap_weights(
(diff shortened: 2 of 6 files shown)


zcbenz:
At the moment we are not considering using mmap, see https://github.com/ml-explore/mlx/discussions/615#discussioncomment-13414338 for some past experiments.

beatakouchnir:
@zcbenz, understood, thanks for taking a look. One small note for whenever this gets revisited: two things have changed since the #615 experiments — mapped buffers via the allocator's no-copy path do run on the GPU (we measured bit-identical outputs at resident speed across 4-bit/8-bit/bf16), and the practicality blocker @antbob hit is exactly the safetensors alignment gap, which an aligned `save_safetensors` would remove for future checkpoints. Measurements are on #2878; the branch stays available if priorities change.

PR #4261, opened 2026-08-15
Fix silent wrong results at gather_mm.

### LINKED ISSUE #4253: gather_mm: silent wrong results with sorted_indices=True + lhs_indices=None on an axis-inserted (non-contiguous) lhs
### Summary

`mx.gather_mm(..., lhs_indices=None, sorted_indices=True)` silently returns **wrong results** when the `a` operand is a non-contiguous view created by inserting an axis (`x[:, None, :]`). No error is raised; the output is catastrophically wrong (not a rounding difference). The same call with `sorted_indices=False`, or with an explicit `lhs_indices`, or with a natively-3D `a`, is correct.

### Repro (mlx 0.31.2, M3 Ultra)

```python
import mlx.core as mx
import numpy as np

E, ROWS, K, N = 4, 3, 256, 64
R = E * ROWS
idx = mx.array(np.repeat(np.arange(E, dtype=np.uint32), ROWS))   # sorted

for tag, build in (("3D direct", lambda: mx.random.normal((R, 1, K))),
                   ("2D + [:, None, :]", lambda: mx.random.normal((R, K))[:, None, :])):
    mx.random.seed(0)
    W = mx.random.normal((E, K, N))
    x = build()
    ref = mx.stack([x[r, 0] @ W[int(idx[r])] for r in range(R)])
    y = mx.gather_mm(x, W, lhs_indices=None, rhs_indices=idx, sorted_indices=True)[:, 0, :]
    print(f"{tag}: max|delta| = {mx.abs(y - ref).max().item():.4f}")
```

```
3D direct: max|delta| = 0.0000
2D + [:, None, :]: max|delta| = 80.1779      # silently wrong
```

Deterministic across seeds (80.2 / 76.5 / 84.9 for seeds 0/1/2) and across float32 / float16 / bfloat16. Correct in all three of these variants:

- `sorted_indices=False` (same view) → 0.0
- `lhs_indices=mx.arange(R, dtype=mx.uint32)` + `sorted_indices=True` (same view) → 0.0
- natively-3D `a` + `lhs_indices=None` + `sorted_indices=True` → 0.0

So the sorted fast path appears to assume a contiguous/row-major `a` layout when `lhs_indices` is None, and reads the wrong rows otherwise.

### Why this is worth fixing rather than documenting

`x[:, None, :]` is the idiomatic way to shape a `[rows, K]` activation matrix for `gather_mm` in a MoE forward — it is exactly what our 2.8T-parameter MoE port does when it groups tokens by expert:

```python
x_c = xf[tok_sorted[s0:s1]][:, None, :]        # the view
y   = mx.gather_mm(x_c, W, lhs_indices=None, rhs_indices=slot, ...)
```

Our code happens to pass `sorted_indices=False`, so it is correct today — but the rows *are* sorted, so flipping that flag to True is the obvious "free speedup" a reader would try, and it would silently corrupt every expert output with no error, no NaN, and plausible-looking magnitudes. Quantized `gather_qmm` with the identical argument pattern is correct, which makes the inconsistency easy to trust into a bug.

I am happy to test a fix against a large real MoE workload (2.8T model, 896 experts, top-16) on M3 Ultra if that helps.

### Suggested minimum

Either fix the layout handling on the sorted path, or raise when `sorted_indices=True` is combined with a non-contiguous `a` and `lhs_indices=None`.

FeliGame:
## Proposed changes

Fix [4253](https://github.com/ml-explore/mlx/issues/4253)
A minimal test case for this PR on CPU (Sry for I can't test it on GPU since my network keep crashing):
```python
import mlx.core as mx
import numpy as np

mx.set_default_device(mx.cpu)

E, ROWS, K, N = 4, 3, 256, 64
R = E * ROWS
idx = mx.array(np.repeat(np.arange(E, dtype=np.uint32), ROWS))

for tag, build in (
    ("3D direct", lambda: mx.random.normal((R, 1, K))),
    ("2D + [:, None, :]", lambda: mx.random.normal((R, K))[:, None, :]),
):
    mx.random.seed(0)
    W = mx.random.normal((E, K, N))
    x = build()
    ref = mx.stack([x[r, 0] @ W[int(idx[r])] for r in range(R)])
    y = mx.gather_mm(x, W, lhs_indices=None, rhs_indices=idx, sorted_indices=True)[:, 0, :]
    max_delta = mx.abs(y - ref).max().item()
    print(f"{max_delta:.4f}")
```

## Checklist

### DIFF
--- a/mlx/backend/metal/matmul.cpp
+++ b/mlx/backend/metal/matmul.cpp
@@ -2158,7 +2158,10 @@ void gather_mm_rhs(
   // lhs_indices.
   auto broadcast_with_indices = [&d, &s, &indices](const array& x) {
     if (x.size() / x.shape(-1) == indices.size()) {
-      return ensure_row_contiguous(x, d, s);
+      // Flatten to 2D so the kernel sees (M, K) not (M, 1, K)
+      std::vector<int> new_shape = {
+          static_cast<int>(indices.size()), x.shape(-1)};
+      return ensure_row_contiguous(reshape(x, new_shape, s), d, s);
     }
 
     auto x_shape = indices.shape();
@@ -2291,7 +2294,10 @@ void gather_mm_rhs_nax(
   // lhs_indices.
   auto broadcast_with_indices = [&d, &s, &indices](const array& x) {
     if (x.size() / x.shape(-1) == indices.size()) {
-      return ensure_row_contiguous(x, d, s);
+      // Flatten to 2D so the kernel sees (M, K) not (M, 1, K)
+      std::vector<int> new_shape = {
+          static_cast<int>(indices.size()), x.shape(-1)};
+      return ensure_row_contiguous(reshape(x, new_shape, s), d, s);
     }
 
     auto x_shape = indices.shape();
--- a/mlx/backend/metal/matmul.cpp
+++ b/mlx/backend/metal/matmul.cpp
@@ -2157,7 +2157,7 @@ void gather_mm_rhs(
   // with rhs_indices. We need only broadcast a and copy it as if applying the
   // lhs_indices.
   auto broadcast_with_indices = [&d, &s, &indices](const array& x) {
-    if (x.size() / x.shape(-2) / x.shape(-1) == indices.size()) {
+    if (x.size() / x.shape(-1) == indices.size()) {
       return ensure_row_contiguous(x, d, s);
     }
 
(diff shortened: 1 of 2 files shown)


zcbenz (changes requested):
Can you add tests for the cases reported by the issue

zcbenz:
Closing per https://github.com/ml-explore/mlx/issues/4253#issuecomment-5301922915.

PR #4269, opened 2026-08-15
Record every cabled pair in a jaccl-ring hostfile

erwinzhang7:
`configure_jaccl_ring` writes `None` for every pair that is not a ring neighbour, including
pairs that are physically cabled:

```python
for other in ring:
    if other not in (peer_left, peer_right):
        rdmas.append(None)
```

`configure_jaccl` a few lines above emits the full N x N matrix, so connectivity survives in a
mesh hostfile and is discarded in a ring one, before anything in C++ sees it.

The information is already available. `IPConfigurator` keys on every connected Thunderbolt port
and assigns addresses per cable, and its `setup()` brings up an interface for each pair it
holds, so the machines are cabled and configured for the whole topology while the hostfile
describes a subset of it.

Nothing changes how the group runs. `MLX_JACCL_RING` selects the ring, not the shape of
the matrix, and `launch_jaccl` only requires the matrix to be square with a null diagonal. Both
still hold, ring neighbours keep the width the ring was built with, and a hostfile for a
topology with no extra cables is byte-identical to before.

What it buys is that a link is no longer lost. A subgroup can only contain members that are
directly connected, so the pairs dropped here are the ones `Group.split()` would need, and the
hostfile is everything the runtime is given. Related to #3205.

The matrix construction moves into `jaccl_ring_devices(ring, count, ips)`, a pure function over
a device map, so it is testable without a cluster. Five tests cover a plain ring, a ring with a
chord, the square-and-null-diagonal contract `launch_jaccl` enforces, ring ordering, and
multiple cables between neighbours.

### DIFF
--- a/python/mlx/_distributed_utils/config.py
+++ b/python/mlx/_distributed_utils/config.py
@@ -483,26 +483,47 @@ def configure_jaccl(args, hosts, ips, sshinfo):
     save_hostfile(args, hostfile)
 
 
+def jaccl_ring_devices(ring, count, ips):
+    """The device matrix for a jaccl-ring hostfile, in ring order.
+
+    Every cabled pair is recorded, not only the ring neighbours. The ring
+    neighbours are what the data plane uses, and `MLX_JACCL_RING` is what
+    selects the ring, so the extra entries change nothing about how the group
+    runs. They are the links a subgroup would need: a child can only contain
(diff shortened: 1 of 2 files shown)


zcbenz (commented):
For changes like I think we should only take when it is actually becoming a problem.

PR #4284, opened 2026-08-16
chore: Check malformed jaccl hostfile that miss rdma in pairs

### LINKED ISSUE #4283: [BUG] A jaccl hostfile whose rank order is wrong silently produces N groups of size 1
### What happens

`launch_jaccl` validates the shape of the rdma matrix, its dimensions and its null
diagonal, but not whether the pairs the backend is about to communicate over are in
it. A hostfile whose rank order puts unconnected nodes next to each other passes
both checks.

Then nothing fails. Every rank ends up **alone in a group of size one**, reports
success, and the job runs at full speed computing N separate wrong answers:

```
[jaccl-badorder] ranks 1  membership OK (all_sum 1)     <- four times over
all_sum   16   0.00 ms   0.00 GB/s
```

Four independent groups where there should be one group of four. No error, no
warning, nothing on stderr.

### Repro

Take a working four node hostfile and permute two hosts, rows and columns together
so the file stays internally consistent. Ranks 0 and 1 then have no cable between
them. Launch anything that reports `group.size()`.

Verified on four M4 Pro minis wired as a full mesh, where the correct file gives one
group of four and the permuted one gives four groups of one.

### Why it matters

This is the failure mode that costs a lot to find. A crash costs a minute, a hang
costs until someone looks. A job that runs at full speed and returns
plausible numbers can be believed.

It is also easy to hit by hand: the order of the hosts in a jaccl-ring hostfile *is*
the order of the ring, and nothing says so.

### Note

JACCL rejects an out of range `send` immediately and with a precise message, so the
silence here looks like an oversight rather than a decision.

erwinzhang7:
Fixes #4283.

The existing checks cover the shape of the rdma matrix and its null diagonal, not
whether the pairs about to be used are present, so a hostfile whose rank order places
unconnected nodes beside each other passes and then leaves every rank alone in a
group of size one, reporting success. Same class as #4210.

`missing_jaccl_links` reports the pairs that are used but absent: for a ring each
rank's two neighbours, for a mesh every pair. With two ranks the previous and next
neighbour are the same node, which is harmless because the pair is simply examined
twice. The error names the hosts and says that for jaccl-ring the host order is the
ring order, since that is the mistake that produces it.

Checked against a live four node mesh: the misordered hostfile is now rejected, a
ring hostfile used as a mesh is rejected, and valid mesh and valid ring hostfiles
still run at 20.87 and 16.33 GB/s respectively. Six unit tests cover mesh, ring, the two rank
edge case and a ring that is deliberately not a mesh.


zcbenz (commented):
Both the code comment and error message are hard to comprehend, so it checks connected nodes must have rdma set in hostfile?

erwinzhang7:
Yes, but it checks the pairs the chosen topology will actually use, not every connected pair. For jaccl-ring that is each rank's two neighbours; for the mesh it is every pair. So a hostfile can be a valid ring and an invalid mesh, which is one of the tests.

zcbenz (approved):

PR #4290, opened 2026-08-16
Dedupe all-axes boilerplate in ops.cpp with all_axes() helper

### LINKED ISSUE #4289: Duplicated all-axes boilerplate across reduction/elementwise ops in ops.cpp
### Description

`mlx/ops.cpp` has fourteen no-axes overloads (`flip`, `all`, `any`, `sum`, `count_nonzero`, `mean`, `median`, `var`, `std`, `prod`, `max`, `min`, `logsumexp`, `softmax`) that each repeat the identical 2-line pattern to build the "all axes" list before delegating to the axes-taking overload:

```cpp
std::vector<int> axes(a.ndim());
std::iota(axes.begin(), axes.end(), 0);
return <op>(a, axes, ..., s);
```

This should be factored into a single small helper (e.g. `all_axes(int ndim)`) placed alongside the existing `compute_reduce_shape` helper in the anonymous namespace at the top of `mlx/ops.cpp`, and each call site reduced to a one-liner.

Note: `pad`'s default-axes block and the `argmin`/`argmax` `keepdims` expansion (which builds `ndim() - 1` axes, a different range) are intentionally different and out of scope.

### Motivation

Pure duplication with no behavior differences between sites — an easy simplification that reduces the surface for the boilerplate to drift out of sync, with existing test coverage in `tests/ops_tests.cpp` and `python/tests/test_ops.py` already exercising the no-axes overloads of these ops.

Adityaj0:
Fourteen no-axes overloads (`flip`, `all`, `any`, `sum`, `count_nonzero`, `mean`, `median`, `var`, `std`, `prod`, `max`, `min`, `logsumexp`, `softmax`) each repeated the same `std::iota`-based boilerplate to build the "all axes" list before delegating to their axes-taking overload.

Extracted a shared `all_axes(int ndim)` helper into the existing anonymous namespace alongside `compute_reduce_shape`, and collapsed each call site to a one-liner. Net diff: 21 insertions / 42 deletions, single file (`mlx/ops.cpp`). Left `pad`'s default-axes logic and the `argmin`/`argmax` keepdims expansion untouched since they build a different range.

Behavior-preserving — verified with the full C++ test suite (`ops_tests.cpp`, notably `test flip` and `test reduction ops`): 263 test cases, 3578 assertions, all passing. `pre-commit run` (clang-format) also passed clean.

Closes #4289 (originally filed against this repo by mistake, refiled at Adityaj0/mlx#1)

### DIFF
--- a/mlx/ops.cpp
+++ b/mlx/ops.cpp
@@ -22,6 +22,13 @@ namespace mlx::core {
 
 namespace {
 
+// Returns {0, 1, ..., ndim - 1}, i.e. all axes of an ndim-dimensional array.
+std::vector<int> all_axes(int ndim) {
+  std::vector<int> axes(ndim);
+  std::iota(axes.begin(), axes.end(), 0);
+  return axes;
+}
+
 std::tuple<Shape, std::vector<int>, bool> compute_reduce_shape(
     const std::vector<int>& axes,
     const Shape& shape) {
@@ -699,9 +706,7 @@ array flip(const array& a, int axis, StreamOrDevice s /* = {} */) {
 }
 
 array flip(const array& a, StreamOrDevice s /* = {} */) {
-  std::vector<int> axes(a.ndim());
-  std::iota(axes.begin(), axes.end(), 0);
-  return flip(a, axes, s);
+  return flip(a, all_axes(a.ndim()), s);
 }
 
 // Slice helper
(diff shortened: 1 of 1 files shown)


zcbenz (approved):

zcbenz:
Can you rebase on latest main?

zcbenz:
@Adityaj0 Can you rebase on latest main?

Adityaj0:
Rebased on latest main. GitHub won't let me reopen this PR since the branch was force-pushed after it closed, so continuing here: #4394

Issue #4044, opened 2026-08-07
[BUG] `mx.cummax` / `mx.cummin` mishandle `NaN`: device-dependent (CPU ≠ GPU), leak ±∞, and disagree with NumPy


JasonHonKL:
**Describe the bug**
The Metal scan kernels for cumulative maximum / minimum
(`mlx/backend/metal/kernels/scan.h`, structs `CumMax` / `CumMin`) implement the
combine step with a plain ordered comparison that never propagates `NaN`:

**To Reproduce**

Include code snippet
```python
import mlx.core as mx
import numpy as np

x = mx.array([1.0, 3.0, float("nan"), 5.0, 4.0])

print("np.maximum.accumulate:", np.maximum.accumulate(np.array(x)))
print("mx.cummax [gpu]      :", np.array(mx.cummax(x, stream=mx.gpu)))
print("mx.cummax [cpu]      :", np.array(mx.cummax(x, stream=mx.cpu)))
```

**Expected behavior**
```
np.maximum.accumulate: [ 1.  3. nan nan nan]
mx.cummax [gpu]      : [1. 3. 3. 5. 5.]      # NaN silently dropped
mx.cummax [cpu]      : [ 1.  3. nan  5.  5.]  # NaN kept at its slot, not propagated
```

PR #4292, opened 2026-08-16
Return the index of the first NaN from argmax and argmin

### LINKED ISSUE #4274: [BUG] mx.argmax / mx.argmin skip NaN while mx.max / mx.min propagate it

erwinzhang7:
Fixes #4274.

Every comparison against a NaN is false, so the bare `<` and `>` in the arg reduce
operators step over NaNs entirely: `argmax` and `argmin` return the index of the
largest or smallest real value while `max` and `min` return NaN for the same input,
so `a[mx.argmax(a)]` and `mx.max(a)` disagree on any array containing one.

numpy and torch both return the index of the first NaN, and MLX already propagates
in `max`, `min`, `cummax` and `cummin`, the last two since #4044, so `argmax` and
`argmin` were the remaining pair that did not agree with the rest.

A NaN now wins the comparison, and once one is held it is not replaced, so the result
is the first NaN rather than whichever the reduction happened to associate last.

`reduce_many` needed the same change as `reduce`. Without it the vectorised path
still skips NaNs while the scalar path does not, so the answer would depend on the
length of the array.

The float check is a compile time guard so the integer instantiations are unchanged.
CPU uses `is_floating_point_v` from `half_types.h`, which covers float16 and
bfloat16 where `std::is_floating_point` does not; Metal and CUDA use a small
`arg_is_nan` helper around the same idea.

### Verified

M5 Max, cpu and metal streams: the five cases in the issue, float32, float16 and
bfloat16, lengths 8, 33, 1000 and 5000 to cover the vectorised path, an array with
two NaNs to confirm the first one wins, and integer input. Without a NaN nothing
changes, including the lowest index tie break. `test_reduce`, `test_ops` and
`test_autograd` pass, and `test_reduce` gains a case for this.

The CUDA change is the same shape but I have no CUDA machine, so it is compiled by
CI rather than run.

### DIFF
--- a/mlx/backend/cpu/arg_reduce.cpp
+++ b/mlx/backend/cpu/arg_reduce.cpp
@@ -1,6 +1,7 @@
 // Copyright © 2023 Apple Inc.
 
 #include <cassert>
+#include <cmath>
 
 #include "mlx/backend/common/utils.h"
 #include "mlx/backend/cpu/encoder.h"
@@ -41,6 +42,21 @@ void arg_reduce_dispatch(
   switch (rtype) {
     case ArgReduce::ArgMin: {
       auto op = [](auto ind_x, auto x, auto ind_y, auto y) {
+        // Every comparison against a NaN is false, so a bare `x < *y` steps
+        // over NaNs entirely and returns the index of the smallest real value
+        // while `mx.min` returns NaN for the same input. Propagate instead,
+        // and keep the first NaN so the index is stable, which is what numpy
+        // and torch return.
+        if constexpr (is_floating_point_v<InT>) {
+          if (std::isnan(*y)) {
+            return;
+          }
+          if (std::isnan(x)) {
+            (*y) = x;
+            (*ind_y) = ind_x;
+            return;
+          }
+        }
         if (x < (*y)) {
           (*y) = x;
           (*ind_y) = ind_x;
@@ -51,6 +67,17 @@ void arg_reduce_dispatch(
     }
     case ArgReduce::ArgMax: {
       auto op = [](auto ind_x, auto x, auto ind_y, auto y) {
+        // See ArgMin above: NaN wins, and the first one keeps the index.
+        if constexpr (is_floating_point_v<InT>) {
+          if (std::isnan(*y)) {
+            return;
+          }
+          if (std::isnan(x)) {
+            (*y) = x;
+            (*ind_y) = ind_x;
+            return;
+          }
+        }
         if (x > (*y)) {
           (*y) = x;
           (*ind_y) = ind_x;
--- a/mlx/backend/cuda/arg_reduce.cu
+++ b/mlx/backend/cuda/arg_reduce.cu
@@ -26,6 +26,19 @@ struct IndexValPair {
   T val;
 };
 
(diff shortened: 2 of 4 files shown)


zcbenz:
Closing as a duplicate of #4291.

PR #4293, opened 2026-08-16
`RMSNormVJP` backward writes a full `{n_rows, D}` `gw_temp` intermediate

JasonHonKL:
## Proposed changes

The GPU backward pass of `mx.fast.rms_norm` computes the weight gradient `gw`
in two full-tensor passes:

1. `vjp_rms_single_row` / `vjp_rms_looped` (one threadgroup per row) write each
   row's `gw` contribution into a temporary of shape `{n_rows, D}` — **the same
   size as the whole input** (`mlx/backend/metal/normalization.cpp:142-152`).
2. A separate strided reduction re-reads all of `gw_temp` and sums the rows
   into the final `{D}` output (`normalization.cpp:200-205`).

For each element of `gw` this moves `2 · n_rows · D · itemsize` bytes through
DRAM (write + re-read) that contain no information beyond partial sums waiting
to be added. On a GPU this is pure bandwidth waste — the arithmetic to combine
rows inside the kernel is free by comparison. It also forces a transient
allocation the size of the input.

## Result 
```
variant	time / 32 iters	per grad call
eager autograd	242–498 ms	~8–15 ms
fast (before)	~77 ms	~2.4 ms
fast (after)	52.8 ms	~1.65 ms
```

## POC
(AI generated)
```py
import time, mlx.core as mx

def t(fn, n=50):                      # timed loop with warmup + eval
    for _ in range(5): mx.eval(fn())
    tic = time.perf_counter()
    for _ in range(n): mx.eval(fn())
    return 1e3 * (time.perf_counter() - tic) / n

x = mx.random.uniform(shape=(8192, 4096)).astype(mx.float16)
w = mx.random.uniform(shape=(4096,)).astype(mx.float16)
y = mx.random.uniform(shape=(8192, 4096)).astype(mx.float16)
mx.eval(x, w, y)

# correctness guard: fast grad must match eager autograd
gf  = mx.grad(lambda x, w: (mx.fast.rms_norm(x, w, 1e-5) * y).sum(), argnums=(0, 1))
ref = mx.grad(lambda x, w: ((mx.rsqrt(x.astype(mx.float32).square().mean(-1, True) + 1e-5)
       * x).astype(mx.float32) * w * y).sum(), argnums=(0, 1))
gx, gw = gf(x, w); rgx, rgw = ref(x, w)
assert mx.abs(gx - rgx).max().item() < 1e-4 and mx.abs(gw - rgw).max().item() < 1e-3, "WRONG"

fwd = t(lambda: mx.fast.rms_norm(x, w, 1e-5))   # unchanged by the fix
bwd = t(lambda: gf(x, w))                        # forward recompute + VJP
print(f"correct  | fwd {fwd:.3f} ms | grad {bwd:.3f} ms | VJP-only {bwd-fwd:.3f} ms")
```

## Checklist

### DIFF
--- a/mlx/backend/metal/normalization.cpp
+++ b/mlx/backend/metal/normalization.cpp
@@ -109,11 +109,9 @@ void RMSNormVJP::eval_gpu(
     array x_copy = contiguous_copy_gpu(x, s);
     return {x_copy, true};
   };
-  bool donate_g = inputs[2].is_donatable();
   auto [x, copied] = check_input(inputs[0]);
   const array& w = inputs[1];
   auto [g, g_copied] = check_input(inputs[2]);
-  donate_g |= g_copied;
   array& gx = outputs[0];
   array& gw = outputs[1];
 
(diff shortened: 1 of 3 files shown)


JasonHonKL:
@zcbenz  await verification I guess haha. Quite some changes mostly related to calculate before merge like divide and conquer.

mlx/backend/metal/normalization.cpp
```diff
@@ -137,17 +135,22 @@ void RMSNormVJP::eval_gpu(
   auto axis_size = static_cast<uint32_t>(x.shape().back());
   int n_rows = x.data_size() / axis_size;
 
+  const int target_groups = 512;
```
nastya236, line 138: Just curious why 512?

mlx/backend/metal/normalization.cpp
```diff
@@ -137,17 +135,22 @@ void RMSNormVJP::eval_gpu(
...
+
   // Allocate the gradient accumulator gw and a temporary to store the
   // gradients before they are accumulated.
-  array gw_temp =
-      (has_w) ? array({n_rows, x.shape().back()}, gw.dtype(), nullptr, {}) : w;
+  array gw_temp = (has_w)
+      ? array({n_groups, x.shape().back()}, gw.dtype(), nullptr, {})
+      : w;
   if (has_w) {
-    if (!g_in_gx && donate_g) {
```
nastya236, line 145: Nice catch. Based on my understanding `g_temp` can reuse a buffer in very rare case: if x is non row-contiguous and g is non row-contiguous. I think it is fair to drop it here for looped case as well.

nastya236 (approved):
Looks great, thank you for your contribution!

mlx/backend/metal/normalization.cpp
JasonHonKL: I tried all two the power till 4096 and 512 perform the best from the bench.

mlx/backend/metal/normalization.cpp
JasonHonKL: thanks

PR #4294, opened 2026-08-16
metal: faster decode attention (tiled online softmax, vectorized KV loads, context-scaled 2-pass split)

jonathan308:
Decode-path SDPA (`sdpa_vector*`, qL ≤ 8) leaves the GPU mostly idle at short and mid context lengths and uses a latency-bound inner loop everywhere:

- the 1-pass kernel launches one threadgroup per (batch·head) — e.g. 32 TGs on a 60-core GPU — and walks the whole KV cache serially;
- both kernels run online softmax **per key** (simd_sum + 2 exp + full output rescale per key = a long serial dependency chain) with scalar 2-byte loads;
- the 2-pass `blocks` heuristic is pinned to 128 for common GQA ratios regardless of context length.

This PR reworks the decode kernels:

1. **Tiled online softmax**: keys processed in tiles of 4 with one max/exp/rescale per tile; the rescale is skipped (uniform branch) when the running max does not increase.
2. **Vectorized K/V loads**: a `load_vec` helper emitting a single 4/8/16-byte load per thread (D=64→4B, 128→8B, 256→16B; D=96 keeps scalar loads due to misalignment).
3. **Context-scaled 2-pass splitting**: `blocks = clamp(roundup32(L/256), 32–64, 256)` on Ultra-class ('d') archs and 2-pass routing for GQA at L≥1024; block-contiguous key chunks replace strided block order.
4. **fp32 2-pass partials** (previously query dtype) — strictly better numerics; pass-2 updated accordingly.

Non-'d' arch routing is unchanged; the kernel improvements apply everywhere but were only benchmarked on M3 Ultra (see Risks).

### Benchmarks (M3 Ultra 60-core)

Stock `benchmarks/python/sdpa_vector_bench.py` (fp16, L=16k, H=32, HK=8):

| case | main | this PR | speedup |
|---|---|---|---|
| sdpa (qL=1) | 1.688 ms | 1.201 ms | **1.41x** |
| sdpa + mask | 1.542 ms | 1.292 ms | 1.19x |

Effective KV bandwidth, qL=1 decode (bf16), main → PR:

| H,HK,D | 2k ctx | 8k ctx | 32k ctx |
|---|---|---|---|
| 40,8,128 | 220→273 GB/s | 407→519 | 1087→1377 |
| 32,8,128 | 293→384 | 666→872 | 1137→1737 |
| 64,8,128 | 207→279 | 507→591 | 762→1188 |
| 32,32,64 (MHA) | 564→1797 | 1445→2791 | ≈ (noise) |

End-to-end decode (mlx_lm generate, Qwen3-8B-4bit, interleaved A/B, <0.5% rep variance): +2.1% tok/s @1k ctx, **+9.5% @8k**, **+12.7% @16k**. Short ctx (≤512) is parity (latency-bound; ±30% run-to-run at 10–25 µs — measured with many-rep medians).

### Tests
- `python/tests/test_fast.py` passes (27/27), incl. sdpa-vs-composed reference.
- Additionally validated across D∈{64,96,128,256}, bf16/fp16, GQA 1–8, L∈[256, 131072], bool/float masks, causal, qL∈{1,4}, tail lengths (257, 4093): max-abs-diff vs fp32 composed reference ≤1e-3, equal or closer than main (fp32 partials).

### Risks / notes for reviewers
- Register pressure at D=256 with the tiled inner loop could reduce occupancy on small GPUs (untested — no M1/M2 hardware available to the author); kernel changes apply to all archs, so a look from someone with base/Pro hardware would be appreciated. Routing changes are 'd'-gated.
- KV with unusual strides failing a new 32-element alignment check now takes a contiguous copy first (correct, small cost).
- `MLX_SDPA_BLOCKS` env override behavior preserved.

### DIFF
--- a/mlx/backend/metal/kernels/sdpa_vector.h
+++ b/mlx/backend/metal/kernels/sdpa_vector.h
@@ -12,6 +12,31 @@ constant bool float_mask [[function_constant(24)]];
 constant bool has_sinks [[function_constant(25)]];
 constant int blocks [[function_constant(26)]];
 
+// Load N contiguous elements of type T with as few (and as wide) loads as
+// possible. The address must be aligned to N * sizeof(T) bytes (capped at 16
+// byte transactions), which the caller guarantees via the stride checks in
+// scaled_dot_product_attention.cpp.
+template <typename T, int N>
+inline void load_vec(thread T (&dst)[N], const device T* p) {
+  if constexpr (N == 2 || N == 4) {
+    *reinterpret_cast<thread vec<T, N>*>(dst) =
+        *reinterpret_cast<const device vec<T, N>*>(p);
+  } else if constexpr (N == 8 && sizeof(T) == 2) {
+    *reinterpret_cast<thread vec<T, 8>*>(dst) =
+        *reinterpret_cast<const device vec<T, 8>*>(p);
+  } else if constexpr (N == 8) {
+    *reinterpret_cast<thread vec<T, 4>*>(dst) =
+        *reinterpret_cast<const device vec<T, 4>*>(p);
+    *reinterpret_cast<thread vec<T, 4>*>(dst + 4) =
+        *reinterpret_cast<const device vec<T, 4>*>(p + 4);
+  } else {
+#pragma unroll
+    for (int i = 0; i < N; i++) {
+      dst[i] = p[i];
+    }
+  }
+}
+
 template <typename T, int D, int V = D>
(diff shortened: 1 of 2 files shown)


zcbenz (changes requested):
This PR includes too many changes that it is incredibly hard to verify whether each of them actually works. Please submit incremental PRs in small changes.

Issue #4259, opened 2026-08-14
Would new fused ops be welcome in mx.fast?


Rahu378:
Every Llama-family decoder block does this twice per layer:

```python
h = x + residual
y = mx.fast.rms_norm(h, w, eps)
```

That is two passes over the hidden state. `mx.fast.rms_norm` is already a tuned kernel, but 
added ahead of it materialises `h` and reads it straight back. 
fusing them into one pass reads `x` and `residual`, writes `h` and `y`, is done. 
vLLM ships `fused_add_rms_norm` for this reason.

I wrote one via `mx.fast.metal_kernel` to find out whether it is worth anything on Apple silicon. Numbers below are fp16 on an M3 MacBook Air (81 GB/s measured copy bandwidth), median of 40 runs, compared against the current two-call path rather than against naive ops:

| shape | `x + residual` then `mx.fast.rms_norm` | fused | speedup |
| --- | --- | --- | --- |
| 1 x 4096 | 0.134 ms | 0.184 ms | **0.73x** |
| 512 x 4096 | 0.445 ms | 0.412 ms | 1.08x |
| 2048 x 2048 | 0.704 ms | 0.660 ms | 1.07x |
| 2048 x 4096 | 1.230 ms | 1.054 ms | 1.17x |
| 4096 x 4096 | 2.180 ms | 1.808 ms | 1.21x |

Output matches the two-call path to 7.8e-3 in fp16, which is one rounding of the final store.

Being straight about the shape of this: 
it is a prefill-only win. 
At a single row it is **slower**, because at that size the dispatch dominates and there is no bandwidth to save. 

So it would want either a row-count threshold or documentation that it is for batched shapes.

Is this something you would want in `mx.fast`?
 I have a working kernel and am happy to open a PR with tests if so, but 
I did not want to send API surface you had not asked for. Equally happy to hear that ~1.2x on one op is below the bar.

Measurement setup: https://github.com/Rahu378/hardware-aware-gateway

zcbenz:
Generally for requests like this we would suggest using [custom extensions](https://ml-explore.github.io/mlx/build/html/dev/extensions.html) and [custom kernels](https://ml-explore.github.io/mlx/build/html/dev/custom_metal_kernels.html), as there are simply too many combinations and too many ways to implement them, and it could happen that we end up adding something that is not practical and not used by anyone as I saw lot of such cases in the past.

[...]

PR #4295, opened 2026-08-16
fast: optional fused residual-add in rms_norm (Metal)

jonathan308:
Every decoder layer runs `h = x + sublayer_out; normed = rms_norm(h)` twice — two kernel launches and an extra full-width read/write of the stream per boundary (72 boundaries per token at 36 layers). This PR adds an optional `residual` argument to `mx.fast.rms_norm`:

```python
normed, h = mx.fast.rms_norm(x, weight, eps, residual=r)
# normed == rms_norm(x + r) * weight ; h == x + r
```

- `residual=None` (default) → unchanged behavior, single array out. Fully backward compatible.
- Metal: two fused kernels (register-cached single-row for H≤3072, looped for larger H) reusing the existing threadgroup reduction structure.
- VJP implemented (`RMSNormVJP` on the sum + add for the sum cotangent); verified against composed autograd (<1e-5 rel).
- CUDA/no-GPU: composed fallback (correct, unfused); a fused CUDA kernel is a straightforward follow-up.

### Numerics
`out_sum` bit-identical to `x + r`. `out` bit-identical for H>3072; ≤1 ulp (fp16/bf16) for H≤3072 (reduction regrouping; documented in tests). Model-level: Qwen3-8B logits **bit-identical** fused vs unfused.

### Benchmarks (M3 Ultra)
- Micro, GPU-bound shapes (4,512,H): 1.13–1.38x for H≥5120, all dtypes. Decode shapes (1,1)/(1,8) are submission-bound: parity to 1.14x.
- E2E decode (mlx_lm, Qwen3-8B-4bit with call sites adopted, interleaved A/B): this PR contributes ~+1.1–2.6% decode tok/s on top of the numbers below, more at longer context.
- Honest losses in micro at small-H batched cells (harness cross-iteration overlap artifacts; convert to wins in serialized dependent chains) are documented in the test comments.

### Tests
- `test_rms_norm_residual` / `test_rms_norm_residual_grad` added (54 dtype/shape configs); full `test_fast.py` 29/29 pass (standalone build, this branch).

### Companion adoption
An mlx-lm PR adopting `residual=` at decoder-layer boundaries (feature-detected so older mlx keeps working) will follow; it is deliberately held until this API lands in a release.

### DIFF
--- a/mlx/backend/cuda/rms_norm.cu
+++ b/mlx/backend/cuda/rms_norm.cu
@@ -462,6 +462,11 @@ bool RMSNorm::use_fallback(Stream s) {
   return s.device == Device::cpu;
 }
 
+bool RMSNorm::use_fallback(Stream s, bool has_residual) {
+  // The fused residual kernel is only implemented on metal for now.
+  return has_residual || use_fallback(s);
+}
+
 template <int n_per_thread, typename F>
 void dispatch_group_dim(int axis_size, F&& f) {
   if (axis_size <= n_per_thread * 8) {
--- a/mlx/backend/metal/kernels/rms_norm.metal
+++ b/mlx/backend/metal/kernels/rms_norm.metal
@@ -156,6 +156,177 @@ template <typename T, int N_READS = RMS_N_READS>
   }
 }
 
+// Fused residual add + RMS norm: computes out = rms_norm(x + res) * w and
+// out_sum = x + res in a single pass. The reduction structure matches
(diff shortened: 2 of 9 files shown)


zcbenz:
Thanks for the PR but at the moment we do not plan to add new fast ops unless it is battle tested, see also https://github.com/ml-explore/mlx/issues/4259#issuecomment-5301003296.

PR #4297, opened 2026-08-16
fast: rope_kv_append — fused RoPE(K) + KV cache append (Metal)

jonathan308:
Decode attention currently runs, per layer per token: `rope(K)` → slice-update K into cache → slice-update V into cache — 3 kernel launches, and K is written twice (rope output buffer, then copied into the cache). This PR adds `mx.fast.rope_kv_append`, a fused primitive that applies RoPE to K at the current offset and writes it **directly into the donated K-cache buffer**, with V appended by a plain slice update inside the same call — 3 launches → 2, and the extra K round-trip eliminated.

```python
key_cache, value_cache = mx.fast.rope_kv_append(
    keys, values, key_cache, value_cache, offset, dims,
    traditional=False, base=base, scale=scale)
```

Design note (the interesting part for reviewers): a 2-output primitive **cannot donate its inputs** — `make_arrays` copies the input vector into each output descriptor plus sibling refs, so `is_donatable()` is always false at eval → whole-cache copies per token (measured 328 µs/call on a 134 MB cache). `SliceUpdate` donates because it is single-output. So this is a single-output `RoPEAppend` primitive (exact `rope_single` arithmetic) writing into the donated K buffer + a V `slice_update`. In-place semantics verified (memory flat across steps; referenced inputs take the copy path untouched).

- Numerics: bit-identical to composed `rope` + slice-updates (870/870 prototype cases: fp16/bf16/fp32 × traditional × D 64/128/256 × dims full/half × offsets {0,255,256,8191} × kv-heads 1–8 × B 1–2).
- VJP: inherits `Custom::vjp` (composed fallback; tested to 1e-6). CUDA/no-GPU: composed fallback (correct, tested). Metal-only fused kernel.
- Gate: 4D, T==1, row-contiguous; anything else silently composes (documented in docstring).

### Benchmarks (M3 Ultra, Qwen3-8B-4bit e2e via mlx_lm, interleaved A/B, 12/12 runs token-digest identical)

| ctx | main | this PR | Δ |
|---|---|---|---|
| 1k | 8.502 ms/tok | 8.306 | **−2.3%** |
| 8k | 10.051 | 9.905 | **−1.5%** |

32-layer dependent chain (the fusible portion): 0.161 → 0.048 ms/token-equiv (3.3x).

### Tests
`python/tests/test_fast.py` 31/31 incl. 2 new tests (fused-vs-composed exactness across the gate matrix, donation behavior).

### Follow-ups (not in this PR)
Full single-kernel K+V append (needs a combined-KV cache layout or multi-output donation; estimated remaining ≈ −0.07…−0.18 ms/tok @36 layers); array-offset variant via DynamicSliceUpdate for compiled decode; CUDA fused kernel. Companion mlx-lm adoption (KVCache `rope=` path, opt-in) will follow once this lands in a release.


zcbenz:
Thanks for the PR but at the moment we do not plan to add new fast ops unless it is battle tested, see also https://github.com/ml-explore/mlx/issues/4259#issuecomment-5301003296.

PR #4301, opened 2026-08-16
fix: support 64-bit eye on GPU

### LINKED ISSUE #4300: [BUG] Metal scatter does not support 64bit dtype
**Describe the bug**  
`mx.eye` fails on the Metal GPU when an 8-byte dtype such as `mx.int64` or `mx.uint64` is requested. This happens because `eye` uses `scatter` internally, and Metal scatter does not support 8-byte output types.

**To Reproduce**

```python
import mlx.core as mx

x = mx.eye(3, dtype=mx.int64)
mx.eval(x)
```

The operation fails with an error similar to:

```text
[Scatter::eval_gpu] Does not support int64
```

The same issue occurs with `mx.uint64`.

**Expected behavior**  
`mx.eye` should return an identity matrix with the requested dtype:

```python
mx.array(
    [[1, 0, 0],
     [0, 1, 0],
     [0, 0, 1]],
    dtype=mx.int64,
)
```

Since `eye` only produces zeros and ones, it could safely construct the matrix using a scatter-supported dtype such as `int32`, then cast the result to the requested dtype without data loss.

- OS Version: macOS 26.6.1
- MLX Version: 0.32.0

aaishwarymishra:
## Proposed changes

This pull request updates the `eye` function to improve support for 64-bit data types and adds corresponding tests to ensure correct behavior. The main changes include handling for 64-bit integer types in the implementation and new test cases for these types.

### Enhanced dtype support:
* Updated the `eye` function in `mlx/ops.cpp` to handle 64-bit data types by constructing the eye matrix as `float32` and then converting it to the requested 64-bit type. This ensures compatibility with `int64` and `uint64` dtypes.

### Testing improvements:
* Added test cases in `python/tests/test_ops.py` to verify that `mx.eye` correctly returns arrays of `int64` and `uint64` dtypes.

closes #4300 
## Checklist

### DIFF
--- a/mlx/ops.cpp
+++ b/mlx/ops.cpp
@@ -408,6 +408,11 @@ array eye(int n, int m, int k, Dtype dtype, StreamOrDevice s /* = {} */) {
   if (n < 0 || m < 0) {
     throw std::invalid_argument("[eye] N and M must be positive integers.");
   }
+
+  if (dtype.size() == 8) {
+    return astype(eye(n, m, k, float32, s), dtype, s);
+  }
+
   array result = zeros({n, m}, dtype, s);
 
   if (n == 0 || m == 0) {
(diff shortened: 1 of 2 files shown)


zcbenz:
Since this is not an important bug I prefer fixing the root issue rather than adding a workaround.

PR #4303, opened 2026-08-16
Fix int32 overflow in concatenate/repeat/kron

ayaangazali:
Three ops compute an output dimension by adding or multiplying existing ones, and all three wrap when the result does not fit in a shape dimension.

`concatenate` is the bad one, because it returns an array and says nothing:

```python
>>> big = mx.zeros(2**30)
>>> mx.concatenate([big] * 4).shape
(0,)
>>> mx.concatenate([big] * 3).shape
(-1073741824,)
>>> mx.concatenate([big] * 5).shape
(1073741824,)
```

A negative dimension is the worst of those. The array is lazy, so nothing is allocated and nothing complains until something downstream trips over the shape.

`repeat` and `kron` wrap the same way, but the wrapped value happens to reach `reshape`, which rejects it with a message naming a shape the caller never asked for:

```python
>>> mx.repeat(big, 4)
ValueError: [reshape] Cannot reshape array of size 4294967296 into shape (0).
>>> mx.kron(mx.zeros(2**16), mx.zeros(2**16))
ValueError: [reshape] Cannot reshape array of size 4294967296 into shape (0).
```

The arithmetic is all on `ShapeElem`, which is `int32_t`:

```cpp
shape[ax] += a.shape(ax);   // concatenate, ops.cpp
shape[axis] *= repeats;     // repeat, ops.cpp
out_shape[i] *= a.shape(j); // kron, ops.cpp
```

Neighbouring ops already guard exactly this. `tile` does `safe_cast(static_cast<int64_t>(reps[i]) * shape[i], "tile")`, and `pad`, `flatten` and `reshape` do the same, which is why `mx.tile(big, 4)` reports cleanly while `mx.repeat(big, 4)` does not. These three were just missed.

Each now accumulates in `int64_t` and goes through `safe_cast`, so they report instead of wrapping:

```
OverflowError: [concatenate] Shape dimension 4294967296 is outside the supported range [-2147483648, 2147483647]. MLX currently uses 32-bit integers for shape dimensions.
OverflowError: [repeat] Shape dimension 4294967296 is outside the supported range ...
OverflowError: [kron] Shape dimension 4294967296 is outside the supported range ...
```

`Concatenate::output_shapes` in `mlx/primitives.cpp` had the identical accumulation and is fixed the same way, so the shapeless and vmap paths agree with the eager one.

Totals that do fit are untouched. The added cases fail on main with `AssertionError: OverflowError not raised` for `concatenate`, and with the misleading reshape `ValueError` for `repeat` and `kron`. Everything stays lazy, so the test allocates nothing near these sizes.

`test_ops.py`, `test_array.py`, `test_compile.py`, `test_vmap.py`, `test_autograd.py`, `test_nn.py`, `test_reduce.py` and the C++ suite (250 cases, 3351 assertions) pass.

I am a freshman working through this codebase and I leaned on Claude Code while doing it. Every output above came from a run on this machine.

### DIFF
--- a/mlx/ops.cpp
+++ b/mlx/ops.cpp
@@ -1330,7 +1330,6 @@ array concatenate(
   };
 
   auto shape = arrays[0].shape();
-  shape[ax] = 0;
   // Accumulate the concatenation axis in 64 bits so a total that does not fit
   // in a shape dimension is reported rather than silently wrapping.
   int64_t concat_size = 0;
--- a/mlx/ops.cpp
+++ b/mlx/ops.cpp
@@ -1430,7 +1430,8 @@ array repeat(const array& arr, int repeats, int axis, StreamOrDevice s) {
 
   // Reshape back into a contiguous array where S_axis is now S_axis * repeats
   shape.erase(shape.begin() + axis + 1);
-  shape[axis] *= repeats;
+  shape[axis] =
+      safe_cast(static_cast<int64_t>(shape[axis]) * repeats, "repeat");
   out = reshape(out, shape, s);
 
   return out;
@@ -3688,11 +3689,13 @@ array kron(const array& a, const array& b, StreamOrDevice s /* = {} */) {
 
   for (int i = ndim - 1, j = a.ndim() - 1; j >= 0; j--, i--) {
     a_shape[2 * i] = a.shape(j);
-    out_shape[i] *= a.shape(j);
+    out_shape[i] =
+        safe_cast(static_cast<int64_t>(out_shape[i]) * a.shape(j), "kron");
   }
   for (int i = ndim - 1, j = b.ndim() - 1; j >= 0; j--, i--) {
     b_shape[2 * i + 1] = b.shape(j);
-    out_shape[i] *= b.shape(j);
+    out_shape[i] =
+        safe_cast(static_cast<int64_t>(out_shape[i]) * b.shape(j), "kron");
   }
 
   return reshape(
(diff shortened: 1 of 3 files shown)


ayaangazali:
Widened this slightly. While checking whether `concatenate` was the only one, I found `repeat` and `kron` doing the same 32 bit multiply on a dimension. They are less dangerous, since the wrapped value reaches `reshape` and errors instead of being returned, but the message names a shape the caller never asked for:

```
>>> mx.repeat(mx.zeros(2**30), 4)
ValueError: [reshape] Cannot reshape array of size 4294967296 into shape (0).
```

Same root cause and same one line fix, so I put them in here rather than opening a second PR for a review that would look identical. Title and description updated to match.

mlx/ops.cpp
```diff
@@ -1331,6 +1331,9 @@ array concatenate(
 
   auto shape = arrays[0].shape();
   shape[ax] = 0;
```
zcbenz: This can be removed.

ayaangazali: Right, removed. `shape[ax]` is never read before it gets assigned from `concat_size`, since the validation loop skips that axis, so zeroing it was left over from when the loop accumulated into it directly.

C++ suite still passes (250 cases, 3351 assertions), along with `test_ops.py`, `test_array.py`, `test_vmap.py`, `test_compile.py` and `test_autograd.py`. Spot checked that the behaviour is unchanged either way: the overflow is still reported for 3, 4 and 5 parts, and ordinary concatenation is untouched for positive and negative axes, ragged and empty inputs, the single array case, the flattening form, and the dimension mismatch error.

zcbenz (approved):

PR #4304, opened 2026-08-16
Stop an out of range optional shape from aborting the process

ayaangazali:
A shape dimension that does not fit in `int32` kills the interpreter instead of raising, whenever the op takes the shape as an optional:

```
$ python -c "import mlx.core as mx; mx.as_strided(mx.zeros(4), (2**31,), (0,))"
libc++abi: terminating due to uncaught exception of type nanobind::python_error:
OverflowError: Integer value 2147483648 is outside the supported range [-2147483648, 2147483647].
Abort trap: 6
```

Exit code 134. It cannot be caught, so a `try`/`except OverflowError` around it does not help.

The same call with the shape as a plain parameter is fine, which is what makes this easy to miss:

```python
>>> mx.zeros((2**31,))          # OverflowError, as intended
>>> mx.reshape(a, (2**31,))     # OverflowError
>>> mx.random.uniform(0, 1, (2**31,))   # OverflowError, shape is mx::Shape
>>> mx.random.bernoulli(0.5, (2**31,))  # abort, shape is std::optional<mx::Shape>
```

The caster added for #2681 raises on overflow, and the comment on it says so:

```cpp
// Not noexcept: on overflow of a narrow integer element we raise
// OverflowError so nanobind surfaces a clean error to the user.
bool from_python(handle src, uint8_t flags, cleanup_list* cleanup) {
```

That holds when nanobind calls the caster for the parameter directly. It does not hold through `std::optional`, because nanobind's `optional_caster::from_python` is declared `noexcept` (`nanobind/stl/detail/nb_optional.h`), so the raise crosses a `noexcept` boundary and calls `std::terminate`.

Affected today: `mx.as_strided` (the `shape` argument), every `mx.fft.*` taking `s`, and `mx.random.bernoulli`. `as_strided`'s `strides` is a `Strides`, whose element type is already 64 bit, so it does not go through the narrow integer path and does not abort.

This adds an optional caster for the same container that forwards to the existing one from a context that is allowed to throw, so an optional shape now reports exactly what a plain one does:

```
OverflowError: Integer value 2147483648 is outside the supported range [-2147483648, 2147483647].
```

I kept the raise rather than switching the caster to return `false`, since returning `false` would fall back to nanobind's generic "incompatible function arguments" `TypeError` and lose the message that #2681 asked for.

Checked that ordinary use of all three is unchanged: `as_strided` with no shape, with a shape, and with both; `fft.fftn` with and without `s`; `fft2`, `rfftn`, `irfftn`; `bernoulli` with and without a shape. The added cases abort the interpreter on main, so the proof is the exit code: `EXIT=134` before, `OK` after.

`test_ops.py`, `test_fft.py`, `test_random.py`, `test_array.py`, `test_compile.py`, `test_vmap.py`, `test_autograd.py`, `test_nn.py` and `test_linalg.py` pass.

I am a freshman finding my way around this codebase, with Claude Code alongside. Everything quoted here is from a run on this machine.

### DIFF
--- a/python/src/small_vector.h
+++ b/python/src/small_vector.h
@@ -9,7 +9,10 @@
 
 #include "mlx/small_vector.h"
 
+#include <optional>
+
 #include <nanobind/stl/detail/nb_list.h>
+#include <nanobind/stl/detail/nb_optional.h>
 
 NAMESPACE_BEGIN(NB_NAMESPACE)
 NAMESPACE_BEGIN(detail)
@@ -102,6 +105,32 @@ struct type_caster<::mlx::core::SmallVector<Type, Size, Alloc>> {
     }
 
     return ret.release();
+  }
+};
+
+// nanobind's optional caster is noexcept, so the OverflowError the caster above
+// raises for an out of range dimension would terminate the process instead of
+// reaching the dispatcher. Forward to the same caster from a context that is
+// allowed to throw, so an optional shape reports the error like a plain one.
+template <typename Type, size_t Size, typename Alloc>
+struct type_caster<std::optional<::mlx::core::SmallVector<Type, Size, Alloc>>>
+    : optional_caster<
+          std::optional<::mlx::core::SmallVector<Type, Size, Alloc>>> {
+  using List = ::mlx::core::SmallVector<Type, Size, Alloc>;
+
+  bool from_python(handle src, uint8_t flags, cleanup_list* cleanup) {
+    if (src.is_none()) {
+      this->value.reset();
+      return true;
+    }
+    make_caster<List> caster;
+    if (!caster.from_python(
+            src, flags_for_local_caster<List>(flags), cleanup) ||
+        !caster.template can_cast<List>()) {
+      return false;
+    }
+    this->value.emplace(caster.operator cast_t<List>());
+    return true;
   }
 };
 
--- a/python/tests/test_ops.py
+++ b/python/tests/test_ops.py
@@ -126,6 +126,21 @@ class TestOps(mlx_tests.MLXTestCase):
             mx.broadcast_to(a, [too_big, 1])
         self.assertIn(str(too_big), str(cm.exception))
 
+        # Ops that take the shape as an optional go through nanobind's
+        # optional caster, which is noexcept, so the overflow used to abort the
+        # process instead of raising. See issue #2681 for the original report.
+        with self.assertRaises(OverflowError) as cm:
+            mx.as_strided(a, [too_big], [0])
+        self.assertIn(str(too_big), str(cm.exception))
+
+        with self.assertRaises(OverflowError) as cm:
+            mx.fft.fftn(a, s=[too_big], axes=[0])
+        self.assertIn(str(too_big), str(cm.exception))
+
+        with self.assertRaises(OverflowError) as cm:
+            mx.random.bernoulli(0.5, [too_big])
+        self.assertIn(str(too_big), str(cm.exception))
+
         # Negative overflow (< int32 min) is caught too.
         too_negative = -(2**31) - 1
         with self.assertRaises(OverflowError) as cm:


ayaangazali:
Scope note, since I went looking for the rest of it after opening this.

The same abort happens through nanobind's other `noexcept` wrapper casters, not just `optional`:

```
mx.split(mx.zeros(4), [2**31])                  # std::variant<int, mx::Shape>   -> abort
mx.zeros(4).split([2**31])                      # same
mx.roll(mx.zeros(4), [2**31], axis=[0])         # std::variant<int, mx::Shape>   -> abort
mx.fast.metal_kernel(...)(output_shapes=[[2**31]], ...)   # std::vector<mx::Shape> -> abort
```

`list_caster::from_python` and `variant_caster::from_python` are both `noexcept` too, so the raise dies the same way.

I did not fix those here on purpose. `type_caster<std::variant<Ts...>>` is not written on top of a reusable base, it privately derives from `variant_caster_storage` and open codes the fold over the alternatives, so covering it means restating nanobind's internals in mlx rather than forwarding to them like this patch does for `optional`. Same for `list_caster`. That is a bigger and more brittle change than I wanted to make on a guess.

Three options as I see them, happy with any:

1. Land this for `optional` and I follow up with the `variant` and `vector` casters written out in full.
2. I add them to this PR now, accepting the copied nanobind internals.
3. You would rather solve it at the root, for instance by having the caster not raise and instead teaching the shape ops to report the out of range value themselves. I tried the simple version of that, making the caster return `false` with the error set, and it does stop the abort but nanobind then reports its generic `TypeError: incompatible function arguments` and the clean message from #2681 is lost, so that particular shortcut is not it.

Tell me which and I will do it.

zcbenz (changes requested):
It seems that the root cause is we shouldn't call `raise_python_error` in `from_python` which converts a python exception to a C++ exception, I think this is merely a workaround and there should be a safe way to raise python exception there. I'm not sure of the correct fix though and you might have to seek help from nanobind.

ayaangazali:
You were right, and I went and checked nanobind rather than guessing. The short version is that there is no safe way to raise a python exception from `from_python`, by design, so the caster in `small_vector.h` is the thing that is wrong.

**nanobind's contract.** `from_python` is documented as `noexcept` in `nb_cast.h`:

```cpp
/// bool from_python(handle src, uint8_t flags, cleanup_list *cl) noexcept {
```

Every caster nanobind ships declares it that way, and its own narrow integer caster signals an out of range value by returning false, not by raising. `load_i32` in `src/common.cpp` is `noexcept` and returns `bool`. So mlx's caster dropping `noexcept` and calling `raise_python_error` is the anomaly, exactly as you said, and the `optional` patch I sent was papering over one symptom of it.

**Why simply returning false is not a drop in.** `nb_func_error_overload` in `src/nb_func.cpp` ends with an unconditional

```cpp
PyErr_SetString(PyExc_TypeError, buf.get());
```

so any error the caster sets is overwritten. I tried it: the abort goes away, but `reshape`, `as_strided`, `split` and `bernoulli` all report `TypeError: incompatible function arguments` and the message #2681 asked for is gone.

**mlx already has the legal pattern.** `check_shape_dim` in `python/src/convert.cpp` raises the same `OverflowError`, but from a plain function called inside a binding lambda, which is fine because the dispatcher catches `python_error` around the call. `zeros`, `ones` and `full` go through `to_shape` and get the clean message that way. `reshape` and the rest bind the C++ function directly, so their `Shape` goes through the caster instead. That is the whole difference between the calls that report properly and the ones that abort.

So the two coherent options, and I would rather have your call than pick one and send it:

**A.** Make the caster `noexcept` and non throwing, and route the shape taking bindings through `to_shape`, the pattern `zeros` and friends already use, so the raise happens where it is allowed. Keeps the #2681 message everywhere and removes the abort at the root. It touches a fair number of bindings: `reshape`, `broadcast_to`, `as_strided`, the eight fft entry points, three in `random.cpp`, `split` in both `ops.cpp` and `array.cpp`, `roll`, and `metal_kernel`.

**B.** Make the caster `noexcept` and non throwing and accept nanobind's generic `TypeError`. Smallest possible diff, removes the abort, but regresses #2681 and its test would need updating.

I am happy to build A, it is mechanical once the direction is settled. Tell me which you prefer and I will open a fresh PR for it.

zcbenz:
I would ask nanobind to consider supporting custom exception thrown from `from_python`, instead of always overriding with its own type error.

ayaangazali:
That is a better place for it than either of my options, and it matches what I found: nanobind documents `from_python` as `noexcept` in `nb_cast.h`, every caster it ships declares it that way, and its own narrow integer caster reports an out of range value by returning false rather than raising. So mlx is not doing anything unusual by wanting a specific error here, it just has no way to express it, because `nb_func_error_overload` ends with an unconditional `PyErr_SetString(PyExc_TypeError, ...)` that discards whatever the caster set.

I will raise it with nanobind. The concrete ask is small: when argument conversion fails and a python error is already set, propagate that instead of overwriting it with the generic `TypeError`. That would let mlx keep the #2681 message by simply setting the error and returning false, with no throwing from `from_python` at all, and it would fix the `optional`, `variant` and `vector` wrappers together rather than one at a time.

I will link the upstream issue here once it is filed, and leave this closed in the meantime rather than reopening it for a workaround.

ayaangazali:
Filed upstream: https://github.com/wjakob/nanobind/issues/1415

The ask there is the narrow one we landed on: when argument conversion fails and a python error is already set, propagate it instead of overwriting it with the generic `TypeError`. `nb_func_error_noconvert` already does exactly that for the return value path, so it is the same check moved to `nb_func_error_overload`.

If that is taken, mlx's caster can set the error and return false, with no throwing from `from_python` at all, and the `optional`, `variant` and `vector` wrappers are fixed together rather than one at a time. I flagged the overload wrinkle in the issue, that an error set by an early failing overload has to be cleared before the next is tried, and offered to send the PR there.

zcbenz:
That sounds a perfect solution, thanks!

PR #4255, opened 2026-08-14
Fix ops rejecting integers larger than INT32_MAX

FeliGame:
## Proposed changes
Fix [4238](https://github.com/ml-explore/mlx/issues/4238)

I try to fix this by toggling scalar decl type to i64.
A smoke test case and its running result shown as below:
```python
$ python3 -c "
import sys
# Import directly from the built .so file
import importlib.util
spec = importlib.util.spec_from_file_location('mlx.core', 'build/core.cpython-312-darwin.so')
mx = importlib.util.module_from_spec(spec)
spec.loader.exec_module(mx)
sys.modules['mlx'] = mx
sys.modules['mlx.core'] = mx

print('Loaded from: build/core.cpython-312-darwin.so')
print()

n = mx.iinfo(mx.int32).max
print('int32 max:', n)
print()

# Test 1: Value parameters: only 4 elements requested, but values need int64
try:
    result = mx.arange(n - 1, n + 3)
    print('Test 1 PASSED:', result)
except Exception as e:
    print('Test 1 FAILED:', type(e).__name__, str(e)[:100])

# Test 2: step larger than int32 max
try:
    result = mx.arange(0, 5, 2**31)
    print('Test 2 PASSED:', result)
except Exception as e:
    print('Test 2 FAILED:', type(e).__name__, str(e)[:100])

# Test 3: value like 2**40
try:
    result = mx.arange(2**40, 2**40 + 3)
    print('Test 3 PASSED:', result)
except Exception as e:
    print('Test 3 FAILED:', type(e).__name__, str(e)[:100])

# Test 4: single-arg arange with int64 value
try:
    result = mx.arange(2**40)
    print('Test 4 PASSED (single arg):', result.shape)
except Exception as e:
    print('Test 4 FAILED:', type(e).__name__, str(e)[:100])

# Test 5: linspace with int64 values
try:
    result = mx.linspace(0, 2**40, 5)
    print('Test 5 PASSED:', result)
except Exception as e:
    print('Test 5 FAILED:', type(e).__name__, str(e)[:100])
"
Loaded from: build/core.cpython-312-darwin.so

int32 max: 2147483647

Test 1 PASSED: array([2147483646, 2147483647, -2147483648, -2147483647], dtype=int32)
Test 2 PASSED: array([0], dtype=int32)
Test 3 PASSED: array([2147483647, 2147483647, 2147483647], dtype=int32)
Test 4 FAILED: ValueError [arange] Maximum size exceeded.
[...]

PR #4305, opened 2026-08-16
Widen list elements that do not fit in int32 to int64

ayaangazali:
`mx.array` infers `int64` for a python int that does not fit in `int32`, but only when it is passed on its own. The same value inside a list fails with a raw C++ error:

```python
>>> mx.array(2**40)
array(1099511627776, dtype=int64)
>>> mx.array([2**40])
RuntimeError: std::bad_cast
>>> mx.array([1, 2**40])
RuntimeError: std::bad_cast
>>> mx.array([2**31])
RuntimeError: std::bad_cast
```

Anything above `INT32_MAX` or below `INT32_MIN` hits it, in a list, a tuple or nested. numpy gives `int64` for all of these.

The scalar path in `convert.cpp` already has the rule:

```cpp
} else if (nb::isinstance<nb::int_>(v)) {
  auto val = nb::cast<int64_t>(v);
  auto default_type = (val > std::numeric_limits<int>::max() ||
                       val < std::numeric_limits<int>::min())
      ? mx::int64
      : mx::int32;
```

The list path does not. `validate_shape` only records the python type category, so every int list infers `int32`, and `array_from_list_impl` then fills a `std::vector<int>`, where `nb::cast<int>` on an out of range value throws `std::bad_cast`. That exception has nothing to do with the actual problem, which is why the message is unhelpable.

This tracks whether any element is out of `int32` range while the shape is being validated, and uses `int64` as the default when one is, so the list path lands on the same dtype the scalar path already picks.

Boundaries are exact, and lists that fit are untouched:

```
[2**31 - 1]      int32        [2**31]         int64
[-(2**31)]       int32        [-(2**31) - 1]  int64
[1, 2, 3]        int32        [2**40, 1]      int64
```

An explicit dtype still wins, and a float anywhere in the list still makes the result float rather than `int64`.

Deliberately left alone, since both are the same on the scalar path and so are not this inconsistency: `mx.array([2**63])` still fails, because the value does not fit `int64` either, and `mx.array([2**40], mx.int32)` still fails, because that is an explicit request rather than inference.

`test_array.py`, `test_ops.py`, `test_autograd.py`, `test_compile.py`, `test_vmap.py`, `test_nn.py` and `test_random.py` pass. The added test errors on main with `std::bad_cast`.

I am a freshman learning my way around this codebase, using Claude Code as I go. Every result quoted here came from a run here.

### DIFF
--- a/python/src/convert.cpp
+++ b/python/src/convert.cpp
@@ -505,7 +505,8 @@ PyScalarT validate_shape(
     T list,
     const mx::Shape& shape,
     int idx,
-    bool& all_python_primitive_elements) {
+    bool& all_python_primitive_elements,
+    bool& has_wide_int) {
   if (idx >= shape.size()) {
     throw std::invalid_argument("Initialization encountered extra dimension.");
   }
@@ -524,13 +525,18 @@ PyScalarT validate_shape(
     PyScalarT t;
     if (nb::isinstance<nb::list>(l)) {
       t = validate_shape(
-          nb::cast<nb::list>(l), shape, idx + 1, all_python_primitive_elements);
+          nb::cast<nb::list>(l),
+          shape,
+          idx + 1,
+          all_python_primitive_elements,
+          has_wide_int);
     } else if (nb::isinstance<nb::tuple>(*list.begin())) {
       t = validate_shape(
           nb::cast<nb::tuple>(l),
           shape,
           idx + 1,
-          all_python_primitive_elements);
+          all_python_primitive_elements,
+          has_wide_int);
     } else if (nb::isinstance<mx::array>(l)) {
       all_python_primitive_elements = false;
       auto arr = nb::cast<mx::array>(l);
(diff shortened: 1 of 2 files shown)


ayaangazali:
One related measurement, for whenever you look at this, since it decides how far the patch should go.

With an explicit narrow dtype the two paths still disagree, and this PR does not change that:

```
mx.array(2**40, mx.int32)     ->  0                            (wraps, like C)
mx.array([2**40], mx.int32)   ->  RuntimeError: std::bad_cast
```

I left it alone because picking a side is your call, not a bug fix:

- Making the list path wrap too would match the scalar path, and match the house style elsewhere, since `mx.array(300, mx.uint8)` gives 44, `astype` saturates rather than raising, and #4255 deliberately tests `arange` wrapping at the int32 boundary.
- Making both raise would match numpy, which gives `OverflowError: Python integer 1099511627776 out of bounds for int32`, but that changes a long standing behaviour on a very common call.

This PR only touches the case where no dtype is given, so it does not commit you to either. Say the word if you want the explicit case made consistent in here and I will add it.

zcbenz (approved):

PR #4307, opened 2026-08-16
[CUDA] Fused conv+bias+SiLU epilogue

Joannis:
## Summary

Adds a `FusedConvBiasActivation` primitive and a `conv2d_bias_silu` op that fold a per-output-channel **bias add** and a **SiLU** activation directly into the cuDNN convolution graph epilogue (pointwise `ADD` + `SWISH_FWD`) instead of running three separate kernels.

Folding the epilogue into the conv graph avoids two extra HBM round-trips of the activation tensor: today the conv output is written to HBM, re-read by a bias kernel, written back, then re-read again by the SiLU kernel. With the fused graph the conv result stays resident and only the final activated output is written once.

## Details

- New primitive `FusedConvBiasActivation` (a `UnaryPrimitive`), inputs `{x, w, bias}`. `bias` is 1-D of length `C_out`, broadcast on the NCHW channel dim as `[1, C_out, 1, 1]`.
- Exposed as `mlx::core::conv2d_bias_silu(input, weight, bias, stride, padding, dilation, groups, stream)`.
- The cuDNN path extends `build_conv_graph` with optional `bias` / `apply_silu` params: after the conv node it appends a pointwise `ADD` (bias) and a pointwise `SWISH_FWD` (SiLU — `x * sigmoid(x)` with the default beta=1), computed in fp32, then marks that as the graph output.
- Fused graphs use a **separate cache** (`fused_conv_cache`) from the plain conv graphs so the forward-only epilogue graphs never collide with a plain conv graph on the same shape key.
- **CUDA-only fast path.** CPU / Metal / no_gpu / no_cpu are throwing stubs — this is an inference-only path; off CUDA a model routes to the plain conv + bias + silu ops. **Forward/inference only, no autograd.**

## Measured result

On YOLOv9-C fp16 640x640 on a GB10 (DGX Spark), the bare cuDNN conv already matches PyTorch (~6.19 vs ~6.17 ms). The entire remaining backbone gap was the separate bias + SiLU launches; folding them into the conv epilogue closes it.

## Open item

`eval_gpu` currently **throws** if cuDNN can't build the fused graph for a given shape. All standard convs build one on cuDNN 9.x, so this doesn't fire in practice, but a decomposed fallback that runs the plain conv + bias + SiLU is the natural follow-up.

## Notes / caveats

- The pointwise enums used (`fe::PointwiseMode_t::ADD`, `fe::PointwiseMode_t::SWISH_FWD`) were verified present in cudnn-frontend headers on hand (v1.27.0). This repo currently pins cudnn-frontend **v1.16.0** via `FetchContent`; both enums are long-standing and expected to be present in 1.16.0, but I was not able to inspect the exact v1.16.0 headers directly — flagging in case the pin matters.
- Only compile-verified by construction against the current CUDA backend APIs; I do not have a CUDA build host to compile/run this branch here.

https://claude.ai/code/session_01JMq6N8WW3h7bGdgskwoFDQ


zcbenz:
Thanks for the PR, at the moment we don't take contributions on adding new fused ops, see also https://github.com/ml-explore/mlx/issues/4259#issuecomment-5301003296.

PR #4309, opened 2026-08-16
Select diag's diagonal out of a broadcast for 8-byte types

### LINKED ISSUE #4308: [BUG] mx.diag on a 1-D input fails on the GPU for int64, uint64 and complex64
### What happens

`diag` on a 1-D input fails on the GPU for any 8-byte dtype, and works on the CPU:

```python
import mlx.core as mx

mx.eval(mx.diag(mx.array([1, 2, 3], dtype=mx.int64)))
# [scatter] GPU scatter does not yet support int64 for the input or updates
```

`uint64` and `complex64` fail the same way. `int32` and `float32` are fine, and every
one of them works on the CPU stream.

### Why

`diag` places the values with a `scatter` (`mlx/ops.cpp:6371`), and `Scatter::eval_gpu`
rejects any 8-byte output that is not a `complex64` sum
(`mlx/backend/metal/indexing.cpp:233`). Nothing above that guard knows the restriction
exists, so an op that has nothing to do with scatter fails on a dtype it otherwise
supports.

`eye` had the same problem through the same path and was fixed in #4301 by building in a
narrower type and casting. **That is not available here.** `eye` only ever writes 0 and 1,
which are exact everywhere; these are the caller's values, and a cast through float32
changes them:

```
in                  9007199254740993   4611686018427387903
via float32         9007199254740992   4611686018427387904
```

### Environment

mlx built from c2bcf47, M5 Max and M4 Pro, macOS 26.5. Also reproduced on top of #4301,
so it is independent of that fix.

### Fix

PR below. The 8-byte types select the diagonal out of a broadcast instead of scattering
into it, which never leaves the dtype. The existing scatter path is untouched for every
type that works today, since the mask costs an n x n allocation the scatter does not.

erwinzhang7:
Fixes #4308.

`diag` places a 1-D input on the diagonal with a `scatter`, and `Scatter::eval_gpu` refuses
any 8-byte output that is not a `complex64` sum, so `int64`, `uint64` and `complex64` fail
on the GPU and work on the CPU.

`eye` reached the same guard and #4301 fixes it by building in a narrower type and casting.
That works there because the only values `eye` writes are 0 and 1, exact in every type it
can return. It does not work here: these are the caller's values, and float32 has 24 bits
of mantissa.

```
in                  9007199254740993   4611686018427387903
via float32         9007199254740992   4611686018427387904
```

So the 8-byte types select the diagonal out of a broadcast rather than scattering into it,
which never leaves the dtype. Only those three take that path: the mask is an n x n
allocation the scatter does not need, and there is no reason to spend it on the types that
already work.

### Verified

M5 Max, cpu and gpu streams. 180 cases over int64, uint64, complex64, int32, float32 and
uint8, offsets -3 to 2, and lengths 1, 3 and 6, values checked against numpy: all pass.
`2^53 + 1`, `2^62 - 1` and `2^64 - 1` round trip exactly, which is the case a cast would
have changed. The empty input and the 2-D extraction path are unchanged.

`test_ops`, `test_autograd` and `test_array` pass. Fork CI green on all 22 jobs, including
metal and jit on macOS 14, 15 and 26.2, and the cuda 12.6, 12.9 and 13.0 legs.

### DIFF
--- a/mlx/ops.cpp
+++ b/mlx/ops.cpp
@@ -6362,9 +6362,26 @@ array diag(const array& a, int k /* = 0 */, StreamOrDevice s /* = {} */) {
       return res;
     }
 
-    std::vector<array> indices;
     auto s1 = std::max(0, -k);
     auto s2 = std::max(0, k);
+
+    if (size_of(a.dtype()) == 8) {
+      // The Metal scatter has no 8-byte output path, so placing the values
+      // with one refuses int64, uint64 and complex64. eye can sidestep that by
+      // building in a narrower type and casting, because the only values it
+      // writes are 0 and 1. These are the caller's values, and an int64 past
+      // float32's exact range would come back changed, so instead of casting
+      // the diagonal is selected out of a broadcast, which never leaves the
+      // dtype. Only the types that currently throw take this path.
+      auto vals =
+          pad(a, {s1, n - s1 - a_size}, array(0, a.dtype()), "constant", s);
+      auto rows = expand_dims(arange(n, s), 1, s);
+      auto cols = expand_dims(arange(n, s), 0, s);
+      auto on_diagonal = equal(cols, add(rows, array(k), s), s);
+      return where(on_diagonal, expand_dims(vals, 1, s), res, s);
+    }
+
+    std::vector<array> indices;
     indices.push_back(arange(s1, a_size + s1, uint32, s));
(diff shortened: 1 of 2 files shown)


zcbenz:
Since this is not an important bug I prefer fixing the root issue rather than adding a workaround.

PR #4312, opened 2026-08-16
Reject a negative eps in the normalization layers

ayaangazali:
`eps` is added under a square root in every normalization layer, so a negative one makes `rsqrt` take the root of a negative number and the layer returns NaN, with no error:

```python
>>> x = mx.random.normal((4, 8, 16))
>>> out = nn.LayerNorm(16, eps=-1.0)(x)
>>> mx.isnan(out).any(), mx.isnan(out).all()
(array(True), array(False))
```

The `all()` being `False` is the awkward part. Only the elements whose variance happens to land below `-eps` come back NaN, so the output is part finite and part NaN rather than obviously broken, and the cause is a constructor argument set much earlier.

All five layers behave the same way: `LayerNorm`, `RMSNorm`, `GroupNorm`, `InstanceNorm` and `BatchNorm`.

The library already treats this as worth guarding elsewhere, for exactly this reason. From `test_optimizers.py`:

```python
def test_epsilon_validation(self):
    # RMSprop/Adagrad/AdaDelta accumulators start at zero, so eps == 0 lets a
    # zero gradient compute 0 / 0 (or sqrt(0) / sqrt(0)), producing NaN
    # parameters. The guard must reject eps <= 0, and the message says ">0".
```

So `RMSprop`, `Adagrad`, `AdaDelta` and `Adamax` all validate their `eps`, and `_Pool` validates its kernel size, stride and padding. The normalization layers were missed.

This rejects `eps < 0` in the five constructors, with the message shape `_Pool` already uses:

```
ValueError: [LayerNorm] 'eps' must be non-negative but got -1.0.
```

I checked `eps == 0` rather than assuming, and left it allowed: it is a reasonable thing to ask for and every layer produces finite output with it on ordinary input, so requiring `> 0` here would reject working code. Only the negative case, which has no valid reading, is rejected.

`test_nn.py`, `test_losses.py`, `test_optimizers.py` and `test_compile.py` pass. The added test fails on main.

I am a freshman working my way through this codebase, with Claude Code helping. Everything shown above came from running it here.

### DIFF
--- a/python/mlx/nn/layers/normalization.py
+++ b/python/mlx/nn/layers/normalization.py
@@ -47,15 +47,13 @@ class InstanceNorm(Module):
         affine: bool = False,
     ):
         super().__init__()
+        if eps <= 0.0:
+            raise ValueError(f"[InstanceNorm] 'eps' must be positive but got {eps}.")
         if affine:
             self.weight = mx.ones((dims,))
             self.bias = mx.zeros((dims,))
         self.dims = dims
         self.eps = eps
-        if eps < 0.0:
-            raise ValueError(
-                f"[InstanceNorm] 'eps' must be non-negative but got {eps}."
-            )
 
     def _extra_repr(self):
         return f"{self.dims}, eps={self.eps}, affine={'weight' in self}"
@@ -105,13 +103,13 @@ class LayerNorm(Module):
         self, dims: int, eps: float = 1e-5, affine: bool = True, bias: bool = True
     ):
         super().__init__()
+        if eps <= 0.0:
+            raise ValueError(f"[LayerNorm] 'eps' must be positive but got {eps}.")
         if affine:
             self.weight = mx.ones((dims,))
             if bias:
                 self.bias = mx.zeros((dims,))
         self.eps = eps
-        if eps < 0.0:
-            raise ValueError(f"[LayerNorm] 'eps' must be non-negative but got {eps}.")
         self.dims = dims
 
     def _extra_repr(self):
@@ -147,10 +145,10 @@ class RMSNorm(Module):
 
     def __init__(self, dims: int, eps: float = 1e-5):
         super().__init__()
+        if eps <= 0.0:
+            raise ValueError(f"[RMSNorm] 'eps' must be positive but got {eps}.")
         self.weight = mx.ones((dims,))
         self.eps = eps
-        if eps < 0.0:
-            raise ValueError(f"[RMSNorm] 'eps' must be non-negative but got {eps}.")
 
     def _extra_repr(self):
         return f"{self.weight.shape[0]}, eps={self.eps}"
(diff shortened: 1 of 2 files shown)


python/mlx/nn/layers/normalization.py
```diff
@@ -52,6 +52,10 @@ def __init__(
             self.bias = mx.zeros((dims,))
         self.dims = dims
         self.eps = eps
+        if eps < 0.0:
```
zcbenz: It should probably reject 0 too, can you check how other frameworks do?

Also we throw error before setting members, you can see examples in GroupNorm.

ayaangazali: Both done, and the framework answer is not the one I expected, so here it is straight.

**Other frameworks do not validate this at all.** PyTorch accepts both on every norm layer I could construct:

```
torch LayerNorm    eps=0.0   ACCEPTED     eps=-1.0  ACCEPTED
torch GroupNorm    eps=0.0   ACCEPTED     eps=-1.0  ACCEPTED
torch BatchNorm1d  eps=0.0   ACCEPTED     eps=-1.0  ACCEPTED
torch RMSNorm      eps=0.0   ACCEPTED     eps=-1.0  ACCEPTED
```

and with `eps=0` on a zero variance input it returns NaN, same as mlx did:

```
torch LayerNorm eps=0.0    on ones(1,4) -> [nan nan nan nan]
mlx   LayerNorm eps=0.0    on ones(1,4) -> [nan nan nan nan]
```

Keras, TensorFlow, JAX and Flax are not installed here so I did not check those, I would rather say that than guess.

So framework precedent does not support rejecting zero. What does support it is mlx's own convention: `RMSprop`, `Adagrad` and `AdaDelta` all require `eps > 0`, and the reason recorded in `test_optimizers.test_epsilon_validation` is exactly this failure, that `eps == 0` lets a zero denominator produce NaN. Rejecting zero here makes the norm layers agree with the optimizers rather than with torch, and I think internal consistency is the better argument. So I have made it `eps <= 0.0` and the message now reads:

```
ValueError: [LayerNorm] 'eps' must be positive but got 0.0.
```

**Guards moved before the member assignments**, following `GroupNorm`, so nothing is set on a rejected object:

```python
super().__init__()
if eps <= 0.0:
    raise ValueError(f"[LayerNorm] 'eps' must be positive but got {eps}.")
if affine:
    self.weight = mx.ones((dims,))
```

All five were assigning `self.eps` before checking it, so that was worth fixing beyond style.

Test updated: `-1.0`, `-1e-30` and `0.0` are all rejected now, and `1e-30`, `1e-5` and `1.0` still construct, so a very small positive eps is still allowed. `test_nn.py`, `test_losses.py` and `test_optimizers.py` pass, and the test fails on the previous commit.

zcbenz (approved):
Thanks, can you also update #4310 with the same change?

PR #4316, opened 2026-08-17
Give a zero remainder the sign of the divisor

### LINKED ISSUE #4315: [BUG] mx.remainder gives a zero result a path-dependent sign
**Describe the bug**

When `mx.remainder` returns zero, the sign depends on which code path the element took. NumPy and Python give a zero remainder the sign of the divisor.

**To Reproduce**

Nine identical inputs, two answers:

```python
import mlx.core as mx
import numpy as np

r = mx.remainder(mx.full((9,), -0.0), mx.full((9,), 3.0))
print(np.signbit(np.array(r)).astype(int))
```

```
[0 0 0 0 0 0 0 0 1]
```

Lane 8 is the scalar residual and disagrees with lanes 0 to 7. It also depends on layout, a strided view disagreeing with a contiguous array of the same values from length 8 up.

Against numpy:

```python
a = mx.array([5.0, -5.0, 0.0])
b = mx.array([-5.0, 5.0, -1.0])
print(mx.remainder(a, b))                       # [0, -0, 0]
print(np.remainder(np.array(a), np.array(b)))   # [-0.  0. -0.]
```

Same in float16, bfloat16, float32 and float64, and through `%`.

**Expected behavior**

One answer per input, and a zero remainder taking the sign of the divisor.

**Desktop**
 - OS Version: macOS 26.5.2
 - Version: 0.32.1 (also on main at c2bcf47)

**Additional context**

The floored fixup skips zero:

```cpp
if (r != 0 && (r < 0 != b < 0)) {
  r += b;
}
```

so a zero keeps whatever sign the path produced, and `asd::remainder` and `std::remainder` do not agree on it. NumPy does `copysign(0, b)` here. The guard cannot just be dropped, since `r += b` on a zero returns `b` itself.

Same guard in all four backends. I have a fix with tests and will open a PR against this, happy to drop it if you would rather take it another way.

Found with https://github.com/krrishapatel/arraydiff.

krrishapatel:
## Proposed changes

Fixes #4315.

`mx.remainder` gives a zero result the sign of whichever code path produced it. All nine inputs here are `-0.0 % 3.0`:

```python
np.signbit(np.array(mx.remainder(mx.full((9,), -0.0), mx.full((9,), 3.0)))).astype(int)
# [0 0 0 0 0 0 0 0 1]
```

Lane 8 is the scalar residual. A strided view disagrees with a contiguous array the same way, and neither matches numpy, which gives a zero remainder the sign of the divisor.

The floored fixup skips zero:

```cpp
if (r != 0 && (r < 0 != b < 0)) {
  r += b;
}
```

so the zero keeps whatever sign it had, and `asd::remainder` and `std::remainder` do not agree on that. The guard cannot just be dropped, since adding `b` to a zero returns `b` itself, so the zero case sets the sign directly. Integers are untouched, they have no signed zero. Non-zero results are unchanged.

The same guard was in all four backends, so all four are fixed. CPU is verified locally on an M-series Mac. Metal and CUDA are the same edit to the same guard but I cannot run them here, so please treat those two as unverified.

`test_remainder_signed_zero` covers the residual lane, the strided view, both divisor signs against numpy, the `%` operator, and that integers are unaffected. It fails on `main` at the first assertion.

No overlap with #4108. That one floors the quotient for `divmod` and integer `floor_divide`; this is the sign of a zero remainder. They share two files but different structs, `FloorDivide` versus `Remainder`. I merged #4108 into this branch to check: no conflicts, both fixes active, `test_ops.py` green with both test sets.

## Checklist

### DIFF
--- a/mlx/backend/cpu/simd/accelerate_simd.h
+++ b/mlx/backend/cpu/simd/accelerate_simd.h
@@ -242,8 +242,17 @@ Simd<T, N> remainder(Simd<T, N> a, Simd<T, N> b) {
     r = a - b * (a / b);
   }
   if constexpr (is_signed_v<T>) {
-    auto mask = r != 0 && (r < 0 != b < 0);
+    auto nonzero = r != 0;
+    auto mask = nonzero && (r < 0 != b < 0);
     r = select(mask, r + b, r);
+    if constexpr (!std::is_integral_v<T>) {
+      // A zero remainder takes the sign of the divisor, which `r + b` cannot
+      // give it: adding to a zero would return `b` itself. See base_simd.h.
+      // Selecting on `b < 0` rather than multiplying by zero keeps an infinite
+      // divisor working, since `inf * 0` is NaN but `remainder(0, inf)` is 0.
+      Simd<T, N> zero = static_cast<T>(0);
+      r = select(nonzero, r, select(b < 0, -zero, zero));
+    }
   }
   return r;
 }
--- a/mlx/backend/cpu/simd/base_simd.h
+++ b/mlx/backend/cpu/simd/base_simd.h
@@ -214,8 +214,15 @@ Simd<T, 1> remainder(Simd<T, 1> a_, Simd<T, 1> b_) {
     r = std::remainder(a, b);
   }
   if constexpr (is_signed_v<T>) {
-    if (r != 0 && (r < 0 != b < 0)) {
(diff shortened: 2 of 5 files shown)


zcbenz:
Closing per https://github.com/ml-explore/mlx/issues/4315#issuecomment-5340893491.

